PR #26, opened 2021-09-10
[Pal/Linux-SGX] Pre-allocate more PAL memory on huge manifest files

dimakuv:
## Description of the changes 

The tomlc99 library to parse TOML files uses a lot of memory during `toml_parse()`. Sometimes apps have very large manifest files (>10MB), and parsing these manifest files requires more than 64MB of preallocated memory pool in Linux-SGX PAL.

Since this function is called before PAL can read `loader.pal_internal_mem_size`, there is no way to learn the additional memory size for PAL from the manifest. Thus we use a heuristic to pre-allocate some PAL memory before `toml_parse()`.

I understand that this fix is... dubious, but couldn't come up with something better.

Fixes https://github.com/gramineproject/graphene/issues/2680

## How to test this PR? 

Try with huge manifest files. E.g., I did this in our Bash example:
```
diff --git a/Examples/bash/manifest.template b/Examples/bash/manifest.template
+# loader.pal_internal_mem_size = "128M"
+sgx.trusted_files.bla = "file:/home/dimakuv/graphene"
```

This resulted in an 18MB `bash.manifest.sgx` file. Without this PR, Graphene fails at `toml_parse()` step:
```
$ graphene-sgx bash -c ls
error: *** Out-of-memory in PAL (try increasing `loader.pal_internal_mem_size`) ***
```

(Increasing `loader.pal_internal_mem_size` doesn't help because `toml_parse()` fails before being able to read this manifest option.)

With this PR and *without* `loader.pal_internal_mem_size = "128M"`, Graphene correctly complains:
```
$ graphene-sgx bash -c ls
error: Too small `loader.pal_internal_mem_size`, need at least 134217728 bytes
```

Finally, with this PR and with `loader.pal_internal_mem_size = "128M"`, Graphene works:
```
$ graphene-sgx bash -c ls
Makefile
README.md
bash.manifest
... and so on ...
```

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/26)

### DIFF
--- a/Pal/src/host/Linux-SGX/db_main.c
+++ b/Pal/src/host/Linux-SGX/db_main.c
@@ -678,6 +678,19 @@ noreturn void pal_linux_main(char* uptr_libpal_uri, size_t libpal_uri_len, char*
         ocall_exit(1, /*is_exitgroup=*/true);
     }
 
+    /* TOML parser (for whatever reason) allocates a lot of memory when parsing the manifest into an
+     * in-memory struct. We heuristically pre-allocate additional PAL internal memory if the
+     * manifest file looks large enough. Hopefully below sizes are sufficient for any manifest.
+     *
+     * FIXME: this is a quick hack, we need proper memory allocation in PAL. */
+    if (manifest_size > 10 * 1024 * 1024) {
+        log_always("Detected a huge manifest, preallocating 128MB of internal memory.");
+        g_pal_internal_mem_size += 128 * 1024 * 1024; /* 10MB manifest -> 64 + 128 MB PAL mem */
+    } else if (manifest_size > 5 * 1024 * 1024) {
+        log_always("Detected a huge manifest, preallocating 64MB of internal memory.");
+        g_pal_internal_mem_size += 64 * 1024 * 1024; /* 5MB manifest -> 64 + 64 MB PAL mem */
+    }
+
     /* parse manifest */
     char errbuf[256];
     toml_table_t* manifest_root = toml_parse(manifest_addr, errbuf, sizeof(errbuf));
@@ -700,12 +713,20 @@ noreturn void pal_linux_main(char* uptr_libpal_uri, size_t libpal_uri_len, char*
             READ_ONCE(*(size_t*)i);
     }
 
+    size_t pal_internal_mem_size;
     ret = toml_sizestring_in(g_pal_state.manifest_root, "loader.pal_internal_mem_size",
-                             /*defaultval=*/0, &g_pal_internal_mem_size);
+                             /*defaultval=*/0, &pal_internal_mem_size);
     if (ret < 0) {
         log_error("Cannot parse 'loader.pal_internal_mem_size'");
         ocall_exit(1, true);
     }
+
+    if (pal_internal_mem_size < g_pal_internal_mem_size) {
+        log_error("Too small `loader.pal_internal_mem_size`, need at least %luMB because the "
+                  "manifest is large", g_pal_internal_mem_size / 1024 / 1024);
+        ocall_exit(1, true);
+    }
+    g_pal_internal_mem_size = pal_internal_mem_size;
 
     if ((ret = init_file_check_policy()) < 0) {
         log_error("Failed to load the file check policy: %d", ret);


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/26)* status: 0 of 2 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[Pal/src/host/Linux-SGX/db_main.c, line 686 at r1](https://reviewable.io/reviews/gramineproject/gramine/26#-MjJ7rOZ0rnjn5c_zZ0e:-MjSnjIq4a8DKAsx6G0m:b-3duf42) ([raw file](https://github.com/gramineproject/gramine/blob/38f86b632f7a70deb85f31a28c68f25e8ff390db/Pal/src/host/Linux-SGX/db_main.c#L686)):*

`log_warning` about that? "This is a huge manifest, preallocating 128MB of internal memory"
</blockquote></details>

Done. Used `log_always` because `log_warning` will not be output (default log level is ERROR, and we didn't yet read the manifest options), and because `log_error` doesn't fit here logically.
___
*[Pal/src/host/Linux-SGX/db_main.c, line 723 at r1](https://reviewable.io/reviews/gramineproject/gramine/26#-MjJ6bv3E_-FmBaObPL6:-MjSo6Kx1uirM2nZbqqu:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/38f86b632f7a70deb85f31a28c68f25e8ff390db/Pal/src/host/Linux-SGX/db_main.c#L723)):*

This will be confusing for the user. How about something like "Too small `loader.pal_internal_mem_size`, need at least %lu bytes because the manifest is large"?

Also, maybe add `%lu bytes (%luMB)` since the user is likely to use megabytes in the manifest?
</blockquote></details>

Done.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/26)* status: 0 of 2 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/26#-MjTPIuIC_t_y8Bdxgcf:-MjTPIuIC_t_y8Bdxgcg:b-o7ij6z) (no related file):*
FYI: I accidentally commited my local changes to `bash/manifest.template` in one of the previous fixup commits. Removed it now.

---

jinengandhi-intel (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/26)* status: 0 of 2 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/26#-MjXxwWd6oT4Dxz53gxz:-MjXxwWhELHCHI_-sU3a:b2yjho2) (no related file):*
We went ahead and tested the PR where like Dmitrii mentioned it prompts the user to add the `loader.pal_internal_mem_size` relevant for the workload in the manifest file and the test passes post that.

---

pwmarcz (approved):
Sorry for the delay. I went through your testing procedure manually and it works for me.

> Reviewed 1 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/26#-:-MjZ-qg47XV1uPahZQ0f:begoba9)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

dimakuv (commented):
Can I request a review from @boryspoplawski or @mkow ? I wonder if they maybe come up with something better than this hack...

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/26#-:-MjZM0AL4lyaCek201aE:bpun5wl)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @jinengandhi-intel)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/26#-MjXxwWd6oT4Dxz53gxz:-MjZLypJCJfhsWI9kaPl:b-p2kyme) (no related file):*
<details><summary><i>Previously, jinengandhi-intel wrote…</i></summary><blockquote>

We went ahead and tested the PR where like Dmitrii mentioned it prompts the user to add the `loader.pal_internal_mem_size` relevant for the workload in the manifest file and the test passes post that.
</blockquote></details>

Thanks for this!

---

boryspoplawski (changes requested):
I guess you meant `18MB` not `GB` :)

> Reviewed 1 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/26#-:-Mj_GQIF2e2xTNYtJ14T:b50g7zs)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @jinengandhi-intel)

___
*[Pal/src/host/Linux-SGX/db_main.c, line 724 at r3](https://reviewable.io/reviews/gramineproject/gramine/26#-Mj_w3abDlg368CQAoH_:-Mj_w3ac1tA2gP4Xv3m0:b-ame75i) ([raw file](https://github.com/gramineproject/gramine/blob/fbbd9afdc4f5959055389ebf1320e3121a16aa0a/Pal/src/host/Linux-SGX/db_main.c#L724)):*
> ```C
>     }
> 
>     if (pal_internal_mem_size < g_pal_internal_mem_size) {
> ```

So if we do not specify `loader.pal_internal_mem_size` (default is `0`) and the manifest is big, this will trigger and exit. Is that intended?
___
*[Pal/src/host/Linux-SGX/db_main.c, line 729 at r3](https://reviewable.io/reviews/gramineproject/gramine/26#-Mj_wWp23QjPb3TXeKGi:-Mj_wWp23QjPb3TXeKGj:b-sdz523) ([raw file](https://github.com/gramineproject/gramine/blob/fbbd9afdc4f5959055389ebf1320e3121a16aa0a/Pal/src/host/Linux-SGX/db_main.c#L729)):*
> ```C
>         ocall_exit(1, true);
>     }
>     g_pal_internal_mem_size = pal_internal_mem_size;
> ```

How does it work? In normal/default case this will be `0` (just like in the old version).

boryspoplawski (commented):
This doesn't look that bad. A bit hacky, but I don't see a better way, without major rework of manifest handling.

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/26#-:-Mj_wnIh2TsM9BvZLXaX:b-5rinoh)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @jinengandhi-intel)

dimakuv (commented):
Fixed the description in the PR. Sure it is `MB`.

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/26#-:-Mj_zzx_7lmgdGCrIIqP:b-houu3z)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @jinengandhi-intel)

___
*[Pal/src/host/Linux-SGX/db_main.c, line 724 at r3](https://reviewable.io/reviews/gramineproject/gramine/26#-Mj_w3abDlg368CQAoH_:-Mja-4Ui0y9GIOk8QXuB:bbyj9d4) ([raw file](https://github.com/gramineproject/gramine/blob/fbbd9afdc4f5959055389ebf1320e3121a16aa0a/Pal/src/host/Linux-SGX/db_main.c#L724)):*

So if we do not specify `loader.pal_internal_mem_size` (default is `0`) and the manifest is big, this will trigger and exit. Is that intended?
</blockquote></details>

This will not trigger. Because `loader.pal_internal_mem_size == g_pal_internal_mem_size == 0`, and this IF clause is a strict `<`
___
*[Pal/src/host/Linux-SGX/db_main.c, line 729 at r3](https://reviewable.io/reviews/gramineproject/gramine/26#-Mj_wWp23QjPb3TXeKGi:-Mja-BbYA0bWRuZeajv_:bpg4ya8) ([raw file](https://github.com/gramineproject/gramine/blob/fbbd9afdc4f5959055389ebf1320e3121a16aa0a/Pal/src/host/Linux-SGX/db_main.c#L729)):*

How does it work? In normal/default case this will be `0` (just like in the old version).
</blockquote></details>

That was exactly my intention -- for normal cases (not-too-big manifest), the behavior is like it was before.

Is your question -- how do we have any PAL internal memory if `g_pal_internal_mem_size == 0`? This is because this `g_pal_internal_mem_size` variable denotes the *dynamically allocated PAL memory*. And we always have the additional *statically allocated 64MB of memory in the form of a BSS global variable `g_mem_pool`.

So the default behavior (when `loader.pal_internal_mem_size == 0`) is to only use 64MB of `g_mem_pool`. This is even described in our ReadTheDocs manifest-syntax page.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/26)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @jinengandhi-intel)

___
*[Pal/src/host/Linux-SGX/db_main.c, line 724 at r3](https://reviewable.io/reviews/gramineproject/gramine/26#-Mj_w3abDlg368CQAoH_:-MjaBD1q4sB2YWy9D4YY:b-xsc631) ([raw file](https://github.com/gramineproject/gramine/blob/fbbd9afdc4f5959055389ebf1320e3121a16aa0a/Pal/src/host/Linux-SGX/db_main.c#L724)):*
> and the manifest is big

By this I meant `g_pal_internal_mem_size > 0` (and default `loader.pal_internal_mem_size == 0` )
___
*[Pal/src/host/Linux-SGX/db_main.c, line 729 at r3](https://reviewable.io/reviews/gramineproject/gramine/26#-Mj_wWp23QjPb3TXeKGi:-MjaBX0zEWksWYy8YsgW:bbbvu4w) ([raw file](https://github.com/gramineproject/gramine/blob/fbbd9afdc4f5959055389ebf1320e3121a16aa0a/Pal/src/host/Linux-SGX/db_main.c#L729)):*

That was exactly my intention -- for normal cases (not-too-big manifest), the behavior is like it was before.

Is your question -- how do we have any PAL internal memory if `g_pal_internal_mem_size == 0`? This is because this `g_pal_internal_mem_size` variable denotes the *dynamically allocated PAL memory*. And we always have the additional *statically allocated 64MB of memory in the form of a BSS global variable `g_mem_pool`.

So the default behavior (when `loader.pal_internal_mem_size == 0`) is to only use 64MB of `g_mem_pool`. This is even described in our ReadTheDocs manifest-syntax page.
</blockquote></details>

Thanks for explanation

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/26)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @jinengandhi-intel)

___
*[Pal/src/host/Linux-SGX/db_main.c, line 724 at r3](https://reviewable.io/reviews/gramineproject/gramine/26#-Mj_w3abDlg368CQAoH_:-Mjc6yDo3Mb0-Pjh1Vab:b-ba49lv) ([raw file](https://github.com/gramineproject/gramine/blob/fbbd9afdc4f5959055389ebf1320e3121a16aa0a/Pal/src/host/Linux-SGX/db_main.c#L724)):*

> and the manifest is big

By this I meant `g_pal_internal_mem_size > 0` (and default `loader.pal_internal_mem_size == 0` )
</blockquote></details>

Ah, sorry, misunderstood the first time.

Yes, this is intended -- your scenario would mean that the user fed a big manifest to Gramine, and Gramine decided to allocate 64/128 MB of additional PAL memory. Since this "allocation of additional PAL memory" is controlled by the `loader.pal_internal_mem_size` manifest option, Gramine cannot just silently continue working. Otherwise what does it mean? The user will think that Gramine is not using additional PAL memory, and if the user will decide later at some point to set `loader.pal_internal_mem_size = 32M`, how should Gramine react?

To remove this ambiguity, I added this check and I force the user to specify `loader.pal_internal_mem_size` explicitly.

boryspoplawski (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/26)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @jinengandhi-intel)

___
*[Pal/src/host/Linux-SGX/db_main.c, line 724 at r3](https://reviewable.io/reviews/gramineproject/gramine/26#-Mj_w3abDlg368CQAoH_:-Mjdr65q7NIVckjhuxmv:b-hfghm4) ([raw file](https://github.com/gramineproject/gramine/blob/fbbd9afdc4f5959055389ebf1320e3121a16aa0a/Pal/src/host/Linux-SGX/db_main.c#L724)):*

Ah, sorry, misunderstood the first time.

Yes, this is intended -- your scenario would mean that the user fed a big manifest to Gramine, and Gramine decided to allocate 64/128 MB of additional PAL memory. Since this "allocation of additional PAL memory" is controlled by the `loader.pal_internal_mem_size` manifest option, Gramine cannot just silently continue working. Otherwise what does it mean? The user will think that Gramine is not using additional PAL memory, and if the user will decide later at some point to set `loader.pal_internal_mem_size = 32M`, how should Gramine react?

To remove this ambiguity, I added this check and I force the user to specify `loader.pal_internal_mem_size` explicitly.
</blockquote></details>

Sounds good

pwmarcz (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/26)* status: all files reviewed, 1 unresolved discussion (waiting on @jinengandhi-intel)

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/26)* status: all files reviewed, 1 unresolved discussion (waiting on @jinengandhi-intel)

mkow (approved):
> Reviewed 1 of 2 files at r2, 1 of 1 files at r4, all commit messages.
> Dismissed @jinengandhi-intel from a discussion.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/26)* status: :shipit: complete! all files reviewed, all discussions resolved

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/26#-MjXxwWd6oT4Dxz53gxz:-MjiK_wV2UooOK-7epWm:bbs6z8t) (no related file):*

Thanks for this!
</blockquote></details>

@jinengandhi-intel: You put a blocking comment, but from its contents it seems that wasn't intended, so I'm dismissing it.

---

Issue #28, opened 2021-09-13
Stress-ng open syscall test failing with assert fd < handle_map->fd_size on RHEL and CentOS


jinengandhi-intel:
## Description of the problem

On RHEL and CentOS, stress-ng open syscall test is failing with the following errors:

`[P1:shim] error: Child process (vmid: 0x84) got disconnected
[P124:T124:stress-ng] assert failed ../LibOS/shim/src/bookkeep/shim_handle.c:419 fd < handle_map->fd_size`

Complete log file is attached below: 
[stress-ng_open_rhel.txt](https://github.com/gramineproject/gramine/files/7151263/stress-ng_open_rhel.txt)

## Steps to reproduce

`graphene-direct stress-ng --open 0 --timeout 60s --verbose`

dimakuv:
I debugged it, and the reason is that this particular workload creates **a lot of** file descriptors (up to 100,000). This seems to be normal for `stress-ng` application.

However, in Gramine we have this: https://github.com/gramineproject/gramine/blob/40e942db6b02555cf2414f20bbd313b63f50e400/LibOS/shim/include/shim_types.h#L341

Because `FDTYPE` is 16-bit in size, it can only count 65536 FDs. This leads to an implicit overflow and truncation here: https://github.com/gramineproject/gramine/blob/40e942db6b02555cf2414f20bbd313b63f50e400/LibOS/shim/src/bookkeep/shim_handle.c#L576

Note that `fd_size` is of type `FDTYPE`: https://github.com/gramineproject/gramine/blob/74420be3a01d82415b36f142f5e52c87385befdd/LibOS/shim/include/shim_handle.h#L270

And that's how we bump into this assert: https://github.com/gramineproject/gramine/blob/40e942db6b02555cf2414f20bbd313b63f50e400/LibOS/shim/src/bookkeep/shim_handle.c#L419

(`handle_map->fd_size` was truncated to the max 16-bit value, so it is less that 32-bit `fd` value.)

**Proposed solution**: there is no reason to keep `FDTYPE` limited to 16 bits. I will change it to `uint32_t` and I'll remove `FDTYPE` user-defined type and will replace with explicit `uint32_t` everywhere in Gramine codebase. (Actually, we should do the same with `IDTYPE` but at least this one doesn't break anything.)

jinengandhi-intel:
Dmitrii, any reason why this is triggered only on Cent and not on Ubuntu?

dimakuv:
[...]

PR #41, opened 2021-09-14
[LibOS] Get rid of `FDTYPE` and replace it with `uint32_t`

dimakuv:
## Description of the changes 

Previously, LibOS used `typedef uint16_t FDTYPE`. This was unnecessary and hard-to-read. Moreover, it led to a bug in `__enlarge_handle_map()`: the new size was of 64-bit type `size_t` but it was assigned to `fd_size` field of 16-bit type `FDTYPE`, which led to silent truncation. This resulted in failing asserts, because active FDs suddenly became greater than the limit in the FD map.

This commit fixes this bug by changing the type from 16-bit to 32-bit, and also removes the opaque and confusing `FDTYPE`.

Fixes https://github.com/gramineproject/gramine/issues/28

## How to test this PR? 

Current tests should be sufficient. I don't think we want a regression test that opens 64 thousand FDs.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/41)

### DIFF
--- a/LibOS/shim/include/shim_handle.h
+++ b/LibOS/shim/include/shim_handle.h
@@ -162,7 +162,7 @@ struct shim_tmpfs_handle {
 DEFINE_LIST(shim_epoll_item);
 DEFINE_LISTP(shim_epoll_item);
 struct shim_epoll_item {
-    FDTYPE fd;
+    int fd;
     uint64_t data;
     unsigned int events;
     unsigned int revents;
@@ -259,16 +259,16 @@ int set_handle_nonblocking(struct shim_handle* hdl, bool on);
 
 /* file descriptor table */
 struct shim_fd_handle {
-    FDTYPE vfd; /* virtual file descriptor */
-    int flags;  /* file descriptor flags, only FD_CLOEXEC */
+    uint32_t vfd; /* virtual file descriptor */
+    int flags;    /* file descriptor flags, only FD_CLOEXEC */
 
     struct shim_handle* handle;
 };
 
 struct shim_handle_map {
     /* the top of created file descriptors */
-    FDTYPE fd_size;
-    FDTYPE fd_top;
+    uint32_t fd_size;
+    uint32_t fd_top;
 
     /* refrence count and lock */
     REFTYPE ref_count;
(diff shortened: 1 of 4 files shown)


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/41)* status: 0 of 4 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners

___
*[LibOS/shim/include/shim_handle.h, line 165 at r1](https://reviewable.io/reviews/gramineproject/gramine/41#-MjZn3iu6qdXgCRaToTM:-MjZn3iu6qdXgCRaToTN:b99kxy3) ([raw file](https://github.com/gramineproject/gramine/blob/20ada53d31011a7faf313c97f059f719d2d4e758/LibOS/shim/include/shim_handle.h#L165)):*
> ```objc
> DEFINE_LISTP(shim_epoll_item);
> struct shim_epoll_item {
>     int fd;
> ```

FYI: I decided to have this `epoll_item->fd` as a signed integer because this is how it is used in `shim_epoll.c` functions. This is also how Linux uses `fd` argument in `epoll_ctl()`, `epoll_wait()`, etc. So it looked easiest to just use signed int (otherwise I would need some type conversions).

pwmarcz (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/41)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

boryspoplawski (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/41)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

pwmarcz (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/41)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/41)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #56, opened 2021-09-15
Add `loader.env.[ENVIRON] = {passthrough=true}` manifest option

dimakuv:
## Description of the changes 

It is now possible to define some envvars that are passed through from the host inside the Graphene instance (previously the only way to pass envvars was to set `loader.insecure__use_host_env`).

This was originally developed as https://github.com/gramineproject/graphene/pull/2641. Based on an offline discussion with @mkow, I simplified the code as much as possible (it overapproximates some things, but we don't care about a handful of envvars.)

Fixes https://github.com/gramineproject/graphene/issues/2356

## How to test this PR? 

A new LibOS regression test is added + old tests should be sufficient.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/56)

### DIFF
--- a/Documentation/manifest-syntax.rst
+++ b/Documentation/manifest-syntax.rst
@@ -156,10 +156,20 @@ both of the following options:
 ::
 
    loader.env.[ENVIRON] = "[VALUE]"
+   or
+   loader.env.[ENVIRON] = { value = "[VALUE]" }
+   or
+   loader.env.[ENVIRON] = { passthrough = true }
+
    loader.env_src_file = "file:file_with_serialized_envs"
 
-``loader.env.[ENVIRON]`` adds/overwrites a single environment variable and can
-be used multiple times to specify more than one variable.
+``loader.env.[ENVIRON]`` adds/overwrites/passes a single environment variable
+and can be used multiple times to specify more than one variable. To
+add/overwrite the environment variable, specify a TOML string (``"[VALUE]"``) or
+a TOML table with the key-value pair ``{ value = "[VALUE]" }``. To pass the
+environment variable from the host, specify a TOML table with the key-value pair
+``{ passthrough = true }``. If you specify a variable, it needs to either have a
+value or be a passthrough.
 
 ``loader.env_src_file`` allows to specify a URI to a file containing serialized
 environment, which can be generated using :program:`gramine-argv-serializer`.
@@ -174,7 +184,17 @@ provided at runtime from an external (trusted) source.
    variables.
 
 If the same variable is set in both, then ``loader.env.[ENVIRON]`` takes
-precedence.
+precedence. It is prohibited to specify both ``value`` and ``passthrough`` keys
+for the same environment variable. If manifest option ``insecure__use_host_env``
+is specified, then ``passthrough = true`` manifest options have no effect (they
+are "consumed" by ``insecure__use_host_env``).
+
+.. note ::
+   It is tempting to try to passthrough all environment variables using
+   ``insecure__use_host_env`` and then disallow some of them using ``passthrough
+   = false``. However, this deny list approach is intentionally prohibited
+   because it's inherently insecure (doesn't provide any real security).
+   Gramine loudly fails if ``passthrough = false`` manifest options are set.
 
 Disabling ASLR
 ^^^^^^^^^^^^^^
@@ -308,7 +328,7 @@ Gramine currently supports two types of mount points:
 * ``chroot``: Host-backed files. All host files and sub-directories found under
   ``[URI]`` are forwarded to the Gramine instance and placed under ``[PATH]``.
   For example, with a host-level path specified as
-  ``fs.mount.lib.uri = "file:/one/path/"`` and forwarded to Graphene via
+  ``fs.mount.lib.uri = "file:/one/path/"`` and forwarded to Gramine via
   ``fs.mount.lib.path = "/another/path"``, a host-level file
   ``/one/path/file`` is visible to graphenized application as
   ``/another/path/file``. This concept is similar to FreeBSD's chroot and to
--- a/LibOS/shim/test/regression/.gitignore
+++ b/LibOS/shim/test/regression/.gitignore
@@ -25,6 +25,7 @@
 /double_fork
 /env_from_file
 /env_from_host
+/env_passthrough
 /epoll_epollet
 /epoll_wait_timeout
 /eventfd
--- a/LibOS/shim/test/regression/Makefile
+++ b/LibOS/shim/test/regression/Makefile
@@ -107,6 +107,7 @@ repo_manifests = \
 	device_passthrough.manifest \
 	env_from_file.manifest \
 	env_from_host.manifest \
+	env_passthrough.manifest \
 	file_check_policy_allow_all_but_log.manifest \
 	file_check_policy_strict.manifest \
 	multi_pthread_exitless.manifest
--- a/LibOS/shim/test/regression/env_passthrough.manifest.template
+++ b/LibOS/shim/test/regression/env_passthrough.manifest.template
@@ -1,0 +1,22 @@
+loader.preload = "file:{{ gramine.libos }}"
+loader.argv0_override = "bootstrap"
+libos.entrypoint = "bootstrap"
+
+loader.env.LD_LIBRARY_PATH = "/lib"
+
+loader.env.A = { passthrough = true }
+loader.env.B = { value = "OVERWRITTEN_VALUE" }
+
+# loader.env.C = { passthrough = false }  # not allowed for security reasons
+# loader.env.E = { passthrough = true, value = "THIS_IS_INCORRECT_SYNTAX" }
+
+fs.mount.lib.type = "chroot"
+fs.mount.lib.path = "/lib"
+fs.mount.lib.uri = "file:{{ gramine.runtimedir() }}"
+
+sgx.nonpie_binary = true
+
+sgx.trusted_files = [
+  "file:{{ gramine.runtimedir() }}/",
+  "file:bootstrap",
+]
--- a/LibOS/shim/test/regression/test_libos.py
+++ b/LibOS/shim/test/regression/test_libos.py
@@ -112,15 +112,33 @@ class TC_01_Bootstrap(RegressionTestCase):
         finally:
             os.remove('env_test_input')
 
+    def test_105_env_passthrough(self):
+        host_envs = {
+            'A': 'THIS_WILL_BE_PASSED',
+            'B': 'THIS_WILL_BE_OVERWRITTEN',
+            'C': 'THIS_SHOULDNT_BE_PASSED',
+            'D': 'THIS_SHOULDNT_BE_PASSED_TOO',
+        }
+        manifest_envs = {'LD_LIBRARY_PATH': '/lib'}
+        stdout, _ = self.run_binary(['env_passthrough'], env=host_envs)
+        self.assertIn('# of envs: %d\n' % (len(host_envs) - 2 + len(manifest_envs)), stdout)
+
+        # We don't enforce any specific order of envs, so we skip checking the index.
+        self.assertIn('] = LD_LIBRARY_PATH=/lib\n', stdout)
+        self.assertIn('] = A=THIS_WILL_BE_PASSED\n', stdout)
+        self.assertIn('] = B=OVERWRITTEN_VALUE\n', stdout)
+        self.assertNotIn('] = C=THIS_SHOULDNT_BE_PASSED\n', stdout)
+        self.assertNotIn('] = D=THIS_SHOULDNT_BE_PASSED_TOO\n', stdout)
+
     @unittest.skipUnless(HAS_SGX,
         'This test is only meaningful on SGX PAL because only SGX catches raw '
         'syscalls and redirects to Gramine\'s LibOS. If we will add seccomp to '
         'Linux PAL, then we should allow this test on Linux PAL as well.')
-    def test_105_basic_bootstrapping_static(self):
+    def test_106_basic_bootstrapping_static(self):
         stdout, _ = self.run_binary(['bootstrap_static'])
         self.assertIn('Hello world (bootstrap_static)!', stdout)
 
-    def test_106_basic_bootstrapping_pie(self):
+    def test_107_basic_bootstrapping_pie(self):
         stdout, _ = self.run_binary(['bootstrap_pie'])
         self.assertIn('User program started', stdout)
         self.assertIn('Local Address in Executable: 0x', stdout)
--- a/Pal/src/db_main.c
+++ b/Pal/src/db_main.c
@@ -63,91 +63,227 @@ static void load_libraries(void) {
     }
 }
 
+/* Finds `loader.env.{key}` in the manifest and returns its value in `out_val`/`out_passthrough`.
+ * Recognizes several formats:
+ *   - `loader.env.{key} = "val"`
+ *   - `loader.env.{key} = { value = "val" }`
+ *   - `loader.env.{key} = { passthrough = true|false }`
+ *
+ * If `loader.env.{key}` doesn't exist, then returns 0 and `*out_exists = false`. If the key exists,
+ * then `*out_exists = true` and:
+ *   - if passthrough is specified, then `*out_val` is NULL and `*out_passthrough` is true/false;
+ *   - if the value is specified, then `*out_val` is set and `*out_passthrough` is false.
+ *
+ * For any other format, returns negative error code. The caller is responsible to free `out_val`.
+ */
+static int get_env_value_from_manifest(toml_table_t* toml_envs, const char* key, bool* out_exists,
+                                       char** out_val, bool* out_passthrough) {
+    int ret;
+
+    assert(out_exists);
+    assert(out_val);
+    assert(out_passthrough);
+
+    toml_table_t* toml_env_table = toml_table_in(toml_envs, key);
+    if (!toml_env_table) {
+        /* not table like `loader.env.key = {...}`, must be string like `loader.env.key = "val"` */
+        toml_raw_t toml_env_val_raw = toml_raw_in(toml_envs, key);
+        if (!toml_env_val_raw) {
+            *out_val = NULL;
+            *out_passthrough = false;
+            *out_exists = false;
+            return 0;
+        }
+
+        ret = toml_rtos(toml_env_val_raw, out_val);
+        if (ret < 0)
+            return -PAL_ERROR_NOMEM;
+
+        *out_passthrough = false;
+        *out_exists = true;
+        return 0;
+    }
+
+    toml_raw_t toml_passthrough_raw = toml_raw_in(toml_env_table, "passthrough");
+    toml_raw_t toml_value_raw = toml_raw_in(toml_env_table, "value");
+
+    if (toml_passthrough_raw && toml_value_raw) {
+        /* disallow `loader.env.key = { passthrough = true, value = "val" }` */
+        return -PAL_ERROR_INVAL;
+    }
+
+    if (!toml_passthrough_raw) {
+        /* not passthrough like `loader.env.key = { passthrough = true }`, must be value-table like
+         * `loader.env.key = { value = "val" }` */
+        if (!toml_value_raw)
+            return -PAL_ERROR_INVAL;
+
+        ret = toml_rtos(toml_value_raw, out_val);
+        if (ret < 0)
+            return -PAL_ERROR_NOMEM;
+
+        *out_passthrough = false;
+        *out_exists = true;
+        return 0;
+    }
+
+    /* at this point, it must be `loader.env.key = { passthrough = true|false }`*/
+    int passthrough_int;
+    ret = toml_rtob(toml_passthrough_raw, &passthrough_int);
+    if (ret < 0)
+        return -PAL_ERROR_INVAL;
+
+    *out_passthrough = !!passthrough_int;
+    *out_val = NULL;
+    *out_exists = true;
+    return 0;
+}
+
+static int create_empty_envs(const char*** out_envp) {
+    const char** new_envp = malloc(sizeof(*new_envp));
+    if (!new_envp)
+        return -PAL_ERROR_NOMEM;
+
+    new_envp[0] = NULL;
+    *out_envp = new_envp;
+    return 0;
+}
+
 /* This function leaks memory on failure (and this is non-trivial to fix), but the assumption is
  * that its failure finishes the execution of the whole process right away. */
-static int insert_envs_from_manifest(const char*** envpp) {
+static int deep_copy_envs(const char** envp, const char*** out_envp) {
+    size_t orig_envs_cnt = 0;
+    for (const char** orig_env = envp; *orig_env; orig_env++)
+        orig_envs_cnt++;
+
+    const char** new_envp = calloc(orig_envs_cnt + 1, sizeof(const char*));
+    if (!new_envp)
+        return -PAL_ERROR_NOMEM;
+
+    size_t idx = 0;
+    for (const char** orig_env = envp; *orig_env; orig_env++) {
+        new_envp[idx] = strdup(*orig_env);
+        if (!new_envp[idx])
+            return -PAL_ERROR_NOMEM;
+        idx++;
+    }
+    assert(idx == orig_envs_cnt);
+
+    *out_envp = new_envp;
+    return 0;
+}
+
+/* Build environment for Gramine process, based on original environment (from host or file) and
+ * manifest. If `propagate` is true, copies all variables from `orig_envp`, except the ones
+ * overriden in manifest. Otherwise, copies only the variables from `orig_envp` that the manifest
+ * specifies as passthrough.
+ *
+ * This function leaks memory on failure (and this is non-trivial to fix), but the assumption is
+ * that its failure finishes the execution of the whole process right away. */
+static int build_envs(const char** orig_envp, bool propagate, const char*** out_envp) {
     int ret;
-    assert(envpp);
 
     toml_table_t* toml_loader = toml_table_in(g_pal_state.manifest_root, "loader");
     if (!toml_loader)
-        return 0;
+        return propagate ? deep_copy_envs(orig_envp, out_envp) : create_empty_envs(out_envp);
 
     toml_table_t* toml_envs = toml_table_in(toml_loader, "env");
     if (!toml_envs)
-        return 0;
+        return propagate ? deep_copy_envs(orig_envp, out_envp) : create_empty_envs(out_envp);
 
     ssize_t toml_envs_cnt = toml_table_nkval(toml_envs);
-    if (toml_envs_cnt <= 0) {
-        /* no env entries found in the manifest */
-        return 0;
-    }
+    if (toml_envs_cnt < 0)
+        return -PAL_ERROR_INVAL;
+
+    ssize_t toml_env_tables_cnt = toml_table_ntab(toml_envs);
+    if (toml_env_tables_cnt < 0)
+        return -PAL_ERROR_INVAL;
+
+    toml_envs_cnt += toml_env_tables_cnt;
+    if (toml_envs_cnt == 0)
+        return propagate ? deep_copy_envs(orig_envp, out_envp) : create_empty_envs(out_envp);
 
     size_t orig_envs_cnt = 0;
-    size_t overwrite_cnt = 0;
-    for (const char** orig_env = *envpp; *orig_env; orig_env++, orig_envs_cnt++) {
-        char* orig_env_key_end = strchr(*orig_env, '=');
-        if (!orig_env_key_end)
-            return -PAL_ERROR_INVAL;
+    for (const char** orig_env = orig_envp; *orig_env; orig_env++)
+        orig_envs_cnt++;
 
-        *orig_env_key_end = '\0';
-        toml_raw_t toml_env_raw = toml_raw_in(toml_envs, *orig_env);
-        if (toml_env_raw) {
-            /* found the original-env key in manifest (i.e., loader.env.<key> exists) */
-            overwrite_cnt++;
-        }
-        *orig_env_key_end = '=';
-    }
-
-    size_t total_envs_cnt = orig_envs_cnt + toml_envs_cnt - overwrite_cnt;
+    /* For simplicity, overapproximate the number of envs by summing up the ones found in the
+     * original environment and the ones found in the manifest. */
+    size_t total_envs_cnt = orig_envs_cnt + toml_envs_cnt;
     const char** new_envp = calloc(total_envs_cnt + 1, sizeof(const char*));
     if (!new_envp)
         return -PAL_ERROR_NOMEM;
 
-    /* For simplicity, allocate each env anew; this is suboptimal but happens only once. First
-     * go through original envs and populate new_envp with only those that are not overwritten by
-     * manifest envs. Then append all manifest envs to new_envp. */
     size_t idx = 0;
-    for (const char** orig_env = *envpp; *orig_env; orig_env++) {
-        char* orig_env_key_end = strchr(*orig_env, '=');
 
-        *orig_env_key_end = '\0';
-        toml_raw_t toml_env_raw = toml_raw_in(toml_envs, *orig_env);
-        if (!toml_env_raw) {
-            /* this original env is not found in manifest (i.e., not overwritten) */
-            *orig_env_key_end = '=';
-            new_envp[idx] = malloc_copy(*orig_env, strlen(*orig_env) + 1);
-            if (!new_envp[idx]) {
-                /* don't care about proper memory cleanup since will terminate anyway */
+    /* First, go through original variables and copy the ones that we're going to use (because of
+     * `propagate`, or because passthrough is specified for that variable in manifest). */
+    for (const char** orig_env = orig_envp; *orig_env; orig_env++) {
+        char* orig_env_key_end = strchr(*orig_env, '=');
+        if (!orig_env_key_end)
+            return -PAL_ERROR_INVAL;
+
+        char* env_key = alloc_substr(*orig_env, orig_env_key_end - *orig_env);
+        if (!env_key)
+            return -PAL_ERROR_NOMEM;
+
+        bool exists;
+        char* env_val;
+        bool passthrough;
+        ret = get_env_value_from_manifest(toml_envs, env_key, &exists, &env_val, &passthrough);
+        if (ret < 0) {
+            log_error("Invalid environment variable in manifest: '%s'", env_key);
+            return ret;
+        }
+
+        if ((propagate && !exists) || (exists && passthrough)) {
+            new_envp[idx] = strdup(*orig_env);
+            if (!new_envp[idx])
                 return -PAL_ERROR_NOMEM;
-            }
             idx++;
         }
-        *orig_env_key_end = '=';
-    }
-    assert(idx < total_envs_cnt);
 
+        free(env_key);
+        free(env_val);
+    }
+
+    /* Then, go through the manifest variables and copy the ones with value provided. */
     for (ssize_t i = 0; i < toml_envs_cnt; i++) {
         const char* toml_env_key = toml_key_in(toml_envs, i);
         assert(toml_env_key);
-        toml_raw_t toml_env_value_raw = toml_raw_in(toml_envs, toml_env_key);
-        assert(toml_env_value_raw);
 
-        char* toml_env_value = NULL;
-        ret = toml_rtos(toml_env_value_raw, &toml_env_value);
+        bool exists;
+        char* env_val;
+        bool passthrough;
+        ret = get_env_value_from_manifest(toml_envs, toml_env_key, &exists, &env_val, &passthrough);
         if (ret < 0) {
-            /* don't care about proper memory cleanup since will terminate anyway */
-            return -PAL_ERROR_NOMEM;
+            log_error("Invalid environment variable in manifest: '%s'", toml_env_key);
+            return ret;
         }
 
-        char* final_env = alloc_concat3(toml_env_key, strlen(toml_env_key), "=", 1, toml_env_value,
-                                        strlen(toml_env_value));
-        new_envp[idx++] = final_env;
-        free(toml_env_value);
-    }
-    assert(idx == total_envs_cnt);
+        assert(exists);
+        if (!env_val && !passthrough) {
+            log_error("Detected environment variable '%s' with `passthrough = false`. For security"
+                      " reasons, Gramine doesn't allow this. Please see documentation on"
+                      " 'loader.env' for details.", toml_env_key);
+            return -PAL_ERROR_DENIED;
+        }
 
-    *envpp = new_envp;
+        if (!env_val) {
+            /* this is envvar with passthrough = true, we accounted for it in previous loop */
+            continue;
+        }
+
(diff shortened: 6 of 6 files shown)


pwmarcz (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/56)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[Documentation/manifest-syntax.rst, line 162 at r1](https://reviewable.io/reviews/gramineproject/gramine/56#-MjirsxG7ioBquQaQeyC:-MjirsxG7ioBquQaQeyD:b-mnffyc) ([raw file](https://github.com/gramineproject/gramine/blob/6ac5a70ae00fae8e0dac42d3dad358a2e3c93a1f/Documentation/manifest-syntax.rst#L162)):*
> ```reStructuredText
>    loader.env.[ENVIRON] = { value = "[VALUE]" }
>    or
>    loader.env.[ENVIRON] = { passthrough = [true|false] }
> ```

Why do we even allow `{ passthrough = false }`? As far as I can tell, it's either disallowed, or has no effect.

(And if we disallow `passthrough = false`, the "forbidden combination" can be detected directly in `get_env_value_from_manifest`.)
___
*[Pal/src/db_main.c, line 146 at r1](https://reviewable.io/reviews/gramineproject/gramine/56#-Mjipgbu-MveI8JCQjU7:-Mjipgbu-MveI8JCQjU8:blgns2l) ([raw file](https://github.com/gramineproject/gramine/blob/6ac5a70ae00fae8e0dac42d3dad358a2e3c93a1f/Pal/src/db_main.c#L146)):*
> ```C
>  * that its failure finishes the execution of the whole process right away. */
> static int deep_copy_envs(bool propagate, const char** envp, const char*** out_envp) {
>     if (!propagate) {
> ```

`propagate=false` essentially converts this function into one creating an empty environment, right?

It's very confusing, I'd prefer to have a separate function, even if it complicates the call sites:

```c
// instead of:
return deep_copy_envs(propagate, envp, out_envp);

// write:
return propagate ? deep_copy_envs(envp, out_envp) : create_empty_envs(out_envp);
```
___
*[Pal/src/db_main.c, line 149 at r1](https://reviewable.io/reviews/gramineproject/gramine/56#-MjioItlBU6vUxNvoz9r:-MjioItlBU6vUxNvoz9s:bexw49y) ([raw file](https://github.com/gramineproject/gramine/blob/6ac5a70ae00fae8e0dac42d3dad358a2e3c93a1f/Pal/src/db_main.c#L149)):*
> ```C
>         const char** new_envp = malloc(sizeof(*new_envp));
>         if (!new_envp)
> 			return -PAL_ERROR_NOMEM;
> ```

You have tabs instead of spaces here.
___
*[Pal/src/db_main.c, line 152 at r1](https://reviewable.io/reviews/gramineproject/gramine/56#-MjioMX00YlQ4c8t5fnb:-MjioMX15pxtp4Gp6jAm:bexw49y) ([raw file](https://github.com/gramineproject/gramine/blob/6ac5a70ae00fae8e0dac42d3dad358a2e3c93a1f/Pal/src/db_main.c#L152)):*
> ```C
>         new_envp[0] = NULL;
> 
> 		*out_envp = new_envp;
> ```

You have tabs instead of spaces here.
___
*[Pal/src/db_main.c, line 178 at r1](https://reviewable.io/reviews/gramineproject/gramine/56#-MjiwQwl0mrpy7XpQunA:-MjiwQwl0mrpy7XpQunB:b8buqsj) ([raw file](https://github.com/gramineproject/gramine/blob/6ac5a70ae00fae8e0dac42d3dad358a2e3c93a1f/Pal/src/db_main.c#L178)):*
Is that true, though? With `propagate` false, there is no deep copy of original environment, just a pass-through of selected variables.

And either way, I think we need to describe what `propagate` does, it wasn't obvious to me.

How about something like:

>Build environment for Gramine process, based on original environment (from host or file) or manifest.
> If `propagate` is true, copies all variables from `orig_envp`, except the ones overriden in manifest.
> Otherwise, copies only the variables from `orig_envp` that the manifest specifies as passthrough.
```c
static int build_envs(const char** orig_envp, bool propagate, const char*** out_envp);
```

The proposed name probably could be better, but my point is: it's not helpful to think about "augmenting" the original environment if you sometimes end up not using it at all.
___
*[Pal/src/db_main.c, line 219 at r1](https://reviewable.io/reviews/gramineproject/gramine/56#-Mjiyjve-UW6KFbzY6bQ:-Mjiyjve-UW6KFbzY6bR:b2x5rii) ([raw file](https://github.com/gramineproject/gramine/blob/6ac5a70ae00fae8e0dac42d3dad358a2e3c93a1f/Pal/src/db_main.c#L219)):*
> ```C
>         return -PAL_ERROR_NOMEM;
> 
>     /* For simplicity, allocate each env anew; this is suboptimal but happens only once. First
> ```

I think the last sentence applies to the second loop. How about:

```c
size_t idx = 0;

/* First, go through original variables and copy the ones that we're going to use
 * (because of `propagate`, or passthrough specified for that variable in manifest). */
for (const char** orig_env = envp; *orig_env; orig_env++) {
    ...
}

/* Then, go through the manifest variables and add copy the ones with value provided. */
for (ssize_t i = 0; i < toml_envs_cnt; i++) {
    ...
}
```

And I don't think the "allocate each env anew" part deserves justification.
___
*[Pal/src/db_main.c, line 268 at r1](https://reviewable.io/reviews/gramineproject/gramine/56#-Mjj-zaN67rDd7pFHef3:-Mjj-zaO3mooN5CV_Xkl:b-ibjyyz) ([raw file](https://github.com/gramineproject/gramine/blob/6ac5a70ae00fae8e0dac42d3dad358a2e3c93a1f/Pal/src/db_main.c#L268)):*
> ```C
>         if (propagate && !env_val && !passthrough) {
>             /* this is not a hard-coded envvar and its passthrough == false */
>             log_error("Forbidden combination: 'loader.insecure__use_host_env' /"
> ```

As said above, I think we should forbid this combination regardless of the value of `propagate`.

This error can seem pretty arbitrary for the user, and doesn't explain what to do instead.

I would add an explanation ("for security reasons, Gramine doesn't allow to selectively disable passthrough for variables, we recommend you specify all variables explicitly instead") or, if that's too long, a starting point like "see documentation for `loader.env` for details".
___
*[Pal/src/db_main.c, line 557 at r1](https://reviewable.io/reviews/gramineproject/gramine/56#-Mjj2aU3Bsw7Q1mzvygt:-Mjj2aU3Bsw7Q1mzvygu:b-kkvqmr) ([raw file](https://github.com/gramineproject/gramine/blob/6ac5a70ae00fae8e0dac42d3dad358a2e3c93a1f/Pal/src/db_main.c#L557)):*
> ```C
>         INIT_FAIL(-ret, "Augmenting environment variables based on the manifest failed");
> 
>     if (initial_environments != environments)
> ```

Is it ever the case that `initial_environments == final_environments`?
___
*[Pal/src/db_main.c, line 558 at r1](https://reviewable.io/reviews/gramineproject/gramine/56#-Mjj2jz7Ac84wzYYGh1-:-Mjj2jz8AyOLUVWjGdq4:b-4b9v94) ([raw file](https://github.com/gramineproject/gramine/blob/6ac5a70ae00fae8e0dac42d3dad358a2e3c93a1f/Pal/src/db_main.c#L558)):*
> ```C
> 
>     if (initial_environments != environments)
>         free(initial_environments);
> ```

You need to also free the individual strings.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/56)* status: all files reviewed, 10 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/56#-Mjj8FGk2eOY3mF73Xoe:-Mjj8FGk2eOY3mF73Xof:b5d5wjn) (no related file):*
Replied with quick comments for now. Tomorrow I'll fix everything.

---
___
*[Documentation/manifest-syntax.rst, line 162 at r1](https://reviewable.io/reviews/gramineproject/gramine/56#-MjirsxG7ioBquQaQeyC:-Mjj7BWoCETcVE325M1T:b-bh5usb) ([raw file](https://github.com/gramineproject/gramine/blob/6ac5a70ae00fae8e0dac42d3dad358a2e3c93a1f/Documentation/manifest-syntax.rst#L162)):*

Why do we even allow `{ passthrough = false }`? As far as I can tell, it's either disallowed, or has no effect.

(And if we disallow `passthrough = false`, the "forbidden combination" can be detected directly in `get_env_value_from_manifest`.)
</blockquote></details>

Do you mean to:
- In the code, fail immediately when we find `passthrough = false`?
- In the docs, don't even mention the possibility of `false`?

I can do both, yeah. It just felt weird to "ignore" the false value in the documentation...
___
*[Pal/src/db_main.c, line 178 at r1](https://reviewable.io/reviews/gramineproject/gramine/56#-MjiwQwl0mrpy7XpQunA:-Mjj7dsGEvkx2GWViedb:b-b5ttzr) ([raw file](https://github.com/gramineproject/gramine/blob/6ac5a70ae00fae8e0dac42d3dad358a2e3c93a1f/Pal/src/db_main.c#L178)):*
> With propagate false, there is no deep copy of original environment, just a pass-through of selected variables.

No, that's not correct. The new environment is allocated, and original envvars are copied via `strdup`. So it is a deep copy.
___
*[Pal/src/db_main.c, line 557 at r1](https://reviewable.io/reviews/gramineproject/gramine/56#-Mjj2aU3Bsw7Q1mzvygt:-Mjj7semA8RMx9vypKu_:b-lv2el0) ([raw file](https://github.com/gramineproject/gramine/blob/6ac5a70ae00fae8e0dac42d3dad358a2e3c93a1f/Pal/src/db_main.c#L557)):*

Is it ever the case that `initial_environments == final_environments`?
</blockquote></details>

No, it is not possible. `final_environments` is always a new deep copy of the `initial_environments` (or a completely new environment object, if there was no `env_src_file` and `use_host_env = false`).
___
*[Pal/src/db_main.c, line 558 at r1](https://reviewable.io/reviews/gramineproject/gramine/56#-Mjj2jz7Ac84wzYYGh1-:-Mjj81yV3e6G4OFDo6FR:b-v6w1q1) ([raw file](https://github.com/gramineproject/gramine/blob/6ac5a70ae00fae8e0dac42d3dad358a2e3c93a1f/Pal/src/db_main.c#L558)):*

You need to also free the individual strings.
</blockquote></details>

No, because `initial_environments` in this case is created via `load_cstring_array()` which mallocs only one object (see the comment there).

pwmarcz (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/56)* status: all files reviewed, 10 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv and @pwmarcz)

___
*[Documentation/manifest-syntax.rst, line 162 at r1](https://reviewable.io/reviews/gramineproject/gramine/56#-MjirsxG7ioBquQaQeyC:-MjjxUfE-q2Dg8-jjSm_:b-8mfjc1) ([raw file](https://github.com/gramineproject/gramine/blob/6ac5a70ae00fae8e0dac42d3dad358a2e3c93a1f/Documentation/manifest-syntax.rst#L162)):*

Do you mean to:
- In the code, fail immediately when we find `passthrough = false`?
- In the docs, don't even mention the possibility of `false`?

I can do both, yeah. It just felt weird to "ignore" the false value in the documentation...
</blockquote></details>

Yes, I mean both. As for justifying it: we could say false is the default, so `{ passthrough = false }` (without `value`) is the same as not saying *anything* about a variable. If you specify a variable, it needs to either have a value, or be a passthrough.

(By that logic, I guess we could admit `{ value = xyz, passthrough = false }` and make it mean the same as `{ value = xyz }`... not sure if that helps, though, or makes it more complicated).
___
*[Pal/src/db_main.c, line 178 at r1](https://reviewable.io/reviews/gramineproject/gramine/56#-MjiwQwl0mrpy7XpQunA:-MjjzfQ-82VC4crrqWWp:bxxa6jy) ([raw file](https://github.com/gramineproject/gramine/blob/6ac5a70ae00fae8e0dac42d3dad358a2e3c93a1f/Pal/src/db_main.c#L178)):*
> No, that's not correct. The new environment is allocated, and original envvars are copied via strdup. So it is a deep copy.

I agree that is what happens. I guess by "deep copy" I understood a copy of the whole structure, so the description was confusing to me.

I still think there should be a better explanation of what this function does and how `propagate` works.
___
*[Pal/src/db_main.c, line 558 at r1](https://reviewable.io/reviews/gramineproject/gramine/56#-Mjj2jz7Ac84wzYYGh1-:-Mjk0qzV4KLNqNkszdwU:bn5n7t6) ([raw file](https://github.com/gramineproject/gramine/blob/6ac5a70ae00fae8e0dac42d3dad358a2e3c93a1f/Pal/src/db_main.c#L558)):*

No, because `initial_environments` in this case is created via `load_cstring_array()` which mallocs only one object (see the comment there).
</blockquote></details>

I see the comment, but I also see `load_cstring_array` make two mallocs (one for `buf` and one for `array`). So it looks like you should `free(initial_environments[0])` and then `free(initial_environments)`?

(And the comment for `load_cstring_array()` should probably make that more clear).

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/56)* status: 3 of 6 files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[Documentation/manifest-syntax.rst, line 162 at r1](https://reviewable.io/reviews/gramineproject/gramine/56#-MjirsxG7ioBquQaQeyC:-Mjmd-CjEmp-sGdGOiUC:b-5t3i4e) ([raw file](https://github.com/gramineproject/gramine/blob/6ac5a70ae00fae8e0dac42d3dad358a2e3c93a1f/Documentation/manifest-syntax.rst#L162)):*
Done.

> (By that logic, I guess we could admit { value = xyz, passthrough = false } and make it mean the same as { value = xyz }... not sure if that helps, though, or makes it more complicated).

To me, it nakes it more complicated. So didn't implement this additional corner case.
___
*[Pal/src/db_main.c, line 146 at r1](https://reviewable.io/reviews/gramineproject/gramine/56#-Mjipgbu-MveI8JCQjU7:-MjmdmV1CBGUG-7Sb4cX:b1c5ibk) ([raw file](https://github.com/gramineproject/gramine/blob/6ac5a70ae00fae8e0dac42d3dad358a2e3c93a1f/Pal/src/db_main.c#L146)):*

`propagate=false` essentially converts this function into one creating an empty environment, right?

It's very confusing, I'd prefer to have a separate function, even if it complicates the call sites:

```c
// instead of:
return deep_copy_envs(propagate, envp, out_envp);

// write:
return propagate ? deep_copy_envs(envp, out_envp) : create_empty_envs(out_envp);
```

</blockquote></details>

Done. Good suggestion.
___
*[Pal/src/db_main.c, line 149 at r1](https://reviewable.io/reviews/gramineproject/gramine/56#-MjioItlBU6vUxNvoz9r:-Mjme6l892m49P4QbBZW:b-b7bipc) ([raw file](https://github.com/gramineproject/gramine/blob/6ac5a70ae00fae8e0dac42d3dad358a2e3c93a1f/Pal/src/db_main.c#L149)):*

You have tabs instead of spaces here.
</blockquote></details>

Done. I wonder how this happened.
___
*[Pal/src/db_main.c, line 152 at r1](https://reviewable.io/reviews/gramineproject/gramine/56#-MjioMX00YlQ4c8t5fnb:-Mjme8Ie97bZmL4Jc-po:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/6ac5a70ae00fae8e0dac42d3dad358a2e3c93a1f/Pal/src/db_main.c#L152)):*

You have tabs instead of spaces here.
</blockquote></details>

Done.
___
*[Pal/src/db_main.c, line 178 at r1](https://reviewable.io/reviews/gramineproject/gramine/56#-MjiwQwl0mrpy7XpQunA:-MjmfnJX5P6Ngcprzlp5:b-ty45v7) ([raw file](https://github.com/gramineproject/gramine/blob/6ac5a70ae00fae8e0dac42d3dad358a2e3c93a1f/Pal/src/db_main.c#L178)):*

> No, that's not correct. The new environment is allocated, and original envvars are copied via strdup. So it is a deep copy.

I agree that is what happens. I guess by "deep copy" I understood a copy of the whole structure, so the description was confusing to me.

I still think there should be a better explanation of what this function does and how `propagate` works.
</blockquote></details>

Done. Thanks for the awesome suggestion on the comment and function name!
___
*[Pal/src/db_main.c, line 219 at r1](https://reviewable.io/reviews/gramineproject/gramine/56#-Mjiyjve-UW6KFbzY6bQ:-MjmgOOM5KeeQyMHGT4v:b-1adtoa) ([raw file](https://github.com/gramineproject/gramine/blob/6ac5a70ae00fae8e0dac42d3dad358a2e3c93a1f/Pal/src/db_main.c#L219)):*

I think the last sentence applies to the second loop. How about:

```c
size_t idx = 0;

/* First, go through original variables and copy the ones that we're going to use
 * (because of `propagate`, or passthrough specified for that variable in manifest). */
for (const char** orig_env = envp; *orig_env; orig_env++) {
    ...
}

/* Then, go through the manifest variables and add copy the ones with value provided. */
for (ssize_t i = 0; i < toml_envs_cnt; i++) {
    ...
}
```

And I don't think the "allocate each env anew" part deserves justification.

</blockquote></details>

Done. Again, thanks for amazing suggestions! Your skill in explaining code succinctly is impressive :)
___
*[Pal/src/db_main.c, line 268 at r1](https://reviewable.io/reviews/gramineproject/gramine/56#-Mjj-zaN67rDd7pFHef3:-MjmhnJMERiJJwG4Hfs_:b-9iwavq) ([raw file](https://github.com/gramineproject/gramine/blob/6ac5a70ae00fae8e0dac42d3dad358a2e3c93a1f/Pal/src/db_main.c#L268)):*

As said above, I think we should forbid this combination regardless of the value of `propagate`.

This error can seem pretty arbitrary for the user, and doesn't explain what to do instead.

I would add an explanation ("for security reasons, Gramine doesn't allow to selectively disable passthrough for variables, we recommend you specify all variables explicitly instead") or, if that's too long, a starting point like "see documentation for `loader.env` for details".
</blockquote></details>

Done. Your explanation was too long, so I cut it a bit.
___
*[Pal/src/db_main.c, line 557 at r1](https://reviewable.io/reviews/gramineproject/gramine/56#-Mjj2aU3Bsw7Q1mzvygt:-MjmjCM4EeViRB6EjfuB:beofpzu) ([raw file](https://github.com/gramineproject/gramine/blob/6ac5a70ae00fae8e0dac42d3dad358a2e3c93a1f/Pal/src/db_main.c#L557)):*

No, it is not possible. `final_environments` is always a new deep copy of the `initial_environments` (or a completely new environment object, if there was no `env_src_file` and `use_host_env = false`).
</blockquote></details>

Pawel, you created a blocking comment here. Is there something I need to fix?
___
*[Pal/src/db_main.c, line 558 at r1](https://reviewable.io/reviews/gramineproject/gramine/56#-Mjj2jz7Ac84wzYYGh1-:-Mjmj3cvCB49BNMwPHXw:bxlz0dr) ([raw file](https://github.com/gramineproject/gramine/blob/6ac5a70ae00fae8e0dac42d3dad358a2e3c93a1f/Pal/src/db_main.c#L558)):*

I see the comment, but I also see `load_cstring_array` make two mallocs (one for `buf` and one for `array`). So it looks like you should `free(initial_environments[0])` and then `free(initial_environments)`?

(And the comment for `load_cstring_array()` should probably make that more clear).
</blockquote></details>

Done. Oops, you are right. Yes, the comment in `load_cstring_array()` was confusing, fixed it.

pwmarcz (approved):
> Reviewed 2 of 3 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/56)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[Documentation/manifest-syntax.rst, line 197 at r3](https://reviewable.io/reviews/gramineproject/gramine/56#-Mjmu4vyARtHFsHPUCOP:-Mjmu4vz-bP0l3_CsU-T:b-84b4lv) ([raw file](https://github.com/gramineproject/gramine/blob/c9769c64f10f5729384fe466c99f67f680a28614/Documentation/manifest-syntax.rst#L197)):*
> ```reStructuredText
>    ``insecure__use_host_env`` and then disallow some of them using ``passthrough
>    = false``. However, this deny list approach is intentionally prohibited for
>    security reasons. Graphene loudly fails if any ``passthrough = false``
> ```

I don't think this requires a lengthy explanation, but "for security reasons" is too little, IMO: it just suggests that the explanation exists elsewhere, so it can be frustrating to encounter e.g. if you just saw the error message and came here to read more about it.

Maybe: "this deny list approach is intentionally prohibited because it's inherently insecure / doesn't provide any real security"?

(Not blocking, though).
___
*[Pal/src/db_main.c, line 557 at r1](https://reviewable.io/reviews/gramineproject/gramine/56#-Mjj2aU3Bsw7Q1mzvygt:-MjmrzWS3fHD1lruMJmF:b44om43) ([raw file](https://github.com/gramineproject/gramine/blob/6ac5a70ae00fae8e0dac42d3dad358a2e3c93a1f/Pal/src/db_main.c#L557)):*

Pawel, you created a blocking comment here. Is there something I need to fix?
</blockquote></details>

Right, I misread this line as `initial_environments = final_environments`. Nothing to fix here.
___
*[Pal/src/db_main.c, line 550 at r3](https://reviewable.io/reviews/gramineproject/gramine/56#-MjmsAjXEqsRm2VpLyII:-MjmsAjXEqsRm2VpLyIJ:b-uga3zv) ([raw file](https://github.com/gramineproject/gramine/blob/c9769c64f10f5729384fe466c99f67f680a28614/Pal/src/db_main.c#L550)):*
> ```C
>     // TODO: Envs from file should be able to override ones from the manifest, but current
>     // code makes this hard to implement.
>     ret = build_envs(orig_environments, use_host_env || env_src_file, &final_environments);
> ```

I think this would read better with `/*propagate=*/use_host_env || env_src_file`, but I won't insist.

dimakuv:
FYI: Jenkins-18.04 timed out on `apps.LTP.sendfile09` (even though both tests PASSED), not related and looks like a real timeout

pwmarcz (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/56)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

mkow (changes requested):
> Reviewed 3 of 6 files at r1, 1 of 3 files at r2, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/56)* status: all files reviewed, 9 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[Documentation/manifest-syntax.rst, line 198 at r4](https://reviewable.io/reviews/gramineproject/gramine/56#-Mkib7tRDQFcE0n3GQ4B:-Mkib7tRDQFcE0n3GQ4C:b-pe01fi) ([raw file](https://github.com/gramineproject/gramine/blob/32ecfd4a60708c82b038605b9f2f8ffb6663669c/Documentation/manifest-syntax.rst#L198)):*
> ```
> Graphene loudly fails if ``passthrough = false`` manifest options are set.
> ```

> ```
> If you do not specify a variable in the manifest, it has the same effect as ``{ passthrough = false }``.
> ```

Aren't these two contradictory? Can I use `passthrough = false` in the manifests or not?
___
*[Pal/src/db_main.c, line 74 at r4](https://reviewable.io/reviews/gramineproject/gramine/56#-MkigkPV0oT4UKMvCGYC:-MkigkPW1pYCB-4UuK80:bu043o2) ([raw file](https://github.com/gramineproject/gramine/blob/32ecfd4a60708c82b038605b9f2f8ffb6663669c/Pal/src/db_main.c#L74)):*
> ```
> and the corresponding `*out_val` or `*out_passthrough` is set
> ```

How is the caller supposed to know which one was set?
___
*[Pal/src/db_main.c, line 93 at r4](https://reviewable.io/reviews/gramineproject/gramine/56#-Mkihpf4179jsJPZd9gG:-Mkihpf4179jsJPZd9gH:b-vstnir) ([raw file](https://github.com/gramineproject/gramine/blob/32ecfd4a60708c82b038605b9f2f8ffb6663669c/Pal/src/db_main.c#L93)):*
> ```C
>         }
> 
>         if (!out_val)
> ```

I'd either drop these checks or convert into an assert at the beginning.
___
*[Pal/src/db_main.c, line 116 at r4](https://reviewable.io/reviews/gramineproject/gramine/56#-MkikoX83oZdrHglVJZZ:-MkikoX83oZdrHglVJZ_:bagikdw) ([raw file](https://github.com/gramineproject/gramine/blob/32ecfd4a60708c82b038605b9f2f8ffb6663669c/Pal/src/db_main.c#L116)):*
> ```
> !out_val
> ```

ditto
___
*[Pal/src/db_main.c, line 129 at r4](https://reviewable.io/reviews/gramineproject/gramine/56#-MkilGiHDAmCboOzbgqK:-MkilGiHDAmCboOzbgqL:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/32ecfd4a60708c82b038605b9f2f8ffb6663669c/Pal/src/db_main.c#L129)):*
> ```C
> 
>     /* at this point, it must be `loader.env.key = { passthrough = true|false }`*/
>     if (!out_passthrough)
> ```

ditto
___
*[Pal/src/db_main.c, line 279 at r4](https://reviewable.io/reviews/gramineproject/gramine/56#-MkitExR8BeGAuyKtP9c:-MkitExR8BeGAuyKtP9d:b-j29ixi) ([raw file](https://github.com/gramineproject/gramine/blob/32ecfd4a60708c82b038605b9f2f8ffb6663669c/Pal/src/db_main.c#L279)):*
> ```C
> 
>         char* final_env = alloc_concat3(toml_env_key, strlen(toml_env_key), "=", 1, env_val,
>                                         strlen(env_val));
> ```

You can pass -1 instead of these strlens
___
*[Pal/src/db_main.c, line 506 at r4](https://reviewable.io/reviews/gramineproject/gramine/56#-MkiueYBBckMx7iDGMAh:-MkiueYBBckMx7iDGMAi:bo7rt4f) ([raw file](https://github.com/gramineproject/gramine/blob/32ecfd4a60708c82b038605b9f2f8ffb6663669c/Pal/src/db_main.c#L506)):*
> ```C
> 
>     /* we don't modify original `environments` but deep-copy them into `final_environments`,
>      * augmented based on `loader.env.key` manifest options */
> ```

I'd drop this comment, because why would you even consider modifying `environments`? The env we're building here is for the emulated app, not for Gramine emulation code.
___
*[Pal/src/db_main.c, line 522 at r4](https://reviewable.io/reviews/gramineproject/gramine/56#-MkitspI0S9qSVmzk8mO:-MkitspI0S9qSVmzk8mP:bjlko4z) ([raw file](https://github.com/gramineproject/gramine/blob/32ecfd4a60708c82b038605b9f2f8ffb6663669c/Pal/src/db_main.c#L522)):*
> ```
> All "'passthrough' environment variables in the manifest are ignored.
> ```

This is weird, what this sentence actually means? I.e. what were you trying to say?
___
*[Pal/src/db_main.c, line 545 at r4](https://reviewable.io/reviews/gramineproject/gramine/56#-MkiuGt5B03sydWZ8dCg:-MkiuGt5B03sydWZ8dCh:b3nrwsh) ([raw file](https://github.com/gramineproject/gramine/blob/32ecfd4a60708c82b038605b9f2f8ffb6663669c/Pal/src/db_main.c#L545)):*
> ```C
>     } else {
>         /* Environment variables are taken from the host. */
>         orig_environments = environments;
> ```

Hmm, but this means that propagation doesn't work with the "env from file"?

dimakuv:
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/56#-MkkOE3L1Jk92-hVi8zP:-MkkOE3L1Jk92-hVi8zQ:b-hzo79u) (no related file):*
Jenkins failed because it is a smart-ass. It merged my PR with the latest master:
```
Commit message: "Merge 22f1b3a5d0788181c31bc288084cf5c70ec8c717 into 5c505d44ab6ffa6d303cde51b4ccf922c3eaab88"
```

And of course everything failed because the latest master uses "Gramine" everywhere and my manifest file still uses "Graphene". I hate this Jenkins auto-merge feature.

---

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/56)* status: 2 of 6 files reviewed, 9 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/56#-MkkOE3L1Jk92-hVi8zP:-MkkPtzc35qjCO5MVxS5:b-p0fgfn) (no related file):*

Jenkins failed because it is a smart-ass. It merged my PR with the latest master:
```
Commit message: "Merge 22f1b3a5d0788181c31bc288084cf5c70ec8c717 into 5c505d44ab6ffa6d303cde51b4ccf922c3eaab88"
```

And of course everything failed because the latest master uses "Gramine" everywhere and my manifest file still uses "Graphene". I hate this Jenkins auto-merge feature.
</blockquote></details>

Done. I had to squash and rebase to the latest master, sorry for this inconvinience. 

---
___
*[Documentation/manifest-syntax.rst, line 198 at r4](https://reviewable.io/reviews/gramineproject/gramine/56#-Mkib7tRDQFcE0n3GQ4B:-MkkEaWPC5TTIeZ5CHmJ:b-jos1tb) ([raw file](https://github.com/gramineproject/gramine/blob/32ecfd4a60708c82b038605b9f2f8ffb6663669c/Documentation/manifest-syntax.rst#L198)):*

> ```
> Graphene loudly fails if ``passthrough = false`` manifest options are set.
> ```

> ```
> If you do not specify a variable in the manifest, it has the same effect as ``{ passthrough = false }``.
> ```

Aren't these two contradictory? Can I use `passthrough = false` in the manifests or not?
</blockquote></details>

Ok, fair enough, bad wording. I removed the `If you do not specify ...` sentence -- it was misleading (I wanted to say "if you don't specify anything, then the behavior is no-passthrough").
___
*[Pal/src/db_main.c, line 74 at r4](https://reviewable.io/reviews/gramineproject/gramine/56#-MkigkPV0oT4UKMvCGYC:-MkkFc-14PDZkTDHctq-:bho48qq) ([raw file](https://github.com/gramineproject/gramine/blob/32ecfd4a60708c82b038605b9f2f8ffb6663669c/Pal/src/db_main.c#L74)):*

> ```
> and the corresponding `*out_val` or `*out_passthrough` is set
> ```

How is the caller supposed to know which one was set?
</blockquote></details>

Done. I reworded the comment. Basically, the caller checks `*out_val == NULL`: if NULL, then it is a passthrough variable, otherwise it is a value variable.
___
*[Pal/src/db_main.c, line 93 at r4](https://reviewable.io/reviews/gramineproject/gramine/56#-Mkihpf4179jsJPZd9gG:-MkkGCc6BiTtMAAIwmGL:b-hmsfh1) ([raw file](https://github.com/gramineproject/gramine/blob/32ecfd4a60708c82b038605b9f2f8ffb6663669c/Pal/src/db_main.c#L93)):*

I'd either drop these checks or convert into an assert at the beginning.
</blockquote></details>

Done. Moved them all on top (we currently always pass all arguments, and if this changes in the future, we'll re-shuffle the asserts).
___
*[Pal/src/db_main.c, line 116 at r4](https://reviewable.io/reviews/gramineproject/gramine/56#-MkikoX83oZdrHglVJZZ:-MkkGLWT5-4Ap2-Svug4:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/32ecfd4a60708c82b038605b9f2f8ffb6663669c/Pal/src/db_main.c#L116)):*

> ```
> !out_val
> ```

ditto
</blockquote></details>

Done.
___
*[Pal/src/db_main.c, line 129 at r4](https://reviewable.io/reviews/gramineproject/gramine/56#-MkilGiHDAmCboOzbgqK:-MkkGLyaAuW53gPXAgqL:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/32ecfd4a60708c82b038605b9f2f8ffb6663669c/Pal/src/db_main.c#L129)):*

ditto
</blockquote></details>

Done.
___
*[Pal/src/db_main.c, line 279 at r4](https://reviewable.io/reviews/gramineproject/gramine/56#-MkitExR8BeGAuyKtP9c:-MkkGdAi0M20Juc9ek-W:bu3lbab) ([raw file](https://github.com/gramineproject/gramine/blob/32ecfd4a60708c82b038605b9f2f8ffb6663669c/Pal/src/db_main.c#L279)):*

You can pass -1 instead of these strlens
</blockquote></details>

Done. Didn't notice this trick, thanks.
___
*[Pal/src/db_main.c, line 506 at r4](https://reviewable.io/reviews/gramineproject/gramine/56#-MkiueYBBckMx7iDGMAh:-MkkGlTu57-ySs1sfYNC:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/32ecfd4a60708c82b038605b9f2f8ffb6663669c/Pal/src/db_main.c#L506)):*

I'd drop this comment, because why would you even consider modifying `environments`? The env we're building here is for the emulated app, not for Gramine emulation code.
</blockquote></details>

Done.
___
*[Pal/src/db_main.c, line 522 at r4](https://reviewable.io/reviews/gramineproject/gramine/56#-MkitspI0S9qSVmzk8mO:-MkkGrsK38FmmIi-e_gY:b9b8k8f) ([raw file](https://github.com/gramineproject/gramine/blob/32ecfd4a60708c82b038605b9f2f8ffb6663669c/Pal/src/db_main.c#L522)):*

> ```
> All "'passthrough' environment variables in the manifest are ignored.
> ```

This is weird, what this sentence actually means? I.e. what were you trying to say?
</blockquote></details>

I meant to be explicit that `loader.env.foo = { passthrough = true }` will have no effect (they are consumed by `loader.insecure__use_host_env`).

I guess this is obvious, right? Let me just remove this sentence.
___
*[Pal/src/db_main.c, line 545 at r4](https://reviewable.io/reviews/gramineproject/gramine/56#-MkiuGt5B03sydWZ8dCg:-MkkHhtB7qtfSnq9pN6K:bidv895) ([raw file](https://github.com/gramineproject/gramine/blob/32ecfd4a60708c82b038605b9f2f8ffb6663669c/Pal/src/db_main.c#L545)):*

Hmm, but this means that propagation doesn't work with the "env from file"?
</blockquote></details>

Yes. "Env from file" approach means "the set of envvars is not taken from the host, it is taken from the file". So *propagation of host envvars* will not happen. But this looks the right way to do this, why would anyone want to mix "envs from file" and "envs from host"? This is also what our documentation seems to imply: `loader.env_src_file allows to specify a URI to a file containing serialized
environment`.

Or maybe you're asking another question: "will we propagate only those envs-from-file that have `{ propagate = true }`"? No, that's not the case. Notice the `build_envs(.., /*propagate=*/use_host_env || env_src_file, ..)` below. If we have the env-from-file approach, then `propagate=true` and all envs from file will be propagated regardless of `{ propagate = true }`. (But envs from file may be overwritten by `{ value = "xxx" }`, just like what the TODO comment says.)

pwmarcz (dismissed):
> Reviewed 4 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/56)* status: all files reviewed, 9 unresolved discussions (waiting on @mkow)

mkow (approved):
> Reviewed 4 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/56)* status: :shipit: complete! all files reviewed, all discussions resolved

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/56#-MkkOE3L1Jk92-hVi8zP:-MklHryl6fJB9kyk2mEl:bkfsscu) (no related file):*
> And of course everything failed because the latest master uses "Gramine" everywhere and my manifest file still uses "Graphene". I hate this Jenkins auto-merge feature.

Everyone hates it :) But no one yet found a way to disable this "feature" of Jenkins.

---
___
*[Documentation/manifest-syntax.rst, line 198 at r4](https://reviewable.io/reviews/gramineproject/gramine/56#-Mkib7tRDQFcE0n3GQ4B:-MklYmpS6oTlfs7-DmIJ:b-dcbopc) ([raw file](https://github.com/gramineproject/gramine/blob/32ecfd4a60708c82b038605b9f2f8ffb6663669c/Documentation/manifest-syntax.rst#L198)):*

Ok, fair enough, bad wording. I removed the `If you do not specify ...` sentence -- it was misleading (I wanted to say "if you don't specify anything, then the behavior is no-passthrough").
</blockquote></details>

Much better now, thanks!
___
*[Pal/src/db_main.c, line 522 at r4](https://reviewable.io/reviews/gramineproject/gramine/56#-MkitspI0S9qSVmzk8mO:-MklZp5bCRgSVVA0LYvx:byncdep) ([raw file](https://github.com/gramineproject/gramine/blob/32ecfd4a60708c82b038605b9f2f8ffb6663669c/Pal/src/db_main.c#L522)):*

I meant to be explicit that `loader.env.foo = { passthrough = true }` will have no effect (they are consumed by `loader.insecure__use_host_env`).

I guess this is obvious, right? Let me just remove this sentence.
</blockquote></details>

Yeah, this part of the meaning is obvious and the other part is probably wrong - `passthrough = false` are probably still errored out?
So yes, I prefer the new version without this sentence at all :)
___
*[Pal/src/db_main.c, line 545 at r4](https://reviewable.io/reviews/gramineproject/gramine/56#-MkiuGt5B03sydWZ8dCg:-Mkl_65u8QVLSOpS3BRJ:b-ghe1gy) ([raw file](https://github.com/gramineproject/gramine/blob/32ecfd4a60708c82b038605b9f2f8ffb6663669c/Pal/src/db_main.c#L545)):*

Yes. "Env from file" approach means "the set of envvars is not taken from the host, it is taken from the file". So *propagation of host envvars* will not happen. But this looks the right way to do this, why would anyone want to mix "envs from file" and "envs from host"? This is also what our documentation seems to imply: `loader.env_src_file allows to specify a URI to a file containing serialized
environment`.

Or maybe you're asking another question: "will we propagate only those envs-from-file that have `{ propagate = true }`"? No, that's not the case. Notice the `build_envs(.., /*propagate=*/use_host_env || env_src_file, ..)` below. If we have the env-from-file approach, then `propagate=true` and all envs from file will be propagated regardless of `{ propagate = true }`. (But envs from file may be overwritten by `{ value = "xxx" }`, just like what the TODO comment says.)
</blockquote></details>

Ok, makes sense, although I previously though of "env-from-file" just as an extension to the env list in the manifest.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/56)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #71, opened 2021-09-17
Compile glibc in meson, retire `Runtime/`

woju:
## Description of the changes 

Glibc is compiled as part of meson build, and in parallel to everything else. Build time on 8 cores is about 2 min (`ninja -C build`). 

## How to test this PR? 

Jenkins

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/71)


woju:
Jenkins, test Jenkins-20.04 please for sendfile09 timeout

dimakuv (changes requested):
> Reviewed 8 of 9 files at r1, 12 of 13 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/71)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @woju)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/71#-Mk7bvadA6qKu4k9Q7kM:-Mk7bvadA6qKu4k9Q7kN:bdxjztj) (no related file):*
You still have this file: https://github.com/gramineproject/gramine/blob/woju/meson-glibc/LibOS/glibc-checksums

Shouldn't you remove it?

---

woju:
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/71#-Mk7bvadA6qKu4k9Q7kM:-Mk7eVa9BRSEwg2w6spP:byqld4o) (no related file):*

You still have this file: https://github.com/gramineproject/gramine/blob/woju/meson-glibc/LibOS/glibc-checksums

Shouldn't you remove it?
</blockquote></details>

Yes, I should. I'll push it together with fixes for @mkow's review, which is still brewing.

---

mkow (changes requested):
> Reviewed 8 of 9 files at r1, 12 of 13 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/71)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @woju)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/71#-Mk7bvadA6qKu4k9Q7kM:-Mk7gDu62cCozLiLmSf4:b-wtsr3x) (no related file):*

Yes, I should. I'll push it together with fixes for @mkow's review, which is still brewing.
</blockquote></details>

Heh, I had exactly the same comment, but @dimakuv was faster :)

---
___
*[-- commits, line 2 at r3](https://reviewable.io/reviews/gramineproject/gramine/71#-Mk7QKeY3VqYeAs9ghyy):*
`meson` -> `Meson`
___
*[LibOS/Makefile, line 7 at r3](https://reviewable.io/reviews/gramineproject/gramine/71#-Mk7eg9N-EI_bN2MrHqG:-Mk7eg9O223KCpN0t6F5:b-yb3nd2) ([raw file](https://github.com/gramineproject/gramine/blob/9590c23ced87b04c7706406421f047b8f6688890/LibOS/Makefile#L7)):*
> ```Makefile
> 
> define LN_SF_TO_RUNTIME_DIR_template =
> $(RUNTIME_DIR)/$(notdir $(1)): $(1)
> ```

Didn't you remove `Runtime/` in this PR? Please grep the whole repo for such stale references.
___
*[LibOS/Makefile, line 13 at r3](https://reviewable.io/reviews/gramineproject/gramine/71#-Mk7ec338R0dSzNgXfjK:-Mk7ec338R0dSzNgXfjL:btx67u9) ([raw file](https://github.com/gramineproject/gramine/blob/9590c23ced87b04c7706406421f047b8f6688890/LibOS/Makefile#L13)):*
> ```Makefile
> GNU_MIRRORS ?= https://ftp.gnu.org/gnu/ \
>                https://mirrors.kernel.org/gnu/ \
>                https://mirrors.ocf.berkeley.edu/gnu/
> ```

Do we still need this?
___
*[subprojects/packagefiles/glibc-2.34/compile.sh, line 16 at r3](https://reviewable.io/reviews/gramineproject/gramine/71#-Mk7gclp8HmwBEna1Ylm:-Mk7gclq5ePHpxQkPTxI:b-keitn6) ([raw file](https://github.com/gramineproject/gramine/blob/9590c23ced87b04c7706406421f047b8f6688890/subprojects/packagefiles/glibc-2.34/compile.sh#L16)):*
> ```Shell
> CXX=g++
> AS=gcc
> CFLAGS="-O2 -Wno-unused-values"
> ```

Didn't you accidentally remove `-g` from debug builds?
___
*[subprojects/packagefiles/glibc-2.34/compile.sh, line 52 at r3](https://reviewable.io/reviews/gramineproject/gramine/71#-Mk7h2Nh3ykJlf4W5iEj:-Mk7h2Nh3ykJlf4W5iEk:b-88q3yb) ([raw file](https://github.com/gramineproject/gramine/blob/9590c23ced87b04c7706406421f047b8f6688890/subprojects/packagefiles/glibc-2.34/compile.sh#L52)):*
> ```
> "$CURRENT_BUILD_DIR"/
> ```

Isn't this a risky practice in general? If for some reason `$CURRENT_BUILD_DIR` ends up empty you'll have `/` here, instead of wrong argument count for `cp`.

This particular place doesn't look dangerous, but looks like a practice we don't want to encourage.

woju:
___
*[subprojects/packagefiles/glibc-2.34/compile.sh, line 52 at r3](https://reviewable.io/reviews/gramineproject/gramine/71#-Mk7h2Nh3ykJlf4W5iEj:-Mk7pP7TChC7zWveePxo:b11x18d) ([raw file](https://github.com/gramineproject/gramine/blob/9590c23ced87b04c7706406421f047b8f6688890/subprojects/packagefiles/glibc-2.34/compile.sh#L52)):*

The problem is, without `/` at the end, `cp` (and friends like `mv`, `ls`) has different behaviour depending on whether the target exists or not, or is it a directory. So if you really like to have this fixed, I'd have to change `cp` to `install`. Should I?
</blockquote></details>

`ln`, obviously

dimakuv (dismissed):
> Reviewed 9 of 9 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/71)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required) (waiting on @mkow)

___
*[LibOS/Makefile, line 13 at r3](https://reviewable.io/reviews/gramineproject/gramine/71#-Mk7ec338R0dSzNgXfjK:-Mk7yta7DsaOrm_PXd35:bay5eou) ([raw file](https://github.com/gramineproject/gramine/blob/9590c23ced87b04c7706406421f047b8f6688890/LibOS/Makefile#L13)):*

For GCC below, which is disabled by default, but still present.
</blockquote></details>

Yeah, we'll need to add Meson build file for GCC as well, at some point...

mkow (commented):
> Reviewed 9 of 9 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/71)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required) (waiting on @woju)

___
*[subprojects/packagefiles/glibc-2.34/compile.sh, line 52 at r3](https://reviewable.io/reviews/gramineproject/gramine/71#-Mk7h2Nh3ykJlf4W5iEj:-Mk81gTj2UzKdzSZOXAV:b-frtveg) ([raw file](https://github.com/gramineproject/gramine/blob/9590c23ced87b04c7706406421f047b8f6688890/subprojects/packagefiles/glibc-2.34/compile.sh#L52)):*

`ln`, obviously
</blockquote></details>

Ah, I thought they look at the type of the source only? And I can't see anything abut this in their manpages.

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/71)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required) (waiting on @mkow)

___
*[subprojects/packagefiles/glibc-2.34/compile.sh, line 52 at r3](https://reviewable.io/reviews/gramineproject/gramine/71#-Mk7h2Nh3ykJlf4W5iEj:-Mk85QMv3V440BHdWtsH:b-x1yved) ([raw file](https://github.com/gramineproject/gramine/blob/9590c23ced87b04c7706406421f047b8f6688890/subprojects/packagefiles/glibc-2.34/compile.sh#L52)):*

Ah, I thought they look at the type of the source only? And I can't see anything abut this in their manpages.
</blockquote></details>

Let's `cp a b`; then:
- if `b` does not exist, it becomes a copy of `a` (if `a` is a directory, you need to add `-r` or you'll get "omitting directory")
- if `b` does exist and is a regular file, then
  - if `a` is also a file, then `b` gets clobbered (unless some options like `-i` or `-n` were given) and becomes a copy of `a`
  - if `a` is a directory, it's an error
- if `b` does exist and is a directory, then `b/a` is a new copy of `a` (with caveats, like `b/a` existing, then the logic from two previous points mostly applies).

All this doesn't happen when you add `/` to destination: if `b` doesn't exist or is a regular file, then `b/` qualifies for ENOTDIR.

About manpages: if you look closely at the SYNOPSIS, you have first two signatures, and the difference between first two cases and the third one is just that. Yes, it could have been explained more clearly.

So again, if you'd like this properly fixed, like *posixly* fixed, then I suggest `install`, which is more explicit: https://manpages.debian.org/install(1)

```sh
install -t "$CURRENT_BUILD_DIR" "$output" # -m 0644 -u root -g root, whatever you'd like else
```

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/71)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required) (waiting on @woju)

___
*[-- commits, line 15 at r4](https://reviewable.io/reviews/gramineproject/gramine/71#-Mk8gJTFAzup-oY346oK):*
There are still some references left, please grep harder ;)
___
*[subprojects/packagefiles/glibc-2.34/compile.sh, line 52 at r3](https://reviewable.io/reviews/gramineproject/gramine/71#-Mk7h2Nh3ykJlf4W5iEj:-Mk8dEKm9bR_4mmhB7-D:bbnwrvs) ([raw file](https://github.com/gramineproject/gramine/blob/9590c23ced87b04c7706406421f047b8f6688890/subprojects/packagefiles/glibc-2.34/compile.sh#L52)):*

Let's `cp a b`; then:
- if `b` does not exist, it becomes a copy of `a` (if `a` is a directory, you need to add `-r` or you'll get "omitting directory")
- if `b` does exist and is a regular file, then
  - if `a` is also a file, then `b` gets clobbered (unless some options like `-i` or `-n` were given) and becomes a copy of `a`
  - if `a` is a directory, it's an error
- if `b` does exist and is a directory, then `b/a` is a new copy of `a` (with caveats, like `b/a` existing, then the logic from two previous points mostly applies).

All this doesn't happen when you add `/` to destination: if `b` doesn't exist or is a regular file, then `b/` qualifies for ENOTDIR.

About manpages: if you look closely at the SYNOPSIS, you have first two signatures, and the difference between first two cases and the third one is just that. Yes, it could have been explained more clearly.

So again, if you'd like this properly fixed, like *posixly* fixed, then I suggest `install`, which is more explicit: https://manpages.debian.org/install(1)

```sh
install -t "$CURRENT_BUILD_DIR" "$output" # -m 0644 -u root -g root, whatever you'd like else
</blockquote></details>

Ok, let's keep it as is, although `install` seems more appropriate for this task.


<!-- Sent from Reviewable.io -->

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/71)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (approved):
> Reviewed 1 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/71)* status: :shipit: complete! all files reviewed, all discussions resolved

dimakuv (approved):
> Reviewed 14 of 15 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/71)* status: 26 of 27 files reviewed, all discussions resolved (waiting on @mkow)

mkow (approved):
> Reviewed 15 of 15 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/71)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #32, opened 2021-09-13
First version of AddressSanitizer

pwmarcz:
## Description of the changes 

As with UBSan, this adds our own handlers to Gramine's common library. The current version of ASan integration supports heap allocation only.

Tracking issue: #19 (but I don't consider this done yet, see "What's next")

## How to test this PR? 

Compile with Clang (`CC=clang CXX=clang++ AS=clang`) and ASan (`ASAN=1` and `meson -Dasan=enabled`). 

Run the regression test: `graphene-direct run_test asan_buffer_overflow`

This is the test code:

```c
    uint8_t* buf = malloc(30);
    buf[30] = 1;
    free(buf);
    return 0;
```

It should abort with the following output:

```
$ graphene-direct run_test asan_buffer_overflow
error: Using insecure argv source. Gramine will continue application execution, but this configuration must not be used in production!
[P1:T1:run_test] run_test("asan_buffer_overflow") ...
[P1:T1:run_test] error: asan: heap-buffer-overflow while trying to store 1 bytes at 0x7ad88073014e
[P1:T1:run_test] error: asan: the bad address is 0x7ad88073014e (0 from beginning)
[P1:T1:run_test] error: asan: IP = 0x7ad880a0bce3 (for a full traceback, use GDB with a breakpoint at "shim_abort")
[P1:T1:run_test] error: asan:
[P1:T1:run_test] error: asan: 0x7ad88072ff00  fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
[P1:T1:run_test] error: asan: 0x7ad88072ff80  fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
[P1:T1:run_test] error: asan: 0x7ad880730000  fa fa fa fa fa fa fd fd fd fd fa fa fa fa fd fd
[P1:T1:run_test] error: asan: 0x7ad880730080  fd fd fa fa fa fa fd fd fd fd fa fa fa fa 00 00
[P1:T1:run_test] error: asan: 0x7ad880730100  00 00 fa fa fa fa 00 00 00[06]fa fa fa fa fa fa
[P1:T1:run_test] error: asan: 0x7ad880730180  fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
[P1:T1:run_test] error: asan: 0x7ad880730200  fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
[...]

PR #94, opened 2021-09-27
[Pal/Linux-SGX] Map enclave pages with MAP_FIXED_NOREPLACE

pwmarcz:
Discussed in #32.

The `mmap()` call should fail if something is mapped in the specified range already.

MAP_FIXED_NOREPLACE is available since Linux 4.17, so we had to wait until we dropped support for Ubuntu 16.04.

## How to test this PR? 

Add an overlapping `mmap` to the beginning of main (`sgx_main.c`):

```c
DO_SYSCALL(mmap, (void*)0x1000, 0x1000, 0, MAP_FIXED | MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
```

Gramine invocation should fail:

```
error: ECREATE failed in allocating EPC memory: -17
error: Creating enclave failed: -17
error: load_enclave() failed with error -17
```

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/94)

### DIFF
--- a/Pal/src/host/Linux-SGX/sgx_framework.c
+++ b/Pal/src/host/Linux-SGX/sgx_framework.c
@@ -143,9 +143,9 @@ int create_enclave(sgx_arch_secs_t* secs, sgx_arch_token_t* token) {
     uint64_t addr = DO_SYSCALL(mmap, request_mmap_addr, request_mmap_size,
                                PROT_NONE, /* newer DCAP driver requires such initial mmap */
 #ifdef SGX_DCAP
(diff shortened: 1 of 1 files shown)


pwmarcz:
Uh, actually Ubuntu 18.04 also doesn't support it (it has older glibc, 2.27, and kernel 4.15 by default). Dropping for now.

mkow:
We'll be dropping 18.04 support in +/- half a year, we can merge it then.

PR #98, opened 2021-09-28
[Glibc] Fix the build of Glibc with default prefix `/usr/local`

### LINKED ISSUE #97: Gramine build command fails when run without prefix option
## Description of the problem
If you follow the steps mentioned in the wiki for building Gramine (which doesn't include any prefix for installation), it would result in a failure at the ninja build step where it's not able to install glibc under `/usr/local`

```
checking for kernel header at least 3.2.0... ok
*** On GNU/Linux systems the GNU C Library should not be installed into
*** /usr/local since this might make your system totally unusable.
*** We strongly advise to use a different prefix.  For details read the FAQ.
*** If you really mean to do this, run configure again using the extra
*** parameter `--disable-sanity-checks'.
ninja: build stopped: subcommand failed.
```

## Steps to reproduce

```
make
make ISGX_DRIVER_PATH="" SGX=1                  # this assumes Linux 5.11+
meson setup build/ --buildtype=release -Ddirect=enabled -Dsgx=enabled
ninja -C build/
sudo ninja -C build/ install
```

## Expected results
Gramine build and installation should PASS, as the commands are from the documentation.

## Actual results
Ninja build results in a failure.

dimakuv:
## Description of the changes 

if Gramine is built with the default prefix `/usr/local`, Glibc requires the option `--disable-sanity-checks` otherwise it complains like this:

"On GNU/Linux systems the GNU C Library should not be installed into /usr/local since this might make your system totally unusable. We strongly advise to use a different prefix."

Note that `--disable-sanity-checks` simply silences this warning -- we actually install Glibc under `gramine/runtime/glibc`.

Fixes #97.

## How to test this PR? 

Try to build with default prefix.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/98)

### DIFF
--- a/subprojects/packagefiles/glibc-2.34/compile.sh
+++ b/subprojects/packagefiles/glibc-2.34/compile.sh
@@ -42,6 +42,7 @@ mkdir -p "$BUILDDIR"
         --with-tls \
         --without-gd \
         --without-selinux \
+        --disable-sanity-checks \
         --disable-test \
         --disable-nscd
 


woju (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/98)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

pwmarcz (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/98)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required) (waiting on @dimakuv)

___
*[-- commits, line 4 at r1](https://reviewable.io/reviews/gramineproject/gramine/98#-Mkfek9RBOh0PW0GnnSO):*
Formatting:

* "if" -> "If"
* comma before "otherwise"
___
*[-- commits, line 11 at r1](https://reviewable.io/reviews/gramineproject/gramine/98#-Mkfem_J7ss_n8ZB46Qp):*
This sentence looks confusing for me, because it looks like one of these clauses implies the other (and as I understand it, they're separate statements: 1. this option simply silences this warning, and doesn't do anything else; 2. we don't install directly into `/usr/local`, so this won't actually break your system.)

Maybe it would be better to split it into two sentences?

And I would say `gramine/runtime/glibc` *under the prefix*, or `/usr/local/gramine/runtime/glibc`, to explain how that path is related to the prefix mentioned earlier.
___
*[subprojects/packagefiles/glibc-2.34/compile.sh, line 39 at r1](https://reviewable.io/reviews/gramineproject/gramine/98#-MkfiL4W8ab5TGz2SwTF:-MkfiL4W8ab5TGz2SwTG:b-ti70ku) ([raw file](https://github.com/gramineproject/gramine/blob/078657a078e3a86cd84814a1fd4b34eb0659016f/subprojects/packagefiles/glibc-2.34/compile.sh#L39)):*
> ```Shell
>     cd "$BUILDDIR"
> 
>     # if Gramine is built with the default prefix `/usr/local`, Glibc requires the option
> ```

This is redundant with the commit message. If you want a longer explanation here, in the comment, maybe the commit message can have just a summary? (Or the other way around).

Also, the formatting in the commit message is better, why not format it that way here?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/98)* status: 0 of 1 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @pwmarcz and @woju)

___
*[-- commits, line 4 at r1](https://reviewable.io/reviews/gramineproject/gramine/98#-Mkfek9RBOh0PW0GnnSO):*

Formatting:

* "if" -> "If"
* comma before "otherwise"
</blockquote></details>

Done.
___
*[-- commits, line 11 at r1](https://reviewable.io/reviews/gramineproject/gramine/98#-Mkfem_J7ss_n8ZB46Qp):*

This sentence looks confusing for me, because it looks like one of these clauses implies the other (and as I understand it, they're separate statements: 1. this option simply silences this warning, and doesn't do anything else; 2. we don't install directly into `/usr/local`, so this won't actually break your system.)

Maybe it would be better to split it into two sentences?

And I would say `gramine/runtime/glibc` *under the prefix*, or `/usr/local/gramine/runtime/glibc`, to explain how that path is related to the prefix mentioned earlier.
</blockquote></details>

Done.
___
*[subprojects/packagefiles/glibc-2.34/compile.sh, line 39 at r1](https://reviewable.io/reviews/gramineproject/gramine/98#-MkfiL4W8ab5TGz2SwTF:-Mkfnn375eASfBsV_stW:b-xuj30f) ([raw file](https://github.com/gramineproject/gramine/blob/078657a078e3a86cd84814a1fd4b34eb0659016f/subprojects/packagefiles/glibc-2.34/compile.sh#L39)):*

This is redundant with the commit message. If you want a longer explanation here, in the comment, maybe the commit message can have just a summary? (Or the other way around).

Also, the formatting in the commit message is better, why not format it that way here?
</blockquote></details>

Done. Just removed it. Curious readers can git blame.

pwmarcz (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/98)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/98)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #99, opened 2021-09-28
[Pal/Linux-SGX] Add Intel AMX to XFRM reg and XSAVE state

dimakuv:
## Description of the changes 

This PR adds the new Intel AMX feature (available on Intel Sapphire Rapids CPU).

This PR also fixes a small inconsistency in how XSAVE area size was calculated in `xsave_size_table` (the size should be calculated as offset of feature + total size of feature).

## How to test this PR? 

Try any test with AMX instructions on an AMX-enabled machine. Without this PR, it will fail with "Illegal instruction". With this PR, the test will succeed.


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/99)* status: 0 of 8 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

___
*[Pal/src/host/Linux-SGX/enclave_xstate.c, line 76 at r1](https://reviewable.io/reviews/gramineproject/gramine/99#-MkgP_pT0iSEnTzsn4r5:-MkgP_pT0iSEnTzsn4r6:b-or6oq7) ([raw file](https://github.com/gramineproject/gramine/blob/5ba0769c9063dd4d2b263bf7ba7c6980237c5457/Pal/src/host/Linux-SGX/enclave_xstate.c#L76)):*
> ```C
>                                                             // header
>         {SGX_XFRM_AVX,    512 + 64 + 256},                  // 256 for YMM0_H - YMM15_H registers
>         {SGX_XFRM_MPX,    512 + 64 + 256 + 256},            // 256 for MPX
> ```

This `256` for MPX was nonsense. The actual size of MPX registers is 128B, and there was a gap of 128B in-between AVX and MPX (you can also see this in a comment in `db_misc.c` above).

This was kinda a bug (which surprisingly resulted in correct math), which cost me several hours of debugging my AMX patch...

chiache (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/99)* status: 0 of 9 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[Pal/src/host/Linux-SGX/db_misc.c, line 234 at r2](https://reviewable.io/reviews/gramineproject/gramine/99#-Mlxs-Kt7FzHvmjsyq_m:-Mlxs-Kt7FzHvmjsyq_n:b-5pqufb) ([raw file](https://github.com/gramineproject/gramine/blob/623b9c92c67e9ae7b59ae0b046b2fbdd995e8d5d/Pal/src/host/Linux-SGX/db_misc.c#L234)):*
> ```C
>                  * SSE state size is implicitly included in the extension's offset, e.g., AVX's
>                  * offset is 576 which includes x87 and SSE state as well as the XSAVE header. */
>                 for (int i = AVX; i <= AMX_2; i++) {
> ```

Can we define some macros like FIRST_EXTENSION and LAST_EXTENSION so that people won't forget to change this line when adding new extensions in the future?
___
*[Pal/src/host/Linux-SGX/enclave_xstate.c, line 81 at r2](https://reviewable.io/reviews/gramineproject/gramine/99#-Mlxq0Eu-oulO-CHeM00:-Mlxq0Eu-oulO-CHeM01:b-iijfr0) ([raw file](https://github.com/gramineproject/gramine/blob/623b9c92c67e9ae7b59ae0b046b2fbdd995e8d5d/Pal/src/host/Linux-SGX/enclave_xstate.c#L81)):*
> ```C
>         {SGX_XFRM_PKRU,   2688 + 8},               // 8 for PKRU register (note the gap!)
>         {SGX_XFRM_AMX,    2752 + 64 + 8192},       // 64 for XTILECFG, 8192 for XTILEDATA
>     };
> ```

This part may be too excessively hardcoding. I understand that all these numbers come from SDM and we need to specify them somewhere, but IMO there is something wrong about: (1) Having the same numbers in two places; (2) hardcoding the offset as the sum of the previous sizes.

Can we find a way of improving this code:
(1) Putting the offset constants into a header file and include it in both db_misc.c and enclave_xstate.c
(2) Calculating this xsave_size_table in a loop instead of hard-coding the total sizes

boryspoplawski (changes requested):
> Reviewed 6 of 8 files at r2, 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/99)* status: all files reviewed, 5 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[Pal/src/host/Linux-SGX/enclave_xstate.c, line 56 at r3](https://reviewable.io/reviews/gramineproject/gramine/99#-MmVO3n05iNj93c7dSov:-MmVO3n05iNj93c7dSow:bwqrdne) ([raw file](https://github.com/gramineproject/gramine/blob/fc0230d2b48de0109abd558f5c30fb70f9f8660b/Pal/src/host/Linux-SGX/enclave_xstate.c#L56)):*
> ```C
> 
> const uint32_t g_cpu_extension_sizes[] = {
>     [AVX] = 256,      [MPX_1] = 64,      [MPX_2] = 64, [AVX512_1] = 64,
> ```

Why such a weird formatting (instead of 1 entry per line)?
___
*[Pal/src/host/Linux-SGX/enclave_xstate.c, line 63 at r3](https://reviewable.io/reviews/gramineproject/gramine/99#-MmVO374BK2jgNX58M8-:-MmVO375B3zt6aNondlN:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/fc0230d2b48de0109abd558f5c30fb70f9f8660b/Pal/src/host/Linux-SGX/enclave_xstate.c#L63)):*
> ```C
>  * PKRU and AMX_1. */
> const uint32_t g_cpu_extension_offsets[] = {
>     [AVX] = 576,       [MPX_1] = 960,     [MPX_2] = 1024, [AVX512_1] = 1088,
> ```

ditto
___
*[Pal/src/host/Linux-SGX/pal_linux.h, line 80 at r3](https://reviewable.io/reviews/gramineproject/gramine/99#-MmVCGkx12piZftAjvDK:-MmVCGkyD5xK67RnwQf4:b-b62xgw) ([raw file](https://github.com/gramineproject/gramine/blob/fc0230d2b48de0109abd558f5c30fb70f9f8660b/Pal/src/host/Linux-SGX/pal_linux.h#L80)):*
> ```objc
> #define DATA_END   ((void*)(&__data_end))
> 
> enum cpu_extension { X87 = 0, SSE, AVX, MPX_1, MPX_2, AVX512_1, AVX512_2, AVX512_3, PKRU = 9,
> ```

Enums starts from `0` by defualt, you don't have to set it.
Also maybe format this like this for readability:
```c
enum cpu_extension {
    X87, SSE, ...
    PKRU = 9,
    AMX_1 = 18, AMX_2,
    LAST_CPU_EXTENSION,
};
```
i.e. group all consecutive options per line?
___
*[Pal/src/host/Linux-SGX/pal_linux.h, line 86 at r3](https://reviewable.io/reviews/gramineproject/gramine/99#-MmVCG5x3e2VeLe2dpM4:-MmVCG5z1uUqYu7CeY8Q:bz201fq) ([raw file](https://github.com/gramineproject/gramine/blob/fc0230d2b48de0109abd558f5c30fb70f9f8660b/Pal/src/host/Linux-SGX/pal_linux.h#L86)):*
> ```objc
> extern const uint32_t g_cpu_extension_offsets[];
> 
> enum register_index { EAX = 0, EBX, ECX, EDX };
> ```

ditto (`0`)

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/99)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[Pal/src/host/Linux-SGX/db_misc.c, line 287 at r5](https://reviewable.io/reviews/gramineproject/gramine/99#-MmXka8X-nA6qhbke8Vr:-MmY-F6E6kfFjZU9xYgj:b33qo2m) ([raw file](https://github.com/gramineproject/gramine/blob/fa07118183bcf55d288d2a22a3bd8cd80e52d374/Pal/src/host/Linux-SGX/db_misc.c#L287)):*

This is theoretically possible. On an older machine (with Intel AMX), the leaves `0x1D` and `0x1E` are considered "not recognized" but still "querable" because there is a higher CPUID leaf (e.g., `0x1F`). In this case, `0x1D` and `0x1E` host CPUID will return the values from this "highest basic information leaf" -- which can be whatever values... That's why I have this guard `extension_enabled(xfrm, AMX_2)`.

Also see my other comment below.
</blockquote></details>

Do we correctly handle the case when enclave disables AMX (in xfrm), but the CPU supports it? Doesn't it return host controlled data to the app directly? In the case above (extended state leaf) we explicitly zero return values on disabled features.
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 300 at r6](https://reviewable.io/reviews/gramineproject/gramine/99#-MmXvn_52OyZkGE5ZSQs:-MmXzh0Z8HAV6OUTZ_Ww:b-kwlpr9) ([raw file](https://github.com/gramineproject/gramine/blob/ec51f7271b75f6b38e77ef50506dde7e10f810eb/Pal/src/host/Linux-SGX/db_misc.c#L300)):*

FYI: didn't change this one and the one below to make it clear that we check 16-bit value against the range `1..256` (note that we allow `256` also) -- I chose this limit specifically because I cannot imagine any AMX hardware with more than 256 tile regs or more than 256 rows in a tile.
</blockquote></details>

But now it looks weird, one decimal among hexes :) Not blocking tho, this can stay

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/99)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[Pal/src/host/Linux-SGX/db_misc.c, line 287 at r5](https://reviewable.io/reviews/gramineproject/gramine/99#-MmXka8X-nA6qhbke8Vr:-MmY6vkoBod1boYwbYxN:b7ghs38) ([raw file](https://github.com/gramineproject/gramine/blob/fa07118183bcf55d288d2a22a3bd8cd80e52d374/Pal/src/host/Linux-SGX/db_misc.c#L287)):*

Cannot app rely on the fact that unrecognized leafs return the highest leaf values (I don't understand this feature, but it's there)? Like query this, see that it differs from highest leaf values -> ok, it must be supported. I mean, this is probably incorrect to do, but for some reason this feature (of returning highest leaf info) is implemented, so something probably uses it.
</blockquote></details>

My point is: in all other cases, even invalid/reserved/not-recognized, we validate/sanitize values, except here.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/99)* status: 9 of 10 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[Pal/src/host/Linux-SGX/db_misc.c, line 411 at r7](https://reviewable.io/reviews/gramineproject/gramine/99#-MmYMqSK9NeSftYHjv3J:-MmYMqSL5dnFBX6i76yE:b-rjmiw2) ([raw file](https://github.com/gramineproject/gramine/blob/16b89618f6dff80253d62b07ec4355016bdc9fac/Pal/src/host/Linux-SGX/db_misc.c#L411)):*
> ```C
>     if (!known_leaf ||
>             (leaf == AMX_TILE_INFO_LEAF && !extension_enabled(xfrm, AMX_2)) ||
>             (leaf == AMX_TMUL_INFO_LEAF && !extension_enabled(xfrm, AMX_2))) {
> ```

FYI: I went through the whole list of CPUID leaves (see our global array above), and -- surprisingly -- the only leaves that depend on any of the XCR0/XFRM features are these two AMX-related leaves. In other words, there are no CPUID leaves controlling the behavior of Intel SSE, MPX, AVX, AVX512, PKRU (MPK in user space).

PR #108, opened 2021-09-30
[Pal/Linux-SGX] Detect signal stack overflow in "enclave_entry.S"

boryspoplawski:
## Description of the changes 
Creating a frame with CPU context on signal stack when handling an exception was done without any bound checks. Malicious host could generate any amount of signals and cause these frames to overflow the stack. This commit fixes it by adding bound checks, effectively disallowing more than ~9 nested exceptions (which should be more than enough in normal case).

## How to test this PR? 
Try running in Gramine with SGX:
```c
#define _GNU_SOURCE
#include <err.h>
#include <sched.h>

int main(void) {
     while (1) {
         if (sched_yield() < 0) {
             err(1, "sched_yield");
         }
     }
}
```
and concurrently `timeout 4 bash -c 'while true; do kill -SIGCONT pid_of_above_program; done'`. This will cause weird behavior on the current master (most likely a crash) and with this PR the process will hang in newly added `FAIL_LOOP`

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/108)

### DIFF
--- a/Pal/src/host/Linux-SGX/enclave_entry.S
+++ b/Pal/src/host/Linux-SGX/enclave_entry.S
@@ -384,6 +384,16 @@ enclave_entry:
 	addq $STACK_FRAME_SUB, %rax
 	subq %rax, %rsi
 
+	# Disallow too many nested exceptions. In normal Gramine flow, this should never happen.
+	# Since addresses need to be canonical, this addition does not overflow.
+	movq %gs:SGX_SIG_STACK_HIGH, %rax
+	addq %gs:SGX_SIG_STACK_LOW, %rax
+	shrq $1, %rax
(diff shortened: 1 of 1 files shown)


mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/108)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/108)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #124, opened 2021-10-04
[Pal/Linux] Exit process in case of errors in `setup_asan`

boryspoplawski:
## Description of the changes 
Previously, if mapping ASAN shadow memory failed, process would just enter an infinite loop. This commit makes it exit the whole process instead.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/124)

### DIFF
--- a/Pal/src/host/Linux/db_main.c
+++ b/Pal/src/host/Linux/db_main.c
@@ -152,8 +152,12 @@ static void setup_asan(void) {
     int flags = MAP_PRIVATE | MAP_ANONYMOUS | MAP_NORESERVE | MAP_FIXED;
     void* addr = (void*)DO_SYSCALL(mmap, (void*)ASAN_SHADOW_START, ASAN_SHADOW_LENGTH, prot, flags,
                                    /*fd=*/-1, /*offset=*/0);
-    if (IS_PTR_ERR(addr))
+    if (IS_PTR_ERR(addr)) {
+        /* We are super early in the init sequence, TCB is not yet set, we probably should not call
+         * any logging functions. */
+        DO_SYSCALL(exit_group, PAL_ERROR_NOMEM);
         die_or_inf_loop();
+    }
 }
 #endif
 


dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/124)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required) (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/124#-MlE7-9O4raSMblMWbOr:-MlE7-9P-zfWJFUdJrWF:b-cd9uts) (no related file):*
We still don't know if this previous `die_or_inf_loop()` was responsible for timeouts in ASan, right? With this PR, we'll at least have some error code if `mmap(<asan shadow memory>)` was responsible for that.

From what I undertood, it is possible that Linux puts the `gramine-direct` executable  somewhere inside "asan shadow memory"? And that's why this `mmap` fails? I still don't get why it can fail.

---

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/124)* status: all files reviewed, 1 unresolved discussion (waiting on @boryspoplawski)

pwmarcz (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/124)* status: all files reviewed, 1 unresolved discussion (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/124#-MlE7-9O4raSMblMWbOr:-MlEtYCs2iqJ0cGKVI1f:bodwot7) (no related file):*

We still don't know if this previous `die_or_inf_loop()` was responsible for timeouts in ASan, right? With this PR, we'll at least have some error code if `mmap(<asan shadow memory>)` was responsible for that.

From what I undertood, it is possible that Linux puts the `gramine-direct` executable  somewhere inside "asan shadow memory"? And that's why this `mmap` fails? I still don't get why it can fail.
</blockquote></details>

I'm not sure exactly where Linux is allowed to map `gramine-direct` (on x86-64, default settings, etc.)

(We allocate the shadow memory at `0x18000000000` (1.5 TB), the range has 16 TB, user part of address space ends at 128 TB. Admittedly this is *not* the default for ASan, the default is `0x7fff8000`, or just before 2 GB - but we can't do that because it will overlap with larger enclaves.)

I tried to trigger this by mapping memory (with the same flags) over the area where program gets usually mapped (`0x700000000000`, len `0x50000000000`). The mmap always succeeds; when it maps over the executable, the process segfaults. Mapping over the area with `[vvar]` and `[vdso]` also doesn't fail.

Another possibility is ASan being triggered in some kind of a loop, or otherwise breaking. The code is not particularly careful about that: the logging functions invoked by ASan handlers are not ASan-exempt. It's also possible ASan is triggered before the logging functions are available, they need relocations to be applied...

---

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/124)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/124)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/124#-MlE7-9O4raSMblMWbOr:-MlFSIl5-rJKc6izNO1Y:b-rfq48i) (no related file):*

I'm not sure exactly where Linux is allowed to map `gramine-direct` (on x86-64, default settings, etc.)

(We allocate the shadow memory at `0x18000000000` (1.5 TB), the range has 16 TB, user part of address space ends at 128 TB. Admittedly this is *not* the default for ASan, the default is `0x7fff8000`, or just before 2 GB - but we can't do that because it will overlap with larger enclaves.)

I tried to trigger this by mapping memory (with the same flags) over the area where program gets usually mapped (`0x700000000000`, len `0x50000000000`). The mmap always succeeds; when it maps over the executable, the process segfaults. Mapping over the area with `[vvar]` and `[vdso]` also doesn't fail.

Another possibility is ASan being triggered in some kind of a loop, or otherwise breaking. The code is not particularly careful about that: the logging functions invoked by ASan handlers are not ASan-exempt. It's also possible ASan is triggered before the logging functions are available, they need relocations to be applied...
</blockquote></details>

@dimakuv We do not know, but this was the only idea we had and this was the only place I've found that does inf-loop without logging or trying to exit earlier.

Yeah, TBH this doesn't look like the place where it could fail, but we still want to merge it (just in case). The question what was wrong unfortunately still stands...

---

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/124)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #139, opened 2021-10-08
[LibOS,Pal/Linux-SGX] Generate offsets using an executable

pwmarcz:
## Description of the changes 

Instead of generating assembly code and parsing it, we compile an executable that outputs desired source (C header file or Python module). As a result, the code is less brittle (no need to account for GCC and Clang assembly format), and some hacks can be removed from Meson. 

## How to test this PR? 

CI should be enough, but you can also look at the generated files (`asm-offsets.h` and `_offsets.py`) in the build directory.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/139)


boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/139)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz and @woju)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/139#-MlVD9Md-QjGP9AiTGqv:-MlVD9Md-QjGP9AiTGqw:b-wwbk5w) (no related file):*
One step closer to sanity...

---
___
*[common/src/generated-offsets-print.c, line 23 at r1](https://reviewable.io/reviews/gramineproject/gramine/139#-MlVCZto5jPYu7aJloIU:-MlVCZtp06HYYRJL9Vmw:bnfpp3j) ([raw file](https://github.com/gramineproject/gramine/blob/367f4ce7a333405c7100f5bba246d0eddd4a27c9/common/src/generated-offsets-print.c#L23)):*
> ```C
>     const struct generated_offset* gen;
>     for (gen = &generated_offsets[0]; gen->name; gen++) {
>         printf("#ifndef %s\n", gen->name);
> ```

Why these guards around each `define`? They don't seem to be useful, just bloat the generated header.
___
*[Pal/include/meson.build, line 18 at r1](https://reviewable.io/reviews/gramineproject/gramine/139#-MlVEjZxEm9CN-bJ7Bp0:-MlVEjZxEm9CN-bJ7Bp1:bhityqr) ([raw file](https://github.com/gramineproject/gramine/blob/367f4ce7a333405c7100f5bba246d0eddd4a27c9/Pal/include/meson.build#L18)):*
> ```Meson
>         join_paths('arch', host_machine.cpu_family(), 'Linux'),
>     ),
>     common_inc,
> ```

Why is this here? Shouldn't this list only things that PAL exposes for LibOS?
___
*[Pal/src/host/Linux-SGX/meson.build, line 31 at r1](https://reviewable.io/reviews/gramineproject/gramine/139#-MlVHEom2XU6PBDi0YOf:-MlVHEom2XU6PBDi0YOg:b-x8a56d) ([raw file](https://github.com/gramineproject/gramine/blob/367f4ce7a333405c7100f5bba246d0eddd4a27c9/Pal/src/host/Linux-SGX/meson.build#L31)):*
> ```Meson
>     include_directories: sgx_inc,
>     c_args: [
>         # no `cflags_pal_common`: we can't use custom stack protector or sanitizers
> ```

Hm, don't we need that `-DIN_PAL` that's stored in `cflags_pal_common`? Afair structs might be defined in a different way, depending on it...

boryspoplawski (approved):
> Reviewed 8 of 8 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/139)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @woju)

___
*[Pal/include/meson.build, line 18 at r1](https://reviewable.io/reviews/gramineproject/gramine/139#-MlVEjZxEm9CN-bJ7Bp0:-MlVUW-V4MZHxEZVJfwn:bh8zrsi) ([raw file](https://github.com/gramineproject/gramine/blob/367f4ce7a333405c7100f5bba246d0eddd4a27c9/Pal/include/meson.build#L18)):*

It's a transitive dependency: the PAL headers require `common_inc` (`api.h` etc.) This is similar to how `common_inc` also contains `tomlc99_inc`: the common module doesn't "expose" the TOML library, but requires it.

At least that's how I think we should do it, but I guess @woju can confirm (or correct me) regarding Meson best practices.
</blockquote></details>

Oh, makes sense. If PAL headers require common stuff, then we should keep it for sure. Maybe it's worth commenting?

boryspoplawski (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/139)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @woju)

mkow (changes requested):
> Reviewed 7 of 10 files at r1, 8 of 8 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/139)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz and @woju)

___
*[common/src/generated-offsets-print.c, line 36 at r3](https://reviewable.io/reviews/gramineproject/gramine/139#-MlbAUFT7XiUPovUZfBw:-MlbAUFT7XiUPovUZfBx:b-k4o7uv) ([raw file](https://github.com/gramineproject/gramine/blob/eade5649e25e9a419590dadb0588fa7310e979e4/common/src/generated-offsets-print.c#L36)):*
> ```C
>     const struct generated_offset* gen;
>     for (gen = &generated_offsets[0]; gen->name; gen++) {
>         printf("%s = %zu\n", gen->name, gen->offset);
> ```

Here and above: could he generate them in hex, for easier manual inspection?
___
*[Pal/src/host/Linux/meson.build, line 55 at r3](https://reviewable.io/reviews/gramineproject/gramine/139#-MlbCb6E4F-KauTBkQbZ:-MlbCb6E4F-KauTBkQb_:b-s6seqj) ([raw file](https://github.com/gramineproject/gramine/blob/eade5649e25e9a419590dadb0588fa7310e979e4/Pal/src/host/Linux/meson.build#L55)):*
> ```Meson
>             cflags_pal_common,
>             cflags_custom_stack_protector,
>             cflags_sanitizers,
> ```

Just out of curiosity: Why pushing the flags down to each PAL?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/139)* status: 11 of 19 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski, @dimakuv, @mkow, and @woju)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/139#-Mli_pap1eITYJf_A9eN:-MlipPbd3JuKL2GYUPKY:b6sgrlw) (no related file):*

I don't like this name `offsets` in filenames. These constants are not only offsets, they can be sizes, masks, etc. But I understand that this will be very cumbersome to fix the filenames. So maybe add a comment in one of the source files about this (`"offsets" is a historic name, this utility actually emits different constants like sizes, masks, offsets, etc.`)?
</blockquote></details>

I tried to replace "offset" with "constant" but somehow I don't like how it looks (`struct generated_constant` etc.) Any other ideas?

I added a TODO for now.

---
___
*[common/src/generated-offsets-print.c, line 36 at r3](https://reviewable.io/reviews/gramineproject/gramine/139#-MlbAUFT7XiUPovUZfBw:-Mliper29HJ27XlRQYWi:b-dc0v5y) ([raw file](https://github.com/gramineproject/gramine/blob/eade5649e25e9a419590dadb0588fa7310e979e4/common/src/generated-offsets-print.c#L36)):*

Here and above: could he generate them in hex, for easier manual inspection?
</blockquote></details>

Done. Some of the values are first defined as decimal, but I agree hex is a better default.
___
*[LibOS/shim/src/arch/x86_64/shim_context.c, line 13 at r4](https://reviewable.io/reviews/gramineproject/gramine/139#-MlixK5XDIUftCknfj4E:-MlixK5YAF1iASkdEYu6:b-hyylaf) ([raw file](https://github.com/gramineproject/gramine/blob/175b1867cbe81250e75d5b8b9b35ba0530a37e7f/LibOS/shim/src/arch/x86_64/shim_context.c#L13)):*
> ```C
> #include <stdnoreturn.h>
> 
> #include "asm-offsets.h"
> ```

FYI: this stopped building when I changed the offsets to hex :) GCC didn't like that I redefined some macros with different content (`128` -> `0x80`).

I think this file should only be included in assembly source, I added an `#error` to ensure that.
___
*[Pal/src/host/Linux/meson.build, line 55 at r3](https://reviewable.io/reviews/gramineproject/gramine/139#-MlbCb6E4F-KauTBkQbZ:-MlipgaE3wDwzCP4LSUV:butnxx1) ([raw file](https://github.com/gramineproject/gramine/blob/eade5649e25e9a419590dadb0588fa7310e979e4/Pal/src/host/Linux/meson.build#L55)):*

Just out of curiosity: Why pushing the flags down to each PAL?
</blockquote></details>

Long story short: because `generated-offsets-print` does not use our custom stack protector or sanitizers. As requested by @dimakuv, I created a separate commit for that.
___
*[Pal/src/host/Linux-SGX/generated-offsets.c, line 31 at r3](https://reviewable.io/reviews/gramineproject/gramine/139#-Mli_DWzDS-B6mLt3r03:-MliuugaBPNJ_WSQx1C8:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/eade5649e25e9a419590dadb0588fa7310e979e4/Pal/src/host/Linux-SGX/generated-offsets.c#L31)):*

These `_HI` and `_LO` masks were split exactly because we had issues with GCC asm output (it outputted a negative number instead of a proper 64-bit unsigned int). Now that we have your sane utility, could you merge the two 32-bit masks into one 64-bit one (for both `FLAGS` and `XFRM`)? Please do it in a separate commit.
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/meson.build, line 31 at r1](https://reviewable.io/reviews/gramineproject/gramine/139#-MlVHEom2XU6PBDi0YOf:-MliwxgD-96hKCoVgykE:bx63ors) ([raw file](https://github.com/gramineproject/gramine/blob/367f4ce7a333405c7100f5bba246d0eddd4a27c9/Pal/src/host/Linux-SGX/meson.build#L31)):*

@mkow This is the answer to your question about pushing stack protector and sanitizers to the level of each PAL.

@pwmarcz Actually, how hard would it be to split this particular change in a separate commit? This PR fixes like 3 different things now... In the worst case, let's at least add a comment in the commit message about these lateral changes.
</blockquote></details>

Makes sense. I created a separate commit.

woju (changes requested):
> Reviewed 4 of 10 files at r1, 2 of 8 files at r2, 1 of 1 files at r3, 1 of 8 files at r4.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/139)* status: 12 of 19 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski, @dimakuv, @mkow, and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/139#-Mli_pap1eITYJf_A9eN:-MlivZ3A9n1DKkYZSZNM:bhdxqzp) (no related file):*

I tried to replace "offset" with "constant" but somehow I don't like how it looks (`struct generated_constant` etc.) Any other ideas?

I added a TODO for now.
</blockquote></details>

"layout"? "config"?

---
___
*[LibOS/shim/src/meson.build, line 140 at r4](https://reviewable.io/reviews/gramineproject/gramine/139#-MlixeH55z5bX62ZOVz2:-MlixeH6E1TubFv0AUVg:b-4lp2z7) ([raw file](https://github.com/gramineproject/gramine/blob/9a06f8374c64d0f3daa5a731d3c0869219e27e4e/LibOS/shim/src/meson.build#L140)):*
> ```Meson
>         # as its underlying PAL which must have a canary in its PAL_TCB at offset 0x8, so no
>         # additional enabling is required in the LibOS code)
>         cflags_custom_stack_protector,
> ```

Plz some whitespace around this, because now you don't know if the comment applies to `cflags_sanitizers` below or not.
___
*[Pal/include/meson.build, line 18 at r1](https://reviewable.io/reviews/gramineproject/gramine/139#-MlVEjZxEm9CN-bJ7Bp0:-Mliy2Gu414Lrc0jX2bH:b5lw3nj) ([raw file](https://github.com/gramineproject/gramine/blob/367f4ce7a333405c7100f5bba246d0eddd4a27c9/Pal/include/meson.build#L18)):*

OK, I added a comment.
</blockquote></details>

Another way is to use dependencies. You can `declare_dependency()` with only includes and another dependencies, so you won't have to reiterate all things inside. 

Yet another way would be to take advantage of the fact that all meson variables are immutable, so this is perfectly fine:

```meson
includes_pal_for_libos = [
    include_pal_common,
    include_directories('elf', join_paths('arch', host_machine.cpu_family(), 'Linux'),
]
```

I don't know if either solution would improve clarity/readability, so your call.

mkow (changes requested):
> Reviewed 8 of 8 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/139)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/139#-Mli_pap1eITYJf_A9eN:-MljA2GE2E9nUlpJF6fd:b-stwddx) (no related file):*

"layout"? "config"?
</blockquote></details>

All these names seem to be worse than just "offsets" to me...

---
___
*[-- commits, line 26 at r4](https://reviewable.io/reviews/gramineproject/gramine/139#-MljAisD1QPAJwKtSTHf):*
`it's` -> `is`
___
*[common/include/generated-offsets-build.h, line 25 at r4](https://reviewable.io/reviews/gramineproject/gramine/139#-MljBvQH49RMgBS5yzuX:-MljBvQH49RMgBS5yzuY:b-nzkpdy) ([raw file](https://github.com/gramineproject/gramine/blob/9a06f8374c64d0f3daa5a731d3c0869219e27e4e/common/include/generated-offsets-build.h#L25)):*
> ```objc
> struct generated_offset {
>     const char* name;
>     size_t value;
> ```

Hmm, I'm not sure if this is a good choice, some constants we use are hardcoded to 64 bits (e.g. some SGX fields). Could you at least statically assert there that they fit into this field? Otherwise we'll silently trim them when compiling for 32-bits.
___
*[common/src/generated-offsets-print.c, line 21 at r4](https://reviewable.io/reviews/gramineproject/gramine/139#-MljBciw1c9_e8AsN7mA:-MljBciw1c9_e8AsN7mB:b-qqw4dc) ([raw file](https://github.com/gramineproject/gramine/blob/9a06f8374c64d0f3daa5a731d3c0869219e27e4e/common/src/generated-offsets-print.c#L21)):*
> ```C
>     printf("\n");
> 
>     printf("#ifndef __ASSEMBLER__\n", generated_offsets_name);
> ```

hmm? is there a `%` missing or was there a copy-paste accident? ;)
___
*[Pal/src/host/Linux/meson.build, line 55 at r3](https://reviewable.io/reviews/gramineproject/gramine/139#-MlbCb6E4F-KauTBkQbZ:-MljCRe80HsporYjEwAE:b-vy5pcl) ([raw file](https://github.com/gramineproject/gramine/blob/eade5649e25e9a419590dadb0588fa7310e979e4/Pal/src/host/Linux/meson.build#L55)):*

Long story short: because `generated-offsets-print` does not use our custom stack protector or sanitizers. As requested by @dimakuv, I created a separate commit for that.
</blockquote></details>

Ah, I see, makes sense.

dimakuv (changes requested):
> Reviewed 5 of 5 files at r6.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/139)* status: 19 of 20 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, @pwmarcz, and @woju)

___
*[common/src/generated-offsets-print.c, line 21 at r4](https://reviewable.io/reviews/gramineproject/gramine/139#-MljBOc4D5p9t0IHpmI4:-MljSNre7lhDkqHldbV3:b-g2w16c) ([raw file](https://github.com/gramineproject/gramine/blob/9a06f8374c64d0f3daa5a731d3c0869219e27e4e/common/src/generated-offsets-print.c#L21)):*

Ouch. So `-fno-builtin` causes the compiler not to check that when we #include a standard library. I think we can do better (maybe even remove `-fno-builtin` everywhere), but I don't want to deal with that now. There's already a TODO in LibOS about removing it.

Fixed.
</blockquote></details>

Not fixed.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/139)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow, @pwmarcz, and @woju)

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/139)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @mkow and @woju)

___
*[-- commits, line 26 at r4](https://reviewable.io/reviews/gramineproject/gramine/139#-MljAisD1QPAJwKtSTHf):*

Will do on final rebase.
</blockquote></details>

Done.

PR #145, opened 2021-10-12
[Meson] Compile `libgomp` (GCC) using Meson

### LINKED ISSUE #132: Move GCC (libgomp) build to Meson
PR #129 moves the main build to Meson, but `LibOS/Makefile` still builds the optional patched `libgomp` library, built together with GCC.

`libgomp` needs to be built with Meson, probably under a feature flag such as `-Dlibgomp=enabled` because the build takes a very long time.

pwmarcz:
## Description of the changes 

This moves `libgomp` compilation from Make to a Meson subproject, and also makes it much faster by disabling the 3-stage build for GCC.

Closes #132.

The GCC build generates huge amounts of output (rather uninteresting, even when running `ninja -v`), so similar to Makefile before it, I capture the logs (see output below). Instead, there is (not buffered) console output from the build script. It uses [Meson's `console: true`](https://mesonbuild.com/Reference-manual_functions.html#custom_target) option which seems to be exactly for this use case:

>Keyword argument conflicts with `capture`, and is meant for commands that are resource-intensive and take a long time to finish. With the Ninja backend, setting this will add this target to [Ninja's `console pool`](https://ninja-build.org/manual.html#_the_literal_console_literal_pool), which has special properties such as not buffering stdout and serializing all targets in this pool.

If you think this is the way to go, maybe I should do the same for `glibc`? I remember @boryspoplawski and @mkow complaining that Ninja output for `glibc` is confusing: it pauses for a long time without displaying that it's compiling `glibc`, and then (if `-v` was specifies) dumps the whole `glibc` compilation output.

## How to test this PR? 

Build with `-Dlibgomp=enabled` (it takes about 16 minutes on my i5 laptop). This is the output from ninja. Note that you can do `tail -f ...gcc-build.log` if you're curious.

```
[0/1] Generating gcc with a custom command
gcc (libgomp): see /home/pawel/gramine/build/subprojects/gcc-10.2.0/gcc-build.log for full build log
gcc (libgomp): preparing sources...
patching file libgomp/config/linux/x86/futex.h
gcc (libgomp): running download_prerequisites...
gcc (libgomp): running configure...
gcc (libgomp): running make...
gcc (libgomp): done
```

I also checked if the `openmp` example still works, and verified (using `objdump -d`) that `libgomp.so.1` does not contain raw `SYSCALL` instructions.


boryspoplawski (commented):
> if you need the full log from build.

I would argue that glibc and Gramine are two different builds (just done together for simplicity and because we always need them both).

Generally I'm fully in favor of having separate log, at least for glibc build. It would be nice to have (tho it's not required) to also have errors from such builds printed directly to console (we used to have something like that in makefiles, idk how it was done, probably just tee stderr).

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/145#-:-Mlyth0SBrKijNnzpISw:b-x4h744)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

woju (dismissed):
> glibc-build.log has 15.5 MB, btw :)  And gcc-build.log has 1.6 MB.

Yeah, OK. I can see the rationale here.

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/145#-:-MlzVB9v6KRUD4RRo1jc:bqasl6b)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

PR #146, opened 2021-10-13
[LibOS] Add minimal emulation of `sysinfo()`

dimakuv:
## Description of the changes 

This minimal implementation was required by some private app.

## How to test this PR? 

~~Unfortunately, the implementation is too minimal to enable any new LTP tests.~~

Enabled two LTP tests (`sysinfo01` and `sysinfo02`). Unfortunately, `sysinfo03` tests unemulated stuff.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/146)


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/146)* status: 4 of 5 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @pwmarcz)

___
*[LibOS/shim/src/sys/shim_getrlimit.c, line 9 at r2](https://reviewable.io/reviews/gramineproject/gramine/146#-Mlt48-h1m3OV15DzMIy:-Mlt6qlG1cP_hK6jjGvd:b-inh3fl) ([raw file](https://github.com/gramineproject/gramine/blob/dd7d3153851854ea8f08b81dc5aebfd924bc38c7/LibOS/shim/src/sys/shim_getrlimit.c#L9)):*

Please do not include stuff from `sys`.
The stuff you are looking for is in `<linux/sysinfo.h>`
</blockquote></details>

Done. Actually, what is the rule of thumb with Linux headers?

pwmarcz (approved):
> Reviewed 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/146)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

chiache (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/146)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[LibOS/shim/src/sys/shim_getrlimit.c, line 170 at r3](https://reviewable.io/reviews/gramineproject/gramine/146#-Mm-GPZk5OGSChPRhye_:-Mm-M3bw80Q_FoknHjUO:b-9avi5l) ([raw file](https://github.com/gramineproject/gramine/blob/d3dfa8e63a39ef17f65abc6995d5a9e6ec2ea0af/LibOS/shim/src/sys/shim_getrlimit.c#L170)):*

To be honest, I haven't seen any application quering uptime for anything interesting. What's the point in adding complexity that is never used by apps?
</blockquote></details>

That's a fair point. Partially, there is some benefit to be able to claim complete backward compatibility--even system calls that are not used in an obvious way (and how do we know for sure that nobody will use this?). But on the other hand, it's important to prioritize the more important, popular system calls and return values.

My point is, suppose if we can implement this easily without too much complexity, wouldn't it be better to implement it than to not implement it?
___
*[LibOS/shim/src/sys/shim_getrlimit.c, line 176 at r3](https://reviewable.io/reviews/gramineproject/gramine/146#-Mm-Gabe3GwHCZujwCN0:-Mm-MobNEJrfD2zjhznB:b-or9sdc) ([raw file](https://github.com/gramineproject/gramine/blob/d3dfa8e63a39ef17f65abc6995d5a9e6ec2ea0af/LibOS/shim/src/sys/shim_getrlimit.c#L176)):*

Same logic here -- I haven't seen applications making any decisions based on the number of processes currently running.

The only thing I observed that applications want from this syscall is "how much free RAM we still have?". Then the app may decide to kill itself.
</blockquote></details>

Similarly, I guess it's hard to know for sure that nobody actually uses this feature.

pwmarcz (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/146)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @chiache, and @dimakuv)

___
*[LibOS/shim/src/sys/shim_getrlimit.c, line 170 at r3](https://reviewable.io/reviews/gramineproject/gramine/146#-Mm-GPZk5OGSChPRhye_:-Mm2LspP29v_wN9U2-Q7:bp4kfi2) ([raw file](https://github.com/gramineproject/gramine/blob/d3dfa8e63a39ef17f65abc6995d5a9e6ec2ea0af/LibOS/shim/src/sys/shim_getrlimit.c#L170)):*

That's a fair point. Partially, there is some benefit to be able to claim complete backward compatibility--even system calls that are not used in an obvious way (and how do we know for sure that nobody will use this?). But on the other hand, it's important to prioritize the more important, popular system calls and return values.

My point is, suppose if we can implement this easily without too much complexity, wouldn't it be better to implement it than to not implement it?
</blockquote></details>

This one sounds easier than the other one, but still not exactly free in terms of complexity: we would need to check and store this timestamp in the initial process, then make sure it's migrated to other processes.
___
*[LibOS/shim/src/sys/shim_getrlimit.c, line 176 at r3](https://reviewable.io/reviews/gramineproject/gramine/146#-Mm-Gabe3GwHCZujwCN0:-Mm2LHPOAgVs8g4UGSXa:b-7j8oit) ([raw file](https://github.com/gramineproject/gramine/blob/d3dfa8e63a39ef17f65abc6995d5a9e6ec2ea0af/LibOS/shim/src/sys/shim_getrlimit.c#L176)):*

Similarly, I guess it's hard to know for sure that nobody actually uses this feature. 
</blockquote></details>

I don't think it's worth the complexity. Recently, we stopped supporting a similar feature (being able to list remote PIDs in `/proc/`: now we just allow the user to access `/proc/<pid>/` directly if you know the PID) in the process of IPC refactoring. (CC @boryspoplawski)

mkow (approved):
> Reviewed 3 of 4 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/146)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required) (waiting on @dimakuv and @pwmarcz)

___
*[LibOS/shim/src/sys/shim_getrlimit.c, line 176 at r3](https://reviewable.io/reviews/gramineproject/gramine/146#-Mm-Gabe3GwHCZujwCN0:-Mm3sgcP5k6jtjXWeM-L:b1jaec3) ([raw file](https://github.com/gramineproject/gramine/blob/d3dfa8e63a39ef17f65abc6995d5a9e6ec2ea0af/LibOS/shim/src/sys/shim_getrlimit.c#L176)):*

I agree that this is probably more complex than we can handle for now.  But in case there is some creative, easy way of implementing this, I would really love to see this feature implemented. After all, I do feel a little uncomfortable seeing that we are providing a dummy, incorrect value without any possibility of warning the users.
</blockquote></details>

@chiache Basically, AFAIK this would require a lot of IPC to get this number. I also don't like that most of the values here are not true, but there's no easy way to implement them, plus it's hard for me to imagine an app using these for something else than statistics / performance decisions.

PR #149, opened 2021-10-15
[tests] Fix warnings

pwmarcz:
## Description of the changes 

This is in preparation for compiling the tests with Meson (which uses `-Wextra` and a few additional warnings). All Meson's warnings are fixed, except for `Wunused-parameter`, there's just too many of them to bother (we define most `main()` functions with `argc, argv, envp` and end up not using them).

## How to test this PR? 

* CI
* Read carefully

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/149)

### DIFF
--- a/LibOS/shim/test/regression/Makefile
+++ b/LibOS/shim/test/regression/Makefile
@@ -138,6 +138,9 @@ extra_rules = \
 include ../../../../Scripts/manifest.mk
 include ../../../../Scripts/Makefile.Test
 
+# TODO: Many files define `main` with parameters, but do not use them.
+CFLAGS += -Wno-unused-parameter
+
 CFLAGS-abort_multithread = -pthread
 CFLAGS-bootstrap_pie = -fPIC -pie
 CFLAGS-bootstrap_static = -static
--- a/Pal/regression/Makefile
+++ b/Pal/regression/Makefile
@@ -10,6 +10,9 @@ CFLAGS	+= -fno-stack-protector -fno-builtin -nostdlib \
 	  -I../include/arch/$(ARCH) \
 	  -I../include/arch/$(ARCH)/$(PAL_HOST)
 
+# TODO: Many files define `main` with parameters, but do not use them.
+CFLAGS += -Wno-unused-parameter
+
 preloads = \
 	Preload1.so \
 	Preload2.so
--- a/Scripts/Makefile.configs
+++ b/Scripts/Makefile.configs
@@ -61,8 +61,8 @@ export OPTIMIZE
 export UBSAN
 export ASAN
 
-CFLAGS += -Wall -std=c11 -Wmissing-prototypes
-CXXFLAGS += -Wall -std=c++14
+CFLAGS += -std=c11 -Wall -Wextra -Wmissing-prototypes -Wstrict-prototypes -Wwrite-strings
+CXXFLAGS += -std=c++14 -Wall -Wextra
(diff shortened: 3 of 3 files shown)


pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/149)* status: all files reviewed, 12 unresolved discussions (waiting on @boryspoplawski and @pwmarcz)

___
*[LibOS/shim/test/regression/exec.c, line 11 at r1](https://reviewable.io/reviews/gramineproject/gramine/149#-Mm447e26JC_ueWBuY4I:-MmHvB8c2K6Vy4yYBmBQ:b-om4014) ([raw file](https://github.com/gramineproject/gramine/blob/45837e7963ee0b8f2c951d4b9303df2374eb726d/LibOS/shim/test/regression/exec.c#L11)):*

I think in very old C string literals were "guaranteed" to be writable, and this example is from these times.

@boryspoplawski: Any thoughts on this?
</blockquote></details>

[Here's what I found](https://stackoverflow.com/a/190208), apparently yes, POSIX guarantees the arguments are guaranteed not to change but the type is for compatibility reasons.

boryspoplawski (commented):
> Reviewed 28 of 41 files at r1.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/149)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[LibOS/shim/test/regression/exec.c, line 11 at r1](https://reviewable.io/reviews/gramineproject/gramine/149#-Mm447e26JC_ueWBuY4I:-MmI_9Vk85fxMtMn5fU-:b-lz4lb1) ([raw file](https://github.com/gramineproject/gramine/blob/45837e7963ee0b8f2c951d4b9303df2374eb726d/LibOS/shim/test/regression/exec.c#L11)):*
> I think in very old C string literals were "guaranteed" to be writable, and this example is from these times.

I don't think it was ever the case. Afaik string literals were always like: if you write to them it's UB (even in pre-const times) even tho the type is `char*`.

The problem is that you cannot auto convert `char **` to `const char **`, so code using `char* argv[] = { something }; execve(argv[0], argv, envp);` would be illegal.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/149)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[LibOS/shim/test/regression/exec.c, line 11 at r1](https://reviewable.io/reviews/gramineproject/gramine/149#-Mm447e26JC_ueWBuY4I:-MmIkXMp-T_MzGQ_BTd3:bt8mm44) ([raw file](https://github.com/gramineproject/gramine/blob/45837e7963ee0b8f2c951d4b9303df2374eb726d/LibOS/shim/test/regression/exec.c#L11)):*

> What I wanted to say is that there was a C version which did this by default,

And my response was that it is the **current** version (all of them to be specific) that does this.
</blockquote></details>

Oh, ok, it actually still is non-const by default, I guess they've changed it only in C++...

PR #150, opened 2021-10-16
[LibOS] Assign UID and GID to stat fields in chroot

dzygann:
Set the UID and GID in the chroot function to the one defined in the manifest
by using the `loader.uid` and `loader.gid` options. Otherwise, a check fails
from the wrapped application, because the UID/GID is set to 0 and the application
expects another one.

This topic has been discussed in the issue [2632](https://github.com/gramineproject/graphene/issues/2632)

Signed-off-by: Denis Zygann <denis@zygann.de>

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/150)

### DIFF
--- a/Documentation/manifest-syntax.rst
+++ b/Documentation/manifest-syntax.rst
@@ -208,7 +208,8 @@ User ID and Group ID
 This specifies the initial, Gramine emulated user/group ID and effective
 user/group ID. It must be non-negative. By default Gramine emulates the
 user/group ID and effective user/group ID as the root user (uid = gid = 0).
-The user/group ID affects also the stat members for ``tmpfs`` and ``chroot``.
+The user and group IDs are also reflected in the information about ``tmpfs``
+and ``chroot`` files returned via ``stat()``.
 
 
 Disabling ASLR
--- a/Documentation/manifest-syntax.rst
+++ b/Documentation/manifest-syntax.rst
@@ -208,6 +208,7 @@ User ID and Group ID
 This specifies the initial, Gramine emulated user/group ID and effective
 user/group ID. It must be non-negative. By default Gramine emulates the
 user/group ID and effective user/group ID as the root user (uid = gid = 0).
+The user/group ID affects also the stat members for ``tmpfs`` and ``chroot``.
 
 
 Disabling ASLR
--- a/LibOS/shim/src/fs/chroot/fs.c
+++ b/LibOS/shim/src/fs/chroot/fs.c
@@ -341,11 +341,14 @@ out:
 }
 
 static int chroot_istat(struct shim_inode* inode, struct stat* buf) {
+    struct shim_thread* cur_thread = get_cur_thread();
     memset(buf, 0, sizeof(*buf));
 
     lock(&inode->lock);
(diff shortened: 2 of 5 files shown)


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/150)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dzygann and @pwmarcz)

___
*[Documentation/manifest-syntax.rst, line 211 at r2](https://reviewable.io/reviews/gramineproject/gramine/150#-MmMSk7s1guo1w07BF-V:-MmMSk7s1guo1w07BF-W:bw7360o) ([raw file](https://github.com/gramineproject/gramine/blob/d30ed4f82ea6d00c2191fb838bb6fbe7cfe88452/Documentation/manifest-syntax.rst#L211)):*
> ```reStructuredText
> user/group ID. It must be non-negative. By default Gramine emulates the
> user/group ID and effective user/group ID as the root user (uid = gid = 0).
> The user/group ID affects also the stat members for ``tmpfs`` and ``chroot``.
> ```

Maybe `The user and group IDs are also reflected in the information about "tmpfs" and "chroot" files returned via stat().`
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 167 at r2](https://reviewable.io/reviews/gramineproject/gramine/150#-MmKf09wC6EEks_CKoW_:-MmMTmAC3wFzkQdsxla6:bbnvr3n) ([raw file](https://github.com/gramineproject/gramine/blob/d30ed4f82ea6d00c2191fb838bb6fbe7cfe88452/LibOS/shim/src/fs/tmpfs/fs.c#L167)):*
<details><summary><i>Previously, dzygann (Denis Zygann) wrote…</i></summary><blockquote>

How can I test this one in Gramine?
</blockquote></details>

In your test `uid_gid.c`, you can open any file under `/mnt/tmpfs` (for example, `open("/mnt/tmpfs/file1", O_RDWR | O_CREAT, S_IRUSR | S_IWUSR)`). After you opened this file, you can do `stat("/mnt/tmpfs/file1", &buf)` and perform the same check as for the `argv[0]` case.

Note that this created file (`/mnt/tmpfs/file1`) is a temporary file and is only created inside enclave memory. So there will be no corresponding file on your host system, don't be surprised.
___
*[LibOS/shim/test/regression/uid_gid.c, line 31 at r1](https://reviewable.io/reviews/gramineproject/gramine/150#-MmHzbH916O_hngphn3s:-MmMUj0d0eef7TpVN7Ai:btih660) ([raw file](https://github.com/gramineproject/gramine/blob/21921e07bbc2fcb65f2bf4302d4c7855c9e3cb16/LibOS/shim/test/regression/uid_gid.c#L31)):*
<details><summary><i>Previously, dzygann (Denis Zygann) wrote…</i></summary><blockquote>

Done.
If I understood it correctly the`err` appends an error message and `errx` doesn't.
</blockquote></details>

Yes, exactly.

PR #162, opened 2021-10-20
New shiny printf implementation

boryspoplawski:
## Description of the changes 
Our old printf implementation hurt my eyes.

## How to test this PR? 
New test was added. If you wish to verify it with an external implementation you can do:
```
[borys@mowmiwuju regression]$ git diff
diff --git a/Pal/regression/printf_test.c b/Pal/regression/printf_test.c
index bac3846e..cb595845 100644
--- a/Pal/regression/printf_test.c
+++ b/Pal/regression/printf_test.c
@@ -1,10 +1,11 @@
-#include "api.h"
-#include "pal_regression.h"
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
 
 #define FAIL(code, fmt...) ({   \
-    pal_printf(fmt);            \
-    pal_printf("\n");           \
-    DkProcessExit(code);        \
+    printf(fmt);                \
+    printf("\n");               \
+    exit(code);                 \
 })
 
 #define TEST(output_str, fmt...) ({                                                                \
@@ -81,27 +82,6 @@ int main(void) {
     TEST("  0x001337", "%#10.6x", 0x1337);
     TEST("0x001337  ", "%#-10.6x", 0x1337);
 
-    /* This one differs from the "normal" printf, but as far as I can tell, it matches the C
-     * standard. */
-    TEST("000x1337", "%#08x", 0x1337);
-
-    char* ptr = NULL;
-    int ret = DkVirtualMemoryAlloc((void**)&ptr, 2 * PAGE_SIZE, PAL_ALLOC_INTERNAL,
-                                   PAL_PROT_READ | PAL_PROT_WRITE);
-    if (ret < 0) {
-        FAIL(1, "DkVirtualMemoryAlloc failed: %d", ret);
-    }
-    ret = DkVirtualMemoryProtect(ptr + PAGE_SIZE, PAGE_SIZE, PAL_PROT_NONE);
-    if (ret < 0) {
-        FAIL(1, "DkVirtualMemoryProtect failed: %d", ret);
-    }
-    memset(ptr + PAGE_SIZE - 7, 'a', 7);
-
-    ret = snprintf(ptr, PAGE_SIZE - 8, "%.7s", ptr + PAGE_SIZE - 7);
-    if (ret != 7) {
-        FAIL(1, "snprintf at %d returned %d, expected %d", __LINE__, ret, 7);
-    }
-
-    pal_printf("TEST OK\n");
+    printf("TEST OK\n");
     return 0;
 }
[borys@mowmiwuju regression]$ gcc -Wall -Wextra -std=c11 printf_test.c -o printf_test
[borys@mowmiwuju regression]$ ./printf_test 
TEST OK
```


boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/162)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

___
*[common/src/printf.c, line 16 at r2](https://reviewable.io/reviews/gramineproject/gramine/162#-MmRyalL35OK21B9d9Fz:-MmSNSjZ26m0t8o2zadA:b-8823vd) ([raw file](https://github.com/gramineproject/gramine/blob/aeffa3b1f947edd7897edbe66d2a0f7485989882/common/src/printf.c#L16)):*

Why do you need these undefs? What can define them?
</blockquote></details>

I was wondering that myself, when I've seen this in the old code, probably deserves a comment.
Turns out that our "api_fortified.h" defines them, so that another version (checking arguments etc) is used.
___
*[common/src/printf.c, line 39 at r2](https://reviewable.io/reviews/gramineproject/gramine/162#-MmRs1I64ZBGPoPIEkrS:-MmSPW5H0fte90Os7T9E:b-aunvxb) ([raw file](https://github.com/gramineproject/gramine/blob/aeffa3b1f947edd7897edbe66d2a0f7485989882/common/src/printf.c#L39)):*

Do you want to explain somewhere that `write_callback()` must return 0 on success and negative error code? Because it is not obvious that `write_callback()` must print all `size` chars from the buffer (i.e., it cannot print only part of the buffer).

I would create a typedef for this and add a comment.
</blockquote></details>

Described in the comment near `vprintf_core`.
I did not add a typedef, it's only used in two places, but I can if you insist.
___
*[common/src/printf.c, line 246 at r2](https://reviewable.io/reviews/gramineproject/gramine/162#-MmRtjKX6c35YPRKv3iu:-MmSPTbjCP5kUxfCg2zQ:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/aeffa3b1f947edd7897edbe66d2a0f7485989882/common/src/printf.c#L246)):*

Typo: `possible`
</blockquote></details>

Done.
___
*[common/src/printf.c, line 373 at r2](https://reviewable.io/reviews/gramineproject/gramine/162#-MmRxmgu4KR2ie6RDqZs:-MmSdCwK1ZQ7JGL5xV5d:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/aeffa3b1f947edd7897edbe66d2a0f7485989882/common/src/printf.c#L373)):*

We use `pos` in `print_buf` struct. Can we also have `pos` here, for uniformity?
</blockquote></details>

Done.
___
*[common/src/printf.c, line 389 at r2](https://reviewable.io/reviews/gramineproject/gramine/162#-MmRxLe_AC7xgUB5te5A:-MmSceC2--ycQ2PWoOXU:b-7gibd4) ([raw file](https://github.com/gramineproject/gramine/blob/aeffa3b1f947edd7897edbe66d2a0f7485989882/common/src/printf.c#L389)):*

What's the benefit of working on `buf_size == 0`? Can we immediately `return 0` in this case? Or you do this because we need to return error code if `fmt/ap` are weird?
</blockquote></details>

We need to calculate the result length. Note that `snprintf` returns the total length that would have been printed if there was enough space (regardless whether there actually is).
___
*[common/src/stdlib/printfmt.c, line 216 at r2](https://reviewable.io/reviews/gramineproject/gramine/162#-MmRyPbk3DBRa78jk0ey:-MmSdfjT8E_URssdE33b:b7s1ape) ([raw file](https://github.com/gramineproject/gramine/blob/ba4a5ffd6f2b63ad160f760b804d37e40c0f82c6/common/src/stdlib/printfmt.c#L216)):*

What was this one? Some non-standard escaping?
</blockquote></details>

F. if I know
___
*[Pal/regression/printf_test.c, line 103 at r2](https://reviewable.io/reviews/gramineproject/gramine/162#-MmRqh0A3-2oWLhXVlfF:-MmSf5hH0382LesfxeHK:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/aeffa3b1f947edd7897edbe66d2a0f7485989882/Pal/regression/printf_test.c#L103)):*

Do I understand correctly that you do all this magic just to catch a segfault if you printf implementation is incorrect? (Because you disallow access to the second mapped page.) Could you add a quick comment on this?
</blockquote></details>

Done.
___
*[Pal/regression/utils.c, line 42 at r1](https://reviewable.io/reviews/gramineproject/gramine/162#-MmQT0QS7wN-qFc9j8-B:-MmSf8pgDyXSx_xl_REM:b-ywj37v) ([raw file](https://github.com/gramineproject/gramine/blob/a4e5f9912346ed434d33ade6a4190645a3c1b677/Pal/regression/utils.c#L42)):*

I don't get it. What was the original purpose of this `malloc()`? Why would we ever need this? In which scenarios we should use it instead of PAL-provided `DkVirtualMemoryAlloc()`?
</blockquote></details>

`DkVirtualMemoryAlloc` allocates whole pages, but I guess in the simple scenarios that's not a problem. I can rewrite this to use `DkVirtualMemoryAlloc`.

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/162)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[Pal/regression/utils.c, line 42 at r1](https://reviewable.io/reviews/gramineproject/gramine/162#-MmQT0QS7wN-qFc9j8-B:-MmT-o8w28mwnM70OhsI:b-z63ciw) ([raw file](https://github.com/gramineproject/gramine/blob/a4e5f9912346ed434d33ade6a4190645a3c1b677/Pal/regression/utils.c#L42)):*

Because `printf` might use them (if precision in format string is high enough).
I just realized: this version is super useless, as it limits the size to 0x100. I've added it when `printf` did not have an inline buffer, which changed at the last stage of implementation/testing.
Changed so that we have such symbols (it won't compile otherwise), but disabled them at runtime.
</blockquote></details>

Ah, understood now. Resolving.

pwmarcz (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/162)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

PR #169, opened 2021-10-23
[WIP][Pal/Linux] Add a seccomp policy to emulate inline syscalls

boryspoplawski:
## Description of the changes 
This commits adds a seccomp policy which traps inline syscalls made by user application and redirects them to LibOS so they can be properly emulated.

## How to test this PR? 
Enabled a test which does inline `syscall` instruction.

### DIFF
--- a/LibOS/shim/test/regression/syscall.c
+++ b/LibOS/shim/test/regression/syscall.c
@@ -2,13 +2,20 @@
 #include <err.h>
 #include <stdlib.h>
 #include <sys/syscall.h>
-#include <unistd.h>
 
 int main(int argc, char** argv) {
     const char buf[] = "Hello world\n";
-    long ret = syscall(__NR_write, 1, buf, sizeof(buf) - 1);
+    long ret = -1;
+#ifdef __x86_64__
+    __asm__ volatile (
+        "syscall"
+        : "=a"(ret)
+        : "0"(__NR_write), "D"(1), "S"(buf), "d"(sizeof(buf) - 1)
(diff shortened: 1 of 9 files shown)


boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/169)* status: 0 of 9 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

___
*[Pal/src/host/Linux-common/debug_map.c, line 173 at r1](https://reviewable.io/reviews/gramineproject/gramine/169#-MmepQNRBsoywG01h_bJ:-MmepQNS7JW8NUDi9PQi:b-an9cil) ([raw file](https://github.com/gramineproject/gramine/blob/581ee1052b747fb450c629152464692cae2b0ed5/Pal/src/host/Linux-common/debug_map.c#L173)):*
> ```C
> /* Example output: "func_name at source_file.c:123" */
> static int run_addr2line(const char* name, uintptr_t offset, char* buf, size_t buf_size) {
>     return -1; // TODO: FIX
> ```

Once we discuss this PR as a whole, we need to decide what to do with this. I see the following options:
1) Use some library instead of `addr2line`. Probably not trivial - libraries usually expect stdlib env, so probably would require patching.
2) Spawn a helper daemon, which would listen on requests from Gramine processes and run `addr2line` on demand. Spawning would happen before seccomp applying.
3) Do not use this at all and restore old behavior or just printing file + address/instruction offset.

Issue #161, opened 2021-10-19
RFC: building tests in Meson


pwmarcz:
## The problem

We would like to move compiling (and possibly running) tests from Makefile to Meson. This is the last part of old Makefiles remaining. However, the transition is difficult for the following reasons:

1. **Building the manifests depends on Gramine being installed** (not only built). That makes it hard to build the tests (binaries + manifests) together with Gramine.

  * Python tools (`gramine-manifest`, `gramine-sgx-sign`) currently require installing
  * SGX manifests depend on the target paths for PAL and LibOS libraries
  * SGX manifests depend on *hashes* of PAL and LibOS libraries, and these hashes change after installing by Meson (Meson adds `rpath` to libraries in the build dir, and strips them during install).

2. **Test binaries depend on Gramine internals** that we don't install. That makes it hard to build the test binaries separately from Gramine.

  * PAL tests use large parts of the `common` library
  * PAL tests use the public PAL interface (`pal.h`)
  * LibOS tests use `mbedtls` (which we install, but without header files)
  * LibOS `attestation` interface test uses `sgx_arch.h`
  
## Wishlist

Some use cases that should be easy:

* **Rebuild Gramine and/or tests**. Ideally, that should be 1-2 commands, after which there are no problems with outdated manifests etc.
* **Run the tests, or a single test, from Python**. This is useful when working with Pytest sources (coming up with the right assertions, etc.)
* **Run a single test binary from command line**, without Python/Pytest. This is indispensable for debugging (GDB, strace, different command line args, etc.)

## Solution ideas

Some of these were discussed with @woju and @mkow already, but I don't think we had a conclusion.

### Start by compiling binaries only, leave manifests to Make

The test binaries are relatively easy to compile in Meson. I have a branch that compiles all test binaries in Meson, and installs them in `/lib/.../graphene/tests/`.

The rest (compiling manifests + running tests) is still handled by Makefiles. In order to run tests, you need to first `ninja install`, then `make regression`.

The upside is that Makefiles no longer compile C code. But is it worth it to merge a temporary solution like this?

### Compile binaries only, build manifests on demand

Since it's problematic to build manifests together with everything else, don't build them! Build them in Python tests when they're necessary. Introduce a utility (`gramine-rebuild-manifest`) or wrapper (`gramine-rebuild-manifest-or-run`)?

I find this idea problematic for several reasons:

[...]

Issue #16, opened 2021-04-15
[RFC] Standard installation paths


woju:
# Statement of the problems

Those are problems that we need to be aware when solving

## Stable paths
After signing the manifest, paths are baked in, so Graphene needs a standardised set of installed paths. The immediate problem concerns `libsysdb.so` and libc. Paths need to be the same across distributions.

## Multiple libcs installed together
There is an ongoing proposal to make graphene work with executables compiled against musl: gramineproject/graphene#2269. It's a choice of enclave developer, which libc will be used (probably using `fs.mount.` manifest option), so we need to make multiple libcs available concurrently.

## Multiple versions installed together, for signing
`libpal.so` is needed to calculate the expected measurement of the enclave. Also, loader might change between versions. So for each Graphene version, we need to install `libpal.so` and the respective python script which emulates loading for the purpose of signing. This signing script will be discovered/loaded/invoked by common `graphene-sgx-sign` command.

# Proposed solution

## PAL, LibOS, runtime

```
/usr/lib/graphene-<version>/direct/libpal.so
/usr/lib/graphene-<version>/direct/loader
/usr/lib/graphene-<version>/sgx/libpal.so
/usr/lib/graphene-<version>/sgx/loader
/usr/lib/graphene-<version>/libsysdb.so
/usr/lib/graphene-<version>/runtime/{glibc,musl,...}-<version>/*.so
```

TODO: gdb, headers, -dev packages, 

## Command line tools

Those might be ELFs, python scripts and/or shell scripts. I'm not sure if those need to be standardised. Possibly related: gramineproject/gramine#13. 
```
/usr/bin/graphene-direct
/usr/bin/graphene-sgx
/usr/bin/graphene-sgx-sign
/usr/bin/graphene-sgx-get-token
/usr/bin/graphene-manifest
# and possibly others
```

# Package names

Packages (whether deb, rpm or else) should be named:

## `graphene`

Metapackage depending on `graphene-direct` and `graphene-sgx`.

## `graphene-direct`

Non-SGX Graphene (`libpal.so`, `loader`, `/usr/bin/` loader script, manpage).

## `graphene-sgx`

metapackage depending on the latest version

## `graphene-sgx-1.23`

Graphene SGX compiled against upstream 5.11+ driver.

## `graphene-sgx-oot-1.23`

Graphene SGX compiled against SDK driver.

## `graphene-sgx-dcap-1.23`

[...]

PR #182, opened 2021-10-29
Python test runner

pwmarcz:
## Description of the changes 

This PR converts building test manifest from Makefiles to our own Python wrapper, `gramine-test`. See #161 for discussion.

## Why Ninja?

The wrapper uses **Ninja** internally, i.e. generates `build.ninja` and runs Ninja to build manifests. I believe I have good reasons for prefering Ninja over a completely custom Python script:

* Ninja handles concurrent building
* Ninja tracks dependencies, i.e. rebuilds only what is necessary
* It's very fast! (see below)
* Most importantly (IMO), it forces us to split the layers, i.e. invoke `gramine-manifest` instead of being tempted to call a Python function.

On the opposite end of the spectrum, there's implementing everything in Meson, but I think would not be much simpler (the part that generates Ninja would translate to `custom_target`s in Meson), and the end result would be less friendly for the user.

## LTP

LTP still uses the Makefiles. My plan for LTP is to treat is the same as other tests: move building the binaries to Meson, and building manifests/running to `gramine-tests`.

When that's done, we'll be able to (finally!) remove all these old scripts: `Makefile.Test`, `manifest.mk`, etc.

## Benchmark

I did a test on an 8-core machine, in `LibOS/shim/test/regression`.

`master` (3db1e28e27622cbf1265650cc92a87281803d31d):

```
$ make clean; time make -j8
real	0m17.696s
user	0m32.239s
sys	0m3.517s

$ make SGX=1 clean; time make -j8 SGX=1 all sgx-tokens
real	0m44.426s
user	3m50.492s
sys	0m16.938s
```

This PR:

```
$ gramine-test clean; time gramine-test build
real	0m4.117s
user	0m28.205s
sys	0m2.719s

$ gramine-test --sgx clean; time gramine-test --sgx build
real	0m26.453s
user	3m13.935s
sys	0m13.083s
```

Also, reducing the static PAL buffer (#192) makes that even faster:
```
$ gramine-test --sgx clean; time gramine-test --sgx build
real	0m18.653s
user	2m12.166s
sys	0m13.671s
```

## How to test this PR? 

See how it works for you. Try using the `gramine-test` utility.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/182)


pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/182)* status: 0 of 37 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

___
*[LICENSE.addendum.txt, line 12 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-MnWKEap53V7O7tcnpR2:-MnWKEap53V7O7tcnpR3:b-dxy47y) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/LICENSE.addendum.txt#L12)):*
> ```help
> tomlc99 - MIT
> uthash - BSD revised
> Ninja (python/graminelibos/ninja_syntax.py) - Apache 2.0
> ```

FYI: This file is a part of Ninja source distribution (https://github.com/ninja-build/ninja/blob/master/misc/ninja_syntax.py), but AFAICT, it's not packaged by Linux distributions. There is an [unofficial package on PyPI](https://pypi.org/project/ninja_syntax/) but it looks abandoned, and [another package that provides the whole Ninja](https://pypi.org/project/ninja/), not only that file.

Overall, I decided it's better to vendor this file, but maybe you'll have a better suggestion.
___
*[LibOS/shim/test/fs/Makefile, line 10 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-MnWKwqc2Lt-B16yCzXQ:-MnWKwqc2Lt-B16yCzXR:bwbnb1h) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/LibOS/shim/test/fs/Makefile#L10)):*
> ```Makefile
> # TODO: Python tests should not produce leftovers
> .PHONY: clean
> clean:
> ```

FYI: `gramine-test` cleans the files that it has built. What remains is the files created by tests themselves.
I want the Python tests to clean up after themselves (i.e. run each test from a temporary directory), but until they do, it's easiest to leave these `make clean` targets.
___
*[LibOS/shim/test/regression/attestation.manifest.template, line 24 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-MnWLWK8Csrhrqarp2r_:-MnWLWK8Csrhrqarp2ra:bkuq9g4) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/LibOS/shim/test/regression/attestation.manifest.template#L24)):*
> ```
> sgx.remote_attestation = true
> 
> sgx.ra_client_spid = "{{ env.get('RA_CLIENT_SPID', '') }}"
> ```

FYI: The new Python-based system is less flexible when it comes to pass arbitrary variables. Since in this case the variables come from the environment anyway, I didn't want to introduce a special mechanism just for handling them.

dimakuv (changes requested):
> Reviewed 37 of 37 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/182)* status: all files reviewed, 13 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_CQZG5IFRu5y9R6_D:-Mn_CQZG5IFRu5y9R6_E:b-6hy84q) (no related file):*
LTP tests use XML format, this PR introduces the TOML format. Do you plan to to move LTP from XML to TOML (in the follow-up PR)?

---
___
*[.ci/lib/stage-clean-check.jenkinsfile, line 9 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_7Uy95ra6oIoPO5py:-Mn_7Uy95ra6oIoPO5pz:b7a7k3o) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/.ci/lib/stage-clean-check.jenkinsfile#L9)):*
> ```
>         cd "$(./Scripts/clean-check-test-copy)"
> 
>         rm -rf "$REL_PREFIX"
> ```

I don't get it... Aren't you removing the same file? Why do you do the same deletion twice, first here and then using `sh 'rm -rf "$PREFIX"'`?
___
*[.ci/lib/stage-clean-check.jenkinsfile, line 50 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_7Ir6A2u1DKmoT4rH:-Mn_7Ir6A2u1DKmoT4rI:b-ew6an4) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/.ci/lib/stage-clean-check.jenkinsfile#L50)):*
> ```
>     sh 'rm -rf "$PREFIX"'
>     sh '''
>         ./Scripts/gitignore-test
> ```

Why did you move this check to the very bottom? Isn't it better to make sure that our git ignore rules cover all tests & CI-Examples in their built state, not after make clean?
___
*[.ci/lib/stage-test.jenkinsfile, line 6 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_7oze4iPOAnCRS3nI:-Mn_7oze4iPOAnCRS3nJ:b5fee2p) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/.ci/lib/stage-test.jenkinsfile#L6)):*
> ```
>             sh '''
>                 cd Pal/regression
>                 gramine-test build -v
> ```

How does it know whether you are in SGX mode or in direct mode? In other words, how does it know if it should generate SGX tokens?
___
*[Documentation/manpages/gramine-sgx-sign.rst, line 50 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_8BuI4cE95BdakD30:-Mn_8BuI4cE95BdakD31:brobiil) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/Documentation/manpages/gramine-sgx-sign.rst#L50)):*
> ```reStructuredText
>     SIGSTRUCT, i.e. files that should trigger rebuilding if they're modified.
>     The dependency file is in Makefile format, and is suitable for using in
>     build systems (Make, Ninja).
> ```

Don't we have this logic in some already-existing Meson file? Shouldn't we remove that logic from that Meson file now?

Or am I mistaken, and we used Makefiles for depfile creation?
___
*[LibOS/shim/test/fs/tests.toml, line 1 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_8jJC0JlxudqQpzYs:-Mn_8jJC0JlxudqQpzYt:b-iofwn3) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/LibOS/shim/test/fs/tests.toml#L1)):*
> ```TOML
> 
> ```

Please remove the empty line
___
*[LibOS/shim/test/fs/tests.toml, line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_8kc78ehKuPGk72Y1:-Mn_8kc78ehKuPGk72Y2:b-jgvnvt) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/LibOS/shim/test/fs/tests.toml#L2)):*
> ```TOML
> 
> binary_dir = '@GRAMINE_PKGLIBDIR@/tests/libos/fs'
> ```

In all other TOML files we use `"` for strings. Why is it different here?
___
*[LibOS/shim/test/regression/tests.toml, line 1 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_93-108Suokb9SqOc:-Mn_93-108Suokb9SqOd:b-pb4q6m) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/LibOS/shim/test/regression/tests.toml#L1)):*
> ```TOML
> 
> ```

Please remove empty line
___
*[LibOS/shim/test/regression/tests.toml, line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_94Mz086LeAQFPnmC:-Mn_94Mz086LeAQFPnmD:b-5963ux) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/LibOS/shim/test/regression/tests.toml#L2)):*
> ```TOML
> 
> binary_dir = '@GRAMINE_PKGLIBDIR@/tests/libos/regression'
> ```

Why `'` instead of `"`
___
*[Pal/regression/tests.toml, line 1 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_9HDG6Fof1IkVBMXn:-Mn_9HDG6Fof1IkVBMXo:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/Pal/regression/tests.toml#L1)):*
> ```TOML
> 
> ```

ditto
___
*[Pal/regression/tests.toml, line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_9Ie3DdS3DaoKwO-O:-Mn_9Ie3DdS3DaoKwO-P:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/Pal/regression/tests.toml#L2)):*
> ```TOML
> 
> binary_dir = '@GRAMINE_PKGLIBDIR@/tests/pal'
> ```

ditto
___
*[python/gramine-test, line 36 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_9pWq0sQ8plnu5LTy:-Mn_9pWq0sQ8plnu5LTz:bvi62zv) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/python/gramine-test#L36)):*
> ```
> 
> 
> @main.command(help='Rebuild manifests. This rebuild either all manifests, or the specified ones.')
> ```

`This rebuild` -> `This rebuilds`
___
*[python/graminelibos/test.py, line 17 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_BJxP8q8NfFZPMWQd:-Mn_BJxP8q8NfFZPMWQe:b727x13) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/python/graminelibos/test.py#L17)):*
> ```Python
> 
> 
> def get_list(data, *keys):
> ```

Could you add a quick comment that this function is to get lists of tests in TOML files, e.g. `[arch.x86_64] manifests`
___
*[python/graminelibos/test.py, line 65 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_C2mj4l_u6jFN32ym:-Mn_C2mj4l_u6jFN32yn:bfldw29) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/python/graminelibos/test.py#L65)):*
> ```Python
> 
>         toplevel = subprocess.check_output(['git', 'rev-parse', '--show-toplevel']).decode().strip()
>         self.key = os.path.join(toplevel, 'Pal/src/host/Linux-SGX/signer/enclave-key.pem')
> ```

This is smart!

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/182)* status: all files reviewed, 13 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_CQZG5IFRu5y9R6_D:-Mn_TLGgCi2TKLXkrrRt:b-bz0lxp) (no related file):*

LTP tests use XML format, this PR introduces the TOML format. Do you plan to to move LTP from XML to TOML (in the follow-up PR)?
</blockquote></details>

I don't know about XML (except for test results, which don't change). AFAIK, LTP tests use [configparser](https://docs.python.org/3/library/configparser.html), which is an `.ini`-like format similar to TOML, for skipping tests. I didn't have any plans to move this configuration, but yeah, it could be also converted to TOML for uniformity. 

(Mostly, I want to convert the LTP runner to Pytest, for several reasons: less custom code, standardized output, better control over concurrent runs).

---
___
*[.ci/lib/stage-clean-check.jenkinsfile, line 9 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_7Uy95ra6oIoPO5py:-Mn_UHe_4y-IOE4Hi1-o:bgcjw7p) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/.ci/lib/stage-clean-check.jenkinsfile#L9)):*

I don't get it... Aren't you removing the same file? Why do you do the same deletion twice, first here and then using `sh 'rm -rf "$PREFIX"'`?
</blockquote></details>

We just made a copy of the whole source tree (including `$PREFIX`, which is `usr/` inside that tree), and this removes the copied directory.
___
*[.ci/lib/stage-clean-check.jenkinsfile, line 50 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_7Ir6A2u1DKmoT4rH:-Mn_UcWu-l6TNmGuxONL:bwrg0zm) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/.ci/lib/stage-clean-check.jenkinsfile#L50)):*

Why did you move this check to the very bottom? Isn't it better to make sure that our git ignore rules cover all tests & CI-Examples in their built state, not after make clean?
</blockquote></details>

This test runs in the original directory. I added comments to make that more clear.
___
*[.ci/lib/stage-test.jenkinsfile, line 6 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_7oze4iPOAnCRS3nI:-Mn_XWB4CEWxxEWDXX3B:b-zc48yi) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/.ci/lib/stage-test.jenkinsfile#L6)):*

How does it know whether you are in SGX mode or in direct mode? In other words, how does it know if it should generate SGX tokens?
</blockquote></details>

This uses `SGX=1`, as before.
___
*[Documentation/manpages/gramine-sgx-sign.rst, line 50 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_8BuI4cE95BdakD30:-Mn_X_Gd4u5NJHKZ_RTH:b-kad846) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/Documentation/manpages/gramine-sgx-sign.rst#L50)):*

Don't we have this logic in some already-existing Meson file? Shouldn't we remove that logic from that Meson file now?

Or am I mistaken, and we used Makefiles for depfile creation?
</blockquote></details>

We never used Meson for manifest creation. We have this logic in Makefiles, it uses `gramine-gen-depend`. I cannot remove that yet because of LTP.

(I could switch Makefiles to use `--depfile` but I'd rather spend the time on removing them :) )
___
*[LibOS/shim/test/fs/tests.toml, line 1 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_8jJC0JlxudqQpzYs:-Mn_YPBpDA_O95qlSmof:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/LibOS/shim/test/fs/tests.toml#L1)):*

Please remove the empty line
</blockquote></details>

Done.
___
*[LibOS/shim/test/fs/tests.toml, line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_8kc78ehKuPGk72Y1:-Mn_YTn73thZdcGhAp5N:b-i8m3ct) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/LibOS/shim/test/fs/tests.toml#L2)):*

In all other TOML files we use `"` for strings. Why is it different here?
</blockquote></details>

I haven't thought of it . I'm switching to double quotes then.
___
*[LibOS/shim/test/regression/tests.toml, line 1 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_93-108Suokb9SqOc:-Mn_YsCk94wq90v7z3k9:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/LibOS/shim/test/regression/tests.toml#L1)):*

Please remove empty line
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/tests.toml, line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_94Mz086LeAQFPnmC:-Mn_YupB7cSmnKCfXsrh:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/LibOS/shim/test/regression/tests.toml#L2)):*

Why `'` instead of `"`
</blockquote></details>

Done.
___
*[Pal/regression/tests.toml, line 1 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_9HDG6Fof1IkVBMXn:-Mn_YuEc6FLe4oF-_RU3:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/Pal/regression/tests.toml#L1)):*

ditto
</blockquote></details>

Done.
___
*[Pal/regression/tests.toml, line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_9Ie3DdS3DaoKwO-O:-Mn_YvE33HDuKCNYk9vK:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/Pal/regression/tests.toml#L2)):*

ditto
</blockquote></details>

Done.
___
*[python/gramine-test, line 36 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_9pWq0sQ8plnu5LTy:-Mn_YyR69a8NAYOWmoVS:bt2vr1b) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/python/gramine-test#L36)):*

`This rebuild` -> `This rebuilds`
</blockquote></details>

Fixed.
___
*[python/graminelibos/test.py, line 17 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_BJxP8q8NfFZPMWQd:-Mn__WwWBpiYVNbXah3w:b-mpjeg) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/python/graminelibos/test.py#L17)):*

Could you add a quick comment that this function is to get lists of tests in TOML files, e.g. `[arch.x86_64] manifests`
</blockquote></details>

OK. I made it a bit more generic too.
___
*[python/graminelibos/test.py, line 65 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_C2mj4l_u6jFN32ym:-Mn__iYG57H1Pz9rfg_M:b-6oyube) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/python/graminelibos/test.py#L65)):*

This is smart!
</blockquote></details>

Thanks, but I'd say it's hacky :) I think we want to keep the key in `~/.config` someday, so this won't be necessary.

woju (changes requested):
> Reviewed 7 of 37 files at r1, 1 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/182)* status: 32 of 37 files reviewed, 17 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @pwmarcz)

___
*[LICENSE.addendum.txt, line 12 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-MnWKEap53V7O7tcnpR2:-Mna4E0CADHrPZRF_G8n:b-npjadl) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/LICENSE.addendum.txt#L12)):*

FYI: This file is a part of Ninja source distribution (https://github.com/ninja-build/ninja/blob/master/misc/ninja_syntax.py), but AFAICT, it's not packaged by Linux distributions. There is an [unofficial package on PyPI](https://pypi.org/project/ninja_syntax/) but it looks abandoned, and [another package that provides the whole Ninja](https://pypi.org/project/ninja/), not only that file.

Overall, I decided it's better to vendor this file, but maybe you'll have a better suggestion.
</blockquote></details>

IDK. Could you, in any case, file a bug against Debian's package to include this file?
___
*[.ci/lib/stage-clean-check.jenkinsfile, line 9 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_7Uy95ra6oIoPO5py:-MnagQxu8Dbzjj6JCZRY:bjrj6jo) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/.ci/lib/stage-clean-check.jenkinsfile#L9)):*
> Aren't you removing the same file?

Note `cd` between `--relative-to=.` and `rm`.
___
*[python/gramine-test, line 39 at r2](https://reviewable.io/reviews/gramineproject/gramine/182#-Mnah9Ng7ulmfqSD6Wp9:-Mnah9Ng7ulmfqSD6WpA:by46r91) ([raw file](https://github.com/gramineproject/gramine/blob/d35298d3b09f61e27d9774b8af70c6f24931c423/python/gramine-test#L39)):*
> ```
> @click.pass_context
> @click.option('--force', '-f', is_flag=True, help='Force rebuild')
> @click.option('--verbose', '-v', is_flag=True, help='Show all command lines while building')
> ```

Why did you use `is_flag=` instead of `'--verbose/--no-verbose'`?
___
*[python/graminelibos/meson.build, line 5 at r2](https://reviewable.io/reviews/gramineproject/gramine/182#-MnajjRI-8S650nSrZxO:-MnajjRI-8S650nSrZxP:b-rwnowz) ([raw file](https://github.com/gramineproject/gramine/blob/d35298d3b09f61e27d9774b8af70c6f24931c423/python/graminelibos/meson.build#L5)):*
> ```Meson
> conf.set('PKGLIBDIR', join_paths(get_option('prefix'), pkglibdir))
> conf.set('LIBDIR', join_paths(get_option('prefix'), get_option('libdir')))
> conf.set('ARCH_LIBDIR', get_option('libdir'))
> ```

Why this is needed? Isn't `LIBDIR` enough?
___
*[python/graminelibos/test.py, line 65 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_C2mj4l_u6jFN32ym:-Mnaffa93DNqgo2aCE8F:bdxozeu) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/python/graminelibos/test.py#L65)):*

Thanks, but I'd say it's hacky :) I think we want to keep the key in `~/.config` someday, so this won't be necessary.
</blockquote></details>

Please no. I already keep my keys in `~/.config`, and have `SGX_SIGNER_KEY` in shell. At least make this `os.getenv('SGX_SIGNER_KEY', os.path.join(...))`.
___
*[python/graminelibos/test.py, line 70 at r2](https://reviewable.io/reviews/gramineproject/gramine/182#-MnakIuwC3qS6eB7txwn:-MnakIuwC3qS6eB7txwo:b-fxivgh) ([raw file](https://github.com/gramineproject/gramine/blob/d35298d3b09f61e27d9774b8af70c6f24931c423/python/graminelibos/test.py#L70)):*
> ```Python
>             self.binary_dir = '.'
> 
>         self.arch_libdir = '/' + _CONFIG_ARCH_LIBDIR
> ```

Shouldn't this be just _CONFIG_LIBDIR, and isn't `/` hardcoded `get_option('prefix')`?

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/182)* status: all files reviewed, 4 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_CQZG5IFRu5y9R6_D:-Mne2zj5Bbj3Tlrit5Lx:b9zjh2o) (no related file):*

I don't know about XML (except for test results, which don't change). AFAIK, LTP tests use [configparser](https://docs.python.org/3/library/configparser.html), which is an `.ini`-like format similar to TOML, for skipping tests. I didn't have any plans to move this configuration, but yeah, it could be also converted to TOML for uniformity. 

(Mostly, I want to convert the LTP runner to Pytest, for several reasons: less custom code, standardized output, better control over concurrent runs).
</blockquote></details>

OK, I'm fine with keeping the XML format for that `ltp.cfg` file. Surely not high-priority.

---

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/182)* status: all files reviewed, 4 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @woju)

___
*[LICENSE.addendum.txt, line 12 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-MnWKEap53V7O7tcnpR2:-MneqKtuB4FmJAbXjlyS:b-kqme0e) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/LICENSE.addendum.txt#L12)):*

IDK. Could you, in any case, file a bug against Debian's package to include this file?
</blockquote></details>

There's [a bug open 2 years ago](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=943540), I sent a short "I want this too" message.
___
*[python/gramine-test, line 39 at r2](https://reviewable.io/reviews/gramineproject/gramine/182#-Mnah9Ng7ulmfqSD6Wp9:-Mneqt_G1_toGiC6C5H6:b9rp28o) ([raw file](https://github.com/gramineproject/gramine/blob/d35298d3b09f61e27d9774b8af70c6f24931c423/python/gramine-test#L39)):*

Why did you use `is_flag=` instead of `'--verbose/--no-verbose'`?
</blockquote></details>

Is that the recommended practice? Does it apply to all boolean flags I just added, i.e. should we have `--no-force` and `--no-quiet` as well?
___
*[python/graminelibos/meson.build, line 5 at r2](https://reviewable.io/reviews/gramineproject/gramine/182#-MnajjRI-8S650nSrZxO:-Mneqzg16jOH9lKYGbfB:bjrvu83) ([raw file](https://github.com/gramineproject/gramine/blob/d35298d3b09f61e27d9774b8af70c6f24931c423/python/graminelibos/meson.build#L5)):*

Why this is needed? Isn't `LIBDIR` enough?
</blockquote></details>

`LIBDIR`, as defined here, is the (possibly custom) directory where Gramine is installed. We want the directory with system libraries (in the regression tests: `libgcc_s`, `libstdc++`, `liblzma`, `libgomp/libomp`, but a similar argument applies to `CI-Examples`); I don't think we should require all of them to be also installed in that custom directory.

(This is `ARCH_LIBDIR` in Makefiles and `{{ arch_libdir }}` in the templates.)
___
*[python/graminelibos/test.py, line 65 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_C2mj4l_u6jFN32ym:-MnetP9TEQPkz0d7XzaX:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/python/graminelibos/test.py#L65)):*

Please no. I already keep my keys in `~/.config`, and have `SGX_SIGNER_KEY` in shell. At least make this `os.getenv('SGX_SIGNER_KEY', os.path.join(...))`.
</blockquote></details>

Done.
___
*[python/graminelibos/test.py, line 70 at r2](https://reviewable.io/reviews/gramineproject/gramine/182#-MnakIuwC3qS6eB7txwn:-MnesuoF8lNDOyKAccoG:b-q64pkt) ([raw file](https://github.com/gramineproject/gramine/blob/d35298d3b09f61e27d9774b8af70c6f24931c423/python/graminelibos/test.py#L70)):*

Shouldn't this be just _CONFIG_LIBDIR, and isn't `/` hardcoded `get_option('prefix')`?
</blockquote></details>

See above.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/182)* status: 36 of 37 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @woju)

___
*[python/graminelibos/meson.build, line 5 at r2](https://reviewable.io/reviews/gramineproject/gramine/182#-MnajjRI-8S650nSrZxO:-MnewSK07xbBPQy1xLEq:bb6is2z) ([raw file](https://github.com/gramineproject/gramine/blob/d35298d3b09f61e27d9774b8af70c6f24931c423/python/graminelibos/meson.build#L5)):*

`LIBDIR`, as defined here, is the (possibly custom) directory where Gramine is installed. We want the directory with system libraries (in the regression tests: `libgcc_s`, `libstdc++`, `liblzma`, `libgomp/libomp`, but a similar argument applies to `CI-Examples`); I don't think we should require all of them to be also installed in that custom directory.

(This is `ARCH_LIBDIR` in Makefiles and `{{ arch_libdir }}` in the templates.)
</blockquote></details>

BTW, using Meson for this is a massive upgrade: before, we had the awkward `try "/lib/$(gcc -dumpmachine)", and "/lib64" if that doesn't exist, and remember that "clang -dumpmachine" has a different format`.

woju (changes requested):
> Reviewed 5 of 37 files at r1.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/182)* status: 36 of 37 files reviewed, 9 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @pwmarcz)

___
*[LICENSE.addendum.txt, line 12 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-MnWKEap53V7O7tcnpR2:-MnfA8KL1FEDUveC1Hb9:b-ucfkn4) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/LICENSE.addendum.txt#L12)):*

There's [a bug open 2 years ago](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=943540), I sent a short "I want this too" message.
</blockquote></details>

Thanks.
___
*[Documentation/manpages/gramine-sgx-sign.rst, line 45 at r2](https://reviewable.io/reviews/gramineproject/gramine/182#-MneF78_BVRwtLJjrQDB:-MneF78_BVRwtLJjrQDC:b-ygqlo9) ([raw file](https://github.com/gramineproject/gramine/blob/d35298d3b09f61e27d9774b8af70c6f24931c423/Documentation/manpages/gramine-sgx-sign.rst#L45)):*
> ```reStructuredText
>     the end and with ".sig" appended.
> 
> .. option:: --depfile depfile, -D depfile
> ```

`-D` is more like `--define`, can you remove the short option?
___
*[python/gramine-sgx-sign, line 22 at r2](https://reviewable.io/reviews/gramineproject/gramine/182#-MneGENZ8VxUlkdTfs5y:-MneGENZ8VxUlkdTfs5z:b-ybhxb3) ([raw file](https://github.com/gramineproject/gramine/blob/d35298d3b09f61e27d9774b8af70c6f24931c423/python/gramine-sgx-sign#L22)):*
> ```
>               required=True, help='Input .manifest file')
> @click.option('--sigfile', '-s', help='Output .sig file')
> @click.option('--depfile', '-D', type=click.Path(), help='Generate dependencies for .manifest.sgx '
> ```

`type=click.File('w')` and you won't have to `open()` it yourself.
___
*[python/gramine-sgx-sign, line 54 at r2](https://reviewable.io/reviews/gramineproject/gramine/182#-MneGRVc0QXoTLSzYcsQ:-MneGRVc0QXoTLSzYcsR:b-s226c4) ([raw file](https://github.com/gramineproject/gramine/blob/d35298d3b09f61e27d9774b8af70c6f24931c423/python/gramine-sgx-sign#L54)):*
> ```
>             f.write(f'{output}:')
>             for filename in expanded:
>                 f.write(f'  \\\n\t{filename}')
> ```

intentionally two spaces? (also below)
___
*[python/gramine-test, line 39 at r2](https://reviewable.io/reviews/gramineproject/gramine/182#-Mnah9Ng7ulmfqSD6Wp9:-Mnf4uIUD65-7sP45-Rj:b-99uj3m) ([raw file](https://github.com/gramineproject/gramine/blob/d35298d3b09f61e27d9774b8af70c6f24931c423/python/gramine-test#L39)):*

Is that the recommended practice? Does it apply to all boolean flags I just added, i.e. should we have `--no-force` and `--no-quiet` as well?
</blockquote></details>

https://click.palletsprojects.com/en/8.0.x/options/#boolean-flags

It is generally expected that you should be able to undo a boolean flag (and re-do it after, if you like), because you might assemble the command line from multiple variables (see CFLAGS). I don't think you necessarily need `--no-` prefix, you might want to do `--verbose/--quiet`, not `.../--no-verbose`.

Yes, it applies to all of them, unless of course you have a very good reason not to do such thing, like security ACLs which should stay shut etc.
___
*[python/graminelibos/meson.build, line 5 at r2](https://reviewable.io/reviews/gramineproject/gramine/182#-MnajjRI-8S650nSrZxO:-Mnf6QFO58ADBgadDSn1:b-jlk4hg) ([raw file](https://github.com/gramineproject/gramine/blob/d35298d3b09f61e27d9774b8af70c6f24931c423/python/graminelibos/meson.build#L5)):*

BTW, using Meson for this is a massive upgrade: before, we had the awkward `try "/lib/$(gcc -dumpmachine)", and "/lib64" if that doesn't exist, and remember that "clang -dumpmachine" has a different format`.
</blockquote></details>

Ah, I understand. In that case, this should be a separate meson option, `-Dsyslibdir`, with default `'/' + get_option('libdir')`, and the `'/'+` should be dropped from python. This is because we might elect to use `libdir` contrary to distro's standard, in fact #16 specifies exactly this.

`sys<something>dir` vs `<something>dir` is a convention for directories outside prefix, for "system's" things as opposed to ours which are in prefix. Unfortunately meson has only `sysconfdir` built-in, so we'd have to define it ourselves. It's rarely used, but sometimes needed. Cf. https://www.musl-libc.org/doc/1.0.0/manual.html (ctrl+f, syslibdir).
___
*[python/graminelibos/sgx_sign.py, line 526 at r2](https://reviewable.io/reviews/gramineproject/gramine/182#-MnbN86n75nb7Po3hss5:-MnbN86oAMMXi86zV7Ix:bii26p) ([raw file](https://github.com/gramineproject/gramine/blob/d35298d3b09f61e27d9774b8af70c6f24931c423/python/graminelibos/sgx_sign.py#L526)):*
> ```Python
>         date (date): Date to put into SIGSTRUCT.
>         libpal (:obj:`str`, optional): Path to the libpal file.
>         libpal (:obj:`bool`, optional): If true, print details to stdout.
> ```

`s/libpal/verbose/`
___
*[python/graminelibos/test.py, line 17 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_BJxP8q8NfFZPMWQd:-MnenR0K8opCDD6EInlb:b7h83dj) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/python/graminelibos/test.py#L17)):*

OK. I made it a bit more generic too.
</blockquote></details>

I feel like this function is unneeded, instead a set of precision `setdefault()`s should do the job:

```py
data = toml.load(...)
data.setdefault('manifests', [])
data.setdefault('arch', {}).setdefault(platform.processor(), {}).setdefault('manifests', [])
data.setdefault('sgx', {}).setdefault('manifests', [])
data.setdefault('binary_dir', '.')
data['binary_dir'] = data['binary_dir'].replace('@...@', ...)
```
?
___
*[python/graminelibos/test.py, line 65 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_C2mj4l_u6jFN32ym:-Mnf9meKB5KY_TBNokzu:b-qt4aaq) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/python/graminelibos/test.py#L65)):*

Done.
</blockquote></details>

`toplevel = subprocess.check_output()`?
___
*[python/graminelibos/test.py, line 1 at r2](https://reviewable.io/reviews/gramineproject/gramine/182#-Mnek07zCjwhQqrwffZo:-Mnek07zCjwhQqrwffZp:b-t4ezwz) ([raw file](https://github.com/gramineproject/gramine/blob/d35298d3b09f61e27d9774b8af70c6f24931c423/python/graminelibos/test.py#L1)):*
> ```Python
> #!/usr/bin/env python3
> ```

I don't think this file should be named `test.py`, everything starting with `test` (and especially `test_` or `tests_`) is a liability for unittest-style autodetection. Can we name this "utils" or "tools" or something like that?
___
*[python/graminelibos/test.py, line 70 at r2](https://reviewable.io/reviews/gramineproject/gramine/182#-MnakIuwC3qS6eB7txwn:-Mnf9TxjA6xN0EA09aqR:b95vfwi) ([raw file](https://github.com/gramineproject/gramine/blob/d35298d3b09f61e27d9774b8af70c6f24931c423/python/graminelibos/test.py#L70)):*

See above.
</blockquote></details>

Resolving here.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/182)* status: 28 of 39 files reviewed, 9 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @woju)

___
*[Documentation/manpages/gramine-sgx-sign.rst, line 45 at r2](https://reviewable.io/reviews/gramineproject/gramine/182#-MneF78_BVRwtLJjrQDB:-MnfJGRu5VzEQJ7Qrh-j:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/d35298d3b09f61e27d9774b8af70c6f24931c423/Documentation/manpages/gramine-sgx-sign.rst#L45)):*

`-D` is more like `--define`, can you remove the short option?
</blockquote></details>

Done.
___
*[python/gramine-sgx-sign, line 22 at r2](https://reviewable.io/reviews/gramineproject/gramine/182#-MneGENZ8VxUlkdTfs5y:-MnfJeeU2ePkMZNK7KCU:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/d35298d3b09f61e27d9774b8af70c6f24931c423/python/gramine-sgx-sign#L22)):*

`type=click.File('w')` and you won't have to `open()` it yourself.
</blockquote></details>

Done.
___
*[python/gramine-sgx-sign, line 54 at r2](https://reviewable.io/reviews/gramineproject/gramine/182#-MneGRVc0QXoTLSzYcsQ:-MnfKLQZ5zPRDvdtvazv:bovsjp9) ([raw file](https://github.com/gramineproject/gramine/blob/d35298d3b09f61e27d9774b8af70c6f24931c423/python/gramine-sgx-sign#L54)):*

intentionally two spaces? (also below)
</blockquote></details>

No. I see `gramine-gen-depend` used one space, and that makes sense. Fixed.
___
*[python/gramine-test, line 39 at r2](https://reviewable.io/reviews/gramineproject/gramine/182#-Mnah9Ng7ulmfqSD6Wp9:-MnfMQW02TAI9p3TQX7m:b9ce24d) ([raw file](https://github.com/gramineproject/gramine/blob/d35298d3b09f61e27d9774b8af70c6f24931c423/python/gramine-test#L39)):*

https://click.palletsprojects.com/en/8.0.x/options/#boolean-flags

It is generally expected that you should be able to undo a boolean flag (and re-do it after, if you like), because you might assemble the command line from multiple variables (see CFLAGS). I don't think you necessarily need `--no-` prefix, you might want to do `--verbose/--quiet`, not `.../--no-verbose`.

Yes, it applies to all of them, unless of course you have a very good reason not to do such thing, like security ACLs which should stay shut etc.
</blockquote></details>

OK. I changed the `--quiet` option to `--verbose/--quiet`, and added `--no-*` variants for the rest.
___
*[python/graminelibos/meson.build, line 5 at r2](https://reviewable.io/reviews/gramineproject/gramine/182#-MnajjRI-8S650nSrZxO:-MnfNb5DA4wqdVpNzdeL:b-k1y5pj) ([raw file](https://github.com/gramineproject/gramine/blob/d35298d3b09f61e27d9774b8af70c6f24931c423/python/graminelibos/meson.build#L5)):*

Ah, I understand. In that case, this should be a separate meson option, `-Dsyslibdir`, with default `'/' + get_option('libdir')`, and the `'/'+` should be dropped from python. This is because we might elect to use `libdir` contrary to distro's standard, in fact #16 specifies exactly this.

`sys<something>dir` vs `<something>dir` is a convention for directories outside prefix, for "system's" things as opposed to ours which are in prefix. Unfortunately meson has only `sysconfdir` built-in, so we'd have to define it ourselves. It's rarely used, but sometimes needed. Cf. https://www.musl-libc.org/doc/1.0.0/manual.html (ctrl+f, syslibdir).
</blockquote></details>

OK. Done.
___
*[python/graminelibos/sgx_sign.py, line 526 at r2](https://reviewable.io/reviews/gramineproject/gramine/182#-MnbN86n75nb7Po3hss5:-MnfNpcgAEzD5mMWMF8t:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/d35298d3b09f61e27d9774b8af70c6f24931c423/python/graminelibos/sgx_sign.py#L526)):*

`s/libpal/verbose/`
</blockquote></details>

Done.
___
*[python/graminelibos/test.py, line 17 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_BJxP8q8NfFZPMWQd:-MnfPSyh3_j-d711CaoJ:b-z3q77y) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/python/graminelibos/test.py#L17)):*

I feel like this function is unneeded, instead a set of precision `setdefault()`s should do the job:

```py
data = toml.load(...)
data.setdefault('manifests', [])
data.setdefault('arch', {}).setdefault(platform.processor(), {}).setdefault('manifests', [])
data.setdefault('sgx', {}).setdefault('manifests', [])
data.setdefault('binary_dir', '.')
data['binary_dir'] = data['binary_dir'].replace('@...@', ...)
```
?
</blockquote></details>

I agree this function can be avoided, I disagree with the specific suggestion of mutating the parsed data in place (I prefer to keep it data unchanged and do the conversion/defaults/anything non-trivial while assigning to variables; something in the spirit of "parse, don't validate").

I refactored the code in a similar way to your suggestion, but with `get()` instead of `setdefault()`.
___
*[python/graminelibos/test.py, line 65 at r1](https://reviewable.io/reviews/gramineproject/gramine/182#-Mn_C2mj4l_u6jFN32ym:-MnfRaHx5Yid3BHJRe-l:bidrs83) ([raw file](https://github.com/gramineproject/gramine/blob/c4c605bef8a4297d62cbea0635148000ceba8b24/python/graminelibos/test.py#L65)):*

`toplevel = subprocess.check_output()`?
</blockquote></details>

Right. Fixed.
___
*[python/graminelibos/test.py, line 1 at r2](https://reviewable.io/reviews/gramineproject/gramine/182#-Mnek07zCjwhQqrwffZo:-MnfO2B4BSB6ZiUZVM50:b-c7pnsf) ([raw file](https://github.com/gramineproject/gramine/blob/d35298d3b09f61e27d9774b8af70c6f24931c423/python/graminelibos/test.py#L1)):*

I don't think this file should be named `test.py`, everything starting with `test` (and especially `test_` or `tests_`) is a liability for unittest-style autodetection. Can we name this "utils" or "tools" or something like that?
</blockquote></details>

I agree `test.py` is a bad name, but I think "utils" or "tools" are even worse: they suggest that this is a "miscellaneous" module for random utilities (and I want it to be specifically for supporting Gramine tests and implementing the `gramine-test` utility).

So if I was renaming it, I would rename it to something like `gramine_test.py` or `test_config.py` or `test_builder.py`. If you think we definitely need to avoid "test" in the name, maybe you'll have a better idea?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/182)* status: 28 of 39 files reviewed, 9 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @woju)

___
*[python/graminelibos/test.py, line 1 at r2](https://reviewable.io/reviews/gramineproject/gramine/182#-Mnek07zCjwhQqrwffZo:-Mo-Mrx72eFhliQ046lv:b-76yen2) ([raw file](https://github.com/gramineproject/gramine/blob/d35298d3b09f61e27d9774b8af70c6f24931c423/python/graminelibos/test.py#L1)):*

I agree `test.py` is a bad name, but I think "utils" or "tools" are even worse: they suggest that this is a "miscellaneous" module for random utilities (and I want it to be specifically for supporting Gramine tests and implementing the `gramine-test` utility).

So if I was renaming it, I would rename it to something like `gramine_test.py` or `test_config.py` or `test_builder.py`. If you think we definitely need to avoid "test" in the name, maybe you'll have a better idea?
</blockquote></details>

Renamed to `util_tests.py` as suggested by @woju.

woju (changes requested):
> Reviewed 2 of 37 files at r1, 9 of 11 files at r5, 3 of 3 files at r7.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/182)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[python/gramine-test, line 22 at r7](https://reviewable.io/reviews/gramineproject/gramine/182#-MoIj5bG6rWfuL9kOtk2:-MoIj5bG6rWfuL9kOtk3:bl3eirr) ([raw file](https://github.com/gramineproject/gramine/blob/77b5e4ad8a9edfa12b52336fda6f020d8febfdb2/python/gramine-test#L22)):*
> ```
> @click.pass_context
> def main(ctx, sgx, directory):
>     if os.environ.get('SGX') == '1':
> ```

This logic is wrong: command line switch should override the environment and not the other way around.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/182)* status: 38 of 39 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @woju)

___
*[python/gramine-test, line 22 at r7](https://reviewable.io/reviews/gramineproject/gramine/182#-MoIj5bG6rWfuL9kOtk2:-MoIm32TEKYDLiJs1XD3:b-deblf4) ([raw file](https://github.com/gramineproject/gramine/blob/77b5e4ad8a9edfa12b52336fda6f020d8febfdb2/python/gramine-test#L22)):*

This logic is wrong: command line switch should override the environment and not the other way around.
</blockquote></details>

Good point.

I was hoping to do `click.option('--sgx/--no-sgx', default=None)` and then check if `sgx is None`, but that doesn't work (Click still gives me `False`). So I'm doing `default=(os.environ.get('SGX') == '1')` instead.

woju (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/182)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

___
*[python/gramine-test, line 22 at r7](https://reviewable.io/reviews/gramineproject/gramine/182#-MoIj5bG6rWfuL9kOtk2:-MoKU9pQ2FOYifSOfn33:b-4ywtmy) ([raw file](https://github.com/gramineproject/gramine/blob/77b5e4ad8a9edfa12b52336fda6f020d8febfdb2/python/gramine-test#L22)):*

Good point.

I was hoping to do `click.option('--sgx/--no-sgx', default=None)` and then check if `sgx is None`, but that doesn't work (Click still gives me `False`). So I'm doing `default=(os.environ.get('SGX') == '1')` instead.
</blockquote></details>

This looks perfect. I skimmed documentation of the built-in envvar mechanism, https://click.palletsprojects.com/en/8.0.x/options/#values-from-environment-variables, but didn't find anything that would fit booleans.

dimakuv (approved):
> Reviewed 6 of 11 files at r5, 2 of 3 files at r7, 1 of 1 files at r8, 6 of 8 files at r9, 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/182)* status: 38 of 39 files reviewed, all discussions resolved (waiting on @woju)

woju (approved):
> Reviewed 1 of 8 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/182)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #192, opened 2021-11-02
[Pal] Reduce the 64MB static buffer

pwmarcz:
## Description of the changes 

This change reduces the initial 64MB buffer to a much smaller size, and allocates 64MB of memory after PAL startup instead (by making that the default value for `loader.pal_internal_mem_size`).

While the memory still has to be explicitely zeroed, this makes calculating the enclave measurement much faster (`gramine-sgx-sign` on test binaries takes 50% less time, because we can omit hashing 64MB worth of zeroes.

## How to test this PR? 

CI should be enough, but keep in mind this is a breaking change if existing manifests define `loader.pal_internal_mem_size`.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/192)

### DIFF
--- a/Pal/include/pal_internal.h
+++ b/Pal/include/pal_internal.h
@@ -244,7 +244,7 @@ int _DkSetProtectedFilesKey(PAL_PTR pf_key_hex);
         _DkProcessExit(exitcode);                                      \
     } while (0)
 
-void init_slab_mgr(void);
+void init_slab_mgr(char* mem_pool, size_t mem_pool_size);
 void* malloc(size_t size);
 void* malloc_copy(const void* mem, size_t size);
 void* calloc(size_t nmem, size_t size);
@@ -277,5 +277,9 @@ const char* pal_event_name(enum PAL_EVENT event);
         _DkProcessExit(PAL_ERROR_NOMEM);       \
     } while (0)
 #include "uthash.h"
+
+/* Size of PAL memory available before parsing the manifest; `loader.pal_internal_mem_size` does not
+ * include this memory */
+#define PAL_INITIAL_MEM_SIZE (64 * 1024 * 1024)
 
 #endif
--- a/Pal/regression/Memory.c
+++ b/Pal/regression/Memory.c
@@ -69,8 +69,10 @@ int main(int argc, char** argv, char** envp) {
             pal_printf("Memory Deallocation OK\n");
     }
 
+    /* TODO: This does not take into account `pal_control.preloaded_ranges`; we are not allowed to
+     * ask for memory overlapping with these ranges */
     void* mem3 = (void*)pal_control.user_address.start;
-    void* mem4 = (void*)pal_control.user_address.end - UNIT;
+    void* mem4 = (void*)pal_control.user_address.start + UNIT;
 
     int ret2 = DkVirtualMemoryAlloc(&mem3, UNIT, 0, PAL_PROT_READ | PAL_PROT_WRITE);
     ret = DkVirtualMemoryAlloc(&mem4, UNIT, 0, PAL_PROT_READ | PAL_PROT_WRITE);
--- a/Pal/src/host/Linux-SGX/db_main.c
+++ b/Pal/src/host/Linux-SGX/db_main.c
@@ -41,10 +41,8 @@ static struct link_map g_pal_map;
 
 PAL_SESSION_KEY g_master_key = {0};
 
-/* for internal PAL objects, Gramine first uses pre-allocated g_mem_pool and then falls back to
- * _DkVirtualMemoryAlloc(PAL_ALLOC_INTERNAL); the amount of available PAL internal memory is
- * limited by the variable below */
-size_t g_pal_internal_mem_size = 0;
+/* Limit of PAL memory available for _DkVirtualMemoryAlloc(PAL_ALLOC_INTERNAL) */
+size_t g_pal_internal_mem_size = PAL_INITIAL_MEM_SIZE;
 
 const size_t g_page_size = PRESET_PAGESIZE;
 
@@ -646,6 +644,12 @@ out:
     return ret;
 }
 
+__attribute_no_sanitize_address
+static void do_preheat_enclave(void) {
+    for (uint8_t* i = g_pal_sec.heap_min; i < (uint8_t*)g_pal_sec.heap_max; i += g_page_size)
+        READ_ONCE(*(size_t*)i);
+}
+
 /* Gramine uses GCC's stack protector that looks for a canary at gs:[0x8], but this function starts
  * with a default canary and then updates it to a random one, so we disable stack protector here */
 __attribute_no_stack_protector
@@ -721,10 +725,11 @@ noreturn void pal_linux_main(char* uptr_libpal_uri, size_t libpal_uri_len, char*
     g_pal_sec.uid = sec_info.uid;
     g_pal_sec.gid = sec_info.gid;
 
-    /* set up page allocator and slab manager */
-    init_slab_mgr();
-    init_untrusted_slab_mgr();
+    /* Set up page allocator and slab manager. There is no need to provide any initial memory pool,
+     * because the slab manager can use normal allocations (`_DkVirtualMemoryAlloc`) right away. */
     init_enclave_pages();
+    init_slab_mgr(/*mem_pool=*/NULL, /*mem_pool_size=*/0);
+    init_untrusted_slab_mgr();
 
     /* initialize enclave properties */
     ret = init_enclave();
@@ -840,25 +845,26 @@ noreturn void pal_linux_main(char* uptr_libpal_uri, size_t libpal_uri_len, char*
         log_error("Cannot parse 'sgx.preheat_enclave' (the value must be `true` or `false`)");
         ocall_exit(1, /*is_exitgroup=*/true);
     }
-    if (preheat_enclave) {
-        for (uint8_t* i = g_pal_sec.heap_min; i < (uint8_t*)g_pal_sec.heap_max; i += g_page_size)
-            READ_ONCE(*(size_t*)i);
-    }
+    if (preheat_enclave)
+        do_preheat_enclave();
 
-    size_t pal_internal_mem_size;
+    /* For backward compatibility, `loader.pal_internal_mem_size` does not include
+     * PAL_INITIAL_MEM_SIZE */
+    size_t extra_mem_size;
     ret = toml_sizestring_in(g_pal_state.manifest_root, "loader.pal_internal_mem_size",
-                             /*defaultval=*/0, &pal_internal_mem_size);
+                             /*defaultval=*/0, &extra_mem_size);
     if (ret < 0) {
         log_error("Cannot parse 'loader.pal_internal_mem_size'");
         ocall_exit(1, /*is_exitgroup=*/true);
     }
 
-    if (pal_internal_mem_size < g_pal_internal_mem_size) {
+    if (extra_mem_size + PAL_INITIAL_MEM_SIZE < g_pal_internal_mem_size) {
         log_error("Too small `loader.pal_internal_mem_size`, need at least %luMB because the "
-                  "manifest is large", g_pal_internal_mem_size / 1024 / 1024);
+                  "manifest is large",
+                  (g_pal_internal_mem_size - PAL_INITIAL_MEM_SIZE) / 1024 / 1024);
         ocall_exit(1, /*is_exitgroup=*/true);
     }
-    g_pal_internal_mem_size = pal_internal_mem_size;
+    g_pal_internal_mem_size = extra_mem_size + PAL_INITIAL_MEM_SIZE;
 
     if ((ret = init_file_check_policy()) < 0) {
         log_error("Failed to load the file check policy: %d", ret);
--- a/Pal/src/host/Linux/db_main.c
+++ b/Pal/src/host/Linux/db_main.c
@@ -241,7 +241,24 @@ noreturn void pal_linux_main(void* initial_rsp, void* fini_callback) {
 
     g_linux_state.host_environ = envp;
 
-    init_slab_mgr();
+    /* Prepare an initial memory pool for the slab allocator. This is necessary because we cannot
+     * allocate the PAL-internal range yet: we need to parse the manifest to know its size. */
+    size_t init_pool_size = PAL_INITIAL_MEM_SIZE;
+    void* init_pool_addr = (void*)DO_SYSCALL(mmap, /*addr=*/NULL, init_pool_size,
+                                             PROT_READ | PROT_WRITE,
+                                             MAP_ANONYMOUS | MAP_PRIVATE, /*fd=*/-1, /*offset=*/0);
+    if (IS_PTR_ERR(init_pool_addr)) {
+        INIT_FAIL(PAL_ERROR_NOMEM, "Cannot allocate initial memory pool");
+    }
+
+    init_slab_mgr(init_pool_addr, init_pool_size);
+
+    ret = add_preloaded_range((uintptr_t)init_pool_addr,
+                              (uintptr_t)init_pool_addr + init_pool_size,
+                              "pal_init_pool");
+    if (ret < 0) {
+        INIT_FAIL(PAL_ERROR_NOMEM, "Out of memory");
+    }
 
 #ifdef DEBUG
     ret = debug_map_init_from_proc_maps();
--- a/Pal/src/slab.c
+++ b/Pal/src/slab.c
@@ -18,14 +18,14 @@ static spinlock_t g_slab_mgr_lock = INIT_SPINLOCK_UNLOCKED;
 #define SYSTEM_UNLOCK() spinlock_unlock(&g_slab_mgr_lock)
 #define SYSTEM_LOCKED() spinlock_is_locked(&g_slab_mgr_lock)
 
-#define POOL_SIZE 64 * 1024 * 1024
-static char g_mem_pool[POOL_SIZE];
+/* Initial memory pool, optionally provided to `init_slab_mgr()` */
+static char* g_mem_pool = NULL;
 static bool g_alloc_from_low = true; /* allocate from low addresses if true, from high if false */
-static void* g_mem_pool_end = &g_mem_pool[POOL_SIZE];
-static void* g_low  = g_mem_pool;
-static void* g_high = &g_mem_pool[POOL_SIZE];
+static void* g_mem_pool_end;
 
-#define STARTUP_SIZE 2
+/* `g_low` and `g_high` are protected by `SYSTEM_LOCK()` */
+static void* g_low;
+static void* g_high;
 
 static inline void* __malloc(size_t size);
 static inline void __free(void* addr, size_t size);
@@ -34,35 +34,35 @@ static inline void __free(void* addr, size_t size);
 
 #include "slabmgr.h"
 
-/* caller (slabmgr.h) releases g_slab_mgr_lock before calling this function (this must be reworked
- * in the future), so grab the lock again to protect g_low/g_high */
 static inline void* __malloc(size_t size) {
     void* addr = NULL;
 
     size = ALIGN_UP(size, MIN_MALLOC_ALIGNMENT);
 
-    SYSTEM_LOCK();
-    if (g_low + size <= g_high) {
-        /* alternate allocating objects from low and high addresses of available memory pool; this
-         * allows to free memory for patterns like "malloc1 - malloc2 - free1" (seen in e.g.
-         * realloc) */
-        if (g_alloc_from_low) {
-            addr = g_low;
-            g_low += size;
-        } else {
-            addr = g_high - size;
-            g_high -= size;
+    /* Use `g_mem_pool`, if available */
+    if (g_mem_pool) {
+        SYSTEM_LOCK();
+        if (g_low + size <= g_high) {
+            /* alternate allocating objects from low and high addresses of available memory pool;
+             * this allows to free memory for patterns like "malloc1 - malloc2 - free1" (seen in
+             * e.g. realloc) */
+            if (g_alloc_from_low) {
+                addr = g_low;
+                g_low += size;
+            } else {
+                addr = g_high - size;
+                g_high -= size;
+            }
+            g_alloc_from_low = !g_alloc_from_low; /* switch alloc direction for next malloc */
         }
-        g_alloc_from_low = !g_alloc_from_low; /* switch alloc direction for next malloc */
+        SYSTEM_UNLOCK();
+        if (addr)
+            return addr;
     }
-    SYSTEM_UNLOCK();
-    if (addr)
-        return addr;
 
-    /* At this point, we depleted the pre-allocated memory pool of POOL_SIZE. Let's fall back to
-     * PAL-internal allocations. PAL allocator must be careful though because LibOS doesn't know
-     * about PAL-internal memory, limited via manifest option `loader.pal_internal_mem_size` and
-     * thus this malloc may return -ENOMEM. */
+    /* We could not use `g_mem_pool`, let's fall back to PAL-internal allocations. PAL allocator
+     * must be careful though because LibOS doesn't know about PAL-internal memory, limited via
+     * manifest option `loader.pal_internal_mem_size` and thus this malloc may return -ENOMEM. */
     int ret = _DkVirtualMemoryAlloc(&addr, ALLOC_ALIGN_UP(size), PAL_ALLOC_INTERNAL,
               PAL_PROT_READ | PAL_PROT_WRITE);
     if (ret < 0) {
@@ -76,15 +76,13 @@ static inline void* __malloc(size_t size) {
     return addr;
 }
 
-/* caller (slabmgr.h) releases g_slab_mgr_lock before calling this function (this must be reworked
- * in the future), so grab the lock again to protect g_low/g_high */
 static inline void __free(void* addr, size_t size) {
     if (!addr)
         return;
 
     size = ALIGN_UP(size, MIN_MALLOC_ALIGNMENT);
 
-    if (addr >= (void*)g_mem_pool && addr < g_mem_pool_end) {
+    if (g_mem_pool && addr >= (void*)g_mem_pool && addr < g_mem_pool_end) {
         SYSTEM_LOCK();
         if (addr == g_high) {
             /* reclaim space of last object allocated at high addresses */
@@ -113,16 +111,25 @@ static inline void __free(void* addr, size_t size) {
 
 static SLAB_MGR g_slab_mgr = NULL;
 
-void init_slab_mgr(void) {
+void init_slab_mgr(char* mem_pool, size_t mem_pool_size) {
     assert(!g_slab_mgr);
+
+    if (mem_pool) {
+        g_mem_pool = mem_pool;
+        g_mem_pool_end = mem_pool + mem_pool_size;
+
+        g_low = g_mem_pool;
+        g_high = g_mem_pool_end;
+
+#ifdef ASAN
+        /* Poison all of `mem_pool` initially */
+        asan_poison_region((uintptr_t)mem_pool, mem_pool_size, ASAN_POISON_HEAP_LEFT_REDZONE);
+#endif
+    }
 
     g_slab_mgr = create_slab_mgr();
     if (!g_slab_mgr)
         INIT_FAIL(PAL_ERROR_NOMEM, "cannot initialize slab manager");
-#ifdef ASAN
-    /* Poison all of `g_mem_pool` initially */
-    asan_poison_region((uintptr_t)&g_mem_pool, sizeof(g_mem_pool), ASAN_POISON_HEAP_LEFT_REDZONE);
-#endif
 }
 
 void* malloc(size_t size) {


boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/192)* status: 0 of 6 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/192#-MnWST1c114rqgAPb8Zt:-MnWST1dBuG5_kpLRLby:b-aces74) (no related file):*
Could we instead just add `64MB` to `pal_internal_mem_size` if the resulting value is smaller than whole enclave size? This way we could be backward compatible (and I don't think it's that much worse of a solution).
@pwmarcz @mkow @dimakuv

---

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/192)* status: 0 of 6 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/192#-MnWST1c114rqgAPb8Zt:-MnWVoI55jdT5-NOTiV6:b-pcsmpp) (no related file):*

Sounds like a good solution. Also, Jenkins just made me aware that we do need some initial memory just to parse a big manifest...
</blockquote></details>

On the other hand, I'm trying to code it, and without any name changes it sounds kind of hacky: `g_pal_internal_mem_size` is actually `loader.pal_internal_mem_size + 64MB`. So if I wanted an option that worked like that, I would rename it to `loader.pal_additional_mem_size` anyway.

I'm curious what others think.

---

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/192)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/192#-MnWST1c114rqgAPb8Zt:-MnWrVOX3aDa6spl_HlV:b-dr276m) (no related file):*

On the other hand, I'm trying to code it, and without any name changes it sounds kind of hacky: `g_pal_internal_mem_size` is actually `loader.pal_internal_mem_size + 64MB`. So if I wanted an option that worked like that, I would rename it to `loader.pal_additional_mem_size` anyway.

I'm curious what others think.
</blockquote></details>

I want to be backward compatible, because this "PAL internal memory" is a hack itself, and all this memory stuff is very brittle and very confusing to end users. Once they found the magic PAL-memory size, let them just stick with it.

In some future, we will simply remove this manifest option and never have this problem again. Until then, let's be backward compatible.

---
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/192#-MnWrrbq7y7T1LcZI_gb:-MnWrrbr7SvjPfSHRnpb:ba40jbh) (no related file):*
@pwmarcz As you already may have realized, you actually need two PAL-internal-memory allocations:
1. Even before reading/parsing the TOML manifest -- representing manifest file itself requires some (theoretically arbitrary) amount of memory, even before you can read `loader.pal_internal_mem_size`.
2. After reading/parsing the TOML manifest -- adjust the size by the amount read from `loader.pal_internal_mem_size`.

Your current PR doesn't do (1) right now. Please add it.

---
___
*[Documentation/manifest-syntax.rst, line 266 at r1](https://reviewable.io/reviews/gramineproject/gramine/192#-MnWr-WY1S4Qd7KLGETg:-MnWr-WY1S4Qd7KLGETh:b-txtw3d) ([raw file](https://github.com/gramineproject/gramine/blob/135e002da906b8ffa170e0c72c32bbeee02d4cf5/Documentation/manifest-syntax.rst#L266)):*
> ```reStructuredText
>    If your manifest previously specified ``loader.pal_internal_mem_size``, you
>    probably need to increase it by 64MB to account for this change (for example,
>    change ``64M`` to ``128M``).
> ```

I dislike this change in semantics. This is *very* visible to end users. I understand that the previous version was weird, and your version is slightly less weird, but this whole "PAL internal memory" manifest option is a terrible hack. Therefore, I don't want users to curse us just because we changed one big hack with a slightly smaller hack.

One day we'll remove all these hacks, by merging LibOS and PAL memory management logic. But until that day, let's just keep the hack the same.
___
*[Pal/regression/manifest.template, line 9 at r1](https://reviewable.io/reviews/gramineproject/gramine/192#-MnWqse_ALTgvGIK2GXy:-MnWqse_ALTgvGIK2GXz:b-m7f72x) ([raw file](https://github.com/gramineproject/gramine/blob/3db1e28e27622cbf1265650cc92a87281803d31d/Pal/regression/manifest.template#L9)):*
> ```
> # PAL tests use `DkVirtualMemoryAlloc(PAL_ALLOC_INTERNAL)` which must allocate in the PAL-internal
> # part of the Gramine memory
> loader.pal_internal_mem_size = "64M"
> ```

Why did you remove this? This should be `128MB` now, with your change in semantics.
___
*[Pal/src/host/Linux/db_main.c, line 328 at r1](https://reviewable.io/reviews/gramineproject/gramine/192#-MnWqZO2D7dqnfV1dWjV:-MnWqZO2D7dqnfV1dWjW:b-r0cr2p) ([raw file](https://github.com/gramineproject/gramine/blob/135e002da906b8ffa170e0c72c32bbeee02d4cf5/Pal/src/host/Linux/db_main.c#L328)):*
> ```C
>             INIT_FAIL(PAL_ERROR_NOMEM, "Out of memory");
> 
>         ret = read_text_file_to_cstr(manifest_path, &manifest);
> ```

Isn't it too brittle? This function will try to malloc as much memory as the size of the manifest file, and if the manifest is big enough, we fail without any chance of recovery in Linux PAL.
___
*[Pal/src/host/Linux-SGX/db_main.c, line 804 at r1](https://reviewable.io/reviews/gramineproject/gramine/192#-MnWpJ0Y2uC0c7VyKSO-:-MnWpJ0Y2uC0c7VyKSO0:bv0twvg) ([raw file](https://github.com/gramineproject/gramine/blob/135e002da906b8ffa170e0c72c32bbeee02d4cf5/Pal/src/host/Linux-SGX/db_main.c#L804)):*
> ```C
>         log_always("Detected a huge manifest, preallocating 64MB of internal memory.");
>         g_pal_internal_mem_size += 128 * 1024 * 1024; /* 5MB manifest -> 1 + 128 MB PAL mem */
>     }
> ```

You need a final `else` here which will set `g_pal_internal_mem_size` to 64MB.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/192)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/192#-MnWrrbq7y7T1LcZI_gb:-Mn_HSivAoGXdzdCLhGr:bsdd7v8) (no related file):*

@pwmarcz As you already may have realized, you actually need two PAL-internal-memory allocations:
1. Even before reading/parsing the TOML manifest -- representing manifest file itself requires some (theoretically arbitrary) amount of memory, even before you can read `loader.pal_internal_mem_size`.
2. After reading/parsing the TOML manifest -- adjust the size by the amount read from `loader.pal_internal_mem_size`.

Your current PR doesn't do (1) right now. Please add it.
</blockquote></details>

Okay, so we do want the 64MB up front. At the same time I don't want to pile on more hacks. Here's what I found out:

* Apparently `Linux-SGX` contains a functional VMA manager inside enclave (`enclave_pages.c`) and by the time we initialize the allocator, we can ask for new pages. (Your "preallocating memory for manifest" trick takes advantage of that: we can bump up `g_pal_internal_mem_size` and the memory is available immediately). So no static buffer needed.

* `Linux` is problematic, because we don't have this VMA tracking. We read `pal_internal_mem_size`, and allocate it as a single range using `mmap()` (and feed new pages to the allocator by moving an index forward, so we cannot ever reuse a freed page...) So here we need some initial pool before we parse the manifest.

I don't want to rewrite PAL memory allocation now, so how about this:

* Remove the static buffer from PAL allocator (`slab.c`)
* In `Linux-SGX`, start with `g_pal_internal_mem_size = 64MB` (which will enable `malloc` right away), then, after parsing the manifest, set `g_pal_internal_mem_size = 64MB + loader.pal_internal_mem_size` (to keep backward compatibility)
* In `Linux`, add the static buffer to the implementation of `_DkVirtualMemoryAlloc`

I tried it out and it seems to work. It doesn't remove the static buffer hack completely, but it moves it from global to host-specific. And it solves my problem with Linux-SGX PAL having a huge static buffer.

---
___
*[Pal/regression/manifest.template, line 9 at r1](https://reviewable.io/reviews/gramineproject/gramine/192#-MnWqse_ALTgvGIK2GXy:-Mn_OfWO5QBQzngtjdsm:ba9isco) ([raw file](https://github.com/gramineproject/gramine/blob/3db1e28e27622cbf1265650cc92a87281803d31d/Pal/regression/manifest.template#L9)):*

Why did you remove this? This should be `128MB` now, with your change in semantics.
</blockquote></details>

It's not about the actual amount. This setting was needed so that `DkVirtualMemoryAlloc(PAL_ALLOC_INTERNAL)` works at all; I want it to work even without customizing the setting.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/192)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/192#-MnWrrbq7y7T1LcZI_gb:-Mn_a7OM63U7ANmHkVqd:b5oltli) (no related file):*

Okay, so we do want the 64MB up front. At the same time I don't want to pile on more hacks. Here's what I found out:

* Apparently `Linux-SGX` contains a functional VMA manager inside enclave (`enclave_pages.c`) and by the time we initialize the allocator, we can ask for new pages. (Your "preallocating memory for manifest" trick takes advantage of that: we can bump up `g_pal_internal_mem_size` and the memory is available immediately). So no static buffer needed.

* `Linux` is problematic, because we don't have this VMA tracking. We read `pal_internal_mem_size`, and allocate it as a single range using `mmap()` (and feed new pages to the allocator by moving an index forward, so we cannot ever reuse a freed page...) So here we need some initial pool before we parse the manifest.

I don't want to rewrite PAL memory allocation now, so how about this:

* Remove the static buffer from PAL allocator (`slab.c`)
* In `Linux-SGX`, start with `g_pal_internal_mem_size = 64MB` (which will enable `malloc` right away), then, after parsing the manifest, set `g_pal_internal_mem_size = 64MB + loader.pal_internal_mem_size` (to keep backward compatibility)
* In `Linux`, add the static buffer to the implementation of `_DkVirtualMemoryAlloc`

I tried it out and it seems to work. It doesn't remove the static buffer hack completely, but it moves it from global to host-specific. And it solves my problem with Linux-SGX PAL having a huge static buffer.
</blockquote></details>

I agree with your proposed change. We don't care about memory constraints in the `Linux` case, so we can have a huge static buffer without any consequences.

---
___
*[Pal/regression/manifest.template, line 9 at r1](https://reviewable.io/reviews/gramineproject/gramine/192#-MnWqse_ALTgvGIK2GXy:-Mn_aPHWE3NfOdBJRS8J:bq2y95e) ([raw file](https://github.com/gramineproject/gramine/blob/3db1e28e27622cbf1265650cc92a87281803d31d/Pal/regression/manifest.template#L9)):*

It's not about the actual amount. This setting was needed so that `DkVirtualMemoryAlloc(PAL_ALLOC_INTERNAL)` works at all; I want it to work even without customizing the setting.
</blockquote></details>

But why does it work in your PR? I don't see any changes to this logic... What I'm missing?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/192)* status: 1 of 7 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/192#-MnWST1c114rqgAPb8Zt:-Mn_j6AS6YdH--38E1Bx:bz42rjl) (no related file):*

I want to be backward compatible, because this "PAL internal memory" is a hack itself, and all this memory stuff is very brittle and very confusing to end users. Once they found the magic PAL-memory size, let them just stick with it.

In some future, we will simply remove this manifest option and never have this problem again. Until then, let's be backward compatible.
</blockquote></details>

OK. The new version is backward compatible: internally the allocation is a bit different, but the initial 64MB is available, and isn't included in the manifest value.

---
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/192#-MnWrrbq7y7T1LcZI_gb:-Mn_jGhP-M2Sip1kteWp:bitysfh) (no related file):*

I agree with your proposed change. We don't care about memory constraints in the `Linux` case, so we can have a huge static buffer without any consequences.
</blockquote></details>

OK. I pushed a new version.

BTW, this change (in the SGX case) is better than measuring 64MB up front for one more reason: if we end up needing less than 64 megabytes, we don't even need to zero it.

---
___
*[Pal/regression/manifest.template, line 9 at r1](https://reviewable.io/reviews/gramineproject/gramine/192#-MnWqse_ALTgvGIK2GXy:-Mn_hY7ZEPE6ObGFqLuZ:b-vevzl7) ([raw file](https://github.com/gramineproject/gramine/blob/3db1e28e27622cbf1265650cc92a87281803d31d/Pal/regression/manifest.template#L9)):*

But why does it work in your PR? I don't see any changes to this logic... What I'm missing?
</blockquote></details>

It worked before because I changed the `loader.pal_internal_mem_size` default from 0 to 64M.

In the current version, it's *still* not necessary: `loader.pal_internal_mem_size` defaults to 0, but `DkVirtualMemoryAlloc(PAL_ALLOC_INTERNAL)` can now use the initial 64 megabytes.
___
*[Pal/src/host/Linux/db_memory.c, line 58 at r2](https://reviewable.io/reviews/gramineproject/gramine/192#-Mn_jRGOEVd2oRfgbwBE:-Mn_jRGOEVd2oRfgbwBF:b-9b1nj6) ([raw file](https://github.com/gramineproject/gramine/blob/511ddfb9ddae5b76d2b989bf3bd18a3d370cc13f/Pal/src/host/Linux/db_memory.c#L58)):*
> ```C
>     /* Try the initial pool first */
>     if (g_mem_pool_used + size <= PAL_INITIAL_MEM_SIZE) {
>         addr = g_mem_pool + g_mem_pool_used;
> ```

FYI: Compared to the previous version, this rounds up to whole pages, and does not use the "alternate allocating from low/high" trick, but I'm not sure if that changes much.
___
*[Pal/src/host/Linux-SGX/db_main.c, line 804 at r1](https://reviewable.io/reviews/gramineproject/gramine/192#-MnWpJ0Y2uC0c7VyKSO-:-Mn_iJZ_3IXNS4in-tBz:b-aw5s7t) ([raw file](https://github.com/gramineproject/gramine/blob/135e002da906b8ffa170e0c72c32bbeee02d4cf5/Pal/src/host/Linux-SGX/db_main.c#L804)):*

You need a final `else` here which will set `g_pal_internal_mem_size` to 64MB.
</blockquote></details>

`g_pal_internal_mem_size` is now initialized just before `init_slab_mgr()`, and increased here if necessary.
___
*[Documentation/manifest-syntax.rst, line 266 at r1](https://reviewable.io/reviews/gramineproject/gramine/192#-MnWr-WY1S4Qd7KLGETg:-Mn_k10oD-AoRf4O2E-N:b2x85fo) ([raw file](https://github.com/gramineproject/gramine/blob/135e002da906b8ffa170e0c72c32bbeee02d4cf5/Documentation/manifest-syntax.rst#L266)):*

I dislike this change in semantics. This is *very* visible to end users. I understand that the previous version was weird, and your version is slightly less weird, but this whole "PAL internal memory" manifest option is a terrible hack. Therefore, I don't want users to curse us just because we changed one big hack with a slightly smaller hack.

One day we'll remove all these hacks, by merging LibOS and PAL memory management logic. But until that day, let's just keep the hack the same.
</blockquote></details>

Reverted, as discussed above.
___
*[Pal/src/host/Linux/db_main.c, line 328 at r1](https://reviewable.io/reviews/gramineproject/gramine/192#-MnWqZO2D7dqnfV1dWjV:-Mn_iBYiBukAjtbKM7c_:bhd8i9h) ([raw file](https://github.com/gramineproject/gramine/blob/135e002da906b8ffa170e0c72c32bbeee02d4cf5/Pal/src/host/Linux/db_main.c#L328)):*

Isn't it too brittle? This function will try to malloc as much memory as the size of the manifest file, and if the manifest is big enough, we fail without any chance of recovery in Linux PAL.
</blockquote></details>

You're right. The new version still keeps a 64MB initial pool here.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/192)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @mkow and @pwmarcz)

___
*[Pal/src/host/Linux/db_memory.c, line 58 at r2](https://reviewable.io/reviews/gramineproject/gramine/192#-Mn_jRGOEVd2oRfgbwBE:-Mn_mQ6vCugSl9Bw98_m:b-7cbrau) ([raw file](https://github.com/gramineproject/gramine/blob/511ddfb9ddae5b76d2b989bf3bd18a3d370cc13f/Pal/src/host/Linux/db_memory.c#L58)):*

FYI: Compared to the previous version, this rounds up to whole pages, and does not use the "alternate allocating from low/high" trick, but I'm not sure if that changes much.
</blockquote></details>

The "alternate allocating from low/high" trick was added historically for Linux-SGX PAL: manifest parsing in the SGX case is much more demanding, with a lot of mallocs and frees of internal-TOML objects. Thus, we could quickly deplete the 64MB of that static buffer we had in common PAL code. To ease that pressure on internal PAL memory, I introduced that poor-man's "alternate allocating" scheme -- to at least catch the cases of `malloc - malloc - free - free` cases which were very frequent in manifest parsing.

But I would still keep it. You forgot to add this case in `_DkVirtualMemoryFree()`, see my comment below. Right now with your "dummy" memory allocator, Linux PAL will only keep accumulating and accumulating internal objects, never freeing them. So in a long-running workload, we will 100% hit this limit of `initial pool is exhausted`. And if the end user didn't specify `pal.internal_mem_size`, then the application is 100% doomed.

At runtime, PAL internal memory allocations almost always happen in two ways: `malloc - do smth - free` or `malloc - malloc - do smth - free -free`. So actually having the old scheme is beneficial for long-running workloads.
___
*[Pal/src/host/Linux/db_memory.c, line 109 at r2](https://reviewable.io/reviews/gramineproject/gramine/192#-Mn_oNBJBIkCRy4PYfow:-Mn_oNBJBIkCRy4PYfox:blebic) ([raw file](https://github.com/gramineproject/gramine/blob/511ddfb9ddae5b76d2b989bf3bd18a3d370cc13f/Pal/src/host/Linux/db_memory.c#L109)):*
> ```C
> 
> int _DkVirtualMemoryFree(void* addr, size_t size) {
>     int ret = DO_SYSCALL(munmap, addr, size);
> ```

But you could have `addr` pointing into the static buffer `g_mem_pool`! Please add this case.
___
*[Pal/src/host/Linux-SGX/db_main.c, line 723 at r2](https://reviewable.io/reviews/gramineproject/gramine/192#-Mn_oadTEYRVhI-2ru9y:-Mn_oadTEYRVhI-2ru9z:bdvpt2w) ([raw file](https://github.com/gramineproject/gramine/blob/511ddfb9ddae5b76d2b989bf3bd18a3d370cc13f/Pal/src/host/Linux-SGX/db_main.c#L723)):*
> ```C
> 
>     /* set up page allocator and slab manager */
>     g_pal_internal_mem_size = PAL_INITIAL_MEM_SIZE;
> ```

Why not just assign it during variable declaration?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/192)* status: 1 of 7 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "squash! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[Pal/src/host/Linux-SGX/db_main.c, line 723 at r2](https://reviewable.io/reviews/gramineproject/gramine/192#-Mn_oadTEYRVhI-2ru9y:-Mn_wNpFAp1jDBaRPZ2m:b-egjqih) ([raw file](https://github.com/gramineproject/gramine/blob/511ddfb9ddae5b76d2b989bf3bd18a3d370cc13f/Pal/src/host/Linux-SGX/db_main.c#L723)):*

Why not just assign it during variable declaration?
</blockquote></details>

It made some sense because we're about to start using it. But a global initialization looks simpler. Changed.
___
*[Pal/src/host/Linux/db_memory.c, line 58 at r2](https://reviewable.io/reviews/gramineproject/gramine/192#-Mn_jRGOEVd2oRfgbwBE:-Mna3KV658E8w-bbgLbq:b-gnp1wf) ([raw file](https://github.com/gramineproject/gramine/blob/511ddfb9ddae5b76d2b989bf3bd18a3d370cc13f/Pal/src/host/Linux/db_memory.c#L58)):*

The "alternate allocating from low/high" trick was added historically for Linux-SGX PAL: manifest parsing in the SGX case is much more demanding, with a lot of mallocs and frees of internal-TOML objects. Thus, we could quickly deplete the 64MB of that static buffer we had in common PAL code. To ease that pressure on internal PAL memory, I introduced that poor-man's "alternate allocating" scheme -- to at least catch the cases of `malloc - malloc - free - free` cases which were very frequent in manifest parsing.

But I would still keep it. You forgot to add this case in `_DkVirtualMemoryFree()`, see my comment below. Right now with your "dummy" memory allocator, Linux PAL will only keep accumulating and accumulating internal objects, never freeing them. So in a long-running workload, we will 100% hit this limit of `initial pool is exhausted`. And if the end user didn't specify `pal.internal_mem_size`, then the application is 100% doomed.

At runtime, PAL internal memory allocations almost always happen in two ways: `malloc - do smth - free` or `malloc - malloc - do smth - free -free`. So actually having the old scheme is beneficial for long-running workloads.
</blockquote></details>

OK, thanks for the explanation. But still, if we exhaust this initial pool, then we fall back to the "dummy" system in this file and can (theoretically) quickly run out of memory, right?

Anyway, as said in a comment below, I reverted to your "low/high" system.
___
*[Pal/src/host/Linux/db_memory.c, line 109 at r2](https://reviewable.io/reviews/gramineproject/gramine/192#-Mn_oNBJBIkCRy4PYfow:-Mn_vt5P8VmRZ36hIGBG:bupzicj) ([raw file](https://github.com/gramineproject/gramine/blob/511ddfb9ddae5b76d2b989bf3bd18a3d370cc13f/Pal/src/host/Linux/db_memory.c#L109)):*

But you could have `addr` pointing into the static buffer `g_mem_pool`! Please add this case.
</blockquote></details>

I assumed we could treat this memory same as dynamically mapped pages (i.e. call `mmap` on it in `_DkVirtualMemoryAlloc`, call `munmap` on it here).

But yeah, it's too hacky. And also PAL regression test show that this breaks some assumptions we have for mapped memory.

So it looks like this memory should not be handled by `DkVirtualMemoryAlloc`. Instead, I'll revert to `slab.c` managing it, but initialize it from outside. As a bonus, I don't need a static array in Linux PAL either.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/192)* status: 1 of 8 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "squash! " and "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[Pal/regression/Memory.c, line 74 at r3](https://reviewable.io/reviews/gramineproject/gramine/192#-Mna_wF794dVH4sDRAGh:-Mna_wF794dVH4sDRAGi:b-odzy47) ([raw file](https://github.com/gramineproject/gramine/blob/bf36f81b92ad1b14711cc93c7bb3dff72c82914a/Pal/regression/Memory.c#L74)):*
> ```C
>     /* TODO: This does not take into account `pal_control.preloaded_ranges`; we are not allowed to
>      * ask for memory overlapping with these ranges */
>     void* mem3 = (void*)pal_control.user_address.start;
> ```

FYI: This test crashes because the memory `user_address.end - UNIT` is actually reserved by PAL for the initial pool (and remapping this memory ends up zeroing the TCB...)

AFAIK, this is exactly why we report all such memory in `preloaded_ranges`, and LibOS respects these ranges when allocating memory, but this test doesn't care.

Seeing the `please someone rm rf this file` here, I decided it's enough to add this workaround, but if you disagree, I can do better and try to avoid these ranges.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/192)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "squash! " and "fixup! " found in commit messages' one-liners (waiting on @mkow and @pwmarcz)

___
*[Pal/regression/Memory.c, line 74 at r3](https://reviewable.io/reviews/gramineproject/gramine/192#-Mna_wF794dVH4sDRAGh:-Mnbwume0eLhyaHaKPsY:b991tw6) ([raw file](https://github.com/gramineproject/gramine/blob/bf36f81b92ad1b14711cc93c7bb3dff72c82914a/Pal/regression/Memory.c#L74)):*

FYI: This test crashes because the memory `user_address.end - UNIT` is actually reserved by PAL for the initial pool (and remapping this memory ends up zeroing the TCB...)

AFAIK, this is exactly why we report all such memory in `preloaded_ranges`, and LibOS respects these ranges when allocating memory, but this test doesn't care.

Seeing the `please someone rm rf this file` here, I decided it's enough to add this workaround, but if you disagree, I can do better and try to avoid these ranges.
</blockquote></details>

I wouldn't mind deleting this file in this PR :) But your workaround is also fine.

Don't you want to add your own `XXX` top-level comment? :) Maybe this file will collect all Gramine devs.
___
*[Pal/src/host/Linux-SGX/db_main.c, line 647 at r3](https://reviewable.io/reviews/gramineproject/gramine/192#-Mnc0-gV3sMff3sITrP2:-Mnc0-gW70pDzEpaQ623:b7e7kog) ([raw file](https://github.com/gramineproject/gramine/blob/bf36f81b92ad1b14711cc93c7bb3dff72c82914a/Pal/src/host/Linux-SGX/db_main.c#L647)):*
> ```C
> }
> 
> ```

Why did it work before, without the "no sanitize" hint?

Or it never did because we don't have a test with this manifest option enabled?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/192)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "squash! " and "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[Pal/regression/Memory.c, line 74 at r3](https://reviewable.io/reviews/gramineproject/gramine/192#-Mna_wF794dVH4sDRAGh:-MneizEr1DCVe3MMk5XQ:bva6zve) ([raw file](https://github.com/gramineproject/gramine/blob/bf36f81b92ad1b14711cc93c7bb3dff72c82914a/Pal/regression/Memory.c#L74)):*

I wouldn't mind deleting this file in this PR :) But your workaround is also fine.

Don't you want to add your own `XXX` top-level comment? :) Maybe this file will collect all Gramine devs.
</blockquote></details>

I'd feel kind of bad deleting it and not replacing with anything else... The `context->rip++` stuff is awful, but testing memory protection (probably *without* recovering from segfault) is a valid thing to do.
___
*[Pal/src/host/Linux-SGX/db_main.c, line 647 at r3](https://reviewable.io/reviews/gramineproject/gramine/192#-Mnc0-gV3sMff3sITrP2:-MnekA3m65F2WHkB60co:b-1oxv2p) ([raw file](https://github.com/gramineproject/gramine/blob/bf36f81b92ad1b14711cc93c7bb3dff72c82914a/Pal/src/host/Linux-SGX/db_main.c#L647)):*

Why did it work before, without the "no sanitize" hint?

Or it never did because we don't have a test with this manifest option enabled?
</blockquote></details>

We have a test that has this option enables specifically to have coverage (`openmp`).

It passed before, because the heap (i.e. the range of pages reserved for `DkVirtualMemoryAlloc`) was completely unused, because all PAL's allocation happened in the initial pool. Now, the slab allocator takes some pages from heap immediately, and poisons them with "heap left redzone" until they're ready to be used. So a read from such a memory might fail.

BTW: ASan assumes that if a page is not mapped, then there's no sense in poisoning the memory, because accessing it will segfault anyway. But that's not true for SGX 1, because pages not "mapped"  by Gramine are still accessible, so as a result ASan's protection is weaker under SGX.

Probably we could keep unused parts of the enclave space poisoned all the time; poison stack guard pages, etc. I'm not sure how complicated that would get. Worth trying out.

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/192)* status: all files reviewed, 1 unresolved discussion, "squash! " and "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[Pal/src/host/Linux-SGX/db_main.c, line 647 at r3](https://reviewable.io/reviews/gramineproject/gramine/192#-Mnc0-gV3sMff3sITrP2:-MnyUV3S4vzYRH3zXxuu:b-r80io7) ([raw file](https://github.com/gramineproject/gramine/blob/bf36f81b92ad1b14711cc93c7bb3dff72c82914a/Pal/src/host/Linux-SGX/db_main.c#L647)):*

We have a test that has this option enables specifically to have coverage (`openmp`).

It passed before, because the heap (i.e. the range of pages reserved for `DkVirtualMemoryAlloc`) was completely unused, because all PAL's allocation happened in the initial pool. Now, the slab allocator takes some pages from heap immediately, and poisons them with "heap left redzone" until they're ready to be used. So a read from such a memory might fail.

BTW: ASan assumes that if a page is not mapped, then there's no sense in poisoning the memory, because accessing it will segfault anyway. But that's not true for SGX 1, because pages not "mapped"  by Gramine are still accessible, so as a result ASan's protection is weaker under SGX.

Probably we could keep unused parts of the enclave space poisoned all the time; poison stack guard pages, etc. I'm not sure how complicated that would get. Worth trying out.
</blockquote></details>

I see. So ASan doesn't do anything about unmapped pages (or more specifically, our current usage of ASan doesn't do anything about unmapped pages). So currently it is possible to access such pages without ASan complaining. Anyway, resolving, this is for another PR (if at all).

boryspoplawski (dismissed):
> Reviewed 5 of 7 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/192)* status: all files reviewed, all discussions resolved, "squash! " and "fixup! " found in commit messages' one-liners

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/192)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/192)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #194, opened 2021-11-03
[Pal,tests,CI-Examples] Deprecate `loader.preload` and use `loader.entrypoint` instead

### LINKED ISSUE #158: Remove preload support in PAL and remove `loader.preload` in manifest
## Description of the problem

After #152, we should remove the "preload libraries" support in PAL and `loader.preload` in the manifest.

There is no reason to have PAL-level preloaded libraries. In reality, we only ever load one ELF binary: either the PAL regression test or the LibOS library.

So we can remove the following:
- Remove preload libraries in PAL regression tests
- Remove `loader.preload = "libsysdb.so"`
- Remove `g_loaded_maps` -- we will only have `g_pal_map` and resolve symbols through it
- Change `CALL_ENTRY` in `db_rtld.c` and the corresponding caller code (only call once)
- Get rid of `enum elf_object_type` -- we will only have EXEC type (for a single loaded ELF binary)

dimakuv:
## Description of the changes 

Previously, PAL had the ability to preload several libraries via `loader.preload = "pal_lib1,pal_lib2,pal_lib3"` syntax. This ability was supposed to be used as an interface for "PAL plugins". In practice, the only loaded library (and simultaneously the entrypoint) was the LibOS.

This PR removes this unused ability. The syntax `loader.preload` is deprecated and is currently used as an alias to `loader.entrypoint`. PAL regression tests that tested this preload functionality are removed. Other tests and examples are updated.

Fixes #158.

Corresponding PRs (must be merged after this one):
- https://github.com/gramineproject/examples/pull/8
- https://github.com/gramineproject/gsc/pull/28

## How to test this PR? 

All tests and examples are updated.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/194)


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/194)* status: 0 of 57 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

___
*[CI-Examples/bash/manifest.template, line 36 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-MnZxjPB7XSfGdHCKIY0:-MnZxjPB7XSfGdHCKIY1:bmb5svg) ([raw file](https://github.com/gramineproject/gramine/blob/fba5653a8c34c3eef72d19faedd6f8fd48f21653/CI-Examples/bash/manifest.template#L36)):*
> ```
> 
> sgx.trusted_files = [
>   "file:{{ gramine.libos }}",
> ```

FYI: Now we require to add explicitly the LibOS binary, similar to how we require it for `libos.entrypoint` file.
___
*[Documentation/manifest-syntax.rst, line 795 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-MnZyFRbBE6rulf6oDmb:-MnZyFRbBE6rulf6oDmc:b-lzr3wv) ([raw file](https://github.com/gramineproject/gramine/blob/fba5653a8c34c3eef72d19faedd6f8fd48f21653/Documentation/manifest-syntax.rst#L795)):*
> ```reStructuredText
> ::
> 
>    loader.preload = "[URI]"
> ```

FYI: We cannot completely remove `loader.preload` in this PR, because we have `examples` and `gsc` repositories which still have the old syntax. I will submit those PRs too.
___
*[Pal/include/pal_rtld.h, line 45 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-MnZywXx4lHUCY0ZppNf:-MnZywXy7GTARsHNrvRk:b4dj6hv) ([raw file](https://github.com/gramineproject/gramine/blob/3db1e28e27622cbf1265650cc92a87281803d31d/Pal/include/pal_rtld.h#L45)):*
> ```objc
> 
> /* loading ELF binaries */
> int load_elf_object(const char* uri, enum elf_object_type type);
> ```

FYI: renamed `load_elf_object` -> `load_entrypoint`
___
*[Pal/src/db_rtld.c, line 625 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-Mn_-0TSDOjAY05W_3pr:-Mn_-0TSDOjAY05W_3ps:b-5g35fa) ([raw file](https://github.com/gramineproject/gramine/blob/fba5653a8c34c3eef72d19faedd6f8fd48f21653/Pal/src/db_rtld.c#L625)):*
> ```C
> void* rbp_before_call = NULL;
> 
> #define CALL_ENTRY(l, cookies)                                                       \
> ```

@mkow I think you didn't like this assembly. What exactly do you want to change here?
___
*[Pal/src/host/Linux/db_main.c, line 31 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-Mn_05Od70nlAB5LjQsQ:-Mn_05Od70nlAB5LjQsR:b-1tp3ms) ([raw file](https://github.com/gramineproject/gramine/blob/3db1e28e27622cbf1265650cc92a87281803d31d/Pal/src/host/Linux/db_main.c#L31)):*
> ```C
> 
> /* pal_start is the entry point of libpal.so, which calls pal_main */
> #define _ENTRY pal_start
> ```

FYI: These two were dead code.
___
*[Pal/src/host/Linux-SGX/db_main.c, line 36 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-Mn_08Uf4aO1uNHTqv8x:-Mn_08Ug-DofBZPWg7-c:b-1tp3ms) ([raw file](https://github.com/gramineproject/gramine/blob/3db1e28e27622cbf1265650cc92a87281803d31d/Pal/src/host/Linux-SGX/db_main.c#L36)):*
> ```C
> 
> #define RTLD_BOOTSTRAP
> #define _ENTRY enclave_entry
> ```

FYI: These two were dead code.
___
*[python/graminelibos/manifest.py, line 189 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-Mn_0Bp-EPWft2IM2NET:-Mn_0Bp-EPWft2IM2NEU:b4eivex) ([raw file](https://github.com/gramineproject/gramine/blob/fba5653a8c34c3eef72d19faedd6f8fd48f21653/python/graminelibos/manifest.py#L189)):*
> ```Python
>         if preload_str:
>             if not any(preload_str == tf['uri'] for tf in trusted_files):
>                 append_trusted_dir_or_file(trusted_files, preload_str)
> ```

@boryspoplawski Now `loader.preload` is a string, not a list of strings. Is my code good enough? Or is there a more pythonic way?
___
*[python/graminelibos/manifest.py, line 210 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-Mn_0Lic0i575RpHXgt4:-Mn_0Lic0i575RpHXgt5:b4eivex) ([raw file](https://github.com/gramineproject/gramine/blob/fba5653a8c34c3eef72d19faedd6f8fd48f21653/python/graminelibos/manifest.py#L210)):*
> ```Python
>         preload_str = self['loader']['preload']
>         if preload_str:
>             deps.add(uri2path(preload_str))
> ```

@boryspoplawski Now `loader.preload` is a string, not a list of strings. Is my code good enough? Or is there a more pythonic way?
___
*[tests/benchmarks/basic.manifest.template, line 1 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-Mn_0OgfCJeIHNFMoRVO:-Mn_0OgfCJeIHNFMoRVP:b5w88m5) ([raw file](https://github.com/gramineproject/gramine/blob/fba5653a8c34c3eef72d19faedd6f8fd48f21653/tests/benchmarks/basic.manifest.template#L1)):*
> ```
> #sgx.enable_stats = true
> ```

FYI: These `tests/benchmarks/` are completely out of sync with anything... Can we fix them or purge them? @mkow @boryspoplawski @woju

chiache (changes requested):
> Reviewed 57 of 57 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/194)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[Documentation/manifest-syntax.rst, line 67 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-Mne_nA10-Be7fAD7qP2:-Mne_nA10-Be7fAD7qP3:b-wb1bzo) ([raw file](https://github.com/gramineproject/gramine/blob/fba5653a8c34c3eef72d19faedd6f8fd48f21653/Documentation/manifest-syntax.rst#L67)):*
> ```reStructuredText
> ::
> 
>    pal.entrypoint= "[URI]"
> ```

My understanding is that this is the only `pal.xxx` syntax, right? Why not use `loader.entrypoint` like every other PAL-related syntax?
___
*[Documentation/manifest-syntax.rst, line 72 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-MneaZiR8e6FbdLvehKW:-MneaZiR8e6FbdLvehKX:b7qugbp) ([raw file](https://github.com/gramineproject/gramine/blob/fba5653a8c34c3eef72d19faedd6f8fd48f21653/Documentation/manifest-syntax.rst#L72)):*
> ```reStructuredText
> the first executable of the user application. Currently, there is only one LibOS
> implementation: ``libsysdb.so``.
> ```

I would also suggest adding some explanation for the difference between this syntax and `libos.entrypoint`, like this one is taking a URL, not a path, and the calling order of these entry points.

pwmarcz (changes requested):
> Reviewed 57 of 57 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/194)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @chiache, @dimakuv, and @mkow)

___
*[Documentation/manifest-syntax.rst, line 67 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo2sGL5-fmMTPcF4JL6:-Mo2sGL5-fmMTPcF4JL7:b-97hhf1) ([raw file](https://github.com/gramineproject/gramine/blob/fba5653a8c34c3eef72d19faedd6f8fd48f21653/Documentation/manifest-syntax.rst#L67)):*
> ```reStructuredText
> ::
> 
>    pal.entrypoint= "[URI]"
> ```

Missing space before `=`.
___
*[Documentation/manifest-syntax.rst, line 72 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-MneaZiR8e6FbdLvehKW:-Mo2sSia1QVtha2qRjfp:b-yiq9v1) ([raw file](https://github.com/gramineproject/gramine/blob/fba5653a8c34c3eef72d19faedd6f8fd48f21653/Documentation/manifest-syntax.rst#L72)):*

I would also suggest adding some explanation for the difference between this syntax and `libos.entrypoint`, like this one is taking a URL, not a path, and the calling order of these entry points.
</blockquote></details>

+1, although I think this can be mentioned under `libos.entrypoint`, since we describe it later.
___
*[Pal/regression/meson.build, line 121 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo2tS3a7STa0vm7rYUX:-Mo2tS3a7STa0vm7rYUY:b38ccj3) ([raw file](https://github.com/gramineproject/gramine/blob/3db1e28e27622cbf1265650cc92a87281803d31d/Pal/regression/meson.build#L121)):*
> ```Meson
> 
>         c_args: [
>             '-DSHARED',
> ```

Please also remove the `#ifdef SHARED` part from `Pal/regression/crt_init/arch/x86_64/user_start.S`.
___
*[Pal/src/db_main.c, line 516 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo2tqL38bCgf4rDp1Zr:-Mo2tqL38bCgf4rDp1Zs:b-9bdmar) ([raw file](https://github.com/gramineproject/gramine/blob/fba5653a8c34c3eef72d19faedd6f8fd48f21653/Pal/src/db_main.c#L516)):*
> ```C
> 
>         if (entrypoint_name)
>             log_warning("'loader.preload' is deprecated; consider switching to 'pal.entrypoint'");
> ```

"Please switch to 'pal.entrypoint'" would be better. "Consider" is IMO too weak, there's no upside to not switching, and the option will be removed sooner or later.

We already say "This trusted files syntax is deprecated, please move to the new one".
___
*[Pal/src/db_rtld.c, line 176 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo2vOCS6Q9j2XBxuNhk:-Mo2vOCS6Q9j2XBxuNhl:b-9fb7ai) ([raw file](https://github.com/gramineproject/gramine/blob/fba5653a8c34c3eef72d19faedd6f8fd48f21653/Pal/src/db_rtld.c#L176)):*
> ```C
>     const char* symbol_name = map->string_table + map->symbol_table[symbol_idx].st_name;
> 
>     /* first try to find in this ELF object itself */
> ```

Can you do something for `link_map == g_pal_map`? Either check it and don't run one of these loops, or assert that this isn't the case.

Context: Right now this doesn't happen, because there are no relocations in PAL that refer to its own symbols. But I'm working on ASan stack sanitization, and found out that in this case LLVM likes to generate some GOT entries anyway, which introduces such relocations. I'm not sure if I'll be able to get rid of them, but anyway, your code is able to handle them (once I move the "load symbol table" step earlier than "apply relocations").

But even if it doesn't happen yet, I think it would be good to acknowledge this case somehow.
___
*[Pal/src/db_rtld.c, line 625 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-Mn_-0TSDOjAY05W_3pr:-Mo2wxQL6d8qVbLeAXhV:b-o27l16) ([raw file](https://github.com/gramineproject/gramine/blob/fba5653a8c34c3eef72d19faedd6f8fd48f21653/Pal/src/db_rtld.c#L625)):*

@mkow I think you didn't like this assembly. What exactly do you want to change here?
</blockquote></details>

Not speaking for @mkow, but here's what I found:

* We start the new stack at `cookies`, which is a stack-allocated array. This is probably fine if we never return to the original function (`start_execution()`), but the code assumes we will return.

  This works because usually there's nothing interesting below `cookies`, but can is not guaranteed to work: it blew up for me when I enabled ASan stack sanitization (which added some instrumentation to `start_execution()`).
* We pass the "address to return to" in RDX. This is not supposed to be a return address, but `atexit` callback. The called program never returns, but is responsible for exiting by itself (and calling this callback during "normal" exit).
  BTW, if I'm reading this correctly, `musl` seems to omit that completely; it's weird because the SysV psABI seems to require it...
  ```c
  // dynlink.c
  CRTJMP((void *)aux[AT_ENTRY], argv-1);

  // reloc.h
  #define CRTJMP(pc,sp) __asm__ __volatile__( \
  	"mov %1,%%rsp ; jmp *%0" : : "r"(pc), "r"(sp) : "memory" )
  ```
* The LibOS actually does exit, but the PAL regression tests do not, so we do return to `start_execution`.

So I would propose:
* Remove saving RSP/RBP.
* Pass 0 in RDX, and explicitly say we ignore this `atexit` callback.
* Treat the code after `CALL_ENTRY` as unreachable (add `die_or_inf_loop()`).
* Ensure that the start code in PAL regression tests (`Pal/regression/crt_init/...`) exits after `main` returns, by calling `DkProcessExit`. As a bonus, the regression tests will finally have use the `main` return value as exit code :)
___
*[tests/benchmarks/basic.manifest.template, line 1 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-Mn_0OgfCJeIHNFMoRVO:-Mo2rVKuAVmTuP_P0sFZ:b-bn0swe) ([raw file](https://github.com/gramineproject/gramine/blob/fba5653a8c34c3eef72d19faedd6f8fd48f21653/tests/benchmarks/basic.manifest.template#L1)):*

FYI: These `tests/benchmarks/` are completely out of sync with anything... Can we fix them or purge them? @mkow @boryspoplawski @woju 
</blockquote></details>

Yeah, should we bother "fixing" this manifest if it won't even work?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/194)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @chiache, @dimakuv, and @mkow)

___
*[Pal/src/db_rtld.c, line 625 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-Mn_-0TSDOjAY05W_3pr:-Mo34SMy9xQHR2NN77vZ:b27925) ([raw file](https://github.com/gramineproject/gramine/blob/fba5653a8c34c3eef72d19faedd6f8fd48f21653/Pal/src/db_rtld.c#L625)):*
Sorry, editing mistakes, I didn't proofread my comment enough:

>can is not guaranteed

is not guaranteed

>have use the

use the

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/194)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @chiache, @dimakuv, and @mkow)

___
*[tests/benchmarks/basic.manifest.template, line 1 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-Mn_0OgfCJeIHNFMoRVO:-Mo3_z4g2nmiRzNzKwDW:b-ps3nkv) ([raw file](https://github.com/gramineproject/gramine/blob/fba5653a8c34c3eef72d19faedd6f8fd48f21653/tests/benchmarks/basic.manifest.template#L1)):*

Yeah, should we bother "fixing" this manifest if it won't even work?
</blockquote></details>

Sure we can. As I've explained numerous times, those were quarterly goals and an exercise in ticking boxes. If we'd like to revive them, we can always check them out from history.

boryspoplawski (changes requested):
> Reviewed 57 of 57 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/194)* status: all files reviewed, 11 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @chiache, @dimakuv, @mkow, and @pwmarcz)

___
*[Pal/src/db_main.c, line 516 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo2tqL38bCgf4rDp1Zr:-Mo4MmCZ2ZyEz3ptJqAg:b32v57) ([raw file](https://github.com/gramineproject/gramine/blob/fba5653a8c34c3eef72d19faedd6f8fd48f21653/Pal/src/db_main.c#L516)):*

"Please switch to 'pal.entrypoint'" would be better. "Consider" is IMO too weak, there's no upside to not switching, and the option will be removed sooner or later.

We already say "This trusted files syntax is deprecated, please move to the new one".
</blockquote></details>

+1
___
*[Pal/src/db_main.c, line 520 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo4Mt5M7HCjdU6mJR0E:-Mo4Mt5M7HCjdU6mJR0F:b-d2c9z0) ([raw file](https://github.com/gramineproject/gramine/blob/fba5653a8c34c3eef72d19faedd6f8fd48f21653/Pal/src/db_main.c#L520)):*
> ```C
> 
>     if (!entrypoint_name)
>         INIT_FAIL(PAL_ERROR_INVAL, "No 'pal.entrypoint'/'loader.preload' is specified in the "
> ```

Please remove `loader.preload` from here, we want users to only specify `pal.entrypoint`.
___
*[Pal/src/db_rtld.c, line 470 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo4OQxb6SuZ5MXs9IKS:-Mo4OQxb6SuZ5MXs9IKT:b-kd5p5d) ([raw file](https://github.com/gramineproject/gramine/blob/fba5653a8c34c3eef72d19faedd6f8fd48f21653/Pal/src/db_rtld.c#L470)):*
> ```C
> out:
>     if (ret < 0) {
>         free((void*)g_entrypoint_map.l_name);
> ```

This is a global variable, please clear it.
___
*[Pal/src/db_rtld.c, line 625 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-Mn_-0TSDOjAY05W_3pr:-Mo4PeN87c1oKoiVHX8Q:b-1kpaj5) ([raw file](https://github.com/gramineproject/gramine/blob/fba5653a8c34c3eef72d19faedd6f8fd48f21653/Pal/src/db_rtld.c#L625)):*
@dimakuv I did not like it and I still do not.
Most of it is useless, it misuses some ABIs, clobbered list is wrong... I could keep going.

Since you removed the preloaded libraries I can restore my old code (well, it was over a year ago, so not sure how useful will that be) which got rid of this and take over from here (clean up entrypoints etc).
But generally speaking Paweł's list seems good.

> BTW, if I'm reading this correctly, musl seems to omit that completely; it's weird because the SysV psABI seems to require it...

Because musl (and we too) controls dynamic loader, which this rdx calling was designed for.
___
*[python/graminelibos/manifest.py, line 189 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-Mn_0Bp-EPWft2IM2NET:-Mo4VTQx2hPBmvIs98hR:bwfe943) ([raw file](https://github.com/gramineproject/gramine/blob/fba5653a8c34c3eef72d19faedd6f8fd48f21653/python/graminelibos/manifest.py#L189)):*

@boryspoplawski Now `loader.preload` is a string, not a list of strings. Is my code good enough? Or is there a more pythonic way?
</blockquote></details>

Looks good (tho you could join the ifs with an `and` (not blocking)
___
*[python/graminelibos/manifest.py, line 210 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-Mn_0Lic0i575RpHXgt4:-Mo4W7b_90IJV4R-Nbbh:bsrfkwe) ([raw file](https://github.com/gramineproject/gramine/blob/fba5653a8c34c3eef72d19faedd6f8fd48f21653/python/graminelibos/manifest.py#L210)):*

@boryspoplawski Now `loader.preload` is a string, not a list of strings. Is my code good enough? Or is there a more pythonic way?
</blockquote></details>

Looks good

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/194)* status: 11 of 63 files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @chiache, and @pwmarcz)

___
*[-- commits, line 3 at r2](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo4qfB81XAwcddzt19T):*
Forgot to change here: `pal.entrypoint` -> `loader.entrypoint`. Will do during final rebase.
___
*[Documentation/manifest-syntax.rst, line 67 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-Mne_nA10-Be7fAD7qP2:-Mo4Vbp28qEWzb9cxM-x:bgo3eov) ([raw file](https://github.com/gramineproject/gramine/blob/fba5653a8c34c3eef72d19faedd6f8fd48f21653/Documentation/manifest-syntax.rst#L67)):*

My understanding is that this is the only `pal.xxx` syntax, right? Why not use `loader.entrypoint` like every other PAL-related syntax?
</blockquote></details>

Done. I actually didn't realize that we use `loader.` for PAL-related (aka "loader"-related) options.
___
*[Documentation/manifest-syntax.rst, line 67 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo2sGL5-fmMTPcF4JL6:-Mo4WqY034x9yidQj5yt:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/fba5653a8c34c3eef72d19faedd6f8fd48f21653/Documentation/manifest-syntax.rst#L67)):*

Missing space before `=`.
</blockquote></details>

Done.
___
*[Documentation/manifest-syntax.rst, line 72 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-MneaZiR8e6FbdLvehKW:-Mo4YSH54yjYQSbE-xss:bzuo63) ([raw file](https://github.com/gramineproject/gramine/blob/fba5653a8c34c3eef72d19faedd6f8fd48f21653/Documentation/manifest-syntax.rst#L72)):*

+1, although I think this can be mentioned under `libos.entrypoint`, since we describe it later.
</blockquote></details>

Done. Hopefully, this text is good enough.

I decided to add the para to `loader.entrypoint` instead of `libos.entrypoint` because it felt more natural.
___
*[Pal/regression/meson.build, line 121 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo2tS3a7STa0vm7rYUX:-Mo4cis77GXVXxbMONJa:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/3db1e28e27622cbf1265650cc92a87281803d31d/Pal/regression/meson.build#L121)):*

Please also remove the `#ifdef SHARED` part from `Pal/regression/crt_init/arch/x86_64/user_start.S`.
</blockquote></details>

Done.
___
*[Pal/src/db_main.c, line 516 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo2tqL38bCgf4rDp1Zr:-Mo4WnfS-aZivHPj0Kn_:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/fba5653a8c34c3eef72d19faedd6f8fd48f21653/Pal/src/db_main.c#L516)):*

"Please switch to 'pal.entrypoint'" would be better. "Consider" is IMO too weak, there's no upside to not switching, and the option will be removed sooner or later.

We already say "This trusted files syntax is deprecated, please move to the new one".
</blockquote></details>

Done.
___
*[Pal/src/db_main.c, line 520 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo4Mt5M7HCjdU6mJR0E:-Mo4bfgW8VGOXEuf8LkD:bkpkhan) ([raw file](https://github.com/gramineproject/gramine/blob/fba5653a8c34c3eef72d19faedd6f8fd48f21653/Pal/src/db_main.c#L520)):*

Please remove `loader.preload` from here, we want users to only specify `pal.entrypoint`.
</blockquote></details>

Done. Also removed from the next message.
___
*[Pal/src/db_rtld.c, line 176 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo2vOCS6Q9j2XBxuNhk:-Mo4dsmg5wskw0GuPO8-:b-w8pdz8) ([raw file](https://github.com/gramineproject/gramine/blob/fba5653a8c34c3eef72d19faedd6f8fd48f21653/Pal/src/db_rtld.c#L176)):*
Done. I allow self-relocations in PAL binary now (and skip the second loop).

> But I'm working on ASan stack sanitization, and found out that in this case LLVM likes to generate some GOT entries anyway, which introduces such relocations.

I would be curious to know why, and if this is strictly necessary. I actually cleaned up PAL code some time ago to get rid of all such stuff.
___
*[Pal/src/db_rtld.c, line 470 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo4OQxb6SuZ5MXs9IKS:-Mo4c-uh5jVVAigKL_mp:bb6un1j) ([raw file](https://github.com/gramineproject/gramine/blob/fba5653a8c34c3eef72d19faedd6f8fd48f21653/Pal/src/db_rtld.c#L470)):*

This is a global variable, please clear it.
</blockquote></details>

Done
___
*[Pal/src/db_rtld.c, line 625 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-Mn_-0TSDOjAY05W_3pr:-Mo4oIWRCjdjswvPJclC:b-3eaqje) ([raw file](https://github.com/gramineproject/gramine/blob/fba5653a8c34c3eef72d19faedd6f8fd48f21653/Pal/src/db_rtld.c#L625)):*

@dimakuv I did not like it and I still do not.
Most of it is useless, it misuses some ABIs, clobbered list is wrong... I could keep going.

Since you removed the preloaded libraries I can restore my old code (well, it was over a year ago, so not sure how useful will that be) which got rid of this and take over from here (clean up entrypoints etc).
But generally speaking Paweł's list seems good.

> BTW, if I'm reading this correctly, musl seems to omit that completely; it's weird because the SysV psABI seems to require it...

Because musl (and we too) controls dynamic loader, which this rdx calling was designed for.
</blockquote></details>

Done.

I wonder why do we bother with additional steps whereas Musl simply sets the new stack and jumps to the entrypoint.
___
*[python/graminelibos/manifest.py, line 189 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-Mn_0Bp-EPWft2IM2NET:-Mo4cLpVA8bZ5jaccXuw:bb6un1j) ([raw file](https://github.com/gramineproject/gramine/blob/fba5653a8c34c3eef72d19faedd6f8fd48f21653/python/graminelibos/manifest.py#L189)):*

Looks good (tho you could join the ifs with an `and` (not blocking)
</blockquote></details>

Done
___
*[tests/benchmarks/basic.manifest.template, line 1 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-Mn_0OgfCJeIHNFMoRVO:-Mo4cObZDMcUuFrBQUHZ:b-pt8t1k) ([raw file](https://github.com/gramineproject/gramine/blob/fba5653a8c34c3eef72d19faedd6f8fd48f21653/tests/benchmarks/basic.manifest.template#L1)):*

Sure we can. As I've explained numerous times, those were quarterly goals and an exercise in ticking boxes. If we'd like to revive them, we can always check them out from history.
</blockquote></details>

I'll submit a separate PR, keeping as-is for now

boryspoplawski (changes requested):
> Reviewed 51 of 52 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/194)* status: all files reviewed, 15 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @chiache, @dimakuv, and @pwmarcz)

___
*[-- commits, line 3 at r2](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo4qfB81XAwcddzt19T):*

Forgot to change here: `pal.entrypoint` -> `loader.entrypoint`. Will do during final rebase.
</blockquote></details>

This applies to all commits.
___
*[Pal/src/db_rtld.c, line 629 at r3](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo5-6Jw8WLuagA3qDkY:-Mo5-6Jw8WLuagA3qDkZ:b-gl5cwa) ([raw file](https://github.com/gramineproject/gramine/blob/9943b5107bccf029543bfe8eaaa65299fb642deb/Pal/src/db_rtld.c#L629)):*
> ```C
> #ifdef __x86_64__
> /*
>  * - RFLAGS is reset (via `push 0, pop`)
> ```

Could you remove this? Or any particular reason for it I'm not aware of?
___
*[Pal/src/db_rtld.c, line 630 at r3](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo50vuS7__Z2pXJvOmR:-Mo50vuS7__Z2pXJvOmS:b-bydgft) ([raw file](https://github.com/gramineproject/gramine/blob/9943b5107bccf029543bfe8eaaa65299fb642deb/Pal/src/db_rtld.c#L630)):*
> ```C
> /*
>  * - RFLAGS is reset (via `push 0, pop`)
>  * - RDX is supposed to point to the `atexit` callback, but our minimal implementation of the
> ```

We could get rid of this. This is a LibOS entrypoint, so we can define our own ABI, which can be as simple as just calling a predefined function/entrypoint without any more guarantees (i.e. just normal function calling stuff).
___
*[Pal/src/db_rtld.c, line 634 at r3](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo526OuD30Z5YnkSXEn:-Mo526OuD30Z5YnkSXEo:b-ulowat) ([raw file](https://github.com/gramineproject/gramine/blob/9943b5107bccf029543bfe8eaaa65299fb642deb/Pal/src/db_rtld.c#L634)):*
> ```C
>  * - RBP is set to zero (to indicate end of backtrace to GDB)
>  */
> #define CALL_ENTRY(l, cookies)                                                       \
> ```

Any reason for this being a macro? It's used in exactly one place now.
___
*[Pal/src/db_rtld.c, line 637 at r3](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo505Lo-Kc4VsSS7Wg0:-Mo4zzU0B8KPhXvaZD1U:bdhqiiw) ([raw file](https://github.com/gramineproject/gramine/blob/9943b5107bccf029543bfe8eaaa65299fb642deb/Pal/src/db_rtld.c#L637)):*
> ```C
>     ({                                                                               \
>         __asm__ volatile(                                                            \
>             "pushq $0\r\n"                                                           \
> ```

Please remove `\r` from each line
___
*[Pal/src/db_rtld.c, line 643 at r3](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo51N3N7jWmR3jbVB2e:-Mo51N3N7jWmR3jbVB2f:b-nwm3bn) ([raw file](https://github.com/gramineproject/gramine/blob/9943b5107bccf029543bfe8eaaa65299fb642deb/Pal/src/db_rtld.c#L643)):*
> ```C
>             "movq %1, %%rsp\r\n"                                                     \
>             "jmp *%0\r\n"                                                            \
>             : : "a"((l)->l_entry), "b"(cookies) : "memory" );                        \
> ```

Please split it to 3 lines (more readable).
___
*[Pal/src/db_rtld.c, line 643 at r3](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo51RO2-EjHTnSu6Y98:-Mo51RO38yX84x-7JerZ:b-3edxxd) ([raw file](https://github.com/gramineproject/gramine/blob/9943b5107bccf029543bfe8eaaa65299fb642deb/Pal/src/db_rtld.c#L643)):*
> ```C
>             "movq %1, %%rsp\r\n"                                                     \
>             "jmp *%0\r\n"                                                            \
>             : : "a"((l)->l_entry), "b"(cookies) : "memory" );                        \
> ```

Please either user input registers directly or allow for any of them (now we have `"a"` but it's used as`%0`).
___
*[Pal/src/db_rtld.c, line 643 at r3](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo51klj5d4qztJjMrUC:-Mo51klk20CGT-JHpWzV:b4tk8al) ([raw file](https://github.com/gramineproject/gramine/blob/9943b5107bccf029543bfe8eaaa65299fb642deb/Pal/src/db_rtld.c#L643)):*
> ```C
>             "movq %1, %%rsp\r\n"                                                     \
>             "jmp *%0\r\n"                                                            \
>             : : "a"((l)->l_entry), "b"(cookies) : "memory" );                        \
> ```

Please change the first argument to just be the entry point, no need to pass whole struct.
___
*[Pal/src/db_rtld.c, line 643 at r3](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo51t7K6Sw26YV1kK5e:-Mo51t7K6Sw26YV1kK5f:bad3go8) ([raw file](https://github.com/gramineproject/gramine/blob/9943b5107bccf029543bfe8eaaa65299fb642deb/Pal/src/db_rtld.c#L643)):*
> ```C
>             "movq %1, %%rsp\r\n"                                                     \
>             "jmp *%0\r\n"                                                            \
>             : : "a"((l)->l_entry), "b"(cookies) : "memory" );                        \
> ```

Please add `__builtin_unreachable()` after this asm.
___
*[Pal/src/db_rtld.c, line 689 at r3](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo520CP2fFk2uGjuGs6:-Mo520CP2fFk2uGjuGs7:bnijcg2) ([raw file](https://github.com/gramineproject/gramine/blob/9943b5107bccf029543bfe8eaaa65299fb642deb/Pal/src/db_rtld.c#L689)):*
> ```C
>     assert(g_entrypoint_map.l_entry);
>     CALL_ENTRY(&g_entrypoint_map, cookies);
>     die_or_inf_loop();
> ```

This can be removed after adding `__buildin_unreachable()`.

pwmarcz (changes requested):
> Reviewed 51 of 52 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/194)* status: all files reviewed, 12 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @chiache and @dimakuv)

___
*[Documentation/manifest-syntax.rst, line 76 at r3](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo83lr4-PpWY6hUZecX:-Mo83lr4-PpWY6hUZecY:b-54bn42) ([raw file](https://github.com/gramineproject/gramine/blob/9943b5107bccf029543bfe8eaaa65299fb642deb/Documentation/manifest-syntax.rst#L76)):*
> ```reStructuredText
> ``loader.entrypoint`` and passes control to this binary. Next, the LibOS binary
> loads the actual executable (the user application) specified in
> ``libos.entrypoint``. Also note that, in constrast to ``libos.entrypoint``, the
> ```

Typo: `constrast`
___
*[Pal/src/db_rtld.c, line 176 at r1](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo2vOCS6Q9j2XBxuNhk:-Mo84Jn9C6mvHg04CvL2:bifb78d) ([raw file](https://github.com/gramineproject/gramine/blob/fba5653a8c34c3eef72d19faedd6f8fd48f21653/Pal/src/db_rtld.c#L176)):*

Done. I allow self-relocations in PAL binary now (and skip the second loop).

> But I'm working on ASan stack sanitization, and found out that in this case LLVM likes to generate some GOT entries anyway, which introduces such relocations.

I would be curious to know why, and if this is strictly necessary. I actually cleaned up PAL code some time ago to get rid of all such stuff.
</blockquote></details>

I'm not sure exactly why that happens. So far I found out that it's in the instrumented code added by AddressSanitizer, and that it makes a few functions (the ones with most interesting patterns of stack allocation, I guess?) store their own address in a local variable. I see that in the generated LLVM IR, but I don't know yet why the address translates to assembly as GOT, not an IP-relative calculation.

Anyway, I'll try to figure out what happens, and get back to you when I finish my PR.
___
*[Pal/src/db_rtld.c, line 632 at r3](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo87X0n0-bdFjemqcLv:-Mo87X0o1cBC0JSPlzHq:b-a2tzuk) ([raw file](https://github.com/gramineproject/gramine/blob/9943b5107bccf029543bfe8eaaa65299fb642deb/Pal/src/db_rtld.c#L632)):*
> ```C
>  * - RDX is supposed to point to the `atexit` callback, but our minimal implementation of the
>  *   dynamic loader doesn't need to do anything for process termination, so we just pass NULL
>  * - RBP is set to zero (to indicate end of backtrace to GDB)
> ```

Why here? Usually that's done by the start code. (Not blocking)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/194)* status: 60 of 82 files reviewed, 12 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @chiache, @dimakuv, and @pwmarcz)

___
*[Documentation/manifest-syntax.rst, line 76 at r3](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo83lr4-PpWY6hUZecX:-Mo8B-7z37ZPmw6JBRWW:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9943b5107bccf029543bfe8eaaa65299fb642deb/Documentation/manifest-syntax.rst#L76)):*

Typo: `constrast`
</blockquote></details>

Done.
___
*[LibOS/shim/src/shim_init.c, line 403 at r4](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo8DLqh7d1AE2sT1wUB:-Mo8DLqh7d1AE2sT1wUC:b-bxhap4) ([raw file](https://github.com/gramineproject/gramine/blob/3db1e28e27622cbf1265650cc92a87281803d31d/LibOS/shim/src/shim_init.c#L403)):*
> ```C
> 
>     const char** argv = args;
>     const char** envp = args + sizeof(char*) * ((argc) + 1);
> ```

FYI: Unrelated to anything, just noticed this in the code and decided to fix.
___
*[Pal/regression/..Bootstrap.c, line 4 at r4](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo8DSXt3eR4EiiS2YRu:-Mo8DSXt3eR4EiiS2YRv:blyzhvm) ([raw file](https://github.com/gramineproject/gramine/blob/3db1e28e27622cbf1265650cc92a87281803d31d/Pal/regression/..Bootstrap.c#L4)):*
> ```C
> #include "pal_regression.h"
> 
> int main(int argc, char** argv, char** envp) {
> ```

FYI: I decided to remove this 3rd argument from our PAL-to-LibOS ABI.
___
*[Pal/regression/crt_init/arch/x86_64/user_start.S, line 22 at r4](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo8DcCk4whBq5pWVMXz:-Mo8DcCk4whBq5pWVMY-:bvgbt6b) ([raw file](https://github.com/gramineproject/gramine/blob/364ab3976624a9f4beee281eee10b1977eef3eef/Pal/regression/crt_init/arch/x86_64/user_start.S#L22)):*
> ```
> 	.type _start,@function
> _start:
>     .cfi_startproc
> ```

FYI: I made this PAL-regression-tests entrypoint exactly the same as our LibOS entrypoint (see `start.S`)
___
*[Pal/src/db_rtld.c, line 629 at r3](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo5-6Jw8WLuagA3qDkY:-Mo7wUO_A5yM_5eBRyC4:bqmwi3) ([raw file](https://github.com/gramineproject/gramine/blob/9943b5107bccf029543bfe8eaaa65299fb642deb/Pal/src/db_rtld.c#L629)):*

Could you remove this? Or any particular reason for it I'm not aware of?
</blockquote></details>

Done. No reason. I actually looked at the entrypoint of LibOS ()
___
*[Pal/src/db_rtld.c, line 630 at r3](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo50vuS7__Z2pXJvOmR:-Mo7wTNpA556jeEJ9z0F:b-20ndbp) ([raw file](https://github.com/gramineproject/gramine/blob/9943b5107bccf029543bfe8eaaa65299fb642deb/Pal/src/db_rtld.c#L630)):*

We could get rid of this. This is a LibOS entrypoint, so we can define our own ABI, which can be as simple as just calling a predefined function/entrypoint without any more guarantees (i.e. just normal function calling stuff).
</blockquote></details>

Done.

As Borys mentioned, we can define our own ABI, so I did this: we only care about the stack layout and the `rsp` register.
___
*[Pal/src/db_rtld.c, line 632 at r3](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo87X0n0-bdFjemqcLv:-Mo8Aqew4uwdEZaw7u7A:b-nrpqjc) ([raw file](https://github.com/gramineproject/gramine/blob/9943b5107bccf029543bfe8eaaa65299fb642deb/Pal/src/db_rtld.c#L632)):*

Why here? Usually that's done by the start code. (Not blocking)
</blockquote></details>

Done. Exactly, there was no need in this, see our `start.S` snippets.
___
*[Pal/src/db_rtld.c, line 634 at r3](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo526OuD30Z5YnkSXEn:-Mo7wRub9R67yNqFxIfq:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9943b5107bccf029543bfe8eaaa65299fb642deb/Pal/src/db_rtld.c#L634)):*

Any reason for this being a macro? It's used in exactly one place now.
</blockquote></details>

Done.
___
*[Pal/src/db_rtld.c, line 637 at r3](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo505Lo-Kc4VsSS7Wg0:-Mo8AsI13w-Rx-VHmliD:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9943b5107bccf029543bfe8eaaa65299fb642deb/Pal/src/db_rtld.c#L637)):*

Please remove `\r` from each line
</blockquote></details>

Done.
___
*[Pal/src/db_rtld.c, line 643 at r3](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo51t7K6Sw26YV1kK5e:-Mo7uwL7BUfc6oT7BekT:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9943b5107bccf029543bfe8eaaa65299fb642deb/Pal/src/db_rtld.c#L643)):*

Please add `__builtin_unreachable()` after this asm.
</blockquote></details>

Done.
___
*[Pal/src/db_rtld.c, line 643 at r3](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo51N3N7jWmR3jbVB2e:-Mo8AskWC_nmCAINQjGY:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9943b5107bccf029543bfe8eaaa65299fb642deb/Pal/src/db_rtld.c#L643)):*

Please split it to 3 lines (more readable).
</blockquote></details>

Done.
___
*[Pal/src/db_rtld.c, line 643 at r3](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo51RO2-EjHTnSu6Y98:-Mo8At6FC9CktLq-QkFR:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9943b5107bccf029543bfe8eaaa65299fb642deb/Pal/src/db_rtld.c#L643)):*

Please either user input registers directly or allow for any of them (now we have `"a"` but it's used as`%0`).
</blockquote></details>

Done.
___
*[Pal/src/db_rtld.c, line 643 at r3](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo51klj5d4qztJjMrUC:-Mo8AtkfDOdZ-PxroxI3:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9943b5107bccf029543bfe8eaaa65299fb642deb/Pal/src/db_rtld.c#L643)):*

Please change the first argument to just be the entry point, no need to pass whole struct.
</blockquote></details>

Done.
___
*[Pal/src/db_rtld.c, line 655 at r3](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo8DnNo43Fw0xOm1LRZ:-Mo8DnNo43Fw0xOm1LR_:bqztauy) ([raw file](https://github.com/gramineproject/gramine/blob/9943b5107bccf029543bfe8eaaa65299fb642deb/Pal/src/db_rtld.c#L655)):*
> ```C
> 
>     /* Let's count the number of cookies, first we will have argc & argv */
>     int ncookies = narguments + 3; /* 1 for argc, argc + 2 for argv */
> ```

All this "cookies" stuff was painful to look at, so I rewrote this function. Hopefully more readable now.
___
*[Pal/src/db_rtld.c, line 665 at r3](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo8DtMJ2SV-SN2EHvLV:-Mo8DtMJ2SV-SN2EHvLW:b-7dqeis) ([raw file](https://github.com/gramineproject/gramine/blob/9943b5107bccf029543bfe8eaaa65299fb642deb/Pal/src/db_rtld.c#L665)):*
> ```C
>     int cookiesz = sizeof(unsigned long int) * ncookies
>                       + sizeof(ElfW(auxv_t)) * 1  /* only AT_NULL */
>                       + sizeof(void*) * 4 + 16;
> ```

FYI: I have no idea what this `4 + 16` was supposed to mean. I removed it. Git blame led me to you-know-which-commit.
___
*[Pal/src/db_rtld.c, line 689 at r3](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo520CP2fFk2uGjuGs6:-Mo7uvsF0s-n1piR6x-c:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9943b5107bccf029543bfe8eaaa65299fb642deb/Pal/src/db_rtld.c#L689)):*

This can be removed after adding `__buildin_unreachable()`.
</blockquote></details>

Done.
___
*[Pal/src/db_rtld.c, line 679 at r4](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo8E3Rh2Q4iRLNcF7Gm:-Mo8E3Rh2Q4iRLNcF7Gn:bj856ob) ([raw file](https://github.com/gramineproject/gramine/blob/364ab3976624a9f4beee281eee10b1977eef3eef/Pal/src/db_rtld.c#L679)):*
> ```C
>                      :
>                      : "r"(g_entrypoint_map.l_entry), "r"(stack_entries)
>                      : "memory");
> ```

FYI: I'm actually not sure why `memory` clobber is here, but I followed the Musl code (pasted by Pawel in another comment thread).

pwmarcz (changes requested):
> Reviewed 22 of 22 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/194)* status: all files reviewed, 13 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @chiache, and @dimakuv)

___
*[Pal/regression/..Bootstrap.c, line 4 at r4](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo8DSXt3eR4EiiS2YRu:-Mo8FbniDfW4tC7KKfCC:b-f9v9gl) ([raw file](https://github.com/gramineproject/gramine/blob/3db1e28e27622cbf1265650cc92a87281803d31d/Pal/regression/..Bootstrap.c#L4)):*

FYI: I decided to remove this 3rd argument from our PAL-to-LibOS ABI.
</blockquote></details>

You mean, the Pal regression `user_start.S` no longer passes the `envp` parameter to `main`?

I think this is surprising (the function is called `main()`, after all), so I would at least mention it in `user_start.S`; i.e. add `NOTE:` saying that our `main()` differs from the Unix `main()` in that the `envp` parameter is not passed.

Alternatively, maybe it's better to pass `envp` in both of these start codes? After all, `shim_init` still ends up needing to calculate it.
___
*[Pal/regression/..Bootstrap.c, line 4 at r4](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo8hpzY1PxttP86IEaM:-Mo8hpzY1PxttP86IEaN:b-xwa254) ([raw file](https://github.com/gramineproject/gramine/blob/364ab3976624a9f4beee281eee10b1977eef3eef/Pal/regression/..Bootstrap.c#L4)):*
> ```C
> #include "pal_regression.h"
> 
> int main(int argc, char** argv) {
> ```

If you're touching all these `main` signatures, maybe you want to fix the `-Wno-unused-parameter` in this directory? It's mostly there because in many cases we have `int main(argc, argv[, envp])` instead of `int main(void)`.

No pressure though, this is already a big PR.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/194)* status: 62 of 83 files reviewed, 13 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @chiache, @dimakuv, and @pwmarcz)

___
*[Pal/regression/..Bootstrap.c, line 4 at r4](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo8DSXt3eR4EiiS2YRu:-Mo943NC0tqIQT6_OGsp:buvdeqo) ([raw file](https://github.com/gramineproject/gramine/blob/3db1e28e27622cbf1265650cc92a87281803d31d/Pal/regression/..Bootstrap.c#L4)):*

You mean, the Pal regression `user_start.S` no longer passes the `envp` parameter to `main`?

I think this is surprising (the function is called `main()`, after all), so I would at least mention it in `user_start.S`; i.e. add `NOTE:` saying that our `main()` differs from the Unix `main()` in that the `envp` parameter is not passed.

Alternatively, maybe it's better to pass `envp` in both of these start codes? After all, `shim_init` still ends up needing to calculate it.
</blockquote></details>

Done. Yes, it actually looks better.
___
*[Pal/regression/..Bootstrap.c, line 4 at r4](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo8hpzY1PxttP86IEaM:-Mo945Ck7pw1XJITOEtZ:b-6pfm09) ([raw file](https://github.com/gramineproject/gramine/blob/364ab3976624a9f4beee281eee10b1977eef3eef/Pal/regression/..Bootstrap.c#L4)):*

If you're touching all these `main` signatures, maybe you want to fix the `-Wno-unused-parameter` in this directory? It's mostly there because in many cases we have `int main(argc, argv[, envp])` instead of `int main(void)`.

No pressure though, this is already a big PR.
</blockquote></details>

I reverted my changes to PAL tests, so not applicable (maybe we'll do it in some future PR).

pwmarcz (changes requested):
> Reviewed 21 of 21 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/194)* status: all files reviewed, 13 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @chiache, and @dimakuv)

___
*[LibOS/shim/src/arch/x86_64/start.S, line 31 at r5](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo9HE_94Y39D1fny8EE:-Mo9HE_A2s2BbZfl6ZSM:bo0m7u) ([raw file](https://github.com/gramineproject/gramine/blob/7f70d873e70a24088d8c30c5d28aba624479c082/LibOS/shim/src/arch/x86_64/start.S#L31)):*
> ```
>     # Arguments for shim_init:
>     movq 0(%rsp), %rdi         # argc
>     leaq 8(%rsp), %rsi         # args
> ```

`argv`?
___
*[Pal/regression/crt_init/arch/x86_64/user_start.S, line 30 at r5](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo9HNQD48wvJhg3ekjz:-Mo9HNQD48wvJhg3ekk-:bo0m7u) ([raw file](https://github.com/gramineproject/gramine/blob/7f70d873e70a24088d8c30c5d28aba624479c082/Pal/regression/crt_init/arch/x86_64/user_start.S#L30)):*
> ```
>     # Arguments for PAL-executable main:
>     movq 0(%rsp), %rdi         # argc
>     leaq 8(%rsp), %rsi         # args
> ```

`argv`?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/194)* status: 81 of 83 files reviewed, 13 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @chiache, @dimakuv, and @pwmarcz)

___
*[LibOS/shim/src/arch/x86_64/start.S, line 31 at r5](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo9HE_94Y39D1fny8EE:-Mo9eQO18jZnFAN9hreZ:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/7f70d873e70a24088d8c30c5d28aba624479c082/LibOS/shim/src/arch/x86_64/start.S#L31)):*

`argv`?
</blockquote></details>

Done.
___
*[Pal/regression/crt_init/arch/x86_64/user_start.S, line 30 at r5](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo9HNQD48wvJhg3ekjz:-Mo9eWDjDGvUgq7cduHF:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/7f70d873e70a24088d8c30c5d28aba624479c082/Pal/regression/crt_init/arch/x86_64/user_start.S#L30)):*

`argv`?
</blockquote></details>

Done.

pwmarcz (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/194)* status: all files reviewed, 11 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @chiache, and @dimakuv)

dimakuv:
Jenkins, retest Jenkins-SGX-20.04-apps please (known bug, will be fixed by #214)

pwmarcz (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/194)* status: all files reviewed, 8 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @chiache and @dimakuv)

___
*[Pal/regression/..Bootstrap.c, line 4 at r4](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo8DSXt3eR4EiiS2YRu:-Mo9tinHBSsXjF_XjG9X:b-vw100k) ([raw file](https://github.com/gramineproject/gramine/blob/3db1e28e27622cbf1265650cc92a87281803d31d/Pal/regression/..Bootstrap.c#L4)):*

Passing `envp` to main is actually an extension, C standard does not define it.
</blockquote></details>

I thought it's defined by POSIX, but apparently that's not true either.

Still, I'd say keeping it here is better than retrieving it manually from after `argv` (like we did in LibOS before).

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/194)* status: 80 of 83 files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @chiache, @dimakuv, and @pwmarcz)

___
*[LibOS/shim/src/arch/x86_64/start.S, line 8 at r6](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo9i_cH0_W5bX5_I1ZB:-MoA-T4q1QZn9zo3Wrlh:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/98a5366313212303d30b157bcc0af95f66fc1af3/LibOS/shim/src/arch/x86_64/start.S#L8)):*

This is wrong, should be `(8*(argc+1))`. Ditto for all below.
</blockquote></details>

Done.
___
*[LibOS/shim/src/arch/x86_64/start.S, line 33 at r6](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo9j7Mj87l2XdE9ATow:-MoA-_GY4LRd-pvwKNFS:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/98a5366313212303d30b157bcc0af95f66fc1af3/LibOS/shim/src/arch/x86_64/start.S#L33)):*

You can just write `leaq 8(%rsi,%rdi,8), %rdx`.
</blockquote></details>

Done.
___
*[Pal/regression/crt_init/arch/x86_64/user_start.S, line 8 at r6](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo9jz_hDqCX1nK0VCxh:-MoA-lGeCSxDqC9T85ma:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/98a5366313212303d30b157bcc0af95f66fc1af3/Pal/regression/crt_init/arch/x86_64/user_start.S#L8)):*

ditto (off by one)
</blockquote></details>

Done.
___
*[Pal/regression/crt_init/arch/x86_64/user_start.S, line 32 at r6](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo9k3fV3nWk8-jhQEAx:-MoA-mfoCaJPz7-Wg1S6:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/98a5366313212303d30b157bcc0af95f66fc1af3/Pal/regression/crt_init/arch/x86_64/user_start.S#L32)):*

ditto (shorted form)
</blockquote></details>

Done.
___
*[Pal/src/db_rtld.c, line 634 at r6](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo9kMWm0v7vfRP396tr:-MoA-x1L9rLzCEHFS7KN:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/98a5366313212303d30b157bcc0af95f66fc1af3/Pal/src/db_rtld.c#L634)):*

off by one (`argc+1`)
</blockquote></details>

Done.
___
*[Pal/src/db_rtld.c, line 652 at r6](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo9ku9LEgaECDPDO6Gq:-MoA0J109_k9LUpHBNP4:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/98a5366313212303d30b157bcc0af95f66fc1af3/Pal/src/db_rtld.c#L652)):*

Why `uinptr_t` instead of `void*`?
</blockquote></details>

Done.
___
*[Pal/regression/..Bootstrap.c, line 4 at r4](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo8DSXt3eR4EiiS2YRu:-MoA0KxJADrg7map42IH:b-z3awr2) ([raw file](https://github.com/gramineproject/gramine/blob/3db1e28e27622cbf1265650cc92a87281803d31d/Pal/regression/..Bootstrap.c#L4)):*

I thought it's defined by POSIX, but apparently that's not true either.

Still, I'd say keeping it here is better than retrieving it manually from after `argv` (like we did in LibOS before).
</blockquote></details>

Yes, let's keep `envp` as the third argument.

boryspoplawski (commented):
> Reviewed 1 of 21 files at r5, 3 of 3 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/194)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @chiache and @dimakuv)

chiache (approved):
> Reviewed 1 of 52 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/194)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

pwmarcz (approved):
> Reviewed 3 of 3 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/194)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/194)* status: 79 of 83 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski, @chiache, and @pwmarcz)

___
*[-- commits, line 3 at r2](https://reviewable.io/reviews/gramineproject/gramine/194#-Mo4qfB81XAwcddzt19T):*

This applies to all commits.
</blockquote></details>

Done

pwmarcz (approved):
> Reviewed 4 of 4 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/194)* status: all files reviewed, 1 unresolved discussion (waiting on @boryspoplawski)

boryspoplawski (approved):
> Reviewed 2 of 4 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/194)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #197, opened 2021-11-05
[CI] Test apps in the SGX sanitizers configuration

pwmarcz:
## Description of the changes 

This adds `CI-Examples` to the SGX sanitizers configuration. ASan and UBSan found no bugs, but the remote attestation examples did not work when compiled with Clang (because it doesn't default to PIE).

Also, I fixed a stupid bug that causes Gramine to continue after failing to map such executables (instead of exiting with a helpful message, which we already have!)

## How to test this PR? 

CI should be enough.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/197)

### DIFF
--- a/.ci/linux-sgx-sanitizers.jenkinsfile
+++ b/.ci/linux-sgx-sanitizers.jenkinsfile
@@ -19,6 +19,7 @@ node('sgx_slave_2.6') {
         load '.ci/lib/stage-clean-check-prepare.jenkinsfile'
         load '.ci/lib/stage-build-sgx.jenkinsfile'
         load '.ci/lib/stage-test.jenkinsfile'
+        load '.ci/lib/stage-test-sgx.jenkinsfile'
         load '.ci/lib/stage-clean-check.jenkinsfile'
     }
 }
--- a/CI-Examples/ra-tls-mbedtls/Makefile
+++ b/CI-Examples/ra-tls-mbedtls/Makefile
@@ -17,6 +17,9 @@ GRAMINE_LOG_LEVEL = error
 CFLAGS += -O2
 endif
 
+CFLAGS += -fPIE
+LDFLAGS += -pie
+
 .PHONY: all
 all: app epid  # by default, only build EPID because it doesn't rely on additional (DCAP) libs
 
@@ -67,13 +70,13 @@ mbedtls/.mbedtls_configured: mbedtls/.mbedtls_downloaded
 ######################### CLIENT/SERVER EXECUTABLES ###########################
 
 CFLAGS += -I./mbedtls/include $(shell pkg-config --cflags mbedtls_gramine)
-LFLAGS += -ldl $(shell pkg-config --libs mbedtls_gramine)
+LDFLAGS += -ldl $(shell pkg-config --libs mbedtls_gramine)
 
 server: src/server.c mbedtls/.mbedtls_configured
-	$(CC) $< $(CFLAGS) $(LFLAGS) -o $@
+	$(CC) $< $(CFLAGS) $(LDFLAGS) -o $@
 
 client: src/client.c mbedtls/.mbedtls_configured
-	$(CC) $< $(CFLAGS) $(LFLAGS) -o $@
+	$(CC) $< $(CFLAGS) $(LDFLAGS) -o $@
 
 ############################### SERVER MANIFEST ###############################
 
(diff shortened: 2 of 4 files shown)


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/197)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

___
*[-- commits, line 6 at r1](https://reviewable.io/reviews/gramineproject/gramine/197#-Mo2gGB0CTO7X8opGqBb):*
So ASan and UBSan only work on PIE executables? I can't believe this. What was the exact problem? Or is it not about ASan/UBSan, but the what? Gramine-SGX is able to load non-PIE executables into SGX enclaves, so what was wrong?
___
*[LibOS/shim/src/shim_rtld.c, line 252 at r1](https://reviewable.io/reviews/gramineproject/gramine/197#-Mo2gk7CBl0F9HRtOnAF:-Mo2gk7D06xaARxzWYX7:baf11rm) ([raw file](https://github.com/gramineproject/gramine/blob/c1eef7ac117ff4827c4ec619e711c3f11cc1316c/LibOS/shim/src/shim_rtld.c#L252)):*
> ```C
> 
>         if ((ret = file->fs->fs_ops->mmap(file, &map_start, map_size, c->prot, map_flags,
>                                           c->map_off)) < 0) {
> ```

Oops :) I thought GCC/Clang warnings would catch this blooper.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/197)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

___
*[-- commits, line 6 at r1](https://reviewable.io/reviews/gramineproject/gramine/197#-Mo2gGB0CTO7X8opGqBb):*

So ASan and UBSan only work on PIE executables? I can't believe this. What was the exact problem? Or is it not about ASan/UBSan, but the what? Gramine-SGX is able to load non-PIE executables into SGX enclaves, so what was wrong?
</blockquote></details>

No, this is not about ASan/UBSan. Gramine-SGX is able to load a non-PIE executable, but only with `sgx.nonpie_executable = true` manifest option (otherwise, the enclave usually starts at too high address). I thought that forcing PIE would be a better solution than adding a manifest option.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/197)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

___
*[-- commits, line 6 at r1](https://reviewable.io/reviews/gramineproject/gramine/197#-Mo2gGB0CTO7X8opGqBb):*

No, this is not about ASan/UBSan. Gramine-SGX is able to load a non-PIE executable, but only with `sgx.nonpie_executable = true` manifest option (otherwise, the enclave usually starts at too high address). I thought that forcing PIE would be a better solution than adding a manifest option.
</blockquote></details>

Ah, I understand now.

1. Building with Clang results in non-PIE executable (by default).
2. A non-PIE executable requires `sgx.nonpie_executable = true` in the manifest.
3. We only hit this problem now because this is the first time we try to build CI-Examples with Clang. Not relevant to sanitizers at all.

boryspoplawski (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/197)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[LibOS/shim/src/shim_rtld.c, line 252 at r1](https://reviewable.io/reviews/gramineproject/gramine/197#-Mo2gk7CBl0F9HRtOnAF:-Mo4oNF46ah64KiTzta-:b-tqg2tj) ([raw file](https://github.com/gramineproject/gramine/blob/c1eef7ac117ff4827c4ec619e711c3f11cc1316c/LibOS/shim/src/shim_rtld.c#L252)):*

Oops :) I thought GCC/Clang warnings would catch this blooper.
</blockquote></details>

That's why I dislike assignments inside IF.

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/197)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #198, opened 2021-11-05
[Docs] Fix minor issues with documentation

### LINKED ISSUE #202: In Gramine build instructions provide instruction to determine kernel version
## Description of the problem
Help new users to more easily set up their Linux SGX environment.
Following build instructions on https://gramine.readthedocs.io/en/latest/building.html
Item 2 might be easier on the reader if titled
title: SGX minimal Linux kernel version"
 text: "SGX requires minimally Linux Kernel version 5.9. To determine your Linux kernel version run 
> uname -r

The kernel requirement is because SGX requires FSGSBASE,  a feature in recent processors which allows direct access to the FS and GS segment base addresses. See link(<to learn more about FSGSBASE>).
"
The remainder for upgrade steps is perfect.
## Steps to reproduce
An Ubuntu environment - bare metal or VM
Goal is to get SGX environment set up for new comers.

### PLEASE ENSURE THAT THE ISSUE REPRODUCES ON THE CURRENT MASTER BRANCH ###

-->

## Expected results
Help the user determine whether they need to apply Linux kernel patches, it they are at a version prior to 5.9. They should not have to read a separate document.

## Actual results
See kernel version with a quick command run
> uname -r
> 5.4.0-89-generic

### LINKED ISSUE #203: Doc: provide actual output example from FSGSBASE test
## Description of the problem
To help new comers use SGX, the easier the instructions are to follow the better.
On https://gramine.readthedocs.io/en/latest/building.html#fsgsbase the user is asked to verify that bit 2 is set. What is the corresponding output that they must expect? An example helps/
## Steps to reproduce
documentation issue

## Expected results
Please share how the output should look.

## Actual results
mkbhanda@sse-tme-wf-01:~$ LD_SHOW_AUXV=1 /bin/true | grep AT_HWCAP2
AT_HWCAP2:            0x0

### LINKED ISSUE #204: Doc: When will the command is_sgx_available become available?
## Description of the problem
in the quick start guide at https://graphene.readthedocs.io/en/latest/quickstart.html
is_sgx_available is alluded to test if SGX is available. Thyat appears to be a command that will become available only after a successful build.
Perhaps it is an after-the-setup test to confirm SGX indeed is available. If so, wording needs to change.
## Expected results
N/A
## Actual results
N/A

vijaydhanraj:
## Description of the changes 
This commit adds steps to build the DCAP version of the RA-TLS library as part of `building` documentation. Also fixes minor naming issues with SGX tools.

Fixes #202
Fixes #203 
Fixes #204 

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)

### DIFF
--- a/Documentation/building.rst
+++ b/Documentation/building.rst
@@ -81,14 +81,17 @@ Run the following commands on Ubuntu to install SGX-related dependencies::
     sudo apt-get install -y libcurl4-openssl-dev libprotobuf-c-dev \
         protobuf-c-compiler python3-pip python3-protobuf
 
-2. Upgrade to the Linux kernel patched with FSGSBASE
-""""""""""""""""""""""""""""""""""""""""""""""""""""
+2. Install Linux kernel with patched FSGSBASE
+"""""""""""""""""""""""""""""""""""""""""""""
 
 FSGSBASE is a feature in recent processors which allows direct access to the FS
 and GS segment base addresses. For more information about FSGSBASE and its
 benefits, see `this discussion <https://lwn.net/Articles/821719>`__. Note that
 if your kernel version is 5.9 or higher, then the FSGSBASE feature is already
-supported and you can skip this step.
+supported and you can skip this step. Kernel version can be checked using the
+following command::
+
+       uname -r
 
 If your current kernel version is lower than 5.9, then you have two options:
 
@@ -107,7 +110,7 @@ version is 5.11 or higher, then the Intel SGX driver is already installed and
 you can skip this step.
 
 If you have an older CPU without :term:`FLC` support, you need to download and
-install the the following Intel SGX driver:
+install the the following out-of-tree (OOT) Intel SGX driver:
 
 - https://github.com/intel/linux-sgx-driver
 
@@ -187,6 +190,14 @@ Additional build options
 - To build test binaries, run :command:`meson -Dtests=enabled`. This is
   necessary if you will be running regression tests. See
   :doc:`devel/contributing` for details.
+
+- In order to run SGX tools with DCAP version of RA-TLS library
+  (``ra_tls_verify_dcap.so``), build with :command:`meson -Ddcap=enabled` option.
+  See `RA-TLS example's README <https://github.com/gramineproject/gramine/blob/master/CI-Examples/ra-tls-mbedtls/README.md>`__.
+
+  .. note::
+     EPID version of RA-TLS library (``ra_tls_verify_epid.so``) is built by
+     default.
 
 - To create a debug build, run :command:`meson --buildtype=debug`. This adds
   debug symbols in all Gramine components, builds them without optimizations,
@@ -304,9 +315,15 @@ instructions ensure that the resulting kernel has FSGSBASE support.
        uname -r
 
 #. Also verify that the patched kernel supports FSGSBASE (the below command
-   must return that bit 2 is set)::
+   must return that bit 1 is set)::
 
-       LD_SHOW_AUXV=1 /bin/true | grep AT_HWCAP2
+       # Linux kernel doesn't support FSGSBASE: patch or use higher version!
+       $ LD_SHOW_AUXV=1 /bin/true | grep AT_HWCAP2
+       AT_HWCAP2:       0x0
+
+       # Linux kernel supports FSGSBASE (example where only bit 1 is set)
+       $ LD_SHOW_AUXV=1 /bin/true | grep AT_HWCAP2
+       AT_HWCAP2:       0x2
 
 After the patched Linux kernel is installed, you may proceed with installations
 of other SGX software infrastructure: the Intel SGX Linux driver, the Intel SGX
--- a/Documentation/quickstart.rst
+++ b/Documentation/quickstart.rst
@@ -49,8 +49,11 @@ Gramine requires several features from your system:
 If your system doesn't meet these requirements, please refer to more detailed
 descriptions in :doc:`building`.
 
-#. Ensure that Intel SGX is enabled on your platform using
-   :program:`is_sgx_available`.
+#. Ensure that Intel SGX is enabled on your platform by doing::
+
+       sudo apt-get install -y cpuid
+       cpuid | grep -i sgx
+       # output must contain lines like "SGX: Software Guard Extensions supported = true"
 
 #. Clone the Gramine repository::
 
--- a/Pal/src/host/Linux-SGX/tools/README.rst
+++ b/Pal/src/host/Linux-SGX/tools/README.rst
@@ -3,12 +3,15 @@ SGX tools
 
 .. TODO move this to Documentation/manpages
 
+All the below tools can be simply run from any directory as they are installed
+globally on the system.
+
 SGX availability checker
 ------------------------
 
 Example output::
 
-    > ./is_sgx_available/is_sgx_available
+    > is-sgx-available
     SGX supported by CPU: true
     SGX1 (ECREATE, EENTER, ...): true
     SGX2 (EAUG, EACCEPT, EMODPR, ...): false
@@ -22,9 +25,10 @@ Example output::
     AESMD installed: true
     SGX PSW/libsgx installed: true
 
-The program terminates successfully if all SGX1 components are detected and running, otherwise
-the program exits with an error code (see the source code for possible values).
-To suppress printing output use the --quiet argument.
+The program terminates successfully if all SGX1 components are detected and
+running, otherwise the program exits with an error code
+(see the source code for possible values). To suppress printing output use the
+--quiet argument.
 
 
 SGX quote dump
@@ -32,12 +36,12 @@ SGX quote dump
 
 Displays internal structure of an SGX quote::
 
-    Usage: quote_dump [options] <quote path>
+    Usage: gramine-sgx-quote-dump [options] <quote path>
     Available options:
       --help, -h  Display this help
       --msb, -m   Display hex strings in big-endian order
 
-    $ quote_dump -m gr.quote
+    $ gramine-sgx-quote-dump -m gr.quote
     version           : 0002
     sign_type         : 0001
     epid_group_id     : 00000aef
@@ -70,10 +74,10 @@ Displays internal structure of an SGX quote::
 Intel Attestation Service submitter
 -----------------------------------
 
-Submits requests to Intel Attestation Service (IAS) for retrieving EPID signature revocation lists
-and for verifying attestation evidence (enclave quote)::
+Submits requests to Intel Attestation Service (IAS) for retrieving EPID
+signature revocation lists and for verifying attestation evidence (enclave quote)::
 
-    Usage: ias_request <request> [options]
+    Usage: gramine-sgx-ias-request <request> [options]
     Available requests:
       sigrl                     Retrieve signature revocation list for a given EPID group
       report                    Verify attestation evidence (quote)
@@ -99,12 +103,12 @@ and for verifying attestation evidence (enclave quote)::
 
 Example SigRL retrieval::
 
-    $ ias_request sigrl -k $IAS_API_KEY -g ef0a0000 -i sigrl
+    $ gramine-sgx-ias-request sigrl -k $IAS_API_KEY -g ef0a0000 -i sigrl
     No SigRL for given EPID group ID ef0a0000
 
 Example quote verification::
 
-    $ ias_request report -k $IAS_API_KEY -q gr.quote -r ias.report -s ias.sig -c ias.cert -a ias.adv -v
+    $ gramine-sgx-ias-request report -k $IAS_API_KEY -q gr.quote -r ias.report -s ias.sig -c ias.cert -a ias.adv -v
     Verbose output enabled
     IAS request:
     {"isvEnclaveQuote":"AgABAO8..."}
@@ -122,10 +126,11 @@ Example quote verification::
 Intel Attestation Report verifier
 ---------------------------------
 
-Verifies attestation report retrieved from IAS (using ``ias_request`` for example). Also verifies
-that the quote from the report contains expected values::
+Verifies attestation report retrieved from IAS
+(using ``gramine-sgx-ias-request`` for example). Also verifies that the quote
+from the report contains expected values::
 
-    Usage: verify_ias_report [options]
+    Usage: gramine-sgx-ias-verify-report [options]
     Available options:
       --help, -h                Display this help
       --verbose, -v             Enable verbose output
@@ -144,7 +149,7 @@ that the quote from the report contains expected values::
 
 Example report verification with all options enabled::
 
-    $ verify_ias_report -v -m -r rp -s sp -i ias.pem -o -d -n thisisnonce -S 14b284525c45c4f526bf1535d05bd88aa73b9e184464f2d97be3dabc0d187b57 -E 4d69102c40401f40a54eb156601be73fb7605db0601845580f036fd284b7b303 -R 0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000004ba476e321e12c720000000000000001 -P 0 -V 0
+    $ gramine-sgx-ias-verify-report -v -m -r rp -s sp -i ias.pem -o -d -n thisisnonce -S 14b284525c45c4f526bf1535d05bd88aa73b9e184464f2d97be3dabc0d187b57 -E 4d69102c40401f40a54eb156601be73fb7605db0601845580f036fd284b7b303 -R 0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000004ba476e321e12c720000000000000001 -P 0 -V 0
     Verbose output enabled
     Endianness set to MSB
     Using IAS public key from file 'ias.pem'
--- a/Pal/src/host/Linux-SGX/tools/ias-request/.gitignore
+++ b/Pal/src/host/Linux-SGX/tools/ias-request/.gitignore
@@ -1,1 +1,0 @@
-/ias_request
--- a/Pal/src/host/Linux-SGX/tools/is-sgx-available/.gitignore
+++ b/Pal/src/host/Linux-SGX/tools/is-sgx-available/.gitignore
@@ -1,1 +1,0 @@
-/is_sgx_available
--- a/Pal/src/host/Linux-SGX/tools/pf_crypt/.gitignore
+++ b/Pal/src/host/Linux-SGX/tools/pf_crypt/.gitignore
@@ -1,1 +1,0 @@
-/pf_crypt
--- a/Pal/src/host/Linux-SGX/tools/pf_tamper/.gitignore
+++ b/Pal/src/host/Linux-SGX/tools/pf_tamper/.gitignore
@@ -1,1 +1,0 @@
-/pf_tamper
--- a/Pal/src/host/Linux-SGX/tools/quote-dump/.gitignore
+++ b/Pal/src/host/Linux-SGX/tools/quote-dump/.gitignore
@@ -1,1 +1,0 @@
-/quote_dump
--- a/Pal/src/host/Linux-SGX/tools/verify-ias-report/.gitignore
+++ b/Pal/src/host/Linux-SGX/tools/verify-ias-report/.gitignore
@@ -1,1 +1,0 @@
-/verify_ias_report


chiache (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: OSCAR Lab)

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: OSCAR Lab) (waiting on @vijaydhanraj)

___
*[Documentation/building.rst, line 190 at r1](https://reviewable.io/reviews/gramineproject/gramine/198#-Mnv52zZ9DIAjaXAmtJB:-Mnv52zZ9DIAjaXAmtJC:bvpj7fd) ([raw file](https://github.com/gramineproject/gramine/blob/ddc05b3709582a47f65586b654187fc1977e37b1/Documentation/building.rst#L190)):*
> ```reStructuredText
>   running regression tests. See :doc:`devel/contributing` for details.
> 
> - In order to run SGX tools with DCAP version of RA-TLS library `ra_tls_verify_dcap.so`, build with
> ```

Please parenthesize `ra_tls_verify_dcap.so`, otherwise it has a different meaning than intended.
___
*[Documentation/building.rst, line 191 at r1](https://reviewable.io/reviews/gramineproject/gramine/198#-MnvDO6x9OcwCIA6sEVz:-MnvDO6y7WmquGQc9XzZ:bcdd73n) ([raw file](https://github.com/gramineproject/gramine/blob/ddc05b3709582a47f65586b654187fc1977e37b1/Documentation/building.rst#L191)):*
> ```
> :doc:`
> ```

This seems to be broken, the text is rendered as usual.
CC: @woju
___
*[Documentation/building.rst, line 195 at r1](https://reviewable.io/reviews/gramineproject/gramine/198#-MnvE6WYCKNWujW2BmqF:-MnvE6WZ8wjS0nRzLmNy:b-pdsfj9) ([raw file](https://github.com/gramineproject/gramine/blob/ddc05b3709582a47f65586b654187fc1977e37b1/Documentation/building.rst#L195)):*
> ```
> `ra_tls_verify_epid.so`
> ```

ditto

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: OSCAR Lab) (waiting on @mkow, @vijaydhanraj, and @woju)

___
*[Documentation/building.rst, line 188 at r1](https://reviewable.io/reviews/gramineproject/gramine/198#-MnyNF3V58vXTA-y3pmz:-MnyNF3V58vXTA-y3pn-:bxnyjj8) ([raw file](https://github.com/gramineproject/gramine/blob/ddc05b3709582a47f65586b654187fc1977e37b1/Documentation/building.rst#L188)):*
> ```reStructuredText
> 
> - To build test binaries, run :command:`meson -Dtests=enabled`. This is necessary if you will be
>   running regression tests. See :doc:`devel/contributing` for details.
> ```

We use 80-char wrapping for our Rst documentation (instead of 100-char wrapping in the rest of the codebase). Please re-wrap all new text, including this one.
___
*[Documentation/building.rst, line 190 at r1](https://reviewable.io/reviews/gramineproject/gramine/198#-Mnv52zZ9DIAjaXAmtJB:-MnyM6qc5OJGnwtoCvmy:bq7gdml) ([raw file](https://github.com/gramineproject/gramine/blob/ddc05b3709582a47f65586b654187fc1977e37b1/Documentation/building.rst#L190)):*

Please parenthesize `ra_tls_verify_dcap.so`, otherwise it has a different meaning than intended.
</blockquote></details>

+1 to parenthesize. And use double-backticks here.
___
*[Documentation/building.rst, line 191 at r1](https://reviewable.io/reviews/gramineproject/gramine/198#-MnvDO6x9OcwCIA6sEVz:-MnyMPKQ0d7_FTCcMnwn:b-mqj156) ([raw file](https://github.com/gramineproject/gramine/blob/ddc05b3709582a47f65586b654187fc1977e37b1/Documentation/building.rst#L191)):*

> ```
> :doc:`
> ```

This seems to be broken, the text is rendered as usual.
CC: @woju
</blockquote></details>

You can't use such files (not part of the ReText documentation) as `:doc:`. So, instead of
```
:doc:`CI-Examples/ra-tls-mbedtls/README.md`
```

Simply use a hyperlink to this README:
```
`RA-TLS example's README <https://github.com/gramineproject/gramine/blob/master/CI-Examples/ra-tls-mbedtls/README.md>`__
```
___
*[Documentation/building.rst, line 195 at r1](https://reviewable.io/reviews/gramineproject/gramine/198#-MnvE6WYCKNWujW2BmqF:-MnyN6Ot-QUVeT0rS9x8:b-gtezqi) ([raw file](https://github.com/gramineproject/gramine/blob/ddc05b3709582a47f65586b654187fc1977e37b1/Documentation/building.rst#L195)):*

> ```
> `ra_tls_verify_epid.so`
> ```

ditto
</blockquote></details>

ditto (parenthesize + double-backticks)
___
*[Pal/src/host/Linux-SGX/tools/README.rst, line 7 at r1](https://reviewable.io/reviews/gramineproject/gramine/198#-MnyNSez7rUVnS_fQdRG:-MnyNSez7rUVnS_fQdRH:bxnyjj8) ([raw file](https://github.com/gramineproject/gramine/blob/ddc05b3709582a47f65586b654187fc1977e37b1/Pal/src/host/Linux-SGX/tools/README.rst#L7)):*
> ```reStructuredText
> 
> All the below tools can be simply run from any directory as they are installed globally on the
> system.
> ```

We use 80-char wrapping for our Rst documentation (instead of 100-char wrapping in the rest of the codebase). Please re-wrap all new text, including this one.
___
*[Pal/src/host/Linux-SGX/tools/README.rst, line 14 at r1](https://reviewable.io/reviews/gramineproject/gramine/198#-MnyLhXd676DqtoHkmme:-MnyLhXd676DqtoHkmmf:bludr05) ([raw file](https://github.com/gramineproject/gramine/blob/ddc05b3709582a47f65586b654187fc1977e37b1/Pal/src/host/Linux-SGX/tools/README.rst#L14)):*
> ```reStructuredText
> Example output::
> 
>     > is_sgx_available
> ```

Isn't it `is-sgx-available`, with `-` instead of `_`? See https://gramine.readthedocs.io/en/latest/manpages/is-sgx-available.html

I remember we changed this recently (for release 1.0) because in all other tools we use `-` (dash). So we decided to be uniform.

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @mkow, and @woju)

___
*[Documentation/building.rst, line 188 at r1](https://reviewable.io/reviews/gramineproject/gramine/198#-MnyNF3V58vXTA-y3pmz:-Mo--DgO8atcTTE1KR7x:b6a1gwb) ([raw file](https://github.com/gramineproject/gramine/blob/ddc05b3709582a47f65586b654187fc1977e37b1/Documentation/building.rst#L188)):*

We use 80-char wrapping for our Rst documentation (instead of 100-char wrapping in the rest of the codebase). Please re-wrap all new text, including this one.
</blockquote></details>

Got it, thanks. Re-wrapped to 80-char limit.
___
*[Documentation/building.rst, line 190 at r1](https://reviewable.io/reviews/gramineproject/gramine/198#-Mnv52zZ9DIAjaXAmtJB:-MnzxTKI4aEjrbLoA6LI:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/ddc05b3709582a47f65586b654187fc1977e37b1/Documentation/building.rst#L190)):*

+1 to parenthesize. And use double-backticks here.
</blockquote></details>

Done.
___
*[Documentation/building.rst, line 191 at r1](https://reviewable.io/reviews/gramineproject/gramine/198#-MnvDO6x9OcwCIA6sEVz:-MnzzJp-AyHzzpXCXk8K:b23lji1) ([raw file](https://github.com/gramineproject/gramine/blob/ddc05b3709582a47f65586b654187fc1977e37b1/Documentation/building.rst#L191)):*

You can't use such files (not part of the ReText documentation) as `:doc:`. So, instead of
```
:doc:`CI-Examples/ra-tls-mbedtls/README.md`
```

Simply use a hyperlink to this README:
```
`RA-TLS example's README <https://github.com/gramineproject/gramine/blob/master/CI-Examples/ra-tls-mbedtls/README.md>`__
```
</blockquote></details>

Done, but this exceeds more than 80 lines, is this fine?
___
*[Documentation/building.rst, line 195 at r1](https://reviewable.io/reviews/gramineproject/gramine/198#-MnvE6WYCKNWujW2BmqF:-Mo--5t31nGDbskOPvM5:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/ddc05b3709582a47f65586b654187fc1977e37b1/Documentation/building.rst#L195)):*

ditto (parenthesize + double-backticks)
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/tools/README.rst, line 7 at r1](https://reviewable.io/reviews/gramineproject/gramine/198#-MnyNSez7rUVnS_fQdRG:-Mo-2sMUB39Dp6oh-i33:b-4c3u9w) ([raw file](https://github.com/gramineproject/gramine/blob/ddc05b3709582a47f65586b654187fc1977e37b1/Pal/src/host/Linux-SGX/tools/README.rst#L7)):*

We use 80-char wrapping for our Rst documentation (instead of 100-char wrapping in the rest of the codebase). Please re-wrap all new text, including this one.
</blockquote></details>

Done, also fixed the old ones which exceeded 80-chars. Just left the commands/outputs to be more than 80-chars.
___
*[Pal/src/host/Linux-SGX/tools/README.rst, line 14 at r1](https://reviewable.io/reviews/gramineproject/gramine/198#-MnyLhXd676DqtoHkmme:-Mo-05u39IENmtQtNV7p:bxzxdzf) ([raw file](https://github.com/gramineproject/gramine/blob/ddc05b3709582a47f65586b654187fc1977e37b1/Pal/src/host/Linux-SGX/tools/README.rst#L14)):*

Isn't it `is-sgx-available`, with `-` instead of `_`? See https://gramine.readthedocs.io/en/latest/manpages/is-sgx-available.html

I remember we changed this recently (for release 1.0) because in all other tools we use `-` (dash). So we decided to be uniform.
</blockquote></details>

My bad, I looked at the file name and assumed `_` instead of `-`. It is indeed `-`.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: 0 of 2 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @chiache, @mkow, and @woju)

___
*[Documentation/building.rst, line 191 at r1](https://reviewable.io/reviews/gramineproject/gramine/198#-MnvDO6x9OcwCIA6sEVz:-Mo2WiAB9Rb-RGy2yoM_:b-am7uda) ([raw file](https://github.com/gramineproject/gramine/blob/ddc05b3709582a47f65586b654187fc1977e37b1/Documentation/building.rst#L191)):*

Done, but this exceeds more than 80 lines, is this fine?
</blockquote></details>

Yes, it is fine, we can't do much about it.

dimakuv:
Jenkins, test this please

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: 2 of 3 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @chiache, @mkow, @vijaydhanraj, and @woju)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo2_IQpCcgfyMqmpqEd:-Mo2_IQpCcgfyMqmpqEe:bsptzop) (no related file):*
@vijaydhanraj Please also fix this occurence of `is_sgx_available`: https://github.com/gramineproject/gramine/issues/204

And also please grep for the old name `is_sgx_available` in all our Gramine code & documentation and fix all occurences.

---
___
*[Documentation/building.rst, line 117 at r3](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo4TMcfE7HuU9AlnjCh:-Mo4TMcfE7HuU9AlnjCi:b-53wf6k) ([raw file](https://github.com/gramineproject/gramine/blob/7bc59d7f0de07019adf4ec21dbf1bd1abd15e07e/Documentation/building.rst#L117)):*
> ```reStructuredText
>    This module is a |~| quick-and-dirty hack with dangerous security hole
>    (allows unauthorized local privilege escalation). "Do not use for production"
>    is not a |~| joke. We use it only for testing on old kernels.
> ```

Now that I look at it... Maybe we should delete this section 2b altogether. Having such a huge warning in our official docs doesn't look good (we are a security-first project and we allow such a terrible security-critical hack in our documentation).

@mkow @boryspoplawski @woju Do we remove just remove this 2b section and never talk about `gsgx.ko` again?
___
*[Documentation/building.rst, line 134 at r3](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo4T3MGA9LeDetiSdgd:-Mo4T3MGA9LeDetiSdge:b12971g) ([raw file](https://github.com/gramineproject/gramine/blob/7bc59d7f0de07019adf4ec21dbf1bd1abd15e07e/Documentation/building.rst#L134)):*
> ```reStructuredText
> 
> If you have an older CPU without :term:`FLC` support, you need to download and
> install the the following out-of-tree(OOT) Intel SGX driver:
> ```

Please add a space between `out-of-tree` and `(OOT)`
___
*[Documentation/quickstart.rst, line 53 at r3](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo4TqqmCzVkP025K54m:-Mo4TqqmCzVkP025K54n:bolblcf) ([raw file](https://github.com/gramineproject/gramine/blob/7bc59d7f0de07019adf4ec21dbf1bd1abd15e07e/Documentation/quickstart.rst#L53)):*
> ```reStructuredText
> 
> #. Ensure that Intel SGX is enabled on your platform using
>    :program:`is-sgx-available`.
> ```

Actually, Pawel had a good suggestion: https://github.com/gramineproject/gramine/issues/204#issuecomment-963914406

We can't ask to use `is-sgx-available` tool because it is part of the Gramine-SGX installation. So at this first step, this tool doesn't exist.

Instead, we can ask people to check for SGX using the CPUID tool:
```
    sudo apt-get install -y cpuid
    cpuid | grep -i sgx
    # output must contain lines like "SGX: Software Guard Extensions supported = true"
```

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: 2 of 3 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @chiache, @dimakuv, @mkow, @vijaydhanraj, and @woju)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo2_IQpCcgfyMqmpqEd:-Mo4X1L51KJZfm9lY609:b-2xwwky) (no related file):*

@vijaydhanraj Please also fix this occurence of `is_sgx_available`: https://github.com/gramineproject/gramine/issues/204

And also please grep for the old name `is_sgx_available` in all our Gramine code & documentation and fix all occurences.
</blockquote></details>

@dimakuv, I don't see any occurrence of `is_sgx_available`. Can you please point if I missed converting `is_sgx_available`->`is-sgx-available`.

Also, I think here the suggestion is to use `cpuid` rather than `is-sgx-available` right?

---
___
*[Documentation/building.rst, line 117 at r3](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo4TMcfE7HuU9AlnjCh:-Mo4XuZY8j49Jfo3oH9F:b-kgerry) ([raw file](https://github.com/gramineproject/gramine/blob/7bc59d7f0de07019adf4ec21dbf1bd1abd15e07e/Documentation/building.rst#L117)):*

Now that I look at it... Maybe we should delete this section 2b altogether. Having such a huge warning in our official docs doesn't look good (we are a security-first project and we allow such a terrible security-critical hack in our documentation).

@mkow @boryspoplawski @woju Do we remove just remove this 2b section and never talk about `gsgx.ko` again?
</blockquote></details>

If we want to remove `gsgx.ko` then should we say we only support 5.4 kernel and upwards?

woju (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: 2 of 3 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @chiache, @dimakuv, @mkow, and @vijaydhanraj)

___
*[Documentation/building.rst, line 191 at r1](https://reviewable.io/reviews/gramineproject/gramine/198#-MnvDO6x9OcwCIA6sEVz:-Mo5KlDZEAthKX2WE-Qm:b-o1eyif) ([raw file](https://github.com/gramineproject/gramine/blob/ddc05b3709582a47f65586b654187fc1977e37b1/Documentation/building.rst#L191)):*

Yes, it is fine, we can't do much about it.
</blockquote></details>

Yes. `:doc:` only works for files under `Documentation/`, and you can't use `../`, obviously. (It might work for markdown though, but I'd prefer to rewrite to rST instead.) Link to github is fine I guess.
___
*[Documentation/building.rst, line 117 at r3](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo4TMcfE7HuU9AlnjCh:-Mo5HFlHElFZSB34NRBs:b-gbktbc) ([raw file](https://github.com/gramineproject/gramine/blob/7bc59d7f0de07019adf4ec21dbf1bd1abd15e07e/Documentation/building.rst#L117)):*
I think we can drop gsgx.ko, it doesn't bring any value now that there are other options. Unless someone speaks up and says s/he's using it, with very good explanation why this insecure contraption is still needed.

> If we want to remove gsgx.ko then should we say we only support 5.4 kernel and upwards?

That's complicated. RHEL8(-compatible) distros have kernel that's nominally 4.18, but has a metric bucketload of patches backported, including inkernel SGX driver, which is supposed to be only in 5.11 kernel.

Ubuntu 18.04LTS has 4.15, and doesn't have this patches, so user needs 5.4 with FSGS patchset or 5.11, either way self-compiled. Ubuntu 20.04LTS has 5.4 without FSGS patchset, same situation. Kernel 5.11 is available only in 21.04, which is not LTS.

Debian 11 has kernel 5.10 by default, but there are newer kernels available in backports (5.14 at the time of writing). We can suggest that on Debian 11 `bullseye-backports` repo should be used to update kernel.
___
*[Documentation/quickstart.rst, line 53 at r3](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo4TqqmCzVkP025K54m:-Mo5KEW7AsJz0Mjb7l5R:b-vyuty8) ([raw file](https://github.com/gramineproject/gramine/blob/7bc59d7f0de07019adf4ec21dbf1bd1abd15e07e/Documentation/quickstart.rst#L53)):*

Actually, Pawel had a good suggestion: https://github.com/gramineproject/gramine/issues/204#issuecomment-963914406

We can't ask to use `is-sgx-available` tool because it is part of the Gramine-SGX installation. So at this first step, this tool doesn't exist.

Instead, we can ask people to check for SGX using the CPUID tool:
```
    sudo apt-get install -y cpuid
    cpuid | grep -i sgx
    # output must contain lines like "SGX: Software Guard Extensions supported = true"
```
</blockquote></details>

Even better, we should drop compilation from `quickstart`, and instead tell people to install from packages. That way is-sgx-available works.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: 2 of 3 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @chiache, @mkow, @vijaydhanraj, and @woju)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo2_IQpCcgfyMqmpqEd:-Mo7r-vQ6_qfWWjH9XnO:bze89sk) (no related file):*
This is what I grepped:
```
~/gramineproject/gramine$ grep -rIw is_sgx_available
Documentation/quickstart.rst:   :program:`is_sgx_available`.
Pal/src/host/Linux-SGX/tools/README.rst:    > ./is_sgx_available/is_sgx_available
Pal/src/host/Linux-SGX/tools/is-sgx-available/.gitignore:/is_sgx_available
Pal/src/host/Linux-SGX/tools/is-sgx-available/meson.build:    'is_sgx_available.cpp',
```

Looks like `Pal/src/host/Linux-SGX/tools/is-sgx-available/.gitignore` still needs to be fixed.

> Also, I think here the suggestion is to use cpuid rather than is-sgx-available right?

Yes, please use this one now.

---
___
*[Documentation/building.rst, line 117 at r3](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo4TMcfE7HuU9AlnjCh:-Mo7rX7n5QwRebHUL9Yw:b-hu0951) ([raw file](https://github.com/gramineproject/gramine/blob/7bc59d7f0de07019adf4ec21dbf1bd1abd15e07e/Documentation/building.rst#L117)):*
> If we want to remove gsgx.ko then should we say we only support 5.4 kernel and upwards?

Maybe we should simply not mention this?.. As Woju points out, the situation is much more complicated (for different distros). I would prefer to *not* write any too-complicated explanations.

Woju also wants to remove this 2b section. I'm pretty sure @mkow will be also for removing :) So let's remove it from here.
___
*[Documentation/quickstart.rst, line 53 at r3](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo4TqqmCzVkP025K54m:-Mo7s-KV-Xu8C6uOXfou:bsm22a5) ([raw file](https://github.com/gramineproject/gramine/blob/7bc59d7f0de07019adf4ec21dbf1bd1abd15e07e/Documentation/quickstart.rst#L53)):*

Even better, we should drop compilation from `quickstart`, and instead tell people to install from packages. That way is-sgx-available works.
</blockquote></details>

What Woju proposes is good, but definitely not for this PR. When we're ready with all packages, we can re-write QuickStart. But not now, not in this PR.

woju (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: 2 of 3 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @chiache, @dimakuv, @mkow, and @vijaydhanraj)

___
*[Documentation/building.rst, line 117 at r3](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo4TMcfE7HuU9AlnjCh:-Mo8ppR-7cDizeAOfhzk:b37ov56) ([raw file](https://github.com/gramineproject/gramine/blob/7bc59d7f0de07019adf4ec21dbf1bd1abd15e07e/Documentation/building.rst#L117)):*

> If we want to remove gsgx.ko then should we say we only support 5.4 kernel and upwards?

Maybe we should simply not mention this?.. As Woju points out, the situation is much more complicated (for different distros). I would prefer to *not* write any too-complicated explanations.

Woju also wants to remove this 2b section. I'm pretty sure @mkow will be also for removing :) So let's remove it from here.
</blockquote></details>

Yes. Blocking until this is done or decided otherwise.
___
*[Documentation/quickstart.rst, line 53 at r3](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo4TqqmCzVkP025K54m:-Mo8phcnDngHJcQzUtF0:b-sxtmed) ([raw file](https://github.com/gramineproject/gramine/blob/7bc59d7f0de07019adf4ec21dbf1bd1abd15e07e/Documentation/quickstart.rst#L53)):*

What Woju proposes is good, but definitely not for this PR. When we're ready with all packages, we can re-write QuickStart. But not now, not in this PR.
</blockquote></details>

Hmm, OK. Unblocking.

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: 0 of 9 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @chiache, @dimakuv, and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo2_IQpCcgfyMqmpqEd:-Mo8mxufCH0qep0QZlG1:b7u15up) (no related file):*
> Looks like Pal/src/host/Linux-SGX/tools/is-sgx-available/.gitignore still needs to be fixed.

I think this `.gitignore` file itself can be removed since the binaries are not generated in this directory anymore.

Thanks for the clarification, will use `cpuid`

---
___
*[Documentation/building.rst, line 117 at r3](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo4TMcfE7HuU9AlnjCh:-Mo8nRr9AW8b_mlPK02f:b-l5nmqi) ([raw file](https://github.com/gramineproject/gramine/blob/7bc59d7f0de07019adf4ec21dbf1bd1abd15e07e/Documentation/building.rst#L117)):*

Yes. Blocking until this is done or decided otherwise.
</blockquote></details>

I have removed this, but my concern is that we don't seem to address the case where the patch doesn't apply cleanly. Also, by not documenting we don't reduce any hacks but simply obscure information about this security hole from our users. What about having a note/warning on this? Something like below,
```
For older kernels (5.3 and lower), where the patchset doesn't apply cleanly install Gramine specific FSGSBASE driver (https://github.com/oscarlab/graphene-sgx-driver). But this does have a dangerous security hole and should NOT be used for production.
```
___
*[Documentation/building.rst, line 134 at r3](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo4T3MGA9LeDetiSdgd:-Mo8uzUp2OYWtV2AV9eN:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/7bc59d7f0de07019adf4ec21dbf1bd1abd15e07e/Documentation/building.rst#L134)):*

Please add a space between `out-of-tree` and `(OOT)`
</blockquote></details>

Done.
___
*[Documentation/quickstart.rst, line 53 at r3](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo4TqqmCzVkP025K54m:-Mo8v4vW653uUmUhDxeb:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/7bc59d7f0de07019adf4ec21dbf1bd1abd15e07e/Documentation/quickstart.rst#L53)):*

Hmm, OK. Unblocking.
</blockquote></details>

Done.

dimakuv (changes requested):
> Reviewed 8 of 8 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: 8 of 9 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @chiache, @mkow, and @vijaydhanraj)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo2_IQpCcgfyMqmpqEd:-Mo9RIkA1ChHV3f50dmT:bxsmcwv) (no related file):*

> Looks like Pal/src/host/Linux-SGX/tools/is-sgx-available/.gitignore still needs to be fixed.

I think this `.gitignore` file itself can be removed since the binaries are generated in this directory anymore.

Thanks for the clarification, will use `cpuid`
</blockquote></details>

OK, thanks for cleaning it up!

---
___
*[Documentation/building.rst, line 117 at r3](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo4TMcfE7HuU9AlnjCh:-Mo9QYNyDMy19jp0i6ED:bxni4fr) ([raw file](https://github.com/gramineproject/gramine/blob/7bc59d7f0de07019adf4ec21dbf1bd1abd15e07e/Documentation/building.rst#L117)):*

I have removed this, but my concern is that we don't seem to address the case where the patch doesn't apply cleanly. Also, by not documenting we don't reduce any hacks but simply obscure information about this security hole from our users. What about having a note/warning on this? Something like below,
```
For older kernels (5.3 and lower), where the patchset doesn't apply cleanly install Gramine specific FSGSBASE driver (https://github.com/oscarlab/graphene-sgx-driver). But this does have a dangerous security hole and should NOT be used for production.
```

</blockquote></details>

What do you mean, we `obscure information about this security hole from our users`? We simply don't tell users about the possibility of using our `gsgx.ko` workaround (which has a security hole). In this case, normal users will not even try to use it because they don't know about it. So no security hole will be there (because users didn't install `gsgx.ko`).

So I am against adding anything at all about Linux 5.3 and lower.
___
*[Documentation/building.rst, line 324 at r4](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo9QvAR2VmgL73pd7Ay:-Mo9QvAR2VmgL73pd7Az:b-jbldjw) ([raw file](https://github.com/gramineproject/gramine/blob/28ed03d64430f4321ea4dd84c3e67e70ca35969d/Documentation/building.rst#L324)):*
> ```reStructuredText
>        AT_HWCAP2:       0x0
> 
>        # Linux kernel supports FSGSBASE
> ```

Per request from @mkow (see https://github.com/gramineproject/gramine/issues/203#issuecomment-965024317), please append ` (this is one example, note that bit 2 is set)` to this comment

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: 8 of 9 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @chiache, @mkow, and @vijaydhanraj)

___
*[Documentation/building.rst, line 117 at r3](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo4TMcfE7HuU9AlnjCh:-Mo9U09CCSnEqwn0rrnL:b-9mnicv) ([raw file](https://github.com/gramineproject/gramine/blob/7bc59d7f0de07019adf4ec21dbf1bd1abd15e07e/Documentation/building.rst#L117)):*

What do you mean, we `obscure information about this security hole from our users`? We simply don't tell users about the possibility of using our `gsgx.ko` workaround (which has a security hole). In this case, normal users will not even try to use it because they don't know about it. So no security hole will be there (because users didn't install `gsgx.ko`).

So I am against adding anything at all about Linux 5.3 and lower.
</blockquote></details>

So today anyone can install `gsgx.ko` module and use it right?  If we don't add this disclaimer in the documentation, we might keep the user in dark of any issues with using this module.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: 8 of 9 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @chiache, @mkow, and @vijaydhanraj)

___
*[Documentation/building.rst, line 117 at r3](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo4TMcfE7HuU9AlnjCh:-Mo9WpR55iR0URvoWl3K:bfwk6n1) ([raw file](https://github.com/gramineproject/gramine/blob/7bc59d7f0de07019adf4ec21dbf1bd1abd15e07e/Documentation/building.rst#L117)):*

So today anyone can install `gsgx.ko` module and use it right?  If we don't add this disclaimer in the documentation, we might keep the user in dark of any issues with using this module.
</blockquote></details>

We don't mention `gsgx.ko` anywhere in the Gramine documentation or on any of our repositories. It can be found only in that legacy https://github.com/oscarlab/graphene-sgx-driver repo -- but the user needs to *find it himself* first. And that legacy repo contains a huge-huge warning in the README.

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: 7 of 9 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @chiache, @dimakuv, and @mkow)

___
*[Documentation/building.rst, line 117 at r3](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo4TMcfE7HuU9AlnjCh:-Mo9w0w_6jIQmjK_m6M8:b-i39rjh) ([raw file](https://github.com/gramineproject/gramine/blob/7bc59d7f0de07019adf4ec21dbf1bd1abd15e07e/Documentation/building.rst#L117)):*

We don't mention `gsgx.ko` anywhere in the Gramine documentation or on any of our repositories. It can be found only in that legacy https://github.com/oscarlab/graphene-sgx-driver repo -- but the user needs to *find it himself* first. And that legacy repo contains a huge-huge warning in the README.
</blockquote></details>

OK, I will skip section 2b altogether.
___
*[Documentation/building.rst, line 324 at r4](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo9QvAR2VmgL73pd7Ay:-Mo9ukUz76wpltZixCXz:bgnkve4) ([raw file](https://github.com/gramineproject/gramine/blob/28ed03d64430f4321ea4dd84c3e67e70ca35969d/Documentation/building.rst#L324)):*

Per request from @mkow (see https://github.com/gramineproject/gramine/issues/203#issuecomment-965024317), please append ` (this is one example, note that bit 2 is set)` to this comment
</blockquote></details>

Just reworded it a little bit.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: 8 of 9 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @chiache, and @mkow)

___
*[Documentation/building.rst, line 324 at r4](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo9QvAR2VmgL73pd7Ay:-MoD85LT44c71Y9VSOnP:b-vem61j) ([raw file](https://github.com/gramineproject/gramine/blob/28ed03d64430f4321ea4dd84c3e67e70ca35969d/Documentation/building.rst#L324)):*

Just reworded it a little bit.
</blockquote></details>

OK, looks good to me, thanks!

dimakuv:
Jenkins, test this please

mkow (changes requested):
> Reviewed 1 of 2 files at r2, 7 of 8 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @vijaydhanraj)

___
*[-- commits, line 2 at r5](https://reviewable.io/reviews/gramineproject/gramine/198#-MoGNmKXCNAXXhBWgtmc):*
> ```
>  with documentation
> ```

You can drop this, it already says `[Docs]`
___
*[Documentation/building.rst, line 324 at r4](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo9QvAR2VmgL73pd7Ay:-MoGQCuO3ZuiWZGINuLu:b-mn3dtx) ([raw file](https://github.com/gramineproject/gramine/blob/28ed03d64430f4321ea4dd84c3e67e70ca35969d/Documentation/building.rst#L324)):*

OK, looks good to me, thanks!
</blockquote></details>

Not done? It doesn't say anywhere that these are just example values, users will be confused if they see "0x6", for example.
___
*[Documentation/quickstart.rst, line 52 at r5](https://reviewable.io/reviews/gramineproject/gramine/198#-MoGQam48pjsFCkC2v01:-MoGQam48pjsFCkC2v02:b-kasug8) ([raw file](https://github.com/gramineproject/gramine/blob/be7b5d3f04735e0ef78435cc39c5a005b7f79887/Documentation/quickstart.rst#L52)):*
> ```reStructuredText
> descriptions in :doc:`building`.
> 
> #. Ensure that Intel SGX is enabled on your platform using::
> ```

`using` -> `by doing`

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, and @vijaydhanraj)

___
*[-- commits, line 2 at r5](https://reviewable.io/reviews/gramineproject/gramine/198#-MoGNmKXCNAXXhBWgtmc):*

> ```
>  with documentation
> ```

You can drop this, it already says `[Docs]`
</blockquote></details>

We'll do it during final rebase.
___
*[Documentation/building.rst, line 324 at r4](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo9QvAR2VmgL73pd7Ay:-MoIBzSRDnSljiioFYoN:b-4au9h7) ([raw file](https://github.com/gramineproject/gramine/blob/28ed03d64430f4321ea4dd84c3e67e70ca35969d/Documentation/building.rst#L324)):*

Not done? It doesn't say anywhere that these are just example values, users will be confused if they see "0x6", for example.
</blockquote></details>

@vijaydhanraj To satisfy Michal, let's also expand this comment like this:
```
       # Linux kernel supports FSGSBASE (example where only bit 2 is set)
```

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: 7 of 9 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

___
*[Documentation/building.rst, line 324 at r4](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo9QvAR2VmgL73pd7Ay:-MoJZby_EUb0GqcKIjt9:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/28ed03d64430f4321ea4dd84c3e67e70ca35969d/Documentation/building.rst#L324)):*

@vijaydhanraj To satisfy Michal, let's also expand this comment like this:
```
       # Linux kernel supports FSGSBASE (example where only bit 2 is set)
```
</blockquote></details>

Done.
___
*[Documentation/quickstart.rst, line 52 at r5](https://reviewable.io/reviews/gramineproject/gramine/198#-MoGQam48pjsFCkC2v01:-MoJZdov1mCI_66frSFh:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/be7b5d3f04735e0ef78435cc39c5a005b7f79887/Documentation/quickstart.rst#L52)):*

`using` -> `by doing`
</blockquote></details>

Done.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @vijaydhanraj)

___
*[Documentation/building.rst, line 324 at r4](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo9QvAR2VmgL73pd7Ay:-MoKIiuj1UKJCcMmzgc0:b-ok1zf0) ([raw file](https://github.com/gramineproject/gramine/blob/28ed03d64430f4321ea4dd84c3e67e70ca35969d/Documentation/building.rst#L324)):*

Done.
</blockquote></details>

It's different than what Dmitrii suggested, you removed the precise explanation altogether and left only this example. We have to say that it's enabled iff bit 2 is set, and additionally may give an example (saying that this is only a sample value).

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, and @vijaydhanraj)

___
*[Documentation/building.rst, line 324 at r4](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo9QvAR2VmgL73pd7Ay:-MoKKjbNCnIZ4pVwTbXy:b98fxg4) ([raw file](https://github.com/gramineproject/gramine/blob/28ed03d64430f4321ea4dd84c3e67e70ca35969d/Documentation/building.rst#L324)):*

It's different than what Dmitrii suggested, you removed the precise explanation altogether and left only this example. We have to say that it's enabled iff bit 2 is set, and additionally may give an example (saying that this is only a sample value).
</blockquote></details>

Can you please clarify what you mean by an additional example? From the kernel code, I see that if FSGSBASE is supported, bit 2 is only set.
https://elixir.bootlin.com/linux/v5.15/source/arch/x86/kernel/cpu/common.c#L1571
https://elixir.bootlin.com/linux/v5.15/source/arch/x86/include/uapi/asm/hwcap2.h#L11

Given this, isn't it enough to say that the `FSGSBASE` support is enabled iff bit 2 is set? All we are asking the user to check is just check bit 2. What is the additional example?

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @vijaydhanraj)

___
*[Documentation/building.rst, line 324 at r4](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo9QvAR2VmgL73pd7Ay:-MoL0trm02zBZ4ueuvVU:bt3e85h) ([raw file](https://github.com/gramineproject/gramine/blob/28ed03d64430f4321ea4dd84c3e67e70ca35969d/Documentation/building.rst#L324)):*
> isn't it enough to say that the `FSGSBASE` support is enabled iff bit 2 is set?

It is, but it's not what you're saying in the comment. The comment says:
```
#. Also verify that the patched kernel supports FSGSBASE (example where only
   bit 2 is set)::
```

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, and @vijaydhanraj)

___
*[Documentation/building.rst, line 324 at r4](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo9QvAR2VmgL73pd7Ay:-MoL30008ggLjOFbJkja:b1bf1w) ([raw file](https://github.com/gramineproject/gramine/blob/28ed03d64430f4321ea4dd84c3e67e70ca35969d/Documentation/building.rst#L324)):*
Sorry getting confused here. I had the same comment before "the below command shows one approach of **verifying the support by checking if bit 2 is set**"

> Not done? It doesn't say anywhere that these are just example values, users will be confused if they see "0x6", for example.

I don't understand why the user would be confused if they see `0x6`. All that the comment is saying is that for `FSGSBASE` support to be present, bit 2 should be set. Also, what do you mean by "...these are example values"?

I think instead of bit 2, we should say `bit position 1` maybe that will clarify things if you are mixing up the value output(0x2) and bit 2.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @vijaydhanraj)

___
*[Documentation/building.rst, line 324 at r4](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo9QvAR2VmgL73pd7Ay:-MoLNanD8l0hKw1TC9s9:bhvylt9) ([raw file](https://github.com/gramineproject/gramine/blob/28ed03d64430f4321ea4dd84c3e67e70ca35969d/Documentation/building.rst#L324)):*

Sorry getting confused here. I had the same comment before "the below command shows one approach of **verifying the support by checking if bit 2 is set**"

> Not done? It doesn't say anywhere that these are just example values, users will be confused if they see "0x6", for example.

I don't understand why the user would be confused if they see `0x6`. All that the comment is saying is that for `FSGSBASE` support to be present, bit 2 should be set. Also, what do you mean by "...these are example values"?

I think instead of bit 2, we should say `bit position 1` maybe that will clarify things if you are mixing up the value output(0x2) and bit 2.

</blockquote></details>

It's not _me_ who had a problem with the original documentation - see https://github.com/gramineproject/gramine/issues/203. I only had a problem with pasting the example output without saying that this value (`0x2`) is just an example, i.e. not the only possible one.

And what Dmitrii proposed here is to only change this very line this comment is pinned at to say that this value is only an example. But you changed the original description instead, removing the part which said that 2nd bit must be set.

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, and @vijaydhanraj)

___
*[Documentation/building.rst, line 324 at r4](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo9QvAR2VmgL73pd7Ay:-MoLqBd0AZ_jouZSSbru:b-a04u43) ([raw file](https://github.com/gramineproject/gramine/blob/28ed03d64430f4321ea4dd84c3e67e70ca35969d/Documentation/building.rst#L324)):*

It's not _me_ who had a problem with the original documentation - see https://github.com/gramineproject/gramine/issues/203. I only had a problem with pasting the example output without saying that this value (`0x2`) is just an example, i.e. not the only possible one.

And what Dmitrii proposed here is to only change this very line this comment is pinned at to say that this value is only an example. But you changed the original description instead, removing the part which said that 2nd bit must be set.
</blockquote></details>

So what was the issue with the original comment which just talks about bit position and not the value of the output? 

If needed, we can modify this as `the below command shows one approach of verifying the support by checking if bit position 1 is set in the output`. Is this fine with you?

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, and @vijaydhanraj)

___
*[Documentation/building.rst, line 324 at r4](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo9QvAR2VmgL73pd7Ay:-MoM0ZPU0Y57tQmy_1Ni:b5if56d) ([raw file](https://github.com/gramineproject/gramine/blob/28ed03d64430f4321ea4dd84c3e67e70ca35969d/Documentation/building.rst#L324)):*
Just for clarification, by original comment I am referring to my earlier comment, `the below command shows one approach of verifying the support by checking if bit 2 is set`

and your response
> Not done? It doesn't say anywhere that these are just example values, users will be confused if they see "0x6", for example

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, and @vijaydhanraj)

___
*[Documentation/building.rst, line 324 at r4](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo9QvAR2VmgL73pd7Ay:-MoXrb2C0SuKkJrng1gS:b5q41kx) ([raw file](https://github.com/gramineproject/gramine/blob/28ed03d64430f4321ea4dd84c3e67e70ca35969d/Documentation/building.rst#L324)):*

Just for clarification, by original comment I am referring to my earlier comment, `the below command shows one approach of verifying the support by checking if bit 2 is set`

and your response
> Not done? It doesn't say anywhere that these are just example values, users will be confused if they see "0x6", for example


</blockquote></details>

Guys, let's just do exactly this:
```
#. Also verify that the patched kernel supports FSGSBASE (the below command
   must return that bit 2 is set)::

       # Linux kernel doesn't support FSGSBASE: patch or use higher version!
       $ LD_SHOW_AUXV=1 /bin/true | grep AT_HWCAP2
       AT_HWCAP2:       0x0

       # Linux kernel supports FSGSBASE (example where only bit 2 is set)
       $ LD_SHOW_AUXV=1 /bin/true | grep AT_HWCAP2
       AT_HWCAP2:       0x2
```

@vijaydhanraj Please use exactly this version. This should satisfy all members of this discussion. We're spending too much time on these nitpicks.

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, and @vijaydhanraj)

___
*[Documentation/building.rst, line 324 at r4](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo9QvAR2VmgL73pd7Ay:-MoZBsLl2I_brb3KTClN:b2zsivl) ([raw file](https://github.com/gramineproject/gramine/blob/28ed03d64430f4321ea4dd84c3e67e70ca35969d/Documentation/building.rst#L324)):*

Guys, let's just do exactly this:
```
#. Also verify that the patched kernel supports FSGSBASE (the below command
   must return that bit 2 is set)::

       # Linux kernel doesn't support FSGSBASE: patch or use higher version!
       $ LD_SHOW_AUXV=1 /bin/true | grep AT_HWCAP2
       AT_HWCAP2:       0x0

       # Linux kernel supports FSGSBASE (example where only bit 2 is set)
       $ LD_SHOW_AUXV=1 /bin/true | grep AT_HWCAP2
       AT_HWCAP2:       0x2
```

@vijaydhanraj Please use exactly this version. This should satisfy all members of this discussion. We're spending too much time on these nitpicks.
</blockquote></details>

@dimakuv Isn't this `(example where only bit 2 is set)`obvious? Why do we need this comment? It is better to keep the text short and clear.

Also, bit numbering usually starts from `0` and so it is better to say bit `1` must be set.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, and @vijaydhanraj)

___
*[Documentation/building.rst, line 324 at r4](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo9QvAR2VmgL73pd7Ay:-MobiP_u2VlvTm1ilnoi:bfob16k) ([raw file](https://github.com/gramineproject/gramine/blob/28ed03d64430f4321ea4dd84c3e67e70ca35969d/Documentation/building.rst#L324)):*
> Also, bit numbering usually starts from 0 and so it is better to say bit 1 must be set.

Yes, you are right. All `2` should be replaced with `1` in the text.

> @dimakuv Isn't this (example where only bit 2 is set)obvious? Why do we need this comment? It is better to keep the text short and clear.

Honestly, I don't know, I was quite ok with the previous version. I just don't want to spend another week on discussing this small comment, so let's just satisfy all reviewers and be done with it.

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: 8 of 9 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, and @vijaydhanraj)

___
*[Documentation/building.rst, line 324 at r4](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo9QvAR2VmgL73pd7Ay:-Mod3E12AwuKJPNHKYql:bm3yqt3) ([raw file](https://github.com/gramineproject/gramine/blob/28ed03d64430f4321ea4dd84c3e67e70ca35969d/Documentation/building.rst#L324)):*

> Also, bit numbering usually starts from 0 and so it is better to say bit 1 must be set.

Yes, you are right. All `2` should be replaced with `1` in the text.

> @dimakuv Isn't this (example where only bit 2 is set)obvious? Why do we need this comment? It is better to keep the text short and clear.

Honestly, I don't know, I was quite ok with the previous version. I just don't want to spend another week on discussing this small comment, so let's just satisfy all reviewers and be done with it.
</blockquote></details>

OK, done.

dimakuv (approved):
> Reviewed 1 of 2 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, and @vijaydhanraj)

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @vijaydhanraj)

___
*[Documentation/building.rst, line 324 at r4](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo9QvAR2VmgL73pd7Ay:-Mpci9eZDqAjphwhXgBr:b-rq2pct) ([raw file](https://github.com/gramineproject/gramine/blob/28ed03d64430f4321ea4dd84c3e67e70ca35969d/Documentation/building.rst#L324)):*

OK, done.
</blockquote></details>

I'm ok with this version, although maybe "bit at position 1" would be less ambiguous (but not blocking).

@vijaydhanraj just FYI: the intended meaning of the old "bit 2" was "bit of value 2", i.e. it could be bit 1, 2, 4, 8, 16, etc, it wasn't about bit position.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @vijaydhanraj)

___
*[Documentation/building.rst, line 117 at r3](https://reviewable.io/reviews/gramineproject/gramine/198#-Mo4TMcfE7HuU9AlnjCh:-MpfHeDC85jj8gLvJKcg:b-2o4fs9) ([raw file](https://github.com/gramineproject/gramine/blob/7bc59d7f0de07019adf4ec21dbf1bd1abd15e07e/Documentation/building.rst#L117)):*

OK, I will skip section 2b altogether.
</blockquote></details>

@woju Could you review this? This PR is blocked on this (your) comment.

woju (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @vijaydhanraj)

dimakuv (approved):
> Reviewed 1 of 2 files at r2, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

___
*[-- commits, line 2 at r5](https://reviewable.io/reviews/gramineproject/gramine/198#-MoGNmKXCNAXXhBWgtmc):*

We'll do it during final rebase.
</blockquote></details>

Done.

dimakuv:
Jenkins, test this please

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/198)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #207, opened 2021-11-09
[LibOS] Return from read/write early with 0 if count is zero

svenkata9:
Signed-off-by: Sankaranarayanan Venkatasubramanian <sankaranarayanan.venkatasubramanian@intel.com>

## Description of the changes 
Return from read/write early with 0 if count is zero.

## How to test this PR? 

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/207)

### DIFF
--- a/LibOS/shim/src/fs/chroot/fs.c
+++ b/LibOS/shim/src/fs/chroot/fs.c
@@ -401,6 +401,9 @@ static ssize_t chroot_read(struct shim_handle* hdl, void* buf, size_t count) {
     if (count > SSIZE_MAX)
         return -EFBIG;
 
+    if (count == 0)
+        return 0;
+
     struct shim_inode* inode = hdl->inode;
     lock(&hdl->lock);
 
@@ -437,6 +440,9 @@ static ssize_t chroot_write(struct shim_handle* hdl, const void* buf, size_t cou
 
     if (count > SSIZE_MAX)
         return -EFBIG;
+
+    if (count == 0)
+        return 0;
 
     struct shim_inode* inode = hdl->inode;
     lock(&hdl->lock);


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/207)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @svenkata9)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/207#-Mo3cavu8aTrDzvUF-t-:-Mo3cavu8aTrDzvUF-t0:bor274q) (no related file):*
@svenkata9 Why do we need this change?

From what I see, if these functions receive `count = 0`, they will call `DkStreamRead(... count = 0, ...)` and `DkStreamWrite(... count = 0, ...)`. In the end, it will be semantically equivalent.

The only difference I currently see is that we will stop earlier. But first of all, this may hide errors (maybe the application wants to test that the file is still accessible or something by abusing `read`/`write`). And second of all, there is no performance win (if the application issues `read(count=0)` then this application clearly doesn't care about performance).

---

svenkata9 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/207)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/207#-Mo3cavu8aTrDzvUF-t-:-Mo3wpCvDVmYTnMompRS:b3id1u4) (no related file):*

@svenkata9 Why do we need this change?

From what I see, if these functions receive `count = 0`, they will call `DkStreamRead(... count = 0, ...)` and `DkStreamWrite(... count = 0, ...)`. In the end, it will be semantically equivalent.

The only difference I currently see is that we will stop earlier. But first of all, this may hide errors (maybe the application wants to test that the file is still accessible or something by abusing `read`/`write`). And second of all, there is no performance win (if the application issues `read(count=0)` then this application clearly doesn't care about performance).
</blockquote></details>

Based on what I read, read and write can return 0 immediately if fd is valid, but count is zero and so it is the right thing to return from here without a need to execute other stuff (including ocalls) - the same way we handle SSIZE_MAX.

---

svenkata9 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/207)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/207#-Mo3cavu8aTrDzvUF-t-:-Mo44xYa2Ts8jONulR2X:b-a9uvv4) (no related file):*
<details><summary><i>Previously, svenkata9 (Sankaranarayanan Venkatasubramanian) wrote…</i></summary><blockquote>

Based on what I read, read and write can return 0 immediately if fd is valid, but count is zero and so it is the right thing to return from here without a need to execute other stuff (including ocalls) - the same way we handle SSIZE_MAX.
</blockquote></details>

We are returning only 0, and not an error. So, it is still a valid return value that we return (and we return after checking the fd that we have internally - it is already open - we check that in shim_do_read and do_handle_read and return EBADF is fd is invalid)). And, whether there is performance benefit or not, we are handling a case as per what is expected - we are not abnormally cutting short the execution. So, I guess, we are still doing the right thing and this can go in.

---

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/207)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @svenkata9)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/207#-Mo3cavu8aTrDzvUF-t-:-Mo4RNaaAPVOrtisNnp8:bfr2d9k) (no related file):*
<details><summary><i>Previously, svenkata9 (Sankaranarayanan Venkatasubramanian) wrote…</i></summary><blockquote>

We are returning only 0, and not an error. So, it is still a valid return value that we return (and we return after checking the fd that we have internally - it is already open - we check that in shim_do_read and do_handle_read and return EBADF is fd is invalid)). And, whether there is performance benefit or not, we are handling a case as per what is expected - we are not abnormally cutting short the execution. So, I guess, we are still doing the right thing and this can go in.
</blockquote></details>

I don't get it. Does it fix anything that didn't work before? To me it just feels redundant.

---

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/207)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/207#-Mo3cavu8aTrDzvUF-t-:-Mo4wEVR7LJKWRWnkn1j:b-ks2ixj) (no related file):*
<details><summary><i>Previously, svenkata9 (Sankaranarayanan Venkatasubramanian) wrote…</i></summary><blockquote>

It doesn't fix anything, but is doing it the right way by returning from the call at an appropriate place and without needing to process further.
</blockquote></details>

I agree with @dimakuv that this is useless and just adds not needed code.

Why do you think that `chroot_read` is appropriate place and `DkStreamRead` is not?

---

svenkata9 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/207)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski and @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/207#-Mo3cavu8aTrDzvUF-t-:-Mo6Yx_58Swy7nLAyvEu:b-dyj0sw) (no related file):*

I agree with @dimakuv that this is useless and just adds not needed code.

Why do you think that `chroot_read` is appropriate place and `DkStreamRead` is not?
</blockquote></details>

For the same reason why the max check is done in chroot_read and not in DkStreamRead.

Why do you call this useless - yes, it doesn't fix anything, but isn't it handling the case early in the flow?

---

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/207)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/207#-Mo3cavu8aTrDzvUF-t-:-Mo6k_oB3MAebB3oq7s0:btrpiew) (no related file):*
> For the same reason why the max check is done in chroot_read and not in DkStreamRead.

No, `SSIZE_MAX` is checked here because it's file read specific.

> Why do you call this useless - yes, it doesn't fix anything, but isn't it handling the case early in the flow?

Handling a case early we do not have to and do not need to handle, which makes this useless.

---

svenkata9 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/207)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski and @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/207#-Mo3cavu8aTrDzvUF-t-:-Mo6meZ25nGVPYs2uUwB:bvled0l) (no related file):*

> For the same reason why the max check is done in chroot_read and not in DkStreamRead.

No, `SSIZE_MAX` is checked here because it's file read specific.

> Why do you call this useless - yes, it doesn't fix anything, but isn't it handling the case early in the flow?

Handling a case early we do not have to and do not need to handle, which makes this useless.
</blockquote></details>

What do you mean by file read specific? The same check is there in write also: https://github.com/gramineproject/gramine/blob/852923d424575256fc0ccdaa56e50b9c4841ac31/LibOS/shim/src/fs/chroot/fs.c#L438

And, I am not sure why you say we don't need to handle. It is handled later - perhaps in the host OS, and we are handling it here early.

---

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/207)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/207#-Mo3cavu8aTrDzvUF-t-:-Mo6pOvT-fdPQMixjry8:b-ag8mly) (no related file):*
> What do you mean by file read specific? The same check is there in write also: 

Yes, it's also in write. I meant it's specific to file operations, you wouldn't find same thing in pipes implementation.

> And, I am not sure why you say we don't need to handle.

Because in principle `read(fd, buf, 0)` is not different from `read(fd, buf, 1)` and does not need special handling. We are not getting any performance, no bugs are fixed, we just get more code for no reason.

---

PR #214, opened 2021-11-10
[Pal/Linux-SGX] Fix user address range calculation

### LINKED ISSUE #215: LTP tests failing with receive_checkpoint_and_restore error after [Pal] Remove the 64MB static buffer commit
## Description of the problem

After the above mentioned commit, seeing multiple LTP tests fail with the following error:

```
[::] error: failed restoring checkpoint at vma (-13)
[::] error: Error during shim_init() in receive_checkpoint_and_restore (-13)
[P1:T1:clock_nanosleep02] error: process creation failed
```

The issue is fixed by #214

pwmarcz:
## Description of the changes 

Fixes the intermittent crashes after fork, introduced in #192. See commit message for more details.

Unfortunately LibOS allocation code still looks pretty brittle (we rely on the fact that in child process does roughly the same allocations as in parent, or at least less of them). But the crash should be gone now.

Fixes #215.

## How to test this PR? 

The problem can be seen on the `fork_and_exec` example. Normally it appears rarely, but you can adjust the memory parameters to force it:

1. `cd LibOS/shim/test/regression; make fork_and_exec.manifest`
2. Edit `fork_and_exec.manifest`: set `sgx.enclave_size = "128M"` and `loader.pal_internal_mem_size = "40M"`.
3. Rebuild: `SGX=1 make fork_and_exec.token` (you'll need to repeat this step after rebuilding Gramine)
4. Run in a loop, e.g.: `while gramine-sgx fork_and_exec; do true; done`

With the manifest modifications, the code on `master` crashes pretty reliably on first try, with the following messages:

```
[::] warning: user app tried to free an internal vma!
[::] error: failed restoring checkpoint at vma (-13)
[::] error: Error during shim_init() in receive_checkpoint_and_restore (-13)
...
[P1:T1:fork_and_exec] debug: process 1 exited with status 1
debug: DkProcessExit: Returning exit code 1
```

After this fix, I've run it in a loop for 300+ iterations and it still works.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/214)

### DIFF
--- a/Pal/src/db_streams.c
+++ b/Pal/src/db_streams.c
@@ -417,21 +417,23 @@ int DkStreamMap(PAL_HANDLE handle, PAL_PTR* addr, pal_prot_flags_t prot, PAL_NUM
     assert(addr);
     void* map_addr = *addr;
 
-    /* TODO: we must not allow NULL addresses here, but sendfile() in LibOS does it -- it must be
-     *       re-written and then this function should enforce `map_addr != NULL` */
-
     if (!handle) {
         return -PAL_ERROR_INVAL;
     }
 
-    if (map_addr && !IS_ALLOC_ALIGNED_PTR(map_addr)) {
+    if (!map_addr) {
         return -PAL_ERROR_INVAL;
     }
+
+    if (!IS_ALLOC_ALIGNED_PTR(map_addr)) {
+        return -PAL_ERROR_INVAL;
+    }
+
     if (!size || !IS_ALLOC_ALIGNED(size) || !IS_ALLOC_ALIGNED(offset)) {
         return -PAL_ERROR_INVAL;
     }
 
-    if (map_addr && _DkCheckMemoryMappable(map_addr, size)) {
+    if (_DkCheckMemoryMappable(map_addr, size)) {
         return -PAL_ERROR_DENIED;
     }
 
--- a/Pal/src/host/Linux-SGX/db_files.c
+++ b/Pal/src/host/Linux-SGX/db_files.c
@@ -416,7 +416,7 @@ static int pf_file_map(struct protected_file* pf, PAL_HANDLE handle, void** addr
               *addr, prot, offset, size);
 
     if (*addr == NULL) {
-        /* LibOS didn't provide address at which to map, can happen on sendfile() */
+        /* No address at which to map, can happen when called from `db_rtld.c` */
         allocated_enclave_pages = get_enclave_pages(/*addr=*/NULL, size, /*is_pal_internal=*/false);
         if (!allocated_enclave_pages)
             return -PAL_ERROR_NOMEM;
--- a/Pal/src/host/Linux-SGX/db_main.c
+++ b/Pal/src/host/Linux-SGX/db_main.c
@@ -47,9 +47,8 @@ void _DkGetAvailableUserAddressRange(PAL_PTR* start, PAL_PTR* end) {
     *end   = (PAL_PTR)g_pal_sec.heap_max;
 
     /* Keep some heap for internal PAL objects allocated at runtime (recall that LibOS does not keep
-     * track of PAL memory, so without this limit it could overwrite internal PAL memory). This
-     * relies on the fact that our memory management allocates memory from higher addresses to lower
-     * addresses (see also enclave_pages.c). */
+     * track of PAL memory, so without this limit it could overwrite internal PAL memory). See also
+     * `enclave_pages.c`. */
     *end = SATURATED_P_SUB(*end, g_pal_internal_mem_size, *start);
 
     if (*end <= *start) {
--- a/Pal/src/host/Linux-SGX/db_main.c
+++ b/Pal/src/host/Linux-SGX/db_main.c
@@ -44,7 +44,7 @@ const size_t g_page_size = PRESET_PAGESIZE;
 
 void _DkGetAvailableUserAddressRange(PAL_PTR* start, PAL_PTR* end) {
     *start = (PAL_PTR)g_pal_sec.heap_min;
-    *end   = (PAL_PTR)get_enclave_heap_top();
+    *end   = (PAL_PTR)g_pal_sec.heap_max;
 
     /* Keep some heap for internal PAL objects allocated at runtime (recall that LibOS does not keep
      * track of PAL memory, so without this limit it could overwrite internal PAL memory). This
--- a/Pal/src/host/Linux-SGX/enclave_pages.c
+++ b/Pal/src/host/Linux-SGX/enclave_pages.c
@@ -319,21 +319,3 @@ out:
     spinlock_unlock(&g_heap_vma_lock);
     return ret;
 }
-
-/* returns current highest available address on the enclave heap */
-void* get_enclave_heap_top(void) {
-    spinlock_lock(&g_heap_vma_lock);
-
-    void* addr = g_heap_top;
-    struct heap_vma* vma;
-    LISTP_FOR_EACH_ENTRY(vma, &g_heap_vma_list, list) {
-        if (vma->top < addr) {
-            goto out;
-        }
-        addr = vma->bottom;
-    }
-
-out:
-    spinlock_unlock(&g_heap_vma_lock);
-    return addr;
-}
--- a/Pal/src/host/Linux-SGX/enclave_pages.c
+++ b/Pal/src/host/Linux-SGX/enclave_pages.c
@@ -13,8 +13,6 @@ struct atomic_int g_allocated_pages;
 static void* g_heap_bottom;
 static void* g_heap_top;
 
-static size_t g_pal_internal_mem_used = 0;
-
 /* list of VMAs of used memory areas kept in DESCENDING order; note that preallocated PAL internal
  * memory relies on this descending order of allocations (from high addresses to low), see
  * _DkGetAvailableUserAddressRange() for more details */
@@ -86,6 +84,10 @@ static void* __create_vma_and_merge(void* addr, size_t size, bool is_pal_interna
     assert(addr && size);
 
     if (addr < g_heap_bottom)
+        return NULL;
+
+    /* PAL-internal memory cannot be allocated below the PAL-internal part of heap */
+    if (is_pal_internal && addr < g_heap_top - g_pal_internal_mem_size)
         return NULL;
 
     /* find enclosing VMAs and check that pal-internal VMAs do not overlap with normal VMAs */
@@ -171,11 +173,6 @@ static void* __create_vma_and_merge(void* addr, size_t size, bool is_pal_interna
     size_t allocated = vma->top - vma->bottom - freed;
     __atomic_add_fetch(&g_allocated_pages.counter, allocated / g_page_size, __ATOMIC_SEQ_CST);
 
-    if (is_pal_internal) {
-        assert(allocated <= g_pal_internal_mem_size - g_pal_internal_mem_used);
-        g_pal_internal_mem_used += allocated;
-    }
-
     return addr;
 }
 
@@ -194,11 +191,6 @@ void* get_enclave_pages(void* addr, size_t size, bool is_pal_internal) {
     struct heap_vma* vma;
 
     spinlock_lock(&g_heap_vma_lock);
-
-    if (is_pal_internal && size > g_pal_internal_mem_size - g_pal_internal_mem_used) {
-        /* requested PAL-internal allocation would exceed the limit, fail */
-        goto out;
-    }
 
     if (addr) {
         /* caller specified concrete address; find VMA right-above this address */
@@ -234,6 +226,17 @@ void* get_enclave_pages(void* addr, size_t size, bool is_pal_internal) {
 
 out:
     spinlock_unlock(&g_heap_vma_lock);
+
+    if (ret) {
+        if (is_pal_internal) {
+            /* This should be guaranteed by the check in `__create_vma_and_merge()` */
+            assert(ret >= g_heap_top - g_pal_internal_mem_size);
+        } else {
+            assert(ret >= g_heap_bottom);
+        }
+        assert(ret + size <= g_heap_top);
+    }
+
     return ret;
 }
 
@@ -309,11 +312,6 @@ int free_enclave_pages(void* addr, size_t size) {
     }
 
     __atomic_sub_fetch(&g_allocated_pages.counter, freed / g_page_size, __ATOMIC_SEQ_CST);
-
-    if (is_pal_internal_set && is_pal_internal) {
-        assert(g_pal_internal_mem_used >= freed);
-        g_pal_internal_mem_used -= freed;
-    }
 
 out:
     spinlock_unlock(&g_heap_vma_lock);
--- a/Pal/src/host/Linux-SGX/enclave_pages.h
+++ b/Pal/src/host/Linux-SGX/enclave_pages.h
@@ -2,6 +2,5 @@
 #include <stddef.h>
 
 void init_enclave_pages(void);
-void* get_enclave_heap_top(void);
 void* get_enclave_pages(void* addr, size_t size, bool is_pal_internal);
 int free_enclave_pages(void* addr, size_t size);


pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/214)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

___
*[Pal/src/host/Linux-SGX/db_main.c, line 56 at r1](https://reviewable.io/reviews/gramineproject/gramine/214#-Mo9igr--ggjtIGHQCH2:-Mo9igr--ggjtIGHQCH3:b-wh0ppz) ([raw file](https://github.com/gramineproject/gramine/blob/478ec63800439c0f779393523036405f2b489e33/Pal/src/host/Linux-SGX/db_main.c#L56)):*
> ```C
>      * track of PAL memory, so without this limit it could overwrite internal PAL memory). This
>      * relies on the fact that our memory management allocates memory from higher addresses to lower
>      * addresses (see also enclave_pages.c). */
> ```

@dimakuv Does that actually work? The comment seems to imply that PAL-internal allocations will always be above the new `*end`. But that is not guaranteed if the heap above `*end` becomes fragmented (i.e. we free some memory from the middle): `get_enclave_pages` looks for a gap big enough, so if it doesn't find one, it will happily allocate memory from a lower address.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/214)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/214#-Mo9oF6T7q1_7Q3wAP3t:-Mo9oF6T7q1_7Q3wAP3u:b-z1jh25) (no related file):*
We can remove `get_enclave_heap_top` function now completely.

---
___
*[Pal/src/host/Linux-SGX/db_main.c, line 51 at r1](https://reviewable.io/reviews/gramineproject/gramine/214#-Mo9o9ot4WVJLhOy_51l:-Mo9o9ot4WVJLhOy_51m:ba2fg9x) ([raw file](https://github.com/gramineproject/gramine/blob/478ec63800439c0f779393523036405f2b489e33/Pal/src/host/Linux-SGX/db_main.c#L51)):*
> ```C
> void _DkGetAvailableUserAddressRange(PAL_PTR* start, PAL_PTR* end) {
>     *start = (PAL_PTR)g_pal_sec.heap_min;
>     *end   = (PAL_PTR)g_pal_sec.heap_max;
> ```

Why does it fix the problem?..
___
*[Pal/src/host/Linux-SGX/db_main.c, line 56 at r1](https://reviewable.io/reviews/gramineproject/gramine/214#-Mo9igr--ggjtIGHQCH2:-Mo9nUtj2Qe5_maRUiwF:bd6v9ee) ([raw file](https://github.com/gramineproject/gramine/blob/478ec63800439c0f779393523036405f2b489e33/Pal/src/host/Linux-SGX/db_main.c#L56)):*

@dimakuv Does that actually work? The comment seems to imply that PAL-internal allocations will always be above the new `*end`. But that is not guaranteed if the heap above `*end` becomes fragmented (i.e. we free some memory from the middle): `get_enclave_pages` looks for a gap big enough, so if it doesn't find one, it will happily allocate memory from a lower address.
</blockquote></details>

It works, but it is very brittle. What you describe can certainly happen in the worst case. This trick is a stop-gap solution (a hack). At some point, we need to re-write this logic.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/214)* status: 1 of 3 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/214#-Mo9oF6T7q1_7Q3wAP3t:-Mo9pqFC66p1ItjPSNjR:b-896fix) (no related file):*

We can remove `get_enclave_heap_top` function now completely.
</blockquote></details>

Done.

---
___
*[Pal/src/host/Linux-SGX/db_main.c, line 51 at r1](https://reviewable.io/reviews/gramineproject/gramine/214#-Mo9o9ot4WVJLhOy_51l:-Mo9pAVR56rH-OBCFU0T:bvltyon) ([raw file](https://github.com/gramineproject/gramine/blob/478ec63800439c0f779393523036405f2b489e33/Pal/src/host/Linux-SGX/db_main.c#L51)):*

Why does it fix the problem?..
</blockquote></details>

Because `heap_max` will be the same for parent and child processes, and `get_enclave_heap_top()` is not guaranteed to. If `user_address.end` is different for parent and child, that means parent LibOS and child LibOS start allocating memory from a different address, so a collision (between a parent "user" allocation and a child "system" allocation) is more likely.
___
*[Pal/src/host/Linux-SGX/db_main.c, line 56 at r1](https://reviewable.io/reviews/gramineproject/gramine/214#-Mo9igr--ggjtIGHQCH2:-Mo9phdrBBzSBy513XPc:b-ykhxfq) ([raw file](https://github.com/gramineproject/gramine/blob/478ec63800439c0f779393523036405f2b489e33/Pal/src/host/Linux-SGX/db_main.c#L56)):*

It works, but it is very brittle. What you describe can certainly happen in the worst case. This trick is a stop-gap solution (a hack). At some point, we need to re-write this logic.
</blockquote></details>

OK. That wasn't clear to me, I thought it was guaranteed.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/214)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

___
*[Pal/src/host/Linux-SGX/db_main.c, line 51 at r1](https://reviewable.io/reviews/gramineproject/gramine/214#-Mo9o9ot4WVJLhOy_51l:-Mo9ufvl3JcA2YdhiM0K:bvypyh8) ([raw file](https://github.com/gramineproject/gramine/blob/478ec63800439c0f779393523036405f2b489e33/Pal/src/host/Linux-SGX/db_main.c#L51)):*

Because `heap_max` will be the same for parent and child processes, and `get_enclave_heap_top()` is not guaranteed to. If `user_address.end` is different for parent and child, that means parent LibOS and child LibOS start allocating memory from a different address, so a collision (between a parent "user" allocation and a child "system" allocation) is more likely.
</blockquote></details>

OK, unblocking. I am still unsure we're not just shifting the problem further, but let's discuss offline.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/214)* status: 1 of 5 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski and @dimakuv)

___
*[Pal/src/db_streams.c, line 427 at r3](https://reviewable.io/reviews/gramineproject/gramine/214#-MoAF6ki9UzywdnUmw1p:-MoAF6ki9UzywdnUmw1q:bkf3f1y) ([raw file](https://github.com/gramineproject/gramine/blob/1c764a2035d4a89893de5f8ba8543c5523584c3a/Pal/src/db_streams.c#L427)):*
> ```C
>     void* map_addr = *addr;
> 
>     /* TODO: we must not allow NULL addresses here, but sendfile() in LibOS does it -- it must be
> ```

FYI: `sendfile()` is rewritten, but the new `db_rtld.c` also does this (although it uses the internal `_DkStreamMap` function).
___
*[Pal/src/host/Linux-SGX/db_main.c, line 51 at r1](https://reviewable.io/reviews/gramineproject/gramine/214#-Mo9o9ot4WVJLhOy_51l:-MoAExDdB2DB8y6H40dL:bd9c7o1) ([raw file](https://github.com/gramineproject/gramine/blob/478ec63800439c0f779393523036405f2b489e33/Pal/src/host/Linux-SGX/db_main.c#L51)):*

OK, unblocking. I am still unsure we're not just shifting the problem further, but let's discuss offline.
</blockquote></details>

As discussed offline: It would be good to ensure that PAL-internal memory does not go below this address). I added a commit that should enforce that. If it turns out to be too problematic, we can split it out to a separate PR.

boryspoplawski (changes requested):
> Reviewed 1 of 2 files at r2, 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/214)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

___
*[Pal/src/host/Linux-SGX/enclave_pages.c, line 194 at r3](https://reviewable.io/reviews/gramineproject/gramine/214#-MoBIe1p3dxJ3uGHH1b5:-MoBIe1p3dxJ3uGHH1b6:b-y695p3) ([raw file](https://github.com/gramineproject/gramine/blob/3ab19384adddd56068a43369ca132a933f26a1c4/Pal/src/host/Linux-SGX/enclave_pages.c#L194)):*
> ```C
>     if (is_pal_internal && (size > g_pal_internal_mem_size
>                             || size > (size_t)(g_heap_top - g_heap_bottom))) {
>         /* requested PAL-internal allocation would not fit in PAL-internal area, fail */
> ```

In what case `g_pal_internal_mem_size > g_heap_top - g_heap_bottom`? Shouldn't this always be false (i.e. PAL initialization would fail if it isn't)?

aniket-intelx:
This PR fixes the issue which was seen for LTP test suite and sandstone.

dimakuv (changes requested):
> Reviewed 3 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/214)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski and @pwmarcz)

___
*[Pal/src/host/Linux-SGX/enclave_pages.c, line 194 at r3](https://reviewable.io/reviews/gramineproject/gramine/214#-MoBIe1p3dxJ3uGHH1b5:-MoD33qg--YzDKPth-UQ:b6hk1o5) ([raw file](https://github.com/gramineproject/gramine/blob/3ab19384adddd56068a43369ca132a933f26a1c4/Pal/src/host/Linux-SGX/enclave_pages.c#L194)):*

In what case `g_pal_internal_mem_size > g_heap_top - g_heap_bottom`? Shouldn't this always be false (i.e. PAL initialization would fail if it isn't)?
</blockquote></details>

+1 to Borys's question. When we get to the second OR condition, we have:
```
(g_heap_top - g_heap_bottom) < size <= g_pal_internal_mem_size
```

In other words, `heap_size < g_pal_internal_mem_size`. But this is impossible.
___
*[Pal/src/host/Linux-SGX/enclave_pages.c, line 195 at r3](https://reviewable.io/reviews/gramineproject/gramine/214#-MoD7MYBEEgjybCowfTf:-MoD7MYBEEgjybCowfTg:b6pvpxx) ([raw file](https://github.com/gramineproject/gramine/blob/3ab19384adddd56068a43369ca132a933f26a1c4/Pal/src/host/Linux-SGX/enclave_pages.c#L195)):*
> ```C
>                             || size > (size_t)(g_heap_top - g_heap_bottom))) {
>         /* requested PAL-internal allocation would not fit in PAL-internal area, fail */
>         return NULL;
> ```

I think you can remove this check completely. See my other comment.
___
*[Pal/src/host/Linux-SGX/enclave_pages.c, line 232 at r3](https://reviewable.io/reviews/gramineproject/gramine/214#-MoD6jNp9HPwq2RXR7sF:-MoD6jNp9HPwq2RXR7sG:b-j58o6b) ([raw file](https://github.com/gramineproject/gramine/blob/3ab19384adddd56068a43369ca132a933f26a1c4/Pal/src/host/Linux-SGX/enclave_pages.c#L232)):*
> ```C
>                 /* we would have to go below PAL-internal area, fail */
>                 goto out;
>             }
> ```

Actually, why so complex? Why not just add something like this to `__create_vma_and_merge()`
```
static void* __create_vma_and_merge(void* addr, size_t size, bool is_pal_internal,
                                    struct heap_vma* vma_above) {
    ...
    if (is_pal_internal && addr < pal_internal_mem_bottom) {
        /* we would have to go below PAL-internal area, fail */
        return NULL;
    }
```

And then you can delete *both* checks that you introduced in this func.
___
*[Pal/src/host/Linux-SGX/enclave_pages.c, line 250 at r3](https://reviewable.io/reviews/gramineproject/gramine/214#-MoD7aOa6rBkeGrH-wg0:-MoD7aOa6rBkeGrH-wg1:b68i7zq) ([raw file](https://github.com/gramineproject/gramine/blob/3ab19384adddd56068a43369ca132a933f26a1c4/Pal/src/host/Linux-SGX/enclave_pages.c#L250)):*
> ```C
>         }
>         assert(ret + size <= g_heap_top);
>     }
> ```

FYI: I love these asserts! Thanks.

dimakuv (commented):
> Reviewed 1 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/214)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski and @pwmarcz)

pwmarcz (commented):
Great, thanks for checking!

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/214#-:-MoENYzA--W8gYSwb-Bz:bpbx239)* status: 4 of 5 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[Pal/src/host/Linux-SGX/enclave_pages.c, line 194 at r3](https://reviewable.io/reviews/gramineproject/gramine/214#-MoBIe1p3dxJ3uGHH1b5:-MoEJwEm9FHZd8RiiSiB:b-wcmhn1) ([raw file](https://github.com/gramineproject/gramine/blob/3ab19384adddd56068a43369ca132a933f26a1c4/Pal/src/host/Linux-SGX/enclave_pages.c#L194)):*

+1 to Borys's question. When we get to the second OR condition, we have:
```
(g_heap_top - g_heap_bottom) < size <= g_pal_internal_mem_size
```

In other words, `heap_size < g_pal_internal_mem_size`. But this is impossible.
</blockquote></details>

You're right. I got confused by the code in `_DkGetAvailableUserAddressRange`, but it does guarantee that `g_pal_internal_mem_size` is smaller than heap.

Check replaced as you suggested.
___
*[Pal/src/host/Linux-SGX/enclave_pages.c, line 195 at r3](https://reviewable.io/reviews/gramineproject/gramine/214#-MoD7MYBEEgjybCowfTf:-MoENMD910ip9h7QU1H1:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/3ab19384adddd56068a43369ca132a933f26a1c4/Pal/src/host/Linux-SGX/enclave_pages.c#L195)):*

I think you can remove this check completely. See my other comment.
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/enclave_pages.c, line 232 at r3](https://reviewable.io/reviews/gramineproject/gramine/214#-MoD6jNp9HPwq2RXR7sF:-MoENNDM05znFa468cHV:bcvjgzp) ([raw file](https://github.com/gramineproject/gramine/blob/3ab19384adddd56068a43369ca132a933f26a1c4/Pal/src/host/Linux-SGX/enclave_pages.c#L232)):*

Actually, why so complex? Why not just add something like this to `__create_vma_and_merge()`
```
static void* __create_vma_and_merge(void* addr, size_t size, bool is_pal_internal,
                                    struct heap_vma* vma_above) {
    ...
    if (is_pal_internal && addr < pal_internal_mem_bottom) {
        /* we would have to go below PAL-internal area, fail */
        return NULL;
    }
```

And then you can delete *both* checks that you introduced in this func.
</blockquote></details>

Good idea. Done.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/214)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/214#-MoERQ4V3wM61fuy5Wsa:-MoERQ4V3wM61fuy5Wsb:bseruxq) (no related file):*
I really like how this PR ended up. Hopefully we now won't have any memory corruptions due to PAL memory areas overlapping with LibOS memory areas!

---

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/214)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[Pal/src/host/Linux-SGX/enclave_pages.c, line 202 at r3](https://reviewable.io/reviews/gramineproject/gramine/214#-MoEjOa_7YClcL78k2wh:-MoEjOaaEDcPbPZUxT4O:bja6p0k) ([raw file](https://github.com/gramineproject/gramine/blob/3ab19384adddd56068a43369ca132a933f26a1c4/Pal/src/host/Linux-SGX/enclave_pages.c#L202)):*
> ```C
>     if (addr) {
>         /* we never request PAL-internal memory with specified address */
>         assert(!is_pal_internal);
> ```

Why did you remove this?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/214)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[Pal/src/host/Linux-SGX/enclave_pages.c, line 202 at r3](https://reviewable.io/reviews/gramineproject/gramine/214#-MoEjOa_7YClcL78k2wh:-MoF2Rdg9rW13sDUSvh0:b7a5m6n) ([raw file](https://github.com/gramineproject/gramine/blob/3ab19384adddd56068a43369ca132a933f26a1c4/Pal/src/host/Linux-SGX/enclave_pages.c#L202)):*

Why did you remove this?
</blockquote></details>

I added it before to ensure the PAL-internal address is correct (we assert `!is_pal_internal` in this branch, so that we can check the address only in the other branch). However, in the updated code the check is another function (`__create_vma_and_merge()`), so there's no need to know anything about `is_pal_internal` here.

And while this specific condition would still be true, it seems a little... accidental with the current messy design, so I don't want to assert it if there is no need to.

boryspoplawski (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/214)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

___
*[Pal/src/host/Linux-SGX/enclave_pages.c, line 202 at r3](https://reviewable.io/reviews/gramineproject/gramine/214#-MoEjOa_7YClcL78k2wh:-MoFhGUI2jTQMUrnqAWE:bjhk9z) ([raw file](https://github.com/gramineproject/gramine/blob/3ab19384adddd56068a43369ca132a933f26a1c4/Pal/src/host/Linux-SGX/enclave_pages.c#L202)):*

I added it before to ensure the PAL-internal address is correct (we assert `!is_pal_internal` in this branch, so that we can check the address only in the other branch). However, in the updated code the check is another function (`__create_vma_and_merge()`), so there's no need to know anything about `is_pal_internal` here.

And while this specific condition would still be true, it seems a little... accidental with the current messy design, so I don't want to assert it if there is no need to.
</blockquote></details>

Okey, but I think we have a higher level assumption, that PAL internal allocations never request a specific address and all other allocations must specify address, so I wouldnt call this accidental. Anyway, resolving.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/214)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

___
*[Pal/src/host/Linux-SGX/enclave_pages.c, line 202 at r3](https://reviewable.io/reviews/gramineproject/gramine/214#-MoEjOa_7YClcL78k2wh:-MoHtiMR0D-NL_rS7bI6:bqisv6l) ([raw file](https://github.com/gramineproject/gramine/blob/3ab19384adddd56068a43369ca132a933f26a1c4/Pal/src/host/Linux-SGX/enclave_pages.c#L202)):*

Okey, but I think we have a higher level assumption, that PAL internal allocations never request a specific address and all other allocations must specify address, so I wouldnt call this accidental. Anyway, resolving.
</blockquote></details>

There was an assumption, it was broken for a long time because of `sendfile`, now it's broken because of new ELF loading code in PAL (which uses both kinds of allocations: with and without address, and doesn't do PAL-internal allocations because the "map file" operations don't know about these).

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/214)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

___
*[Pal/src/host/Linux-SGX/enclave_pages.c, line 202 at r3](https://reviewable.io/reviews/gramineproject/gramine/214#-MoEjOa_7YClcL78k2wh:-MoI24_PCauJix0y6J2b:b3uy1gp) ([raw file](https://github.com/gramineproject/gramine/blob/3ab19384adddd56068a43369ca132a933f26a1c4/Pal/src/host/Linux-SGX/enclave_pages.c#L202)):*

There was an assumption, it was broken for a long time because of `sendfile`, now it's broken because of new ELF loading code in PAL (which uses both kinds of allocations: with and without address, and doesn't do PAL-internal allocations because the "map file" operations don't know about these).
</blockquote></details>

Actually, this sounds like an unintended issue in the new ELF loading code. We may want to fix it at some point.

On the other hand, we have a clear separation between PAL-internal memory space (`g_heap_top - g_pal_internal_mem_size .. g_heap_top`) and LibOS memory space (all the rest memory), so the assumption "PAL internal allocs never request a specific address, whereas LibOS allocs always request an address" is kinda arbitrary. The first part of this assumption (PAL never allocs with a fixed address) is unnecessary -- PAL should be able to allocate at a chosen fixed address in its memory space. The second part seems (LibOS always allocs with a fixed address) logical for now -- but I have plans to allow for "LibOS allocs in untrusted shared memory space", and in this case LibOS should outsource all decisions to the host allocator and *not* specify a fixed address.

TLDR: I think we shouldn't have this assumption at all.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/214)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

mkow (approved):
> Reviewed 1 of 2 files at r2, 1 of 1 files at r4, 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/214)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #217, opened 2021-11-12
[Pal/Linux-SGX] Add a message on `sgx.debug = true` on GDB attach

dimakuv:
## Description of the changes 

I forgot to add `sgx.debug = true` in the manifest file before starting my GDB sessions on a production-ready app. This made my GDB hang, and I spent some time figuring out what's going on. Then I remembered that `sgx.debug` is false by default (this was flipped starting from Gramine v1.0).

## How to test this PR? 

Run GDB.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/217)

### DIFF
--- a/Pal/src/host/Linux-SGX/gdb_integration/gramine_sgx_gdb.py
+++ b/Pal/src/host/Linux-SGX/gdb_integration/gramine_sgx_gdb.py
@@ -20,6 +20,9 @@ def main():
         path = os.path.join(os.path.dirname(__file__), filename)
         gdb.execute('source ' + path)
 
+    print('[%s] For SGX enclave debugging, don\'t forget to specify `sgx.debug=true`.' %
+              os.path.basename(__file__))
+
 
 if __name__ == '__main__':
     main()


mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/217)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[Pal/src/host/Linux-SGX/gdb_integration/gramine_sgx_gdb.py, line 24 at r1](https://reviewable.io/reviews/gramineproject/gramine/217#-MoJ62dVCvt45mCn26bM:-MoJ62dVCvt45mCn26bN:b-eucdmk) ([raw file](https://github.com/gramineproject/gramine/blob/d2d7cae754f890abb0112e3529d031c162c4d22d/Pal/src/host/Linux-SGX/gdb_integration/gramine_sgx_gdb.py#L24)):*
> ```Python
> 
>     print('[%s] For SGX enclave debugging, don\'t forget to specify `sgx.debug=true`.' %
>               os.path.basename(__file__))
> ```

Isn't there are viable way to print this only when it's actually missing? I don't like printing log messages when they aren't useful :)
CC: @pwmarcz

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/217)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @pwmarcz)

___
*[Pal/src/host/Linux-SGX/gdb_integration/gramine_sgx_gdb.py, line 24 at r1](https://reviewable.io/reviews/gramineproject/gramine/217#-MoJ62dVCvt45mCn26bM:-MoJ6YHM0Grnr_R8ws7T:b7w2059) ([raw file](https://github.com/gramineproject/gramine/blob/d2d7cae754f890abb0112e3529d031c162c4d22d/Pal/src/host/Linux-SGX/gdb_integration/gramine_sgx_gdb.py#L24)):*

Isn't there are viable way to print this only when it's actually missing? I don't like printing log messages when they aren't useful :)
CC: @pwmarcz
</blockquote></details>

When I talked with Pawel, he didn't have a good way to check it.

Also, this is our GDB message, only hardcore devs will ever see this message.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/217)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @pwmarcz)

___
*[Pal/src/host/Linux-SGX/gdb_integration/gramine_sgx_gdb.py, line 24 at r1](https://reviewable.io/reviews/gramineproject/gramine/217#-MoJ62dVCvt45mCn26bM:-MoJ8wEu3lEAadKSxcfl:byfrbiy) ([raw file](https://github.com/gramineproject/gramine/blob/d2d7cae754f890abb0112e3529d031c162c4d22d/Pal/src/host/Linux-SGX/gdb_integration/gramine_sgx_gdb.py#L24)):*
> only hardcore devs will ever see this message

I don't like seeing a wall of text each time I start something :) I quickly default to not reading it at all, so I'll most likely not read this warning in case of forgetting this option, because I'll be used to gdb printing it regardless of what I did ;)

Where is it failing/inf looping in case of the problems? Maybe we could just handle the error properly and print this message there?

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/217)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @pwmarcz)

___
*[Pal/src/host/Linux-SGX/gdb_integration/gramine_sgx_gdb.py, line 24 at r1](https://reviewable.io/reviews/gramineproject/gramine/217#-MoJ62dVCvt45mCn26bM:-MoJhO2o8SJJJaRG8t2v:blcar7f) ([raw file](https://github.com/gramineproject/gramine/blob/d2d7cae754f890abb0112e3529d031c162c4d22d/Pal/src/host/Linux-SGX/gdb_integration/gramine_sgx_gdb.py#L24)):*

> only hardcore devs will ever see this message

I don't like seeing a wall of text each time I start something :) I quickly default to not reading it at all, so I'll most likely not read this warning in case of forgetting this option, because I'll be used to gdb printing it regardless of what I did ;)

Where is it failing/inf looping in case of the problems? Maybe we could just handle the error properly and print this message there?
</blockquote></details>

+1 I would prefer this to be printed only on actual errors

dimakuv:
Let me close it for now. Maybe I'll return back to this when I have more time to debug what exactly happens.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/217)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski and @mkow)

___
*[Pal/src/host/Linux-SGX/gdb_integration/gramine_sgx_gdb.py, line 24 at r1](https://reviewable.io/reviews/gramineproject/gramine/217#-MoJ62dVCvt45mCn26bM:-MoJsRdMBnqsVTd2Yd5f:b-dim8ao) ([raw file](https://github.com/gramineproject/gramine/blob/d2d7cae754f890abb0112e3529d031c162c4d22d/Pal/src/host/Linux-SGX/gdb_integration/gramine_sgx_gdb.py#L24)):*

+1 I would prefer this to be printed only on actual errors
</blockquote></details>

I think I found a simple way.

In the SGX GDB integration, the application keeps an [`enclave_dbginfo`](https://github.com/gramineproject/gramine/blob/master/Pal/src/host/Linux-SGX/gdb_integration/sgx_gdb.h) struct, in untrusted memory, at a predefined location. On attaching to a new process, `sgx_gdb.so`  [reads this structure](https://github.com/gramineproject/gramine/blob/master/Pal/src/host/Linux-SGX/gdb_integration/sgx_gdb.c#L395).

So there can be a `bool is_debug_enclave` in that structure. On reading it, `sgx_gdb.so` should complain loudly that this is not a debug enclave and that we're unable to look inside.

PR #219, opened 2021-11-12
[Pal/Linux-SGX] Restructure sgx_quote_t to split signature from body

### LINKED ISSUE #113: [Pal/Linux-SGX] `sgx_quote_t` has more fields than actually returned from IAS
## Description of the problem

`sgx_quote_t` is declared like this: https://github.com/gramineproject/gramine/blob/9a4fd153b489dbfcad1112a6aae47c35f3503c5a/Pal/src/host/Linux-SGX/sgx_attest.h#L47-L58

The last two fields (`signature_len` and `signature`) are "optional". Indeed, the official IAS specification `IAS-API-Spec-rev-5.0.pdf` tells us:

> isvEnclaveQuoteBody String Base 64-encoded BODY of QUOTE structure (i.e., QUOTE structure without signature related fields: SIG_LEN and SIG) as received in Attestation Evidence Payload. See Quoting Data Structures for details.

This is why we have a convinience macro like this (to calculate the size of the SGX quote returned in the IAS attestation report): https://github.com/gramineproject/gramine/blob/9a4fd153b489dbfcad1112a6aae47c35f3503c5a/Pal/src/host/Linux-SGX/sgx_attest.h#L60

This is a pretty bad design, because we use this type `sgx_quote_t` everywhere in our verification functions. For example:
- https://github.com/gramineproject/gramine/blob/9a4fd153b489dbfcad1112a6aae47c35f3503c5a/Pal/src/host/Linux-SGX/tools/common/attestation.c#L343
- https://github.com/gramineproject/gramine/blob/9a4fd153b489dbfcad1112a6aae47c35f3503c5a/Pal/src/host/Linux-SGX/tools/ra-tls/ra_tls_verify_common.c#L203
- (well, these examples will be clearer once #100 is merged; currently the conversion to `sgx_quote_t` happens inside these funcs)

This may blow up spectacularly in the future, if any of these verification functions will decide to use e.g. `sgx_quote_t::signature_len` (buffer overflow).

So we must split `sgx_quote_t` into two parts: `sgx_quote_body_t` and `sgx_quote_sig_t`. And only pass `sgx_quote_body_t` to verification functions.

See more discussions in #100.

vijaydhanraj:
## Description of the changes 
SGX quote returned by IAS does not contain signature and signature length fields. This might result in a buffer overflow if we accidentally end up using these fields as part of the quote. To avoid such an issue, this commit splits sgx_quote structure into sgx_quote_body and signature.

Also, modified `attestation` LibOS regression to incorporate the new sgx_quote structural change.

Signed-off-by: Vijay Dhanraj <vijay.dhanraj@intel.com>

Fixes #113 

## How to test this PR? 
Please run LibOS regression tests. One can also run the `ra-tls-mbedtls` test as part of our CI-Examples.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/219)

### DIFF
--- a/LibOS/shim/test/regression/attestation.c
+++ b/LibOS/shim/test/regression/attestation.c
@@ -226,20 +226,20 @@ static int test_quote_interface(void) {
     }
 
     /* 3. verify report data read from `quote` */
-    if ((size_t)bytes < sizeof(sgx_quote_t)) {
+    if ((size_t)bytes < sizeof(sgx_quote_body_t)) {
         fprintf(stderr, "obtained SGX quote is too small: %ldB (must be at least %ldB)\n", bytes,
-                sizeof(sgx_quote_t));
+                sizeof(sgx_quote_body_t));
         return FAILURE;
     }
 
-    sgx_quote_t* typed_quote = (sgx_quote_t*)g_quote;
+    sgx_quote_body_t* quote_body = (sgx_quote_body_t*)g_quote;
 
-    if (typed_quote->version != /*EPID*/2 && typed_quote->version != /*DCAP*/3) {
+    if (quote_body->version != /*EPID*/2 && quote_body->version != /*DCAP*/3) {
         fprintf(stderr, "version of SGX quote is not EPID (2) and not ECDSA/DCAP (3)\n");
         return FAILURE;
     }
 
-    int ret = memcmp(typed_quote->report_body.report_data.d, user_report_data.d,
+    int ret = memcmp(quote_body->report_body.report_data.d, user_report_data.d,
                      sizeof(user_report_data));
     if (ret) {
         fprintf(stderr, "comparison of report data in SGX quote failed\n");
--- a/Pal/src/host/Linux-SGX/sgx_attest.h
+++ b/Pal/src/host/Linux-SGX/sgx_attest.h
@@ -44,11 +44,7 @@ typedef struct _sgx_basename_t {
     uint8_t name[32];
 } sgx_basename_t;
 
-/* TODO: IAS returns an IAS report object that contains a truncated SGX quote inside, without
- * `signature_len` and `signature` fields. This is called "SGX quote body". We must split the
- * current `sgx_quote_t` struct into two structs, and use the "SGX quote body" in all relevant
- * functions instead of the full `sgx_quote_t` (otherwise it will blow up someday). */
-typedef struct _sgx_quote_t {
+typedef struct _sgx_quote_body_t {
     uint16_t version;
     uint16_t sign_type;
     sgx_epid_group_id_t epid_group_id;
@@ -57,11 +53,13 @@ typedef struct _sgx_quote_t {
     uint32_t xeid;
     sgx_basename_t basename;
     sgx_report_body_t report_body;
-    uint32_t signature_len;
+} sgx_quote_body_t;
+
+typedef struct _sgx_quote_t {
+    sgx_quote_body_t body;
+    uint32_t signature_size;
     uint8_t signature[];
 } sgx_quote_t;
-
-#define SGX_QUOTE_BODY_SIZE (offsetof(sgx_quote_t, signature_len))
 
 typedef uint8_t sgx_spid_t[16];
 typedef uint8_t sgx_quote_nonce_t[16];
--- a/Pal/src/host/Linux-SGX/sgx_platform.c
+++ b/Pal/src/host/Linux-SGX/sgx_platform.c
@@ -305,7 +305,7 @@ int retrieve_quote(const sgx_spid_t* spid, bool linkable, const sgx_report_t* re
     /* Intel SGX SDK implementation of the Quoting Enclave always sets `quote.len` to user-provided
      * `getreq.buf_size` (see above) instead of the actual size. We calculate the actual size here
      * by peeking into the quote and determining the size of the signature. */
-    size_t actual_quote_size = sizeof(sgx_quote_t) + actual_quote->signature_len;
+    size_t actual_quote_size = sizeof(sgx_quote_t) + actual_quote->signature_size;
     if (actual_quote_size > SGX_QUOTE_MAX_SIZE) {
         log_error("Size of the obtained SGX quote exceeds %d", SGX_QUOTE_MAX_SIZE);
         goto out;
--- a/Pal/src/host/Linux-SGX/tools/common/attestation.c
+++ b/Pal/src/host/Linux-SGX/tools/common/attestation.c
@@ -46,7 +46,7 @@ const char* g_ias_public_key_pem =
 #define IS_ALIGNED_PTR_POW2(val, alignment) IS_ALIGNED_POW2((uintptr_t)(val), alignment)
 
 // TODO: decode some known values (flags etc)
-void display_report_body(const sgx_report_body_t* body) {
+static void display_report_body(const sgx_report_body_t* body) {
     INFO(" cpu_svn          : ");
     HEXDUMP(body->cpu_svn);
     INFO(" misc_select      : ");
@@ -83,39 +83,45 @@ void display_report_body(const sgx_report_body_t* body) {
     HEXDUMP(body->report_data);
 }
 
+static void display_quote_body(const sgx_quote_body_t* quote_body) {
+    INFO(" version          : ");
+    HEXDUMP(quote_body->version);
+    INFO(" sign_type        : ");
+    HEXDUMP(quote_body->sign_type);
+    INFO(" epid_group_id    : ");
+    HEXDUMP(quote_body->epid_group_id);
+    INFO(" qe_svn           : ");
+    HEXDUMP(quote_body->qe_svn);
+    INFO(" pce_svn          : ");
+    HEXDUMP(quote_body->pce_svn);
+    INFO(" xeid             : ");
+    HEXDUMP(quote_body->xeid);
+    INFO(" basename         : ");
+    HEXDUMP(quote_body->basename);
+}
+
 void display_quote(const void* quote_data, size_t quote_size) {
-    if (quote_size < SGX_QUOTE_BODY_SIZE) {
+    if (quote_size < sizeof(sgx_quote_body_t)) {
         ERROR("Quote size too small\n");
         return;
     }
 
     assert(IS_ALIGNED_PTR_POW2(quote_data, alignof(sgx_quote_t)));
     sgx_quote_t* quote = (sgx_quote_t*)quote_data;
-    INFO("version           : ");
-    HEXDUMP(quote->version);
-    INFO("sign_type         : ");
-    HEXDUMP(quote->sign_type);
-    INFO("epid_group_id     : ");
-    HEXDUMP(quote->epid_group_id);
-    INFO("qe_svn            : ");
-    HEXDUMP(quote->qe_svn);
-    INFO("pce_svn           : ");
-    HEXDUMP(quote->pce_svn);
-    INFO("xeid              : ");
-    HEXDUMP(quote->xeid);
-    INFO("basename          : ");
-    HEXDUMP(quote->basename);
+    INFO("quote_body        :\n");
+    display_quote_body(&quote->body);
     INFO("report_body       :\n");
-    display_report_body(&quote->report_body);
+    display_report_body(&quote->body.report_body);
 
-    // quotes from IAS reports are missing signature fields
-    if (quote_size >= sizeof(sgx_quote_t)) {
-        INFO("signature_len     : %d (0x%x)\n", quote->signature_len, quote->signature_len);
+    /* Quotes from IAS reports are missing signature fields. So display signature and signature_size
+       fields only for DCAP-based quotes */
+    if (quote_size >= sizeof(sgx_quote_body_t) + sizeof(quote->signature_size)) {
+        INFO("signature_size    : %d (0x%x)\n", quote->signature_size, quote->signature_size);
     }
 
-    if (quote_size >= sizeof(sgx_quote_t) + quote->signature_len) {
+    if (quote_size >= sizeof(sgx_quote_t) + quote->signature_size) {
         INFO("signature         : ");
-        hexdump_mem(&quote->signature, quote->signature_len);
+        hexdump_mem(&quote->signature, quote->signature_size);
         INFO("\n");
     }
 }
@@ -315,25 +321,17 @@ out:
     return ret ? -1 : 0;
 }
 
-int verify_quote(const void* quote_data, size_t quote_size, const char* mr_signer,
-                 const char* mr_enclave, const char* isv_prod_id, const char* isv_svn,
-                 const char* report_data, bool expected_as_str) {
+int verify_quote_body(const sgx_quote_body_t* quote_body, const char* mr_signer,
+                      const char* mr_enclave, const char* isv_prod_id, const char* isv_svn,
+                      const char* report_data, bool expected_as_str) {
     int ret = -1;
 
-    assert(IS_ALIGNED_PTR_POW2(quote_data, alignof(sgx_quote_t)));
-    sgx_quote_t* quote = (sgx_quote_t*)quote_data;
-
-    // Quote contained in the IAS report doesn't contain signature_len and signature fields
-    // Reject any smaller quotes as invalid
-    if (quote_size < SGX_QUOTE_BODY_SIZE) {
-        ERROR("Quote: Bad size %zu < %zu\n", quote_size, SGX_QUOTE_BODY_SIZE);
-        goto out;
-    }
+    sgx_quote_body_t* body = (sgx_quote_body_t*)quote_body;
 
     if (get_verbose())
-        display_quote(quote_data, quote_size);
+        display_quote_body(body);
 
-    sgx_report_body_t* body = &quote->report_body;
+    sgx_report_body_t* report_body = &body->report_body;
 
     sgx_measurement_t expected_mr;
     if (mr_signer) {
@@ -344,11 +342,11 @@ int verify_quote(const void* quote_data, size_t quote_size, const char* mr_signe
             memcpy(&expected_mr, mr_signer, sizeof(expected_mr));
         }
 
-        if (memcmp(&body->mr_signer, &expected_mr, sizeof(expected_mr)) != 0) {
+        if (memcmp(&report_body->mr_signer, &expected_mr, sizeof(expected_mr)) != 0) {
             ERROR("Quote: mr_signer doesn't match the expected value\n");
             if (get_verbose()) {
                 ERROR("Quote mr_signer:\n");
-                HEXDUMP(body->mr_signer);
+                HEXDUMP(report_body->mr_signer);
                 ERROR("Expected mr_signer:\n");
                 HEXDUMP(expected_mr);
             }
@@ -366,11 +364,11 @@ int verify_quote(const void* quote_data, size_t quote_size, const char* mr_signe
             memcpy(&expected_mr, mr_enclave, sizeof(expected_mr));
         }
 
-        if (memcmp(&body->mr_enclave, &expected_mr, sizeof(expected_mr)) != 0) {
+        if (memcmp(&report_body->mr_enclave, &expected_mr, sizeof(expected_mr)) != 0) {
             ERROR("Quote: mr_enclave doesn't match the expected value\n");
             if (get_verbose()) {
                 ERROR("Quote mr_enclave:\n");
-                HEXDUMP(body->mr_enclave);
+                HEXDUMP(report_body->mr_enclave);
                 ERROR("Expected mr_enclave:\n");
                 HEXDUMP(expected_mr);
             }
@@ -390,8 +388,9 @@ int verify_quote(const void* quote_data, size_t quote_size, const char* mr_signe
             memcpy(&prod_id, isv_prod_id, sizeof(prod_id));
         }
 
-        if (body->isv_prod_id != prod_id) {
-            ERROR("Quote: invalid isv_prod_id (%u, expected %u)\n", body->isv_prod_id, prod_id);
+        if (report_body->isv_prod_id != prod_id) {
+            ERROR("Quote: invalid isv_prod_id (%u, expected %u)\n", report_body->isv_prod_id,
+                  prod_id);
             goto out;
         }
 
@@ -407,8 +406,8 @@ int verify_quote(const void* quote_data, size_t quote_size, const char* mr_signe
             memcpy(&svn, isv_svn, sizeof(svn));
         }
 
-        if (body->isv_svn < svn) {
-            ERROR("Quote: invalid isv_svn (%u < expected %u)\n", body->isv_svn, svn);
+        if (report_body->isv_svn < svn) {
+            ERROR("Quote: invalid isv_svn (%u < expected %u)\n", report_body->isv_svn, svn);
             goto out;
         }
 
@@ -425,11 +424,11 @@ int verify_quote(const void* quote_data, size_t quote_size, const char* mr_signe
             memcpy(&rd, report_data, sizeof(rd));
         }
 
-        if (memcmp(&body->report_data, &rd, sizeof(rd)) != 0) {
+        if (memcmp(&report_body->report_data, &rd, sizeof(rd)) != 0) {
             ERROR("Quote: report_data doesn't match the expected value\n");
             if (get_verbose()) {
                 ERROR("Quote report_data:\n");
-                HEXDUMP(body->report_data);
+                HEXDUMP(report_body->report_data);
                 ERROR("Expected report_data:\n");
                 HEXDUMP(rd);
             }
@@ -445,27 +444,27 @@ out:
     return ret;
 }
 
-int verify_quote_enclave_attributes(sgx_quote_t* quote, bool allow_debug_enclave) {
-    if (!allow_debug_enclave && (quote->report_body.attributes.flags & SGX_FLAGS_DEBUG)) {
+int verify_quote_body_enclave_attributes(sgx_quote_body_t* quote_body, bool allow_debug_enclave) {
+    if (!allow_debug_enclave && (quote_body->report_body.attributes.flags & SGX_FLAGS_DEBUG)) {
         ERROR("Quote: DEBUG bit in enclave attributes is set\n");
         return -1;
     }
 
     /* sanity check: enclave must be initialized */
-    if (!(quote->report_body.attributes.flags & SGX_FLAGS_INITIALIZED)) {
+    if (!(quote_body->report_body.attributes.flags & SGX_FLAGS_INITIALIZED)) {
         ERROR("Quote: INIT bit in enclave attributes is not set\n");
         return -1;
     }
 
     /* sanity check: enclave must not have provision/EINIT token key */
-    if ((quote->report_body.attributes.flags & SGX_FLAGS_PROVISION_KEY) ||
-            (quote->report_body.attributes.flags & SGX_FLAGS_LICENSE_KEY)) {
+    if ((quote_body->report_body.attributes.flags & SGX_FLAGS_PROVISION_KEY) ||
+            (quote_body->report_body.attributes.flags & SGX_FLAGS_LICENSE_KEY)) {
         ERROR("Quote: PROVISION_KEY or LICENSE_KEY bit in enclave attributes is set\n");
         return -1;
     }
 
     /* currently only support 64-bit enclaves */
-    if (!(quote->report_body.attributes.flags & SGX_FLAGS_MODE64BIT)) {
+    if (!(quote_body->report_body.attributes.flags & SGX_FLAGS_MODE64BIT)) {
         ERROR("Quote: MODE64 bit in enclave attributes is not set\n");
         return -1;
     }
--- a/Pal/src/host/Linux-SGX/tools/common/attestation.h
+++ b/Pal/src/host/Linux-SGX/tools/common/attestation.h
@@ -23,13 +23,6 @@
 void display_quote(const void* quote_data, size_t quote_size);
 
 /*!
- *  \brief Display internal SGX report body structure (sgx_report_body_t).
- *
- *  \param[in] body Buffer with report body data.
- */
-void display_report_body(const sgx_report_body_t* body);
-
-/*!
  *  \brief Verify IAS attestation report. Also extract the SGX quote contained in IAS report:
  *         allocate enough memory to hold the quote and pass it to the user.
  *
@@ -54,10 +47,9 @@ int verify_ias_report_extract_quote(const uint8_t* ias_report, size_t ias_report
                                     size_t* out_quote_size);
 
 /*!
- *  \brief Verify that the provided SGX quote contains expected values.
+ *  \brief Verify that the provided SGX quote body contains expected values.
  *
- *  \param[in] quote_data      Quote to verify.
- *  \param[in] quote_size      Size of \a quote_data in bytes.
+ *  \param[in] quote_body      Quote body to verify.
  *  \param[in] mr_signer       (Optional) Expected mr_signer quote field.
  *  \param[in] mr_enclave      (Optional) Expected mr_enclave quote field.
  *  \param[in] isv_prod_id     (Optional) Expected isv_prod_id quote field.
@@ -72,18 +64,18 @@ int verify_ias_report_extract_quote(const uint8_t* ias_report, size_t ias_report
  *
  *  \return 0 on successful verification, negative value on error.
  */
-int verify_quote(const void* quote_data, size_t quote_size, const char* mr_signer,
-                 const char* mr_enclave, const char* isv_prod_id, const char* isv_svn,
-                 const char* report_data, bool expected_as_str);
+int verify_quote_body(const sgx_quote_body_t* quote_body, const char* mr_signer,
+                      const char* mr_enclave, const char* isv_prod_id, const char* isv_svn,
+                      const char* report_data, bool expected_as_str);
 
 /*!
- *  \brief Verify enclave attributes of the provided SGX quote.
+ *  \brief Verify enclave attributes of the provided SGX quote body.
  *
- *  \param[in] quote                Quote to verify.
+ *  \param[in] quote_body           Quote body to verify.
  *  \param[in] allow_debug_enclave  If true, then SGXREPORT.ATTRIBUTES.DEBUG can be 1.
  *
  *  \return 0 on successful verification, negative value on error.
  */
-int verify_quote_enclave_attributes(sgx_quote_t* quote, bool allow_debug_enclave);
+int verify_quote_body_enclave_attributes(sgx_quote_body_t* quote_body, bool allow_debug_enclave);
 
 #endif /* ATTESTATION_H */
--- a/Pal/src/host/Linux-SGX/tools/ias-request/ias_request.c
+++ b/Pal/src/host/Linux-SGX/tools/ias-request/ias_request.c
@@ -92,11 +92,11 @@ static int report(struct ias_context_t* ias, const char* quote_path, const char*
     }
 
     sgx_quote_t* quote = (sgx_quote_t*)quote_data;
-    if ((size_t)quote_size < sizeof(sgx_quote_t) + quote->signature_len) {
+    if ((size_t)quote_size < sizeof(sgx_quote_t) + quote->signature_size) {
         ERROR("Quote is too small\n");
         goto out;
     }
-    quote_size = sizeof(sgx_quote_t) + quote->signature_len;
+    quote_size = sizeof(sgx_quote_t) + quote->signature_size;
 
     ret = ias_verify_quote(ias, quote_data, quote_size, nonce, report_path, sig_path, cert_path,
                            advisory_path);
--- a/Pal/src/host/Linux-SGX/tools/ra-tls/ra_tls.h
+++ b/Pal/src/host/Linux-SGX/tools/ra-tls/ra_tls.h
@@ -54,7 +54,7 @@ __attribute__ ((visibility("hidden")))
 int cmp_crt_pk_against_quote_report_data(mbedtls_x509_crt* crt, sgx_quote_t* quote);
 
 __attribute__ ((visibility("hidden")))
-int verify_quote_against_envvar_measurements(const void* quote, size_t quote_size);
+int verify_quote_body_against_envvar_measurements(const sgx_quote_body_t* quote_body);
 
 /*!
  * \brief Callback for user-specific verification of measurements in SGX quote.
--- a/Pal/src/host/Linux-SGX/tools/ra-tls/ra_tls_verify_common.c
+++ b/Pal/src/host/Linux-SGX/tools/ra-tls/ra_tls_verify_common.c
@@ -166,7 +166,7 @@ int cmp_crt_pk_against_quote_report_data(mbedtls_x509_crt* crt, sgx_quote_t* quo
     if (ret < 0)
         return ret;
 
-    ret = memcmp(quote->report_body.report_data.d, sha, SHA256_DIGEST_SIZE);
+    ret = memcmp(quote->body.report_body.report_data.d, sha, SHA256_DIGEST_SIZE);
     if (ret)
         return MBEDTLS_ERR_X509_SIG_MISMATCH;
 
@@ -197,7 +197,7 @@ out:
     return ret;
 }
 
-int verify_quote_against_envvar_measurements(const void* quote, size_t quote_size) {
+int verify_quote_body_against_envvar_measurements(const sgx_quote_body_t* quote_body) {
     int ret;
 
     sgx_measurement_t expected_mrsigner;
@@ -217,12 +217,11 @@ int verify_quote_against_envvar_measurements(const void* quote, size_t quote_siz
     if (ret < 0)
         return MBEDTLS_ERR_X509_BAD_INPUT_DATA;
 
-    ret = verify_quote(quote, quote_size,
-                       validate_mrsigner ? (char*)&expected_mrsigner : NULL,
-                       validate_mrenclave ? (char*)&expected_mrenclave : NULL,
-                       validate_isv_prod_id ? (char*)&expected_isv_prod_id : NULL,
-                       validate_isv_svn ? (char*)&expected_isv_svn : NULL,
-                       /*report_data=*/NULL, /*expected_as_str=*/false);
+    ret = verify_quote_body(quote_body, validate_mrsigner ? (char*)&expected_mrsigner : NULL,
+                            validate_mrenclave ? (char*)&expected_mrenclave : NULL,
+                            validate_isv_prod_id ? (char*)&expected_isv_prod_id : NULL,
+                            validate_isv_svn ? (char*)&expected_isv_svn : NULL,
+                            /*report_data=*/NULL, /*expected_as_str=*/false);
     if (ret < 0)
         return MBEDTLS_ERR_X509_CERT_VERIFY_FAILED;
 
--- a/Pal/src/host/Linux-SGX/tools/ra-tls/ra_tls_verify_dcap.c
+++ b/Pal/src/host/Linux-SGX/tools/ra-tls/ra_tls_verify_dcap.c
@@ -163,8 +163,10 @@ int ra_tls_verify_callback(void* data, mbedtls_x509_crt* crt, int depth, uint32_
             break;
     }
 
-    /* verify enclave attributes from the SGX quote */
-    ret = verify_quote_enclave_attributes(quote, allow_debug_enclave);
+    sgx_quote_body_t* quote_body = &quote->body;
+
+    /* verify enclave attributes from the SGX quote body */
+    ret = verify_quote_body_enclave_attributes(quote_body, allow_debug_enclave);
     if (ret < 0) {
         ret = MBEDTLS_ERR_X509_CERT_VERIFY_FAILED;
         goto out;
@@ -173,13 +175,13 @@ int ra_tls_verify_callback(void* data, mbedtls_x509_crt* crt, int depth, uint32_
     /* verify other relevant enclave information from the SGX quote */
     if (g_verify_measurements_cb) {
         /* use user-supplied callback to verify measurements */
-        ret = g_verify_measurements_cb((const char*)&quote->report_body.mr_enclave,
-                                       (const char*)&quote->report_body.mr_signer,
-                                       (const char*)&quote->report_body.isv_prod_id,
-                                       (const char*)&quote->report_body.isv_svn);
+        ret = g_verify_measurements_cb((const char*)&quote_body->report_body.mr_enclave,
+                                       (const char*)&quote_body->report_body.mr_signer,
+                                       (const char*)&quote_body->report_body.isv_prod_id,
+                                       (const char*)&quote_body->report_body.isv_svn);
     } else {
         /* use default logic to verify measurements */
-        ret = verify_quote_against_envvar_measurements(quote, quote_size);
+        ret = verify_quote_body_against_envvar_measurements(quote_body);
     }
     if (ret < 0) {
         ret = MBEDTLS_ERR_X509_CERT_VERIFY_FAILED;
--- a/Pal/src/host/Linux-SGX/tools/ra-tls/ra_tls_verify_epid.c
+++ b/Pal/src/host/Linux-SGX/tools/ra-tls/ra_tls_verify_epid.c
@@ -233,16 +233,16 @@ int ra_tls_verify_callback(void* data, mbedtls_x509_crt* crt, int depth, uint32_
     }
 
     /* verify that obtained SGX quote (extracted from IAS report) has reasonable size */
-    if (quote_from_ias_size < SGX_QUOTE_BODY_SIZE || quote_from_ias_size > SGX_QUOTE_MAX_SIZE) {
+    if (quote_from_ias_size < sizeof(sgx_quote_body_t) ||
+            quote_from_ias_size > SGX_QUOTE_MAX_SIZE) {
         ret = MBEDTLS_ERR_X509_CERT_VERIFY_FAILED;
         goto out;
     }
 
-    /* TODO: we must pass only the "SGX quote body" struct in the below verify_* functions, because
-     * the IAS report returns a truncated SGX quote (without signature_len and signature fields) */
+    sgx_quote_body_t* quote_body = (sgx_quote_body_t*)quote_from_ias;
 
-    /* verify enclave attributes from the SGX quote */
-    ret = verify_quote_enclave_attributes((sgx_quote_t*)quote_from_ias, allow_debug_enclave);
+    /* verify enclave attributes from the SGX quote body */
+    ret = verify_quote_body_enclave_attributes(quote_body, allow_debug_enclave);
     if (ret < 0) {
         ret = MBEDTLS_ERR_X509_CERT_VERIFY_FAILED;
         goto out;
@@ -251,14 +251,13 @@ int ra_tls_verify_callback(void* data, mbedtls_x509_crt* crt, int depth, uint32_
     /* verify other relevant enclave information from the SGX quote */
     if (g_verify_measurements_cb) {
         /* use user-supplied callback to verify measurements */
-        sgx_quote_t* q = (sgx_quote_t*)quote_from_ias;
-        ret = g_verify_measurements_cb((const char*)&q->report_body.mr_enclave,
-                                       (const char*)&q->report_body.mr_signer,
-                                       (const char*)&q->report_body.isv_prod_id,
-                                       (const char*)&q->report_body.isv_svn);
+        ret = g_verify_measurements_cb((const char*)&quote_body->report_body.mr_enclave,
+                                       (const char*)&quote_body->report_body.mr_signer,
+                                       (const char*)&quote_body->report_body.isv_prod_id,
+                                       (const char*)&quote_body->report_body.isv_svn);
     } else {
         /* use default logic to verify measurements */
-        ret = verify_quote_against_envvar_measurements(quote_from_ias, quote_from_ias_size);
+        ret = verify_quote_body_against_envvar_measurements(quote_body);
     }
     if (ret < 0) {
         ret = MBEDTLS_ERR_X509_CERT_VERIFY_FAILED;
--- a/Pal/src/host/Linux-SGX/tools/verify-ias-report/verify_ias_report.c
+++ b/Pal/src/host/Linux-SGX/tools/verify-ias-report/verify_ias_report.c
@@ -164,33 +164,32 @@ int main(int argc, char* argv[]) {
         DBG("Using IAS public key from file '%s'\n", ias_pubkey_path);
     }
 
-    uint8_t* report_quote = NULL;
-    size_t quote_size = 0;
+    uint8_t* report_quote_body = NULL;
+    size_t quote_body_size = 0;
 
+    /* IAS returns a truncated SGX quote without signature fields (only the SGX quote body) */
     ret = verify_ias_report_extract_quote(report, report_size, sig, sig_size,
                                           allow_outdated_tcb, nonce, ias_pubkey,
-                                          &report_quote, &quote_size);
+                                          &report_quote_body, &quote_body_size);
     if (ret < 0)
         return ret;
 
     /* verify that obtained SGX quote (extracted from IAS report) has reasonable size */
-    if (quote_size < SGX_QUOTE_BODY_SIZE || quote_size > SGX_QUOTE_MAX_SIZE) {
-        ERROR("SGX quote returned in IAS report has unexpected size %lu\n", quote_size);
-        free(report_quote);
+    if (quote_body_size < sizeof(sgx_quote_body_t) || quote_body_size > SGX_QUOTE_MAX_SIZE) {
+        ERROR("SGX quote returned in IAS report has unexpected size %lu\n", quote_body_size);
+        free(report_quote_body);
         return -1;
     }
 
-    /* TODO: we must pass only the "SGX quote body" struct in the below verify_* functions, because
-     * the IAS report returns a truncated SGX quote (without signature_len and signature fields) */
-
-    ret = verify_quote(report_quote, quote_size, mrsigner, mrenclave, isv_prod_id, isv_svn,
-                       report_data, /*expected_as_str=*/true);
+    ret = verify_quote_body((sgx_quote_body_t*)report_quote_body, mrsigner, mrenclave, isv_prod_id,
+                            isv_svn, report_data, /*expected_as_str=*/true);
     if (ret < 0) {
-        free(report_quote);
+        free(report_quote_body);
         return ret;
     }
 
-    ret = verify_quote_enclave_attributes((sgx_quote_t*)report_quote, allow_debug_enclave);
-    free(report_quote);
+    ret = verify_quote_body_enclave_attributes((sgx_quote_body_t*)report_quote_body,
+                                               allow_debug_enclave);
+    free(report_quote_body);
     return ret;
 }


dimakuv (changes requested):
> Reviewed 8 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/219)* status: all files reviewed, 10 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @vijaydhanraj)

___
*[Pal/src/host/Linux-SGX/tools/common/attestation.h, line 30 at r1](https://reviewable.io/reviews/gramineproject/gramine/219#-MoY6bPdCsPiumWEzCPS:-MoY6bPe6Jag40jhNXl0:bte2ycs) ([raw file](https://github.com/gramineproject/gramine/blob/825f22207adb2bbf109b7361d4b11336d227d23d/Pal/src/host/Linux-SGX/tools/common/attestation.h#L30)):*
> ```objc
>  *  \param[in] body Buffer with quote body data.
>  */
> void display_quote_body(const sgx_quote_body_t* body);
> ```

Please rename `body` argument to `quote_body` -- to be uniform with the below function `verify_quote_enclave_attributes()`.

Don't forget to change the name in the comment for this func (above).
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.h, line 67 at r1](https://reviewable.io/reviews/gramineproject/gramine/219#-MoY6spFD7-7ev31qXUZ:-MoY6spFD7-7ev31qXU_:b-ug78ki) ([raw file](https://github.com/gramineproject/gramine/blob/825f22207adb2bbf109b7361d4b11336d227d23d/Pal/src/host/Linux-SGX/tools/common/attestation.h#L67)):*
> ```objc
>  *
>  *  \param[in] quote_data      Quote body to verify.
>  *  \param[in] quote_size      Size of \a quote body in bytes.
> ```

Typo: `quote body` -> `quote_body` (forgot the underscore)
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.h, line 82 at r1](https://reviewable.io/reviews/gramineproject/gramine/219#-MoY8eB76ZWeDVem5Kbn:-MoY8eB76ZWeDVem5Kbo:b-l2abqy) ([raw file](https://github.com/gramineproject/gramine/blob/825f22207adb2bbf109b7361d4b11336d227d23d/Pal/src/host/Linux-SGX/tools/common/attestation.h#L82)):*
> ```objc
>  *  \return 0 on successful verification, negative value on error.
>  */
> int verify_quote(const void* quote_data, size_t quote_size, const char* mr_signer,
> ```

You forgot to change names in this func: should be `quote_body` and `quote_body_size`. Also please change the names in the comment above.
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.h, line 82 at r1](https://reviewable.io/reviews/gramineproject/gramine/219#-MoYAxJX3I0BfUOk5NTc:-MoYAxJX3I0BfUOk5NTd:b-vab996) ([raw file](https://github.com/gramineproject/gramine/blob/825f22207adb2bbf109b7361d4b11336d227d23d/Pal/src/host/Linux-SGX/tools/common/attestation.h#L82)):*
> ```objc
>  *  \return 0 on successful verification, negative value on error.
>  */
> int verify_quote(const void* quote_data, size_t quote_size, const char* mr_signer,
> ```

Actually, we should use the real type instead of `const void*` and we can get rid of `quote_size` (because callers of this function already checked the size of the quote, no need to double-check). So the function signature should look like this:
```
int verify_quote(const sgx_quote_body_t* quote_body, const char* mr_signer, ...
```
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.c, line 86 at r1](https://reviewable.io/reviews/gramineproject/gramine/219#-MoY7-xl6Gki7nV2B19Z:-MoY7-xl6Gki7nV2B19_:b-l1aqfg) ([raw file](https://github.com/gramineproject/gramine/blob/825f22207adb2bbf109b7361d4b11336d227d23d/Pal/src/host/Linux-SGX/tools/common/attestation.c#L86)):*
> ```C
> }
> 
> void display_quote_body(const sgx_quote_body_t* body) {
> ```

ditto (`body` -> `quote_body`)
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.c, line 117 at r1](https://reviewable.io/reviews/gramineproject/gramine/219#-MoY7K2X6zN8jImRICXf:-MoY7K2X6zN8jImRICXg:b35urqw) ([raw file](https://github.com/gramineproject/gramine/blob/825f22207adb2bbf109b7361d4b11336d227d23d/Pal/src/host/Linux-SGX/tools/common/attestation.c#L117)):*
> ```C
> 
>     // quotes from IAS reports are missing signature fields
>     if (quote_size >= sizeof(sgx_quote_body_t)) {
> ```

This change is wrong. E.g., if `quote_size` is only one byte greater than `sgx_quote_body_t`, you are **not** supposed to read `quote->signature_len` (you'll get only first byte from the quote object, the rest 3 bytes will be garbage).

So either:
- revert this change (old code was fine because `sizeof(sgx_quote_t)` covers the `signature_len` field)
- or change to something like: `if (quote_size >= sizeof(sgx_quote_body_t) + sizeof(quote->signature_len)) `
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.c, line 333 at r1](https://reviewable.io/reviews/gramineproject/gramine/219#-MoYBOVG7imQxvlsOo28:-MoYBOVG7imQxvlsOo29:b-d7fmz2) ([raw file](https://github.com/gramineproject/gramine/blob/825f22207adb2bbf109b7361d4b11336d227d23d/Pal/src/host/Linux-SGX/tools/common/attestation.c#L333)):*
> ```C
>         ERROR("Quote: Bad size %zu < %zu\n", quote_body_size, sizeof(sgx_quote_body_t));
>         goto out;
>     }
> ```

You won't need these 5 lines if you change the signature of `verify_quote()` as I explained in the previous comment
___
*[Pal/src/host/Linux-SGX/tools/ra-tls/ra_tls_verify_common.c, line 200 at r1](https://reviewable.io/reviews/gramineproject/gramine/219#-MoY9HmtDOOnPaoVB9Za:-MoY9HmtDOOnPaoVB9Zb:b-f5kquz) ([raw file](https://github.com/gramineproject/gramine/blob/825f22207adb2bbf109b7361d4b11336d227d23d/Pal/src/host/Linux-SGX/tools/ra-tls/ra_tls_verify_common.c#L200)):*
> ```C
> }
> 
> int verify_quote_against_envvar_measurements(const void* quote_body, size_t quote_body_size) {
> ```

You forgot to update `ra_tls.h` -- argument names should be changed in that file as well.
___
*[Pal/src/host/Linux-SGX/tools/ra-tls/ra_tls_verify_common.c, line 200 at r1](https://reviewable.io/reviews/gramineproject/gramine/219#-MoYBX7MB9wz5vqmG6xN:-MoYBX7MB9wz5vqmG6xO:b-djhjk) ([raw file](https://github.com/gramineproject/gramine/blob/825f22207adb2bbf109b7361d4b11336d227d23d/Pal/src/host/Linux-SGX/tools/ra-tls/ra_tls_verify_common.c#L200)):*
> ```C
> }
> 
> int verify_quote_against_envvar_measurements(const void* quote_body, size_t quote_body_size) {
> ```

Actually, we should use the real type instead of `const void*` and we can get rid of `quote_body_size` (because callers of this function already checked the size of the quote, no need to double-check). So the function signature should look like this:
```
int verify_quote_against_envvar_measurements(const sgx_quote_body_t* quote_body)
```
___
*[Pal/src/host/Linux-SGX/tools/verify-ias-report/verify_ias_report.c, line 172 at r1](https://reviewable.io/reviews/gramineproject/gramine/219#-MoYA7J0Ad1udzruTa1e:-MoYA7J0Ad1udzruTa1f:bcbwf8m) ([raw file](https://github.com/gramineproject/gramine/blob/825f22207adb2bbf109b7361d4b11336d227d23d/Pal/src/host/Linux-SGX/tools/verify-ias-report/verify_ias_report.c#L172)):*
> ```C
>     ret = verify_ias_report_extract_quote(report, report_size, sig, sig_size,
>                                           allow_outdated_tcb, nonce, ias_pubkey,
>                                           &report_quote_body, &quote_body_size);
> ```

This renaming is wrong. You're actually extracting the full SGX quote, not just the body. Please revert these renames.

But in `verify_quote...` functions below you of course need to use the SGX quote body, so you'll need to cast there:
```
verify_quote((sgx_quote_body_t*)report_quote, ...
verify_quote_enclave_attributes((sgx_quote_body_t*)report_quote_body, ...
```

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/219)* status: 2 of 9 files reviewed, 10 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[Pal/src/host/Linux-SGX/tools/common/attestation.h, line 30 at r1](https://reviewable.io/reviews/gramineproject/gramine/219#-MoY6bPdCsPiumWEzCPS:-ModC3Ev6zCQ7BstLb6k:bs7wumb) ([raw file](https://github.com/gramineproject/gramine/blob/825f22207adb2bbf109b7361d4b11336d227d23d/Pal/src/host/Linux-SGX/tools/common/attestation.h#L30)):*

Please rename `body` argument to `quote_body` -- to be uniform with the below function `verify_quote_enclave_attributes()`.

Don't forget to change the name in the comment for this func (above).
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.h, line 67 at r1](https://reviewable.io/reviews/gramineproject/gramine/219#-MoY6spFD7-7ev31qXUZ:-ModCllk0GuQdQxamoTQ:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/825f22207adb2bbf109b7361d4b11336d227d23d/Pal/src/host/Linux-SGX/tools/common/attestation.h#L67)):*

Typo: `quote body` -> `quote_body` (forgot the underscore)
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.h, line 82 at r1](https://reviewable.io/reviews/gramineproject/gramine/219#-MoY8eB76ZWeDVem5Kbn:-ModCnMo1HOwfjVEzyFh:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/825f22207adb2bbf109b7361d4b11336d227d23d/Pal/src/host/Linux-SGX/tools/common/attestation.h#L82)):*

You forgot to change names in this func: should be `quote_body` and `quote_body_size`. Also please change the names in the comment above.
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.h, line 82 at r1](https://reviewable.io/reviews/gramineproject/gramine/219#-MoYAxJX3I0BfUOk5NTc:-ModEit-87Rhm4EMk6uI:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/825f22207adb2bbf109b7361d4b11336d227d23d/Pal/src/host/Linux-SGX/tools/common/attestation.h#L82)):*

Actually, we should use the real type instead of `const void*` and we can get rid of `quote_size` (because callers of this function already checked the size of the quote, no need to double-check). So the function signature should look like this:
```
int verify_quote(const sgx_quote_body_t* quote_body, const char* mr_signer, ...
```
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.c, line 86 at r1](https://reviewable.io/reviews/gramineproject/gramine/219#-MoY7-xl6Gki7nV2B19Z:-ModF4kKD9RPXOqNRFUK:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/825f22207adb2bbf109b7361d4b11336d227d23d/Pal/src/host/Linux-SGX/tools/common/attestation.c#L86)):*

ditto (`body` -> `quote_body`)
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.c, line 117 at r1](https://reviewable.io/reviews/gramineproject/gramine/219#-MoY7K2X6zN8jImRICXf:-ModGQpxDsipckQswdnY:bgeqfz2) ([raw file](https://github.com/gramineproject/gramine/blob/825f22207adb2bbf109b7361d4b11336d227d23d/Pal/src/host/Linux-SGX/tools/common/attestation.c#L117)):*

This change is wrong. E.g., if `quote_size` is only one byte greater than `sgx_quote_body_t`, you are **not** supposed to read `quote->signature_len` (you'll get only first byte from the quote object, the rest 3 bytes will be garbage).

So either:
- revert this change (old code was fine because `sizeof(sgx_quote_t)` covers the `signature_len` field)
- or change to something like: `if (quote_size >= sizeof(sgx_quote_body_t) + sizeof(quote->signature_len)) `
</blockquote></details>

Good catch. Yes, I have changed the condition to check `if (quote_size >= sizeof(sgx_quote_body_t) + sizeof(quote->signature_len))`
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.c, line 333 at r1](https://reviewable.io/reviews/gramineproject/gramine/219#-MoYBOVG7imQxvlsOo28:-ModH8mR2gvjp4skl6BQ:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/825f22207adb2bbf109b7361d4b11336d227d23d/Pal/src/host/Linux-SGX/tools/common/attestation.c#L333)):*

You won't need these 5 lines if you change the signature of `verify_quote()` as I explained in the previous comment
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/tools/ra-tls/ra_tls_verify_common.c, line 200 at r1](https://reviewable.io/reviews/gramineproject/gramine/219#-MoY9HmtDOOnPaoVB9Za:-ModI1Gc0lwVCsj3iU63:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/825f22207adb2bbf109b7361d4b11336d227d23d/Pal/src/host/Linux-SGX/tools/ra-tls/ra_tls_verify_common.c#L200)):*

You forgot to update `ra_tls.h` -- argument names should be changed in that file as well.
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/tools/ra-tls/ra_tls_verify_common.c, line 200 at r1](https://reviewable.io/reviews/gramineproject/gramine/219#-MoYBX7MB9wz5vqmG6xN:-ModIp4X2QF22F4TD_va:b-3vwnbr) ([raw file](https://github.com/gramineproject/gramine/blob/825f22207adb2bbf109b7361d4b11336d227d23d/Pal/src/host/Linux-SGX/tools/ra-tls/ra_tls_verify_common.c#L200)):*

Actually, we should use the real type instead of `const void*` and we can get rid of `quote_body_size` (because callers of this function already checked the size of the quote, no need to double-check). So the function signature should look like this:
```
int verify_quote_against_envvar_measurements(const sgx_quote_body_t* quote_body)
```
</blockquote></details>

Yes, the only user of the `quote_body_size` parameter was the `verify_quote()` function. Since we removed the `quote_body_size` parameter from `verify_quote()`, it makes sense to remove it here as well.
___
*[Pal/src/host/Linux-SGX/tools/verify-ias-report/verify_ias_report.c, line 172 at r1](https://reviewable.io/reviews/gramineproject/gramine/219#-MoYA7J0Ad1udzruTa1e:-ModN-TX0qBKM3NnjdC6:b-ju8ov) ([raw file](https://github.com/gramineproject/gramine/blob/825f22207adb2bbf109b7361d4b11336d227d23d/Pal/src/host/Linux-SGX/tools/verify-ias-report/verify_ias_report.c#L172)):*

This renaming is wrong. You're actually extracting the full SGX quote, not just the body. Please revert these renames.

But in `verify_quote...` functions below you of course need to use the SGX quote body, so you'll need to cast there:
```
verify_quote((sgx_quote_body_t*)report_quote, ...
verify_quote_enclave_attributes((sgx_quote_body_t*)report_quote_body, ...
```
</blockquote></details>

I did the renaming after checking this, https://github.com/gramineproject/gramine/blob/master/Pal/src/host/Linux-SGX/tools/common/attestation.c#L270. Here we actually extract `isvEnclaveQuoteBody` object instead of `quote` and so the renaming seemed appropriate.

dimakuv (changes requested):
> Reviewed 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/219)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @vijaydhanraj)

___
*[Pal/src/host/Linux-SGX/tools/common/attestation.h, line 67 at r2](https://reviewable.io/reviews/gramineproject/gramine/219#-Mogq4tK15wqCegKM9aY:-Mogq4tK15wqCegKM9aZ:b-q6lz8k) ([raw file](https://github.com/gramineproject/gramine/blob/7fd1fd3dbc1962c9e06964538f2c48f1169be448/Pal/src/host/Linux-SGX/tools/common/attestation.h#L67)):*
> ```objc
>  *
>  *  \param[in] quote_body      Quote body to verify.
>  *  \param[in] quote_body_size Size of \a quote_body in bytes.
> ```

We don't have this argument anymore, please remove this comment
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.c, line 328 at r2](https://reviewable.io/reviews/gramineproject/gramine/219#-MogqLN0-xLPCpk66ko7:-MogqLN0-xLPCpk66ko8:b-z8yy9h) ([raw file](https://github.com/gramineproject/gramine/blob/7fd1fd3dbc1962c9e06964538f2c48f1169be448/Pal/src/host/Linux-SGX/tools/common/attestation.c#L328)):*
> ```C
>     int ret = -1;
> 
>     sgx_quote_body_t* body = (sgx_quote_body_t*)quote_body;
> ```

This is a no-op assignment now. Please remove it and just use `quote_body` everywhere in this function.
___
*[Pal/src/host/Linux-SGX/tools/verify-ias-report/verify_ias_report.c, line 172 at r1](https://reviewable.io/reviews/gramineproject/gramine/219#-MoYA7J0Ad1udzruTa1e:-Mogs43i0imHu0F5j9m1:b5vexem) ([raw file](https://github.com/gramineproject/gramine/blob/825f22207adb2bbf109b7361d4b11336d227d23d/Pal/src/host/Linux-SGX/tools/verify-ias-report/verify_ias_report.c#L172)):*

I did the renaming after checking this, https://github.com/gramineproject/gramine/blob/master/Pal/src/host/Linux-SGX/tools/common/attestation.c#L270. Here we actually extract `isvEnclaveQuoteBody` object instead of `quote` and so the renaming seemed appropriate.
</blockquote></details>

Ah, true. This is rather obscure, so could you add a comment before this function, smth like:
```
/* IAS returns a truncated SGX quote without signature fields (only the SGX quote body) */
```

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/219)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @vijaydhanraj)

___
*[Pal/src/host/Linux-SGX/tools/common/attestation.c, line 328 at r2](https://reviewable.io/reviews/gramineproject/gramine/219#-MogqLN0-xLPCpk66ko7:-MoiU5CwBlvxflEtw-D8:b9hox92) ([raw file](https://github.com/gramineproject/gramine/blob/7fd1fd3dbc1962c9e06964538f2c48f1169be448/Pal/src/host/Linux-SGX/tools/common/attestation.c#L328)):*

This is a no-op assignment now. Please remove it and just use `quote_body` everywhere in this function.
</blockquote></details>

We will get an error due to `const` qualifier of `const sgx_quote_body_t*` at the below usage,
```
sgx_report_body_t* report_body = &quote_body->report_body;
```
 So, shall we remove the `const` qualifier? I prefer to keep it the current way.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/219)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @vijaydhanraj)

___
*[Pal/src/host/Linux-SGX/tools/common/attestation.c, line 328 at r2](https://reviewable.io/reviews/gramineproject/gramine/219#-MogqLN0-xLPCpk66ko7:-Moi_bKtE5OpefawMEHC:b-z8nf9o) ([raw file](https://github.com/gramineproject/gramine/blob/7fd1fd3dbc1962c9e06964538f2c48f1169be448/Pal/src/host/Linux-SGX/tools/common/attestation.c#L328)):*

We will get an error due to `const` qualifier of `const sgx_quote_body_t*` at the below usage,
```
sgx_report_body_t* report_body = &quote_body->report_body;
```
 So, shall we remove the `const` qualifier? I prefer to keep it the current way.
</blockquote></details>

Ah, I see. I really dislike the `const` rules in the C language, but ok. Resolving, I'm fine with the current version.

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/219)* status: 7 of 9 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[Pal/src/host/Linux-SGX/tools/common/attestation.h, line 67 at r2](https://reviewable.io/reviews/gramineproject/gramine/219#-Mogq4tK15wqCegKM9aY:-Moic4aN69ZkRlkUwXdm:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/7fd1fd3dbc1962c9e06964538f2c48f1169be448/Pal/src/host/Linux-SGX/tools/common/attestation.h#L67)):*

We don't have this argument anymore, please remove this comment
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.c, line 328 at r2](https://reviewable.io/reviews/gramineproject/gramine/219#-MogqLN0-xLPCpk66ko7:-Moid6yrA8UNVlG2Adkn:b-ucfkn4) ([raw file](https://github.com/gramineproject/gramine/blob/7fd1fd3dbc1962c9e06964538f2c48f1169be448/Pal/src/host/Linux-SGX/tools/common/attestation.c#L328)):*

Ah, I see. I really dislike the `const` rules in the C language, but ok. Resolving, I'm fine with the current version.
</blockquote></details>

Thanks.
___
*[Pal/src/host/Linux-SGX/tools/verify-ias-report/verify_ias_report.c, line 172 at r1](https://reviewable.io/reviews/gramineproject/gramine/219#-MoYA7J0Ad1udzruTa1e:-MoiclBv6hgZRwIyTA1S:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/825f22207adb2bbf109b7361d4b11336d227d23d/Pal/src/host/Linux-SGX/tools/verify-ias-report/verify_ias_report.c#L172)):*

Ah, true. This is rather obscure, so could you add a comment before this function, smth like:
```
/* IAS returns a truncated SGX quote without signature fields (only the SGX quote body) */
```
</blockquote></details>

Done.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/219)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners

dimakuv:
Jenkins, test this please

mkow (changes requested):
> Reviewed 2 of 8 files at r1, 5 of 7 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/219)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @vijaydhanraj)

___
*[Pal/src/host/Linux-SGX/sgx_attest.h, line 60 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphdRNC98O30KM6cHgi:-MphdRNC98O30KM6cHgj:b3sstkc) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/sgx_attest.h#L60)):*
> ```objc
> typedef struct _sgx_quote_t {
>     sgx_quote_body_t body;
>     uint32_t signature_len;
> ```

Does this include this field itself or not? For me it sounds like it doesn't, but the code in this PR assumes that it does.
___
*[Pal/src/host/Linux-SGX/sgx_attest.h, line 60 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphddSm1xLaYsR_t8MH:-MphddSn5379UUvwxmUA:b-cictx8) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/sgx_attest.h#L60)):*
> ```objc
> typedef struct _sgx_quote_t {
>     sgx_quote_body_t body;
>     uint32_t signature_len;
> ```

`len` -> `size` (for uniformity with our naming conventions)
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.h, line 81 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphbqLy8dx4JgGDwfNc:-MphbqLy8dx4JgGDwfNd:btzna8e) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/tools/common/attestation.h#L81)):*
> ```objc
>  *  \return 0 on successful verification, negative value on error.
>  */
> int verify_quote(const sgx_quote_body_t* quote_body, const char* mr_signer,
> ```

Wait, why does `verify_quote()` skips the signature and blindly trusts the quote fields?
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.c, line 86 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphcJIV9Hbq5XQFRVVa:-MphcJIV9Hbq5XQFRVVb:b2felz3) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/tools/common/attestation.c#L86)):*
> ```C
> }
> 
> void display_quote_body(const sgx_quote_body_t* quote_body) {
> ```

this should be `static`
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.c, line 99 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphchueCPhWomLgvHOc:-MphchueCPhWomLgvHOd:b7c9s1q) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/tools/common/attestation.c#L99)):*
> ```C
>     INFO("xeid              : ");
>     HEXDUMP(quote_body->xeid);
>     INFO("basename          : ");
> ```

`display_report_body()` indents this section, but here you miss this.
You'll also need to fix the colon below so that it's still aligned.
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.c, line 117 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphehWrBeWEL3aVIpU1:-MphehWrBeWEL3aVIpU2:b-yp5ljn) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/tools/common/attestation.c#L117)):*
> ```C
> 
>     // quotes from IAS reports are missing signature fields
>     if (quote_size >= sizeof(sgx_quote_body_t) + sizeof(quote->signature_len)) {
> ```

Ad the comment about `signature_len`: here you assume that `signature_len` includes its own size.
___
*[Pal/src/host/Linux-SGX/tools/ra-tls/ra_tls_verify_dcap.c, line 166 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphfmILDHf-Z3AowrBd:-MphfmILDHf-Z3AowrBe:b-26pxko) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/tools/ra-tls/ra_tls_verify_dcap.c#L166)):*
> ```C
>     }
> 
>     sgx_quote_body_t quote_body = ((sgx_quote_t*)quote)->body;
> ```

Was it intended to copy the whole structure here?

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/219)* status: 5 of 11 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[Pal/src/host/Linux-SGX/sgx_attest.h, line 60 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphdRNC98O30KM6cHgi:-MpjTVJN4_OpVphNbKza:b-viiakp) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/sgx_attest.h#L60)):*

Does this include this field itself or not? For me it sounds like it doesn't, but the code in this PR assumes that it does.
</blockquote></details>

Can you please clarify? Are you asking if the sgx quote struct includes the `signature_len` field?
___
*[Pal/src/host/Linux-SGX/sgx_attest.h, line 60 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphddSm1xLaYsR_t8MH:-MpjUGTmDkTw9fz4fVE2:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/sgx_attest.h#L60)):*

`len` -> `size` (for uniformity with our naming conventions)
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.h, line 81 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphbqLy8dx4JgGDwfNc:-MpjV2VO1hlwCzchtoX4:b-sjdmnv) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/tools/common/attestation.h#L81)):*

Wait, why does `verify_quote()` skips the signature and blindly trusts the quote fields?
</blockquote></details>

Callers of `verify_quote()` already validate `quote` and so we skip doing a check again.
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.c, line 86 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphcJIV9Hbq5XQFRVVa:-Mpjd2eDCyNGSTfLJf7n:b-u0qeii) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/tools/common/attestation.c#L86)):*

this should be `static`
</blockquote></details>

yes agree. I also made `display_report_body()` static.
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.c, line 99 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphchueCPhWomLgvHOc:-MpjfCxKAOhccDa_D_A9:b-4c5ezp) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/tools/common/attestation.c#L99)):*

`display_report_body()` indents this section, but here you miss this.
You'll also need to fix the colon below so that it's still aligned.
</blockquote></details>

Good catch, done.
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.c, line 117 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphehWrBeWEL3aVIpU1:-MpjhMiA73xnH9zYec44:berye6v) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/tools/common/attestation.c#L117)):*

Ad the comment about `signature_len`: here you assume that `signature_len` includes its own size.
</blockquote></details>

Added the following comment. Does this clarify?
```
    /* quotes from IAS reports are missing signature fields. So display signature and signature_size
       fields only for dcap based quotes */
```
___
*[Pal/src/host/Linux-SGX/tools/ra-tls/ra_tls_verify_dcap.c, line 166 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphfmILDHf-Z3AowrBd:-MpjiLew6fATnCeGmaLc:b-evc2kc) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/tools/ra-tls/ra_tls_verify_dcap.c#L166)):*

Was it intended to copy the whole structure here?
</blockquote></details>

Changed this to a pointer.

dimakuv (changes requested):
> Reviewed 6 of 6 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/219)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @mkow and @vijaydhanraj)

___
*[Pal/src/host/Linux-SGX/sgx_attest.h, line 60 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphdRNC98O30KM6cHgi:-Mpjwx3d111B2Z-wucjd:b-5jue07) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/sgx_attest.h#L60)):*

Can you please clarify? Are you asking if the sgx quote struct includes the `signature_len` field?
</blockquote></details>

@mkow There are two entities that use the term "SGX quote".

- The Quoting Enclave generates the full SGX quote. This full SGX quote is of `sgx_quote_t` type and lways includes `signature_len` and `signature`.
- The IAS (for EPID-based attestation) returns back the trimmed SGX quote. This trimmed SGX quote is of `sgx_quote_body_t` type (see https://api.trustedservices.intel.com/documents/IAS-API-Spec-rev-5.0.pdf, Section 3.2.2, search for `isvEnclaveQuoteBody`).
___
*[Pal/src/host/Linux-SGX/sgx_attest.h, line 60 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphddSm1xLaYsR_t8MH:-MpkD7Ma3yXt6sZvfrCQ:b-hy2wtg) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/sgx_attest.h#L60)):*

Done.
</blockquote></details>

I'm ok with this change, but just want to point out -- there are SGX API specifications from Intel where this field is called "length". I'm not sure what to do with this problem: sometimes we consider terms/keywords from specifications to be not clear or not complying to our Gramine standards -- is it fine to rename in our code?
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.h, line 81 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphbqLy8dx4JgGDwfNc:-MpkENKjDIHV51j04UEK:biezqfd) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/tools/common/attestation.h#L81)):*

Callers of `verify_quote()` already validate `quote` and so we skip doing a check again.
</blockquote></details>

@mkow Do you want to rename this function to `verify_quote_body()`? I wouldn't mind that. Or what would be the suggestion?
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.c, line 117 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphehWrBeWEL3aVIpU1:-MpkExzc8HOqLt5AR-30:bxulodv) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/tools/common/attestation.c#L117)):*
> here you assume that signature_len includes its own size.

@mkow Why are you saying this? `signature_len` reflects only the size of the `sgx_quote_t::signature` field. It *does not* include its own size. Maybe you didn't notice two *different* structs in the two IF checks (`sizeof(sgx_quote_body_t)` vs `sizeof(sgx_quote_t)`)?

I actually don't see much need in adding any comment at all here. But I don't mind Vijay's added comment (though @vijaydhanraj please replace `dcap based` with `DCAP-based`).
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.c, line 119 at r4](https://reviewable.io/reviews/gramineproject/gramine/219#-MpkFcwkDvKhpQQYazLj:-MpkFcwkDvKhpQQYazLk:bk8akad) ([raw file](https://github.com/gramineproject/gramine/blob/3a65628a05123add0f7718ad57cf0f8f54a7a8d6/Pal/src/host/Linux-SGX/tools/common/attestation.c#L119)):*
> ```C
>        fields only for dcap based quotes */
>     if (quote_size >= sizeof(sgx_quote_body_t) + sizeof(quote->signature_size)) {
>         INFO("signature_size     : %d (0x%x)\n", quote->signature_size, quote->signature_size);
> ```

You need to remove one space before `:` now, for nice alignment
___
*[Pal/src/host/Linux-SGX/tools/ra-tls/ra_tls_verify_dcap.c, line 166 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphfmILDHf-Z3AowrBd:-MpkFoQb5CmkaYhmoaIu:b-ajrzp4) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/tools/ra-tls/ra_tls_verify_dcap.c#L166)):*

Changed this to a pointer.
</blockquote></details>

Good catch :) Though it didn't matter much.

mkow (changes requested):
> Reviewed 6 of 6 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/219)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @vijaydhanraj)

___
*[Pal/src/host/Linux-SGX/sgx_attest.h, line 60 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphddSm1xLaYsR_t8MH:-Mpkm7On9EPO3yI28vON:b-m8g8cv) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/sgx_attest.h#L60)):*

I'm ok with this change, but just want to point out -- there are SGX API specifications from Intel where this field is called "length". I'm not sure what to do with this problem: sometimes we consider terms/keywords from specifications to be not clear or not complying to our Gramine standards -- is it fine to rename in our code?
</blockquote></details>

Yeah, I assumed it's just copied from the spec, but I prefer to use our conventions for consistency.
___
*[Pal/src/host/Linux-SGX/sgx_attest.h, line 60 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphdRNC98O30KM6cHgi:-MpkmGsY0NWJd2rG0UiN:bmj9h25) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/sgx_attest.h#L60)):*
> @mkow There are two entities that use the term "SGX quote".

Yes, I know, this is what this PR is about :P

My question is whether this field is `size(signature)` or `size(signature_len || signature)`.
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.h, line 81 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphbqLy8dx4JgGDwfNc:-MpknccABQ1RbBhoHggt:b-7rkc81) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/tools/common/attestation.h#L81)):*
> Callers of `verify_quote()` already validate `quote` and so we skip doing a check again.

But _this_ function is named ~"quote verification", why should the caller _also_ verify quote, if it just called "verify quote"?

> @mkow Do you want to rename this function to `verify_quote_body()`? I wouldn't mind that. Or what would be the suggestion?

Sounds good, but it also needs and explicit comment IMO that this doesn't verify the signature. Same with `verify_quote_against_envvar_measurements()`.

> Callers of `verify_quote()` already validate `quote` and so we skip doing a check again.

Another question related to this: Do they actually do this? Just want to ensure that this function name didn't make anyone skip the additional checks.
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.c, line 99 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphchueCPhWomLgvHOc:-MpkqeDUEnWC_KJd3pyY:b98kl7x) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/tools/common/attestation.c#L99)):*
> You'll also need to fix the colon below so that it's still aligned.

You didn't do this ;) The colons are misaligned now.
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.c, line 117 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphehWrBeWEL3aVIpU1:-MpkrP7IBv7AY5UwWanY:b-bygez6) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/tools/common/attestation.c#L117)):*
I think you both misunderstood my comment or the code ;)

> Added the following comment. Does this clarify?

I said "Ad the comment", not "Add the comment" ;) This is the example for the discussion thread pinned at sgx_attest.h.

@dimakuv: You say "It does _not_ include its own size.". Then what's the meaning of `sizeof(sgx_quote_t) + quote->signature_size` expression if it skips `signature_size` field, a few lines below this comment? (I should have pinned the discussion there, sorry)

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/219)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @vijaydhanraj)

___
*[Pal/src/host/Linux-SGX/sgx_attest.h, line 60 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphdRNC98O30KM6cHgi:-Mpl4BVk2U0vt3CUzSw_:b-5m7pdz) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/sgx_attest.h#L60)):*

This field is `size(signature)`.
</blockquote></details>

Ok, resolving, see explanation below.
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.c, line 117 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphehWrBeWEL3aVIpU1:-Mpl4HOJ-2PmWGe4CRSV:b-l8mmii) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/tools/common/attestation.c#L117)):*

> Then what's the meaning of `sizeof(sgx_quote_t) + quote->signature_size` expression if it skips `signature_size` field

It doesn't skip `signature_size` field. Check the definition of `sgx_quote_t`:
```
typedef struct _sgx_quote_t {
    sgx_quote_body_t body;
    uint32_t signature_size;
    uint8_t signature[];
} sgx_quote_t;
```

So `sizeof(sgx_quote_t)` *already includes* the `signature_size` field. Again, I think you're confusing `sgx_quote_body_t` vs `sgx_quote_t`.
</blockquote></details>

Ahh, I see what happened. I first too a look at the `if` with `sizeof(sgx_quote_body_t) + sizeof(quote->signature_size)` and then at the one with `sizeof(sgx_quote_t) + quote->signature_size)` and didn't notice these are different lines, so I kind of took the first addend from the first one and the second from the second :) (as both `if`s look visually very similarly)

Ok, this code if fine then, sorry for confusion.

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/219)* status: 4 of 11 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[Pal/src/host/Linux-SGX/tools/common/attestation.h, line 81 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphbqLy8dx4JgGDwfNc:-Mpo8ht20kFgINxmXcRJ:b-h34pny) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/tools/common/attestation.h#L81)):*
Changed `verify_quote`->`verify_quote_body`, 
`verify_quote_against_envvar_measurements`-> `verify_quote_body_against_envvar_measurements` and also changed
`verify_quote_enclave_attributes` -> `verify_quote_body_enclave_attributes`

Given this do we need a comment saying that these functions do not verify the signature? PS: Quote body does not have the signature field itself in the first place.

> Another question related to this: Do they actually do this? Just want to ensure that this function name didn't make anyone skip the additional checks.

For DCAP, yes, this `sgx_qv_verify_quote` function does seem to verify the signature and returns `SGX_QL_QE_REPORT_INVALID_SIGNATURE`in case of failure. Please see https://download.01.org/intel-sgx/latest/dcap-latest/linux/docs/Intel_SGX_ECDSA_QuoteLibReference_DCAP_API.pdf, pages `29-34`.

For EPID based trimmed quote, there is no `signature` or `signature_len` but we do a check on `size(sgx_quote_body_t)` before calling `verify_quote_body()`.
See here, https://github.com/gramineproject/gramine/pull/219/files#diff-7fcaed3ad7a104dd5d4e90335cab93c9f76e87fc0aea49976ab5c83d802220ddR236 and
https://github.com/gramineproject/gramine/pull/219/files#diff-2487ced5606ccff3c3343ca96f8e4cba9bbf13e03352b6634b7f5d74ad268d34R178
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.c, line 99 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphchueCPhWomLgvHOc:-MpoMTaX62Z7_aJ64_gK:bgbo24o) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/tools/common/attestation.c#L99)):*

> You'll also need to fix the colon below so that it's still aligned.

You didn't do this ;) The colons are misaligned now.
</blockquote></details>

Here is the current output from `gramine-sgx_quote_dump` tool which calls `display_quote`. Does this look fine?
```
quote_body        :
 version          : 0300
 sign_type        : 0200
 epid_group_id    : 00000000
 qe_svn           : 0600
 pce_svn          : 0b00
 xeid             : 939a7233
 basename         : f79c4ca9940a0db3957f0607562125326eef0c2a45d525f19f50129800000000
report_body       :
 cpu_svn          : 04050708ffff01000000000000000000
 misc_select      : 00000000
 reserved1        : 000000000000000000000000
 isv_ext_prod_id  : 00000000000000000000000000000000
 attributes.flags : 0500000000000000
 attributes.xfrm  : e700000000000000
 mr_enclave       : ea1d8689b474c9c46ef0de6fc136222908f91bc3aa17b0d165d6ed8507fcd42d
 reserved2        : 0000000000000000000000000000000000000000000000000000000000000000
 mr_signer        : d412a4f07ef83892a5915fb2ab584be31e186e5a4f95ab5f6950fd4eb8694d7b
 reserved3        : 0000000000000000000000000000000000000000000000000000000000000000
 config_id        : 00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
 isv_prod_id      : 0000
 isv_svn          : 0000
 config_svn       : 0000
 reserved4        : 000000000000000000000000000000000000000000000000000000000000000000000000000000000000
 isv_family_id    : 00000000000000000000000000000000
 report_data      : 00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
signature_size    : 4288 (0x10c0)
signature         : 0a8cc7b347efe3d....
```
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.c, line 117 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphehWrBeWEL3aVIpU1:-Mpo0B0O2poRArha9pPn:b-cadpkj) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/tools/common/attestation.c#L117)):*

Ahh, I see what happened. I first too a look at the `if` with `sizeof(sgx_quote_body_t) + sizeof(quote->signature_size)` and then at the one with `sizeof(sgx_quote_t) + quote->signature_size)` and didn't notice these are different lines, so I kind of took the first addend from the first one and the second from the second :) (as both `if`s look visually very similarly)

Ok, this code if fine then, sorry for confusion.
</blockquote></details>

@dimakuv I have changed `dcap based` to `DCAP-based`. But if is not needed, I can remove it.

@mkow For future references, do you mean `And` when you say `Ad`?
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.c, line 119 at r4](https://reviewable.io/reviews/gramineproject/gramine/219#-MpkFcwkDvKhpQQYazLj:-Mpo1U8yALs58JoJ8Cjw:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/3a65628a05123add0f7718ad57cf0f8f54a7a8d6/Pal/src/host/Linux-SGX/tools/common/attestation.c#L119)):*

You need to remove one space before `:` now, for nice alignment
</blockquote></details>

Done.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/219)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @mkow and @vijaydhanraj)

___
*[Pal/src/host/Linux-SGX/tools/common/attestation.h, line 81 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphbqLy8dx4JgGDwfNc:-MpoujdPERQRC3rOuuJ_:b-ayv7gw) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/tools/common/attestation.h#L81)):*

Changed `verify_quote`->`verify_quote_body`, 
`verify_quote_against_envvar_measurements`-> `verify_quote_body_against_envvar_measurements` and also changed
`verify_quote_enclave_attributes` -> `verify_quote_body_enclave_attributes`

Given this do we need a comment saying that these functions do not verify the signature? PS: Quote body does not have the signature field itself in the first place.

> Another question related to this: Do they actually do this? Just want to ensure that this function name didn't make anyone skip the additional checks.

For DCAP, yes, this `sgx_qv_verify_quote` function does seem to verify the signature and returns `SGX_QL_QE_REPORT_INVALID_SIGNATURE`in case of failure. Please see https://download.01.org/intel-sgx/latest/dcap-latest/linux/docs/Intel_SGX_ECDSA_QuoteLibReference_DCAP_API.pdf, pages `29-34`.

For EPID based trimmed quote, there is no `signature` or `signature_len` but we do a check on `size(sgx_quote_body_t)` before calling `verify_quote_body()`.
See here, https://github.com/gramineproject/gramine/pull/219/files#diff-7fcaed3ad7a104dd5d4e90335cab93c9f76e87fc0aea49976ab5c83d802220ddR236 and
https://github.com/gramineproject/gramine/pull/219/files#diff-2487ced5606ccff3c3343ca96f8e4cba9bbf13e03352b6634b7f5d74ad268d34R178


</blockquote></details>

I agree with Vijay that we don't need an explicit comment for these functions, because they already have it (`quote_body`) in their names, and these functions don't have any "access" to the signature field.
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.h, line 50 at r5](https://reviewable.io/reviews/gramineproject/gramine/219#-MpouR2sCJC9pDoy65c8:-MpouR2sCJC9pDoy65c9:b-onys77) ([raw file](https://github.com/gramineproject/gramine/blob/73773f8e0b3dee7fd5760617470ee0e94ff3ba1b/Pal/src/host/Linux-SGX/tools/common/attestation.h#L50)):*
> ```objc
> 
> /*!
>  *  \brief Verify that the provided SGX quote contains expected values.
> ```

Please also fix this comment: `SGX quote contains` -> `SGX quote body contains`
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.h, line 72 at r5](https://reviewable.io/reviews/gramineproject/gramine/219#-Mpov0fC9HpAGteAq2P3:-Mpov0fC9HpAGteAq2P4:b-ob7gkd) ([raw file](https://github.com/gramineproject/gramine/blob/73773f8e0b3dee7fd5760617470ee0e94ff3ba1b/Pal/src/host/Linux-SGX/tools/common/attestation.h#L72)):*
> ```objc
> 
> /*!
>  *  \brief Verify enclave attributes of the provided SGX quote.
> ```

`SGX quote body`
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.c, line 99 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphchueCPhWomLgvHOc:-Mpov89RCbwDe6M7YQQE:b-nbxz02) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/tools/common/attestation.c#L99)):*

Here is the current output from `gramine-sgx_quote_dump` tool which calls `display_quote`. Does this look fine?
```
quote_body        :
 version          : 0300
 sign_type        : 0200
 epid_group_id    : 00000000
 qe_svn           : 0600
 pce_svn          : 0b00
 xeid             : 939a7233
 basename         : f79c4ca9940a0db3957f0607562125326eef0c2a45d525f19f50129800000000
report_body       :
 cpu_svn          : 04050708ffff01000000000000000000
 misc_select      : 00000000
 reserved1        : 000000000000000000000000
 isv_ext_prod_id  : 00000000000000000000000000000000
 attributes.flags : 0500000000000000
 attributes.xfrm  : e700000000000000
 mr_enclave       : ea1d8689b474c9c46ef0de6fc136222908f91bc3aa17b0d165d6ed8507fcd42d
 reserved2        : 0000000000000000000000000000000000000000000000000000000000000000
 mr_signer        : d412a4f07ef83892a5915fb2ab584be31e186e5a4f95ab5f6950fd4eb8694d7b
 reserved3        : 0000000000000000000000000000000000000000000000000000000000000000
 config_id        : 00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
 isv_prod_id      : 0000
 isv_svn          : 0000
 config_svn       : 0000
 reserved4        : 000000000000000000000000000000000000000000000000000000000000000000000000000000000000
 isv_family_id    : 00000000000000000000000000000000
 report_data      : 00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
signature_size    : 4288 (0x10c0)
signature         : 0a8cc7b347efe3d....
```
</blockquote></details>

This looks fine to me.
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.c, line 117 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphehWrBeWEL3aVIpU1:-MpovF_x272PZheWbC-E:bgunjr) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/tools/common/attestation.c#L117)):*

@dimakuv I have changed `dcap based` to `DCAP-based`. But if is not needed, I can remove it.

@mkow For future references, do you mean `And` when you say `Ad`?
</blockquote></details>

Thanks, the comment is good now.
___
*[Pal/src/host/Linux-SGX/tools/ra-tls/ra_tls_verify_epid.c, line 244 at r5](https://reviewable.io/reviews/gramineproject/gramine/219#-MpovYng3WJpRpRvM9-B:-MpovYng3WJpRpRvM9-C:bb4ih04) ([raw file](https://github.com/gramineproject/gramine/blob/73773f8e0b3dee7fd5760617470ee0e94ff3ba1b/Pal/src/host/Linux-SGX/tools/ra-tls/ra_tls_verify_epid.c#L244)):*
> ```C
>     sgx_quote_body_t* quote_body = (sgx_quote_body_t*)quote_from_ias;
> 
>     /* verify enclave attributes from the SGX quote */
> ```

`from the SGX quote body`

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/219)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @vijaydhanraj)

___
*[Pal/src/host/Linux-SGX/tools/common/attestation.h, line 81 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphbqLy8dx4JgGDwfNc:-Mpr_q_379hcNg8B3TQ_:b-xe3yi9) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/tools/common/attestation.h#L81)):*

I agree with Vijay that we don't need an explicit comment for these functions, because they already have it (`quote_body`) in their names, and these functions don't have any "access" to the signature field.
</blockquote></details>

Ok, I'd still prefer an explicit comment, but won't block on this.
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.c, line 99 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphchueCPhWomLgvHOc:-Mpra44-BgZ-1XioO0DU:b67haji) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/tools/common/attestation.c#L99)):*

This looks fine to me.
</blockquote></details>

Looks good after the fix.
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.c, line 117 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphehWrBeWEL3aVIpU1:-MpraFSr1oLuizM1ONT0:b-r68ef5) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/tools/common/attestation.c#L117)):*

Thanks, the comment is good now.
</blockquote></details>

@vijaydhanraj: No, I mean "with regards to", I think this is a borrowed word from latin, but not really sure, maybe I'm actually using it incorrectly... ¯\\_(ツ)_/¯

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/219)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @vijaydhanraj)

___
*[Pal/src/host/Linux-SGX/tools/common/attestation.h, line 50 at r5](https://reviewable.io/reviews/gramineproject/gramine/219#-MpouR2sCJC9pDoy65c8:-MpsR7e90NAAQxICNxuy:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/73773f8e0b3dee7fd5760617470ee0e94ff3ba1b/Pal/src/host/Linux-SGX/tools/common/attestation.h#L50)):*

Please also fix this comment: `SGX quote contains` -> `SGX quote body contains`
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.h, line 72 at r5](https://reviewable.io/reviews/gramineproject/gramine/219#-Mpov0fC9HpAGteAq2P3:-MpsRATF2w5m9Q-Gj0ji:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/73773f8e0b3dee7fd5760617470ee0e94ff3ba1b/Pal/src/host/Linux-SGX/tools/common/attestation.h#L72)):*

`SGX quote body`
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/tools/common/attestation.c, line 117 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphehWrBeWEL3aVIpU1:-MpsRHs_Ao8anvrAFKoc:btejdo1) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/tools/common/attestation.c#L117)):*

@vijaydhanraj: No, I mean "with regards to", I think this is a borrowed word from latin, but not really sure, maybe I'm actually using it incorrectly... ¯\\_(ツ)_/¯
</blockquote></details>

ok got it :)
___
*[Pal/src/host/Linux-SGX/tools/ra-tls/ra_tls_verify_epid.c, line 244 at r5](https://reviewable.io/reviews/gramineproject/gramine/219#-MpovYng3WJpRpRvM9-B:-MpsREBU6VC910aDAB4N:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/73773f8e0b3dee7fd5760617470ee0e94ff3ba1b/Pal/src/host/Linux-SGX/tools/ra-tls/ra_tls_verify_epid.c#L244)):*

`from the SGX quote body`
</blockquote></details>

Done.

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/219)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/219)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

___
*[Pal/src/host/Linux-SGX/tools/common/attestation.c, line 117 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphehWrBeWEL3aVIpU1:-Mpu2UevAwdff5X2Sklw:b-rcs1ud) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/tools/common/attestation.c#L117)):*

ok got it :)
</blockquote></details>

Yeah, `ad` sounds like a typo in `and` or `add`, and it's easy to misinterpret it. Maybe we should always replace it with "wrt" or with the full "with respect/regards to" version. I was also confused (thought you meant `add`).

dimakuv:
Jenkins, test this please

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/219)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

___
*[Pal/src/host/Linux-SGX/tools/common/attestation.c, line 117 at r3](https://reviewable.io/reviews/gramineproject/gramine/219#-MphehWrBeWEL3aVIpU1:-Mpv0YI2BC8XY_a02M_Q:b-9gefkk) ([raw file](https://github.com/gramineproject/gramine/blob/ca6fa5bcc5ff0ad5acd93b31122f9a217888edc7/Pal/src/host/Linux-SGX/tools/common/attestation.c#L117)):*

Yeah, `ad` sounds like a typo in `and` or `add`, and it's easy to misinterpret it. Maybe we should always replace it with "wrt" or with the full "with respect/regards to" version. I was also confused (thought you meant `add`).
</blockquote></details>

Ok, will try to use "w.r.t." instead in the future.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/219)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

dimakuv:
Jenkins, test this please

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/219)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #208, opened 2021-11-09
Test failures seen with ASAN enabled on Ubuntu 20.04 running 5.12, 5.13 upstream kernel


jinengandhi-intel:
## Description of the problem

Seeing multiple issues when enabling ASAN on local CI Ubuntu 20.04 machines running 5.12 and 5.13 kernel.

**Issue 1: Gramine SGX on Ubuntu 20.04 + 5.12 kernel**

Below error is seen for multiple PAL, PF, LibOS tests. Logfiles for a few of the tests are attached here. When run via pytest the test is hung and eventually times out.

[LIBOS_mprotect_filefork_SGX_ASAN.txt](https://github.com/gramineproject/gramine/files/7503628/LIBOS_mprotect_filefork_SGX_ASAN.txt)
[LIBOS_forandexec_SGX_ASAN.txt](https://github.com/gramineproject/gramine/files/7503629/LIBOS_forandexec_SGX_ASAN.txt)
[PAL_Process_ASAN_SGX.txt](https://github.com/gramineproject/gramine/files/7503630/PAL_Process_ASAN_SGX.txt)
[asan_gsgx_fail_test-list.txt](https://github.com/gramineproject/gramine/files/7503718/asan_gsgx_fail_test-list.txt)
```
error: Unexpected memory fault occurred inside PAL (0xba3fc1f)
debug: (untrusted PAL sent PAL event 0x2)
debug: DkProcessExit: Returning exit code 1
```

**Issue 2: test_asan_buffer_overflow failing intermittently with Gramine-direct on Ubuntu 20.04 + 5.12 kernel**
Except for this 1 test, all LibOS, PAL, FS tests run without any issues. Log file for the failing  run is attached.
[LibOS_ASAN_unit_test.txt](https://github.com/gramineproject/gramine/files/7503655/LibOS_ASAN_unit_test.txt)

**Issue 3: Gramine-direct tests hang on Ubuntu 20.04 + 5.13 upstream kernel**

Prompt is never returned and even Ctrl + C doesn't work, the process has to be killed from a new session. On the same machine Gramine-SGX exhibits the same behavior as 5.12 kernel i.e. passes for some tests and fails for some.
 

## Steps to reproduce

Build Gramine with ASAN enabled and run the tests. Let me or @aniket-intelx   know if you need access to the specific machine where we see the issues.

pwmarcz:
[...]

PR #228, opened 2021-11-16
ASan improvements

pwmarcz:
## Description of the changes 

A few changes related to ASan. The "Fix shadow memory size" commit fixes one of the problems described in #208.

## How to test this PR? 

* This fixes `gramine-direct` hanging on some new Linux systems (I reproduced this on Linux 5.11, over at #208 people also reproduced it on 5.13) which like to map vvar/vdso above 0x800000000000 (I'm not sure why that happens, though).
* Accessing "invalid" enclave memory should now trigger ASan (which is why I had to disable it for some PAL tests).
* Overflowing the LibOS stack (you can try adding a big array in `shim_emulate_syscall`) also triggers ASan, because we hit the stack guard page.
  
  However, the user experience is not very good, as ASan then gets triggered recursively when trying to format the report, and the actual report we see is for memory *below* the stack guard page.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/228)

### DIFF
--- a/LibOS/shim/test/fs/test_pf.py
+++ b/LibOS/shim/test/fs/test_pf.py
@@ -172,15 +172,21 @@ class TC_50_ProtectedFiles(test_fs.TC_00_FileSystem):
                                          timeout=timeout)
         self.verify_copy(stdout, stderr, self.ENCRYPTED_DIR, executable)
 
+    # TODO: `mmap` on protected files is broken, because we fail to properly register that memory is
+    # unmapped. The `copy_mmap*` tests technically work, but trigger AddressSanitizer.
+
     # overrides TC_00_FileSystem to not skip this on SGX
+    @unittest.skipIf(os.environ.get('ASAN') == '1', 'mapping protected files is broken')
     def test_204_copy_dir_mmap_whole(self):
         self.do_copy_test('copy_mmap_whole', 30)
 
     # overrides TC_00_FileSystem to not skip this on SGX
+    @unittest.skipIf(os.environ.get('ASAN') == '1', 'mapping protected files is broken')
     def test_205_copy_dir_mmap_seq(self):
         self.do_copy_test('copy_mmap_seq', 60)
 
     # overrides TC_00_FileSystem to not skip this on SGX
+    @unittest.skipIf(os.environ.get('ASAN') == '1', 'mapping protected files is broken')
     def test_206_copy_dir_mmap_rev(self):
         self.do_copy_test('copy_mmap_rev', 60)
 
--- a/Pal/regression/Memory.c
+++ b/Pal/regression/Memory.c
@@ -26,6 +26,9 @@ static void handler(bool is_in_pal, PAL_NUM arg, PAL_CONTEXT* context) {
 #endif
 }
 
+/* Disable AddressSanitizer: this code tries to trigger a memory fault by accessing memory that's
+ * supposed to be inaccessible, but SGX PAL poisons such memory. */
+__attribute_no_sanitize_address
 int main(int argc, char** argv, char** envp) {
     volatile int c;
     DkSetExceptionHandler(handler, PAL_EVENT_MEMFAULT);
--- a/Pal/src/host/Linux-SGX/db_files.c
+++ b/Pal/src/host/Linux-SGX/db_files.c
@@ -12,6 +12,7 @@
 #include <sys/types.h>
 
 #include "api.h"
+#include "asan.h"
 #include "enclave_pf.h"
 #include "enclave_tf.h"
 #include "pal.h"
@@ -422,6 +423,21 @@ static int pf_file_map(struct protected_file* pf, PAL_HANDLE handle, void** addr
             return -PAL_ERROR_NOMEM;
 
         *addr = allocated_enclave_pages;
+    } else {
+        /*
+         * TODO: We should call `get_enclave_pages` here, but it's unclear how we should clean up in
+         * case of errors, because we don't know if anything was mapped over this range before.
+         *
+         * An unconditional `free_enclave_pages` during cleanup would poison this memory range, even
+         * if was being used before. The current code errs on the side of caution, and always leaves
+         * this memory unpoisoned (since we keep `allocated_enclave_pages` set to NULL).
+         *
+         * A proper implementation should first load the protected file, and allocate pages only
+         * after that succeeds.
+         */
+#ifdef ASAN
+        asan_unpoison_region((uintptr_t)*addr, size);
+#endif
     }
 
     if (prot & PAL_PROT_WRITE) {
--- a/Pal/src/host/Linux-SGX/db_memory.c
+++ b/Pal/src/host/Linux-SGX/db_memory.c
@@ -6,6 +6,7 @@
  */
 
 #include "api.h"
+#include "asan.h"
 #include "enclave_pages.h"
 #include "pal.h"
 #include "pal_error.h"
@@ -71,6 +72,16 @@ int _DkVirtualMemoryProtect(void* addr, uint64_t size, pal_prot_flags_t prot) {
     __UNUSED(prot);
 
     assert(WITHIN_MASK(prot, PAL_PROT_MASK));
+
+#ifdef ASAN
+    if (sgx_is_completely_within_enclave(addr, size)) {
+        if (prot) {
+            asan_unpoison_region((uintptr_t)addr, size);
+        } else {
+            asan_poison_region((uintptr_t)addr, size, ASAN_POISON_USER);
+        }
+    }
+#endif
 
     static struct atomic_int at_cnt = {.counter = 0};
     int64_t t = 0;
--- a/Pal/src/host/Linux-SGX/enclave_pages.c
+++ b/Pal/src/host/Linux-SGX/enclave_pages.c
@@ -1,6 +1,7 @@
 #include "enclave_pages.h"
 
 #include "api.h"
+#include "asan.h"
 #include "list.h"
 #include "pal_error.h"
 #include "pal_internal.h"
@@ -235,6 +236,10 @@ out:
             assert(ret >= g_heap_bottom);
         }
         assert(ret + size <= g_heap_top);
+
+#ifdef ASAN
+        asan_unpoison_region((uintptr_t)ret, size);
+#endif
     }
 
     return ret;
@@ -312,6 +317,10 @@ int free_enclave_pages(void* addr, size_t size) {
     }
 
     __atomic_sub_fetch(&g_allocated_pages.counter, freed / g_page_size, __ATOMIC_SEQ_CST);
+
+#ifdef ASAN
+    asan_poison_region((uintptr_t)addr, size, ASAN_POISON_USER);
+#endif
 
 out:
     spinlock_unlock(&g_heap_vma_lock);
--- a/Pal/src/host/Linux-SGX/gdb_integration/sgx_gdb.h
+++ b/Pal/src/host/Linux-SGX/gdb_integration/sgx_gdb.h
@@ -2,12 +2,14 @@
 
 #define MAX_DBG_THREADS 1024
 
-/* This address is shared between our GDB and Gramine-SGX and must
- * reside in non-enclave memory. Gramine-SGX puts an enclave_dbginfo
- * object at this address and periodically updates it. Our GDB
- * reads the object from this address to update its internal structs
- * and learn about enclave layout, active threads, etc. */
-#define DBGINFO_ADDR 0x100000000000
+/*
+ * Pre-defined address at which Gramine-SGX writes the `enclave_dbginfo` object, and periodically
+ * updates it. The object is being read by the GDB integration (`sgx_gdb.c`) using `ptrace`.
+ *
+ * The address should be in untrusted memory (outside of enclave), and should not overlap with the
+ * ASan shadow memory area (see `asan.h`).
+ */
+#define DBGINFO_ADDR 0x10000000000ULL /* 1 TB */
 
 /* This struct is read using PTRACE_PEEKDATA in 8B increments
  * therefore it is aligned as uint64_t. */
--- a/Pal/src/host/Linux-SGX/sgx_main.c
+++ b/Pal/src/host/Linux-SGX/sgx_main.c
@@ -413,6 +413,7 @@ static int initialize_enclave(struct pal_enclave* enclave, const char* manifest_
 
     enclave_entry_addr += pal_area->addr;
 
+    struct mem_area* free_area = NULL;
     if (last_populated_addr > enclave_heap_min) {
         areas[area_num] = (struct mem_area){.desc         = "free",
                                             .skip_eextend = true,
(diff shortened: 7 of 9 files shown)


pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/228)* status: 0 of 10 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

___
*[LibOS/shim/test/fs/test_pf.py, line 175 at r1](https://reviewable.io/reviews/gramineproject/gramine/228#-MoiUM04B9mSuHvPs7Mi:-MoiUM04B9mSuHvPs7Mj:bxhjl19) ([raw file](https://github.com/gramineproject/gramine/blob/b652ec6d0b59b555e60d8372a70870d723536573/LibOS/shim/test/fs/test_pf.py#L175)):*
> ```Python
>         self.verify_copy(stdout, stderr, self.ENCRYPTED_DIR, executable)
> 
>     # TODO: `mmap` on protected files is broken, because we fail to properly register that memory is
> ```

FYI: this is ticketed as #231.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/228)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/228)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[common/include/asan.h, line 94 at r1](https://reviewable.io/reviews/gramineproject/gramine/228#-Mom5xsa7VtfvwPzI_PH:-Mom5xsa7VtfvwPzI_PI:bwrmhlf) ([raw file](https://github.com/gramineproject/gramine/blob/b652ec6d0b59b555e60d8372a70870d723536573/common/include/asan.h#L94)):*
> ```objc
>  * two, LLVM tries to optimize the generated code by emitting bitwise OR instead of addition in the
>  * mem-to-shadow conversion. As a result, low values (such as 1 TB) will not work correctly. A value
>  * at least as high as the shadow map length (1 << 44) should work, but it's probably better to stay
> ```

`44` -> `45`
___
*[common/include/asan.h, line 108 at r1](https://reviewable.io/reviews/gramineproject/gramine/228#-Mom67Ef9bANFsDKwU78:-Mom67Ef9bANFsDKwU79:b1w25mp) ([raw file](https://github.com/gramineproject/gramine/blob/b652ec6d0b59b555e60d8372a70870d723536573/common/include/asan.h#L108)):*
> ```objc
> #define ASAN_SHADOW_START 0x18000000000ULL /* 1.5 TB */
> #define ASAN_SHADOW_SHIFT 3
> #define ASAN_SHADOW_LENGTH (1ULL << 45)
> ```

Can you add a sentence somewhere in the above top-level comment that
```
The shadow map length is 1<<45 to cover the whole 48-bit address space:
Linux sometimes maps `vvar` and `vdso` just above 0x800000000000
(= 1<<48).
```
___
*[common/include/asan.h, line 119 at r1](https://reviewable.io/reviews/gramineproject/gramine/228#-Mom8tDD-nVdO1_UhnaV:-Mom8tDD-nVdO1_UhnaW:b-6l6h2o) ([raw file](https://github.com/gramineproject/gramine/blob/b652ec6d0b59b555e60d8372a70870d723536573/common/include/asan.h#L119)):*
> ```objc
> #define ASAN_POISON_HEAP_LEFT_REDZONE     0xfa
> #define ASAN_POISON_HEAP_AFTER_FREE       0xfd
> #define ASAN_POISON_USER                  0xf7
> ```

From what I understand, this kind is SGX-only and means "logically marked as no-prot or freed enclave pages", is this understanding correct? Could you add a comment for this then (or a better "bug classification" message).
___
*[Pal/regression/Exception.c, line 110 at r1](https://reviewable.io/reviews/gramineproject/gramine/228#-Mom75574yNNCil1Vk-O:-Mom75574yNNCil1Vk-P:b30n01j) ([raw file](https://github.com/gramineproject/gramine/blob/b652ec6d0b59b555e60d8372a70870d723536573/Pal/regression/Exception.c#L110)):*
> ```C
> }
> 
> ```

Why do you disable it here? Can you put a short comment of why ASan is triggered here? Shouldn't this be marked as TODO?
___
*[Pal/regression/Memory.c, line 29 at r1](https://reviewable.io/reviews/gramineproject/gramine/228#-Mom7C6Z3MFBj-OqYam4:-Mom7C6Z3MFBj-OqYam5:b30n01j) ([raw file](https://github.com/gramineproject/gramine/blob/b652ec6d0b59b555e60d8372a70870d723536573/Pal/regression/Memory.c#L29)):*
> ```C
> }
> 
> ```

Why do you disable it here? Can you put a short comment of why ASan is triggered here? Shouldn't this be marked as TODO?
___
*[Pal/src/host/Linux-SGX/db_files.c, line 427 at r1](https://reviewable.io/reviews/gramineproject/gramine/228#-Mom7TMN1ShpdOasQEd1:-Mom7TMN1ShpdOasQEd2:b-vtwsyr) ([raw file](https://github.com/gramineproject/gramine/blob/b652ec6d0b59b555e60d8372a70870d723536573/Pal/src/host/Linux-SGX/db_files.c#L427)):*
> ```C
>         *addr = allocated_enclave_pages;
>     } else {
>         /* TODO: Shouldn't we also call `get_enclave_pages` here? */
> ```

Yeah, good question. I have a feeling this code path is never triggered?
___
*[Pal/src/host/Linux-SGX/sgx_main.c, line 579 at r1](https://reviewable.io/reviews/gramineproject/gramine/228#-Mom8DC95EcS0AjyJu46:-Mom8DC95EcS0AjyJu47:bv3ebco) ([raw file](https://github.com/gramineproject/gramine/blob/b652ec6d0b59b555e60d8372a70870d723536573/Pal/src/host/Linux-SGX/sgx_main.c#L579)):*
> ```C
>         }
>     }
> #endif
> ```

Why are you doing this? I don't get the logic here: "mark all enclave memory as inaccessible other than the free regions (heap)".

UPDATE: I incorrectly read `strcmp` check -- you're actually amrking all enclave memory as accessible, but mark free regions as inaccessible. Could you add a comment or make the `strcmp` check somehow more explicit (`!= 0`), so that future readers don't make the same mistake?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/228)* status: 4 of 10 files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @pwmarcz)

___
*[common/include/asan.h, line 94 at r1](https://reviewable.io/reviews/gramineproject/gramine/228#-Mom5xsa7VtfvwPzI_PH:-MomGpLG90kqJ6xvH6V3:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/b652ec6d0b59b555e60d8372a70870d723536573/common/include/asan.h#L94)):*

`44` -> `45`
</blockquote></details>

Done.
___
*[common/include/asan.h, line 108 at r1](https://reviewable.io/reviews/gramineproject/gramine/228#-Mom67Ef9bANFsDKwU78:-MomHIz99g_q8bauCUQh:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/b652ec6d0b59b555e60d8372a70870d723536573/common/include/asan.h#L108)):*

Can you add a sentence somewhere in the above top-level comment that
```
The shadow map length is 1<<45 to cover the whole 48-bit address space:
Linux sometimes maps `vvar` and `vdso` just above 0x800000000000
(= 1<<48).
```
</blockquote></details>

Done.
___
*[common/include/asan.h, line 119 at r1](https://reviewable.io/reviews/gramineproject/gramine/228#-Mom8tDD-nVdO1_UhnaV:-MomIIPJ8CaU4O-xgBgG:b-3hic1c) ([raw file](https://github.com/gramineproject/gramine/blob/b652ec6d0b59b555e60d8372a70870d723536573/common/include/asan.h#L119)):*

From what I understand, this kind is SGX-only and means "logically marked as no-prot or freed enclave pages", is this understanding correct? Could you add a comment for this then (or a better "bug classification" message).
</blockquote></details>

The constant is defined by LLVM, but that's how we used it. I added comments to explain that, and improved the message.
___
*[common/include/asan.h, line 83 at r2](https://reviewable.io/reviews/gramineproject/gramine/228#-MomcAXNELctQk8FLSh9:-MomcAXNELctQk8FLShA:bpejm9d) ([raw file](https://github.com/gramineproject/gramine/blob/0af0ea5e2b9a43301a4125653b1e8b70b3326476/common/include/asan.h#L83)):*
> ```objc
> /*
>  * Parameters of the shadow memory area. Each byte of shadow memory corresponds to ASAN_SHADOW_ALIGN
>  * (by default 8) bytes of user memory. The shadow map length is 1 << 45 to cover the whole 48-bit
> ```

TODO while rebasing: I put the edits to this comment in a wrong `fixup!` commit, I will have to sort it out.
___
*[Pal/regression/Memory.c, line 29 at r1](https://reviewable.io/reviews/gramineproject/gramine/228#-Mom7C6Z3MFBj-OqYam4:-MomVNcd9nvFWFI1-jwb:bs9z80e) ([raw file](https://github.com/gramineproject/gramine/blob/b652ec6d0b59b555e60d8372a70870d723536573/Pal/regression/Memory.c#L29)):*

Why do you disable it here? Can you put a short comment of why ASan is triggered here? Shouldn't this be marked as TODO?
</blockquote></details>

I added a comment. I don't think that's a TODO, the whole point of this test is to access memory that we're not supposed to.
___
*[Pal/src/host/Linux-SGX/db_files.c, line 427 at r1](https://reviewable.io/reviews/gramineproject/gramine/228#-Mom7TMN1ShpdOasQEd1:-MomVq1m3tXNkIguO44w:brha1z4) ([raw file](https://github.com/gramineproject/gramine/blob/b652ec6d0b59b555e60d8372a70870d723536573/Pal/src/host/Linux-SGX/db_files.c#L427)):*

Yeah, good question. I have a feeling this code path is never triggered?
</blockquote></details>

It definitely is triggered, that's why I had to add the `asan_unpoison_region` below.

I described the problem in the comment. This is yet another way in which this protected files code is badly broken. When we get to reimplementing it, I think moving it out of SGX would help: nobody wouldn't be tempted to take such shortcuts as "use the memory without actually allocating it" that are possible with SGX.
___
*[Pal/src/host/Linux-SGX/sgx_main.c, line 579 at r1](https://reviewable.io/reviews/gramineproject/gramine/228#-Mom8DC95EcS0AjyJu46:-Mom_xiX3TG__FOkgi2f:b-sgwlzp) ([raw file](https://github.com/gramineproject/gramine/blob/b652ec6d0b59b555e60d8372a70870d723536573/Pal/src/host/Linux-SGX/sgx_main.c#L579)):*

Why are you doing this? I don't get the logic here: "mark all enclave memory as inaccessible other than the free regions (heap)".

UPDATE: I incorrectly read `strcmp` check -- you're actually amrking all enclave memory as accessible, but mark free regions as inaccessible. Could you add a comment or make the `strcmp` check somehow more explicit (`!= 0`), so that future readers don't make the same mistake?
</blockquote></details>

I realized this is not right. The only range we care about (i.e. we add it to enclave, but it should not be accessed) is "free", so I changed the code to poison only that range instead.
___
*[Pal/regression/Exception.c, line 110 at r1](https://reviewable.io/reviews/gramineproject/gramine/228#-Mom75574yNNCil1Vk-O:-MomS-K19MpxdqKi1CU2:bjis1nd) ([raw file](https://github.com/gramineproject/gramine/blob/b652ec6d0b59b555e60d8372a70870d723536573/Pal/regression/Exception.c#L110)):*

Why do you disable it here? Can you put a short comment of why ASan is triggered here? Shouldn't this be marked as TODO?
</blockquote></details>

Actually, this one should not be disabled. I did it because it accesses address 0x1000 to trigger a memory fault, but that's because we poisoned that memory instead of leaving it alone.

mkow (approved):
> Reviewed 5 of 6 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/228)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @pwmarcz)

___
*[Pal/src/host/Linux-SGX/db_files.c, line 427 at r1](https://reviewable.io/reviews/gramineproject/gramine/228#-Mom7TMN1ShpdOasQEd1:-MonCqgt06mh2FogmcQ8:b-rkdubw) ([raw file](https://github.com/gramineproject/gramine/blob/b652ec6d0b59b555e60d8372a70870d723536573/Pal/src/host/Linux-SGX/db_files.c#L427)):*

It definitely is triggered, that's why I had to add the `asan_unpoison_region` below.

I described the problem in the comment. This is yet another way in which this protected files code is badly broken. When we get to reimplementing it, I think moving it out of SGX would help: nobody wouldn't be tempted to take such shortcuts as "use the memory without actually allocating it" that are possible with SGX.
</blockquote></details>

+1 to making them non-SGX-specific.

dimakuv (approved):
> Reviewed 5 of 6 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/228)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[Pal/src/host/Linux-SGX/db_files.c, line 427 at r1](https://reviewable.io/reviews/gramineproject/gramine/228#-Mom7TMN1ShpdOasQEd1:-Mp5eTRYA1rFBVd9kJuz:b-28socr) ([raw file](https://github.com/gramineproject/gramine/blob/b652ec6d0b59b555e60d8372a70870d723536573/Pal/src/host/Linux-SGX/db_files.c#L427)):*

+1 to making them non-SGX-specific.
</blockquote></details>

Heh, yes, agreed.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/228)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

___
*[common/include/asan.h, line 83 at r2](https://reviewable.io/reviews/gramineproject/gramine/228#-MomcAXNELctQk8FLSh9:-Mp6IOtvDEFTXe5KsLc-:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/0af0ea5e2b9a43301a4125653b1e8b70b3326476/common/include/asan.h#L83)):*

TODO while rebasing: I put the edits to this comment in a wrong `fixup!` commit, I will have to sort it out.
</blockquote></details>

Done.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/228)* status: :shipit: complete! all files reviewed, all discussions resolved

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/228)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #250, opened 2021-11-26
db_rtld: Allow EXEC (non-pie executable) PAL binaries

dimakuv:
## Description of the changes 

Previously, the common PAL loader assumed that the PAL binary is a shared object file (DYN). In some PAL implementations, the PAL binary may be an EXEC, and thus doesn't need relocations (in other words, its base address is `0x00`).

UPDATE: This PR is superseded by https://github.com/gramineproject/gramine/pull/255 and https://github.com/gramineproject/gramine/pull/264

## How to test this PR? 

N/A. Technically, this PR is currently not needed, but I am implementing a new PAL that uses an EXEC PAL binary. So I decided to upstream this, since it doesn't hurt and it is the only change I needed in the common PAL code.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/250)

### DIFF
--- a/LibOS/shim/src/shim_rtld.c
+++ b/LibOS/shim/src/shim_rtld.c
@@ -40,7 +40,8 @@
  * Structure describing a loaded ELF object. Originally based on glibc link_map structure.
  */
 struct link_map {
-    /* Base address shared object is loaded at. */
+    /* Difference between the address in ELF file and the address in memory (for DYNs, base address
+     * where shared object is loaded at; for EXECs, 0x00). */
     ElfW(Addr) l_addr;
 
     /* Object identifier: file path, or PAL URI if path is unavailable. */
@@ -385,7 +386,9 @@ static struct link_map* map_elf_object(struct shim_handle* file, ElfW(Ehdr)* ehd
 
         l->l_addr = (ElfW(Addr))addr;
     } else {
-        l->l_addr = 0;
+        /* This is a non-pie shared object (EXEC, executable), so the difference between address in
+         * the ELF file and the actual address in memory is zero. */
+        l->l_addr = 0x00;
     }
     l->l_map_start = load_start + l->l_addr;
     l->l_map_end   = load_end + l->l_addr;
--- a/LibOS/shim/src/shim_rtld.c
+++ b/LibOS/shim/src/shim_rtld.c
@@ -41,8 +41,8 @@
  */
 struct link_map {
     /* Difference between the address in ELF file and the address in memory (for DYNs, base address
-     * where shared object is loaded at; for EXECs, 0x00). */
-    ElfW(Addr) l_addr;
+     * where shared object is loaded at; for EXECs, 0x0). */
+    ElfW(Addr) l_diff;
 
     /* Object identifier: file path, or PAL URI if path is unavailable. */
     const char* l_name;
@@ -56,8 +56,7 @@ struct link_map {
     /* Number of program header entries.  */
     ElfW(Half) l_phnum;
 
-    /* Start and finish of memory map for this object.  l_map_start need not be the same as
-     * l_addr. */
+    /* Start and finish of memory map for this object. */
     ElfW(Addr) l_map_start, l_map_end;
 
     const char* l_interp_libname;
@@ -77,7 +76,7 @@ struct loadcmd {
      *   - start, map_end, alloc_end are page-aligned
      *   - map_off is page-aligned
      *
-     * The addresses are not relocated (i.e. you need to add l_addr to them).
+     * The addresses are not relocated (i.e. you need to add l_diff to them).
      */
 
     /* Start of memory area */
@@ -232,7 +231,7 @@ static int reserve_dyn(size_t total_size, void** addr) {
  * This function doesn't undo allocations in case of error: if it fails, it may leave some segments
  * already allocated.
  */
-static int execute_loadcmd(const struct loadcmd* c, ElfW(Addr) load_addr,
+static int execute_loadcmd(const struct loadcmd* c, ElfW(Addr) addr_diff,
                            struct shim_handle* file) {
     int ret;
     int map_flags = MAP_FIXED | MAP_PRIVATE;
@@ -240,7 +239,7 @@ static int execute_loadcmd(const struct loadcmd* c, ElfW(Addr) load_addr,
 
     /* Map the part that should be loaded from file, rounded up to page size. */
     if (c->start < c->map_end) {
-        void* map_start = (void*)(load_addr + c->start);
+        void* map_start = (void*)(c->start + addr_diff);
         size_t map_size = c->map_end - c->start;
 
         if ((ret = bkeep_mmap_fixed(map_start, map_size, c->prot, map_flags, file, c->map_off,
@@ -259,7 +258,7 @@ static int execute_loadcmd(const struct loadcmd* c, ElfW(Addr) load_addr,
     /* Zero out the extra data at the end of mapped area. If necessary, temporarily remap the last
      * page as writable. */
     if (c->data_end < c->map_end) {
-        void* zero_start = (void*)(load_addr + c->data_end);
+        void* zero_start = (void*)(c->data_end + addr_diff);
         size_t zero_size = c->map_end - c->data_end;
         void* last_page_start = ALLOC_ALIGN_DOWN_PTR(zero_start);
 
@@ -283,7 +282,7 @@ static int execute_loadcmd(const struct loadcmd* c, ElfW(Addr) load_addr,
 
     /* Allocate extra pages after the mapped area. */
     if (c->map_end < c->alloc_end) {
-        void* zero_page_start = (void*)(load_addr + c->map_end);
+        void* zero_page_start = (void*)(c->map_end + addr_diff);
         size_t zero_page_size = c->alloc_end - c->map_end;
         int zero_map_flags = MAP_FIXED | MAP_PRIVATE | MAP_ANONYMOUS;
         pal_prot_flags_t zero_pal_prot = LINUX_PROT_TO_PAL(c->prot, zero_map_flags);
@@ -375,7 +374,7 @@ static struct link_map* map_elf_object(struct shim_handle* file, ElfW(Ehdr)* ehd
          * address.
          *
          * Note that we reserve memory starting from offset 0, not from load_start. This is to
-         * ensure that the load base (l_addr) will not be lower than 0.
+         * ensure that the load base address will not be lower than 0.
          */
         void* addr;
 
@@ -384,19 +383,19 @@ static struct link_map* map_elf_object(struct shim_handle* file, ElfW(Ehdr)* ehd
             goto err;
         }
 
-        l->l_addr = (ElfW(Addr))addr;
+        l->l_diff = (ElfW(Addr))addr;
     } else {
         /* This is a non-pie shared object (EXEC, executable), so the difference between address in
          * the ELF file and the actual address in memory is zero. */
-        l->l_addr = 0x00;
+        l->l_diff = 0x0;
     }
-    l->l_map_start = load_start + l->l_addr;
-    l->l_map_end   = load_end + l->l_addr;
+    l->l_map_start = load_start + l->l_diff;
+    l->l_map_end   = load_end + l->l_diff;
 
     /* Execute load commands. */
     l->l_data_segment_size = 0;
     for (struct loadcmd* c = &loadcmds[0]; c < &loadcmds[n_loadcmds]; c++) {
-        if ((ret = execute_loadcmd(c, l->l_addr, file)) < 0) {
+        if ((ret = execute_loadcmd(c, l->l_diff, file)) < 0) {
             errstring = "failed to execute load command";
             goto err;
         }
@@ -405,14 +404,14 @@ static struct link_map* map_elf_object(struct shim_handle* file, ElfW(Ehdr)* ehd
                 && ehdr->e_phoff + phdr_size <= c->map_off + (c->data_end - c->start)) {
             /* Found the program header in this segment. */
             ElfW(Addr) phdr_vaddr = ehdr->e_phoff - c->map_off + c->start;
-            l->l_phdr = (ElfW(Phdr)*)(phdr_vaddr + l->l_addr);
+            l->l_phdr = (ElfW(Phdr)*)(phdr_vaddr + l->l_diff);
         }
 
 
         if (interp_libname_vaddr != 0 && !l->l_interp_libname && c->start <= interp_libname_vaddr
                 && interp_libname_vaddr < c->data_end) {
             /* Found the interpreter name in this segment (but we need to validate length). */
-            const char* interp_libname = (const char*)(interp_libname_vaddr + l->l_addr);
+            const char* interp_libname = (const char*)(interp_libname_vaddr + l->l_diff);
             size_t maxlen = c->data_end - interp_libname_vaddr;
             size_t len = strnlen(interp_libname, maxlen);
             if (len == maxlen) {
@@ -426,7 +425,7 @@ static struct link_map* map_elf_object(struct shim_handle* file, ElfW(Ehdr)* ehd
         if (ehdr->e_entry != 0 && !l->l_entry && c->start <= ehdr->e_entry
                 && ehdr->e_entry < c->data_end) {
             /* Found the entry point in this segment. */
-            l->l_entry = (ElfW(Addr))(ehdr->e_entry + l->l_addr);
+            l->l_entry = (ElfW(Addr))(ehdr->e_entry + l->l_diff);
         }
 
         if (!(c->prot & PROT_EXEC))
@@ -470,7 +469,7 @@ err:
 }
 
 static void remove_elf_object(struct link_map* l) {
-    remove_r_debug((void*)l->l_addr);
+    remove_r_debug((void*)l->l_diff);
     free(l);
 }
 
@@ -626,7 +625,7 @@ int load_elf_object(struct shim_handle* file, struct link_map** out_map) {
     map->l_file = file;
 
     if (map->l_file && !qstrempty(&map->l_file->uri)) {
-        append_r_debug(qstrgetstr(&map->l_file->uri), (void*)map->l_addr);
+        append_r_debug(qstrgetstr(&map->l_file->uri), (void*)map->l_diff);
     }
 
     *out_map = map;
@@ -866,7 +865,7 @@ noreturn void execute_elf_object(struct link_map* exec_map, void* argp, ElfW(aux
     auxp[3].a_type     = AT_ENTRY;
     auxp[3].a_un.a_val = g_exec_map->l_entry;
     auxp[4].a_type     = AT_BASE;
-    auxp[4].a_un.a_val = g_interp_map ? g_interp_map->l_addr : 0;
+    auxp[4].a_un.a_val = g_interp_map ? g_interp_map->l_diff : 0;
     auxp[5].a_type     = AT_RANDOM;
     auxp[5].a_un.a_val = 0; /* filled later */
     if (vdso_addr) {
@@ -941,7 +940,6 @@ BEGIN_RS_FUNC(elf_object) {
 
     CP_REBASE(map->l_name);
     CP_REBASE(map->l_file);
-    DEBUG_RS("base=0x%08lx,name=%s", map->l_addr, map->l_name);
 }
 END_RS_FUNC(elf_object)
 
--- a/Pal/include/pal_rtld.h
+++ b/Pal/include/pal_rtld.h
@@ -15,11 +15,12 @@ enum elf_object_type { ELF_OBJECT_INTERNAL, ELF_OBJECT_EXEC, ELF_OBJECT_PRELOAD 
 
 /* Loaded shared object. */
 struct link_map {
-    ElfW(Addr)  l_addr;       /* Address shared object (its first LOAD segment) is loaded at. */
-    ElfW(Addr)  l_base;       /* Base address (0x0 for EXECs, same as l_addr for DYNs). */
-    const char* l_name;       /* Absolute file name object was found in. */
-    ElfW(Dyn)*  l_ld;         /* Dynamic section of the shared object. */
-    ElfW(Addr)  l_entry;      /* Entry point location (may be empty, e.g., for libs). */
+    ElfW(Addr)  l_addr;  /* Address shared object (its first LOAD segment) is loaded at. */
+    ElfW(Addr)  l_diff;  /* Diff between address in ELF file and address in memory
+                            (0x0 for EXECs, same as l_addr for DYNs). */
+    const char* l_name;  /* Absolute file name object was found in. */
+    ElfW(Dyn)*  l_ld;    /* Dynamic section of the shared object. */
+    ElfW(Addr)  l_entry; /* Entry point location (may be empty, e.g., for libs). */
 
     /* Chain of all shared objects loaded at startup. */
     struct link_map* l_next;
--- a/Pal/src/db_rtld.c
+++ b/Pal/src/db_rtld.c
@@ -22,10 +22,10 @@
  *     Corresponding linker flag is `-Wl,--hash-style=sysv`.
  *
  *  - They must have DYN or EXEC object file type. Notice that addresses in DYN binaries are
- *    actually offsets from the base address (`l_base`) and thus need adjustment, whereas addresses
- *    in EXEC binaries are hard-coded and do not need any adjustment (thus `l_base == 0x0`). The
- *    LibOS shared library is built as DYN, but some PAL regression tests are built as EXEC, so we
- *    support both.
+ *    actually offsets from the actual base address in memory (`l_diff`) and thus need adjustment,
+ *    whereas addresses in EXEC binaries are hard-coded and do not need any adjustment (thus
+ *    `l_diff == 0x00`). The LibOS shared library is built as DYN, but some PAL regression tests are
+ *    built as EXEC, so we support both.
  */
 
 #include <stdbool.h>
@@ -105,14 +105,14 @@ static pal_prot_flags_t elf_segment_prot_to_pal_prot(int elf_segment_prot) {
 }
 
 /* iterate through DSO's program headers to find dynamic section (for dynamic linking) */
-static ElfW(Dyn)* find_dynamic_section(ElfW(Addr) ehdr_addr, ElfW(Addr) base_addr) {
+static ElfW(Dyn)* find_dynamic_section(ElfW(Addr) ehdr_addr, ElfW(Addr) addr_diff) {
     const ElfW(Ehdr)* header = (void*)ehdr_addr;
     const ElfW(Phdr)* phdr   = (void*)(ehdr_addr + header->e_phoff);
 
     ElfW(Dyn)* dynamic_section = NULL;
     for (const ElfW(Phdr)* ph = phdr; ph < &phdr[header->e_phnum]; ph++) {
         if (ph->p_type == PT_DYNAMIC) {
-            dynamic_section = (void*)base_addr + ph->p_vaddr;
+            dynamic_section = (void*)ph->p_vaddr + addr_diff;
             break;
         }
     }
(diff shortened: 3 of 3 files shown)


boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/250)* status: 0 of 1 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[-- commits, line 7 at r1](https://reviewable.io/reviews/gramineproject/gramine/250#-MpReURy5NuqCI6qSQsK):*
Wait, EXEC does not imply no relocations and definitely does not imply 0 base address.
___
*[Pal/src/db_rtld.c, line 560 at r1](https://reviewable.io/reviews/gramineproject/gramine/250#-MpRep7IBxQq91jVqsOM:-MpRep7IBxQq91jVqsON:b-c5otdc) ([raw file](https://github.com/gramineproject/gramine/blob/1f50bf9a79510d79f264ec0d20de8a7a2281ed47/Pal/src/db_rtld.c#L560)):*
> ```C
> 
>     ElfW(Addr) pal_binary_addr = (ElfW(Addr))&__ehdr_start;
>     ElfW(Addr) pal_base_addr   = (ehdr->e_type == ET_EXEC) ? 0x0 : pal_binary_addr;
> ```

I'm greatly confused, what's going on in here?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/250)* status: 0 of 1 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski)

___
*[-- commits, line 7 at r1](https://reviewable.io/reviews/gramineproject/gramine/250#-MpReURy5NuqCI6qSQsK):*

Wait, EXEC does not imply no relocations and definitely does not imply 0 base address.
</blockquote></details>

Actually, yes, this message is very vague (and looks like misleading).

Basically, what I'm trying to say is that ET_EXEC binaries are non-pie, i.e. they are loaded at a specific location and thus the "base address" (offset of the address in virtual memory at which the object was loaded in comparison to what is in the LOAD segment? how to call it properly) is 0x00.
___
*[Pal/src/db_rtld.c, line 560 at r1](https://reviewable.io/reviews/gramineproject/gramine/250#-MpRep7IBxQq91jVqsOM:-MpRiQ8F29_eDYOhGw0H:b-1ya4r5) ([raw file](https://github.com/gramineproject/gramine/blob/1f50bf9a79510d79f264ec0d20de8a7a2281ed47/Pal/src/db_rtld.c#L560)):*

I'm greatly confused, what's going on in here?
</blockquote></details>

This is the base address from which to relocate PAL symbols (relocation happens as `base_addr + symbol.p_vaddr`). For EXECs, there is no such base address (symbols already contain hard-coded addresses).

I feel like I'm misunderstanding something...

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/250)* status: 0 of 1 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[-- commits, line 7 at r1](https://reviewable.io/reviews/gramineproject/gramine/250#-MpReURy5NuqCI6qSQsK):*

Actually, yes, this message is very vague (and looks like misleading).

Basically, what I'm trying to say is that ET_EXEC binaries are non-pie, i.e. they are loaded at a specific location and thus the "base address" (offset of the address in virtual memory at which the object was loaded in comparison to what is in the LOAD segment? how to call it properly) is 0x00.
</blockquote></details>

Base address is just hardcoded in binary. Why can't we just apply the relocations as specified in the ELF? If it needs no, there it won't ask for any...
___
*[Pal/src/db_rtld.c, line 560 at r1](https://reviewable.io/reviews/gramineproject/gramine/250#-MpRep7IBxQq91jVqsOM:-MpRl3yP5cX6W45LlUxN:bmj4bu6) ([raw file](https://github.com/gramineproject/gramine/blob/1f50bf9a79510d79f264ec0d20de8a7a2281ed47/Pal/src/db_rtld.c#L560)):*

This is the base address from which to relocate PAL symbols (relocation happens as `base_addr + symbol.p_vaddr`). For EXECs, there is no such base address (symbols already contain hard-coded addresses).

I feel like I'm misunderstanding something...
</blockquote></details>

Base address is base address - where the elf is loaded at. Why can't we just proceed normally? If there are no relocations, then, well, we won't apply them by definition.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/250)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski)

___
*[-- commits, line 7 at r1](https://reviewable.io/reviews/gramineproject/gramine/250#-MpReURy5NuqCI6qSQsK):*

Base address is just hardcoded in binary. Why can't we just apply the relocations as specified in the ELF? If it needs no, there it won't ask for any...
</blockquote></details>

+1 to Borys' comments. Also, isn't "base address" used more commonly as "the virtual address of the first segment in binary", not the _delta_ between memory and headers?
___
*[Pal/src/db_rtld.c, line 560 at r1](https://reviewable.io/reviews/gramineproject/gramine/250#-MpRep7IBxQq91jVqsOM:-MpcgEBe4MatRExJZyJa:b-xw20zv) ([raw file](https://github.com/gramineproject/gramine/blob/1f50bf9a79510d79f264ec0d20de8a7a2281ed47/Pal/src/db_rtld.c#L560)):*

Base address is base address - where the elf is loaded at. Why can't we just proceed normally? If there are no relocations, then, well, we won't apply them by definition.
</blockquote></details>

I think the confusion stems from @dimakuv using an uncommon definition for base address ;) See the thread above.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/250)* status: 0 of 3 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "squash! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[-- commits, line 7 at r1](https://reviewable.io/reviews/gramineproject/gramine/250#-MpReURy5NuqCI6qSQsK):*

+1 to Borys' comments. Also, isn't "base address" used more commonly as "the virtual address of the first segment in binary", not the _delta_ between memory and headers?
</blockquote></details>

Done. See the new commit message.
___
*[Pal/src/db_rtld.c, line 560 at r1](https://reviewable.io/reviews/gramineproject/gramine/250#-MpRep7IBxQq91jVqsOM:-MpfCJxs0isCkFA07qcN:b-vmrw27) ([raw file](https://github.com/gramineproject/gramine/blob/1f50bf9a79510d79f264ec0d20de8a7a2281ed47/Pal/src/db_rtld.c#L560)):*

I think the confusion stems from @dimakuv using an uncommon definition for base address ;) See the thread above.
</blockquote></details>

Done. Indeed, my choice of the word `base` was wrong. I changed to `diff` everywhere and added some comments.

It also doesn't help that Glibc has `l_addr` which doesn't really mean an actual address: https://github.com/bminor/glibc/blob/e49c3c5d7a553040948a20b982387a097dc1eb2c/include/link.h#L100

I also improved comments in the LibOS rtld a bit (notice that it doesn't need `l_addr` vs `l_diff` fields, because LibOS rtld doesn't perform any relocations).

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/250)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "squash! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[LibOS/shim/src/shim_rtld.c, line 44 at r2](https://reviewable.io/reviews/gramineproject/gramine/250#-MpfLv4y1tm6AWth30h2:-MpfLv4y1tm6AWth30h3:b24hl70) ([raw file](https://github.com/gramineproject/gramine/blob/502f705eb93d72305ad04de891017824eb0c774b/LibOS/shim/src/shim_rtld.c#L44)):*
> ```C
> struct link_map {
>     /* Difference between the address in ELF file and the address in memory (for DYNs, base address
>      * where shared object is loaded at; for EXECs, 0x00). */
> ```

Why `0x00` not `0x0` or `0`?
___
*[LibOS/shim/src/shim_rtld.c, line 45 at r2](https://reviewable.io/reviews/gramineproject/gramine/250#-MpfK46n-vSW6LvooJbw:-MpfK46oEbKO6BB6qwir:b-ryus1u) ([raw file](https://github.com/gramineproject/gramine/blob/502f705eb93d72305ad04de891017824eb0c774b/LibOS/shim/src/shim_rtld.c#L45)):*
> ```C
>     /* Difference between the address in ELF file and the address in memory (for DYNs, base address
>      * where shared object is loaded at; for EXECs, 0x00). */
>     ElfW(Addr) l_addr;
> ```

Could you also rename this field? Judging from comment it's the same as `l_diff` in PAL.
___
*[LibOS/shim/src/shim_rtld.c, line 391 at r2](https://reviewable.io/reviews/gramineproject/gramine/250#-MpfLzkk3Qno9w9MOzGf:-MpfLzkk3Qno9w9MOzGg:bkk7a38) ([raw file](https://github.com/gramineproject/gramine/blob/502f705eb93d72305ad04de891017824eb0c774b/LibOS/shim/src/shim_rtld.c#L391)):*
> ```C
>         /* This is a non-pie shared object (EXEC, executable), so the difference between address in
>          * the ELF file and the actual address in memory is zero. */
>         l->l_addr = 0x00;
> ```

ditto (0x00)
___
*[Pal/src/db_rtld.c, line 115 at r2](https://reviewable.io/reviews/gramineproject/gramine/250#-MpfKSR7A_CdXy9VDq2D:-MpfKSR82NUro6HMy0UH:b6q04ze) ([raw file](https://github.com/gramineproject/gramine/blob/502f705eb93d72305ad04de891017824eb0c774b/Pal/src/db_rtld.c#L115)):*
> ```C
>     for (const ElfW(Phdr)* ph = phdr; ph < &phdr[header->e_phnum]; ph++) {
>         if (ph->p_type == PT_DYNAMIC) {
>             dynamic_section = (void*)ph->p_vaddr + addr_diff;
> ```

Please use `char*` for pointer arithmetic (since you change this already).
___
*[Pal/src/db_rtld.c, line 290 at r2](https://reviewable.io/reviews/gramineproject/gramine/250#-MpfLDOU8yPGZC75XqC5:-MpfLDOU8yPGZC75XqC6:b-wz4sax) ([raw file](https://github.com/gramineproject/gramine/blob/502f705eb93d72305ad04de891017824eb0c774b/Pal/src/db_rtld.c#L290)):*
> ```C
>     struct loadcmd* loadcmds = NULL;
> 
>     ElfW(Addr) l_relro_addr = 0x00;
> ```

Why such a weird change? `0x00` looks weird (unless it's of type char/uint8_t).
___
*[Pal/src/db_rtld.c, line 398 at r2](https://reviewable.io/reviews/gramineproject/gramine/250#-MpfLNr-7j2-UC7AYNB-:-MpfLNr-7j2-UC7AYNB0:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/502f705eb93d72305ad04de891017824eb0c774b/Pal/src/db_rtld.c#L398)):*
> ```C
>             /* memorize where the ELF file (its first loadable segment) was loaded */
>             g_entrypoint_map.l_addr = (ElfW(Addr))map_addr;
>             g_entrypoint_map.l_diff = (ehdr->e_type == ET_EXEC) ? 0x00 : g_entrypoint_map.l_addr;
> ```

ditto
___
*[Pal/src/db_rtld.c, line 560 at r2](https://reviewable.io/reviews/gramineproject/gramine/250#-MpfLgnC7OnesS6XlIUq:-MpfLgnC7OnesS6XlIUr:bkk7a38) ([raw file](https://github.com/gramineproject/gramine/blob/502f705eb93d72305ad04de891017824eb0c774b/Pal/src/db_rtld.c#L560)):*
> ```C
> 
>     ElfW(Addr) pal_binary_addr = (ElfW(Addr))&__ehdr_start;
>     ElfW(Addr) pal_addr_diff   = (ehdr->e_type == ET_EXEC) ? 0x00 : pal_binary_addr;
> ```

ditto (0x00)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/250)* status: 1 of 3 files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "squash! " and "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/250#-Mpff_MV2cXPGiKCeRpS:-Mpff_MV2cXPGiKCeRpT:bgrxiyy) (no related file):*
Since I'm changing the LibOS rtld now in this PR, I would like to ask @pwmarcz to also take a look at the PR.

---
___
*[LibOS/shim/src/shim_rtld.c, line 44 at r2](https://reviewable.io/reviews/gramineproject/gramine/250#-MpfLv4y1tm6AWth30h2:-MpfbBYg7-7d1lVlt0uQ:bvslj4f) ([raw file](https://github.com/gramineproject/gramine/blob/502f705eb93d72305ad04de891017824eb0c774b/LibOS/shim/src/shim_rtld.c#L44)):*

Why `0x00` not `0x0` or `0`?
</blockquote></details>

Done. I thought it was you who prefered to have `0x00` for addresses (represented as ints). But apparently not you, maybe @mkow? Whatever, I also feel that `0x00` is weird.
___
*[LibOS/shim/src/shim_rtld.c, line 45 at r2](https://reviewable.io/reviews/gramineproject/gramine/250#-MpfK46n-vSW6LvooJbw:-MpfbAFL6uIpNWH4Jckf:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/502f705eb93d72305ad04de891017824eb0c774b/LibOS/shim/src/shim_rtld.c#L45)):*

Could you also rename this field? Judging from comment it's the same as `l_diff` in PAL.
</blockquote></details>

Done.
___
*[LibOS/shim/src/shim_rtld.c, line 391 at r2](https://reviewable.io/reviews/gramineproject/gramine/250#-MpfLzkk3Qno9w9MOzGf:-MpfbB8T58mSd2Mfisr6:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/502f705eb93d72305ad04de891017824eb0c774b/LibOS/shim/src/shim_rtld.c#L391)):*

ditto (0x00)
</blockquote></details>

Done.
___
*[Pal/src/db_rtld.c, line 115 at r2](https://reviewable.io/reviews/gramineproject/gramine/250#-MpfKSR7A_CdXy9VDq2D:-MpfcQRf3KOanmflqVZe:b1szsrd) ([raw file](https://github.com/gramineproject/gramine/blob/502f705eb93d72305ad04de891017824eb0c774b/Pal/src/db_rtld.c#L115)):*

Please use `char*` for pointer arithmetic (since you change this already).
</blockquote></details>

Done. Actually, here `p_vaddr` is already a 64-bit int, so there was no need to cast this variable, but rather the result of this equation.

Also, fixed all similar casts to avoid `void*` where possible.
___
*[Pal/src/db_rtld.c, line 290 at r2](https://reviewable.io/reviews/gramineproject/gramine/250#-MpfLDOU8yPGZC75XqC5:-Mpfc2KM1U-ZWpukQevU:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/502f705eb93d72305ad04de891017824eb0c774b/Pal/src/db_rtld.c#L290)):*

Why such a weird change? `0x00` looks weird (unless it's of type char/uint8_t).
</blockquote></details>

Done.
___
*[Pal/src/db_rtld.c, line 398 at r2](https://reviewable.io/reviews/gramineproject/gramine/250#-MpfLNr-7j2-UC7AYNB-:-Mpfc3Ug3kWf5m3dHVJM:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/502f705eb93d72305ad04de891017824eb0c774b/Pal/src/db_rtld.c#L398)):*

ditto
</blockquote></details>

Done.
___
*[Pal/src/db_rtld.c, line 560 at r2](https://reviewable.io/reviews/gramineproject/gramine/250#-MpfLgnC7OnesS6XlIUq:-Mpfc4P2Cw31vsAOWH2M:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/502f705eb93d72305ad04de891017824eb0c774b/Pal/src/db_rtld.c#L560)):*

ditto (0x00)
</blockquote></details>

Done.

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/250)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "squash! " and "fixup! " found in commit messages' one-liners (waiting on @mkow)

mkow (approved):
> Reviewed 1 of 3 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/250)* status: all files reviewed, 1 unresolved discussion, "squash! " and "fixup! " found in commit messages' one-liners

___
*[LibOS/shim/src/shim_rtld.c, line 44 at r2](https://reviewable.io/reviews/gramineproject/gramine/250#-MpfLv4y1tm6AWth30h2:-MpfkreB6bDmjqOsOvu1:b-ebf7fu) ([raw file](https://github.com/gramineproject/gramine/blob/502f705eb93d72305ad04de891017824eb0c774b/LibOS/shim/src/shim_rtld.c#L44)):*

Done. I thought it was you who prefered to have `0x00` for addresses (represented as ints). But apparently not you, maybe @mkow? Whatever, I also feel that `0x00` is weird.
</blockquote></details>

I think I preferred 0x00 only in case of byte arrays, like `{0x20, 0x00, 0x61}`. Here I prefer `0` or `0x0`.

pwmarcz (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/250)* status: all files reviewed, 2 unresolved discussions, "squash! " and "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/250#-Mpff_MV2cXPGiKCeRpS:-Mpfpr801ozdraMbdK5t:barjxdw) (no related file):*

Since I'm changing the LibOS rtld now in this PR, I would like to ask @pwmarcz to also take a look at the PR.
</blockquote></details>

Thanks. Apart from terminology change, this is mostly about handling the `ET_EXEC` case by setting the diff address to 0, right?

I think it would also be cleaner to make the `ET_EXEC` change a separate commit from changing the names in PAL (and maybe doing the PAL and LibOS changes in a single commit instead?). Right now that commit is a few lines of logic changes in `setup_pal_library`, and many more lines of changing the names. I don't insist, though.

---
___
*[LibOS/shim/src/shim_rtld.c, line 44 at r3](https://reviewable.io/reviews/gramineproject/gramine/250#-Mpftmer5Ab0SubL4XkR:-Mpftmet2iaSKYBVwnkN:b-f7v3ob) ([raw file](https://github.com/gramineproject/gramine/blob/20d508605e0abab8f135e7ff0143dc780a5a575b/LibOS/shim/src/shim_rtld.c#L44)):*
> ```C
> struct link_map {
>     /* Difference between the address in ELF file and the address in memory (for DYNs, base address
>      * where shared object is loaded at; for EXECs, 0x0). */
> ```

1. By the "address in ELF file" you don't mean the offset in file, but addresses (vaddr) specified in ELF program headers, right? I'd maybe make that more explicit in this explanation.
2. Do I understand correctly that this is, indeed, the *base address* as defined by the ELF spec[1], but the common usage of "base address" is different than what ELF spec says?

   I'm not sure that's a good idea, `diff` sounds weird/non-standard, and musl seems to just use `base` for this value. By the way, is "load base" the same thing, and is it similarly an ambiguous term?

   But if you keep that, then I would be careful with using the "base address" term in the comments - maybe avoid it altogether?

If I got it wrong, please explain.

[1] https://refspecs.linuxfoundation.org/elf/elf.pdf page 73.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/250)* status: all files reviewed, 2 unresolved discussions, "squash! " and "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/250#-Mpff_MV2cXPGiKCeRpS:-MpkRDiRE6zfAe3rWie-:b-3vg7ao) (no related file):*

Thanks. Apart from terminology change, this is mostly about handling the `ET_EXEC` case by setting the diff address to 0, right?

I think it would also be cleaner to make the `ET_EXEC` change a separate commit from changing the names in PAL (and maybe doing the PAL and LibOS changes in a single commit instead?). Right now that commit is a few lines of logic changes in `setup_pal_library`, and many more lines of changing the names. I don't insist, though.
</blockquote></details>

This PR got messy. I will create a new PR with only the name changes (in both PAL and LibOS), as discussed here. We'll first decide on that new "change-names" PR, and then I'll submit a PR on top of it to support `ET_EXEC` in PAL.

---
___
*[LibOS/shim/src/shim_rtld.c, line 44 at r3](https://reviewable.io/reviews/gramineproject/gramine/250#-Mpftmer5Ab0SubL4XkR:-MpkRReD7jXVwwgz-v6_:bc5zjzs) ([raw file](https://github.com/gramineproject/gramine/blob/20d508605e0abab8f135e7ff0143dc780a5a575b/LibOS/shim/src/shim_rtld.c#L44)):*
1. You are correct, I'm talking about "virtual addresses in the program headers", or `p_vaddr`. I'll make it more explicit.
2. Yes, this is the "base address" as defined in the ELF spec. The point of @boryspoplawski and @mkow is that this is a historic name/term and it is very confusing, and we shouldn't drag this historic baggage in our code. I actually agree -- changing `l_addr` to `l_diff` made the RTLD code cleaner in my mind. So in the new PR, I'll add a huge comment explaining this departure from the ELF standard.

By the way, the phrase `Difference between the address ...` is taken directly from Glibc sources, see https://github.com/bminor/glibc/blob/e49c3c5d7a553040948a20b982387a097dc1eb2c/include/link.h#L100. I blame Glibc for all this mess with names :)

For everyone, I'll copy-paste the relevant paragraphs from the ELF standard (page 73, section "Base address"):

> The virtual addresses in the program headers might not represent the actual virtual addresses
> of the program's memory image. Executable files typically contain absolute code. To let the
> process execute correctly, the segments must reside at the virtual addresses used to build the
> executable file. On the other hand, shared object segments typically contain
> position-independent code. This lets a segment's virtual address change from one process to
> another, without invalidating execution behavior. Though the system chooses virtual addresses
> for individual processes, it maintains the segments’ relative positions. Because
> position-independent code uses relative addressing between segments, the difference between
> virtual addresses in memory must match the difference between virtual addresses in the file.
> The difference between the virtual address of any segment in memory and the corresponding
> virtual address in the file is thus a single constant value for any one executable or shared object
> in a given process. This difference is the base address. One use of the base address is to relocate
> the memory image of the program during dynamic linking.

> An executable or shared object file's base address is calculated during execution from three
> values: the virtual memory load address, the maximum page size, and the lowest virtual address
> of a program's loadable segment. To compute the base address, one determines the memory
> address associated with the lowest p_vaddr value for a PT_LOAD segment. This address is
> truncated to the nearest multiple of the maximum page size. The corresponding p_vaddr value
> itself is also truncated to the nearest multiple of the maximum page size. The base address is
> the difference between the truncated memory address and the truncated p_vaddr value.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/250)* status: all files reviewed, 3 unresolved discussions, "squash! " and "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, and @pwmarcz)

___
*[LibOS/shim/src/shim_rtld.c, line 377 at r3](https://reviewable.io/reviews/gramineproject/gramine/250#-MpkUGh0DVumnK-W84Nq:-MpkUGh15RZIXs2JwiXw:bvca6z7) ([raw file](https://github.com/gramineproject/gramine/blob/20d508605e0abab8f135e7ff0143dc780a5a575b/LibOS/shim/src/shim_rtld.c#L377)):*
> ```C
>          *
>          * Note that we reserve memory starting from offset 0, not from load_start. This is to
>          * ensure that the load base address will not be lower than 0.
> ```

@pwmarcz What does this mean actually?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/250)* status: all files reviewed, 3 unresolved discussions, "squash! " and "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @pwmarcz)

___
*[LibOS/shim/src/shim_rtld.c, line 44 at r3](https://reviewable.io/reviews/gramineproject/gramine/250#-Mpftmer5Ab0SubL4XkR:-MpkwmWH-ynYWsyhYBQ1:bexfpp4) ([raw file](https://github.com/gramineproject/gramine/blob/20d508605e0abab8f135e7ff0143dc780a5a575b/LibOS/shim/src/shim_rtld.c#L44)):*

> Do I understand correctly that this is, indeed, the _base address_ as defined by the ELF spec[1], but the common usage of "base address" is different than what ELF spec says?

Yes. It's super-unintuitive for me that "base address" may mean an adress where there's actually nothing in memory and it's just a delta by which to shift everything.

Maybe "base_shift" then?
</blockquote></details>

@mkow I created https://github.com/gramineproject/gramine/pull/255, let's continue the discussion there.

pwmarcz (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/250)* status: all files reviewed, 3 unresolved discussions, "squash! " and "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[LibOS/shim/src/shim_rtld.c, line 44 at r3](https://reviewable.io/reviews/gramineproject/gramine/250#-Mpftmer5Ab0SubL4XkR:-MpkyZLcDcygIyrOA6Is:bc6vvdh) ([raw file](https://github.com/gramineproject/gramine/blob/20d508605e0abab8f135e7ff0143dc780a5a575b/LibOS/shim/src/shim_rtld.c#L44)):*

@mkow I created https://github.com/gramineproject/gramine/pull/255, let's continue the discussion there.
</blockquote></details>

OK, let's continue there.
___
*[LibOS/shim/src/shim_rtld.c, line 377 at r3](https://reviewable.io/reviews/gramineproject/gramine/250#-MpkUGh0DVumnK-W84Nq:-Mpkyg6F5zaNk9uOZBpH:bxvxx7x) ([raw file](https://github.com/gramineproject/gramine/blob/20d508605e0abab8f135e7ff0143dc780a5a575b/LibOS/shim/src/shim_rtld.c#L377)):*

@pwmarcz What does this mean actually?
</blockquote></details>

Imagine an object for which the fist segment begins at vaddr 0x10000, and last segments ends at vaddr 0x20000 (so `load_start = 0x10000, load_end = 0x20000`). Since we need a memory area of size `load_end - load_start`, it's tempting to reserve it like this:

```c
reserve_dyn(load_end - load_start, &addr);  /* error check omitted */
l->l_diff = addr - load_start;
```

However, it's possible that this will reserve memory at, say, `addr = 0x8000`, making `l_diff = -0x8000` (or rather, underflowing it, since it's unsigned). To make sure `l_diff` doesn't go below zero, we reserve a range of size `load_end` (i.e. pretend it starts at zero), which is slightly wasteful but should not matter too much.

[Here is where we discussed this](https://reviewable.io/reviews/gramineproject/graphene/2231#-MVxguep4jt_ur0L4hXw).

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/250)* status: all files reviewed, 2 unresolved discussions, "squash! " and "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[LibOS/shim/src/shim_rtld.c, line 377 at r3](https://reviewable.io/reviews/gramineproject/gramine/250#-MpkUGh0DVumnK-W84Nq:-MplOHq--plo-yustNPN:bi4ldxt) ([raw file](https://github.com/gramineproject/gramine/blob/20d508605e0abab8f135e7ff0143dc780a5a575b/LibOS/shim/src/shim_rtld.c#L377)):*

Imagine an object for which the fist segment begins at vaddr 0x10000, and last segments ends at vaddr 0x20000 (so `load_start = 0x10000, load_end = 0x20000`). Since we need a memory area of size `load_end - load_start`, it's tempting to reserve it like this:

```c
reserve_dyn(load_end - load_start, &addr);  /* error check omitted */
l->l_diff = addr - load_start;
```

However, it's possible that this will reserve memory at, say, `addr = 0x8000`, making `l_diff = -0x8000` (or rather, underflowing it, since it's unsigned). To make sure `l_diff` doesn't go below zero, we reserve a range of size `load_end` (i.e. pretend it starts at zero), which is slightly wasteful but should not matter too much.

[Here is where we discussed this](https://reviewable.io/reviews/gramineproject/graphene/2231#-MVxguep4jt_ur0L4hXw).
</blockquote></details>

OK, thanks, understood.

PR #252, opened 2021-11-26
gramine-sgx-gen-private-key

woju:
## Description of the changes 

Supply a tool to help generating proper RSA key in proper location. If the key is kept in default location, `--key` argument to `gramine-sgx-sign` is optional.

## How to test this PR? 

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/252)


jinengandhi-intel (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/252)* status: 0 of 8 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @woju)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/252#-MpjvDbc4t43L6UXyaTM:-MpjvDbc4t43L6UXyaTN:b-nobcpn) (no related file):*
Woju, I was trying to have the changes ready for my local CI and hence was testing the gramine-sgx-gen-key command on the CI machines and everywhere I am getting an error:
```
gramine-sgx-gen-key Pal/src/host/Linux-SGX/signer/enclave-key.pem
Traceback (most recent call last):
  File "/usr/local/bin/gramine-sgx-gen-key", line 8, in <module>
    from graminelibos import sgx_sign
  File "/usr/local/lib/python3.6/dist-packages/graminelibos/__init__.py", line 25, in <module>
    from .sgx_sign import get_tbssigstruct, sign_with_local_key, SGX_LIBPAL, SGX_RSA_KEY_PATH
  File "/usr/local/lib/python3.6/dist-packages/graminelibos/sgx_sign.py", line 38, in <module>
    SGX_RSA_KEY_PATH = xdg.XDG_CONFIG_HOME / 'gramine' / 'enclave-key.pem'
AttributeError: module 'xdg' has no attribute 'XDG_CONFIG_HOME'
```
Have you encountered this? I have already installed python3-xdg module.

---

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/252)* status: 0 of 8 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @jinengandhi-intel)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/252#-MpjvDbc4t43L6UXyaTM:-Mpkaaoq0AMLMXtkkXN_:bfkwo8d) (no related file):*
<details><summary><i>Previously, jinengandhi-intel wrote…</i></summary><blockquote>

Woju, I was trying to have the changes ready for my local CI and hence was testing the gramine-sgx-gen-key command on the CI machines and everywhere I am getting an error:
```
gramine-sgx-gen-key Pal/src/host/Linux-SGX/signer/enclave-key.pem
Traceback (most recent call last):
  File "/usr/local/bin/gramine-sgx-gen-key", line 8, in <module>
    from graminelibos import sgx_sign
  File "/usr/local/lib/python3.6/dist-packages/graminelibos/__init__.py", line 25, in <module>
    from .sgx_sign import get_tbssigstruct, sign_with_local_key, SGX_LIBPAL, SGX_RSA_KEY_PATH
  File "/usr/local/lib/python3.6/dist-packages/graminelibos/sgx_sign.py", line 38, in <module>
    SGX_RSA_KEY_PATH = xdg.XDG_CONFIG_HOME / 'gramine' / 'enclave-key.pem'
AttributeError: module 'xdg' has no attribute 'XDG_CONFIG_HOME'
```
Have you encountered this? I have already installed python3-xdg module.
</blockquote></details>

Yes, Jenkins failed for this exact reason (in pylint, but the root cause is the same):

```
************* Module graminelibos.sgx_sign
E: 38,19: Module 'xdg' has no 'XDG_CONFIG_HOME' member (no-member)
```


The root cause: there are two `xdg` modules in circulation:

* https://www.freedesktop.org/wiki/Software/pyxdg/ https://pypi.org/project/pyxdg/, which is what you installed as `python3-xdg`
* https://github.com/srstevenson/xdg https://pypi.org/project/xdg/

They both are `import xdg`, but have incompatible API. PyXDG has some compatibility layer which I hoped to use so that if anyone accidentally installed wrong one thing wouldn't break, but it looks like Ubuntu has too old version for that. I'll have to have my own compat layer somewhere in our module.

---

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/252)* status: 0 of 31 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @jinengandhi-intel and @woju)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/252#-Mv8Xn4P6eh6S-H84FZr:-Mv8Xn4Q1nP4ZJNmKJ0b:b-82pqw5) (no related file):*
Similar fixes (as in the second commit) would need to be pushed in other Gramine repos:
- https://github.com/gramineproject/gsc/
- https://github.com/gramineproject/examples

From what I understand, the old style of explicitly specifying the `--key` option still works, so this is not a breaking change -- GSC and Examples can still work with Gramine after this PR. But we'll need to fix those repos anyway at some point. Am I correct?

---

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/252)* status: 0 of 31 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @jinengandhi-intel)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/252#-Mv8Xn4P6eh6S-H84FZr:-MvIyJAQB8E6FVCRB-kk:b-o24jvp) (no related file):*

Similar fixes (as in the second commit) would need to be pushed in other Gramine repos:
- https://github.com/gramineproject/gsc/
- https://github.com/gramineproject/examples

From what I understand, the old style of explicitly specifying the `--key` option still works, so this is not a breaking change -- GSC and Examples can still work with Gramine after this PR. But we'll need to fix those repos anyway at some point. Am I correct?
</blockquote></details>

Yes, explicit `--key` still works. Examples probably can be fixed; GSC, I'm not sure, it might not be a good idea, since we have that machinery which prevents the key from being included in the last container and it operates on explicit paths, so keeping the key in `/` is probably warranted.

---

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/252)* status: 0 of 31 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @jinengandhi-intel)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/252#-MpjvDbc4t43L6UXyaTM:-MvJ2_e99R-yH_2Wm77R:b-yhrgft) (no related file):*

Yes, Jenkins failed for this exact reason (in pylint, but the root cause is the same):

```
************* Module graminelibos.sgx_sign
E: 38,19: Module 'xdg' has no 'XDG_CONFIG_HOME' member (no-member)
```


The root cause: there are two `xdg` modules in circulation:

* https://www.freedesktop.org/wiki/Software/pyxdg/ https://pypi.org/project/pyxdg/, which is what you installed as `python3-xdg`
* https://github.com/srstevenson/xdg https://pypi.org/project/xdg/

They both are `import xdg`, but have incompatible API. PyXDG has some compatibility layer which I hoped to use so that if anyone accidentally installed wrong one thing wouldn't break, but it looks like Ubuntu has too old version for that. I'll have to have my own compat layer somewhere in our module.
</blockquote></details>

This is now fixed by removing `xdg` module altogether, we just `getenv()` the relevant variable.

---

woju:
Jenkins, test Jenkins-20.04 please (fdatasync01 timed out)

mkow (changes requested):
> Reviewed 20 of 26 files at r2, 5 of 9 files at r3, 12 of 12 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/252)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski, @dimakuv, @jinengandhi-intel, and @woju)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/252#-MvyzT1m11ZYTzjBN1xV:-MvyzT1oEH-JbkH_pY6v:bmajl2c) (no related file):*
You left one reference to `SGX_SIGNER_KEY` in `Documentation/quickstart.rst`.

---
___
*[Documentation/conf.py, line 169 at r4](https://reviewable.io/reviews/gramineproject/gramine/252#-MvyzcSRCy0y6tpo2Mfp:-MvyzcSRCy0y6tpo2Mfq:bhuips3) ([raw file](https://github.com/gramineproject/gramine/blob/21112f739503652bc272d9e5fd43ba54e146e346/Documentation/conf.py#L169)):*
> ```Python
>     ('manpages/gramine', 'gramine-sgx', 'Gramine', [author], 1),
>     ('manpages/gramine-manifest', 'gramine-manifest', 'Gramine manifest preprocessor', [author], 1),
>     ('manpages/gramine-sgx-gen-private-key', 'gramine-sgx-gen-private-key', 'Gramine SGX RSA key generator', [author], 1),
> ```

Not sure if mentioning RSA here makes sense, it's just an implementation / SGX design detail.
___
*[Documentation/manpages/gramine-sgx-gen-private-key.rst, line 16 at r4](https://reviewable.io/reviews/gramineproject/gramine/252#-Mvz3Lly0Ko8Af2SCQG4:-Mvz3LlzDtGc_FNIJCTW:b-kjudxc) ([raw file](https://github.com/gramineproject/gramine/blob/21112f739503652bc272d9e5fd43ba54e146e346/Documentation/manpages/gramine-sgx-gen-private-key.rst#L16)):*
> ```reStructuredText
> ===========
> 
> :program:`gramine-sgx-get-token` is used to generate the private key suitable
> ```

`gramine-sgx-get-token` -> `gramine-sgx-gen-private-key`
___
*[python/graminelibos/sgx_sign.py, line 4 at r4](https://reviewable.io/reviews/gramineproject/gramine/252#-Mvz616LCchWGa-RNtBY:-Mvz616LCchWGa-RNtBZ:b-vaugpb) ([raw file](https://github.com/gramineproject/gramine/blob/21112f739503652bc272d9e5fd43ba54e146e346/python/graminelibos/sgx_sign.py#L4)):*
> ```Python
> # Copyright (C) 2014 Stony Brook University
> # Copyright (C) 2021 Intel Corporation
> ```

`2021-2022` -> `2022` + please just merge with the previous copyrights:



_Suggestion:_
```Python
# Copyright (C) 2022 Intel Corporation
#                    Michał Kowalczyk <mkow@invisiblethingslab.com>
#                    Borys Popławski <borysp@invisiblethingslab.com>
#                    Wojtek Porczyk <woju@invisiblethingslab.com>
```
___
*[python/gramine-sgx-gen-private-key, line 23 at r4](https://reviewable.io/reviews/gramineproject/gramine/252#-Mvz5Ch14DsyQqxCe0m8:-Mvz5Ch14DsyQqxCe0m9:b-nrxixl) ([raw file](https://github.com/gramineproject/gramine/blob/21112f739503652bc272d9e5fd43ba54e146e346/python/gramine-sgx-gen-private-key#L23)):*
> ```
> def main(output, force, quiet):
>     output = pathlib.Path(output)
>     old_umask = os.umask(0o77)
> ```

You sure you meant `0o77` here?

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/252)* status: 33 of 37 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, @jinengandhi-intel, and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/252#-MvyzT1m11ZYTzjBN1xV:-MvzFGtOAlf6uVDfJ5sE:b-896fix) (no related file):*

You left one reference to `SGX_SIGNER_KEY` in `Documentation/quickstart.rst`.
</blockquote></details>

Done.

---
___
*[Documentation/conf.py, line 169 at r4](https://reviewable.io/reviews/gramineproject/gramine/252#-MvyzcSRCy0y6tpo2Mfp:-Mvz9QBt6fallKwEVYKE:b-muaelp) ([raw file](https://github.com/gramineproject/gramine/blob/21112f739503652bc272d9e5fd43ba54e146e346/Documentation/conf.py#L169)):*

Not sure if mentioning RSA here makes sense, it's just an implementation / SGX design detail.
</blockquote></details>

In the manpage title, yeah, I'm also not sure. Removed unless someone objects. The key details are mentioned in the manpage contents.
___
*[Documentation/manpages/gramine-sgx-gen-private-key.rst, line 16 at r4](https://reviewable.io/reviews/gramineproject/gramine/252#-Mvz3Lly0Ko8Af2SCQG4:-MvzAy340pIOwroDjCt-:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/21112f739503652bc272d9e5fd43ba54e146e346/Documentation/manpages/gramine-sgx-gen-private-key.rst#L16)):*

`gramine-sgx-get-token` -> `gramine-sgx-gen-private-key`
</blockquote></details>

Done.
___
*[python/graminelibos/sgx_sign.py, line 4 at r4](https://reviewable.io/reviews/gramineproject/gramine/252#-Mvz616LCchWGa-RNtBY:-MvzFFkcB5XZGTisktjD:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/21112f739503652bc272d9e5fd43ba54e146e346/python/graminelibos/sgx_sign.py#L4)):*

`2021-2022` -> `2022` + please just merge with the previous copyrights:
</blockquote></details>

Done.
___
*[python/gramine-sgx-gen-private-key, line 23 at r4](https://reviewable.io/reviews/gramineproject/gramine/252#-Mvz5Ch14DsyQqxCe0m8:-MvzAzXZ0yc1YR1-A8iz:b-sb3nfz) ([raw file](https://github.com/gramineproject/gramine/blob/21112f739503652bc272d9e5fd43ba54e146e346/python/gramine-sgx-gen-private-key#L23)):*

You sure you meant `0o77` here?
</blockquote></details>

Yes, I did. This is "mask", i.e. any bits lighted here will be masked out of the resulting file permissions. The manual (https://manpages.debian.org/bullseye/manpages-dev/umask.2.en.html) is misleading there, but I'm quite sure I'm right.

The exact maths under this is `mode = mode & ~current_mask()`, and `current_mask()` is whatever I set with `umask` (`& 0o777`, so you can't mask set{u,g}id and sticky).
https://elixir.bootlin.com/linux/v5.16.9/source/kernel/sys.c#L1843
https://elixir.bootlin.com/linux/v5.16.9/source/fs/namei.c#L3241

I think creating the key to be accessible only by the owner (not group, not others) is the right way. This is also what OpenSSH does (https://github.com/openssh/openssh-portable/blob/a23698c3082ffe661abed14b020eac9b0c25eb9f/authfile.c#L61) and they even check file permissions when *reading* the file and refuse to use certain files if they are accessible a bit too much (https://github.com/openssh/openssh-portable/blob/a23698c3082ffe661abed14b020eac9b0c25eb9f/authfile.c#L104-L113 — we can consider also doing this, maybe in another PR).

mkow (changes requested):
> Reviewed 4 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/252)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, @jinengandhi-intel, and @woju)

___
*[python/gramine-sgx-gen-private-key, line 23 at r4](https://reviewable.io/reviews/gramineproject/gramine/252#-Mvz5Ch14DsyQqxCe0m8:-MvzGVbH4Qcs6NkFOM9T:bkzoo2c) ([raw file](https://github.com/gramineproject/gramine/blob/21112f739503652bc272d9e5fd43ba54e146e346/python/gramine-sgx-gen-private-key#L23)):*

Yes, I did. This is "mask", i.e. any bits lighted here will be masked out of the resulting file permissions. The manual (https://manpages.debian.org/bullseye/manpages-dev/umask.2.en.html) is misleading there, but I'm quite sure I'm right.

The exact maths under this is `mode = mode & ~current_mask()`, and `current_mask()` is whatever I set with `umask` (`& 0o777`, so you can't mask set{u,g}id and sticky).
https://elixir.bootlin.com/linux/v5.16.9/source/kernel/sys.c#L1843
https://elixir.bootlin.com/linux/v5.16.9/source/fs/namei.c#L3241

I think creating the key to be accessible only by the owner (not group, not others) is the right way. This is also what OpenSSH does (https://github.com/openssh/openssh-portable/blob/a23698c3082ffe661abed14b020eac9b0c25eb9f/authfile.c#L61) and they even check file permissions when *reading* the file and refuse to use certain files if they are accessible a bit too much (https://github.com/openssh/openssh-portable/blob/a23698c3082ffe661abed14b020eac9b0c25eb9f/authfile.c#L104-L113 — we can consider also doing this, maybe in another PR).
</blockquote></details>

Ah, right. But then please write it as `0o077`, looks less like a typo.

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/252)* status: 36 of 37 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, @jinengandhi-intel, and @mkow)

___
*[python/gramine-sgx-gen-private-key, line 23 at r4](https://reviewable.io/reviews/gramineproject/gramine/252#-Mvz5Ch14DsyQqxCe0m8:-Mvzb0T2B_uP3CZktvha:b-pm2lkd) ([raw file](https://github.com/gramineproject/gramine/blob/21112f739503652bc272d9e5fd43ba54e146e346/python/gramine-sgx-gen-private-key#L23)):*

Ah, right. But then please write it as `0o077`, looks less like a typo.
</blockquote></details>

IDK. This doesn't look more readable to me, it more like someone who didn't use much umask.

I can also do `S_IRWXG | S_IRWXO` for even more readability, but outside of the kernel proper I've never encountered this in any serious code (even though this is "recommended" by the manpage).

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/252)* status: all files reviewed, 2 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @jinengandhi-intel)

___
*[python/gramine-sgx-gen-private-key, line 23 at r4](https://reviewable.io/reviews/gramineproject/gramine/252#-Mvz5Ch14DsyQqxCe0m8:-MvzfoOo7_eGmkiQT8tO:b-levpiu) ([raw file](https://github.com/gramineproject/gramine/blob/21112f739503652bc272d9e5fd43ba54e146e346/python/gramine-sgx-gen-private-key#L23)):*
> I can also do `S_IRWXG | S_IRWXO`

Please don't, it's actually much less readable IMO ;) See https://lwn.net/Articles/696231/ (actually these `PERM_*` macros are what we adopted in Gramine C code).

boryspoplawski (changes requested):
> Reviewed 20 of 26 files at r2, 2 of 9 files at r3, 10 of 12 files at r4, 4 of 4 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/252)* status: all files reviewed, 8 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @jinengandhi-intel, and @woju)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/252#-MvzuS0n1pai6b6Drm4i:-MvzuS0n1pai6b6Drm4j:b-ik10zi) (no related file):*
Suppose I've used `gramine-sgx-gen-private-key --key /abc` to generate key. How do I run tests (from what I can tell it's impossible)?

---
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/252#-MvzunmHAFqAc35aIdcb:-MvzunmHAFqAc35aIdcc:b-v2lbqn) (no related file):*
Maybe `gramine-sgx-gen-priv-key`? It's a bit shorter and `priv` is a well know abbreviation.
Not blocking.

---
___
*[.ci/lib/stage-sgx-gen-key.jenkinsfile, line 1 at r4](https://reviewable.io/reviews/gramineproject/gramine/252#-MvzAYosBwipGLIkVsAw:-MvzAYot1nea3cvYHp-I:b-z060yc) ([raw file](https://github.com/gramineproject/gramine/blob/21112f739503652bc272d9e5fd43ba54e146e346/.ci/lib/stage-sgx-gen-key.jenkinsfile#L1)):*
> ```
> stage('sgx-gen-key') {
> ```

Is it worth creating another jenkinsfile for this? I get that modularity is nice, but this is SGX specific and cannot be reused anywhere. I think leaving this in building (more precisely installing) steps is fine.
___
*[Documentation/devel/building.rst, line 132 at r4](https://reviewable.io/reviews/gramineproject/gramine/252#-MvzBjc-3hkLuTvjzZeT:-MvzBjc-3hkLuTvjzZeU:b-306qpy) ([raw file](https://github.com/gramineproject/gramine/blob/9eafa5f7731650dc54043e4aa4804b6a2e34e292/Documentation/devel/building.rst#L132)):*
> ```reStructuredText
> - https://github.com/intel/linux-sgx
> 
> 5. Generate signing keys
> ```

How does an user which goes by full documentation (not quickstart) knows they need to create such a key?
Please leave this information here or add somewhere else (but I don't think we have a suitable place at the moment)
___
*[python/graminelibos/sgx_sign.py, line 26 at r6](https://reviewable.io/reviews/gramineproject/gramine/252#-Mvzs4Qj8MnkQL1E5KTm:-Mvzs4Qj8MnkQL1E5KTn:b2t5f5i) ([raw file](https://github.com/gramineproject/gramine/blob/a43fcbda34e144ee955d05fae2186d33604d287f/python/graminelibos/sgx_sign.py#L26)):*
> ```Python
> 
> 
> _backend = backends.default_backend()
> ```

`crypto_backend`? With this name reader has no idea what it's about.
Also why import `backends`? This is a pretty generic name, I think `cryptography.hazmat.backends` would look better (and it's only used here).
___
*[python/gramine-sgx-gen-private-key, line 11 at r6](https://reviewable.io/reviews/gramineproject/gramine/252#-MvztPh20RPXOzsgaRI-:-MvztPh20RPXOzsgaRI0:b-cpcpd3) ([raw file](https://github.com/gramineproject/gramine/blob/a43fcbda34e144ee955d05fae2186d33604d287f/python/gramine-sgx-gen-private-key#L11)):*
> ```
> import click
> 
> from graminelibos import sgx_sign
> ```

Shouldn't these tools use only things directly exported by `graminelibos`, not poke into specific modules/parts of implementation?
___
*[python/gramine-sgx-gen-private-key, line 23 at r6](https://reviewable.io/reviews/gramineproject/gramine/252#-MvzrDbJ8ImYGJAm0j2p:-MvzrDbJ8ImYGJAm0j2q:b-mnaq7q) ([raw file](https://github.com/gramineproject/gramine/blob/a43fcbda34e144ee955d05fae2186d33604d287f/python/gramine-sgx-gen-private-key#L23)):*
> ```
> def main(output, force, quiet):
>     output = pathlib.Path(output)
>     old_umask = os.umask(0o077)
> ```

Why not set mode in calls to mkdir and file creation?

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/252)* status: all files reviewed, 8 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, @jinengandhi-intel, and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/252#-MvzuS0n1pai6b6Drm4i:-Mw-8kZEAfF34-U0iFRe:bbquq19) (no related file):*

Suppose I've used `gramine-sgx-gen-private-key --key /abc` to generate key. How do I run tests (from what I can tell it's impossible)?
</blockquote></details>

Is this really a problem? If yes, I can revert the change in `util_tests.py`, only change the default to be consistent with this PR, but still allow override via env `SGX_SIGNER_KEY=/abc`.

cc @pwmarcz

---
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/252#-MvzunmHAFqAc35aIdcb:-Mw-8qEdDAnmS3cnIuEB:brob9hl) (no related file):*

Maybe `gramine-sgx-gen-priv-key`? It's a bit shorter and `priv` is a well know abbreviation.
Not blocking.
</blockquote></details>

No strong opinion. Shell completion works from `g' in `gen` anyway.

---
___
*[.ci/lib/stage-sgx-gen-key.jenkinsfile, line 1 at r4](https://reviewable.io/reviews/gramineproject/gramine/252#-MvzAYosBwipGLIkVsAw:-Mw-97vAAt4z5Gedm2KB:bffr54l) ([raw file](https://github.com/gramineproject/gramine/blob/21112f739503652bc272d9e5fd43ba54e146e346/.ci/lib/stage-sgx-gen-key.jenkinsfile#L1)):*

Is it worth creating another jenkinsfile for this? I get that modularity is nice, but this is SGX specific and cannot be reused anywhere. I think leaving this in building (more precisely installing) steps is fine.
</blockquote></details>

"build" is not a good place, because when I'll split the jenkins pipeline to build package/install and test, it won't be packaged. It should be in something testing-related, but it can be in neither "test" nor "test-sgx": we have one of those in sgx and -apps pipelines, but both in -sanitizers, so either the key is not generated in one of the pipelines, or sanitizers attempt to generate the key twice, and it fails because the key already exists (as it should). This is the reason why I have pushed so many times to jenkins.

Another solution would be to have some switch in the tool itself to no-op when the key exists (and not fail).
___
*[Documentation/devel/building.rst, line 132 at r4](https://reviewable.io/reviews/gramineproject/gramine/252#-MvzBjc-3hkLuTvjzZeT:-Mw-Ie3D0zz3irM6CiDL:bfznqes) ([raw file](https://github.com/gramineproject/gramine/blob/9eafa5f7731650dc54043e4aa4804b6a2e34e292/Documentation/devel/building.rst#L132)):*

How does an user which goes by full documentation (not quickstart) knows they need to create such a key?
Please leave this information here or add somewhere else (but I don't think we have a suitable place at the moment)
</blockquote></details>

The key is not a requirement for building (it was before). Note this document doesn't mention running any tests or examples.
___
*[python/graminelibos/sgx_sign.py, line 26 at r6](https://reviewable.io/reviews/gramineproject/gramine/252#-Mvzs4Qj8MnkQL1E5KTm:-Mw-4IR_093s3EwuoRkt:bc8vyaw) ([raw file](https://github.com/gramineproject/gramine/blob/a43fcbda34e144ee955d05fae2186d33604d287f/python/graminelibos/sgx_sign.py#L26)):*

`crypto_backend`? With this name reader has no idea what it's about.
Also why import `backends`? This is a pretty generic name, I think `cryptography.hazmat.backends` would look better (and it's only used here).
</blockquote></details>

`backend=` mandatory argument goes away in newer cryptography version, because it was a misfeature: https://cryptography.io/en/2.8/hazmat/backends/. Can't find the exact commit, but it's optional in 3.3 (Ubuntu 21.04) and entirely missing in latest 36 (not yet packaged).

The problem is, cryptography.has.those.very.nested.package.names so there's a convention that you `from cryptography... import` the actual module. Otherwise we'd have monstrosities like:
```python
cryptography.hazmat.primitives.asymmetric.rsa.generate_private_key(
    public_exponent=3, key_size=3072
).private_bytes(
    encoding=cryptography.hazmat.primitives.serialization.Encoding.PEM,
    format=cryptography.hazmat.primitives.serialization.PrivateFormat.TraditionalOpenSSL,
    encryption_algorithm=cryptography.hazmat.primitives.serialization.NoEncryption())
)
```

(X.509 generation, which we don't do, is even worse, where you have classes, modules and enums for all X.509 v3 extensions).

Examples in the cryptography.io documentation even import just the function `default_backend()`: https://cryptography.io/en/2.8/hazmat/primitives/asymmetric/rsa/#cryptography.hazmat.primitives.asymmetric.rsa.generate_private_key. I could do this.
___
*[python/gramine-sgx-gen-private-key, line 23 at r4](https://reviewable.io/reviews/gramineproject/gramine/252#-Mvz5Ch14DsyQqxCe0m8:-Mw-B6ZKBhoSc1_Zc4Ww:bim31nn) ([raw file](https://github.com/gramineproject/gramine/blob/21112f739503652bc272d9e5fd43ba54e146e346/python/gramine-sgx-gen-private-key#L23)):*

> I can also do `S_IRWXG | S_IRWXO`

Please don't, it's actually much less readable IMO ;) See https://lwn.net/Articles/696231/ (actually these `PERM_*` macros are what we adopted in Gramine C code).
</blockquote></details>

Oh, no. I hate those `PERM_` macros. One underline miscounted and baam, undefined variable. Fortunately they're not exposed in python.
___
*[python/gramine-sgx-gen-private-key, line 11 at r6](https://reviewable.io/reviews/gramineproject/gramine/252#-MvztPh20RPXOzsgaRI-:-Mw-Al_lEqHEGQ0A3STT:bq1x7o6) ([raw file](https://github.com/gramineproject/gramine/blob/a43fcbda34e144ee955d05fae2186d33604d287f/python/gramine-sgx-gen-private-key#L11)):*

Shouldn't these tools use only things directly exported by `graminelibos`, not poke into specific modules/parts of implementation?
</blockquote></details>

No, why? I think `sgx_sign` is OK to be imported and used.
___
*[python/gramine-sgx-gen-private-key, line 23 at r6](https://reviewable.io/reviews/gramineproject/gramine/252#-MvzrDbJ8ImYGJAm0j2p:-Mw-BBEH3s4YxbX_IMO2:b-f31k69) ([raw file](https://github.com/gramineproject/gramine/blob/a43fcbda34e144ee955d05fae2186d33604d287f/python/gramine-sgx-gen-private-key#L23)):*

Why not set mode in calls to mkdir and file creation?
</blockquote></details>

It's usually done with `umask`, because it works once for all calls between this and restoring `old_umask`. Also, there's no such argument to either python's `open()` or click's `open_file()`, we'd have to `open(os.open())` and watch ourselves for `-`, which click's version handles for us.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/252)* status: all files reviewed, 8 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @jinengandhi-intel)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/252#-MvzuS0n1pai6b6Drm4i:-Mw13qpEAUktcHfuQD6Z:b-clg32p) (no related file):*

Is this really a problem? If yes, I can revert the change in `util_tests.py`, only change the default to be consistent with this PR, but still allow override via env `SGX_SIGNER_KEY=/abc`.

cc @pwmarcz
</blockquote></details>

+1 from me. And more importantly, I would prefer `util_tests.py` to still add the dependency on the key file to Ninja. Otherwise, if you change the key, the manifests will not get rebuilt. This is a rare corner case, but corner cases like this are what made the old Makefiles so annoying.

---

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/252)* status: 36 of 37 files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, @jinengandhi-intel, and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/252#-MvzuS0n1pai6b6Drm4i:-Mw1LYrp9Rh_4CDyUHAI:basoa73) (no related file):*

+1 from me. And more importantly, I would prefer `util_tests.py` to still add the dependency on the key file to Ninja. Otherwise, if you change the key, the manifests will not get rebuilt. This is a rare corner case, but corner cases like this are what made the old Makefiles so annoying.
</blockquote></details>

Oh, right. The dependency part is important.

---

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/252)* status: 36 of 37 files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, @jinengandhi-intel, and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/252#-MvzuS0n1pai6b6Drm4i:-Mw1tjz0622PJmIUniVf:b-896fix) (no related file):*

Oh, right. The dependency part is important.
</blockquote></details>

Done.

---

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/252)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @jinengandhi-intel, and @woju)

___
*[.ci/lib/stage-sgx-gen-key.jenkinsfile, line 1 at r4](https://reviewable.io/reviews/gramineproject/gramine/252#-MvzAYosBwipGLIkVsAw:-Mw3-Ydi-kJGTZfsQv6k:b-ufawha) ([raw file](https://github.com/gramineproject/gramine/blob/21112f739503652bc272d9e5fd43ba54e146e346/.ci/lib/stage-sgx-gen-key.jenkinsfile#L1)):*
> more precisely installing

Why cannot you put it in install step after split, like suggested in first message?
___
*[Documentation/devel/building.rst, line 132 at r4](https://reviewable.io/reviews/gramineproject/gramine/252#-MvzBjc-3hkLuTvjzZeT:-Mw3-kX1DQ6HsdFLsGBK:b-6s6hj5) ([raw file](https://github.com/gramineproject/gramine/blob/9eafa5f7731650dc54043e4aa4804b6a2e34e292/Documentation/devel/building.rst#L132)):*

The key is not a requirement for building (it was before). Note this document doesn't mention running any tests or examples.
</blockquote></details>

It is a requirement for running **anything**. Also half of this file is not required if you only want to build (not run), e.g. fsgs support.
___
*[python/graminelibos/sgx_sign.py, line 26 at r6](https://reviewable.io/reviews/gramineproject/gramine/252#-Mvzs4Qj8MnkQL1E5KTm:-Mw326XeEXCK-gsGLaz-:b-op9t9a) ([raw file](https://github.com/gramineproject/gramine/blob/a43fcbda34e144ee955d05fae2186d33604d287f/python/graminelibos/sgx_sign.py#L26)):*
> The problem is, cryptography.has.those.very.nested.package.names so there's a convention that you from cryptography... import the actual module.

That's why I didn't mention `serialization`, it would be PITA.  I just suggested:
```python
import cryptography
crypto_backend = cryptography.hazmat.backends.default_backend()
```
so that it's less confusing.
___
*[python/graminelibos/util_tests.py, line 80 at r7](https://reviewable.io/reviews/gramineproject/gramine/252#-Mw32ye09GYP-gDjZWht:-Mw32ye09GYP-gDjZWhu:bdcdbat) ([raw file](https://github.com/gramineproject/gramine/blob/a4a88af41f6b01e72ba8a1a4324443c3e2a50848/python/graminelibos/util_tests.py#L80)):*
> ```Python
>         self.key = os.environ.get('SGX_SIGNER_KEY', None)
>         if not self.key:
>             self.key = os.fspath(_SGX_RSA_KEY_PATH)
> ```

Why not:
```python
if not self.key:
    try:
        from .sgx_sign import SGX_RSA_KEY_PATH as _SGX_RSA_KEY_PATH
        self.key = os.fspath(SGX_RSA_KEY_PATH)
    except:
        raise KEYNOTFOUNDSUPEREXCEPTION
```
instead of import at top-level?
___
*[python/gramine-sgx-gen-private-key, line 23 at r4](https://reviewable.io/reviews/gramineproject/gramine/252#-Mvz5Ch14DsyQqxCe0m8:-Mw30kaQ9ZCHU0BYdlvy:b8d4p0p) ([raw file](https://github.com/gramineproject/gramine/blob/21112f739503652bc272d9e5fd43ba54e146e346/python/gramine-sgx-gen-private-key#L23)):*

Oh, no. I hate those `PERM_` macros. One underline miscounted and baam, undefined variable. Fortunately they're not exposed in python.
</blockquote></details>

In C (unlike in python) it's detected at compile time, so it's all good. If you make a mistake in `0123` then you will never know.
___
*[python/gramine-sgx-gen-private-key, line 11 at r6](https://reviewable.io/reviews/gramineproject/gramine/252#-MvztPh20RPXOzsgaRI-:-Mw30TSvECU1cIwG-W4-:b-ijcvf7) ([raw file](https://github.com/gramineproject/gramine/blob/a43fcbda34e144ee955d05fae2186d33604d287f/python/gramine-sgx-gen-private-key#L11)):*

No, why? I think `sgx_sign` is OK to be imported and used.
</blockquote></details>

It's ok if we do it (because we provide both parts and control the API). I just wouldn't want users to check this file and do the same, because `sgx_sign` is not part of our API.

dimakuv (changes requested):
> Reviewed 19 of 26 files at r2, 10 of 12 files at r4, 4 of 4 files at r5, 1 of 1 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/252)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @jinengandhi-intel, @mkow, and @woju)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/252#-MvzunmHAFqAc35aIdcb:-Mw5wa6mCI7mrTx9fnxo:b-s9y7lq) (no related file):*

No strong opinion. Shell completion works from `g' in `gen` anyway.
</blockquote></details>

I wouldn't care saving 3 letters in an already-long name :)

---
___
*[Documentation/quickstart.rst, line 45 at r7](https://reviewable.io/reviews/gramineproject/gramine/252#-Mw5xpim7vuIsj-vgD6-:-Mw5xpim7vuIsj-vgD60:b-9slt6x) ([raw file](https://github.com/gramineproject/gramine/blob/a4a88af41f6b01e72ba8a1a4324443c3e2a50848/Documentation/quickstart.rst#L45)):*
> ```reStructuredText
> ---------------------
> 
> Only for SGX, and if you haven't already::
> ```

What about "Cloud deployment"? https://github.com/gramineproject/gramine/blob/95b934de1dc67d0a2499a224a72d68b817d39d24/Documentation/cloud-deployment.rst

Also, we don't have any Gramine package that actually contains this tool, so the documentation steps will not work. Shouldn't we split these docs changes until we have Gramine v1.2 ?
___
*[Documentation/devel/building.rst, line 132 at r4](https://reviewable.io/reviews/gramineproject/gramine/252#-MvzBjc-3hkLuTvjzZeT:-Mw5y_ny0TDmO4GXK-P7:bievqb3) ([raw file](https://github.com/gramineproject/gramine/blob/9eafa5f7731650dc54043e4aa4804b6a2e34e292/Documentation/devel/building.rst#L132)):*

It is a requirement for running **anything**. Also half of this file is not required if you only want to build (not run), e.g. fsgs support.
</blockquote></details>

+1 to Borys

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/252)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, @jinengandhi-intel, and @mkow)

___
*[.ci/lib/stage-sgx-gen-key.jenkinsfile, line 1 at r4](https://reviewable.io/reviews/gramineproject/gramine/252#-MvzAYosBwipGLIkVsAw:-MwCPMtX3FEbDoUuLPeb:b-5e3xg0) ([raw file](https://github.com/gramineproject/gramine/blob/21112f739503652bc272d9e5fd43ba54e146e346/.ci/lib/stage-sgx-gen-key.jenkinsfile#L1)):*

> more precisely installing

Why cannot you put it in install step after split, like suggested in first message?
</blockquote></details>

Hmm, yes, that also works.
___
*[Documentation/quickstart.rst, line 45 at r7](https://reviewable.io/reviews/gramineproject/gramine/252#-Mw5xpim7vuIsj-vgD6-:-Mw73YIn8yTnqUEvgTWD:bwrx9c0) ([raw file](https://github.com/gramineproject/gramine/blob/a4a88af41f6b01e72ba8a1a4324443c3e2a50848/Documentation/quickstart.rst#L45)):*

What about "Cloud deployment"? https://github.com/gramineproject/gramine/blob/95b934de1dc67d0a2499a224a72d68b817d39d24/Documentation/cloud-deployment.rst

Also, we don't have any Gramine package that actually contains this tool, so the documentation steps will not work. Shouldn't we split these docs changes until we have Gramine v1.2 ?
</blockquote></details>

Do you deploy signing key to the cloud, aka "other people's computers"?

Yes, we probably should have "stable" documents, tracking last packaged version, and don't link to "latest" everywhere.
___
*[Documentation/devel/building.rst, line 132 at r4](https://reviewable.io/reviews/gramineproject/gramine/252#-MvzBjc-3hkLuTvjzZeT:-MwCTgseEBK-WBzTQ-_5:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9eafa5f7731650dc54043e4aa4804b6a2e34e292/Documentation/devel/building.rst#L132)):*

+1 to Borys
</blockquote></details>

Done.
___
*[python/graminelibos/sgx_sign.py, line 26 at r6](https://reviewable.io/reviews/gramineproject/gramine/252#-Mvzs4Qj8MnkQL1E5KTm:-MwCUCNS5Vb4Kemn08M6:b-h2pf1y) ([raw file](https://github.com/gramineproject/gramine/blob/a43fcbda34e144ee955d05fae2186d33604d287f/python/graminelibos/sgx_sign.py#L26)):*

> The problem is, cryptography.has.those.very.nested.package.names so there's a convention that you from cryptography... import the actual module.

That's why I didn't mention `serialization`, it would be PITA.  I just suggested:
```python
import cryptography
crypto_backend = cryptography.hazmat.backends.default_backend()
```
so that it's less confusing.
</blockquote></details>

`import cryptography` won't work, we'd have to `import cryptography.hazmat.backends`. I don't want to do that, because that would be inconsistent, because we `from import` everything apart from this one.

I've renamed to `_cryptography_backend`, that I can do.
___
*[python/graminelibos/util_tests.py, line 80 at r7](https://reviewable.io/reviews/gramineproject/gramine/252#-Mw32ye09GYP-gDjZWht:-Mw72Q_lA0vGHacDUK6f:bhjk9qz) ([raw file](https://github.com/gramineproject/gramine/blob/a4a88af41f6b01e72ba8a1a4324443c3e2a50848/python/graminelibos/util_tests.py#L80)):*

Why not:
```python
if not self.key:
    try:
        from .sgx_sign import SGX_RSA_KEY_PATH as _SGX_RSA_KEY_PATH
        self.key = os.fspath(SGX_RSA_KEY_PATH)
    except:
        raise KEYNOTFOUNDSUPEREXCEPTION
```
instead of import at top-level?
</blockquote></details>

1) because local imports are last resort, mainly used to break import cycles; here we don't have import cycle
2) because this will always trigger, as `self.key` is always filled and present in `.ninja` file, even withoug SGX — just the rules don't get executed; we'd have to have this try/except (or equivalent `if CONFIG_SGX_ENABLED`) in few other places where we generate rules
___
*[python/gramine-sgx-gen-private-key, line 11 at r6](https://reviewable.io/reviews/gramineproject/gramine/252#-MvztPh20RPXOzsgaRI-:-Mw6WGJJ4DIJQz4nP_Sf:b-w2zsi3) ([raw file](https://github.com/gramineproject/gramine/blob/a43fcbda34e144ee955d05fae2186d33604d287f/python/gramine-sgx-gen-private-key#L11)):*

It's ok if we do it (because we provide both parts and control the API). I just wouldn't want users to check this file and do the same, because `sgx_sign` is not part of our API.
</blockquote></details>

I would like, eventually, to have a documented `graminelibos.sgx` module, to avoid the `if CONFIG_SGX_ENABLED` in `__init__.py`. But that's obviously out of scope here.

boryspoplawski (changes requested):
> Reviewed 9 of 9 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/252)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, @jinengandhi-intel, and @mkow)

___
*[Documentation/devel/building.rst, line 132 at r4](https://reviewable.io/reviews/gramineproject/gramine/252#-MvzBjc-3hkLuTvjzZeT:-MwJMq9ECPLjFjwSOwdy:b-bqd2ea) ([raw file](https://github.com/gramineproject/gramine/blob/9eafa5f7731650dc54043e4aa4804b6a2e34e292/Documentation/devel/building.rst#L132)):*

Done.
</blockquote></details>

Ok, but why did you move it? It's now after some additional build options, to me it fit well here, near SGX related things.

dimakuv (changes requested):
> Reviewed 3 of 9 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/252)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @jinengandhi-intel, @mkow, @pwmarcz, and @woju)

___
*[Documentation/quickstart.rst, line 45 at r7](https://reviewable.io/reviews/gramineproject/gramine/252#-Mw5xpim7vuIsj-vgD6-:-MwRBRBA1uux4EYwrWK_:bsflurf) ([raw file](https://github.com/gramineproject/gramine/blob/a4a88af41f6b01e72ba8a1a4324443c3e2a50848/Documentation/quickstart.rst#L45)):*
> Do you deploy signing key to the cloud, aka "other people's computers"?

I actually wonder why we have that "Cloud deployment" page at all. It has zero interesting information; all the same could be done by reading "Quick start" and "Building" pages. What do others think, maybe we simply remove this page? @mkow @boryspoplawski @pwmarcz 

> Yes, we probably should have "stable" documents, tracking last packaged version, and don't link to "latest" everywhere.

So what would be the step to resolve my comment? I guess you should split this change in a separate PR and mark this PR as *for v1.2*?

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/252)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @jinengandhi-intel, @mkow, and @pwmarcz)

___
*[Documentation/quickstart.rst, line 45 at r7](https://reviewable.io/reviews/gramineproject/gramine/252#-Mw5xpim7vuIsj-vgD6-:-MwRRQqbC24pFvm_SWOW:bs7577d) ([raw file](https://github.com/gramineproject/gramine/blob/a4a88af41f6b01e72ba8a1a4324443c3e2a50848/Documentation/quickstart.rst#L45)):*

> I actually wonder why we have that "Cloud deployment" page at all. It has zero interesting information; all the same could be done by reading "Quick start" and "Building" pages. What do others think, maybe we simply remove this page? @mkow @boryspoplawski @pwmarcz 

+1 to removing it if you say it's redundant. I'm not into using Gramine with cloud providers, so I'll rely on your judgement ;)

> I guess you should split this change in a separate PR and mark this PR as for v1.2?

No, please keep the docs in the repo synced with master. If someone needs the docs for the latest release, then we should just point them to https://gramine.readthedocs.io/en/v1.1/.
</blockquote></details>

This is not mutually exclusive, we can have https://gramine.rtfd.io/en/stable/ for people to bookmark, to point to the latest version in the package repo, whatever that version would be (/v1.1/ won't be the current released version forever). "latest" should obviously be master, as it is currently.

I would be fine with removing `cloud-deployment.rst` in a separate PR, then rebasing this PR on top of said separate PR.
___
*[Documentation/devel/building.rst, line 132 at r4](https://reviewable.io/reviews/gramineproject/gramine/252#-MvzBjc-3hkLuTvjzZeT:-MwRS85nEncgBwEN_7Gy:b57jvza) ([raw file](https://github.com/gramineproject/gramine/blob/9eafa5f7731650dc54043e4aa4804b6a2e34e292/Documentation/devel/building.rst#L132)):*

Ok, but why did you move it? It's now after some additional build options, to me it fit well here, near SGX related things.
</blockquote></details>

Because you need to install the tool to execute it, so it needs to be after build and install for people who read the document top-to-bottom (previously you didn't need to install gramine itself, because key generation used openssl, installed as part of dependencies above). Current document headings have building options as a subsection to "Building", so it needs to be after that. I can rearrange the document so that building options are another section at the same level as building, after key generation.
___
*[python/gramine-sgx-gen-private-key, line 23 at r4](https://reviewable.io/reviews/gramineproject/gramine/252#-Mvz5Ch14DsyQqxCe0m8:-MwRT7CTC6qxIyGuHjyi:bpzn3wt) ([raw file](https://github.com/gramineproject/gramine/blob/21112f739503652bc272d9e5fd43ba54e146e346/python/gramine-sgx-gen-private-key#L23)):*

+1 to Borys, better to fail compile-time than ship broken code.
</blockquote></details>

In python we have compile-time linter, so if you reference the non-existent constant from a module, it will similarly fail. But it doesn't change my hatred towards those constants, which I find over the top. "0o77" is readable enough.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/252)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @jinengandhi-intel, @mkow, and @pwmarcz)

___
*[python/gramine-sgx-gen-private-key, line 23 at r4](https://reviewable.io/reviews/gramineproject/gramine/252#-Mvz5Ch14DsyQqxCe0m8:-MwRV_gICerhjYt6Pw_R:b5tmams) ([raw file](https://github.com/gramineproject/gramine/blob/21112f739503652bc272d9e5fd43ba54e146e346/python/gramine-sgx-gen-private-key#L23)):*

In python we have compile-time linter, so if you reference the non-existent constant from a module, it will similarly fail. But it doesn't change my hatred towards those constants, which I find over the top. "0o77" is readable enough.
</blockquote></details>

I'm not against `0o770` notation here, I'm against these unreadable `S_IRWXG | S_IRWXO` macros. I'm fine with both octal and `PERM_` (and here we use octal because we don't have `PERM_` defined for Python right now).

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/252)* status: all files reviewed, 5 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @jinengandhi-intel, and @mkow)

___
*[Documentation/quickstart.rst, line 45 at r7](https://reviewable.io/reviews/gramineproject/gramine/252#-Mw5xpim7vuIsj-vgD6-:-MwRdlAS4UcJqEgOTxyf:b-4xcfha) ([raw file](https://github.com/gramineproject/gramine/blob/a4a88af41f6b01e72ba8a1a4324443c3e2a50848/Documentation/quickstart.rst#L45)):*
> I would be fine with removing cloud-deployment.rst in a separate PR, then rebasing this PR on top of said separate PR.

No, no need to rebase your PR or modify it in any way. I'll submit a separate PR to remove `cloud-deployment.rst` at some point.

> No, please keep the docs in the repo synced with master.

Ok, I'm fine with it. So for now I'll point people to https://gramine.readthedocs.io/en/v1.1/ to read about the stable release usage.

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/252)* status: all files reviewed, 5 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @jinengandhi-intel, and @mkow)

___
*[Documentation/quickstart.rst, line 45 at r7](https://reviewable.io/reviews/gramineproject/gramine/252#-Mw5xpim7vuIsj-vgD6-:-MwSRFGcCmKu3fJDXyV4:b-pzfifu) ([raw file](https://github.com/gramineproject/gramine/blob/a4a88af41f6b01e72ba8a1a4324443c3e2a50848/Documentation/quickstart.rst#L45)):*

> I would be fine with removing cloud-deployment.rst in a separate PR, then rebasing this PR on top of said separate PR.

No, no need to rebase your PR or modify it in any way. I'll submit a separate PR to remove `cloud-deployment.rst` at some point.

> No, please keep the docs in the repo synced with master.

Ok, I'm fine with it. So for now I'll point people to https://gramine.readthedocs.io/en/v1.1/ to read about the stable release usage.
</blockquote></details>

Sure.

/stable/ also works, points to 1.1.

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/252)* status: all files reviewed, 3 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @jinengandhi-intel, and @mkow)

mkow (changes requested):
> Reviewed 1 of 1 files at r7, 9 of 9 files at r8, 7 of 7 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/252)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski, @jinengandhi-intel, @mkow, and @woju)

___
*[Documentation/devel/building.rst, line 280 at r8](https://reviewable.io/reviews/gramineproject/gramine/252#-MwWmlQ3Dvgq6MgpYzG2:-MwWmlQ3Dvgq6MgpYzG3:b-cageyi) ([raw file](https://github.com/gramineproject/gramine/blob/592f631a3692eb2f3611c90360785a4b235e832f/Documentation/devel/building.rst#L280)):*
> ```reStructuredText
>    gramine-sgx-gen-private-key
> 
> This command generates a |~| RSA 3072 key suitable for signing SGX enclaves and
> ```

Should be "an".



_Code quote:_
```reStructuredText
a |~| RSA 3072
```
___
*[python/graminelibos/sgx_sign.py, line 8 at r9](https://reviewable.io/reviews/gramineproject/gramine/252#-MwWnYwF0kSsqufiEotX:-MwWnYwF0kSsqufiEotY:b-3vmjqc) ([raw file](https://github.com/gramineproject/gramine/blob/8342e1d72322ba8dd3a68d908132105597b22b42/python/graminelibos/sgx_sign.py#L8)):*
> ```Python
> # Copyright (C) 2021 Intel Corporation
> #                    Borys Popławski <borysp@invisiblethingslab.com>
> # Copyright (c) 2021-2022 Intel Corporation
> ```

Should be only `2022` and merged with the copyrights above (see the new rules)

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/252)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski, @jinengandhi-intel, and @mkow)

___
*[Documentation/devel/building.rst, line 280 at r8](https://reviewable.io/reviews/gramineproject/gramine/252#-MwWmlQ3Dvgq6MgpYzG2:-MwWp2G8319tIMuRCnsB:b-elhus3) ([raw file](https://github.com/gramineproject/gramine/blob/592f631a3692eb2f3611c90360785a4b235e832f/Documentation/devel/building.rst#L280)):*

Should be "an".
</blockquote></details>

Done also in `quickstart.rst`.
___
*[python/graminelibos/sgx_sign.py, line 8 at r9](https://reviewable.io/reviews/gramineproject/gramine/252#-MwWnYwF0kSsqufiEotX:-MwWp1G94f_HZ8uNL60r:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/8342e1d72322ba8dd3a68d908132105597b22b42/python/graminelibos/sgx_sign.py#L8)):*

Should be only `2022` and merged with the copyrights above (see the new rules)
</blockquote></details>

Done.

mkow (approved):
> Reviewed 3 of 3 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/252)* status: all files reviewed, 2 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski and @jinengandhi-intel)

boryspoplawski (approved):
> Reviewed 5 of 7 files at r9, 3 of 3 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/252)* status: all files reviewed, 1 unresolved discussion (waiting on @jinengandhi-intel)

PR #255, opened 2021-11-30
[Pal,LibOS] rtld: Rename `l_addr`/`l_base` to `l_base_diff`

dimakuv:
## Description of the changes 

Previously, `l_addr`/`l_base` field names were used to denote the "difference between virtual addresses in ELF file and actual virtual addresses in memory". This commit renames them to `l_base_diff` which reflects the actual meaning better.

Link to ELF spec, see page 73: https://refspecs.linuxfoundation.org/elf/elf.pdf.

## How to test this PR? 

All tests must pass, no functional change.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/255)

### DIFF
--- a/LibOS/shim/src/shim_rtld.c
+++ b/LibOS/shim/src/shim_rtld.c
@@ -41,8 +41,23 @@
  * Structure describing a loaded ELF object. Originally based on glibc link_map structure.
  */
 struct link_map {
-    /* Base address shared object is loaded at. */
-    ElfW(Addr) l_addr;
+    /*
+     * Difference between virtual addresses (p_vaddr) in ELF file and actual virtual addresses in
+     * memory. Equal to 0x0 for ET_EXECs.
+     *
+     * Note that `l_base_diff` name is a departure from the ELF standard and Glibc code: the ELF
+     * standard uses the term "Base Address" and Glibc uses `l_addr`. We find ELF/Glibc terms
+     * misleading and therefore replace them with `l_base_diff`. For the context, below is the
+     * exerpt from the ELF spec, Section "Program Header", Sub-section "Base Address":
+     *
+     *   The virtual addresses in the program headers might not represent the actual virtual
+     *   addresses of the program's memory image. [...] The difference between the virtual address
+     *   of any segment in memory and the corresponding virtual address in the file is thus a single
+     *   constant value for any one executable or shared object in a given process. This difference
+     *   is the *base address*. One use of the base address is to relocate the memory image of the
+     *   program during dynamic linking.
+     */
+    ElfW(Addr) l_base_diff;
 
     /* Object identifier: file path, or PAL URI if path is unavailable. */
     const char* l_name;
@@ -56,8 +71,7 @@ struct link_map {
     /* Number of program header entries.  */
     ElfW(Half) l_phnum;
 
-    /* Start and finish of memory map for this object.  l_map_start need not be the same as
-     * l_addr. */
+    /* Start and finish of memory map for this object. */
     ElfW(Addr) l_map_start, l_map_end;
 
     const char* l_interp_libname;
@@ -77,7 +91,7 @@ struct loadcmd {
      *   - start, map_end, alloc_end are page-aligned
      *   - map_off is page-aligned
      *
-     * The addresses are not relocated (i.e. you need to add l_addr to them).
+     * The addresses are not relocated (i.e. you need to add l_base_diff to them).
      */
 
     /* Start of memory area */
@@ -232,7 +246,7 @@ static int reserve_dyn(size_t total_size, void** addr) {
  * This function doesn't undo allocations in case of error: if it fails, it may leave some segments
  * already allocated.
  */
-static int execute_loadcmd(const struct loadcmd* c, ElfW(Addr) load_addr,
+static int execute_loadcmd(const struct loadcmd* c, ElfW(Addr) base_diff,
                            struct shim_handle* file) {
     int ret;
     int map_flags = MAP_FIXED | MAP_PRIVATE;
@@ -240,7 +254,7 @@ static int execute_loadcmd(const struct loadcmd* c, ElfW(Addr) load_addr,
 
     /* Map the part that should be loaded from file, rounded up to page size. */
     if (c->start < c->map_end) {
-        void* map_start = (void*)(load_addr + c->start);
+        void* map_start = (void*)(c->start + base_diff);
         size_t map_size = c->map_end - c->start;
 
         if ((ret = bkeep_mmap_fixed(map_start, map_size, c->prot, map_flags, file, c->map_off,
@@ -259,7 +273,7 @@ static int execute_loadcmd(const struct loadcmd* c, ElfW(Addr) load_addr,
     /* Zero out the extra data at the end of mapped area. If necessary, temporarily remap the last
      * page as writable. */
     if (c->data_end < c->map_end) {
-        void* zero_start = (void*)(load_addr + c->data_end);
+        void* zero_start = (void*)(c->data_end + base_diff);
         size_t zero_size = c->map_end - c->data_end;
         void* last_page_start = ALLOC_ALIGN_DOWN_PTR(zero_start);
 
@@ -283,7 +297,7 @@ static int execute_loadcmd(const struct loadcmd* c, ElfW(Addr) load_addr,
 
     /* Allocate extra pages after the mapped area. */
     if (c->map_end < c->alloc_end) {
-        void* zero_page_start = (void*)(load_addr + c->map_end);
+        void* zero_page_start = (void*)(c->map_end + base_diff);
         size_t zero_page_size = c->alloc_end - c->map_end;
         int zero_map_flags = MAP_FIXED | MAP_PRIVATE | MAP_ANONYMOUS;
         pal_prot_flags_t zero_pal_prot = LINUX_PROT_TO_PAL(c->prot, zero_map_flags);
@@ -375,7 +389,7 @@ static struct link_map* map_elf_object(struct shim_handle* file, ElfW(Ehdr)* ehd
          * address.
          *
          * Note that we reserve memory starting from offset 0, not from load_start. This is to
-         * ensure that the load base (l_addr) will not be lower than 0.
+         * ensure that l_base_diff will not be less than 0.
          */
         void* addr;
 
@@ -384,17 +398,19 @@ static struct link_map* map_elf_object(struct shim_handle* file, ElfW(Ehdr)* ehd
             goto err;
         }
 
-        l->l_addr = (ElfW(Addr))addr;
+        l->l_base_diff = (ElfW(Addr))addr;
     } else {
-        l->l_addr = 0;
+        /* This is a non-pie shared object (EXEC, executable), so the difference between virtual
+         * addresses (p_vaddr) in the ELF file and the actual addresses in memory is zero. */
+        l->l_base_diff = 0x0;
     }
-    l->l_map_start = load_start + l->l_addr;
-    l->l_map_end   = load_end + l->l_addr;
+    l->l_map_start = load_start + l->l_base_diff;
+    l->l_map_end   = load_end + l->l_base_diff;
 
     /* Execute load commands. */
     l->l_data_segment_size = 0;
     for (struct loadcmd* c = &loadcmds[0]; c < &loadcmds[n_loadcmds]; c++) {
-        if ((ret = execute_loadcmd(c, l->l_addr, file)) < 0) {
+        if ((ret = execute_loadcmd(c, l->l_base_diff, file)) < 0) {
             errstring = "failed to execute load command";
             goto err;
         }
@@ -403,14 +419,13 @@ static struct link_map* map_elf_object(struct shim_handle* file, ElfW(Ehdr)* ehd
                 && ehdr->e_phoff + phdr_size <= c->map_off + (c->data_end - c->start)) {
             /* Found the program header in this segment. */
             ElfW(Addr) phdr_vaddr = ehdr->e_phoff - c->map_off + c->start;
-            l->l_phdr = (ElfW(Phdr)*)(phdr_vaddr + l->l_addr);
+            l->l_phdr = (ElfW(Phdr)*)(phdr_vaddr + l->l_base_diff);
         }
-
 
         if (interp_libname_vaddr != 0 && !l->l_interp_libname && c->start <= interp_libname_vaddr
                 && interp_libname_vaddr < c->data_end) {
             /* Found the interpreter name in this segment (but we need to validate length). */
-            const char* interp_libname = (const char*)(interp_libname_vaddr + l->l_addr);
+            const char* interp_libname = (const char*)(interp_libname_vaddr + l->l_base_diff);
             size_t maxlen = c->data_end - interp_libname_vaddr;
             size_t len = strnlen(interp_libname, maxlen);
             if (len == maxlen) {
@@ -424,7 +439,7 @@ static struct link_map* map_elf_object(struct shim_handle* file, ElfW(Ehdr)* ehd
         if (ehdr->e_entry != 0 && !l->l_entry && c->start <= ehdr->e_entry
                 && ehdr->e_entry < c->data_end) {
             /* Found the entry point in this segment. */
-            l->l_entry = (ElfW(Addr))(ehdr->e_entry + l->l_addr);
+            l->l_entry = (ElfW(Addr))(ehdr->e_entry + l->l_base_diff);
         }
 
         if (!(c->prot & PROT_EXEC))
@@ -468,7 +483,7 @@ err:
 }
 
 static void remove_elf_object(struct link_map* l) {
-    remove_r_debug((void*)l->l_addr);
+    remove_r_debug((void*)l->l_base_diff);
     free(l);
 }
 
@@ -624,7 +639,7 @@ int load_elf_object(struct shim_handle* file, struct link_map** out_map) {
     map->l_file = file;
 
     if (map->l_file && !qstrempty(&map->l_file->uri)) {
-        append_r_debug(qstrgetstr(&map->l_file->uri), (void*)map->l_addr);
+        append_r_debug(qstrgetstr(&map->l_file->uri), (void*)map->l_base_diff);
     }
 
     *out_map = map;
@@ -882,7 +897,7 @@ noreturn void execute_elf_object(struct link_map* exec_map, void* argp, ElfW(aux
     auxp[3].a_type     = AT_ENTRY;
     auxp[3].a_un.a_val = g_exec_map->l_entry;
     auxp[4].a_type     = AT_BASE;
-    auxp[4].a_un.a_val = g_interp_map ? g_interp_map->l_addr : 0;
+    auxp[4].a_un.a_val = g_interp_map ? g_interp_map->l_map_start : 0;
     auxp[5].a_type     = AT_RANDOM;
     auxp[5].a_un.a_val = 0; /* filled later */
     if (vdso_addr) {
@@ -955,7 +970,6 @@ BEGIN_RS_FUNC(elf_object) {
 
     CP_REBASE(map->l_name);
     CP_REBASE(map->l_file);
-    DEBUG_RS("base=0x%08lx,name=%s", map->l_addr, map->l_name);
 }
 END_RS_FUNC(elf_object)
 
--- a/Pal/include/pal_rtld.h
+++ b/Pal/include/pal_rtld.h
@@ -11,12 +11,27 @@
 #include "api.h"
 #include "elf/elf.h"
 
-enum elf_object_type { ELF_OBJECT_INTERNAL, ELF_OBJECT_EXEC, ELF_OBJECT_PRELOAD };
-
 /* Loaded shared object. */
 struct link_map {
-    ElfW(Addr)  l_addr;       /* Address shared object (its first LOAD segment) is loaded at. */
-    ElfW(Addr)  l_base;       /* Base address (0x0 for EXECs, same as l_addr for DYNs). */
+    /*
+     * Difference between virtual addresses (p_vaddr) in ELF file and actual virtual addresses in
+     * memory. Equal to 0x0 for ET_EXECs.
+     *
+     * Note that `l_base_diff` name is a departure from the ELF standard and Glibc code: the ELF
+     * standard uses the term "Base Address" and Glibc uses `l_addr`. We find ELF/Glibc terms
+     * misleading and therefore replace them with `l_base_diff`. For the context, below is the
+     * exerpt from the ELF spec, Section "Program Header", Sub-section "Base Address":
+     *
+     *   The virtual addresses in the program headers might not represent the actual virtual
+     *   addresses of the program's memory image. [...] The difference between the virtual address
+     *   of any segment in memory and the corresponding virtual address in the file is thus a single
+     *   constant value for any one executable or shared object in a given process. This difference
+     *   is the *base address*. One use of the base address is to relocate the memory image of the
+     *   program during dynamic linking.
+     */
+    ElfW(Addr)  l_base_diff;
+
+    ElfW(Addr)  l_map_start;  /* Address shared object (its first LOAD segment) is loaded at. */
     const char* l_name;       /* Absolute file name object was found in. */
     ElfW(Dyn)*  l_ld;         /* Dynamic section of the shared object. */
     ElfW(Addr)  l_entry;      /* Entry point location (may be empty, e.g., for libs). */
--- a/Pal/src/db_rtld.c
+++ b/Pal/src/db_rtld.c
@@ -21,11 +21,10 @@
  *     non-loadable and may be missing from final binaries (i.e., in vDSO).
  *     Corresponding linker flag is `-Wl,--hash-style=sysv`.
  *
- *  - They must have DYN or EXEC object file type. Notice that addresses in DYN binaries are
- *    actually offsets from the base address (`l_base`) and thus need adjustment, whereas addresses
- *    in EXEC binaries are hard-coded and do not need any adjustment (thus `l_base == 0x0`). The
- *    LibOS shared library is built as DYN, but some PAL regression tests are built as EXEC, so we
- *    support both.
+ *  - They must have DYN or EXEC object file type. Notice that virtual addresses (p_vaddr) in DYN
+ *    binaries are actually offsets from the address where DYN is loaded at and thus need adjustment
+ *    (via `l_base_diff`), whereas virtual addresses in EXEC binaries are hard-coded and do not need
+ *    any adjustment (thus `l_base_diff == 0x0`).
  */
 
 #include <stdbool.h>
@@ -71,7 +70,7 @@ struct loadcmd {
      *   - start, map_end, alloc_end are page-aligned
      *   - map_off is page-aligned
      *
-     *   The addresses are not relocated (i.e. you need to add l_addr to them).
+     *   The addresses are not relocated (i.e. you need to add l_base_diff to them).
      *
      *   The same struct is used also in LibOS/shim/src/shim_rtld.c code.
      */
@@ -105,14 +104,14 @@ static pal_prot_flags_t elf_segment_prot_to_pal_prot(int elf_segment_prot) {
 }
 
 /* iterate through DSO's program headers to find dynamic section (for dynamic linking) */
-static ElfW(Dyn)* find_dynamic_section(ElfW(Addr) ehdr_addr, ElfW(Addr) base_addr) {
-    const ElfW(Ehdr)* header = (void*)ehdr_addr;
-    const ElfW(Phdr)* phdr   = (void*)(ehdr_addr + header->e_phoff);
+static ElfW(Dyn)* find_dynamic_section(ElfW(Addr) ehdr_addr, ElfW(Addr) base_diff) {
+    const ElfW(Ehdr)* header = (const ElfW(Ehdr)*)ehdr_addr;
+    const ElfW(Phdr)* phdr   = (const ElfW(Phdr)*)(ehdr_addr + header->e_phoff);
 
     ElfW(Dyn)* dynamic_section = NULL;
     for (const ElfW(Phdr)* ph = phdr; ph < &phdr[header->e_phnum]; ph++) {
         if (ph->p_type == PT_DYNAMIC) {
-            dynamic_section = (void*)base_addr + ph->p_vaddr;
+            dynamic_section = (ElfW(Dyn)*)(ph->p_vaddr + base_diff);
             break;
         }
     }
@@ -120,14 +119,14 @@ static ElfW(Dyn)* find_dynamic_section(ElfW(Addr) ehdr_addr, ElfW(Addr) base_add
     return dynamic_section;
 }
 
-int find_string_and_symbol_tables(ElfW(Addr) ehdr_addr, ElfW(Addr) base_addr,
+int find_string_and_symbol_tables(ElfW(Addr) ehdr_addr, ElfW(Addr) base_diff,
                                   const char** out_string_table, ElfW(Sym)** out_symbol_table,
                                   uint32_t* out_symbol_table_cnt) {
     const char* string_table  = NULL;
     ElfW(Sym)* symbol_table   = NULL;
     uint32_t symbol_table_cnt = 0;
 
-    ElfW(Dyn)* dynamic_section = find_dynamic_section(ehdr_addr, base_addr);
+    ElfW(Dyn)* dynamic_section = find_dynamic_section(ehdr_addr, base_diff);
     if (!dynamic_section) {
         log_error("Loaded binary doesn't have dynamic section (required for symbol resolution)");
         return -PAL_ERROR_DENIED;
@@ -138,15 +137,15 @@ int find_string_and_symbol_tables(ElfW(Addr) ehdr_addr, ElfW(Addr) base_addr,
     while (dynamic_section_entry->d_tag != DT_NULL) {
         switch (dynamic_section_entry->d_tag) {
             case DT_STRTAB:
-                string_table = (const char*)(dynamic_section_entry->d_un.d_ptr + base_addr);
+                string_table = (const char*)(dynamic_section_entry->d_un.d_ptr + base_diff);
                 break;
             case DT_SYMTAB:
-                symbol_table = (ElfW(Sym)*)(dynamic_section_entry->d_un.d_ptr + base_addr);
+                symbol_table = (ElfW(Sym)*)(dynamic_section_entry->d_un.d_ptr + base_diff);
                 break;
             case DT_HASH: {
                 /* symbol table size can only be found via ELF hash table's nchain (which is the
                  * second word in the ELF hash table struct) */
-                ElfW(Word)* ht = (ElfW(Word)*)(dynamic_section_entry->d_un.d_ptr + base_addr);
+                ElfW(Word)* ht = (ElfW(Word)*)(dynamic_section_entry->d_un.d_ptr + base_diff);
                 symbol_table_cnt = ht[1];
                 break;
             }
@@ -176,7 +175,7 @@ static int find_symbol_in_loaded_maps(struct link_map* map, ElfW(Rela)* rela,
     /* first try to find in this ELF object itself */
     if (map->symbol_table[symbol_idx].st_value &&
             map->symbol_table[symbol_idx].st_shndx != SHN_UNDEF) {
-        *out_symbol_addr = map->l_base + map->symbol_table[symbol_idx].st_value;
+        *out_symbol_addr = map->symbol_table[symbol_idx].st_value + map->l_base_diff;
         return 0;
     }
 
@@ -195,7 +194,7 @@ static int find_symbol_in_loaded_maps(struct link_map* map, ElfW(Rela)* rela,
         if (!strcmp(symbol_name, pal_symbol_name)) {
             /* NOTE: we currently don't take into account weak symbols and return the first symbol
              *       found (we don't have weak symbols in PAL) */
-            *out_symbol_addr = g_pal_map.l_base + g_pal_map.symbol_table[i].st_value;
+            *out_symbol_addr = g_pal_map.symbol_table[i].st_value + g_pal_map.l_base_diff;
             return 0;
         }
     }
@@ -208,7 +207,7 @@ static int find_symbol_in_loaded_maps(struct link_map* map, ElfW(Rela)* rela,
 static int perform_relocations(struct link_map* map) {
     int ret;
 
-    ElfW(Addr) base_addr = map->l_base;
+    ElfW(Addr) base_diff = map->l_base_diff;
     ElfW(Dyn)* dynamic_section_entry = map->l_ld;
 
     ElfW(Rela)* relas_addr = NULL;
@@ -220,13 +219,13 @@ static int perform_relocations(struct link_map* map) {
     while (dynamic_section_entry->d_tag != DT_NULL) {
         switch (dynamic_section_entry->d_tag) {
             case DT_RELA:
-                relas_addr = (ElfW(Rela)*)(base_addr + dynamic_section_entry->d_un.d_ptr);
+                relas_addr = (ElfW(Rela)*)(dynamic_section_entry->d_un.d_ptr + base_diff);
                 break;
             case DT_RELASZ:
                 relas_size = dynamic_section_entry->d_un.d_val;
                 break;
             case DT_JMPREL:
-                plt_relas_addr = (ElfW(Rela)*)(base_addr + dynamic_section_entry->d_un.d_ptr);
+                plt_relas_addr = (ElfW(Rela)*)(dynamic_section_entry->d_un.d_ptr + base_diff);
                 break;
             case DT_PLTRELSZ:
                 plt_relas_size = dynamic_section_entry->d_un.d_val;
@@ -236,18 +235,18 @@ static int perform_relocations(struct link_map* map) {
     }
 
     /* perform relocs: supported binaries may have only R_X86_64_RELATIVE/R_X86_64_GLOB_DAT relas */
-    ElfW(Rela)* relas_addr_end = (void*)relas_addr + relas_size;
+    ElfW(Rela)* relas_addr_end = (ElfW(Rela)*)((uintptr_t)relas_addr + relas_size);
     for (ElfW(Rela)* rela = relas_addr; rela < relas_addr_end; rela++) {
         if (ELFW(R_TYPE)(rela->r_info) == R_X86_64_RELATIVE) {
-            ElfW(Addr)* addr_to_relocate = (ElfW(Addr)*)(base_addr + rela->r_offset);
-            *addr_to_relocate = base_addr + *addr_to_relocate;
+            ElfW(Addr)* addr_to_relocate = (ElfW(Addr)*)(rela->r_offset + base_diff);
+            *addr_to_relocate = *addr_to_relocate + base_diff;
         } else if (ELFW(R_TYPE)(rela->r_info) == R_X86_64_GLOB_DAT) {
             ElfW(Addr) symbol_addr;
             ret = find_symbol_in_loaded_maps(map, rela, &symbol_addr);
             if (ret < 0)
                 return ret;
 
-            ElfW(Addr)* addr_to_relocate = (ElfW(Addr)*)(base_addr + rela->r_offset);
+            ElfW(Addr)* addr_to_relocate = (ElfW(Addr)*)(rela->r_offset + base_diff);
             *addr_to_relocate = symbol_addr + rela->r_addend;
         } else {
             log_error("Unrecognized relocation type; PAL loader currently supports only "
@@ -274,7 +273,7 @@ static int perform_relocations(struct link_map* map) {
         if (ret < 0)
             return ret;
 
-        ElfW(Addr)* addr_to_relocate = (ElfW(Addr)*)(base_addr + plt_rela->r_offset);
+        ElfW(Addr)* addr_to_relocate = (ElfW(Addr)*)(plt_rela->r_offset + base_diff);
         *addr_to_relocate = symbol_addr + plt_rela->r_addend;
     }
 
@@ -317,7 +316,7 @@ static int create_and_relocate_entrypoint(PAL_HANDLE handle, const char* elf_fil
     for (const ElfW(Phdr)* ph = phdr; ph < &phdr[ehdr->e_phnum]; ph++) {
         switch (ph->p_type) {
             case PT_DYNAMIC:
-                g_entrypoint_map.l_ld = (void*)ph->p_vaddr;
+                g_entrypoint_map.l_ld = (ElfW(Dyn)*)ph->p_vaddr;
                 break;
 
             case PT_LOAD:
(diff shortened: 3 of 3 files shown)


pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/255)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski, @dimakuv, and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/255#-MpkxkLi3c854vrKDevi:-MpkxkLj5ClYDNzszDXV:b-51hlfh) (no related file):*
I'm still not a fan of "diff": for me, it sounds not like a difference between numbers/addresses, but something like a textual diff. And how do we use it in English? "The diff of this ELF object"? "The diff at/with which this ELF was loaded"?

Alternatives:
* @mkow suggested "base_shift"
* synonyms of "difference": "shift"? "delta"? There's "offset", but usually it means offset in file.
* synonyms of "base": "origin"? (as in: origin of coordinate system)

Not blocking, though: I'm fine with "diff" if it sounds best to you.

---

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/255)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/255#-MpkxkLi3c854vrKDevi:-Mpl3dmM1chNOqpbIX7s:b-zek0hy) (no related file):*

I'm still not a fan of "diff": for me, it sounds not like a difference between numbers/addresses, but something like a textual diff. And how do we use it in English? "The diff of this ELF object"? "The diff at/with which this ELF was loaded"?

Alternatives:
* @mkow suggested "base_shift"
* synonyms of "difference": "shift"? "delta"? There's "offset", but usually it means offset in file.
* synonyms of "base": "origin"? (as in: origin of coordinate system)

Not blocking, though: I'm fine with "diff" if it sounds best to you.
</blockquote></details>

I'd vote for `base_shift`

---

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/255)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/255#-MpkxkLi3c854vrKDevi:-MplOcYg3-hSAAF9aCy8:bhkm051) (no related file):*

I'd vote for `base_shift`
</blockquote></details>

I'm not a big fan of `base_shift`. This name is not easily recognizable (shift from the base? `shift` is a very weird word in the context of memory addresses). I suggest:
- `base_diff`
- `base_offset`
-  `addr_diff` (I vote for this)

I guess @boryspoplawski will be our tie-breaker :)

---

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/255)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv, @mkow, and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/255#-MpkxkLi3c854vrKDevi:-Mpm6fcr84fnYrtwySp9:b3vkg0t) (no related file):*
> And how do we use it in English? "The diff of this ELF object"? "The diff at/with which this ELF was loaded"?

Non of the proposed names are any better with regard to this. I guess the reason is that it's the difference of in memory and in file base addresses, so unless your sentence contains references to both of these (in mem and in file) then any term will be missing the context.

> I'd vote for base_shift

I don't like this either, sounds like base is aligned to some power of 2 and you use "shift operator" to obtain the actual base (e.g. `encoded_base << base_shift == real_base`).

> I guess @boryspoplawski will be our tie-breaker :)

I vote for `base_diff`, but `addr_diff` will do as well.

---

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/255)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski, @mkow, and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/255#-MpkxkLi3c854vrKDevi:-MpowdXgELguWKgKuRmF:bs4xikn) (no related file):*

> And how do we use it in English? "The diff of this ELF object"? "The diff at/with which this ELF was loaded"?

Non of the proposed names are any better with regard to this. I guess the reason is that it's the difference of in memory and in file base addresses, so unless your sentence contains references to both of these (in mem and in file) then any term will be missing the context.

> I'd vote for base_shift

I don't like this either, sounds like base is aligned to some power of 2 and you use "shift operator" to obtain the actual base (e.g. `encoded_base << base_shift == real_base`).

> I guess @boryspoplawski will be our tie-breaker :)

I vote for `base_diff`, but `addr_diff` will do as well.
</blockquote></details>

`base_diff` it is. Will submit a fixup shortly.

---

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/255)* status: 0 of 3 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/255#-MpkxkLi3c854vrKDevi:-MpozZwt2JB649eNEzHg:b-896fix) (no related file):*

`base_diff` it is. Will submit a fixup shortly.
</blockquote></details>

Done.

---

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/255)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/255#-MpkxkLi3c854vrKDevi:-MppYQHt2RwuSdeGh3nt:b5rs6wm) (no related file):*

Done.
</blockquote></details>

After checking the PR it seems we had a couple of bugs, stemming from the poor name choice. Good we changed it!

---
___
*[LibOS/shim/src/shim_rtld.c, line 46 at r2](https://reviewable.io/reviews/gramineproject/gramine/255#-MppSd5h2vEiy2q89jE5:-MppS1U4AuYgUzjt4yPV:bgzj0xs) ([raw file](https://github.com/gramineproject/gramine/blob/a8ca87ce11dd8662bf34c828d171bf6bb3b1726b/LibOS/shim/src/shim_rtld.c#L46)):*
> ```C
>     /*
>      * Difference between virtual addresses (p_vaddr) in ELF file and actual virtual addresses in
>      * memory. Equal to the address shared object (its first LOAD segment) is loaded at for ET_DYNs,
> ```

Isn't there (rather pathological and probably not happening in the wild) a case where the first LOAD segment has non 0 address for ET_DYNs? In such case this would not be the address where the shared object is loaded at (but well, the diff described in the first sentence).
___
*[LibOS/shim/src/shim_rtld.c, line 487 at r2](https://reviewable.io/reviews/gramineproject/gramine/255#-MppXiZTE-5I6jM-FYWM:-MppXiZU010ZF8a0VTjJ:b-dxblqn) ([raw file](https://github.com/gramineproject/gramine/blob/a8ca87ce11dd8662bf34c828d171bf6bb3b1726b/LibOS/shim/src/shim_rtld.c#L487)):*
> ```C
> 
> static void remove_elf_object(struct link_map* l) {
>     remove_r_debug((void*)l->l_base_diff);
> ```

I believe this is a bug. These functions (`remove_r_debug` and `append_r_debug`) expect the address where the elf is loaded at, not `base_diff`. This would only prove that the choice of the old name was beyond broken.
This works only because in most "normal" elfs these two values (base address and `base_diff`) are equal.
___
*[LibOS/shim/src/shim_rtld.c, line 643 at r2](https://reviewable.io/reviews/gramineproject/gramine/255#-MppVWz8-KrTMyUOthDi:-MppVWz94pgp9U876mn-:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/a8ca87ce11dd8662bf34c828d171bf6bb3b1726b/LibOS/shim/src/shim_rtld.c#L643)):*
> ```C
> 
>     if (map->l_file && !qstrempty(&map->l_file->uri)) {
>         append_r_debug(qstrgetstr(&map->l_file->uri), (void*)map->l_base_diff);
> ```

ditto
___
*[LibOS/shim/src/shim_rtld.c, line 901 at r2](https://reviewable.io/reviews/gramineproject/gramine/255#-MppYJBEBz4SIqlRnMtN:-MppYJBEBz4SIqlRnMtO:b-kc313u) ([raw file](https://github.com/gramineproject/gramine/blob/a8ca87ce11dd8662bf34c828d171bf6bb3b1726b/LibOS/shim/src/shim_rtld.c#L901)):*
> ```C
>     auxp[3].a_un.a_val = g_exec_map->l_entry;
>     auxp[4].a_type     = AT_BASE;
>     auxp[4].a_un.a_val = g_interp_map ? g_interp_map->l_base_diff : 0;
> ```

ditto - this needs to be an actual address, not `base_diff`
___
*[Pal/include/pal_rtld.h, line 18 at r2](https://reviewable.io/reviews/gramineproject/gramine/255#-MppYbzN-FM3h9yTFlg0:-MppYbzO6m4e_W6ixuGu:b-63mdd2) ([raw file](https://github.com/gramineproject/gramine/blob/a8ca87ce11dd8662bf34c828d171bf6bb3b1726b/Pal/include/pal_rtld.h#L18)):*
> ```objc
>     /*
>      * Difference between virtual addresses (p_vaddr) in ELF file and actual virtual addresses in
>      * memory. Equal to the address shared object (its first LOAD segment) is loaded at for ET_DYNs,
> ```

ditto (as in LibOS)
___
*[Pal/src/db_rtld.c, line 28 at r2](https://reviewable.io/reviews/gramineproject/gramine/255#-MppYr2h6NzF8_c9vmOv:-MppYr2i3NdQ1aw9eYGx:bfneerh) ([raw file](https://github.com/gramineproject/gramine/blob/a8ca87ce11dd8662bf34c828d171bf6bb3b1726b/Pal/src/db_rtld.c#L28)):*
> ```C
>  *    (via `l_base_diff`), whereas virtual addresses in EXEC binaries are hard-coded and do not need
>  *    any adjustment (thus `l_base_diff == 0x0`). The LibOS shared library is built as DYN, but some
>  *    PAL regression tests are built as EXEC, so we support both.
> ```

I would remove the last sentence, it adds nothing and might become invalid soon. Your new PAL (if/when it comes) will be an EXEC, right?
___
*[Pal/src/db_rtld.c, line 399 at r2](https://reviewable.io/reviews/gramineproject/gramine/255#-MppZnirAHwHEXtBsT1f:-MppZnirAHwHEXtBsT1g:b-ucy6ch) ([raw file](https://github.com/gramineproject/gramine/blob/a8ca87ce11dd8662bf34c828d171bf6bb3b1726b/Pal/src/db_rtld.c#L399)):*
> ```C
>             g_entrypoint_map.l_map_start = (ElfW(Addr))map_addr;
>             g_entrypoint_map.l_base_diff = (ehdr->e_type == ET_EXEC) ? 0x0
>                                                                      : g_entrypoint_map.l_map_start;
> ```

Shouldn't we adjust this by subtracting first (this) segments `p_vaddr`?
___
*[Pal/src/db_rtld.c, line 543 at r2](https://reviewable.io/reviews/gramineproject/gramine/255#-Mpp_gSB4Tf2fmCxer6m:-Mpp_gSB4Tf2fmCxer6n:b-mtpqg4) ([raw file](https://github.com/gramineproject/gramine/blob/a8ca87ce11dd8662bf34c828d171bf6bb3b1726b/Pal/src/db_rtld.c#L543)):*
> ```C
> #ifdef DEBUG
>     assert(g_entrypoint_map.l_name);
>     _DkDebugMapAdd(g_entrypoint_map.l_name, (void*)g_entrypoint_map.l_base_diff);
> ```

ditto - should be address not diff
___
*[Pal/src/db_rtld.c, line 559 at r2](https://reviewable.io/reviews/gramineproject/gramine/255#-Mpp_l6rEx1uGHK9RTbc:-Mpp_l6rEx1uGHK9RTbd:b-epy2h9) ([raw file](https://github.com/gramineproject/gramine/blob/a8ca87ce11dd8662bf34c828d171bf6bb3b1726b/Pal/src/db_rtld.c#L559)):*
> ```C
> 
>     ElfW(Addr) pal_binary_addr = (ElfW(Addr))&__ehdr_start;
>     ElfW(Addr) pal_base_diff   = pal_binary_addr;
> ```

Shouldn't this be adjusted by first load segment's `p_vaddr`?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/255)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv and @pwmarcz)

___
*[LibOS/shim/src/shim_rtld.c, line 487 at r2](https://reviewable.io/reviews/gramineproject/gramine/255#-MppXiZTE-5I6jM-FYWM:-MppbUsz8wJTr3ibfk1A:bo1phbc) ([raw file](https://github.com/gramineproject/gramine/blob/a8ca87ce11dd8662bf34c828d171bf6bb3b1726b/LibOS/shim/src/shim_rtld.c#L487)):*

I believe this is a bug. These functions (`remove_r_debug` and `append_r_debug`) expect the address where the elf is loaded at, not `base_diff`. This would only prove that the choice of the old name was beyond broken.
This works only because in most "normal" elfs these two values (base address and `base_diff`) are equal.
</blockquote></details>

I also thought so originally, but all these functions work with `ET_EXEC` files as well (e.g., I checked this on our PAL regression tests, where most PAL files are non-pie executables, i.e., `ET_EXEC`). And with `ET_EXEC`, `l_base_diff` is zero. So all these functions seem to accept the address difference, not the address itself.

@pwmarcz Could you shed some light on these functions and their arguments?

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/255)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv and @pwmarcz)

___
*[LibOS/shim/src/shim_rtld.c, line 487 at r2](https://reviewable.io/reviews/gramineproject/gramine/255#-MppXiZTE-5I6jM-FYWM:-MppcacP0f0J6w0ToKEl:byfptgw) ([raw file](https://github.com/gramineproject/gramine/blob/a8ca87ce11dd8662bf34c828d171bf6bb3b1726b/LibOS/shim/src/shim_rtld.c#L487)):*

I also thought so originally, but all these functions work with `ET_EXEC` files as well (e.g., I checked this on our PAL regression tests, where most PAL files are non-pie executables, i.e., `ET_EXEC`). And with `ET_EXEC`, `l_base_diff` is zero. So all these functions seem to accept the address difference, not the address itself.

@pwmarcz Could you shed some light on these functions and their arguments?
</blockquote></details>

Could you verify in GDB that they work as expected? I think this might work by accident - there is only one EXEC at a time, so only one elf will have this address == 0 (since we disallow mapping anything there normally). Maybe we do not get some symbols (or whatever this address is used for), but nothing brakes thanks to that.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/255)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv and @pwmarcz)

___
*[LibOS/shim/src/shim_rtld.c, line 487 at r2](https://reviewable.io/reviews/gramineproject/gramine/255#-MppXiZTE-5I6jM-FYWM:-MpphXY4EB3HZKAUQq59:bfg8dkt) ([raw file](https://github.com/gramineproject/gramine/blob/a8ca87ce11dd8662bf34c828d171bf6bb3b1726b/LibOS/shim/src/shim_rtld.c#L487)):*
> Could you verify in GDB that they work as expected?

I verified that e.g. HelloWorld is debuggable as usual. All HelloWorld symbols are there, I can put a breakpoint on `main` and I can step through it. So it looks totally fine. Or did you mean something else?

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/255)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv and @pwmarcz)

___
*[LibOS/shim/src/shim_rtld.c, line 487 at r2](https://reviewable.io/reviews/gramineproject/gramine/255#-MppXiZTE-5I6jM-FYWM:-Mppo0xO3LWJeX4anPsB:b-p0g58w) ([raw file](https://github.com/gramineproject/gramine/blob/a8ca87ce11dd8662bf34c828d171bf6bb3b1726b/LibOS/shim/src/shim_rtld.c#L487)):*

> Could you verify in GDB that they work as expected?

I verified that e.g. HelloWorld is debuggable as usual. All HelloWorld symbols are there, I can put a breakpoint on `main` and I can step through it. So it looks totally fine. Or did you mean something else?
</blockquote></details>

I've also checked - created an DYN ELF, which had first load segment address != 0 (unfortunately you cannot do this with ld, see the patch at https://sourceware.org/legacy-ml/binutils/2013-12/msg00081.html). Gramine loaded it without problems, debugging worked as well. Seems that this `l_addr` is not used for anything, except detecting collisions of libraries/not loading same lib twice in debug map list?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/255)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski and @dimakuv)

___
*[LibOS/shim/src/shim_rtld.c, line 46 at r2](https://reviewable.io/reviews/gramineproject/gramine/255#-MppSd5h2vEiy2q89jE5:-MpqBB1X3IjO9D8EHHHZ:b-gkxx3w) ([raw file](https://github.com/gramineproject/gramine/blob/a8ca87ce11dd8662bf34c828d171bf6bb3b1726b/LibOS/shim/src/shim_rtld.c#L46)):*

Isn't there (rather pathological and probably not happening in the wild) a case where the first LOAD segment has non 0 address for ET_DYNs? In such case this would not be the address where the shared object is loaded at (but well, the diff described in the first sentence).
</blockquote></details>

Yes, `(its first LOAD segment)` sounds wrong, and also that's why we have `l_map_start`.
___
*[LibOS/shim/src/shim_rtld.c, line 487 at r2](https://reviewable.io/reviews/gramineproject/gramine/255#-MppXiZTE-5I6jM-FYWM:-MpqBWZB65Cib8iAmaCJ:bojezqd) ([raw file](https://github.com/gramineproject/gramine/blob/a8ca87ce11dd8662bf34c828d171bf6bb3b1726b/LibOS/shim/src/shim_rtld.c#L487)):*

I've also checked - created an DYN ELF, which had first load segment address != 0 (unfortunately you cannot do this with ld, see the patch at https://sourceware.org/legacy-ml/binutils/2013-12/msg00081.html). Gramine loaded it without problems, debugging worked as well. Seems that this `l_addr` is not used for anything, except detecting collisions of libraries/not loading same lib twice in debug map list?
</blockquote></details>

I believe all these functions already operate on "base diff" as defined currently:
* This value is used in the GDB integration, and we use it to compute section addresses that we pass to GDB. See `debug_map_gdb.py`.
* It's also used by `describe_location` to calculate symbol address: we subtract this value from the memory address, then compare it with output of `nm`. See `debug_map.c` (`debug_map_find()`).

What will _not_ work in the case you describe is unloading such objects when unmapping memory (see `shim_vma.c`), and there's a TODO there about exactly this.
___
*[LibOS/shim/src/shim_rtld.c, line 643 at r2](https://reviewable.io/reviews/gramineproject/gramine/255#-MppVWz8-KrTMyUOthDi:-MpqCyN72sYWGRXzGqDN:br1hn2u) ([raw file](https://github.com/gramineproject/gramine/blob/a8ca87ce11dd8662bf34c828d171bf6bb3b1726b/LibOS/shim/src/shim_rtld.c#L643)):*

ditto
</blockquote></details>

As said above, I disagree.
___
*[Pal/src/db_rtld.c, line 543 at r2](https://reviewable.io/reviews/gramineproject/gramine/255#-Mpp_gSB4Tf2fmCxer6m:-MpqDS7bDos-PpJEoc5t:br1hn2u) ([raw file](https://github.com/gramineproject/gramine/blob/a8ca87ce11dd8662bf34c828d171bf6bb3b1726b/Pal/src/db_rtld.c#L543)):*

ditto - should be address not diff
</blockquote></details>

As said above, I disagree.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/255)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski, @dimakuv, and @pwmarcz)

___
*[LibOS/shim/src/shim_rtld.c, line 487 at r2](https://reviewable.io/reviews/gramineproject/gramine/255#-MppXiZTE-5I6jM-FYWM:-MpqVsYnECW1UWVyUgnX:bwcx3j2) ([raw file](https://github.com/gramineproject/gramine/blob/a8ca87ce11dd8662bf34c828d171bf6bb3b1726b/LibOS/shim/src/shim_rtld.c#L487)):*

I believe all these functions already operate on "base diff" as defined currently:
* This value is used in the GDB integration, and we use it to compute section addresses that we pass to GDB. See `debug_map_gdb.py`.
* It's also used by `describe_location` to calculate symbol address: we subtract this value from the memory address, then compare it with output of `nm`. See `debug_map.c` (`debug_map_find()`).

What will _not_ work in the case you describe is unloading such objects when unmapping memory (see `shim_vma.c`), and there's a TODO there about exactly this.
</blockquote></details>

All right, it makes sense, but I would rename these arguments (because atm they are just `addr`, which is clearly misleading).

Also the comment in `debug_map_gdb.py` mentions `-o <load_addr>`:
1) `load_addr` is actually okay-ish name,
2) gdb help says `-o OFFSET` and always describes this as an offset (not an address).
___
*[Pal/src/db_rtld.c, line 543 at r2](https://reviewable.io/reviews/gramineproject/gramine/255#-Mpp_gSB4Tf2fmCxer6m:-MpqWxxS64CRSInBfeof:bkk47yg) ([raw file](https://github.com/gramineproject/gramine/blob/a8ca87ce11dd8662bf34c828d171bf6bb3b1726b/Pal/src/db_rtld.c#L543)):*

As said above, I disagree.
</blockquote></details>

ditto (argument name)

pwmarcz (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/255)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[LibOS/shim/src/shim_rtld.c, line 901 at r2](https://reviewable.io/reviews/gramineproject/gramine/255#-MppYJBEBz4SIqlRnMtN:-Mpqw_mC6xIQ8Akk_b2K:b-juxc9y) ([raw file](https://github.com/gramineproject/gramine/blob/a8ca87ce11dd8662bf34c828d171bf6bb3b1726b/LibOS/shim/src/shim_rtld.c#L901)):*

Done. From what I understand in the corresponding Linux source, you are right, this should be the actual base address (here: https://elixir.bootlin.com/linux/latest/source/fs/binfmt_elf.c#L591).

@pwmarcz It looks like `l_map_start` was not used at all in the LibOS?
</blockquote></details>

That sounds likely - the interpreter is usually from Glibc, so probably we didn't give much thought to this part.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/255)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @pwmarcz)

___
*[LibOS/shim/src/shim_rtld.c, line 46 at r2](https://reviewable.io/reviews/gramineproject/gramine/255#-MppSd5h2vEiy2q89jE5:-Mpqjw-UBxNotXXFuhvT:b-lfc7dt) ([raw file](https://github.com/gramineproject/gramine/blob/a8ca87ce11dd8662bf34c828d171bf6bb3b1726b/LibOS/shim/src/shim_rtld.c#L46)):*

Done.
</blockquote></details>

What about `address shared object is loaded at`?
___
*[LibOS/shim/src/shim_rtld.c, line 487 at r2](https://reviewable.io/reviews/gramineproject/gramine/255#-MppXiZTE-5I6jM-FYWM:-Mpqk9PN8DYj88GTDVub:b238o44) ([raw file](https://github.com/gramineproject/gramine/blob/a8ca87ce11dd8662bf34c828d171bf6bb3b1726b/LibOS/shim/src/shim_rtld.c#L487)):*

Do you want me to rename the names in this PR (in a separate commit), or have a separate PR with these debug-funcs renames?
</blockquote></details>

I would even say they belong to this commit (the only one in this PR).
___
*[Pal/include/pal_rtld.h, line 18 at r2](https://reviewable.io/reviews/gramineproject/gramine/255#-MppYbzN-FM3h9yTFlg0:-Mpqxj0N-ltYt9SZbqUA:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/a8ca87ce11dd8662bf34c828d171bf6bb3b1726b/Pal/include/pal_rtld.h#L18)):*

Done.
</blockquote></details>

ditto
___
*[Pal/src/db_rtld.c, line 399 at r2](https://reviewable.io/reviews/gramineproject/gramine/255#-MppZnirAHwHEXtBsT1f:-MpqzLsc1dv7Bfg6tukl:b-btg1c8) ([raw file](https://github.com/gramineproject/gramine/blob/a8ca87ce11dd8662bf34c828d171bf6bb3b1726b/Pal/src/db_rtld.c#L399)):*

I added the comment describing that `p_vaddr == 0`, otherwise this PAL rtld code would return with error.
</blockquote></details>

Isn't the fix just adding `- c->start`? I think that's even easier than explaining why we don't do it.
___
*[Pal/src/db_rtld.c, line 559 at r2](https://reviewable.io/reviews/gramineproject/gramine/255#-Mpp_l6rEx1uGHK9RTbc:-Mpr1QT93_T1nPI6ObCF:bovzf65) ([raw file](https://github.com/gramineproject/gramine/blob/a8ca87ce11dd8662bf34c828d171bf6bb3b1726b/Pal/src/db_rtld.c#L559)):*

We assume that the PAL binary is `ET_DYN` with the first load segment having `p_vaddr == 0`. Actually, I don't even know how to get to the first LOAD segment in this code -- the PAL binary was loaded by the kernel (in the Linux PAL case).

I added a comment; not sure if it helps...
</blockquote></details>

Okay, but this might become a problem for your new PAL (since it's an EXEC).

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/255)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @pwmarcz)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/255)* status: 0 of 3 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, and @pwmarcz)

___
*[LibOS/shim/src/shim_rtld.c, line 46 at r2](https://reviewable.io/reviews/gramineproject/gramine/255#-MppSd5h2vEiy2q89jE5:-MpuPtKZ0WuYDiXdSs2Z:b-je8umi) ([raw file](https://github.com/gramineproject/gramine/blob/a8ca87ce11dd8662bf34c828d171bf6bb3b1726b/LibOS/shim/src/shim_rtld.c#L46)):*

What about `address shared object is loaded at`?
</blockquote></details>

Done. Couldn't come up with a short desc, so just removed it. Probably was not needed anyway, since the "difference between ..." already says it all.
___
*[LibOS/shim/src/shim_rtld.c, line 487 at r2](https://reviewable.io/reviews/gramineproject/gramine/255#-MppXiZTE-5I6jM-FYWM:-MpuZT4566FAHgMVFi-0:b-fhu8np) ([raw file](https://github.com/gramineproject/gramine/blob/a8ca87ce11dd8662bf34c828d171bf6bb3b1726b/LibOS/shim/src/shim_rtld.c#L487)):*

I would even say they belong to this commit (the only one in this PR).
</blockquote></details>

No, I don't think we should do it in this commit. I submitted a separate issue:https://github.com/gramineproject/gramine/issues/259

I first tried to fix all these debug-related functions (including `DkDebug...`), but this exploded. There is a *lot* of code where we use `addr` but in fact it means `base_diff`. We should fix it in a separate PR.
___
*[Pal/include/pal_rtld.h, line 18 at r2](https://reviewable.io/reviews/gramineproject/gramine/255#-MppYbzN-FM3h9yTFlg0:-MpuQ-QP8EW1IBJQGjGs:bb6un1j) ([raw file](https://github.com/gramineproject/gramine/blob/a8ca87ce11dd8662bf34c828d171bf6bb3b1726b/Pal/include/pal_rtld.h#L18)):*

ditto
</blockquote></details>

Done
___
*[Pal/src/db_rtld.c, line 399 at r2](https://reviewable.io/reviews/gramineproject/gramine/255#-MppZnirAHwHEXtBsT1f:-MpuTQnBAa-TK16wN8lC:b-11tm5v) ([raw file](https://github.com/gramineproject/gramine/blob/a8ca87ce11dd8662bf34c828d171bf6bb3b1726b/Pal/src/db_rtld.c#L399)):*

Isn't the fix just adding `- c->start`? I think that's even easier than explaining why we don't do it.
</blockquote></details>

Done. Ok, actually I reworked it; now this code snippet looks and behaves similarly to how it's done in `shim_rtld.c`

pwmarcz (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/255)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/255)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[Pal/src/db_rtld.c, line 399 at r2](https://reviewable.io/reviews/gramineproject/gramine/255#-MppZnirAHwHEXtBsT1f:-Mpuq_IYA9SIT5zKAtIH:bi1kyu5) ([raw file](https://github.com/gramineproject/gramine/blob/a8ca87ce11dd8662bf34c828d171bf6bb3b1726b/Pal/src/db_rtld.c#L399)):*

Done. Ok, actually I reworked it; now this code snippet looks and behaves similarly to how it's done in `shim_rtld.c`
</blockquote></details>

This seems to be more complicated, I expected a simple `- c->start` is enough.
___
*[Pal/src/db_rtld.c, line 385 at r4](https://reviewable.io/reviews/gramineproject/gramine/255#-MpuqiNT1u6-GFU9M4eB:-MpuqiNU71pfE6z6cZs3:b-w10ifc) ([raw file](https://github.com/gramineproject/gramine/blob/a7074cce776b12fedeaa57210f3a1c4d87ec2421/Pal/src/db_rtld.c#L385)):*
> ```C
>              * overwrite memory on subsequent segments.
>              *
>              * Note that we reserve memory starting from offset 0, not from c->start. This is to
> ```

Wait, this does not reserve memory - it actually maps it.
___
*[Pal/src/db_rtld.c, line 392 at r4](https://reviewable.io/reviews/gramineproject/gramine/255#-MpuqljQ8BHTt9DVBUaw:-MpuqljQ8BHTt9DVBUax:b-jjvmyh) ([raw file](https://github.com/gramineproject/gramine/blob/a7074cce776b12fedeaa57210f3a1c4d87ec2421/Pal/src/db_rtld.c#L392)):*
> ```C
>         }
> 
>         ret = _DkStreamMap(handle, &map_addr, c->prot | PAL_PROT_WRITECOPY, c->map_off, map_size);
> ```

Are you sure this is correct? Wouldn't it map first segment at the start of this range, regardless of it's `p_vaddr`?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/255)* status: 2 of 3 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @pwmarcz)

___
*[LibOS/shim/src/shim_rtld.c, line 386 at r5](https://reviewable.io/reviews/gramineproject/gramine/255#-MpvF9lk0QL3p-XME5WG:-MpvF9lk0QL3p-XME5WH:b-m9l49s) ([raw file](https://github.com/gramineproject/gramine/blob/cdfedcc014d14159464f54e2f4052e827bed4361/LibOS/shim/src/shim_rtld.c#L386)):*
> ```C
>     size_t load_end = loadcmds[n_loadcmds - 1].alloc_end;
> 
>     if (ehdr->e_type == ET_DYN) {
> ```

FYI: @boryspoplawski This is where my logic in PAL mirrors the logic in LibOS now.
___
*[Pal/src/db_rtld.c, line 399 at r2](https://reviewable.io/reviews/gramineproject/gramine/255#-MppZnirAHwHEXtBsT1f:-Mpusb1i1qyFe-nxWPyR:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/a8ca87ce11dd8662bf34c828d171bf6bb3b1726b/Pal/src/db_rtld.c#L399)):*

This seems to be more complicated, I expected a simple `- c->start` is enough.
</blockquote></details>

Done.
___
*[Pal/src/db_rtld.c, line 385 at r4](https://reviewable.io/reviews/gramineproject/gramine/255#-MpuqiNT1u6-GFU9M4eB:-MpvEJN77l30TjMlL6Dw:b-8vzdg) ([raw file](https://github.com/gramineproject/gramine/blob/a7074cce776b12fedeaa57210f3a1c4d87ec2421/Pal/src/db_rtld.c#L385)):*

Wait, this does not reserve memory - it actually maps it.
</blockquote></details>

Done. I still keep `reserve` because that's what we want to do -- reserve a sufficiently large memory region that is enough to put all LOAD segments into this region. We reserve this region via a dummy `_DkStreamMap()` because this is the only function in PAL I could find that finds some unoccupied address :(
___
*[Pal/src/db_rtld.c, line 392 at r4](https://reviewable.io/reviews/gramineproject/gramine/255#-MpuqljQ8BHTt9DVBUaw:-MpuriuJ4Q_xwWJV0lDv:b-127kg7) ([raw file](https://github.com/gramineproject/gramine/blob/a7074cce776b12fedeaa57210f3a1c4d87ec2421/Pal/src/db_rtld.c#L392)):*

Are you sure this is correct? Wouldn't it map first segment at the start of this range, regardless of it's `p_vaddr`?
</blockquote></details>

Done. You are right, I mixed everything. This is because I tried to "merge" two code paths into one:
1. Reserve a sufficiently large memory region
2. File-map the contents from each LOAD segment in the ELF file to this memory region

Now I split these two steps in two separate code snippets, and the logic makes sense now. In fact, it is the same as in LibOS (see my other comment).

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/255)* status: 2 of 3 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @pwmarcz)

___
*[Pal/src/db_rtld.c, line 385 at r4](https://reviewable.io/reviews/gramineproject/gramine/255#-MpuqiNT1u6-GFU9M4eB:-MpveWIO4GWb11OmOpKf:b3vo9e4) ([raw file](https://github.com/gramineproject/gramine/blob/a7074cce776b12fedeaa57210f3a1c4d87ec2421/Pal/src/db_rtld.c#L385)):*

Done. I still keep `reserve` because that's what we want to do -- reserve a sufficiently large memory region that is enough to put all LOAD segments into this region. We reserve this region via a dummy `_DkStreamMap()` because this is the only function in PAL I could find that finds some unoccupied address :(
</blockquote></details>

We still map the memory, not reserve it. Note that on LibOS level we merely mark the memory as taken in our memory manager, here you actually map a file there.
___
*[Pal/src/db_rtld.c, line 371 at r5](https://reviewable.io/reviews/gramineproject/gramine/255#-Mpvf_lY-q4NpniYAF_O:-Mpvf_lZBEjJ2Dg7HWxG:bwgexqv) ([raw file](https://github.com/gramineproject/gramine/blob/cdfedcc014d14159464f54e2f4052e827bed4361/Pal/src/db_rtld.c#L371)):*
> ```C
> 
>     if (ehdr->e_type == ET_DYN) {
>         /*
> ```

My initial comment about handling this was because I thought it's enough to add `- c->start` to one line and that's all. Since we do not support first `p_vaddr != 0` anyway, any more complicated solution is probably not worth it.
Since you've already done the work, we could have merge it, if it was correct, but I'm not really sure it is. I'll leave the decision what to do with this to you. Whatever you decide to do (keep this version or restore the old one) please keep the setting of starting address out of the loop - it's a nice cleanup :)
___
*[Pal/src/db_rtld.c, line 382 at r5](https://reviewable.io/reviews/gramineproject/gramine/255#-MpvdzD3-wC5hR2hOvEn:-MpvdzD49QU1Z_kK5mkr:b-2f6hpt) ([raw file](https://github.com/gramineproject/gramine/blob/cdfedcc014d14159464f54e2f4052e827bed4361/Pal/src/db_rtld.c#L382)):*
> ```C
>          */
>         void* map_addr = NULL;
>         ret = _DkStreamMap(handle, &map_addr, /*prot=*/0, /*offset=*/0,
> ```

How does it actually work? We map a file, which might be far smaller than the map size (in the case of first p_vaddr != 0). Doesn't this mean that `map_addr` now contain elf header?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/255)* status: 2 of 3 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @pwmarcz)

___
*[Pal/src/db_rtld.c, line 385 at r4](https://reviewable.io/reviews/gramineproject/gramine/255#-MpuqiNT1u6-GFU9M4eB:-Mpvm8Rt5rMuZxMIGJdN:b-rl2h6f) ([raw file](https://github.com/gramineproject/gramine/blob/a7074cce776b12fedeaa57210f3a1c4d87ec2421/Pal/src/db_rtld.c#L385)):*

We still map the memory, not reserve it. Note that on LibOS level we merely mark the memory as taken in our memory manager, here you actually map a file there.
</blockquote></details>

Done. Hope new comments explain what's going on better.
___
*[Pal/src/db_rtld.c, line 371 at r5](https://reviewable.io/reviews/gramineproject/gramine/255#-Mpvf_lY-q4NpniYAF_O:-MpvmIGR5PaCeDulmLto:bh8rw3p) ([raw file](https://github.com/gramineproject/gramine/blob/cdfedcc014d14159464f54e2f4052e827bed4361/Pal/src/db_rtld.c#L371)):*

My initial comment about handling this was because I thought it's enough to add `- c->start` to one line and that's all. Since we do not support first `p_vaddr != 0` anyway, any more complicated solution is probably not worth it.
Since you've already done the work, we could have merge it, if it was correct, but I'm not really sure it is. I'll leave the decision what to do with this to you. Whatever you decide to do (keep this version or restore the old one) please keep the setting of starting address out of the loop - it's a nice cleanup :)
</blockquote></details>

The latest implementation is definitely more correct (and it mirrors the LibOS rtld implementation). I prefer to keep this implementation.
___
*[Pal/src/db_rtld.c, line 382 at r5](https://reviewable.io/reviews/gramineproject/gramine/255#-MpvdzD3-wC5hR2hOvEn:-MpvjjzqBICWLKIpx2-a:bzii3h4) ([raw file](https://github.com/gramineproject/gramine/blob/cdfedcc014d14159464f54e2f4052e827bed4361/Pal/src/db_rtld.c#L382)):*

How does it actually work? We map a file, which might be far smaller than the map size (in the case of first p_vaddr != 0). Doesn't this mean that `map_addr` now contain elf header?
</blockquote></details>

But we don't care about the actual contents. Yes, `map_addr` will contain the ELF header, but this is irrelevant. What's relevant is that we will later map each LOAD segment over this memory region. And the ELf header (which is in the first LOAD segment) will be at the `l_map_start` -- this is what we will use anyway.

So if say first LOAD segment has `p_vaddr = 0x1000`, then `l_map_start = 0x1000 + l_base_diff` (which is where we allocated this dummy memory region). So the ELF header will be read from the correct final address (`l_map_start = 0x1000 + l_base_diff`).

Just to reiterate: I have to use `DkStreamMap()` to get a dummy memory region at some address. I will not use the contents of this memory region at all -- they will be overwritten by the contents of `DkStreamMap(<actual LOAD segment>)` invocations below. I have to use `DkStreamMap()` here simply because we have a bug (feature?) of disallowing `DkVirtualMemoryAlloc(/*addr=*/NULL)`.

Also, pay attention that I have `/*prot=*/0` here, which means that this is a "dummy" memory region. We can't even read from it.

pwmarcz (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/255)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/255)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

___
*[Pal/src/db_rtld.c, line 382 at r5](https://reviewable.io/reviews/gramineproject/gramine/255#-MpvdzD3-wC5hR2hOvEn:-MpwJckZ9THepP3yNV0-:bhw9j2s) ([raw file](https://github.com/gramineproject/gramine/blob/cdfedcc014d14159464f54e2f4052e827bed4361/Pal/src/db_rtld.c#L382)):*
> what are the semantics (e.g. is file 0 extended to the requested size?)

Yes, the "exceeding file size" part of the requested memory region is filled with zeros. That's the semantics of `DkStreamMap()`, similar to how file-backed `mmap()` does it.

boryspoplawski (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/255)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

___
*[Pal/src/db_rtld.c, line 382 at r5](https://reviewable.io/reviews/gramineproject/gramine/255#-MpvdzD3-wC5hR2hOvEn:-MpwdlyiBsCQgPRpucLJ:b-uzg7pt) ([raw file](https://github.com/gramineproject/gramine/blob/cdfedcc014d14159464f54e2f4052e827bed4361/Pal/src/db_rtld.c#L382)):*

> what are the semantics (e.g. is file 0 extended to the requested size?)

Yes, the "exceeding file size" part of the requested memory region is filled with zeros. That's the semantics of `DkStreamMap()`, similar to how file-backed `mmap()` does it. 
</blockquote></details>

cc @mkow hahaha :)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/255)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

___
*[Pal/src/db_rtld.c, line 382 at r5](https://reviewable.io/reviews/gramineproject/gramine/255#-MpvdzD3-wC5hR2hOvEn:-MpzCZkU4J3nfNvpAAEz:b-j8mggi) ([raw file](https://github.com/gramineproject/gramine/blob/cdfedcc014d14159464f54e2f4052e827bed4361/Pal/src/db_rtld.c#L382)):*

cc @mkow hahaha :)
</blockquote></details>

I take it back. I looked at the actual `mmap()` semantics and how it is implemented in our PALs. TLDR: no consistency in our PALs.

`mmap()` semantics are as follows:
1. The part of the mapping that exceeds the file size but is still in the last page of file contents is filled with zeros.
2. The part of the mapping that exceeds the file size and also exceeds the last page of file contents produces SIGBUS on any access.

In other words, accesses to pages that contain actual file contents are allowed, all other accesses produce SIGBUS
(See e.g. https://stackoverflow.com/questions/59024621/can-i-mmap-a-file-with-length-greater-than-the-size-of-the-file).

Now how does it work in PALs:
- Linux: `DkStreamMap()` ends up in `file_map()` which simply calls host-level `mmap()`. So semantics are the same.
- Linux-SGX: `DkStreamMap()` ends up in `file_map()` which ultimately calls (for protected, allowed, trusted files) `get_enclave_pages()` which fills the whole mapping with zeros. So semantics are *different* -- SIGBUS is never triggered. (There is probably not much that can be done about this anyway, we're on SGX v1.)

pwmarcz (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/255)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/255)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[Pal/src/db_rtld.c, line 382 at r5](https://reviewable.io/reviews/gramineproject/gramine/255#-MpvdzD3-wC5hR2hOvEn:-MpzWlFIEWm2_FBxeflg:b8dszrs) ([raw file](https://github.com/gramineproject/gramine/blob/cdfedcc014d14159464f54e2f4052e827bed4361/Pal/src/db_rtld.c#L382)):*

I take it back. I looked at the actual `mmap()` semantics and how it is implemented in our PALs. TLDR: no consistency in our PALs.

`mmap()` semantics are as follows:
1. The part of the mapping that exceeds the file size but is still in the last page of file contents is filled with zeros.
2. The part of the mapping that exceeds the file size and also exceeds the last page of file contents produces SIGBUS on any access.

In other words, accesses to pages that contain actual file contents are allowed, all other accesses produce SIGBUS
(See e.g. https://stackoverflow.com/questions/59024621/can-i-mmap-a-file-with-length-greater-than-the-size-of-the-file).

Now how does it work in PALs:
- Linux: `DkStreamMap()` ends up in `file_map()` which simply calls host-level `mmap()`. So semantics are the same.
- Linux-SGX: `DkStreamMap()` ends up in `file_map()` which ultimately calls (for protected, allowed, trusted files) `get_enclave_pages()` which fills the whole mapping with zeros. So semantics are *different* -- SIGBUS is never triggered. (There is probably not much that can be done about this anyway, we're on SGX v1.)
</blockquote></details>

For the context: I was laughing, because I've created a task on a CTF contest we organized last weekend, for which the solution based on the fact that you can violate the rule number 1 in your list :) Basically, there is also file page cache, so if you map a file, write beyond it on the last page, unmap, then successive mapping of this page (even in other processes) will see these changes, as long as the page stays in cache. Which basically means you cannot assume the page is zero filled.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/255)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[Pal/src/db_rtld.c, line 382 at r5](https://reviewable.io/reviews/gramineproject/gramine/255#-MpvdzD3-wC5hR2hOvEn:-MpzgzTf9ajneEr3nBH9:b-o8x90q) ([raw file](https://github.com/gramineproject/gramine/blob/cdfedcc014d14159464f54e2f4052e827bed4361/Pal/src/db_rtld.c#L382)):*

For the context: I was laughing, because I've created a task on a CTF contest we organized last weekend, for which the solution based on the fact that you can violate the rule number 1 in your list :) Basically, there is also file page cache, so if you map a file, write beyond it on the last page, unmap, then successive mapping of this page (even in other processes) will see these changes, as long as the page stays in cache. Which basically means you cannot assume the page is zero filled.
</blockquote></details>

I see. True, there is a note somewhere in `man 2 mmap` about this Linux behavior.

mkow (commented):
> Reviewed 2 of 3 files at r4, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/255)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[Pal/src/db_rtld.c, line 382 at r5](https://reviewable.io/reviews/gramineproject/gramine/255#-MpvdzD3-wC5hR2hOvEn:-Mq-SrDeByc9HwRE65Zh:bpyt9kx) ([raw file](https://github.com/gramineproject/gramine/blob/cdfedcc014d14159464f54e2f4052e827bed4361/Pal/src/db_rtld.c#L382)):*
> Yes, the "exceeding file size" part of the requested memory region is filled with zeros. That's the semantics of `DkStreamMap()`, similar to how file-backed `mmap()` does it.

hehe :)

> The part of the mapping that exceeds the file size but is still in the last page of file contents is filled with zeros.

This nicely explains the very low number of solves for this challenge :)

PR #258, opened 2021-12-02
[Pal] Partially remove `PAL_PTR` type

mkow:
## Description of the changes 

Only less controversial places for now, where the replacement type is quite obvious. Next will come `PAL_NUM`, and `PAL_PTR` occurrences in memory management (not sure yet whether to use `void*` or `uintptr_t` there).

## How to test this PR? 

CI.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/258)

### DIFF
--- a/Documentation/conf.py
+++ b/Documentation/conf.py
@@ -105,8 +105,15 @@ nitpicky = True
 nitpick_ignore = [
     ('c:type', 'bool'),
     ('c:type', 'toml_table_t'),
+    ('c:type', 'uint8_t'),
+    ('c:type', 'uint16_t'),
     ('c:type', 'uint32_t'),
     ('c:type', 'uint64_t'),
+    ('c:type', 'int8_t'),
+    ('c:type', 'int16_t'),
+    ('c:type', 'int32_t'),
+    ('c:type', 'int64_t'),
+    ('c:type', 'uintptr_t'),
     ('c:type', 'union'),
     ('c:type', 'enum'), # parsing of these seems to be broken:
                         #     WARNING: c:type reference target not found: enum
--- a/LibOS/shim/include/arch/x86_64/shim_tcb-arch.h
+++ b/LibOS/shim/include/arch/x86_64/shim_tcb-arch.h
@@ -143,18 +143,18 @@ struct shim_xstate {
         }                                                               \
     } while (0)
 
-static inline void set_tls(unsigned long tls) {
-    DkSegmentBaseSet(PAL_SEGMENT_FS, (PAL_PTR)tls);
+static inline void set_tls(uintptr_t tls) {
+    DkSegmentBaseSet(PAL_SEGMENT_FS, tls);
 }
 
 static inline void set_default_tls(void) {
     set_tls(0);
 }
 
-static inline unsigned long get_tls(void) {
-    void* addr = NULL;
+static inline uintptr_t get_tls(void) {
+    uintptr_t addr = 0;
     (void)DkSegmentBaseGet(PAL_SEGMENT_FS, &addr);
-    return (unsigned long)addr;
+    return addr;
 }
 
 #endif /* _SHIM_TCB_ARCH_H_ */
--- a/LibOS/shim/include/shim_table.h
+++ b/LibOS/shim/include/shim_table.h
@@ -138,7 +138,7 @@ long shim_do_sched_get_priority_max(int policy);
 long shim_do_sched_get_priority_min(int policy);
 long shim_do_sched_rr_get_interval(pid_t pid, struct timespec* interval);
 long shim_do_rt_sigsuspend(const __sigset_t* mask, size_t setsize);
-long shim_do_arch_prctl(int code, void* addr);
+long shim_do_arch_prctl(int code, unsigned long addr);
 long shim_do_setrlimit(int resource, struct __kernel_rlimit* rlim);
 long shim_do_chroot(const char* filename);
 long shim_do_sethostname(char* name, int len);
--- a/LibOS/shim/include/shim_tcb.h
+++ b/LibOS/shim/include/shim_tcb.h
@@ -12,7 +12,7 @@
 struct shim_context {
     PAL_CONTEXT* regs;
     long syscall_nr;
-    unsigned long tls; /* Used only in clone. */
+    uintptr_t tls; /* Used only in clone. */
 };
 
 typedef struct shim_tcb shim_tcb_t;
--- a/LibOS/shim/src/arch/x86_64/shim_arch_prctl.c
+++ b/LibOS/shim/src/arch/x86_64/shim_arch_prctl.c
@@ -12,14 +12,14 @@
 #include "shim_table.h"
 #include "shim_tcb.h"
 
-long shim_do_arch_prctl(int code, void* addr) {
+long shim_do_arch_prctl(int code, unsigned long addr) {
     switch (code) {
         case ARCH_SET_FS:
-            set_tls((unsigned long)addr);
+            set_tls(addr);
             return 0;
 
         case ARCH_GET_FS:
-            return pal_to_unix_errno(DkSegmentBaseGet(PAL_SEGMENT_FS, addr));
+            return pal_to_unix_errno(DkSegmentBaseGet(PAL_SEGMENT_FS, (unsigned long*)addr));
 
         default:
             log_warning("Not supported flag (0x%x) passed to arch_prctl", code);
--- a/LibOS/shim/src/fs/dev/attestation.c
+++ b/LibOS/shim/src/fs/dev/attestation.c
@@ -275,7 +275,7 @@ static int pfkey_save(struct shim_dentry* dent, const char* data, size_t size) {
     char buffer[sizeof(g_pf_key_hex) + 1];
     memcpy(buffer, data, sizeof(g_pf_key_hex));
     buffer[sizeof(g_pf_key_hex)] = '\0';
-    int ret = DkSetProtectedFilesKey(&buffer);
+    int ret = DkSetProtectedFilesKey(buffer);
     if (ret < 0)
         return -EACCES;
 
--- a/LibOS/shim/src/ipc/shim_ipc_worker.c
+++ b/LibOS/shim/src/ipc/shim_ipc_worker.c
@@ -357,7 +357,7 @@ out_die:
     DkProcessExit(1);
 }
 
-static void ipc_worker_wrapper(void* arg) {
+static int ipc_worker_wrapper(void* arg) {
     __UNUSED(arg);
     assert(g_worker_thread);
 
--- a/LibOS/shim/src/shim_async.c
+++ b/LibOS/shim/src/shim_async.c
@@ -139,10 +139,10 @@ int init_async_worker(void) {
     return 0;
 }
 
-static void shim_async_worker(void* arg) {
+static int shim_async_worker(void* arg) {
     struct shim_thread* self = (struct shim_thread*)arg;
     if (!arg)
-        return;
+        return -1;
 
     shim_tcb_init();
     set_cur_thread(self);
--- a/LibOS/shim/src/shim_rtld.c
+++ b/LibOS/shim/src/shim_rtld.c
@@ -915,7 +915,7 @@ noreturn void execute_elf_object(struct link_map* exec_map, void* argp, ElfW(aux
     ElfW(Addr) auxp_extra = (ElfW(Addr))&auxp[8];
 
     ElfW(Addr) random = auxp_extra; /* random 16B for AT_RANDOM */
-    ret = DkRandomBitsRead((PAL_PTR)random, 16);
+    ret = DkRandomBitsRead((void*)random, 16);
     if (ret < 0) {
         log_error("execute_elf_object: DkRandomBitsRead failed: %d", ret);
         DkProcessExit(1);
--- a/LibOS/shim/src/sys/shim_clone.c
+++ b/LibOS/shim/src/sys/shim_clone.c
@@ -48,12 +48,13 @@ struct shim_clone_args {
  * 7.In the wrapper function ,we just do the stack switch to user
  *   Provided stack and execute the user Provided function.
  */
-static int clone_implementation_wrapper(struct shim_clone_args* arg) {
+static int clone_implementation_wrapper(void* arg_) {
     /* The child thread created by PAL is now running on the PAL allocated stack. We need to switch
      * to the user provided stack. */
 
     /* We acquired ownership of arg->thread from the caller, hence there is
      * no need to call get_thread. */
+    struct shim_clone_args* arg = (struct shim_clone_args*)arg_;
     struct shim_thread* my_thread = arg->thread;
     assert(my_thread);
 
--- a/LibOS/shim/src/utils/log.c
+++ b/LibOS/shim/src/utils/log.c
@@ -78,7 +78,7 @@ void log_setprefix(shim_tcb_t* tcb) {
 
 static int buf_write_all(const char* str, size_t size, void* arg) {
     __UNUSED(arg);
-    DkDebugLog((PAL_PTR)str, size);
+    DkDebugLog(str, size);
     return 0;
 }
 
--- a/Pal/include/pal/pal.h
+++ b/Pal/include/pal/pal.h
@@ -336,7 +336,7 @@ int DkStreamWaitForClient(PAL_HANDLE handle, PAL_HANDLE* client);
  * If \p handle is a directory, DkStreamRead fills the buffer with the null-terminated names of the
  * directory entries.
  */
-int DkStreamRead(PAL_HANDLE handle, PAL_NUM offset, PAL_NUM* count, PAL_PTR buffer, PAL_PTR source,
+int DkStreamRead(PAL_HANDLE handle, PAL_NUM offset, PAL_NUM* count, void* buffer, char* source,
                  PAL_NUM size);
 
 /*!
@@ -352,7 +352,7 @@ int DkStreamRead(PAL_HANDLE handle, PAL_NUM offset, PAL_NUM* count, PAL_PTR buff
  *
  * \return 0 on success, negative error code on failure.
  */
-int DkStreamWrite(PAL_HANDLE handle, PAL_NUM offset, PAL_NUM* count, PAL_PTR buffer,
+int DkStreamWrite(PAL_HANDLE handle, PAL_NUM offset, PAL_NUM* count, void* buffer,
                   const char* dest);
 
 enum pal_delete_mode {
@@ -469,7 +469,7 @@ int DkStreamAttributesSetByHandle(PAL_HANDLE handle, PAL_STREAM_ATTR* attr);
 /*!
  * \brief Query the name of an open stream. On success `buffer` contains a null-terminated string.
  */
-int DkStreamGetName(PAL_HANDLE handle, PAL_PTR buffer, PAL_NUM size);
+int DkStreamGetName(PAL_HANDLE handle, char* buffer, PAL_NUM size);
 
 /*!
  * \brief This API changes the name of an open stream.
@@ -487,7 +487,7 @@ int DkStreamChangeName(PAL_HANDLE handle, const char* uri);
  * \param param is the pointer argument that is passed to the new thread
  * \param[out] handle on success contains the thread handle
  */
-int DkThreadCreate(PAL_PTR addr, PAL_PTR param, PAL_HANDLE* handle);
+int DkThreadCreate(int (*callback)(void*), void* param, PAL_HANDLE* handle);
 
 /*!
  * \brief Yield the current thread such that the host scheduler can reschedule it.
@@ -501,7 +501,7 @@ void DkThreadYieldExecution(void);
  *  to notify LibOS (which in turn notifies the parent thread if any); if
  *  `clear_child_tid` is NULL, then PAL doesn't do the clearing.
  */
-noreturn void DkThreadExit(PAL_PTR clear_child_tid);
+noreturn void DkThreadExit(int* clear_child_tid);
 
 /*!
  * \brief Resume a thread.
@@ -520,7 +520,7 @@ int DkThreadResume(PAL_HANDLE thread);
  *
  * \return Returns 0 on success, negative error code on failure.
  */
-int DkThreadSetCpuAffinity(PAL_HANDLE thread, PAL_NUM cpumask_size, PAL_PTR cpu_mask);
+int DkThreadSetCpuAffinity(PAL_HANDLE thread, PAL_NUM cpumask_size, unsigned long* cpu_mask);
 
 /*!
  * \brief Gets the CPU affinity of a thread.
@@ -535,7 +535,7 @@ int DkThreadSetCpuAffinity(PAL_HANDLE thread, PAL_NUM cpumask_size, PAL_PTR cpu_
  *
  * \return Returns 0 on success, negative error code on failure.
  */
-int DkThreadGetCpuAffinity(PAL_HANDLE thread, PAL_NUM cpumask_size, PAL_PTR cpu_mask);
+int DkThreadGetCpuAffinity(PAL_HANDLE thread, PAL_NUM cpumask_size, unsigned long* cpu_mask);
 
 /*
  * Exception Handling
@@ -669,7 +669,7 @@ void DkObjectClose(PAL_HANDLE object_handle);
  *
  * \return 0 on success, negative error code on failure.
  */
-int DkDebugLog(PAL_PTR buffer, PAL_NUM size);
+int DkDebugLog(const void* buffer, PAL_NUM size);
 
 /*!
  * \brief Get the current time
@@ -685,7 +685,7 @@ int DkSystemTimeQuery(PAL_NUM* time);
  * \param[in] size buffer size
  * \return 0 on success, negative on failure
  */
-int DkRandomBitsRead(PAL_PTR buffer, PAL_NUM size);
+int DkRandomBitsRead(void* buffer, PAL_NUM size);
 
 enum pal_segment_reg {
     PAL_SEGMENT_FS,
@@ -700,7 +700,7 @@ enum pal_segment_reg {
  *
  * \return 0 on success, negative error value on failure
  */
-int DkSegmentBaseGet(enum pal_segment_reg reg, PAL_PTR* addr);
+int DkSegmentBaseGet(enum pal_segment_reg reg, uintptr_t* addr);
 
 /*!
  * \brief Set segment register
@@ -710,7 +710,7 @@ int DkSegmentBaseGet(enum pal_segment_reg reg, PAL_PTR* addr);
  *
  * \return 0 on success, negative error value on failure
  */
-int DkSegmentBaseSet(enum pal_segment_reg reg, PAL_PTR addr);
+int DkSegmentBaseSet(enum pal_segment_reg reg, uintptr_t addr);
 
 /*!
  * \brief Return the amount of currently available memory for LibOS/application
@@ -748,8 +748,8 @@ PAL_NUM DkMemoryAvailableQuota(void);
  * \param[in,out] report_size            Caller specifies size of `report`; on return, contains
  *                                       PAL-enforced size of `report` (432B in case of SGX PAL).
  */
-int DkAttestationReport(PAL_PTR user_report_data, PAL_NUM* user_report_data_size,
-                        PAL_PTR target_info, PAL_NUM* target_info_size, PAL_PTR report,
+int DkAttestationReport(const void* user_report_data, PAL_NUM* user_report_data_size,
+                        void* target_info, PAL_NUM* target_info_size, void* report,
                         PAL_NUM* report_size);
 
 /*!
@@ -767,7 +767,7 @@ int DkAttestationReport(PAL_PTR user_report_data, PAL_NUM* user_report_data_size
  * \param[in,out] quote_size             Caller specifies maximum size allocated for `quote`; on
  *                                       return, contains actual size of obtained quote.
  */
-int DkAttestationQuote(PAL_PTR user_report_data, PAL_NUM user_report_data_size, PAL_PTR quote,
+int DkAttestationQuote(const void* user_report_data, PAL_NUM user_report_data_size, void* quote,
                        PAL_NUM* quote_size);
 
 /*!
@@ -779,7 +779,7 @@ int DkAttestationQuote(PAL_PTR user_report_data, PAL_NUM user_report_data_size, 
  * \param[in]     pf_key_hex       Wrap key for protected files. Must be a 32-char null-terminated
  *                                 hex string in case of SGX PAL (AES-GCM encryption key).
  */
-int DkSetProtectedFilesKey(PAL_PTR pf_key_hex);
+int DkSetProtectedFilesKey(const char* pf_key_hex);
 
 #ifdef __GNUC__
 #define symbol_version_default(real, name, version) \
--- a/Pal/include/pal_internal.h
+++ b/Pal/include/pal_internal.h
@@ -188,14 +188,14 @@ int _DkSendHandle(PAL_HANDLE hdl, PAL_HANDLE cargo);
 int _DkReceiveHandle(PAL_HANDLE hdl, PAL_HANDLE* cargo);
 
 /* DkProcess and DkThread calls */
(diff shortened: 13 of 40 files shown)


dimakuv (changes requested):
> Reviewed 39 of 39 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/258)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @mkow)

___
*[Pal/regression/Event.c, line 29 at r1](https://reviewable.io/reviews/gramineproject/gramine/258#-MpuLXycElqPAQ9jIyOL:-MpuLXycElqPAQ9jIyOM:b77s1t1) ([raw file](https://github.com/gramineproject/gramine/blob/b9d61494df77074d2ad50e6d7ddee8a16d7decb7/Pal/regression/Event.c#L29)):*
> ```C
> static int g_ready = 0;
> 
> static int thread_func(void* arg) {
> ```

You don't have any `return` in this func
___
*[Pal/regression/Thread.c, line 11 at r1](https://reviewable.io/reviews/gramineproject/gramine/258#-MpuLhBO17R3sGnTBspM:-MpuLhBO17R3sGnTBspN:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/b9d61494df77074d2ad50e6d7ddee8a16d7decb7/Pal/regression/Thread.c#L11)):*
> ```C
> static atomic_int count = 0;
> 
> static int callback(void* args) {
> ```

ditto
___
*[Pal/src/host/Linux/pal_host.h, line 87 at r1](https://reviewable.io/reviews/gramineproject/gramine/258#-MpuMKek6Sh1BhwGQTO1:-MpuMKek6Sh1BhwGQTO2:b2nu0dj) ([raw file](https://github.com/gramineproject/gramine/blob/b9d61494df77074d2ad50e6d7ddee8a16d7decb7/Pal/src/host/Linux/pal_host.h#L87)):*
> ```objc
>             PAL_IDX fd;
>             void* bind;
>             void* conn;
> ```

Shouldn't these two be `struct sockaddr*`? These are TCP/UDP addresses, and we're inside Linux host, so it's always `struct sockaddr` here.
___
*[Pal/src/host/Linux/arch/x86_64/db_arch_misc.c, line 233 at r1](https://reviewable.io/reviews/gramineproject/gramine/258#-MpuMg6o9V7B4tEWdJlb:-MpuMg6o9V7B4tEWdJlc:b-6w9mcm) ([raw file](https://github.com/gramineproject/gramine/blob/b9d61494df77074d2ad50e6d7ddee8a16d7decb7/Pal/src/host/Linux/arch/x86_64/db_arch_misc.c#L233)):*
> ```C
>     switch (reg) {
>         case PAL_SEGMENT_FS:
>             return unix_to_pal_error(DO_SYSCALL(arch_prctl, (unsigned long)ARCH_GET_FS, addr));
> ```

Why do you need this cast? Looks like you put it at the wrong argument :)
___
*[Pal/src/host/Linux-SGX/db_sockets.c, line 233 at r1](https://reviewable.io/reviews/gramineproject/gramine/258#-MpuN7HW2L7ypf149YHW:-MpuN7HW2L7ypf149YHX:b-vwbqn9) ([raw file](https://github.com/gramineproject/gramine/blob/b9d61494df77074d2ad50e6d7ddee8a16d7decb7/Pal/src/host/Linux-SGX/db_sockets.c#L233)):*
> ```C
>     HANDLE_HDR(hdl)->flags |= RFD(0) | (type != PAL_TYPE_TCPSRV ? WFD(0) : 0);
>     hdl->sock.fd = fd;
>     void* addr = (void*)hdl + HANDLE_SIZE(sock);
> ```

As Borys likes to point out, the cast should be to char: `(char*)hdl`

And `void* addr` should actually be `struct sockaddr*`, but on the other hand, I don't mind to cast it later, during `hdl->sock.bind/conn` assignment.
___
*[Pal/src/host/Linux-SGX/pal_host.h, line 103 at r1](https://reviewable.io/reviews/gramineproject/gramine/258#-MpuOCc7--zxTxwERsqM:-MpuOCc7--zxTxwERsqN:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/b9d61494df77074d2ad50e6d7ddee8a16d7decb7/Pal/src/host/Linux-SGX/pal_host.h#L103)):*
> ```objc
>             PAL_IDX fd;
>             void* bind;
>             void* conn;
> ```

ditto

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/258)* status: 38 of 39 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[Pal/regression/Event.c, line 29 at r1](https://reviewable.io/reviews/gramineproject/gramine/258#-MpuLXycElqPAQ9jIyOL:-MpvA8ao6WKXl2p_JhUS:b-kzjjj2) ([raw file](https://github.com/gramineproject/gramine/blob/b9d61494df77074d2ad50e6d7ddee8a16d7decb7/Pal/regression/Event.c#L29)):*

You don't have any `return` in this func
</blockquote></details>

Yes, and why is this a problem? `DkThreadExit()` doesn't return.
I fixed this type because it's passed as a callback, and before we just erased it type to `void*` and the type mismatch wasn't detected.
___
*[Pal/src/host/Linux/pal_host.h, line 87 at r1](https://reviewable.io/reviews/gramineproject/gramine/258#-MpuMKek6Sh1BhwGQTO1:-MpvAXqMEdwMjN5QPOzx:b-vbql6n) ([raw file](https://github.com/gramineproject/gramine/blob/b9d61494df77074d2ad50e6d7ddee8a16d7decb7/Pal/src/host/Linux/pal_host.h#L87)):*

Shouldn't these two be `struct sockaddr*`? These are TCP/UDP addresses, and we're inside Linux host, so it's always `struct sockaddr` here.
</blockquote></details>

Hmm, maybe. I though it depends on something and may actually differ, but seems you're right.

OTOH, we end up in the land of poor-man's polymorphism of C, because in some places we use it as `struct addr_inet*` and I'd need to add casts there. Should I do this or keep as `void*`?
___
*[Pal/src/host/Linux/arch/x86_64/db_arch_misc.c, line 233 at r1](https://reviewable.io/reviews/gramineproject/gramine/258#-MpuMg6o9V7B4tEWdJlb:-MpvGQXxDeK8_wS7BTGL:b-mxdjix) ([raw file](https://github.com/gramineproject/gramine/blob/b9d61494df77074d2ad50e6d7ddee8a16d7decb7/Pal/src/host/Linux/arch/x86_64/db_arch_misc.c#L233)):*

Why do you need this cast? Looks like you put it at the wrong argument :)
</blockquote></details>

Oh, too many replaces -.- Sorry.
Not sure whether to add a static assert about the sizes here, if so, I should add it to other similar places also.
___
*[Pal/src/host/Linux-SGX/db_sockets.c, line 233 at r1](https://reviewable.io/reviews/gramineproject/gramine/258#-MpuN7HW2L7ypf149YHW:-MpvHe6g-XbWNg__rVCU:b-mqg0oz) ([raw file](https://github.com/gramineproject/gramine/blob/b9d61494df77074d2ad50e6d7ddee8a16d7decb7/Pal/src/host/Linux-SGX/db_sockets.c#L233)):*

As Borys likes to point out, the cast should be to char: `(char*)hdl`

And `void* addr` should actually be `struct sockaddr*`, but on the other hand, I don't mind to cast it later, during `hdl->sock.bind/conn` assignment.
</blockquote></details>

Let's resolve the discussion above before fixing this place.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/258)* status: 31 of 40 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[Pal/regression/Event.c, line 29 at r1](https://reviewable.io/reviews/gramineproject/gramine/258#-MpuLXycElqPAQ9jIyOL:-MpvTxdMCrFhjrNDVmku:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/b9d61494df77074d2ad50e6d7ddee8a16d7decb7/Pal/regression/Event.c#L29)):*

Maybe at least add `noreturn`? Not blocking though.

But now I understand why this change at all.
</blockquote></details>

Done.
___
*[Pal/regression/Thread.c, line 11 at r1](https://reviewable.io/reviews/gramineproject/gramine/258#-MpuLhBO17R3sGnTBspM:-MpvU2ZV7mCrj3l81QTl:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/b9d61494df77074d2ad50e6d7ddee8a16d7decb7/Pal/regression/Thread.c#L11)):*

`noreturn`? Not blocking
</blockquote></details>

Done.
___
*[Pal/src/host/Linux/pal_host.h, line 87 at r1](https://reviewable.io/reviews/gramineproject/gramine/258#-MpuMKek6Sh1BhwGQTO1:-MpvVCw3Cy0xr5LAz-hn:b-ak5o6d) ([raw file](https://github.com/gramineproject/gramine/blob/b9d61494df77074d2ad50e6d7ddee8a16d7decb7/Pal/src/host/Linux/pal_host.h#L87)):*

I would prefer `struct sockaddr*` instead of `void*`. It at least gives you some idea what these fields contain (in contrast to `void*` which could be interpreted as `bind/conn` being human-readable strings).
</blockquote></details>

Done. It actually removed more casts than added, so it looks nice now.
___
*[Pal/src/host/Linux/arch/x86_64/db_arch_misc.c, line 233 at r1](https://reviewable.io/reviews/gramineproject/gramine/258#-MpuMg6o9V7B4tEWdJlb:-MpvYVMG5vkaEmjaCtsQ:btq15yl) ([raw file](https://github.com/gramineproject/gramine/blob/b9d61494df77074d2ad50e6d7ddee8a16d7decb7/Pal/src/host/Linux/arch/x86_64/db_arch_misc.c#L233)):*

That the sizes `uintptr_t` and `unsigned long` match? I don't think you need it, you're under `arch/x86_64/`
</blockquote></details>

Yup, that thing.
x86_x64 is not enough, C standard allows for different sizes for `unsigned long` (and e.g. MSVC does that), but OTOH the whole Linux-related codebase already assumes that `unsigned long` is the same as `uintptr_t`.
___
*[Pal/src/host/Linux-SGX/db_sockets.c, line 233 at r1](https://reviewable.io/reviews/gramineproject/gramine/258#-MpuN7HW2L7ypf149YHW:-MpvVsboCMxYInQ2h0j0:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/b9d61494df77074d2ad50e6d7ddee8a16d7decb7/Pal/src/host/Linux-SGX/db_sockets.c#L233)):*

Let's resolve the discussion above before fixing this place.
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/pal_host.h, line 103 at r1](https://reviewable.io/reviews/gramineproject/gramine/258#-MpuOCc7--zxTxwERsqM:-MpvVxa5BRlXY2T87xgg:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/b9d61494df77074d2ad50e6d7ddee8a16d7decb7/Pal/src/host/Linux-SGX/pal_host.h#L103)):*

ditto
</blockquote></details>

Done.

dimakuv (changes requested):
> Reviewed 9 of 9 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/258)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[Pal/src/host/Linux/arch/x86_64/db_arch_misc.c, line 233 at r1](https://reviewable.io/reviews/gramineproject/gramine/258#-MpuMg6o9V7B4tEWdJlb:-MpvbL6-Cn3wGq9yVX4L:b-946giw) ([raw file](https://github.com/gramineproject/gramine/blob/b9d61494df77074d2ad50e6d7ddee8a16d7decb7/Pal/src/host/Linux/arch/x86_64/db_arch_misc.c#L233)):*
> MSVC does that

Mind blown.
___
*[Pal/src/host/Linux-SGX/db_sockets.c, line 233 at r1](https://reviewable.io/reviews/gramineproject/gramine/258#-MpuN7HW2L7ypf149YHW:-MpvbRs26vXD8GVVEFNj:bmsdjh4) ([raw file](https://github.com/gramineproject/gramine/blob/b9d61494df77074d2ad50e6d7ddee8a16d7decb7/Pal/src/host/Linux-SGX/db_sockets.c#L233)):*

Done.
</blockquote></details>

But you still need a cast to `char*` (or `uint8_t*`) instead of `void*`

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/258)* status: 38 of 40 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[Pal/src/host/Linux-SGX/db_sockets.c, line 233 at r1](https://reviewable.io/reviews/gramineproject/gramine/258#-MpuN7HW2L7ypf149YHW:-MpvdQO02bnArN1BZ2nc:b-kq024x) ([raw file](https://github.com/gramineproject/gramine/blob/b9d61494df77074d2ad50e6d7ddee8a16d7decb7/Pal/src/host/Linux-SGX/db_sockets.c#L233)):*

But you still need a cast to `char*` (or `uint8_t*`) instead of `void*`
</blockquote></details>

Oops, I just missed this place, sorry :)

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/258)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

boryspoplawski (changes requested):
> Reviewed 29 of 39 files at r1, 1 of 1 files at r2, 8 of 9 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/258)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/258#-MpwbDjX5gv-twt4_GJ_:-MpwbDjX5gv-twt4_GJa:b-snllc) (no related file):*
You've opted for casting things to `uint8_t*` when doing pointer arithmetic, why not `char*`? It's shorter :)

---
___
*[LibOS/shim/src/arch/x86_64/shim_arch_prctl.c, line 22 at r4](https://reviewable.io/reviews/gramineproject/gramine/258#-MpwVMlTDMlJNi9PpK6K:-MpwVMlUC3r0FXageMMN:b-e9n62w) ([raw file](https://github.com/gramineproject/gramine/blob/d6006b7258e93857154478506c11222b78dde120/LibOS/shim/src/arch/x86_64/shim_arch_prctl.c#L22)):*
> ```C
> 
>         case ARCH_GET_FS:
>             return pal_to_unix_errno(DkSegmentBaseGet(PAL_SEGMENT_FS, (uintptr_t*)addr));
> ```

This should be `unsigned long*` - that's the type syscall callers put here (and it either matches `uintptr_t*` - ok - or it does not - compiler error, also ok).
___
*[Pal/include/pal/pal.h, line 490 at r4](https://reviewable.io/reviews/gramineproject/gramine/258#-MpvzbVL39ur5ovLoav-:-MpvzbVL39ur5ovLoav0:b-gs4qci) ([raw file](https://github.com/gramineproject/gramine/blob/d6006b7258e93857154478506c11222b78dde120/Pal/include/pal/pal.h#L490)):*
> ```objc
>  * \param[out] handle on success contains the thread handle
>  */
> int DkThreadCreate(int (*callback)(void*), const void* param, PAL_HANDLE* handle);
> ```

Why `callback` gets a `void*` and `param` is `const`? I think they should be the same type (most likely not `const`).
___
*[Pal/src/host/Linux/db_streams.c, line 167 at r4](https://reviewable.io/reviews/gramineproject/gramine/258#-Mpw1abf4leTfzTNAmCF:-Mpw1abf4leTfzTNAmCG:b-javkwp) ([raw file](https://github.com/gramineproject/gramine/blob/d6006b7258e93857154478506c11222b78dde120/Pal/src/host/Linux/db_streams.c#L167)):*
> ```C
>     switch (PAL_GET_TYPE(hdl)) {
>         case PAL_TYPE_FILE:
>             hdl->file.realpath = (const char*)(hdl->file.realpath ? (uint8_t*)hdl + hdlsz : NULL);
> ```

Why the inner cast is not `const char*`? Then the outer could be removed.
___
*[Pal/src/host/Linux/db_streams.c, line 177 at r4](https://reviewable.io/reviews/gramineproject/gramine/258#-Mpw1l7tBsFMoOhjdKvk:-Mpw1l7tBsFMoOhjdKvl:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/d6006b7258e93857154478506c11222b78dde120/Pal/src/host/Linux/db_streams.c#L177)):*
> ```C
>             break;
>         case PAL_TYPE_DIR:
>             hdl->dir.realpath = (const char*)(hdl->dir.realpath ? (uint8_t*)hdl + hdlsz : NULL);
> ```

ditto
___
*[Pal/src/host/Linux/arch/x86_64/db_arch_misc.c, line 233 at r1](https://reviewable.io/reviews/gramineproject/gramine/258#-MpuMg6o9V7B4tEWdJlb:-MpwPgX05ZD8ybXAXwe_:bsudrf) ([raw file](https://github.com/gramineproject/gramine/blob/b9d61494df77074d2ad50e6d7ddee8a16d7decb7/Pal/src/host/Linux/arch/x86_64/db_arch_misc.c#L233)):*

> MSVC does that

Mind blown.
</blockquote></details>

You are under `Linux/arch/x86_64` where you **can** safely assume LP64 model.
Besides, what do you want to achieve with this cast?
___
*[Pal/src/host/Linux-SGX/enclave_tf.h, line 15 at r4](https://reviewable.io/reviews/gramineproject/gramine/258#-MpwS7E27sWudBYzGe_w:-MpwS7E27sWudBYzGe_x:b-m2w85d) ([raw file](https://github.com/gramineproject/gramine/blob/a498b73657bf6978d7ee9df208eb018f60919e8f/Pal/src/host/Linux-SGX/enclave_tf.h#L15)):*
> ```objc
>  * verification in future reads, to avoid re-verifying the whole file again or the need of caching
>  * file contents.
>  *
> ```

Why this split of headers? Cannot we just include this header instead of the one newly created?

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/258)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/258#-MpwbDjX5gv-twt4_GJ_:-MpyPFpv7yTL-iAfm-6S:b-elc4sr) (no related file):*

You've opted for casting things to `uint8_t*` when doing pointer arithmetic, why not `char*`? It's shorter :)
</blockquote></details>

I just dislike `char` for its ambiguous definition in the standard, same as other "old" C types like `int` ;) It's not even specified if it's signed or not.

`uint8_t` is just what it says, although I guess it may be problematic for non-8-bit-bytes archs, but in this case we are doomed anyways for tons of other reasons.

---
___
*[LibOS/shim/src/arch/x86_64/shim_arch_prctl.c, line 22 at r4](https://reviewable.io/reviews/gramineproject/gramine/258#-MpwVMlTDMlJNi9PpK6K:-MpyRB9f8-YbNA4tTL9W:bx1y80f) ([raw file](https://github.com/gramineproject/gramine/blob/d6006b7258e93857154478506c11222b78dde120/LibOS/shim/src/arch/x86_64/shim_arch_prctl.c#L22)):*

This should be `unsigned long*` - that's the type syscall callers put here (and it either matches `uintptr_t*` - ok - or it does not - compiler error, also ok).
</blockquote></details>

Hmm, maybe then just leave it without the cast? The effective result will be the same, but at least it won't have this confusing cast to another type than the real argument.
___
*[Pal/include/pal/pal.h, line 490 at r4](https://reviewable.io/reviews/gramineproject/gramine/258#-MpvzbVL39ur5ovLoav-:-MpySseHAuQTloaPl5WO:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/d6006b7258e93857154478506c11222b78dde120/Pal/include/pal/pal.h#L490)):*

Why `callback` gets a `void*` and `param` is `const`? I think they should be the same type (most likely not `const`).
</blockquote></details>

Done.
___
*[Pal/src/host/Linux/db_streams.c, line 167 at r4](https://reviewable.io/reviews/gramineproject/gramine/258#-Mpw1abf4leTfzTNAmCF:-MpySwI19iRYA-SzCOk6:bv64xkf) ([raw file](https://github.com/gramineproject/gramine/blob/d6006b7258e93857154478506c11222b78dde120/Pal/src/host/Linux/db_streams.c#L167)):*

Why the inner cast is not `const char*`? Then the outer could be removed.
</blockquote></details>

I prefer to use `uint8_t*` or `uintptr_t` for pointer arithmetic (see the discussion above), but if we decide to use `char` then I'll change it here.
___
*[Pal/src/host/Linux/arch/x86_64/db_arch_misc.c, line 233 at r1](https://reviewable.io/reviews/gramineproject/gramine/258#-MpuMg6o9V7B4tEWdJlb:-MpyTJ_H-y83c2AC6krA:bsh3avi) ([raw file](https://github.com/gramineproject/gramine/blob/b9d61494df77074d2ad50e6d7ddee8a16d7decb7/Pal/src/host/Linux/arch/x86_64/db_arch_misc.c#L233)):*
> You are under `Linux/arch/x86_64` where you **can** safely assume LP64 model.

Hmm, not sure, I think this is a compiler property, not platform+arch.

Anyways, I just want to always pass exactly the same type as in the declarations in variadics (although `DO_SYSCALL` is kind of special, I still feel safer when passing an exactly-matching type).
___
*[Pal/src/host/Linux-SGX/enclave_tf.h, line 15 at r4](https://reviewable.io/reviews/gramineproject/gramine/258#-MpwS7E27sWudBYzGe_w:-MpyUTv56TM4KejQ60Uq:b-d6zb9b) ([raw file](https://github.com/gramineproject/gramine/blob/a498b73657bf6978d7ee9df208eb018f60919e8f/Pal/src/host/Linux-SGX/enclave_tf.h#L15)):*

Why this split of headers? Cannot we just include this header instead of the one newly created?
</blockquote></details>

Unfortunately not, because we would end up with cyclic dependency. Welcome to C.

Without this split `pal_host.h` uses `sgx_chunk_hash_t` (defined in `enclave_tf.h`), but `enclave_tf.h` uses `PAL_HANDLE` defined in `pal_host.h`.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/258)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/258#-MpwbDjX5gv-twt4_GJ_:-Mq-vCpCA2BPMZShUCIh:b-92r7ky) (no related file):*

I just dislike `char` for its ambiguous definition in the standard, same as other "old" C types like `int` ;) It's not even specified if it's signed or not.

`uint8_t` is just what it says, although I guess it may be problematic for non-8-bit-bytes archs, but in this case we are doomed anyways for tons of other reasons.
</blockquote></details>

But you are using this only for the fact that `sizeof(char) == 1` and this is well defined. I don't see how signedness applies here.

Let's not talk about non 8-bit archs, that's not gonna happen (in Gramine) like ever©

---
___
*[LibOS/shim/src/arch/x86_64/shim_arch_prctl.c, line 22 at r4](https://reviewable.io/reviews/gramineproject/gramine/258#-MpwVMlTDMlJNi9PpK6K:-Mq-vi5R3TSYnabnHtHF:b-s1oeyh) ([raw file](https://github.com/gramineproject/gramine/blob/d6006b7258e93857154478506c11222b78dde120/LibOS/shim/src/arch/x86_64/shim_arch_prctl.c#L22)):*

Hmm, maybe then just leave it without the cast? The effective result will be the same, but at least it won't have this confusing cast to another type than the real argument.
</blockquote></details>

Unfortunately we need the cast - this functions has 2 signatures, one taking `unsigned long` one `unsigned long*` depending on the value of `code` :(((
___
*[Pal/src/host/Linux/db_streams.c, line 167 at r4](https://reviewable.io/reviews/gramineproject/gramine/258#-Mpw1abf4leTfzTNAmCF:-Mq-wGv34eJlvN7j0QFf:byygwqu) ([raw file](https://github.com/gramineproject/gramine/blob/d6006b7258e93857154478506c11222b78dde120/Pal/src/host/Linux/db_streams.c#L167)):*

I prefer to use `uint8_t*` or `uintptr_t` for pointer arithmetic (see the discussion above), but if we decide to use `char` then I'll change it here.
</blockquote></details>

I don't mind either being used, I think it depends on the specific case. I usually use `char` because it's shorted; I don't think we need uniformity across whole codebase w.r.t. to this.

As for this specific place: I think here would be better to have `char` even if every other place uses `uint8_t` - it's just simpler (1 cast less) and equally well defined.
___
*[Pal/src/host/Linux/arch/x86_64/db_arch_misc.c, line 233 at r1](https://reviewable.io/reviews/gramineproject/gramine/258#-MpuMg6o9V7B4tEWdJlb:-Mq-wxefAyON0AwkeyLC:bpxqqr0) ([raw file](https://github.com/gramineproject/gramine/blob/b9d61494df77074d2ad50e6d7ddee8a16d7decb7/Pal/src/host/Linux/arch/x86_64/db_arch_misc.c#L233)):*
> Hmm, not sure, I think this is a compiler property, not platform+arch.

You wouldn't be able to use syscalls if your compiler could have different model than the one required by the system (of course compiler used on that system only).

> Anyways, I just want to always pass exactly the same type as in the declarations in variadics (although DO_SYSCALL is kind of special, I still feel safer when passing an exactly-matching type).

But it doesn't change anything here, both are pointers (if they were integers that would be a different story). In such case casting can only hide problems, never solve them.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/258)* status: 38 of 40 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[LibOS/shim/src/arch/x86_64/shim_arch_prctl.c, line 22 at r4](https://reviewable.io/reviews/gramineproject/gramine/258#-MpwVMlTDMlJNi9PpK6K:-MqAvAEyB81Nv2AJev4d:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/d6006b7258e93857154478506c11222b78dde120/LibOS/shim/src/arch/x86_64/shim_arch_prctl.c#L22)):*

Unfortunately we need the cast - this functions has 2 signatures, one taking `unsigned long` one `unsigned long*` depending on the value of `code` :(((
</blockquote></details>

Done.
___
*[Pal/src/host/Linux/db_streams.c, line 167 at r4](https://reviewable.io/reviews/gramineproject/gramine/258#-Mpw1abf4leTfzTNAmCF:-MqAwS3H-uhGGwywOR5u:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/d6006b7258e93857154478506c11222b78dde120/Pal/src/host/Linux/db_streams.c#L167)):*

I don't mind either being used, I think it depends on the specific case. I usually use `char` because it's shorted; I don't think we need uniformity across whole codebase w.r.t. to this.

As for this specific place: I think here would be better to have `char` even if every other place uses `uint8_t` - it's just simpler (1 cast less) and equally well defined.
</blockquote></details>

Done.
___
*[Pal/src/host/Linux/db_streams.c, line 177 at r4](https://reviewable.io/reviews/gramineproject/gramine/258#-Mpw1l7tBsFMoOhjdKvk:-MqAwZGo-nDy5Rbn5rn7:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/d6006b7258e93857154478506c11222b78dde120/Pal/src/host/Linux/db_streams.c#L177)):*

ditto
</blockquote></details>

Done.
___
*[Pal/src/host/Linux/arch/x86_64/db_arch_misc.c, line 233 at r1](https://reviewable.io/reviews/gramineproject/gramine/258#-MpuMg6o9V7B4tEWdJlb:-MqAxIXT7XqGNK-JRWvF:b5orkl7) ([raw file](https://github.com/gramineproject/gramine/blob/b9d61494df77074d2ad50e6d7ddee8a16d7decb7/Pal/src/host/Linux/arch/x86_64/db_arch_misc.c#L233)):*
> You wouldn't be able to use syscalls if your compiler could have different model than the one required by the system (of course compiler used on that system only).

Depends how you think about system interface. If you treat it as ABI (for Linux I think this is the right "definition") and provide your own headers then you can write in anything you want.

> But it doesn't change anything here, both are pointers (if they were integers that would be a different story). In such case casting can only hide problems, never solve them.

I feel safer if I pass the correct types to a variadic. Also, this is kind of a "comment" saying explicitly that the syscall takes an argument of this type (which is not checked by `DO_SYSCALL` at all).

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/258)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

___
*[Pal/src/host/Linux/arch/x86_64/db_arch_misc.c, line 233 at r1](https://reviewable.io/reviews/gramineproject/gramine/258#-MpuMg6o9V7B4tEWdJlb:-MqB4k6c6qohaWpIBtTJ:b-t28brg) ([raw file](https://github.com/gramineproject/gramine/blob/b9d61494df77074d2ad50e6d7ddee8a16d7decb7/Pal/src/host/Linux/arch/x86_64/db_arch_misc.c#L233)):*
> If you treat it as ABI (for Linux I think this is the right "definition") and provide your own headers then you can write in anything you want.

These would be just wrappers, moving the problem to the moment when you compile these wrappers.

> I feel safer if I pass the correct types to a variadic.

Data pointer is a data pointer. Either the underlying type mach (and there is no problem) or it does not and the cast hides the problem. But this case is quite specific, because it's `DO_SYSCALL` macro, which hides the types anyway.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/258)* status: 38 of 40 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski and @dimakuv)

___
*[Pal/src/host/Linux/arch/x86_64/db_arch_misc.c, line 233 at r1](https://reviewable.io/reviews/gramineproject/gramine/258#-MpuMg6o9V7B4tEWdJlb:-MqBBMQh9Hb27oxpt4NY:btya45i) ([raw file](https://github.com/gramineproject/gramine/blob/b9d61494df77074d2ad50e6d7ddee8a16d7decb7/Pal/src/host/Linux/arch/x86_64/db_arch_misc.c#L233)):*
> These would be just wrappers, moving the problem to the moment when you compile these wrappers.

But you don't have to write them in C if you treat this as an ABI.

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/258)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (approved):
> Reviewed 7 of 7 files at r5, 2 of 2 files at r6, 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/258)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #275, opened 2021-12-09
[Pal/Linux-SGX] Allow different masks for SGX sealing key derivation

### LINKED ISSUE #274: [Pal/Linux-SGX] Different sealing keys generated via EGETKEY on different versions of Gramine
## Description of the problem

(I am using the "Intel® 64 and IA-32 Architectures Software Developer’s Manual Combined Volumes: 1, 2A, 2B, 2C, 2D, 3A, 3B, 3C, 3D and 4" document; version 325462-075US, June 2021. To understand the context and details of the below issue, please use this document: Section 33.18 "KEY REQUEST (KEYREQUEST)" and Section 36.4 "EGETKEY". Please carefully read through the Table 36-64 and through the **complete pseudo-code of EGETKEY instruction**. Be aware that the EGETKEY documentation is a bit misleading and hard to read! Your best bet is to analyze the pseudo-code -- it is correct.)

(If you don't have the Intel SDM document at hand, you can refer to https://www.felixcloutier.com/x86/egetkey.)

Gramine provides support for sealing keys. In particular, there is the `sgx.protected_mrsigner_files = [ list of files ]` manifest option.

Under the hood, Gramine's Linux-SGX PAL generates an MRSIGNER-bound sealing key (`SEAL_KEY`). This key is used to seal (encrypt) and unseal (decrypt) files listed in `sgx.protected_mrsigner_files`.

This MRSIGNED-bound sealing allows to e.g. seal a file in one SGX enclave and unseal this file in another SGX enclave. The main requirement is that both SGX enclaves are signed by the same key (i.e., both enclaves have the same signer). Another requirement is that both SGX enclaves must execute on the same physical machine (this is the SGX hardware requirement).

In particular, this means that technically, one SGX enclave created with an *older* version of Gramine can seal the file on the machine, and then the same SGX enclave but created with a *newer* version of Gramine should be able to unseal this file.

However, this is not the case currently. We have a bug in Gramine which goes as follows.

### Bug

`EGETKEY` instruction generates the Sealing Key based on the key derivation material (supplied to EGETKEY in the `KEYREQUEST` struct). This key derivation material is quite complex, see Section 36.4 "EGETKEY". Note that the exact key derivation material is different for different KeyNames (we care only about the Sealing Key in this issue).

The overlooked problem in Gramine implementation is this. In addition to the MRSIGNER field and other obvious fields in KEYREQUEST struct, there is something more subtle: **the fields `CET_ATTRIBUTES_MASK`, `ATTRIBUTEMASK`, `MISCMASK` do not only define which attribute bits will be included in key derivation, but these mask fields themselves are included in key derivation**.

This can be seen in the pseudo-code of the EGETKEY instruction. One relevant piece as an example:
```
CASE (DS:RBX.KEYNAME)
    SEAL_KEY:
        [ ... irrelevant stuff ... ]
        //Determine values key is based on
        TMP_KEYDEPENDENCIES.ATTRIBUTES := TMP_ATTRIBUTES;
        TMP_KEYDEPENDENCIES.ATTRIBUTESMASK := DS:RBX.ATTRIBUTEMASK;  // <--- KEYREQUEST.ATTRIBUTEMASK is a key-derivation material!
```

Gramine produces a Sealing Key like this: https://github.com/gramineproject/gramine/blob/16c91f594a0feb2c2417929e045331576fada322/Pal/src/host/Linux-SGX/enclave_framework.c#L193-L221

This line of code in particular is important:
https://github.com/gramineproject/gramine/blob/16c91f594a0feb2c2417929e045331576fada322/Pal/src/host/Linux-SGX/enclave_framework.c#L212

So now the problem becomes obvious: when `KEYREQUEST.ATTRIBUTEMASK` is changed even by one bit, the produced Sealing Key will be different. This is what happened recently in Gramine: we changed the `ATTRIBUTEMASK` (more specifically, we changed the `XFRM` part of the `ATTRIBUTEMASK`):

- Gramine v1.0 had `ATTRIBUTEMASK.XFRM = 0xffffffffffffff1b`: https://github.com/gramineproject/gramine/blob/v1.0/Pal/src/host/Linux-SGX/sgx_arch.h#L117-L118
- Gramine commit 497847c ("[Pal/Linux-SGX] Add sgx.protected_mr{enclave,signer}_files manifest options") changed it to `ATTRIBUTEMASK.XFRM = 0xfffffffffff9ff1b`: https://github.com/gramineproject/gramine/blob/497847c0353a13c9e83c0ec4c0cbe99f11d4a75d/Pal/src/host/Linux-SGX/sgx_arch.h#L113?plain=1

Because of this change and because `ATTRIBUTEMASK` is used as key-derivation material, the produced Sealing Keys are different, even if all other key-derivation parameters are exactly the same.

## Steps to reproduce

```bash
# --- seal file using older version of Gramine ---
$ git checkout v1.0
$ ninja -C build-debug/ install
$ cd LibOS/shim/test/regression/
$ make SGX=1 sgx-tokens -j20

# create sealed file `sealed_file_mrsigner.dat` anew
$ rm -rf sealed_file_mrsigner.dat
$ gramine-sgx sealed_file sealed_file_mrsigner.dat
CREATION OK

# copy the sealed file to home dir, just to make sure it doesn't disappear
$ cp sealed_file_mrsigner.dat ~/
$ make clean

# --- try to unseal file using newer version of Gramine ---
$ git checkout master
$ ninja -C build-debug/ install
$ cd LibOS/shim/test/regression/
$ SGX=1 gramine-test build

# copy the sealed file from home dir
$ cp ~/sealed_file_mrsigner.dat ./

# try to unseal the file
$ gramine-sgx sealed_file sealed_file_mrsigner.dat
sealed_file: access failed: Permission denied   # BUG!!!
```

## Proposed fix

Intel SGX SDK introduced `MISCMASK` option in its Enclave Configuration File (for similar reasons). See e.g. https://github.com/intel/linux-sgx/blob/edfe42a517b3e4b1d81204c3cdef6da6cb35fefc/sdk/sign_tool/SignTool/sign_tool.cpp#L1300

Interestingly, Intel SGX SDK doesn't provide a backed-in solution for `ATTRIBUTEMASK` but instead provides two APIs:
- `sgx_seal_data()` has a hard-coded ATTRIBUTES.XFRM mask of all-zeros: https://github.com/intel/linux-sgx/blob/26c458905b72e66db7ac1feae04b43461ce1b76f/sdk/tseal/tSeal.cpp#L46-L54
- `sgx_seal_data_ex()` forces the user to specify `ATTRIBUTEMASK` explicitly: https://github.com/intel/linux-sgx/blob/26c458905b72e66db7ac1feae04b43461ce1b76f/sdk/tseal/tSeal.cpp#L68-L70

In Gramine, we cannot use a mask of all-zeros -- this is insecure (e.g., this would mean that one SGX enclave running with security-critical Intel MPX enabled and another SGX enclave running without Intel MPX could seal/unseal files of each other, worsening security guarantees).

So we need to specify *some reasonably secure mask*. The general problem here is that **there is no way to say which future Intel hardware features (reflected in ATTRIBUTES.XFRM) will be secure or insecure**. So Gramine ends up with an ever-changing `ATTRIBUTEMASK`, probably with each new generation of Intel CPUs.

So there seems to be the only reasonable solution -- to introduce a new `sgx.seal_key.xfrm_mask = [ hex ]` manifest option. It will default to whatever Gramine currently considers a secure mask (as of this writing, `0xfffffffffff9ff1b`). And the user may change this value if there is a need for backwards-compatibility.

### Example of the fix

So, to unseal the file in newer Gramine, the user will do:
```
$ echo 'sgx.seal_key.xfrm_mask = "ffffffffffffff1b"' >> manifest.template
$ SGX=1 gramine-test build

$ gramine-sgx sealed_file sealed_file_mrsigner.dat
READING OK
```

Unfortunately, specifying such a mask is still problematic: this mask implies that Intel AMX availability on the machine changes the Sealing Key. So there will be a backwards-compatibility with older Gramine versions, but now there will be incompatibility between AMX-enabled and AMX-disabled machines.

I don't see a good fix for this last problem. The only thing that can be done (it seems) is to periodically re-seal files with newer versions of Gramine...

dimakuv:
## Description of the changes 

This PR adds three new manifest options:

- `sgx.seal_key.flags_mask` for KEYREQUEST.ATTRIBUTESMASK.FLAGS
- `sgx.seal_key.xfrm_mask` for KEYREQUEST.ATTRIBUTESMASK.XFRM
- `sgx.seal_key.misc_mask` for KEYREQUEST.MISCMASK

These new manifest options allow the user to control how the SGX sealing key is generated. This flexibility is especially important for cases where files are sealed with one version of Gramine and unsealed with another version of Gramine (differing in default attribute masks).

For a very detailed explanation of this subtle issue, please check https://github.com/gramineproject/gramine/issues/274.

Fixes #274.

## How to test this PR? 

See #274. But here is a snippet that proves this PR works:
```
$ echo 'sgx.seal_key.xfrm_mask = "ffffffffffffff1b"' >> manifest.template
$ SGX=1 gramine-test build

$ gramine-sgx sealed_file sealed_file_mrsigner.dat
READING OK
```

I'm not sure how to test this in CI. I could create a couple new tests with manifest files containing compatible/incompatible `sgx.seal_key.xfrm_mask`, `sgx.seal_key.flags_mask`, `sgx.seal_key.misc_mask`, but I don't know if it is worth it.

UPDATE: The person who initially reported this issue tested this PR, and it works on their workload.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/275)

### DIFF
--- a/Documentation/manifest-syntax.rst
+++ b/Documentation/manifest-syntax.rst
@@ -532,6 +532,27 @@ ISV Product ID and SVN
 This syntax specifies the ISV Product ID and SVN to be added to the enclave
 signature.
 
+Attribute masks for SGX sealing key derivation
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+::
+
+    sgx.seal_key.flags_mask = "[8-byte hex value]"  (default: "0xffffffffffffffff")
+    sgx.seal_key.xfrm_mask  = "[8-byte hex value]"  (default: "0xfffffffffff9ff1b")
+    sgx.seal_key.misc_mask  = "[4-byte hex value]"  (default: "0xffffffff")
+
+This syntax specifies masks used to generate the SGX sealing key. These masks
+correspond to the following SGX ``KEYREQUEST`` struct fields:
+
+- ``flags_mask``: ``KEYREQUEST.ATTRIBUTESMASK.FLAGS``
+- ``xfrm_mask``: ``KEYREQUEST.ATTRIBUTESMASK.XFRM``
+- ``misc_mask``: ``KEYREQUEST.MISCMASK``
+
+Most users do *not* need to set these masks. Only advanced users with knowledge
+of SGX sealing should use these masks. In particular, these masks allow to
+specify a subset of enclave/machine attributes to be used in sealing key
+derivation. Moreover, these masks themselves are used in sealing key derivation.
+
 Allowed files
 ^^^^^^^^^^^^^
 
--- a/LibOS/shim/test/regression/attestation.manifest.template
+++ b/LibOS/shim/test/regression/attestation.manifest.template
@@ -21,6 +21,11 @@ sgx.nonpie_binary = true
 sgx.debug = true
 sgx.remote_attestation = true
 
+# below three entries are irrelevant, only for test purposes
+sgx.seal_key.flags_mask = "0xffffffffffffffff"
+sgx.seal_key.xfrm_mask  = "0xfffffffffff9ff1b"
+sgx.seal_key.misc_mask  = "0xffffffff"
+
 sgx.ra_client_spid = "{{ env.get('RA_CLIENT_SPID', '') }}"
 sgx.ra_client_linkable = {{ 'true' if env.get('RA_CLIENT_LINKABLE', '0') == '1' else 'false' }}
 
--- a/Pal/src/host/Linux-SGX/db_main.c
+++ b/Pal/src/host/Linux-SGX/db_main.c
@@ -858,6 +858,12 @@ noreturn void pal_linux_main(char* uptr_libpal_uri, size_t libpal_uri_len, char*
     }
     g_pal_internal_mem_size = extra_mem_size + PAL_INITIAL_MEM_SIZE;
 
+    /* seal-key material initialization must come before protected-files initialization */
+    if ((ret = init_seal_key_material()) < 0) {
+        log_error("Failed to initialize SGX sealing key material: %d", ret);
+        ocall_exit(1, /*is_exitgroup=*/true);
+    }
+
     if ((ret = init_file_check_policy()) < 0) {
         log_error("Failed to load the file check policy: %d", ret);
         ocall_exit(1, /*is_exitgroup=*/true);
--- a/Pal/src/host/Linux-SGX/enclave_framework.c
+++ b/Pal/src/host/Linux-SGX/enclave_framework.c
@@ -19,11 +19,31 @@
 #define LOCAL_ATTESTATION_TAG_PARENT_STR "GRAMINE_LOCAL_ATTESTATION_TAG_PARENT"
 #define LOCAL_ATTESTATION_TAG_CHILD_STR "GRAMINE_LOCAL_ATTESTATION_TAG_CHILD"
 
+static int register_file(const char* uri, const char* checksum_str, bool check_duplicates);
+
 void* g_enclave_base;
 void* g_enclave_top;
 bool g_allowed_files_warn = false;
 
-static int register_file(const char* uri, const char* checksum_str, bool check_duplicates);
+/*
+ * SGX's EGETKEY(SEAL_KEY) uses three masks as key-derivation material:
+ *   - KEYREQUEST.ATTRIBUTESMASK.FLAGS
+ *   - KEYREQUEST.ATTRIBUTESMASK.XFRM
+ *   - KEYREQUEST.MISCMASK
+ *
+ * These default masks may be replaced by user-defined ones (specified in the manifest file).
+ * Corresponding manifest keys are:
+ *   - `sgx.seal_key.flags_mask`
+ *   - `sgx.seal_key.xfrm_mask`
+ *   - `sgx.seal_key.misc_mask`
+ */
+static uint64_t g_seal_key_flags_mask = SGX_FLAGS_MASK_CONST;
+static uint64_t g_seal_key_xfrm_mask  = SGX_XFRM_MASK_CONST;
+static uint32_t g_seal_key_misc_mask  = SGX_MISCSELECT_MASK_CONST;
+
+static_assert(sizeof(g_seal_key_flags_mask) + sizeof(g_seal_key_xfrm_mask) ==
+                  sizeof(sgx_attributes_t), "wrong types");
+static_assert(sizeof(g_seal_key_misc_mask) == sizeof(sgx_misc_select_t), "wrong types");
 
 bool sgx_is_completely_within_enclave(const void* addr, size_t size) {
     if ((uintptr_t)addr > UINTPTR_MAX - size) {
@@ -209,9 +229,9 @@ int sgx_get_seal_key(uint16_t key_policy, sgx_key_128bit_t* out_seal_key) {
     memcpy(&key_request.config_svn, &g_pal_linuxsgx_state.enclave_info.config_svn,
            sizeof(sgx_config_svn_t));
 
-    key_request.attribute_mask.flags = SGX_FLAGS_MASK_CONST;
-    key_request.attribute_mask.xfrm  = SGX_XFRM_MASK_CONST;
-    key_request.misc_mask            = SGX_MISCSELECT_MASK_CONST;
+    key_request.attribute_mask.flags = g_seal_key_flags_mask;
+    key_request.attribute_mask.xfrm  = g_seal_key_xfrm_mask;
+    key_request.misc_mask            = g_seal_key_misc_mask;
 
     int ret = sgx_getkey(&key_request, out_seal_key);
     if (ret) {
@@ -976,6 +996,70 @@ int init_file_check_policy(void) {
     log_debug("File check policy: %s", file_check_policy_str);
     free(file_check_policy_str);
     return 0;
+}
+
+static int update_seal_key_mask(const char* mask_name, uint8_t* mask_ptr, size_t mask_size) {
+    int ret;
+
+    char* mask_str = NULL;
+    ret = toml_string_in(g_pal_public_state.manifest_root, mask_name, &mask_str);
+    if (ret < 0) {
+        log_error("Cannot parse '%s'", mask_name);
+        return -PAL_ERROR_INVAL;
+    }
+
+    if (!mask_str) {
+        /* no mask specified in the manifest, use the default */
+        return 0;
+    }
+
+    if (strlen(mask_str) != 2 + mask_size * 2) {
+        log_error("Malformed '%s' value in the manifest (wrong size)", mask_name);
+        ret = -PAL_ERROR_INVAL;
+        goto out;
+    }
+
+    if (mask_str[0] != '0' || (mask_str[1] != 'x' && mask_str[1] != 'X')) {
+        log_error("Malformed '%s' value in the manifest (must start with '0x')", mask_name);
+        ret = -PAL_ERROR_INVAL;
+        goto out;
+    }
+
+    memset(mask_ptr, 0, mask_size);
+
+    for (size_t i = 0; i < mask_size * 2; i++) {
+        int8_t val = hex2dec(mask_str[i + 2]); /* skip first two chars (the '0x' prefix) */
+        if (val < 0) {
+            log_error("Malformed '%s' value in the manifest (not a hex number)", mask_name);
+            ret = -PAL_ERROR_INVAL;
+            goto out;
+        }
+        uint8_t* mask_byte = &mask_ptr[mask_size - i / 2 - 1];
+        *mask_byte = *mask_byte * 16 + (uint8_t)val;
+    }
+
+    ret = 0;
+out:
+    free(mask_str);
+    return ret;
+}
+
+int init_seal_key_material(void) {
+    int ret;
+
+    /* below parsing of TOML strings assumes little-endianness (which is true for this SGX PAL) */
+    ret = update_seal_key_mask("sgx.seal_key.flags_mask", (uint8_t*)&g_seal_key_flags_mask,
+                               sizeof(g_seal_key_flags_mask));
+    if (ret < 0)
+        return ret;
+
+    ret = update_seal_key_mask("sgx.seal_key.xfrm_mask", (uint8_t*)&g_seal_key_xfrm_mask,
+                               sizeof(g_seal_key_xfrm_mask));
+    if (ret < 0)
+        return ret;
+
+    return update_seal_key_mask("sgx.seal_key.misc_mask", (uint8_t*)&g_seal_key_misc_mask,
+                                sizeof(g_seal_key_misc_mask));
 }
 
 int init_enclave(void) {
--- a/Pal/src/host/Linux-SGX/enclave_tf.h
+++ b/Pal/src/host/Linux-SGX/enclave_tf.h
@@ -28,8 +28,9 @@
 #include "enclave_tf_structs.h"
 #include "pal.h"
 
-int init_file_check_policy(void);
+int init_seal_key_material(void);
 
+int init_file_check_policy(void);
 int get_file_check_policy(void);
 
 /*!


boryspoplawski (changes requested):
> Reviewed 3 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/275)* status: 3 of 4 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[Documentation/manifest-syntax.rst, line 541 at r1](https://reviewable.io/reviews/gramineproject/gramine/275#-Mr-2rpiACHJpsrNrAlf:-Mr-2rpiACHJpsrNrAlg:b-ln3ood) ([raw file](https://github.com/gramineproject/gramine/blob/8c035f7ffbd34e3fd807de018802ebed9f6742c7/Documentation/manifest-syntax.rst#L541)):*
> ```reStructuredText
>     sgx.seal_key.flags_mask = "[8-byte hex value]"  (default: "ffffffffffffffff")
>     sgx.seal_key.xfrm_mask  = "[8-byte hex value]"  (default: "fffffffffff9ff1b")
>     sgx.seal_key.misc_mask  = "[4-byte hex value]"  (default: "ffffffff")
> ```

Uh, why these do not have the leading `0x`? Actually why they are not just numbers (instead of hex strings)?
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 28 at r1](https://reviewable.io/reviews/gramineproject/gramine/275#-Mr-0uwgA1aRL95AfdO3:-Mr-0uwgA1aRL95AfdO4:b-mwu3z6) ([raw file](https://github.com/gramineproject/gramine/blob/8c035f7ffbd34e3fd807de018802ebed9f6742c7/Pal/src/host/Linux-SGX/enclave_framework.c#L28)):*
> ```C
> bool g_allowed_files_warn = false;
> 
> /* SGX's EGETKEY(SEAL_KEY) uses three masks as key-derivation material:
> ```

NP: either add an empty line at the top of this comment or remove it from the end (former preferred by me).
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 33 at r1](https://reviewable.io/reviews/gramineproject/gramine/275#-Mr-16CEBpIadiKMv9en:-Mr-16CEBpIadiKMv9eo:b-x0deya) ([raw file](https://github.com/gramineproject/gramine/blob/8c035f7ffbd34e3fd807de018802ebed9f6742c7/Pal/src/host/Linux-SGX/enclave_framework.c#L33)):*
> ```C
>  *   - KEYREQUEST.MISCMASK
>  *
>  * These default masks may be replaced by used-defined ones, correspondingly:
> ```

Please add some information that the list below are manifest keys.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 40 at r1](https://reviewable.io/reviews/gramineproject/gramine/275#-Mr-1QxIAGRe-EUkeWMj:-Mr-1QxIAGRe-EUkeWMk:b-cfsjmn) ([raw file](https://github.com/gramineproject/gramine/blob/8c035f7ffbd34e3fd807de018802ebed9f6742c7/Pal/src/host/Linux-SGX/enclave_framework.c#L40)):*
> ```C
> uint64_t g_seal_key_flags_mask = SGX_FLAGS_MASK_CONST;
> uint64_t g_seal_key_xfrm_mask  = SGX_XFRM_MASK_CONST;
> uint32_t g_seal_key_misc_mask  = SGX_MISCSELECT_MASK_CONST;
> ```

Shouldn't these 3 be `static`? I don't see any usages outside of this file.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 40 at r1](https://reviewable.io/reviews/gramineproject/gramine/275#-Mr-2UCgB7eAzz3RfMZ4:-Mr-2UCgB7eAzz3RfMZ5:bamk4nm) ([raw file](https://github.com/gramineproject/gramine/blob/8c035f7ffbd34e3fd807de018802ebed9f6742c7/Pal/src/host/Linux-SGX/enclave_framework.c#L40)):*
> ```C
> uint64_t g_seal_key_flags_mask = SGX_FLAGS_MASK_CONST;
> uint64_t g_seal_key_xfrm_mask  = SGX_XFRM_MASK_CONST;
> uint32_t g_seal_key_misc_mask  = SGX_MISCSELECT_MASK_CONST;
> ```

We do have a special type for this `sgx_misc_select_t`.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 1032 at r1](https://reviewable.io/reviews/gramineproject/gramine/275#-Mr-3TqtEZEep-SULQeR:-Mr-3TqtEZEep-SULQeS:bp0htwy) ([raw file](https://github.com/gramineproject/gramine/blob/8c035f7ffbd34e3fd807de018802ebed9f6742c7/Pal/src/host/Linux-SGX/enclave_framework.c#L1032)):*
> ```C
>     free(mask_str);
>     return ret;
> }
> ```

I don't like this "advanced" parsing, I think a simple number would be enough. See the comment in documentation.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 1034 at r1](https://reviewable.io/reviews/gramineproject/gramine/275#-Mr-4Hqo-Ji2xdb9SSNE:-Mr-4Hqo-Ji2xdb9SSNF:b-vo6a9w) ([raw file](https://github.com/gramineproject/gramine/blob/8c035f7ffbd34e3fd807de018802ebed9f6742c7/Pal/src/host/Linux-SGX/enclave_framework.c#L1034)):*
> ```C
> }
> 
> int init_seal_key(void) {
> ```

Hm, this initializes seal key material, not the key itself (if there is a difference)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/275)* status: 3 of 4 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski)

___
*[Documentation/manifest-syntax.rst, line 541 at r1](https://reviewable.io/reviews/gramineproject/gramine/275#-Mr-2rpiACHJpsrNrAlf:-Mr1VAFO8go8mZjrvlPX:b-ca9sfk) ([raw file](https://github.com/gramineproject/gramine/blob/8c035f7ffbd34e3fd807de018802ebed9f6742c7/Documentation/manifest-syntax.rst#L541)):*

Uh, why these do not have the leading `0x`? Actually why they are not just numbers (instead of hex strings)?
</blockquote></details>

I don't like them to be represented as numbers. I think masks should have a very specific format (they must be exactly 32/64 bits in size), and allowing to specify something like `sgx.seal_key.flags_mask = 123` will be abused by users in any possible way.

Having a strict 4/8-byte hex string seems like a natural fit for the mask, and some guarantee that users won't abuse this without understanding the consequences.

I can add the prefix `0x`, no problem. But I don't want to have it as a simple TOML number. So what would you recommend?
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 28 at r1](https://reviewable.io/reviews/gramineproject/gramine/275#-Mr-0uwgA1aRL95AfdO3:-Mr1Ohaa5fmK37R8xO0o:b-gu810e) ([raw file](https://github.com/gramineproject/gramine/blob/8c035f7ffbd34e3fd807de018802ebed9f6742c7/Pal/src/host/Linux-SGX/enclave_framework.c#L28)):*

NP: either add an empty line at the top of this comment or remove it from the end (former preferred by me).
</blockquote></details>

Done. Always forget about this.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 33 at r1](https://reviewable.io/reviews/gramineproject/gramine/275#-Mr-16CEBpIadiKMv9en:-Mr1PIqx1kCbsrcwWQSy:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/8c035f7ffbd34e3fd807de018802ebed9f6742c7/Pal/src/host/Linux-SGX/enclave_framework.c#L33)):*

Please add some information that the list below are manifest keys.
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 40 at r1](https://reviewable.io/reviews/gramineproject/gramine/275#-Mr-1QxIAGRe-EUkeWMj:-Mr1PRps27F01Cz9N_HU:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/8c035f7ffbd34e3fd807de018802ebed9f6742c7/Pal/src/host/Linux-SGX/enclave_framework.c#L40)):*

Shouldn't these 3 be `static`? I don't see any usages outside of this file.
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 40 at r1](https://reviewable.io/reviews/gramineproject/gramine/275#-Mr-2UCgB7eAzz3RfMZ4:-Mr1UOpB9nJ2Ai4qUdax:b-9i1akp) ([raw file](https://github.com/gramineproject/gramine/blob/8c035f7ffbd34e3fd807de018802ebed9f6742c7/Pal/src/host/Linux-SGX/enclave_framework.c#L40)):*

We do have a special type for this `sgx_misc_select_t`.
</blockquote></details>

I don't like those types too much. We have a type for MISCSELECT and thus MISCMASK, but we don't have special types for FLAGS and XFRM. So I wanted to be uniform and use standard types.

But I added two static asserts to make it explicit.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 1032 at r1](https://reviewable.io/reviews/gramineproject/gramine/275#-Mr-3TqtEZEep-SULQeR:-Mr1W1j-3Z1QIV4c4-dD:b-volb2) ([raw file](https://github.com/gramineproject/gramine/blob/8c035f7ffbd34e3fd807de018802ebed9f6742c7/Pal/src/host/Linux-SGX/enclave_framework.c#L1032)):*

I don't like this "advanced" parsing, I think a simple number would be enough. See the comment in documentation.
</blockquote></details>

See my reply there.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 1034 at r1](https://reviewable.io/reviews/gramineproject/gramine/275#-Mr-4Hqo-Ji2xdb9SSNE:-Mr1Uj4L408yiZ1_fsyu:b-1ijo5) ([raw file](https://github.com/gramineproject/gramine/blob/8c035f7ffbd34e3fd807de018802ebed9f6742c7/Pal/src/host/Linux-SGX/enclave_framework.c#L1034)):*

Hm, this initializes seal key material, not the key itself (if there is a difference)
</blockquote></details>

Done. Fair enough, change the name.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/275)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[Documentation/manifest-syntax.rst, line 541 at r1](https://reviewable.io/reviews/gramineproject/gramine/275#-Mr-2rpiACHJpsrNrAlf:-Mr2dGdi2BZVccTZqDS7:bvzx4v2) ([raw file](https://github.com/gramineproject/gramine/blob/8c035f7ffbd34e3fd807de018802ebed9f6742c7/Documentation/manifest-syntax.rst#L541)):*

I don't like them to be represented as numbers. I think masks should have a very specific format (they must be exactly 32/64 bits in size), and allowing to specify something like `sgx.seal_key.flags_mask = 123` will be abused by users in any possible way.

Having a strict 4/8-byte hex string seems like a natural fit for the mask, and some guarantee that users won't abuse this without understanding the consequences.

I can add the prefix `0x`, no problem. But I don't want to have it as a simple TOML number. So what would you recommend?
</blockquote></details>

The main problem with the current version is that I have no idea what the endianness is. Is the first nibble pair (byte) MSB (like it was a hex encoded number) or is it LSB (like it was hex encoded string/bytes array)?
But yeah, I see some value in the fact that the size must match (e.g. you cannot remove a leading `f` accidentally).

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/275)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[Documentation/manifest-syntax.rst, line 541 at r1](https://reviewable.io/reviews/gramineproject/gramine/275#-Mr-2rpiACHJpsrNrAlf:-Mr6zrlH7RyLbjrgpYE5:bimshx5) ([raw file](https://github.com/gramineproject/gramine/blob/8c035f7ffbd34e3fd807de018802ebed9f6742c7/Documentation/manifest-syntax.rst#L541)):*

The main problem with the current version is that I have no idea what the endianness is. Is the first nibble pair (byte) MSB (like it was a hex encoded number) or is it LSB (like it was hex encoded string/bytes array)?
But yeah, I see some value in the fact that the size must match (e.g. you cannot remove a leading `f` accidentally).
</blockquote></details>

This string represents a number in LSB format.

So e.g. `fffffffffff9ff1b` is `1111111111111111111111111111111111111111111110011111111100011011` in binary.

Adding `0x` prefix would suffice for you? In addition, I can add the remark in the text about LSB.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/275)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[Documentation/manifest-syntax.rst, line 541 at r1](https://reviewable.io/reviews/gramineproject/gramine/275#-Mr-2rpiACHJpsrNrAlf:-Mr7ApgQ6gWAx2ovk1Jw:b9yzfsd) ([raw file](https://github.com/gramineproject/gramine/blob/8c035f7ffbd34e3fd807de018802ebed9f6742c7/Documentation/manifest-syntax.rst#L541)):*

This string represents a number in LSB format.

So e.g. `fffffffffff9ff1b` is `1111111111111111111111111111111111111111111110011111111100011011` in binary.

Adding `0x` prefix would suffice for you? In addition, I can add the remark in the text about LSB.
</blockquote></details>

I know, I was just pointing out it's not clear just from the manifest documentation.

Yeah, `0x` should be enough.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/275)* status: 2 of 5 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[Documentation/manifest-syntax.rst, line 541 at r1](https://reviewable.io/reviews/gramineproject/gramine/275#-Mr-2rpiACHJpsrNrAlf:-Mr7IaKJ1FhUS9cchV4E:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/8c035f7ffbd34e3fd807de018802ebed9f6742c7/Documentation/manifest-syntax.rst#L541)):*

I know, I was just pointing out it's not clear just from the manifest documentation.

Yeah, `0x` should be enough.
</blockquote></details>

Done.

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/275)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/275)* status: :shipit: complete! all files reviewed, all discussions resolved

mkow (changes requested):
> Reviewed 1 of 3 files at r2, 1 of 3 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/275)* status: all files reviewed, 1 unresolved discussion (waiting on @dimakuv)

___
*[Documentation/manifest-syntax.rst, line 542 at r4](https://reviewable.io/reviews/gramineproject/gramine/275#-MrXDUwWBoUNHpVo9DGE:-MrXDUwWBoUNHpVo9DGF:b-oafylw) ([raw file](https://github.com/gramineproject/gramine/blob/a163844ab42285fe624d57d44a53973b685ce4fa/Documentation/manifest-syntax.rst#L542)):*
> ```reStructuredText
>     sgx.seal_key.flags_mask = "[8-byte hex value]"  (default: "0xffffffffffffffff")
>     sgx.seal_key.xfrm_mask  = "[8-byte hex value]"  (default: "0xfffffffffff9ff1b")
>     sgx.seal_key.misc_mask  = "[4-byte hex value]"  (default: "0xffffffff")
> ```

If these are valid hex numbers, then why are we passing them as strings? I guess because tomlc99 doesn't handle unsigned 64-bit numbers well?
If ints are actually possible to use then it would make `update_seal_key_mask()` much simpler.
___
*[Pal/src/host/Linux-SGX/db_main.c, line 861 at r4](https://reviewable.io/reviews/gramineproject/gramine/275#-MrXDhoi62tDJcVv0vbL:-MrXDhoi62tDJcVv0vbM:b-7fnmtc) ([raw file](https://github.com/gramineproject/gramine/blob/a163844ab42285fe624d57d44a53973b685ce4fa/Pal/src/host/Linux-SGX/db_main.c#L861)):*
> ```C
>     g_pal_internal_mem_size = extra_mem_size + PAL_INITIAL_MEM_SIZE;
> 
>     /* seal-key material initialization must come before protected-files initialization */
> ```

We need some better way to assert dependencies in initialization order, but that's a bigger problem for another time...
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 45 at r4](https://reviewable.io/reviews/gramineproject/gramine/275#-MrXFHp364xPYLf9CD2b:-MrXFHp364xPYLf9CD2c:bktw28i) ([raw file](https://github.com/gramineproject/gramine/blob/a163844ab42285fe624d57d44a53973b685ce4fa/Pal/src/host/Linux-SGX/enclave_framework.c#L45)):*
> ```C
> 
> static_assert(sizeof(g_seal_key_flags_mask) + sizeof(g_seal_key_xfrm_mask) ==
>         sizeof(sgx_attributes_t), "wrong types");
> ```

This indentation seems wrong
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 1051 at r4](https://reviewable.io/reviews/gramineproject/gramine/275#-MrXGYVNBSoJFJGuo3LV:-MrXGYVNBSoJFJGuo3LW:bsowowm) ([raw file](https://github.com/gramineproject/gramine/blob/a163844ab42285fe624d57d44a53973b685ce4fa/Pal/src/host/Linux-SGX/enclave_framework.c#L1051)):*
> ```C
> 
>     ret = update_seal_key_mask("sgx.seal_key.flags_mask", (uint8_t*)&g_seal_key_flags_mask,
>                                sizeof(g_seal_key_flags_mask));
> ```

This won't work on big-endian machines, because `update_seal_key_mask()` always writes the output in little-endian. We can probably silently assume little-endianness here because SGX works only on little endian, but it's quite ugly.

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/275)* status: all files reviewed, 1 unresolved discussion (waiting on @dimakuv)

___
*[Documentation/manifest-syntax.rst, line 542 at r4](https://reviewable.io/reviews/gramineproject/gramine/275#-MrXDUwWBoUNHpVo9DGE:-MrXMl9sEs1j0PmXBgda:b1wfjrq) ([raw file](https://github.com/gramineproject/gramine/blob/a163844ab42285fe624d57d44a53973b685ce4fa/Documentation/manifest-syntax.rst#L542)):*

If these are valid hex numbers, then why are we passing them as strings? I guess because tomlc99 doesn't handle unsigned 64-bit numbers well?
If ints are actually possible to use then it would make `update_seal_key_mask()` much simpler.
</blockquote></details>

See the closed conversation (here and near parsing), we already talked about this
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 1051 at r4](https://reviewable.io/reviews/gramineproject/gramine/275#-MrXGYVNBSoJFJGuo3LV:-MrXMt2-41bWGiuShF0y:b-ehv57v) ([raw file](https://github.com/gramineproject/gramine/blob/a163844ab42285fe624d57d44a53973b685ce4fa/Pal/src/host/Linux-SGX/enclave_framework.c#L1051)):*

This won't work on big-endian machines, because `update_seal_key_mask()` always writes the output in little-endian. We can probably silently assume little-endianness here because SGX works only on little endian, but it's quite ugly.
</blockquote></details>

We assume that we are running on x64 everywhere in SGX code (which is a valid assumption, I believe :smiley: ), which is always little endian.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/275)* status: all files reviewed, 1 unresolved discussion (waiting on @dimakuv)

___
*[Documentation/manifest-syntax.rst, line 542 at r4](https://reviewable.io/reviews/gramineproject/gramine/275#-MrXDUwWBoUNHpVo9DGE:-MrXP9kYA4ymPqSFhRl8:b-zavtxu) ([raw file](https://github.com/gramineproject/gramine/blob/a163844ab42285fe624d57d44a53973b685ce4fa/Documentation/manifest-syntax.rst#L542)):*

See the closed conversation (here and near parsing), we already talked about this
</blockquote></details>

Ok, I see, although I don't know why dropping the double quotes would change anything w.r.t. readability/ambiguity of this.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 1051 at r4](https://reviewable.io/reviews/gramineproject/gramine/275#-MrXGYVNBSoJFJGuo3LV:-MrXPTnz3XZjm2KP6c0_:b-7e4uag) ([raw file](https://github.com/gramineproject/gramine/blob/a163844ab42285fe624d57d44a53973b685ce4fa/Pal/src/host/Linux-SGX/enclave_framework.c#L1051)):*

We assume that we are running on x64 everywhere in SGX code (which is a valid assumption, I believe :smiley: ), which is always little endian.
</blockquote></details>

yeah, but I don't like that this is assumed so silently here

boryspoplawski (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/275)* status: all files reviewed, 1 unresolved discussion (waiting on @dimakuv)

___
*[Documentation/manifest-syntax.rst, line 542 at r4](https://reviewable.io/reviews/gramineproject/gramine/275#-MrXDUwWBoUNHpVo9DGE:-MrXQY9j-3e3Ne8qe1Oh:b-l19un1) ([raw file](https://github.com/gramineproject/gramine/blob/a163844ab42285fe624d57d44a53973b685ce4fa/Documentation/manifest-syntax.rst#L542)):*

Ok, I see, although I don't know why dropping the double quotes would change anything w.r.t. readability/ambiguity of this.
</blockquote></details>

There is a check that the size of this string is exactly what's expected, so if you e.g. drop one `f` then it will complain loudly. At least I think that was the reasoning of @dimakuv for this

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/275)* status: 4 of 5 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[Documentation/manifest-syntax.rst, line 542 at r4](https://reviewable.io/reviews/gramineproject/gramine/275#-MrXDUwWBoUNHpVo9DGE:-MrXcnhlA2IZ75gVK2QG:b-py3w1r) ([raw file](https://github.com/gramineproject/gramine/blob/a163844ab42285fe624d57d44a53973b685ce4fa/Documentation/manifest-syntax.rst#L542)):*
Yes, Borys is correct, that was my reasoning. Citing myself from a previous (resolved) discussion:

> I don't like them to be represented as numbers. I think masks should have a very specific format (they must be exactly 32/64 bits in size), and allowing to specify something like `sgx.seal_key.flags_mask = 123` will be abused by users in any possible way.

Basically, I want to force users to specify all bits explicitly. If users don't (the string is too short), then my parser will fail loudly.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 45 at r4](https://reviewable.io/reviews/gramineproject/gramine/275#-MrXFHp364xPYLf9CD2b:-MrXdXV92htCW1lsthY1:bzabb07) ([raw file](https://github.com/gramineproject/gramine/blob/a163844ab42285fe624d57d44a53973b685ce4fa/Pal/src/host/Linux-SGX/enclave_framework.c#L45)):*

This indentation seems wrong
</blockquote></details>

Done. I wasn't sure how to indent it better. I think now is good enough.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 1051 at r4](https://reviewable.io/reviews/gramineproject/gramine/275#-MrXGYVNBSoJFJGuo3LV:-MrXdyXn43hQBIbFzgs_:bs6xkwd) ([raw file](https://github.com/gramineproject/gramine/blob/a163844ab42285fe624d57d44a53973b685ce4fa/Pal/src/host/Linux-SGX/enclave_framework.c#L1051)):*

yeah, but I don't like that this is assumed so silently here
</blockquote></details>

Done. I added a comment (in case someone will want to copy it to some big-endian machine).

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/275)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

___
*[Documentation/manifest-syntax.rst, line 542 at r4](https://reviewable.io/reviews/gramineproject/gramine/275#-MrXDUwWBoUNHpVo9DGE:-MrXf4fm0lP8XL91W_h1:b-nec2kw) ([raw file](https://github.com/gramineproject/gramine/blob/a163844ab42285fe624d57d44a53973b685ce4fa/Documentation/manifest-syntax.rst#L542)):*

Yes, Borys is correct, that was my reasoning. Citing myself from a previous (resolved) discussion:

> I don't like them to be represented as numbers. I think masks should have a very specific format (they must be exactly 32/64 bits in size), and allowing to specify something like `sgx.seal_key.flags_mask = 123` will be abused by users in any possible way.

Basically, I want to force users to specify all bits explicitly. If users don't (the string is too short), then my parser will fail loudly.
</blockquote></details>

Ok, makes sense.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 45 at r4](https://reviewable.io/reviews/gramineproject/gramine/275#-MrXFHp364xPYLf9CD2b:-MrXfDzS7GN323IRCbgK:b-a8jcui) ([raw file](https://github.com/gramineproject/gramine/blob/a163844ab42285fe624d57d44a53973b685ce4fa/Pal/src/host/Linux-SGX/enclave_framework.c#L45)):*

Done. I wasn't sure how to indent it better. I think now is good enough.
</blockquote></details>

Yup, now it's good :)

boryspoplawski (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/275)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/275)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #278, opened 2021-12-13
[Pal] Refactor and rename global structures

mkow:
## Description of the changes 

The new names should better reflect what these structures really are:
- `pal_linux_state` - PAL/Linux private state
- `pal_linuxsgx_state` - PAL/Linux-SGX private state
- `pal_common_state` - common state between all PALs, private
- `pal_public_state` - common state between all PALs, but public (shared with LibOS)

I made `private` and `uncommon` attributes implicit (i.e. not added to the names when they apply, instead adding only negations), to not make the names even longer, but if you have better ideas for consistent naming then I'm open for changing them. And `pal_public_state` already breaks this (should be `pal_common_public_state`), but that would be really long :/

Renamed/removed:
- `PAL_CONTROL` - No idea where this name came from, but it's actually a part of the PAL state which is visible by LibOS, so it was renamed to `pal_public_state`.
- `pal_sec` - Another riddle-name :) Partially un-globalled, mostly migrated to `pal_linuxsgx_state`, will be finally deleted in the next PR.
- `pal_internal_state` - Renamed to `pal_common_state`.

There's also a question whether something should be a global variable in PAL or a member of one of these structs. Technically all variables could be moved to "direct" globals or all moved to those global structs, but I think it's better to split them, and the question is how exactly. I'd put in these structs +/- only the state which sounds like global to the whole PAL, not just to a single submodule. But it's quite subjective and I'm not very convinced about this approach, so I'd like to hear your ideas (and whatever we choose, this is for future PRs, not this one).

This PR also makes passing of the initial state to the start ecall more explicit (previously `pal_sec` was partially used for this and partially for storing other global state).

The next step will be moving topology information to a separate struct and getting rid of `pal_sec`.

## How to test this PR? 

CI, no functional changes intended.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/278)

### DIFF
--- a/Documentation/pal/host-abi.rst
+++ b/Documentation/pal/host-abi.rst
@@ -79,21 +79,20 @@ Basic types
    :project: pal
    :members:
 
-Gramine control block
-^^^^^^^^^^^^^^^^^^^^^^
+PAL public state
+^^^^^^^^^^^^^^^^
 
-The control block in Gramine is a structure that provides static immutable
-information about the current process and its host. The address of the control
-block is resolved via :func:`DkGetPalControl()` and can be memorized in a global
-variable for ease of use.
+All PALs in Gramine expose a structure that provides static immutable
+information about the current process and its host. The address of the
+structure can be retrieved via :func:`DkGetPalPublicState()` and can be
+memorized in a global variable for ease of use.
 
-The fields of the Gramine control block are defined as follows:
-
-.. doxygentypedef:: PAL_CONTROL
-   :project: pal
-.. doxygenstruct:: PAL_CONTROL_
+.. doxygenstruct:: pal_public_state
    :project: pal
    :members:
+
+PAL public state - topology information
+"""""""""""""""""""""""""""""""""""""""
 
 .. doxygentypedef:: PAL_CPU_INFO
    :project: pal
@@ -113,7 +112,7 @@ The fields of the Gramine control block are defined as follows:
    :project: pal
    :members:
 
-Pal APIs
+PAL APIs
 --------
 
 The PAL APIs contain a |~| number of functions that can be called from the
@@ -331,7 +330,7 @@ and to obtain an attestation report and quote.
 .. doxygenfunction:: DkDebugLog
    :project: pal
 
-.. doxygenfunction:: DkGetPalControl
+.. doxygenfunction:: DkGetPalPublicState
    :project: pal
 
 .. doxygenfunction:: DkSystemTimeQuery
--- a/LibOS/shim/include/shim_internal.h
+++ b/LibOS/shim/include/shim_internal.h
@@ -25,7 +25,7 @@ noreturn void* shim_init(int argc, const char** argv, const char** envp);
 
 extern int g_log_level;
 
-extern const PAL_CONTROL* g_pal_control;
+extern const struct pal_public_state* g_pal_public_state;
 
 // TODO(mkow): We should make it cross-object-inlinable, ideally by enabling LTO, less ideally by
 // pasting it here and making `inline`, but our current linker scripts prevent both.
@@ -310,7 +310,7 @@ static inline int64_t __ref_dec(REFTYPE* ref) {
 #define __alloca __builtin_alloca
 #endif
 
-#define ALLOC_ALIGNMENT         (g_pal_control->alloc_align)
+#define ALLOC_ALIGNMENT         (g_pal_public_state->alloc_align)
 #define IS_ALLOC_ALIGNED(x)     IS_ALIGNED_POW2(x, ALLOC_ALIGNMENT)
 #define IS_ALLOC_ALIGNED_PTR(x) IS_ALIGNED_PTR_POW2(x, ALLOC_ALIGNMENT)
 #define ALLOC_ALIGN_DOWN(x)     ALIGN_DOWN_POW2(x, ALLOC_ALIGNMENT)
--- a/LibOS/shim/include/shim_vma.h
+++ b/LibOS/shim/include/shim_vma.h
@@ -94,11 +94,11 @@ int bkeep_mmap_any_in_range(void* bottom_addr, void* top_addr, size_t length, in
                             void** ret_val_ptr);
 
 /* Shorthand for `bkeep_mmap_any_in_range` with the range
- * [`g_pal_control->user_address.start`, `g_pal_control->user_address.end`). */
+ * [`g_pal_public_state->user_address.start`, `g_pal_public_state->user_address.end`). */
 int bkeep_mmap_any(size_t length, int prot, int flags, struct shim_handle* file, uint64_t offset,
                    const char* comment, void** ret_val_ptr);
 
-/* First tries to bookkeep in the range [`g_pal_control->user_address.start`, `aslr_addr_top`)
+/* First tries to bookkeep in the range [`g_pal_public_state->user_address.start`, `aslr_addr_top`)
  * and if it fails calls `bkeep_mmap_any`. `aslr_addr_top` is a value randomized on each program
  * run. */
 int bkeep_mmap_any_aslr(size_t length, int prot, int flags, struct shim_handle* file,
--- a/LibOS/shim/src/bookkeep/shim_thread.c
+++ b/LibOS/shim/src/bookkeep/shim_thread.c
@@ -212,7 +212,7 @@ static int init_main_thread(void) {
         return ret;
     }
 
-    cur_thread->pal_handle = g_pal_control->first_thread;
+    cur_thread->pal_handle = g_pal_public_state->first_thread;
 
     set_cur_thread(cur_thread);
     add_thread(cur_thread);
@@ -356,7 +356,7 @@ void put_thread(struct shim_thread* thread) {
 
         free(thread->groups_info.groups);
 
-        if (thread->pal_handle && thread->pal_handle != g_pal_control->first_thread)
+        if (thread->pal_handle && thread->pal_handle != g_pal_public_state->first_thread)
             DkObjectClose(thread->pal_handle);
 
         if (thread->handle_map) {
@@ -659,7 +659,7 @@ BEGIN_RS_FUNC(thread) {
     assert(tcb->context.regs);
     set_tls(tcb->context.tls);
 
-    thread->pal_handle = g_pal_control->first_thread;
+    thread->pal_handle = g_pal_public_state->first_thread;
 
     set_cur_thread(thread);
     log_setprefix(thread->shim_tcb);
--- a/LibOS/shim/src/bookkeep/shim_vma.c
+++ b/LibOS/shim/src/bookkeep/shim_vma.c
@@ -524,7 +524,7 @@ static int _bkeep_initial_vma(struct shim_vma* new_vma) {
 static void* g_aslr_addr_top = NULL;
 
 int init_vma(void) {
-    struct shim_vma init_vmas[2 + g_pal_control->preloaded_ranges_cnt];
+    struct shim_vma init_vmas[2 + g_pal_public_state->preloaded_ranges_cnt];
 
     init_vmas[0].begin = 0; // vma for creation of memory manager
 
@@ -536,14 +536,14 @@ int init_vma(void) {
     init_vmas[1].offset = 0;
     copy_comment(&init_vmas[1], "LibOS");
 
-    for (size_t i = 0; i < g_pal_control->preloaded_ranges_cnt; i++) {
-        init_vmas[2 + i].begin  = ALLOC_ALIGN_DOWN(g_pal_control->preloaded_ranges[i].start);
-        init_vmas[2 + i].end    = ALLOC_ALIGN_UP(g_pal_control->preloaded_ranges[i].end);
+    for (size_t i = 0; i < g_pal_public_state->preloaded_ranges_cnt; i++) {
+        init_vmas[2 + i].begin  = ALLOC_ALIGN_DOWN(g_pal_public_state->preloaded_ranges[i].start);
+        init_vmas[2 + i].end    = ALLOC_ALIGN_UP(g_pal_public_state->preloaded_ranges[i].end);
         init_vmas[2 + i].prot   = PROT_NONE;
         init_vmas[2 + i].flags  = MAP_PRIVATE | MAP_ANONYMOUS | VMA_INTERNAL;
         init_vmas[2 + i].file   = NULL;
         init_vmas[2 + i].offset = 0;
-        copy_comment(&init_vmas[2 + i], g_pal_control->preloaded_ranges[i].comment);
+        copy_comment(&init_vmas[2 + i], g_pal_public_state->preloaded_ranges[i].comment);
     }
 
     spinlock_lock(&vma_tree_lock);
@@ -581,12 +581,12 @@ int init_vma(void) {
         return ret;
     }
 
-    g_aslr_addr_top = g_pal_control->user_address.end;
+    g_aslr_addr_top = g_pal_public_state->user_address.end;
 
-    if (!g_pal_control->disable_aslr) {
+    if (!g_pal_public_state->disable_aslr) {
         /* Inspired by: https://elixir.bootlin.com/linux/v5.6.3/source/arch/x86/mm/mmap.c#L80 */
         size_t gap_max_size =
-            (g_pal_control->user_address.end - g_pal_control->user_address.start) / 6 * 5;
+            (g_pal_public_state->user_address.end - g_pal_public_state->user_address.start) / 6 * 5;
         /* We do address space randomization only if we have at least ASLR_BITS to randomize. */
         if (gap_max_size / ALLOC_ALIGNMENT >= (1ul << ASLR_BITS)) {
             size_t gap = 0;
@@ -1030,16 +1030,16 @@ out:
 
 int bkeep_mmap_any(size_t length, int prot, int flags, struct shim_handle* file, uint64_t offset,
                    const char* comment, void** ret_val_ptr) {
-    return bkeep_mmap_any_in_range(g_pal_control->user_address.start,
-                                   g_pal_control->user_address.end,
+    return bkeep_mmap_any_in_range(g_pal_public_state->user_address.start,
+                                   g_pal_public_state->user_address.end,
                                    length, prot, flags, file, offset, comment, ret_val_ptr);
 }
 
 int bkeep_mmap_any_aslr(size_t length, int prot, int flags, struct shim_handle* file,
                         uint64_t offset, const char* comment, void** ret_val_ptr) {
     int ret;
-    ret = bkeep_mmap_any_in_range(g_pal_control->user_address.start, g_aslr_addr_top, length, prot,
-                                  flags, file, offset, comment, ret_val_ptr);
+    ret = bkeep_mmap_any_in_range(g_pal_public_state->user_address.start, g_aslr_addr_top, length,
+                                  prot, flags, file, offset, comment, ret_val_ptr);
     if (ret >= 0) {
         return ret;
     }
--- a/LibOS/shim/src/fs/dev/attestation.c
+++ b/LibOS/shim/src/fs/dev/attestation.c
@@ -284,7 +284,7 @@ static int pfkey_save(struct shim_dentry* dent, const char* data, size_t size) {
 }
 
 int init_attestation(struct pseudo_node* dev) {
-    if (strcmp(g_pal_control->host_type, "Linux-SGX")) {
+    if (strcmp(g_pal_public_state->host_type, "Linux-SGX")) {
         log_debug("host is not Linux-SGX, skipping /dev/attestation setup");
         return 0;
     }
--- a/LibOS/shim/src/fs/proc/info.c
+++ b/LibOS/shim/src/fs/proc/info.c
@@ -23,7 +23,7 @@ int proc_meminfo_load(struct shim_dentry* dent, char** out_data, size_t* out_siz
     } meminfo[] = {
         {
             "MemTotal:      %8lu kB\n",
-            g_pal_control->mem_info.mem_total / 1024,
+            g_pal_public_state->mem_info.mem_total / 1024,
         },
         {
             "MemFree:       %8lu kB\n",
@@ -113,19 +113,19 @@ int proc_cpuinfo_load(struct shim_dentry* dent, char** out_data, size_t* out_siz
         size += ret;                                                  \
     } while (0)
 
-    for (size_t n = 0; n < g_pal_control->cpu_info.online_logical_cores; n++) {
+    for (size_t n = 0; n < g_pal_public_state->cpu_info.online_logical_cores; n++) {
         /* Below strings must match exactly the strings retrieved from /proc/cpuinfo
          * (see Linux's arch/x86/kernel/cpu/proc.c) */
         ADD_INFO("processor\t: %lu\n", n);
-        ADD_INFO("vendor_id\t: %s\n", g_pal_control->cpu_info.cpu_vendor);
-        ADD_INFO("cpu family\t: %lu\n", g_pal_control->cpu_info.cpu_family);
-        ADD_INFO("model\t\t: %lu\n", g_pal_control->cpu_info.cpu_model);
-        ADD_INFO("model name\t: %s\n", g_pal_control->cpu_info.cpu_brand);
-        ADD_INFO("stepping\t: %lu\n", g_pal_control->cpu_info.cpu_stepping);
-        ADD_INFO("physical id\t: %d\n", g_pal_control->cpu_info.cpu_socket[n]);
+        ADD_INFO("vendor_id\t: %s\n", g_pal_public_state->cpu_info.cpu_vendor);
+        ADD_INFO("cpu family\t: %lu\n", g_pal_public_state->cpu_info.cpu_family);
+        ADD_INFO("model\t\t: %lu\n", g_pal_public_state->cpu_info.cpu_model);
+        ADD_INFO("model name\t: %s\n", g_pal_public_state->cpu_info.cpu_brand);
+        ADD_INFO("stepping\t: %lu\n", g_pal_public_state->cpu_info.cpu_stepping);
+        ADD_INFO("physical id\t: %d\n", g_pal_public_state->cpu_info.cpu_socket[n]);
         ADD_INFO("core id\t\t: %lu\n", n);
-        ADD_INFO("cpu cores\t: %lu\n", g_pal_control->cpu_info.physical_cores_per_socket);
-        double bogomips = g_pal_control->cpu_info.cpu_bogomips;
+        ADD_INFO("cpu cores\t: %lu\n", g_pal_public_state->cpu_info.physical_cores_per_socket);
+        double bogomips = g_pal_public_state->cpu_info.cpu_bogomips;
         // Apparently Gramine snprintf cannot into floats.
         ADD_INFO("bogomips\t: %lu.%02lu\n", (unsigned long)bogomips,
                  (unsigned long)(bogomips * 100.0 + 0.5) % 100);
--- a/LibOS/shim/src/fs/shim_dcache.c
+++ b/LibOS/shim/src/fs/shim_dcache.c
@@ -64,7 +64,7 @@ int init_dcache(void) {
 
     dentry_mgr = create_mem_mgr(init_align_up(DCACHE_MGR_ALLOC));
 
-    if (g_pal_control->parent_process) {
+    if (g_pal_public_state->parent_process) {
         /* In a child process, `g_dentry_root` will be restored from a checkpoint. */
         return 0;
     }
--- a/LibOS/shim/src/fs/shim_fs.c
+++ b/LibOS/shim/src/fs/shim_fs.c
@@ -151,7 +151,7 @@ static int __mount_sys(void) {
         return ret;
     }
 
-    if (g_pal_control->enable_sysfs_topology) {
+    if (g_pal_public_state->enable_sysfs_topology) {
         log_debug("Mounting special sys filesystem: /sys");
         if ((ret = mount_fs("pseudo", "sys", "/sys")) < 0) {
             log_error("Mounting sys filesystem failed (%d)", ret);
--- a/LibOS/shim/src/fs/sys/cache_info.c
+++ b/LibOS/shim/src/fs/sys/cache_info.c
@@ -26,7 +26,7 @@ int sys_cache_load(struct shim_dentry* dent, char** out_data, size_t* out_size) 
         return ret;
 
     const char* name = dent->name;
-    PAL_CORE_CACHE_INFO* cache = &g_pal_control->topo_info.core_topology[cpu_num].cache[cache_num];
+    PAL_CORE_CACHE_INFO* cache = &g_pal_public_state->topo_info.core_topology[cpu_num].cache[cache_num];
     const char* str;
     if (strcmp(name, "shared_cpu_map") == 0) {
         str = cache->shared_cpu_map;
--- a/LibOS/shim/src/fs/sys/cpu_info.c
+++ b/LibOS/shim/src/fs/sys/cpu_info.c
@@ -17,9 +17,9 @@ int sys_cpu_general_load(struct shim_dentry* dent, char** out_data, size_t* out_
     const char* str;
 
     if (strcmp(name, "online") == 0) {
-        str = g_pal_control->topo_info.online_logical_cores;
+        str = g_pal_public_state->topo_info.online_logical_cores;
     } else if (strcmp(name, "possible") == 0) {
-        str = g_pal_control->topo_info.possible_logical_cores;
+        str = g_pal_public_state->topo_info.possible_logical_cores;
     } else {
         log_debug("unrecognized file: %s", name);
         return -ENOENT;
@@ -36,7 +36,7 @@ int sys_cpu_load(struct shim_dentry* dent, char** out_data, size_t* out_size) {
         return ret;
 
     const char* name = dent->name;
-    PAL_CORE_TOPO_INFO* core_topology = &g_pal_control->topo_info.core_topology[cpu_num];
+    PAL_CORE_TOPO_INFO* core_topology = &g_pal_public_state->topo_info.core_topology[cpu_num];
     const char* str;
     char buf[12];
     if (strcmp(name, "online") == 0) {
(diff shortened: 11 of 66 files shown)


boryspoplawski (changes requested):
> Reviewed 66 of 66 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/278)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[Pal/include/pal_internal.h, line 33 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-Mqr05JO3HkEe5eyUF_S:-Mqr05JP-FOWPDV85p1F:ba0s3z7) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/include/pal_internal.h#L33)):*
> ```objc
>     uint64_t instance_id;
> 
>     /* Both of these are obviously untrusted on SGX, but we use them only for operations on the
> ```

1) Both? There are 4 items here
2) Why did you add these here? Previous `pal_internal_state` did not have them and they seem Linux specific (as comment above states). We have `pal_linux_state` or something similar for such stuff.
___
*[Pal/regression/pal_regression.h, line 10 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-Mqr10Oi80gTNOG7z7uL:-Mqr10Oi80gTNOG7z7uM:b-p1kghz) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/regression/pal_regression.h#L10)):*
> ```objc
> 
> // TODO: this needs to be fixed, it pretends to be a variable but it actually is a function call.
> #define pal_public_state (*DkGetPalPublicState())
> ```

Why not change it now? You've changed the name, so had to modify each usage anyway. It seems like you can just `s/pal_public_state\./DkGetPublicState()->/g`
___
*[Pal/src/host/Linux/pal_linux.h, line 30 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-Mqr3itr9Sya0jpVrPPn:-Mqr3itr9Sya0jpVrPPo:bfzrsfr) ([raw file](https://github.com/gramineproject/gramine/blob/63a73ae54c25eb0490c9fb0633d9b96db7cd72ed/Pal/src/host/Linux/pal_linux.h#L30)):*
> ```objc
> 
> extern struct pal_linux_state {
>     const char**    host_environ;
> ```

Why did you move them out? They seem to belong to this struct.

pwmarcz (changes requested):
>I'd put in these structs +/- only the state which sounds like global to the whole PAL, not just to a single submodule.

You mean, keep variables like `static int g_foo;` as static, but move the ones used in many modules to one of the structures?

I think it's a good approach. Especially for the host-specific globals; it would make it clearer that they're host-specific.

> Reviewed 66 of 66 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/278#-:-MqsZVLbD2INNHvg59Jh:b-q4qkno)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[-- commits, line 3 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-MqsUiUi5kUjSx6XOxDd):*
This is a big diff, and it would take me some time to determine what exactly was renamed. Can you add more information to this message? At least the list of new structures (`pal_public_state`, `pal_common_state` etc.)
___
*[Documentation/pal/host-abi.rst, line 87 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-MqsSJ318bTW_oyxJJbG:-MqsSJ318bTW_oyxJJbH:b5vs0ov) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Documentation/pal/host-abi.rst#L87)):*
> ```reStructuredText
> The control block in Gramine is a structure that provides static immutable
> information about the current process and its host. The address of the control
> block is resolved via :func:`DkGetPalPublicState()` and can be memorized in a global
> ```

I'm not sure if "state" is the best name here, given that the information is immutable. I don't have a better idea, though.

By the way, are all other structures immutable as well? (After the initial setup, of course).
___
*[Pal/regression/pal_regression.h, line 10 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-Mqr10Oi80gTNOG7z7uL:-MqsTNzgCK4l7R1Ls-ar:b-4ecu02) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/regression/pal_regression.h#L10)):*

Why not change it now? You've changed the name, so had to modify each usage anyway. It seems like you can just `s/pal_public_state\./DkGetPublicState()->/g`
</blockquote></details>

+1. At first I thought it's a global variable and you didn't add `g_`.
___
*[Pal/src/host/Linux-SGX/pal_linux.h, line 35 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-MqsULwA5JM6RcZjpwmO:-MqsULwA5JM6RcZjpwmP:bvv7fng) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/src/host/Linux-SGX/pal_linux.h#L35)):*
> ```objc
> 
> /* Part of Linux-SGX PAL private state which is not shared with other PALs. */
> extern struct pal_linuxsgx_state {
> ```

Looking at all this code, `{g_}pal_linuxsgx_state` is tripping me up - the `linuxsgx` part is hard to read.

Maybe `g_pal_linux_sgx_state` would be better? (I'm not sure).

dimakuv (changes requested):
I agree with Pawel.

> pal_linux_state - PAL/Linux private state
> pal_linuxsgx_state - PAL/Linux-SGX private state

While I agree with the general naming scheme, this notation `pal_{name}_state` is a bit cumbersome. Technically, these structs are only used inside a PAL-specific directory (e.g. `Pal/host/Linux`), and the naming feels redundant. I would prefer a uniform name `pal_specific_state` (or something similar), where all PAL-specific fields will be located.

But I don't insist, maybe it's just me.

> Reviewed 66 of 66 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/278#-:-Mqsl1q92IvDS7UhrALo:b-nt42d9)* status: all files reviewed, 16 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, and @pwmarcz)

___
*[Documentation/pal/host-abi.rst, line 82 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-MqrwPV28sdc4iVLeFLd:-MqrwPV28sdc4iVLeFLe:b9mmsds) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Documentation/pal/host-abi.rst#L82)):*
> ```reStructuredText
>    :members:
> 
> Gramine control block
> ```

It's not really `control block` anymore, it is `Gramine PAL (public) state`. Please rename everywhere in this doc.
___
*[Documentation/pal/host-abi.rst, line 90 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-MqrwsbDCqJLH4jagIoc:-MqrwsbDCqJLH4jagIod:bkhje52) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Documentation/pal/host-abi.rst#L90)):*
> ```reStructuredText
> variable for ease of use.
> 
> The fields of the Gramine control block are defined as follows:
> ```

This list doesn't make sense. `PAL_CONTROL` (now `pal_public_state`) is a struct that *contains* `PAL_*_INFO` fields.

I guess we need to split this list into two paragraphs. One just gives fields of `PAL_CONTROL` (now `pal_public_state`). Another one talks about sub-fields of the fields `PAL_CPU_INFO`, ...
___
*[Pal/include/pal_internal.h, line 28 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-Mqrxsir5I0-QahuhLEC:-Mqrxsir5I0-QahuhLED:b-kaoqlm) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/include/pal_internal.h#L28)):*
> ```objc
>  * Part of PAL state which is common to all PALs (trusted part, in case of TEE PALs).
>  * Most of it is actually very Linux-specific, we'll need to refactor it if we ever add a non-Linux
>  * PAL.
> ```

This can happen pretty soon :) Could you add a TODO/FIXME here?
___
*[Pal/include/pal_internal.h, line 33 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-Mqr05JO3HkEe5eyUF_S:-MqryABBBGJ34NlvcYHU:b-4jkv37) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/include/pal_internal.h#L33)):*

1) Both? There are 4 items here
2) Why did you add these here? Previous `pal_internal_state` did not have them and they seem Linux specific (as comment above states). We have `pal_linux_state` or something similar for such stuff.
</blockquote></details>

I agree with Borys. But I would go even further and completely decouple them -- you spend more LoCs explaining them in your comments than simply duplicating them (from what I see, e.g. `host_environ` is not even used in Linux-SGX PAL).

Also, in new PALs (think of Gramine-VM), we don't have "host environment variables" etc, so these fields will be useless.
___
*[Pal/include/pal/pal.h, line 113 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-MqrzD-YBcO0d6o_fKD8:-MqrzD-YBcO0d6o_fKD9:bpmfr80) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/include/pal/pal.h#L113)):*
> ```objc
> 
> /* Part of PAL state which is shared between all PALs and accessible (read-only) by the code run by
>  * PAL (usually our LibOS). */
> ```

`by the code run by PAL` is ambigious. Maybe `by the binary started by PAL` or `by the entrypoint loaded by PAL`?
___
*[Pal/regression/pal_regression.h, line 10 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-Mqr10Oi80gTNOG7z7uL:-MqrzcVD44XgPv-lXQmR:b-eevk75) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/regression/pal_regression.h#L10)):*

+1. At first I thought it's a global variable and you didn't add `g_`.
</blockquote></details>

+1. When I wrote that hack, I didn't want to modify any regression tests' sources (well, at that time I really hated PAL regression tests and wanted them to be removed :)). Now you modify them anyway, so let's just use the function.
___
*[Pal/src/db_main.c, line 29 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-Mqs-3jt5xDsv6EUJABX:-Mqs-3jt5xDsv6EUJABY:btz0jrh) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/src/db_main.c#L29)):*
> ```C
> }
> 
> struct pal_common_state g_pal_common_state;
> ```

Maybe move it right-after includes? Right now it's weirdly situated in-between two functions.
___
*[Pal/src/db_main.c, line 529 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-Mqs-Sai1CnldY1nn_x9:-Mqs-Sai1CnldY1nn_xA:b44nbsa) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/src/db_main.c#L529)):*
> ```C
> 
>     g_pal_public_state.host_type       = XSTRINGIFY(HOST_TYPE);
>     g_pal_public_state.manifest_root   = g_pal_common_state.manifest_root;
> ```

Why do we need this duplication? Why not init `g_pal_public_state` immediately?
___
*[Pal/src/db_main.c, line 537 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-Mqs-YS32dMibe3OO7yh:-Mqs-YS32dMibe3OO7yi:b44nbsa) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/src/db_main.c#L537)):*
> ```C
>                                     &g_pal_public_state.user_address.end);
> 
>     g_pal_public_state.alloc_align = g_pal_common_state.alloc_align;
> ```

Why do we need this duplication? Why not init `g_pal_public_state` immediately?
___
*[Pal/src/host/Linux/db_main.c, line 44 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-MqsiXtu9jf0TMHyzr7E:-MqsiXtu9jf0TMHyzr7F:b-utao28) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/src/host/Linux/db_main.c#L44)):*
> ```C
> const size_t g_page_size = PRESET_PAGESIZE;
> 
> static void read_args_from_stack(void* initial_rsp, int* out_argc, const char*** out_argv,
> ```

Maybe also change a name? Like `read_info_from_stack`?
___
*[Pal/src/host/Linux/db_threading.c, line 27 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-MqsiwyuENBSUe_Zdm2c:-MqsiwyuENBSUe_Zdm2d:bi6dz8s) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/src/host/Linux/db_threading.c#L27)):*
> ```C
> 
> /* Linux PAL cannot use mmap/unmap to manage thread stacks because this may overlap with
>  * pal_public_state.user_address. Linux PAL also cannot just use malloc/free because DkThreadExit
> ```

`g_` missing?
___
*[Pal/src/host/Linux/pal_linux.h, line 30 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-Mqr3itr9Sya0jpVrPPn:-Mqsj1hJEQJXAbaYyECR:b-5kx2ei) ([raw file](https://github.com/gramineproject/gramine/blob/63a73ae54c25eb0490c9fb0633d9b96db7cd72ed/Pal/src/host/Linux/pal_linux.h#L30)):*

Why did you move them out? They seem to belong to this struct.
</blockquote></details>

+1 to Borys's request. They seem to not even be used in Linux-SGX (at least not all of them).
___
*[Pal/src/host/Linux-SGX/db_main.c, line 36 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-MqsjE3j6pG1mN9uwjRP:-MqsjE3j6pG1mN9uwjRQ:bznqee) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/src/host/Linux-SGX/db_main.c#L36)):*
> ```C
> #include "toml_utils.h"
> 
> struct pal_sec g_pal_sec;
> ```

Yeah, we definitely need to remove this one.
___
*[Pal/src/host/Linux-SGX/pal_linux.h, line 35 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-MqsULwA5JM6RcZjpwmO:-MqskaQRDKqu6o2hlCNZ:b-ibhnhd) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/src/host/Linux-SGX/pal_linux.h#L35)):*

Looking at all this code, `{g_}pal_linuxsgx_state` is tripping me up - the `linuxsgx` part is hard to read.

Maybe `g_pal_linux_sgx_state` would be better? (I'm not sure).
</blockquote></details>

I found this name easy to read, so I'm neutral on this.
___
*[Pal/src/host/Linux-SGX/pal_linux.h, line 37 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-Mqskltb79Sl3Qtmk2Qs:-Mqskltb79Sl3Qtmk2Qt:blp0n89) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/src/host/Linux-SGX/pal_linux.h#L37)):*
> ```objc
> extern struct pal_linuxsgx_state {
>     /* enclave information */
>     bool enclave_initialized; /* thread creation ECALL is allowed only after this is set */
> ```

Could you indent the comment, for uniformity?

mkow (commented):
> You mean, keep variables like `static int g_foo`; as static, but move the ones used in many modules to one of the structures?

Yup, something like this. Ofc with possible exceptions, we'll decide case-by-case.

> I would prefer a uniform name `pal_specific_state` (or something similar), where all PAL-specific fields will be located.

I thought about something like this, but having multiple structures and functions with the same name and different contents is a bit PITA when changing the code (e.g. when grepping for something). But that's not a strong preference. What do others think?

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/278#-:-Mr33N7A3MkhSOGcbxi-:b-xagwu5)* status: 43 of 66 files reviewed, 16 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "squash! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @pwmarcz)

___
*[-- commits, line 3 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-MqsUiUi5kUjSx6XOxDd):*

This is a big diff, and it would take me some time to determine what exactly was renamed. Can you add more information to this message? At least the list of new structures (`pal_public_state`, `pal_common_state` etc.)
</blockquote></details>

Done. I moved here a part of the PR description, is this enough?
___
*[Documentation/pal/host-abi.rst, line 82 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-MqrwPV28sdc4iVLeFLd:-Mr36ezZ3Ob4B_RqOxqv:b-n7vh9r) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Documentation/pal/host-abi.rst#L82)):*

It's not really `control block` anymore, it is `Gramine PAL (public) state`. Please rename everywhere in this doc.
</blockquote></details>

Oh, right, I missed this whole thing here.
___
*[Documentation/pal/host-abi.rst, line 87 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-MqsSJ318bTW_oyxJJbG:-Mr37hH63NOLpXCDoJcb:b5pdo57) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Documentation/pal/host-abi.rst#L87)):*
Good point, "state" doesn't connect well with "immutable". But I'd leave it for uniformity with the naming of the rest of the global structures.

> By the way, are all other structures immutable as well? (After the initial setup, of course).

I think this is the case currently, although it may change if we add support for hot-plug in topology.
___
*[Documentation/pal/host-abi.rst, line 90 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-MqrwsbDCqJLH4jagIoc:-Mr3C25o94sYJNIfLPGl:b-y538cy) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Documentation/pal/host-abi.rst#L90)):*

This list doesn't make sense. `PAL_CONTROL` (now `pal_public_state`) is a struct that *contains* `PAL_*_INFO` fields.

I guess we need to split this list into two paragraphs. One just gives fields of `PAL_CONTROL` (now `pal_public_state`). Another one talks about sub-fields of the fields `PAL_CPU_INFO`, ...
</blockquote></details>

What about now? I moved them into a sub-sub-sub-sub-sub-section ;) Anyways, in the next PR I want to create just a single "topology structure" to cover all of this.
___
*[Pal/include/pal_internal.h, line 28 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-Mqrxsir5I0-QahuhLEC:-Mr3DvjvCIwbXMQAoYBz:bhfgl6u) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/include/pal_internal.h#L28)):*

This can happen pretty soon :) Could you add a TODO/FIXME here?
</blockquote></details>

Hmm, but it's not a "TODO" in the current state of the project, i.e. for me a "TODO" suggest that this is something actionable - something we should fix when we find some time for it. But here it's something which makes sense to change only after we add another PAL.
___
*[Pal/include/pal_internal.h, line 33 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-Mqr05JO3HkEe5eyUF_S:-Mr3F5Xw48LOmU82rV88:b93y6un) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/include/pal_internal.h#L33)):*
> Both? There are 4 items here

Fixed.

> Why did you add these here? Previous `pal_internal_state` did not have them and they seem Linux specific (as comment above states). We have `pal_linux_state` or something similar for such stuff.

But what do you propose? To duplicate them to both `pal_linux_state` and `pal_linuxsgx_state`?

> Also, in new PALs (think of Gramine-VM), we don't have "host environment variables" etc, so these fields will be useless.

Yeah, but adding this PAL will require much bigger rework of PAL anyways, I think.
I missed that `host_environ` and `host_pid` are used only on Linux, fixed.
___
*[Pal/include/pal/pal.h, line 113 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-MqrzD-YBcO0d6o_fKD8:-Mr3I9wp64Mkr_PpnZRh:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/include/pal/pal.h#L113)):*

`by the code run by PAL` is ambigious. Maybe `by the binary started by PAL` or `by the entrypoint loaded by PAL`?
</blockquote></details>

Done.
___
*[Pal/regression/pal_regression.h, line 10 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-Mqr10Oi80gTNOG7z7uL:-Mr3Julf1ArHWE4qOkXH:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/regression/pal_regression.h#L10)):*

+1. When I wrote that hack, I didn't want to modify any regression tests' sources (well, at that time I really hated PAL regression tests and wanted them to be removed :)). Now you modify them anyway, so let's just use the function.
</blockquote></details>

Done.
___
*[Pal/src/db_main.c, line 29 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-Mqs-3jt5xDsv6EUJABX:-Mr3JzsN1VDN3MCZCWuD:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/src/db_main.c#L29)):*

Maybe move it right-after includes? Right now it's weirdly situated in-between two functions.
</blockquote></details>

Done.
___
*[Pal/src/db_main.c, line 529 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-Mqs-Sai1CnldY1nn_x9:-Mr3L7yb-Le7cqjGV-Dt:b-66313w) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/src/db_main.c#L529)):*

Why do we need this duplication? Why not init `g_pal_public_state` immediately?
</blockquote></details>

No idea, here I only applied the renaming. I guess it doesn't make sense, let me change it...
___
*[Pal/src/db_main.c, line 537 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-Mqs-YS32dMibe3OO7yh:-Mr3Rkae3Y_7M_1tKX3U:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/src/db_main.c#L537)):*

Why do we need this duplication? Why not init `g_pal_public_state` immediately?
</blockquote></details>

Done.
___
*[Pal/src/host/Linux/db_main.c, line 44 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-MqsiXtu9jf0TMHyzr7E:-Mr3TEUeDhnmPssJoDN7:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/src/host/Linux/db_main.c#L44)):*

Maybe also change a name? Like `read_info_from_stack`?
</blockquote></details>

Done.
___
*[Pal/src/host/Linux/db_threading.c, line 27 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-MqsiwyuENBSUe_Zdm2c:-Mr3TMBV5rEoSA5d10_x:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/src/host/Linux/db_threading.c#L27)):*

`g_` missing?
</blockquote></details>

Done.
___
*[Pal/src/host/Linux/pal_linux.h, line 30 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-Mqr3itr9Sya0jpVrPPn:-Mr3TPLF4pPSpRBWyvMB:bxacs3i) ([raw file](https://github.com/gramineproject/gramine/blob/63a73ae54c25eb0490c9fb0633d9b96db7cd72ed/Pal/src/host/Linux/pal_linux.h#L30)):*

+1 to Borys's request. They seem to not even be used in Linux-SGX (at least not all of them).
</blockquote></details>

I accidentally moved too much. The intention was to deduplicate them, as some are shared between both PALs.
But I can move them back, see the other discussion.
___
*[Pal/src/host/Linux-SGX/pal_linux.h, line 35 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-MqsULwA5JM6RcZjpwmO:-Mr3Tsbk5OO1J8MdZTpr:b5oyjze) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/src/host/Linux-SGX/pal_linux.h#L35)):*

I found this name easy to read, so I'm neutral on this.
</blockquote></details>

It's a bit easier to read for me if the structure is like `{toplevel_component}_{pal_name}_{struct_purpose}`. But also no strong preference.
@boryspoplawski any thoughts?
___
*[Pal/src/host/Linux-SGX/pal_linux.h, line 37 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-Mqskltb79Sl3Qtmk2Qs:-Mr3Unj1C6wyhqZZ2vZ3:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/src/host/Linux-SGX/pal_linux.h#L37)):*

Could you indent the comment, for uniformity?
</blockquote></details>

Done.

pwmarcz (approved):
I'd also like to have different-named structures rather than `pal_specific_state`. Not only it's easier to grep, it's also easier to read such code (you don't need to remember that because we're in this directory, *this* `pal_specific_state` has such-and-such fields), and probably also easier to talk about it.

> Reviewed 23 of 23 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/278#-:-Mr6kT_YCY476KOz6LDK:b-4nb0d6)* status: all files reviewed, 13 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "squash! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[-- commits, line 3 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-MqsUiUi5kUjSx6XOxDd):*

Done. I moved here a part of the PR description, is this enough?
</blockquote></details>

Looks good to me, thanks.
___
*[Pal/include/pal_internal.h, line 33 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-Mqr05JO3HkEe5eyUF_S:-Mr6mCJv8FpdTJNuCuO2:bj8510b) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/include/pal_internal.h#L33)):*

> Both? There are 4 items here

Fixed.

> Why did you add these here? Previous `pal_internal_state` did not have them and they seem Linux specific (as comment above states). We have `pal_linux_state` or something similar for such stuff.

But what do you propose? To duplicate them to both `pal_linux_state` and `pal_linuxsgx_state`?

> Also, in new PALs (think of Gramine-VM), we don't have "host environment variables" etc, so these fields will be useless.

Yeah, but adding this PAL will require much bigger rework of PAL anyways, I think.
I missed that `host_environ` and `host_pid` are used only on Linux, fixed.
</blockquote></details>

I'd also be careful about putting too much Linux-specific stuff in the common part. I agree we need to deduplicate `Linux` and `Linux-SGX` but the way to go is to use `Linux-common`.

Still, it's probably not a good solution here (we would need to add *another* global structure), so I don't have a strong opinion of what to do here. Keeping these variables duplicated is fine for me, keeping them here is also fine (as long as we revisit that once we add a non-Linux PAL).
___
*[Pal/src/host/Linux-SGX/pal_linux.h, line 35 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-MqsULwA5JM6RcZjpwmO:-Mr6nJIH7XoKkcyMcS2T:b-42ap4u) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/src/host/Linux-SGX/pal_linux.h#L35)):*

It's a bit easier to read for me if the structure is like `{toplevel_component}_{pal_name}_{struct_purpose}`. But also no strong preference.
@boryspoplawski any thoughts?
</blockquote></details>

If nobody else is bothered by it, let's leave it this way. I agree with the argument about structure.

dimakuv (dismissed):
Ok, then I concur.

> Reviewed 23 of 23 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/278#-:-Mr6wlpKDOnAfnYGoJe8:b-p498c5)* status: all files reviewed, 3 unresolved discussions, "squash! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[Pal/include/pal_internal.h, line 28 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-Mqrxsir5I0-QahuhLEC:-Mr6xKCh9hrtGI8goHcF:bs4ixp3) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/include/pal_internal.h#L28)):*

Hmm, but it's not a "TODO" in the current state of the project, i.e. for me a "TODO" suggest that this is something actionable - something we should fix when we find some time for it. But here it's something which makes sense to change only after we add another PAL.
</blockquote></details>

OK, fair enough.
___
*[Pal/include/pal_internal.h, line 33 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-Mqr05JO3HkEe5eyUF_S:-Mr6xd1N-WJUbzFAgXI0:bolx4vb) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/include/pal_internal.h#L33)):*

I'd also be careful about putting too much Linux-specific stuff in the common part. I agree we need to deduplicate `Linux` and `Linux-SGX` but the way to go is to use `Linux-common`.

Still, it's probably not a good solution here (we would need to add *another* global structure), so I don't have a strong opinion of what to do here. Keeping these variables duplicated is fine for me, keeping them here is also fine (as long as we revisit that once we add a non-Linux PAL).
</blockquote></details>

I'm fine with the current state now.

boryspoplawski (changes requested):
> Reviewed 23 of 23 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/278)* status: all files reviewed, 1 unresolved discussion, "squash! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[Pal/include/pal_internal.h, line 33 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-Mqr05JO3HkEe5eyUF_S:-Mr75VBj4cZDvmiRNsPS:b69wlcq) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/include/pal_internal.h#L33)):*
> But what do you propose? To duplicate them to both pal_linux_state and pal_linuxsgx_state?

Yes. I would move the remaining Linux specific fields to PAL specific structs (duplicating them).
___
*[Pal/src/host/Linux-SGX/pal_linux.h, line 35 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-MqsULwA5JM6RcZjpwmO:-Mr78-sO-Ze7IAbUXRm-:b-3opnz5) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/src/host/Linux-SGX/pal_linux.h#L35)):*

If nobody else is bothered by it, let's leave it this way. I agree with the argument about structure.
</blockquote></details>

I don't mind either version

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/278)* status: 59 of 66 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @pwmarcz)

___
*[Pal/include/pal_internal.h, line 33 at r1](https://reviewable.io/reviews/gramineproject/gramine/278#-Mqr05JO3HkEe5eyUF_S:-Mr7FXfxAVlbo7dnhU55:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/672e5300fec61b9489822dfac69079b0f145724a/Pal/include/pal_internal.h#L33)):*

> But what do you propose? To duplicate them to both pal_linux_state and pal_linuxsgx_state?

Yes. I would move the remaining Linux specific fields to PAL specific structs (duplicating them).
</blockquote></details>

Done.

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/278)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/278)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

pwmarcz (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/278)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/278)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/278)* status: :shipit: complete! all files reviewed, all discussions resolved

pwmarcz (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/278)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #152, opened 2021-10-18
[Pal] Rework ELF parsing/RTLD code

dimakuv:
## Description of the changes 

This PR completely reworks the code to load ELF binaries into the memory and link them against each other. Note that PAL loads only two kinds of ELF binaries: the LibOS shared library and the PAL regression tests. This allows us to simplify the code by posing the following restrictions:

- ELF binaries must be linked with RELRO (Relocation Read-Only).
- They must have old-style hash (DT_HASH) table (new code doesn't use hashing at all; the table is needed solely to figure out the number of dynamic symbols in the SYMTAB).
- They must have DYN or EXEC object file type.

Closes #86.

## How to test this PR? 

All tests must pass. Also, try GDBing manually.

As a side effect, some previously-disabled PAL tests are enabled.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/152)

PR #281, opened 2021-12-15
[Pal] Always map LibOS binary inside enclave memory

dimakuv:
## Description of the changes 

In the [recent RTLD refactoring](https://github.com/gramineproject/gramine/pull/152), the LibOS binary was initially mapped into memory via `_DkStreamMap(..., /*prot=*/0, ...)`. This led to a corner case in the Linux-SGX PAL: if `sgx.file_check_policy = "allow_all_but_log"` was set in the manifest (or the LibOS binary was marked in `sgx.allowed_files`), the LibOS binary was considered an allowed file. And because protections were `0` (aka MAP_SHARED), Linux-SGX's `file_map()` callback decided to map the file in untrusted memory, which was incorrect and led to failures later.

This commit fixes this bug by specifying `PAL_PROT_WRITECOPY` -- this guarantees that the file will be mapped inside enclave memory.

Kudos to @aneessahib who reported this issue.

## How to test this PR? 

I modified one of the LibOS regression tests to catch this in the future.

A simple way to test it is to remove all `sgx.trusted_files`/`sgx.allowed_files` in some example and replace with `sgx.file_check_policy = "allow_all_but_log"`. For example, doing this in the Bash example:
```
diff --git a/CI-Examples/bash/manifest.template b/CI-Examples/bash/manifest.template
@@ -32,14 +32,4 @@ sgx.nonpie_binary = true
 sgx.enclave_size = "256M"
 sgx.thread_num = 4

-sgx.trusted_files = [
-  "file:{{ gramine.libos }}",
-  "file:{{ execdir }}/",
-  "file:{{ gramine.runtimedir() }}/",
-  "file:{{ arch_libdir }}/",
-  "file:/usr/{{ arch_libdir }}/",
-]
-
-sgx.allowed_files = [
-  "file:scripts/",
-]
+sgx.file_check_policy = "allow_all_but_log"
```

- Without this PR, fails like this:
```
warning: Allowing access to unknown file '/home/dimakuv/gramineproject/gramine/built-debug/lib/x86_64-linux-gnu/gramine/libsysdb.so' due to file_check_policy settings.
error: Failed to map segment from ELF file
error: Could not map the ELF file into memory and then relocate it
error: PAL failed at ../Pal/src/db_main.c:pal_main:558 (exitcode = 4, reason=Unable to load loader.entrypoint)
debug: DkProcessExit: Returning exit code 4
```

- With this PR, succeeds.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/281)

### DIFF
--- a/LibOS/shim/test/regression/file_check_policy_allow_all_but_log.manifest.template
+++ b/LibOS/shim/test/regression/file_check_policy_allow_all_but_log.manifest.template
@@ -22,10 +22,9 @@ sgx.debug = true
 
 sgx.file_check_policy = "allow_all_but_log"
 
-sgx.trusted_files = [
-  "file:{{ gramine.libos }}",
-  "file:{{ gramine.runtimedir(libc) }}/",
-]
+# there are no "file:{{ gramine.libos }}" and "file:{{ gramine.runtimedir() }}/"
+# entries in `sgx.trusted_files` -- this is on purpose; we want to test that
+# `allow_all_but_log` also applies to Gramine-runtime files (e.g., LibOS binary)
 
 # below entry in sgx.trusted_files is to test TOML-table syntax without `sha256`
 [[sgx.trusted_files]]
--- a/Pal/src/db_rtld.c
+++ b/Pal/src/db_rtld.c
@@ -377,10 +377,12 @@ static int create_and_relocate_entrypoint(PAL_HANDLE handle, const char* elf_fil
          * first segment's p_vaddr. This is to ensure that l_base_diff will not be less than 0.
          *
          * FIXME: We (ab)use _DkStreamMap() because _DkVirtualMemoryAlloc() cannot be used to
-         *        allocate memory at PAL-chosen address (it expects `map_addr` to be fixed).
+         *        allocate memory at PAL-chosen address (it expects `map_addr` to be fixed). Note
+         *        that `PAL_PROT_WRITECOPY` is specified to prevent allocating memory in untrusted
+         *        memory in case of Linux-SGX PAL (see Linux-SGX/db_files.c:file_map).
          */
         void* map_addr = NULL;
-        ret = _DkStreamMap(handle, &map_addr, /*prot=*/0, /*offset=*/0,
+        ret = _DkStreamMap(handle, &map_addr, /*prot=*/PAL_PROT_WRITECOPY, /*offset=*/0,
                            loadcmds[loadcmds_cnt - 1].alloc_end);
         if (ret < 0) {
             log_error("Failed to allocate memory for all LOAD segments of DYN ELF file");


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/281)* status: 0 of 1 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

___
*[Pal/src/db_rtld.c, line 382 at r1](https://reviewable.io/reviews/gramineproject/gramine/281#-Mqx6g6Y1tEXJNnzeHqB:-Mqx6g6Y1tEXJNnzeHqC:bhcsgm2) ([raw file](https://github.com/gramineproject/gramine/blob/ebc380dc24e04cb4418da71ea49c9c6e63bf3e83/Pal/src/db_rtld.c#L382)):*
> ```C
>          *        allocate memory at PAL-chosen address (it expects `map_addr` to be fixed). Note
>          *        that `PAL_PROT_WRITECOPY` is specified to prevent allocating memory in untrusted
>          *        memory in case of Linux-SGX PAL (see Linux-SGX/db_files.c:file_map).
> ```

FYI: We should probably disable that weird "improvement" of mmapping files in untrusted memory in Linux-SGX PAL. For now, this quick fix does the job.

(We need a proper rehaul of the complete file-system implementation in Linux-SGX PAL.)

pwmarcz (approved):
I verified the fix manually.

> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/281#-:-MqxktB_-Fj9HrkDw_ny:bn67t5x)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

___
*[Pal/src/db_rtld.c, line 382 at r1](https://reviewable.io/reviews/gramineproject/gramine/281#-Mqx6g6Y1tEXJNnzeHqB:-MqxjWJD7zUv0qbEklm-:b-nk9hms) ([raw file](https://github.com/gramineproject/gramine/blob/ebc380dc24e04cb4418da71ea49c9c6e63bf3e83/Pal/src/db_rtld.c#L382)):*

FYI: We should probably disable that weird "improvement" of mmapping files in untrusted memory in Linux-SGX PAL. For now, this quick fix does the job.

(We need a proper rehaul of the complete file-system implementation in Linux-SGX PAL.)
</blockquote></details>

Wow, this hack is getting worse :)

Apart from this surprising behavior, `_DkStreamMap` also makes no distiction between "user" and "PAL-internal" memory (I think it always assumes it's allocating "user" memory).

And, as noted by you, we should use `_DkVirtualMemoryAlloc()` here but it needs a fixed address. So I think what we need is actually an overhaul of the memory API.

boryspoplawski (approved):
> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/281)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #279, opened 2021-12-14
Inodes


pwmarcz:
Parent issue: #24

This is an issue to track the inodes rewrite project, mostly for myself to keep my thoughts in order.

## Motivation

Currently Gramine uses `shim_dentry` as both "directory entry" and "file metadata" object. This is inconvenient for several reasons:

* It's hard or even impossible to implement unlink and rename correctly: if we we unlink/rename a file that is open, things often break. Old implementation had a hacky workaround that essentially copied file metadata to a handle, but it worked only partially.
* Locking is weird and broken. There's a global lock (`g_dcache_lock`) but *also* a per-object lock (`shim_dentry.lock`), and many operations do not use locks at all (https://github.com/gramineproject/graphene/issues/944).
* The `shim_dentry` structure is complicated. I think splitting it into dentry and inode will make both structures much simpler, and the interactions more manageable.

## Goal

* `shim_dentry` is a mapping from name to inode. It doesn't contain file information or filesystem-specific data.
* All access to `shim_dentry` is guarded by a global lock (`g_dcache_lock`). Operations on a dentry (e.g. lookup, open, unlink) all take `g_dcache_lock`.
* `shim_inode` holds file metadata (enough to answer `stat()`). It has a per-object lock for mutable data.

## Plan

Rough draft, later steps might evolve.

* :heavy_check_mark: Introduce `shim_inode`, convert `chroot` fs (PR #5)
* :heavy_check_mark: Define semantics for `d_ops` callbacks, make sure they take the global lock (`g_dcache_lock`) (PR #282)
* :heavy_check_mark: Get rid of `shim_dentry.lock` (make `g_dcache_lock` the global lock for dentry access)
  * (I want to simplify before converting more code to inodes)
  * :heavy_check_mark: Don't use `shim_dentry.lock` in the `fs_lock` module (POSIX locks) (PR #283)
  * :heavy_check_mark: Don't use `shim_dentry.lock` in filesystem callbacks (PR #282)
  * :heavy_check_mark: Remove `shim_dentry.data` and `shim_dentry.lock` fields (PR #414)
* :heavy_check_mark: Convert individual filesystems to use inodes
  * :heavy_check_mark: : tmpfs (PR #297)
  * :heavy_check_mark: synthetic nodes (PR #383)
  * :heavy_check_mark: pseudofs (PR #402)
  * :heavy_check_mark: : socket, pipe (PR #414)
[...]

PR #283, opened 2021-12-15
[LibOS] fs_lock: Use a separate lock for all operations

pwmarcz:
## Description of the changes 

Instead of using `shim_dentry.lock` for data stores in individual dentries, we use a global local (`g_fs_lock_lock`) for all fs_lock
related operations. This is in preparation for removing `shim_dentry.lock`.

Context: #279.

I realize that using a lock for guarding a field on another object (`shim_dentry`) doesn't look nice. However, I tried the alternative (taking the global lock for dentries, `g_dcache_lock`) and it breaks because of the spaghetti in filesystem code[1]. While I could probably work around this specific issue, I think it's better to keep this module separate (i.e. using its own lock) so that it's easier to reason about.

[1] If you want to know the gory details, it was `shim_pipe`: because it uses extra file descriptor, the callback for `open` actually "closes" one of these artificial file descriptors, triggering fs_lock while we're already holding `g_dcache_lock`. Because we don't expect to be holding `g_dcache_lock` at this stage, we try to lock it for the second time. IMO `shim_pipe` breaks some implicit assumptions here, but there might be more places like this.

## How to test this PR? 

CI should be enough.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/283)

### DIFF
--- a/LibOS/shim/include/shim_fs.h
+++ b/LibOS/shim/include/shim_fs.h
@@ -161,12 +161,11 @@ struct shim_dentry {
      * filesystem (see `shim_inode` below). Protected by `lock`. */
     struct shim_inode* inode;
 
-    /* File lock information, stored only in the main process. Protected by `lock`. See
-     * `shim_fs_lock.c`. */
+    /* File lock information, stored only in the main process. Managed by `shim_fs_lock.c`. */
     struct fs_lock* fs_lock;
 
     /* True if the file might have locks placed by current process. Used in processes other than
-     * main process, to prevent unnecessary IPC calls on handle close. Protected by `lock`. See
+     * main process, to prevent unnecessary IPC calls on handle close. Managed by
      * `shim_fs_lock.c`. */
     bool maybe_has_fs_locks;
 
--- a/LibOS/shim/src/fs/shim_fs_lock.c
+++ b/LibOS/shim/src/fs/shim_fs_lock.c
@@ -11,6 +11,12 @@
 #include "shim_lock.h"
 
 /*
+ * Global lock for the whole subsystem. Protects access to `g_fs_lock_list`, and also to dentry
+ * fields (`fs_lock` and `maybe_has_fs_locks`).
+ */
+static struct shim_lock g_fs_lock_lock;
+
+/*
  * Describes a pending request for a POSIX lock. After processing the request, the object is
  * removed, and a possible waiter is notified (see below).
  *
@@ -46,10 +52,10 @@ struct fs_lock {
     struct shim_dentry* dent;
 
     /* POSIX locks, sorted by PID and then by start position (so that we are able to merge and split
-     * locks). The ranges do not overlap within a given PID. Protected by `dent->lock`. */
+     * locks). The ranges do not overlap within a given PID. */
     LISTP_TYPE(posix_lock) posix_locks;
 
-    /* Pending requests. Protected by `dent->lock`. */
+    /* Pending requests. */
     LISTP_TYPE(posix_lock_request) posix_lock_requests;
 
     /* List node, for `g_fs_lock_list`. */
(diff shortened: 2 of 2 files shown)


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/283)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

___
*[LibOS/shim/src/fs/shim_fs_lock.c, line 15 at r1](https://reviewable.io/reviews/gramineproject/gramine/283#-Mr1jIWWBBbNJc76dfF1:-Mr1jIWWBBbNJc76dfF2:br53th6) ([raw file](https://github.com/gramineproject/gramine/blob/7d21531b80ad69becdec9309aa3b79a3b12e6cc8/LibOS/shim/src/fs/shim_fs_lock.c#L15)):*
> ```C
> /*
>  * Global lock for the whole subsystem. Protects access to `g_fs_lock_list`, and also to dentry
>  * fields (`fs_lock and `maybe_has_fs_locks`).
> ```

Missing closing backtick in `fs_lock`
___
*[LibOS/shim/src/fs/shim_fs_lock.c, line 628 at r1](https://reviewable.io/reviews/gramineproject/gramine/283#-Mr1jpbk5PoOUzzJ557R:-Mr1jpbk5PoOUzzJ557S:b3safci) ([raw file](https://github.com/gramineproject/gramine/blob/63a73ae54c25eb0490c9fb0633d9b96db7cd72ed/LibOS/shim/src/fs/shim_fs_lock.c#L628)):*
> ```C
>     *out_dents = dents;
>     return 0;
> }
> ```

I guess removing this two-phase deletion of fs locks is a huge benefit of this PR. I'm pretty sure the performance overhead of a single global lock is negligible in real-world cases.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/283)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

___
*[LibOS/shim/src/fs/shim_fs_lock.c, line 15 at r1](https://reviewable.io/reviews/gramineproject/gramine/283#-Mr1jIWWBBbNJc76dfF1:-Mr7lvYs-CW5aPH2H5_t:bt2vr1b) ([raw file](https://github.com/gramineproject/gramine/blob/7d21531b80ad69becdec9309aa3b79a3b12e6cc8/LibOS/shim/src/fs/shim_fs_lock.c#L15)):*

Missing closing backtick in `fs_lock`
</blockquote></details>

Fixed.
___
*[LibOS/shim/src/fs/shim_fs_lock.c, line 628 at r1](https://reviewable.io/reviews/gramineproject/gramine/283#-Mr1jpbk5PoOUzzJ557R:-Mr7lysCET_1owyQ60Ni:b-r2hcqy) ([raw file](https://github.com/gramineproject/gramine/blob/63a73ae54c25eb0490c9fb0633d9b96db7cd72ed/LibOS/shim/src/fs/shim_fs_lock.c#L628)):*

I guess removing this two-phase deletion of fs locks is a huge benefit of this PR. I'm pretty sure the performance overhead of a single global lock is negligible in real-world cases.
</blockquote></details>

Yes, I'm happy about it. But also about untangling this module from the rest of the system, it's more independent now.

I think a global lock could have an impact on performance if an application takes many locks for various ranges of a single file (since lock operations are `O(n)` in the number of file ranges). But I'm not sure how likely that is, and anyway, to fix it we'd also need to rewrite the lock list to something else.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/283)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/283)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #288, opened 2021-12-16
[tests] Work around LTP tests leaving running processes

pwmarcz:
## Description of the changes 

The previous (async) LTP runner only waited for the main process to exit. Right now we wait for all processes, which uncovered a (known) problem with signal delivery in Gramine: apparently LTP tests sometimes leave running child processes.

Because these tests are still useful for us, we do not treat such situations as a timeout, only log a warning.

## Decisions

I decided it'd be better to enable this workaround globally, than selectively enable it for "known-bad" tests. While there is a risk this will mask some real errors, this issue (as I understand) can happen with basically any program that relies on signal delivery, and deciding (probably experimentally) where this workaround is needed sounds like more trouble than it's worth.

Unfortunately, this implementation is slower: in case some processes remain, it ends up waiting the whole `timeout` seconds (the old implementation finished waiting as soon as the main process exited). My justification is that this is significantly harder to achieve using Python `subprocess` than it was using `asyncio` (I would probably have to either rewrite this function using `asyncio`, or using threads), and we would be doing the wrong thing anyway (we *should* wait for all processes and treat them not exiting as a bug...)

## How to test this PR? 

There should be no timeouts on CI.

You can also reproduce the issue manually, although it happens rarely. To do so, edit `test_ltp.py` to repeat `run_command` many times:

```diff
-    returncode, stdout, _stderr = run_command(full_cmd, timeout=timeout, can_fail=True)
+    for i in range(1000):
+        returncode, stdout, _stderr = run_command(full_cmd, timeout=timeout, can_fail=True)
```

Then, run one of the problematic tests:

```
python3 -m pytest -k sendfile04_64 -rA
```

(the `-rA` will display all the details even when the tests succeeed)

On `master`, this will occasionally report a timeout. On this branch, it should not time out, and will occasionally log a warning instead:

```
...
sendfile04_64    2  TPASS  :  sendfile() returned 14 : Bad address
sendfile04_64    3  TPASS  :  sendfile() returned 14 : Bad address
sendfile04_64    4  TPASS  :  sendfile() returned 14 : Bad address
sendfile04_64    5  TPASS  :  sendfile() returned 14 : Bad address
sendfile04_64    1  TPASS  :  sendfile() returned 14 : Bad address
sendfile04_64    2  TPASS  :  sendfile() returned 14 : Bad address
sendfile04_64    3  TPASS  :  sendfile() returned 14 : Bad address
sendfile04_64    4  TPASS  :  sendfile() returned 14 : Bad address
sendfile04_64    5  TPASS  :  sendfile() returned 14 : Bad address
---------------------------------------------------- Captured stderr call -----------------------------------------------------
[P6:T6:sendfile04_64] error: Sending IPC process-exit notification failed: -13
[P6:T6:sendfile04_64] error: Sending IPC process-exit notification failed: -13
[P6:T6:sendfile04_64] error: Sending IPC process-exit notification failed: -13
[P6:T6:sendfile04_64] error: Sending IPC process-exit notification failed: -13
[P6:T6:sendfile04_64] error: Sending IPC process-exit notification failed: -13
[P6:T6:sendfile04_64] error: Sending IPC process-exit notification failed: -13
[P6:T6:sendfile04_64] error: Sending IPC process-exit notification failed: -13
------------------------------------------------------ Captured log call ------------------------------------------------------
command: ['gramine-direct', 'sendfile04_64']
must_pass: all
run_command: Command ['gramine-direct', 'sendfile04_64'] timed out, but the main process exited. This might be due to a known issue with Gramine failing to deliver a signal. Continuing.
run_command: Command ['gramine-direct', 'sendfile04_64'] timed out, but the main process exited. This might be due to a known issue with Gramine failing to deliver a signal. Continuing.
returncode: 0
passed: [1, 2, 3, 4, 5]
failed: []
=================================================== short test summary info ===================================================
PASSED test_ltp.py::test_ltp[sendfile04_64]
=================================== 1 passed, 1327 deselected, 2 warnings in 97.28 seconds ====================================
```

(The above is an aggregated output from these 1000 Gramine runs).

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/288)

### DIFF
--- a/python/graminelibos/regression.py
+++ b/python/graminelibos/regression.py
@@ -32,6 +32,9 @@ def run_command(cmd, *, timeout, can_fail=False, **kwds):
         except subprocess.TimeoutExpired:
             timed_out = True
 
+        proc.poll()
+        main_returncode = proc.returncode
+
         # Kill the whole process group: even if we did not time out, there might be some processes
         # remaining
 
@@ -61,12 +64,27 @@ def run_command(cmd, *, timeout, can_fail=False, **kwds):
         sys.stderr.write(stderr)
 
         if timed_out:
-            raise AssertionError('Command {} timed out after {} s'.format(cmd, timeout))
+            if main_returncode is not None:
+                # XXX: Don't fail the test as long the main process exited (i.e. if it left dangling
+                # child processes). This can happen due to a known issue with Gramine failing to
+                # deliver a signal for an arbitrary amount of time. See the comment in
+                # `shim_internal.h:handle_signal` for details.
+                #
+                # This happens occasionally when running LTP tests (e.g. `sendfile04`,
+                # `fdatasync01`, `recvfrom01`, `sendto01`) that send SIGKILL to child processes.
+                logging.warning(
+                    'run_command: Command %s timed out, but the main process exited. This might be '
+                    'due to a known issue with Gramine failing to deliver a signal. Continuing.',
+                    cmd)
+            else:
+                raise AssertionError('Command {} timed out after {} s'.format(cmd, timeout))
 
-        if proc.returncode and not can_fail:
+        assert main_returncode is not None
+
+        if main_returncode != 0 and not can_fail:
             raise subprocess.CalledProcessError(proc.returncode, cmd, raw_stdout, raw_stderr)
 
-        return proc.returncode, stdout, stderr
+        return main_returncode, stdout, stderr
 
 
 class RegressionTestCase(unittest.TestCase):


boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/288)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/288#-Mr33g5F4MMDwtRa4kS1:-Mr33g5F4MMDwtRa4kS2:b4exas5) (no related file):*
> Unfortunately, this implementation is slower: in case some processes remain, it ends up waiting the whole timeout seconds (the old implementation finished waiting as soon as the main process exited). My justification is that this is significantly harder to achieve using Python subprocess than it was using asyncio (I would probably have to either rewrite this function using asyncio, or using threads), and we would be doing the wrong thing anyway (we should wait for all processes and treat them not exiting as a bug...)

I would prefer to not wait the full timeout (as probably multiple LTP tests have this issue, so it might extend the testing time substantially), but I also don't see a nice option of implementing this (without rewriting to asyncio).
@mkow @dimakuv ideas?

---
___
*[python/graminelibos/regression.py, line 35 at r1](https://reviewable.io/reviews/gramineproject/gramine/288#-Mr350Ec2C8cQWCjPuWs:-Mr350Ec2C8cQWCjPuWt:blj4bsp) ([raw file](https://github.com/gramineproject/gramine/blob/facde8e7f101f880923ebe204218f97f60ed1dfb/python/graminelibos/regression.py#L35)):*
> ```Python
>             timed_out = True
> 
>         proc.poll()
> ```

Is this correct? If the process timesout, then it's still alive and this will not set the return code. I think you should move it after `os.killpg` (or before `if timed_out:` below).

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/288)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/288#-Mr33g5F4MMDwtRa4kS1:-Mr3C5Wt84VyiKFmRfSZ:b3xalvj) (no related file):*

> Unfortunately, this implementation is slower: in case some processes remain, it ends up waiting the whole timeout seconds (the old implementation finished waiting as soon as the main process exited). My justification is that this is significantly harder to achieve using Python subprocess than it was using asyncio (I would probably have to either rewrite this function using asyncio, or using threads), and we would be doing the wrong thing anyway (we should wait for all processes and treat them not exiting as a bug...)

I would prefer to not wait the full timeout (as probably multiple LTP tests have this issue, so it might extend the testing time substantially), but I also don't see a nice option of implementing this (without rewriting to asyncio).
@mkow @dimakuv ideas?
</blockquote></details>

Multiple LTP tests are having this issue, but only occasionally. As you can see, on my machine I have to do 1000s of runs, on CI unfortunately it happens more often (but not on every run). The fact that we run LTP tests in parallel probably makes it happen more often, too.

Rewriting to asyncio is not such a big deal, I think (it would be the old async function, wrapped into `asyncio.run()`). I'm just not happy about doing this: having to rewrite the function in a completely different style just to apply a workaround. But I will do it if I'm outvoted. :)

---
___
*[python/graminelibos/regression.py, line 35 at r1](https://reviewable.io/reviews/gramineproject/gramine/288#-Mr350Ec2C8cQWCjPuWs:-Mr3CStx9hrA5OGSuz0D:blo1cib) ([raw file](https://github.com/gramineproject/gramine/blob/facde8e7f101f880923ebe204218f97f60ed1dfb/python/graminelibos/regression.py#L35)):*

Is this correct? If the process timesout, then it's still alive and this will not set the return code. I think you should move it after `os.killpg` (or before `if timed_out:` below).
</blockquote></details>

`proc.communicate()` times out not because the main process is still alive, but because stdout/stderr is not closed. So, if `proc.communicate()` timed out, the returncode here is either None (if the main process is still alive) or a number (if the main process exited, but other processes are still alive).

After `os.killpg()` everything should be dead, so we no longer will be able to distinguish these two cases.

boryspoplawski (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/288)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/288#-Mr33g5F4MMDwtRa4kS1:-Mr3TmLu-jkmjpwp0AO6:ba33c8r) (no related file):*

Multiple LTP tests are having this issue, but only occasionally. As you can see, on my machine I have to do 1000s of runs, on CI unfortunately it happens more often (but not on every run). The fact that we run LTP tests in parallel probably makes it happen more often, too.

Rewriting to asyncio is not such a big deal, I think (it would be the old async function, wrapped into `asyncio.run()`). I'm just not happy about doing this: having to rewrite the function in a completely different style just to apply a workaround. But I will do it if I'm outvoted. :)
</blockquote></details>

Ah, forgot we need to lose a race for it to deadlock, so it's not that bad.
Anyway, I'm fine with the current approach.

---
___
*[python/graminelibos/regression.py, line 82 at r1](https://reviewable.io/reviews/gramineproject/gramine/288#-Mr3Ugso4YOyajfZV2fE:-Mr3Ugso4YOyajfZV2fF:b4qtnf4) ([raw file](https://github.com/gramineproject/gramine/blob/facde8e7f101f880923ebe204218f97f60ed1dfb/python/graminelibos/regression.py#L82)):*
> ```Python
>                 raise AssertionError('Command {} timed out after {} s'.format(cmd, timeout))
> 
>         assert main_returncode is not None
> ```

Hm, maybe we should throw an exception here (instead of an assert)? I think it could happen if process closes all fds but does not exit (which is a failed test, I suppose).

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/288)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

___
*[python/graminelibos/regression.py, line 82 at r1](https://reviewable.io/reviews/gramineproject/gramine/288#-Mr3Ugso4YOyajfZV2fE:-Mr6fw1vCvXB2ojDojlp:bkajve4) ([raw file](https://github.com/gramineproject/gramine/blob/facde8e7f101f880923ebe204218f97f60ed1dfb/python/graminelibos/regression.py#L82)):*

Hm, maybe we should throw an exception here (instead of an assert)? I think it could happen if process closes all fds but does not exit (which is a failed test, I suppose).
</blockquote></details>

`communicate` [waits for the process to terminate](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate) ([source code](https://github.com/python/cpython/blob/3.10/Lib/subprocess.py#L1105)).

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/288)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/288#-Mr33g5F4MMDwtRa4kS1:-Mr6soteAHCSYuG519P7:bevk8c0) (no related file):*

Ah, forgot we need to lose a race for it to deadlock, so it's not that bad.
Anyway, I'm fine with the current approach.
</blockquote></details>

I'm fine with the current workaround.

---

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/288)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/288)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #282, opened 2021-12-15
Take `g_dcache_lock` in dentry-related functions

pwmarcz:
## Description of the changes 

This commit makes holding `g_dcache_lock` mandatory when calling `path_lookupat`, `dentry_open` and dentry callbacks (`shim_d_ops`). The callbacks are now also better documented.

This is in preparation for removing the per-object dentry lock (`shim_dentry.lock`), and later, switching to inodes. It should also make the filesystem functions less prone to race conditions.

Context: #279 (inodes migration). Should also help with https://github.com/gramineproject/graphene/issues/944 (but since I touch only some of the filesystem code, I won't claim it's fixed yet).

## How to test this PR? 

I spent some time making sure the callbacks are called according to documentation; I would be grateful for a double-check.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/282)

PR #297, opened 2021-12-22
Convert `tmpfs` to use inodes

pwmarcz:
## Description of the changes 

Next step in inodes conversion (#279). Based on open PR #282.

- Use `g_dcache_lock` for the dentry inode field
- Take `g_dcache_lock` for the whole checkpointing process
- Convert `tmpfs` to use inodes (fixes unlinking, renaming, and corner cases when migrating handles)
- Support `inode->data` (with deallocation and migration)
- Factor out some generic callbacks (seek, stat, hstat, poll)

## How to test this PR? 

CI should be enough. Some tests are now enabled for tmpfs (`send_handle`, `rename_unlink`).

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/297)

### DIFF
--- a/LibOS/shim/include/shim_fs.h
+++ b/LibOS/shim/include/shim_fs.h
@@ -158,7 +158,7 @@ struct shim_dentry {
     void* data;
 
     /* Inode associated with this dentry. Currently optional, and only for the use of underlying
-     * filesystem (see `shim_inode` below). Protected by `lock`. */
+     * filesystem (see `shim_inode` below). Protected by `g_dcache_lock`. */
     struct shim_inode* inode;
 
     /* File lock information, stored only in the main process. Managed by `shim_fs_lock.c`. */
--- a/LibOS/shim/include/shim_fs.h
+++ b/LibOS/shim/include/shim_fs.h
@@ -204,6 +204,9 @@ struct shim_inode {
      * different in case of special files (such as named pipes or sockets). */
     struct shim_fs* fs;
 
+    /* Filesystem-specific data */
+    void* data;
+
     struct shim_lock lock;
     REFTYPE ref_count;
 };
@@ -355,6 +358,42 @@ struct shim_d_ops {
      * The caller should hold `g_dcache_lock`.
      */
     int (*readdir)(struct shim_dentry* dent, readdir_callback_t callback, void* arg);
+
+    /*!
+     * \brief Deallocate inode data
+     *
+     * \param inode the inode about to be freed
+     *
+     * Deallocates any custom data stored in the inode by filesystem. Called before deleting the
+     * inode.
+     *
+     * The caller should hold `inode->lock`.
+     */
+    void (*idrop)(struct shim_inode* inode);
+
+    /*!
+     * \brief Checkpoint inode data
+     *
+     * \param inode the inode to be checkpointed
+     * \param out_data on success, contains a newly allocated buffer with data
+     * \param out_size on success, contains data size
+     *
+     * Prepares any custom data necessary to migrate the inode. Called when checkpointing the inode.
+     *
+     * The caller should hold `inode->lock`. On success, the caller should free `*out_data`.
+     */
+    int (*icheckpoint)(struct shim_inode* inode, void** out_data, size_t* out_size);
+
+    /*!
+     * \brief Restore inode data from checkpoint
+     *
+     * \param inode newly restored inode
+     * \param data checkpoint data (prepared by the `icheckpoint` callback)
+     *
+     * Restores custom state of the inode. Called when restoring the inode from checkpoint, after
+     * all other fields are set.
+     */
+    int (*irestore)(struct shim_inode* inode, void* data);
 };
 
 /*
@@ -648,9 +687,6 @@ void get_dentry(struct shim_dentry* dent);
 /* Decrement the reference count on dent */
 void put_dentry(struct shim_dentry* dent);
 
-void lock_two_dentries(struct shim_dentry* dent1, struct shim_dentry* dent2);
-void unlock_two_dentries(struct shim_dentry* dent1, struct shim_dentry* dent2);
-
 /*!
  * \brief Get the dentry one level up
  *
@@ -841,5 +877,10 @@ struct shim_fs* find_fs(const char* name);
  */
 int generic_seek(file_off_t pos, file_off_t size, file_off_t offset, int origin,
                  file_off_t* out_pos);
+
+int generic_inode_stat(struct shim_dentry* dent, struct stat* buf);
+int generic_inode_hstat(struct shim_handle* hdl, struct stat* buf);
+file_off_t generic_inode_seek(struct shim_handle* hdl, file_off_t offset, int origin);
+int generic_inode_poll(struct shim_handle* hdl, int poll_type);
 
 #endif /* _SHIM_FS_H_ */
--- a/LibOS/shim/include/shim_handle.h
+++ b/LibOS/shim/include/shim_handle.h
@@ -179,6 +179,9 @@ struct shim_handle {
      * filesystem (see `shim_inode` in `shim_fs.h`). Eventually, should replace `dentry` fields.
      *
      * This field does not change, so reading it does not require holding `lock`.
+     *
+     * When taking locks for both handle and inode (`hdl->lock` and `hdl->inode->lock`), you should
+     * lock the *inode* first.
      */
     struct shim_inode* inode;
 
--- a/LibOS/shim/src/bookkeep/shim_handle.c
+++ b/LibOS/shim/src/bookkeep/shim_handle.c
@@ -732,10 +732,6 @@ BEGIN_CP_FUNC(handle) {
             DO_CP_MEMBER(dentry, hdl, new_hdl, dentry);
         }
 
-        if (hdl->inode) {
-            DO_CP_MEMBER(inode, hdl, new_hdl, inode);
-        }
-
         if (new_hdl->pal_handle) {
             struct shim_palhdl_entry* entry;
             DO_CP(palhdl_ptr, &hdl->pal_handle, &entry);
@@ -767,6 +763,12 @@ BEGIN_CP_FUNC(handle) {
         }
 
         unlock(&hdl->lock);
+        if (hdl->inode) {
+            /* NOTE: Checkpointing `inode` will take `inode->lock`, so we need to do it after
+             * `hdl->lock` is released. */
+            DO_CP_MEMBER(inode, hdl, new_hdl, inode);
+        }
+
         ADD_CP_FUNC_ENTRY(off);
     } else {
         new_hdl = (struct shim_handle*)(base + off);
--- a/LibOS/shim/src/fs/chroot/fs.c
+++ b/LibOS/shim/src/fs/chroot/fs.c
@@ -314,42 +314,6 @@ static int chroot_mkdir(struct shim_dentry* dent, mode_t perm) {
     return chroot_setup_dentry(dent, type, perm, /*size=*/0);
 }
 
-static int chroot_istat(struct shim_inode* inode, struct stat* buf) {
-    memset(buf, 0, sizeof(*buf));
-
-    lock(&inode->lock);
-    buf->st_mode = inode->type | inode->perm;
-    buf->st_size = inode->size;
-    buf->st_dev = hash_str(inode->mount->uri);
-    /*
-     * Pretend `nlink` is 2 for directories (to account for "." and ".."), 1 for other files.
-     *
-     * Applications are unlikely to depend on exact value of `nlink`, and for us, it's inconvenient
-     * to keep track of the exact value (we would have to list the directory, and also take into
-     * account synthetic files created by Gramine, such as named pipes and sockets).
-     *
-     * TODO: Make this a default for filesystems that don't provide `nlink`?
-     */
-    buf->st_nlink = (inode->type == FILE_DIR ? 2 : 1);
-    unlock(&inode->lock);
-    return 0;
-}
-
-static int chroot_stat(struct shim_dentry* dent, struct stat* buf) {
-    assert(locked(&g_dcache_lock));
-
-    if (!dent->inode)
-        return -ENOENT;
-
-    return chroot_istat(dent->inode, buf);
-}
-
-static int chroot_hstat(struct shim_handle* hdl, struct stat* buf) {
-    assert(hdl->type == TYPE_CHROOT);
-
-    return chroot_istat(hdl->inode, buf);
-}
-
 static int chroot_flush(struct shim_handle* hdl) {
     assert(hdl->type == TYPE_CHROOT);
 
@@ -403,6 +367,7 @@ static ssize_t chroot_write(struct shim_handle* hdl, const void* buf, size_t cou
         return -EFBIG;
 
     struct shim_inode* inode = hdl->inode;
+    lock(&inode->lock);
     lock(&hdl->lock);
 
     file_off_t pos = hdl->pos;
@@ -426,15 +391,14 @@ static ssize_t chroot_write(struct shim_handle* hdl, const void* buf, size_t cou
         hdl->pos = pos;
 
         /* Update file size if we just wrote past the end of file */
-        lock(&inode->lock);
         if (inode->size < pos)
             inode->size = pos;
-        unlock(&inode->lock);
     }
     ret = actual_count;
 
 out:
     unlock(&hdl->lock);
+    unlock(&inode->lock);
     return ret;
 }
 
@@ -449,30 +413,6 @@ static int chroot_mmap(struct shim_handle* hdl, void** addr, size_t size, int pr
 
     int ret = DkStreamMap(hdl->pal_handle, addr, pal_prot, offset, size);
     return pal_to_unix_errno(ret);
-}
-
-/* TODO: this function emulates lseek() completely inside the LibOS, but some device files may
- * report size == 0 during fstat() and may provide device-specific lseek() logic; this emulation
- * breaks for such device-specific cases */
-static file_off_t chroot_seek(struct shim_handle* hdl, file_off_t offset, int whence) {
-    assert(hdl->type == TYPE_CHROOT);
-
-    file_off_t ret;
-
-    lock(&hdl->lock);
-    file_off_t pos = hdl->pos;
-
-    lock(&hdl->inode->lock);
-    file_off_t size = hdl->inode->size;
-    unlock(&hdl->inode->lock);
-
-    ret = generic_seek(pos, size, offset, whence, &pos);
-    if (ret == 0) {
-        hdl->pos = pos;
-        ret = pos;
-    }
-    unlock(&hdl->lock);
-    return ret;
 }
 
 static int chroot_truncate(struct shim_handle* hdl, file_off_t size) {
@@ -570,29 +510,6 @@ static int chroot_unlink(struct shim_dentry* dent) {
     dent->inode = NULL;
     put_inode(inode);
     return 0;
-}
-
-static int chroot_poll(struct shim_handle* hdl, int poll_type) {
-    assert(hdl->type == TYPE_CHROOT);
-
-    int ret;
-
-    lock(&hdl->lock);
-    lock(&hdl->inode->lock);
-
-    if (hdl->inode->type == S_IFREG) {
-        ret = 0;
-        if (poll_type & FS_POLL_WR)
-            ret |= FS_POLL_WR;
-        if ((poll_type & FS_POLL_RD) && hdl->pos < hdl->inode->size)
-            ret |= FS_POLL_RD;
-    } else {
-        ret = -EAGAIN;
-    }
-
-    unlock(&hdl->inode->lock);
-    unlock(&hdl->lock);
-    return ret;
 }
 
 static int chroot_rename(struct shim_dentry* old, struct shim_dentry* new) {
@@ -741,10 +658,13 @@ struct shim_fs_ops chroot_fs_ops = {
     .read       = &chroot_read,
     .write      = &chroot_write,
     .mmap       = &chroot_mmap,
-    .seek       = &chroot_seek,
-    .hstat      = &chroot_hstat,
+    /* TODO: this function emulates lseek() completely inside the LibOS, but some device files may
+     * report size == 0 during fstat() and may provide device-specific lseek() logic; this emulation
+     * breaks for such device-specific cases */
+    .seek       = &generic_inode_seek,
+    .hstat      = &generic_inode_hstat,
     .truncate   = &chroot_truncate,
-    .poll       = &chroot_poll,
+    .poll       = &generic_inode_poll,
     .checkout   = &chroot_checkout,
     .checkin    = &chroot_checkin,
 };
@@ -754,7 +674,7 @@ struct shim_d_ops chroot_d_ops = {
     .lookup  = &chroot_lookup,
     .creat   = &chroot_creat,
     .mkdir   = &chroot_mkdir,
-    .stat    = &chroot_stat,
+    .stat    = &generic_inode_stat,
     .readdir = &chroot_readdir,
     .unlink  = &chroot_unlink,
     .rename  = &chroot_rename,
--- a/LibOS/shim/src/fs/chroot/fs.c
+++ b/LibOS/shim/src/fs/chroot/fs.c
@@ -135,7 +135,7 @@ out:
 
 static int chroot_setup_dentry(struct shim_dentry* dent, mode_t type, mode_t perm,
                                file_off_t size) {
-    assert(locked(&dent->lock));
+    assert(locked(&g_dcache_lock));
     assert(!dent->inode);
 
     dent->type = type;
@@ -226,7 +226,7 @@ static int chroot_temp_open(struct shim_dentry* dent, mode_t type, PAL_HANDLE* o
 /* Open a PAL handle, and associate it with a LibOS handle (if provided). */
 static int chroot_do_open(struct shim_handle* hdl, struct shim_dentry* dent, mode_t type,
                           int flags, mode_t perm) {
-    assert(locked(&dent->lock));
+    assert(locked(&g_dcache_lock));
 
     int ret;
 
@@ -264,25 +264,20 @@ out:
 }
 
 static int chroot_open(struct shim_handle* hdl, struct shim_dentry* dent, int flags) {
+    assert(locked(&g_dcache_lock));
+
     int ret;
 
-    lock(&dent->lock);
-    if (!dent->inode) {
-        ret = -ENOENT;
-        goto out;
-    }
+    if (!dent->inode)
+        return -ENOENT;
 
     ret = chroot_do_open(hdl, dent, dent->type, flags, /*perm=*/0);
     if (ret < 0)
-        goto out;
+        return ret;
 
     hdl->inode = dent->inode;
     get_inode(dent->inode);
-    ret = 0;
-
-out:
-    unlock(&dent->lock);
-    return ret;
+    return 0;
 }
 
 static int chroot_creat(struct shim_handle* hdl, struct shim_dentry* dent, int flags, mode_t perm) {
@@ -292,22 +287,17 @@ static int chroot_creat(struct shim_handle* hdl, struct shim_dentry* dent, int f
 
     mode_t type = S_IFREG;
 
-    lock(&dent->lock);
     ret = chroot_do_open(hdl, dent, type, flags | O_CREAT | O_EXCL, perm);
     if (ret < 0)
-        goto out;
+        return ret;
 
     ret = chroot_setup_dentry(dent, type, perm, /*size=*/0);
     if (ret < 0)
-        goto out;
+        return ret;
 
     hdl->inode = dent->inode;
     get_inode(dent->inode);
-    ret = 0;
-
-out:
-    unlock(&dent->lock);
-    return ret;
+    return 0;
 }
 
 static int chroot_mkdir(struct shim_dentry* dent, mode_t perm) {
@@ -317,15 +307,11 @@ static int chroot_mkdir(struct shim_dentry* dent, mode_t perm) {
 
     mode_t type = S_IFDIR;
 
-    lock(&dent->lock);
     ret = chroot_do_open(/*hdl=*/NULL, dent, type, O_CREAT | O_EXCL, perm);
     if (ret < 0)
-        goto out;
+        return ret;
 
-    ret = chroot_setup_dentry(dent, type, perm, /*size=*/0);
-out:
-    unlock(&dent->lock);
-    return ret;
+    return chroot_setup_dentry(dent, type, perm, /*size=*/0);
 }
 
 static int chroot_istat(struct shim_inode* inode, struct stat* buf) {
@@ -350,19 +336,12 @@ static int chroot_istat(struct shim_inode* inode, struct stat* buf) {
 }
 
 static int chroot_stat(struct shim_dentry* dent, struct stat* buf) {
-    int ret;
+    assert(locked(&g_dcache_lock));
 
-    lock(&dent->lock);
-    if (!dent->inode) {
-        ret = -ENOENT;
-        goto out;
-    }
+    if (!dent->inode)
+        return -ENOENT;
 
-    ret = chroot_istat(dent->inode, buf);
-
-out:
-    unlock(&dent->lock);
-    return ret;
+    return chroot_istat(dent->inode, buf);
 }
 
 static int chroot_hstat(struct shim_handle* hdl, struct stat* buf) {
@@ -573,29 +552,24 @@ out:
 }
 
 static int chroot_unlink(struct shim_dentry* dent) {
-    int ret;
+    assert(locked(&g_dcache_lock));
 
-    lock(&dent->lock);
+    int ret;
 
     PAL_HANDLE palhdl;
     ret = chroot_temp_open(dent, dent->type, &palhdl);
     if (ret < 0)
-        goto out;
+        return ret;
 
     ret = DkStreamDelete(palhdl, PAL_DELETE_ALL);
     DkObjectClose(palhdl);
-    if (ret < 0) {
-        ret = pal_to_unix_errno(ret);
-        goto out;
-    }
+    if (ret < 0)
+        return pal_to_unix_errno(ret);
 
     struct shim_inode* inode = dent->inode;
     dent->inode = NULL;
     put_inode(inode);
-
-out:
-    unlock(&dent->lock);
-    return ret;
+    return 0;
 }
 
 static int chroot_poll(struct shim_handle* hdl, int poll_type) {
@@ -622,10 +596,10 @@ static int chroot_poll(struct shim_handle* hdl, int poll_type) {
 }
 
 static int chroot_rename(struct shim_dentry* old, struct shim_dentry* new) {
+    assert(locked(&g_dcache_lock));
+
     int ret;
     char* new_uri = NULL;
-
-    lock_two_dentries(old, new);
 
     ret = chroot_dentry_uri(new, old->type, &new_uri);
     if (ret < 0)
@@ -659,15 +633,15 @@ static int chroot_rename(struct shim_dentry* old, struct shim_dentry* new) {
     ret = 0;
 
 out:
-    unlock_two_dentries(old, new);
     free(new_uri);
     return ret;
 }
 
 static int chroot_chmod(struct shim_dentry* dent, mode_t perm) {
+    assert(locked(&g_dcache_lock));
+
     int ret;
 
-    lock(&dent->lock);
     lock(&dent->inode->lock);
 
     PAL_HANDLE palhdl;
@@ -690,7 +664,6 @@ static int chroot_chmod(struct shim_dentry* dent, mode_t perm) {
 
 out:
     unlock(&dent->inode->lock);
-    unlock(&dent->lock);
     return ret;
 }
 
@@ -721,18 +694,19 @@ static int chroot_checkout(struct shim_handle* hdl) {
     assert(hdl->type == TYPE_CHROOT);
     assert(hdl->pal_handle);
 
+    /* We should be holding `g_dcache_lock` for the whole checkpointing process. */
+    assert(locked(&g_dcache_lock));
+
     /* We don't take `hdl->lock` because this is actually the handle *copied* for checkpointing (and
      * the lock isn't even properly initialized). */
 
     /* First, check if we have not deleted or renamed the file (the dentry contains the same
      * inode). */
-    lock(&hdl->dentry->lock);
     bool is_in_dentry = (hdl->dentry->inode == hdl->inode);
-    unlock(&hdl->dentry->lock);
 
-    /* Then check if the file still exists on host. If so, we assume it can be opened by the child
-     * process, so the PAL handle doesn't need sending. */
     if (is_in_dentry) {
+        /* Then check if the file still exists on host. If so, we assume it can be opened by the
+         * child process, so the PAL handle doesn't need sending. */
         PAL_STREAM_ATTR attr;
         if (DkStreamAttributesQuery(hdl->uri, &attr) == 0) {
             hdl->pal_handle = NULL;
--- a/LibOS/shim/src/fs/shim_dcache.c
+++ b/LibOS/shim/src/fs/shim_dcache.c
@@ -536,6 +536,9 @@ BEGIN_CP_FUNC(dentry) {
     __UNUSED(size);
     assert(size == sizeof(struct shim_dentry));
 
+    /* We should be holding `g_dcache_lock` for the whole checkpointing process. */
+    assert(locked(&g_dcache_lock));
+
     struct shim_dentry* dent     = (struct shim_dentry*)obj;
     struct shim_dentry* new_dent = NULL;
 
--- a/LibOS/shim/src/fs/shim_dcache.c
+++ b/LibOS/shim/src/fs/shim_dcache.c
@@ -354,28 +354,6 @@ int dentry_rel_path(struct shim_dentry* dent, char** path, size_t* size) {
     return dentry_path(dent, /*relative=*/true, path, size);
 }
 
-void lock_two_dentries(struct shim_dentry* dent1, struct shim_dentry* dent2) {
-    assert(dent1 != dent2);
-    if ((uintptr_t)dent1 < (uintptr_t)dent2) {
-        lock(&dent1->lock);
-        lock(&dent2->lock);
-    } else {
-        lock(&dent2->lock);
-        lock(&dent1->lock);
-    }
-}
-
-void unlock_two_dentries(struct shim_dentry* dent1, struct shim_dentry* dent2) {
-    assert(dent1 != dent2);
-    if ((uintptr_t)dent1 < (uintptr_t)dent2) {
-        unlock(&dent2->lock);
-        unlock(&dent1->lock);
-    } else {
-        unlock(&dent1->lock);
-        unlock(&dent2->lock);
-    }
-}
-
 struct shim_inode* get_new_inode(struct shim_mount* mount, mode_t type, mode_t perm) {
     assert(mount);
 
@@ -399,6 +377,8 @@ struct shim_inode* get_new_inode(struct shim_mount* mount, mode_t type, mode_t p
     get_mount(mount);
     inode->fs = mount->fs;
 
+    inode->data = NULL;
+
     REF_SET(inode->ref_count, 1);
     return inode;
 }
@@ -409,6 +389,12 @@ void get_inode(struct shim_inode* inode) {
 
 void put_inode(struct shim_inode* inode) {
     if (REF_DEC(inode->ref_count) == 0) {
+        if (inode->fs->d_ops && inode->fs->d_ops->idrop) {
+            lock(&inode->lock);
+            inode->fs->d_ops->idrop(inode);
+            unlock(&inode->lock);
+        }
+
         put_mount(inode->mount);
 
         destroy_lock(&inode->lock);
@@ -667,6 +653,22 @@ BEGIN_CP_FUNC(inode) {
 
         REF_SET(new_inode->ref_count, 0);
 
+        if (inode->fs->d_ops && inode->fs->d_ops->icheckpoint) {
+            void* cp_data;
+            size_t cp_size;
+            int ret = inode->fs->d_ops->icheckpoint(inode, &cp_data, &cp_size);
+            if (ret < 0)
+                return ret;
+
+            size_t cp_off = ADD_CP_OFFSET(cp_size);
+            new_inode->data = (char*)base + cp_off;
+            memcpy(new_inode->data, cp_data, cp_size);
+            free(cp_data);
+        } else {
+            assert(!inode->data);
+            new_inode->data = NULL;
+        }
+
         unlock(&inode->lock);
 
         ADD_CP_FUNC_ENTRY(off);
@@ -690,6 +692,19 @@ BEGIN_RS_FUNC(inode) {
 
     if (!create_lock(&inode->lock)) {
         return -ENOMEM;
+    }
+
+    if (inode->fs->d_ops && inode->fs->d_ops->irestore) {
+        assert(inode->data);
+        CP_REBASE(inode->data);
+        void* cp_data = inode->data;
+        inode->data = NULL;
+
+        int ret = inode->fs->d_ops->irestore(inode, cp_data);
+        if (ret < 0)
+            return ret;
+    } else {
+        assert(!inode->data);
     }
 }
 END_RS_FUNC(inode)
--- a/LibOS/shim/src/fs/shim_fs_util.c
+++ b/LibOS/shim/src/fs/shim_fs_util.c
@@ -6,6 +6,8 @@
 #include <stdint.h>
 
 #include "shim_fs.h"
+#include "shim_lock.h"
+#include "stat.h"
 
 int generic_seek(file_off_t pos, file_off_t size, file_off_t offset, int origin,
                  file_off_t* out_pos) {
@@ -36,4 +38,90 @@ int generic_seek(file_off_t pos, file_off_t size, file_off_t offset, int origin,
 
     *out_pos = pos;
     return 0;
+}
+
+static int generic_istat(struct shim_inode* inode, struct stat* buf) {
+    memset(buf, 0, sizeof(*buf));
+
+    lock(&inode->lock);
+    buf->st_mode = inode->type | inode->perm;
+    buf->st_size = inode->size;
+    /*
+     * Pretend `nlink` is 2 for directories (to account for "." and ".."), 1 for other files.
+     *
+     * Applications are unlikely to depend on exact value of `nlink`, and for us, it's inconvenient
+     * to keep track of the exact value (we would have to list the directory, and also take into
+     * account synthetic files created by Graphene, such as named pipes and sockets).
+     */
+    buf->st_nlink = (inode->type == S_IFDIR ? 2 : 1);
+
+    if (inode->mount->uri)
+        buf->st_dev = hash_str(inode->mount->uri);
+
+    unlock(&inode->lock);
+    return 0;
+}
+
+int generic_inode_stat(struct shim_dentry* dent, struct stat* buf) {
+    assert(locked(&g_dcache_lock));
+    if (!dent->inode) {
+        /*
+         * TODO: This can happen for synthetic dentries (DENTRY_SYNTHETIC) created as intermediate
+         * directories to a mountpoint. These should be treated specially, same as other
+         * Gramine-internal files (pipes, sockets). For now, we fail `stat()` with -ENOENT.
+         */
+        assert(dent->state & DENTRY_SYNTHETIC);
+        log_debug("%s: disregarding synthetic dentry: %s", __func__, dent->name);
+        return -ENOENT;
+    }
+
+    return generic_istat(dent->inode, buf);
+}
+
+int generic_inode_hstat(struct shim_handle* hdl, struct stat* buf) {
+    assert(hdl->inode);
+
+    return generic_istat(hdl->inode, buf);
+}
+
+file_off_t generic_inode_seek(struct shim_handle* hdl, file_off_t offset, int origin) {
+    file_off_t ret;
+
+    lock(&hdl->inode->lock);
+    lock(&hdl->lock);
+    file_off_t pos = hdl->pos;
+    file_off_t size = hdl->inode->size;
+
+    ret = generic_seek(pos, size, offset, origin, &pos);
+    if (ret == 0) {
+        hdl->pos = pos;
+        ret = pos;
+    }
+    unlock(&hdl->lock);
+    unlock(&hdl->inode->lock);
+    return ret;
+}
+
+int generic_inode_poll(struct shim_handle* hdl, int poll_type) {
+    int ret;
+
+    lock(&hdl->inode->lock);
+    lock(&hdl->lock);
+
+    if (hdl->inode->type == S_IFREG) {
+        ret = 0;
+        if (poll_type & FS_POLL_WR)
+            ret |= FS_POLL_WR;
+        /* TODO: The `hdl->pos < hdl->inode->size` condition is wrong, the `poll` syscall treats
+         * end-of-file as readable. Check if removing this condition doesn't break anything
+         * in our `poll` implementation. */
+        if ((poll_type & FS_POLL_RD) && hdl->pos < hdl->inode->size)
+            ret |= FS_POLL_RD;
+    } else {
+        ret = -EAGAIN;
+    }
+
+    unlock(&hdl->lock);
+    unlock(&hdl->inode->lock);
+    return ret;
 }
--- a/LibOS/shim/src/fs/tmpfs/fs.c
+++ b/LibOS/shim/src/fs/tmpfs/fs.c
@@ -6,10 +6,12 @@
 
 /*
  * This file contains code for implementation of 'tmpfs' filesystem. The tmpfs files are *not*
- * cloned during fork/clone and cannot be synchronized between processes.
+ * cloned during fork/clone (except for files currently open) and cannot be synchronized between
+ * processes.
  *
- * The tmpfs files are directly represented by their dentries (i.e. a file exists whenever
- * corresponding dentry exists). The file data is stored in the dentries.
+ * The tmpfs files are directly represented by their dentries and inodes (i.e. a file exists
+ * whenever corresponding dentry exists, and is associated with inode). The file data is stored in
+ * the `data` field of the inode (as a pointer to `struct shim_mem_file`).
  */
 
 #include <asm/mman.h>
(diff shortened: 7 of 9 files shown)


pwmarcz:
Jenkins, retest Jenkins-Direct-Sanitizers please (`sendfile04` timed out, known issue, fixed on `master`, but this PR is not yet based on `master`)

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/297)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 84 at r1](https://reviewable.io/reviews/gramineproject/gramine/297#-MruxArTDANZQARJlA5X:-MruxArTDANZQARJlA5Y:b5ysca8) ([raw file](https://github.com/gramineproject/gramine/blob/3fb9cf5c6069af94c78932fff78692c329788dac/LibOS/shim/src/fs/tmpfs/fs.c#L84)):*
> ```C
>     struct tmpfs_checkpoint* cp;
>     size_t cp_size = sizeof(*cp) + mem->size;
>     cp = malloc(cp_size);
> ```

Why do we have this level of indirection? Why couldn't we have a pointer to the memory region, where the tmpfs file content is located? (These memory regions -- VMAs -- are checkpointed and sent to the child anyway, and are located at exactly the same addresses in the child as in the parent.)

I'm afraid that this "copy the contents of the tmpfs file into the checkpoint blob, and then restore into child's memory from the checkpoint blob" will explode on large tmpfs files. I can easily imagine some database engine creating temporary files of 10GB in size, and copying such a file during fork will most probably fail (we cap the size of the checkpoint blob).

Or maybe I misunderstand the code? I actually thought that we had FS tests for such cases (copying of pretty-big files during fork), and I would expect them to fail in our CI.
___
*[LibOS/shim/src/sys/shim_clone.c, line 129 at r1](https://reviewable.io/reviews/gramineproject/gramine/297#-MruzhIs81_Rh7gwghk0:-MruzhIs81_Rh7gwghk1:b-gx66az) ([raw file](https://github.com/gramineproject/gramine/blob/3fb9cf5c6069af94c78932fff78692c329788dac/LibOS/shim/src/sys/shim_clone.c#L129)):*
> ```C
>                         struct shim_ipc_ids* process_ipc_ids, va_list ap) {
>     __UNUSED(ap);
>     lock(&g_dcache_lock);
> ```

It's a bit weird to grab the dcache lock here -- we will hold the lock even when we checkpoint unrelated objects (like VMAs). There is no way to grab this lock at some root-of-dcache object?

Alternatively, please put a comment that "for simplicity, we grab the dcache lock here, even though it will cover unrelated stuff".

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/297)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 84 at r1](https://reviewable.io/reviews/gramineproject/gramine/297#-MruxArTDANZQARJlA5X:-MsFg43I63UqlLkfThBl:b-pc6clq) ([raw file](https://github.com/gramineproject/gramine/blob/3fb9cf5c6069af94c78932fff78692c329788dac/LibOS/shim/src/fs/tmpfs/fs.c#L84)):*
> Why do we have this level of indirection? Why couldn't we have a pointer to the memory region, where the tmpfs file content is located? (These memory regions -- VMAs -- are checkpointed and sent to the child anyway, and are located at exactly the same addresses in the child as in the parent.)

Wow, looks like we do that, but I'm very surprised to see this. I assumed we only copy *application*'s data, not internal LibOS memory, and that's why we need the whole checkpointing process which copies all the data.

Do we actually rely on the fact we copy all VMAs? I haven't seen anything like this in the checkpointing code. Will anything break if I exclude internal VMAs from checkpointing?

Mostly I'm just really confused. Using the memory from the checkpoint blob directly (which we do right now) already sounded like a pretty bad design, but now I hear that we also transfer the *original memory* containing the objects that we just copied to the checkpoint blob, and that sounds really broken (or at least wasteful).

As for the solution you propose:

1. While we could use this memory directly, probably bad things will happen if we `free()` it, right? It is my understanding that `free()` is a no-op only when called on memory inside the checkpoint blob, not inside these other copied VMAs.

2. If I understand correctly, the 10GB-temporary-file scenario should fail even before any of this, because we will be checkpointing the VMA that contains it, so we will try to copy it to the checkpoint blob, and that will fail. Is that right?

>Or maybe I misunderstand the code? I actually thought that we had FS tests for such cases (copying of pretty-big files during fork), and I would expect them to fail in our CI.

Not sure what tests you have in mind? If there are any, they're probably still skipped for tmpfs.
___
*[LibOS/shim/src/sys/shim_clone.c, line 129 at r1](https://reviewable.io/reviews/gramineproject/gramine/297#-MruzhIs81_Rh7gwghk0:-MsFlhF44hKO6X7N-EGL:b940vjq) ([raw file](https://github.com/gramineproject/gramine/blob/3fb9cf5c6069af94c78932fff78692c329788dac/LibOS/shim/src/sys/shim_clone.c#L129)):*
>It's a bit weird to grab the dcache lock here -- we will hold the lock even when we checkpoint unrelated objects (like VMAs). There is no way to grab this lock at some root-of-dcache object?

While we checkpoint root of dcache, currently we don't checkpoint children (I'm not sure why, maybe there were some weird cyclic dependencies?) But we checkpoint dentries for open handles, and, recursively, all their parents. And, separately, the mount objects that also point to root dentries.

My point is, the checkpoint procedure isn't neatly split between object types. Checkpointing an object of one type can pull in other objects (e.g. VMA -> file handle -> dentry), as long as they haven't been checkpointed before. So even if I could claim that `g_dcache_lock` is needed only for a part of this operation, I think it would be pretty brittle.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/297)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski)

___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 84 at r1](https://reviewable.io/reviews/gramineproject/gramine/297#-MruxArTDANZQARJlA5X:-MsTzIJq5q929w16t5jw:bs5919n) ([raw file](https://github.com/gramineproject/gramine/blob/3fb9cf5c6069af94c78932fff78692c329788dac/LibOS/shim/src/fs/tmpfs/fs.c#L84)):*
Just a note: I don't think any one of us understands the design and implementation of Gramine checkpointing completely. It is probably the only part of Gramine that was not touched or deeply analyzed by any active developer.

> If I understand correctly, the 10GB-temporary-file scenario should fail even before any of this, because we will be checkpointing the VMA that contains it, so we will try to copy it to the checkpoint blob, and that will fail. Is that right?

No, VMAs are not copied into the checkpoint blob. They are transferred to the child process separately: https://github.com/gramineproject/gramine/blob/57e5cd3cd48db24f4bb2e8cd9d36b74d5cb2fa6b/LibOS/shim/src/shim_checkpoint.c#L214

> Wow, looks like we do that, but I'm very surprised to see this. I assumed we only copy application's data, not internal LibOS memory, and that's why we need the whole checkpointing process which copies all the data.

> Do we actually rely on the fact we copy all VMAs? I haven't seen anything like this in the checkpointing code. Will anything break if I exclude internal VMAs from checkpointing?

We need @boryspoplawski here. He re-wrote the VMA subsystem. To be honest, I don't know/remember the details of *checkpointing internal VMAs*.

> Not sure what tests you have in mind? If there are any, they're probably still skipped for tmpfs.

I mean these tests: https://github.com/gramineproject/gramine/blob/master/LibOS/shim/test/fs/test_tmpfs.py

TLDR: If it turns out that checkpointing of internal VMAs is an unwanted side-effect of the current implementation, I'm fine with the current approach. We'll need to fix Gramine checkpointing in a separate series of PRs...
___
*[LibOS/shim/src/sys/shim_clone.c, line 129 at r1](https://reviewable.io/reviews/gramineproject/gramine/297#-MruzhIs81_Rh7gwghk0:-MsU-jzoELgjqJ5fdIE0:btkyv5q) ([raw file](https://github.com/gramineproject/gramine/blob/3fb9cf5c6069af94c78932fff78692c329788dac/LibOS/shim/src/sys/shim_clone.c#L129)):*

>It's a bit weird to grab the dcache lock here -- we will hold the lock even when we checkpoint unrelated objects (like VMAs). There is no way to grab this lock at some root-of-dcache object?

While we checkpoint root of dcache, currently we don't checkpoint children (I'm not sure why, maybe there were some weird cyclic dependencies?) But we checkpoint dentries for open handles, and, recursively, all their parents. And, separately, the mount objects that also point to root dentries.

My point is, the checkpoint procedure isn't neatly split between object types. Checkpointing an object of one type can pull in other objects (e.g. VMA -> file handle -> dentry), as long as they haven't been checkpointed before. So even if I could claim that `g_dcache_lock` is needed only for a part of this operation, I think it would be pretty brittle.
</blockquote></details>

Ah, true, I understand now. Our checkpointing is more of a hodgepodge approach: upon encountering a new LibOS object, checkpoint all its related objects.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/297)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 84 at r1](https://reviewable.io/reviews/gramineproject/gramine/297#-MruxArTDANZQARJlA5X:-MsYBQy91LklP17DnUg4:bapynqd) ([raw file](https://github.com/gramineproject/gramine/blob/3fb9cf5c6069af94c78932fff78692c329788dac/LibOS/shim/src/fs/tmpfs/fs.c#L84)):*

Just a note: I don't think any one of us understands the design and implementation of Gramine checkpointing completely. It is probably the only part of Gramine that was not touched or deeply analyzed by any active developer.

> If I understand correctly, the 10GB-temporary-file scenario should fail even before any of this, because we will be checkpointing the VMA that contains it, so we will try to copy it to the checkpoint blob, and that will fail. Is that right?

No, VMAs are not copied into the checkpoint blob. They are transferred to the child process separately: https://github.com/gramineproject/gramine/blob/57e5cd3cd48db24f4bb2e8cd9d36b74d5cb2fa6b/LibOS/shim/src/shim_checkpoint.c#L214

> Wow, looks like we do that, but I'm very surprised to see this. I assumed we only copy application's data, not internal LibOS memory, and that's why we need the whole checkpointing process which copies all the data.

> Do we actually rely on the fact we copy all VMAs? I haven't seen anything like this in the checkpointing code. Will anything break if I exclude internal VMAs from checkpointing?

We need @boryspoplawski here. He re-wrote the VMA subsystem. To be honest, I don't know/remember the details of *checkpointing internal VMAs*.

> Not sure what tests you have in mind? If there are any, they're probably still skipped for tmpfs.

I mean these tests: https://github.com/gramineproject/gramine/blob/master/LibOS/shim/test/fs/test_tmpfs.py

TLDR: If it turns out that checkpointing of internal VMAs is an unwanted side-effect of the current implementation, I'm fine with the current approach. We'll need to fix Gramine checkpointing in a separate series of PRs...
</blockquote></details>

We do **not** checkpoint internal VMAs.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/297)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 84 at r1](https://reviewable.io/reviews/gramineproject/gramine/297#-MruxArTDANZQARJlA5X:-MsYCK8hBQ70-mnNjbnI:b-p4nah4) ([raw file](https://github.com/gramineproject/gramine/blob/3fb9cf5c6069af94c78932fff78692c329788dac/LibOS/shim/src/fs/tmpfs/fs.c#L84)):*
> Just a note: I don't think any one of us understands the design and implementation of Gramine checkpointing completely. It is probably the only part of Gramine that was not touched or deeply analyzed by any active developer.

I've dug into it some/long time ago, but there were always changes with more priority, so I've never finished it.

FWIW: checkpointing of VMAs and memory in general was mostly left as it was, when I rewrote VMAs subsystem.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/297)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski and @pwmarcz)

___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 84 at r1](https://reviewable.io/reviews/gramineproject/gramine/297#-MruxArTDANZQARJlA5X:-MsZ9Ugr9iep1KxMZp9D:bp3fej2) ([raw file](https://github.com/gramineproject/gramine/blob/3fb9cf5c6069af94c78932fff78692c329788dac/LibOS/shim/src/fs/tmpfs/fs.c#L84)):*

> Just a note: I don't think any one of us understands the design and implementation of Gramine checkpointing completely. It is probably the only part of Gramine that was not touched or deeply analyzed by any active developer.

I've dug into it some/long time ago, but there were always changes with more priority, so I've never finished it.

FWIW: checkpointing of VMAs and memory in general was mostly left as it was, when I rewrote VMAs subsystem.
</blockquote></details>

Indeed, we do not checkpoint internal VMAs, I just looked at the VMA checkpointing procedure. These are the relevant lines: https://github.com/gramineproject/gramine/blob/57e5cd3cd48db24f4bb2e8cd9d36b74d5cb2fa6b/LibOS/shim/src/bookkeep/shim_vma.c#L1121-L1122

@pwmarcz So I guess there is no other way to transfer tmpfs files, other than your current copy-into-checkpoint-blob? If there is no other way, then I'll resolve my comment.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/297)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 84 at r1](https://reviewable.io/reviews/gramineproject/gramine/297#-MruxArTDANZQARJlA5X:-MsZMFRjBwXpeCKuW88K:bfzh9mh) ([raw file](https://github.com/gramineproject/gramine/blob/3fb9cf5c6069af94c78932fff78692c329788dac/LibOS/shim/src/fs/tmpfs/fs.c#L84)):*

Indeed, we do not checkpoint internal VMAs, I just looked at the VMA checkpointing procedure. These are the relevant lines: https://github.com/gramineproject/gramine/blob/57e5cd3cd48db24f4bb2e8cd9d36b74d5cb2fa6b/LibOS/shim/src/bookkeep/shim_vma.c#L1121-L1122

@pwmarcz So I guess there is no other way to transfer tmpfs files, other than your current copy-into-checkpoint-blob? If there is no other way, then I'll resolve my comment.
</blockquote></details>

Memory is checkpointed using https://github.com/gramineproject/gramine/blob/master/LibOS/shim/src/shim_checkpoint.c#L129 so you probably want to use that.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/297)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 84 at r1](https://reviewable.io/reviews/gramineproject/gramine/297#-MruxArTDANZQARJlA5X:-Muk2zGjB-CtFMHy7Sjn:b3reliv) ([raw file](https://github.com/gramineproject/gramine/blob/3fb9cf5c6069af94c78932fff78692c329788dac/LibOS/shim/src/fs/tmpfs/fs.c#L84)):*
OK, my bad, I didn't notice that we skip `VMA_INTERNAL` vmas. Too bad the code consistently claims to copy "all vmas" but that's Gramine for you :)

>Memory is checkpointed using https://github.com/gramineproject/gramine/blob/master/LibOS/shim/src/shim_checkpoint.c#L129 so you probably want to use that.

Correct me if I'm wrong, but it looks like this is for copying a region of memory _to the same address in child_ (the `receive_memory_on_stream` function uses `entry->addr` as-is). So 

1. it won't work for objects allocated by our internal `malloc()`, since we don't migrate the corresponding VMA, and
2. even if we migrated the internal VMA as well, the resulting memory would be managed by the slab allocator (e.g. I cannot `free()` it). So I don't see an easy way of using this.

I think this is a good input for redesigning the checkpointing system: ideally, _all_ serialized data should be received as a stream, and not kept in a "blob received from parent" where it can never be freed. (That would also force us to _process_ all fields, instead of using them directly with some "rebasing" hacks.)

But within the current system, I don't think we have an easy way of migrating `malloc`ed data without copying it?

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/297)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 84 at r1](https://reviewable.io/reviews/gramineproject/gramine/297#-MruxArTDANZQARJlA5X:-MukTbcc6DH9k1xsLypS:bq8hs1t) ([raw file](https://github.com/gramineproject/gramine/blob/3fb9cf5c6069af94c78932fff78692c329788dac/LibOS/shim/src/fs/tmpfs/fs.c#L84)):*
> But within the current system, I don't think we have an easy way of migrating `malloc`ed data without copying it?

Yes, I have the same impression.

> I think this is a good input for redesigning the checkpointing system: ideally, all serialized data should be received as a stream, and not kept in a "blob received from parent" where it can never be freed. (That would also force us to process all fields, instead of using them directly with some "rebasing" hacks.)

Yes, this makes sense. It's much cleaner to re-create the objects (including nested objects) from the stream, and forget about the "checkpoint blob" after the restore process is finished.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/297)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 84 at r1](https://reviewable.io/reviews/gramineproject/gramine/297#-MruxArTDANZQARJlA5X:-MulFMDp2rEwoITpZFFE:b-29yv7h) ([raw file](https://github.com/gramineproject/gramine/blob/3fb9cf5c6069af94c78932fff78692c329788dac/LibOS/shim/src/fs/tmpfs/fs.c#L84)):*
> Too bad the code consistently claims to copy "all vmas" but that's Gramine for you :)

I think we only have `all_vma` as checkpointing function, which might be misleading.

> Correct me if I'm wrong, but it looks like this is for copying a region of memory to the same address in child

As, you're right.

> I think this is a good input for redesigning the checkpointing system: ideally, all serialized data should be received as a stream, and not kept in a "blob received from parent" where it can never be freed. (That would also force us to process all fields, instead of using them directly with some "rebasing" hacks.)

Yes, I think we already had some discussions about this (or maybe I've just talked privately with Michał, don't remember).

> But within the current system, I don't think we have an easy way of migrating malloced data without copying it?

From what I can see you just malloc this buffer so that you can copy data into it in `icheckpoint` and then copy that data into actual checkpoint and free it? Why not copy data directly into checkpoint? It should be doable in the current state, just add a checkpointing function for tmpfs inodes/data.
Disclaimer: I've not read the whole PR, maybe I'm talking nonsense.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/297)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 84 at r1](https://reviewable.io/reviews/gramineproject/gramine/297#-MruxArTDANZQARJlA5X:-MulO5Xo0xSYx4QffixT:b3eb07h) ([raw file](https://github.com/gramineproject/gramine/blob/3fb9cf5c6069af94c78932fff78692c329788dac/LibOS/shim/src/fs/tmpfs/fs.c#L84)):*
> ideally, all serialized data should be received as a stream, and not kept in a "blob received from parent" where it can never be freed

And we could finally have `relloc()` then! :)

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/297)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 84 at r1](https://reviewable.io/reviews/gramineproject/gramine/297#-MruxArTDANZQARJlA5X:-MupeNqZ6SpAkrrBnPR_:b-mlnp45) ([raw file](https://github.com/gramineproject/gramine/blob/3fb9cf5c6069af94c78932fff78692c329788dac/LibOS/shim/src/fs/tmpfs/fs.c#L84)):*
> From what I can see you just malloc this buffer so that you can copy data into it in `icheckpoint` and then copy that data into actual checkpoint and free it? Why not copy data directly into checkpoint? It should be doable in the current state, just add a checkpointing function for tmpfs inodes/data.

I mean, I could define something like `BEGIN_CP_FUNC(tmpfs_inode)` but then I don't know how to call it based on the filesystem type. It sounds like I'd have to store a pointer to it in the filesystem callbacks (e.g. `shim_d_ops`, and I don't like where that leads (for example, both the name and the signature of that function are hidden behind macros...)

So, for the sake of sanity, I wanted to keep `icheckpoint/irestore` as a set of simple callbacks that are completely independent of the current checkpointing system. I hope it's worth it even with the extra copy.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/297)* status: 4 of 7 files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 84 at r1](https://reviewable.io/reviews/gramineproject/gramine/297#-MruxArTDANZQARJlA5X:-MupqFaS0fGDJ6lZWc_z:bz2r3ah) ([raw file](https://github.com/gramineproject/gramine/blob/3fb9cf5c6069af94c78932fff78692c329788dac/LibOS/shim/src/fs/tmpfs/fs.c#L84)):*

> From what I can see you just malloc this buffer so that you can copy data into it in `icheckpoint` and then copy that data into actual checkpoint and free it? Why not copy data directly into checkpoint? It should be doable in the current state, just add a checkpointing function for tmpfs inodes/data.

I mean, I could define something like `BEGIN_CP_FUNC(tmpfs_inode)` but then I don't know how to call it based on the filesystem type. It sounds like I'd have to store a pointer to it in the filesystem callbacks (e.g. `shim_d_ops`, and I don't like where that leads (for example, both the name and the signature of that function are hidden behind macros...)

So, for the sake of sanity, I wanted to keep `icheckpoint/irestore` as a set of simple callbacks that are completely independent of the current checkpointing system. I hope it's worth it even with the extra copy.
</blockquote></details>

I think you can just:
```c
switch (inode->type) {
    case tmpfs:
        DO_CP(tmpfs_inode, ptr1, ptr2);
        break;
    case X:
         ....
    default:
        ...
```

Idk if it's worth

mkow (changes requested):
> Reviewed 3 of 7 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/297)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

___
*[LibOS/shim/include/shim_fs.h, line 172 at r2](https://reviewable.io/reviews/gramineproject/gramine/297#-MurnRus7Pnd7YLzoz8v:-MurnRut3VfFXsuDjD3a:b-5i5vji) ([raw file](https://github.com/gramineproject/gramine/blob/41f583578288eafd7f1294efe6e8122158cf3bed/LibOS/shim/include/shim_fs.h#L172)):*
> ```objc
>     bool maybe_has_fs_locks;
> 
>     struct shim_lock lock;
> ```

Do I see correctly that only `.data` is now protected by this lock? If so, maybe we should rename it to `data_lock`?
___
*[LibOS/shim/include/shim_fs.h, line 370 at r2](https://reviewable.io/reviews/gramineproject/gramine/297#-Muromv4CmLL_DA21Z2j:-Muromv4CmLL_DA21Z2k:biwvu40) ([raw file](https://github.com/gramineproject/gramine/blob/41f583578288eafd7f1294efe6e8122158cf3bed/LibOS/shim/include/shim_fs.h#L370)):*
> ```objc
>      * inode.
>      */
>     void (*iput)(struct shim_inode* inode);
> ```

I think the convention we use is that "put"/"get" are used for refcounted objects. Here the description suggests that this function unconditionally frees the data, so I'd rather name it like "ifree" or something similar.
___
*[LibOS/shim/include/shim_fs.h, line 389 at r2](https://reviewable.io/reviews/gramineproject/gramine/297#-MurpUuwEQflWrF6M17-:-MurpUuwEQflWrF6M170:b-6cc28w) ([raw file](https://github.com/gramineproject/gramine/blob/41f583578288eafd7f1294efe6e8122158cf3bed/LibOS/shim/include/shim_fs.h#L389)):*
> ```objc
>      *
>      * \param inode newly restored inode
>      * \param data checkpoint data (prepared by the `icheckpoint` callback)
> ```

[not related to this PR directly]
This Doxygen formatting of parameter description really confuses and annoys me, we need to change it. I always start with reading like here "data checkpoint data", wondering wtf and then noticing that the first "data" is the parameter name.
Doxygen website capitalizes the description to make it a bit more distinct. We could also consider just spacing it a bit, e.g. minimum 2 spaces and aligning all params descriptions together.
___
*[LibOS/shim/src/fs/shim_fs_util.c, line 68 at r2](https://reviewable.io/reviews/gramineproject/gramine/297#-MurtIwC5cuyHNaoadfb:-MurtIwD-dV_Mr5MfCQC:b-npkg03) ([raw file](https://github.com/gramineproject/gramine/blob/41f583578288eafd7f1294efe6e8122158cf3bed/LibOS/shim/src/fs/shim_fs_util.c#L68)):*
> ```C
>     assert(locked(&g_dcache_lock));
> 
>     if (!dent->inode)
> ```

When is this condition possible? Shouldn't this be an assert?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/297)* status: 6 of 7 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[LibOS/shim/include/shim_fs.h, line 172 at r2](https://reviewable.io/reviews/gramineproject/gramine/297#-MurnRus7Pnd7YLzoz8v:-MuuGWJCARqgb-czZj1t:b8l17z2) ([raw file](https://github.com/gramineproject/gramine/blob/41f583578288eafd7f1294efe6e8122158cf3bed/LibOS/shim/include/shim_fs.h#L172)):*

Do I see correctly that only `.data` is now protected by this lock? If so, maybe we should rename it to `data_lock`?
</blockquote></details>

Both `data` and `lock` are on their way out (filesystem-specific data, when necessary, will be stored in inode), so I'd rather not change it for now. There is a comment at the top of this structure that explains the end goal.
___
*[LibOS/shim/include/shim_fs.h, line 370 at r2](https://reviewable.io/reviews/gramineproject/gramine/297#-Muromv4CmLL_DA21Z2j:-MuuHlIS04_YViZaZRrG:bfhyw58) ([raw file](https://github.com/gramineproject/gramine/blob/41f583578288eafd7f1294efe6e8122158cf3bed/LibOS/shim/include/shim_fs.h#L370)):*

I think the convention we use is that "put"/"get" are used for refcounted objects. Here the description suggests that this function unconditionally frees the data, so I'd rather name it like "ifree" or something similar.
</blockquote></details>

I agree, and this convention caused problems before - someone used `dput` to implement reference counting...

Before I change, what do you think about "idrop" or "irelease"? Since this could also be about closing handles or releasing other resources, not only freeing memory.
___
*[LibOS/shim/include/shim_fs.h, line 389 at r2](https://reviewable.io/reviews/gramineproject/gramine/297#-MurpUuwEQflWrF6M17-:-MuuIcefDL1xPzu2xE5c:bdawxqi) ([raw file](https://github.com/gramineproject/gramine/blob/41f583578288eafd7f1294efe6e8122158cf3bed/LibOS/shim/include/shim_fs.h#L389)):*

[not related to this PR directly]
This Doxygen formatting of parameter description really confuses and annoys me, we need to change it. I always start with reading like here "data checkpoint data", wondering wtf and then noticing that the first "data" is the parameter name.
Doxygen website capitalizes the description to make it a bit more distinct. We could also consider just spacing it a bit, e.g. minimum 2 spaces and aligning all params descriptions together.
</blockquote></details>

+1. "Minimum 2 spaces + align all params" sounds good, I can try it out (separately from this PR).
___
*[LibOS/shim/src/fs/shim_fs_util.c, line 68 at r2](https://reviewable.io/reviews/gramineproject/gramine/297#-MurtIwC5cuyHNaoadfb:-MuuJNnODARqQTI7INsS:b-q33c9v) ([raw file](https://github.com/gramineproject/gramine/blob/41f583578288eafd7f1294efe6e8122158cf3bed/LibOS/shim/src/fs/shim_fs_util.c#L68)):*

When is this condition possible? Shouldn't this be an assert?
</blockquote></details>

Yes, I think so (there were race conditions possible before, but now we call this immediately after lookup, under `g_dcache_lock`). Done.
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 84 at r1](https://reviewable.io/reviews/gramineproject/gramine/297#-MruxArTDANZQARJlA5X:-MuuJnYBDLCtmB3wlmHA:b-uv6xvt) ([raw file](https://github.com/gramineproject/gramine/blob/3fb9cf5c6069af94c78932fff78692c329788dac/LibOS/shim/src/fs/tmpfs/fs.c#L84)):*

I think you can just:
```c
switch (inode->type) {
    case tmpfs:
        DO_CP(tmpfs_inode, ptr1, ptr2);
        break;
    case X:
         ....
    default:
        ...
```

Idk if it's worth
</blockquote></details>

I thought about it, but we already have some special cases for filesystems in the "generic" checkpointing code, and they are a pain to work with. This one would be a simple dispatch, but I'd still rather avoid it.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/297)* status: 6 of 7 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @pwmarcz)

___
*[LibOS/shim/src/fs/shim_fs_util.c, line 68 at r2](https://reviewable.io/reviews/gramineproject/gramine/297#-MurtIwC5cuyHNaoadfb:-Muugd7AAaVwjOfT6gr3:b-2148mq) ([raw file](https://github.com/gramineproject/gramine/blob/41f583578288eafd7f1294efe6e8122158cf3bed/LibOS/shim/src/fs/shim_fs_util.c#L68)):*

Yes, I think so (there were race conditions possible before, but now we call this immediately after lookup, under `g_dcache_lock`). Done.
</blockquote></details>

Jenkins points out an issue with synthetic dentries that actually do *not* have an inode... Let me think of a least-hacky way of handling that.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/297)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[LibOS/shim/include/shim_fs.h, line 172 at r2](https://reviewable.io/reviews/gramineproject/gramine/297#-MurnRus7Pnd7YLzoz8v:-Muuig2b6TuU7nKtwGCl:b-tyvqxt) ([raw file](https://github.com/gramineproject/gramine/blob/41f583578288eafd7f1294efe6e8122158cf3bed/LibOS/shim/include/shim_fs.h#L172)):*

Both `data` and `lock` are on their way out (filesystem-specific data, when necessary, will be stored in inode), so I'd rather not change it for now. There is a comment at the top of this structure that explains the end goal.
</blockquote></details>

Ah, I missed it. Then it's fine to leave this as is.
___
*[LibOS/shim/include/shim_fs.h, line 370 at r2](https://reviewable.io/reviews/gramineproject/gramine/297#-Muromv4CmLL_DA21Z2j:-MuuiJDP-HkIaRXswHlb:brxp9sq) ([raw file](https://github.com/gramineproject/gramine/blob/41f583578288eafd7f1294efe6e8122158cf3bed/LibOS/shim/include/shim_fs.h#L370)):*

I agree, and this convention caused problems before - someone used `dput` to implement reference counting...

Before I change, what do you think about "idrop" or "irelease"? Since this could also be about closing handles or releasing other resources, not only freeing memory.
</blockquote></details>

Sounds good.
___
*[LibOS/shim/include/shim_fs.h, line 389 at r2](https://reviewable.io/reviews/gramineproject/gramine/297#-MurpUuwEQflWrF6M17-:-MuuiLjTEv8xa_TVrC6Q:b4c9lyr) ([raw file](https://github.com/gramineproject/gramine/blob/41f583578288eafd7f1294efe6e8122158cf3bed/LibOS/shim/include/shim_fs.h#L389)):*

+1. "Minimum 2 spaces + align all params" sounds good, I can try it out (separately from this PR).
</blockquote></details>

:thumbsup:

mkow (commented):
> Reviewed 1 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/297)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/297)* status: 3 of 7 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @pwmarcz)

___
*[LibOS/shim/include/shim_fs.h, line 370 at r2](https://reviewable.io/reviews/gramineproject/gramine/297#-Muromv4CmLL_DA21Z2j:-Muusx534LNG99Gw4Q9D:bxnmpli) ([raw file](https://github.com/gramineproject/gramine/blob/41f583578288eafd7f1294efe6e8122158cf3bed/LibOS/shim/include/shim_fs.h#L370)):*

Sounds good.
</blockquote></details>

Changed to "idrop".
___
*[LibOS/shim/include/shim_fs.h, line 389 at r2](https://reviewable.io/reviews/gramineproject/gramine/297#-MurpUuwEQflWrF6M17-:-MuusTeI7bmb_srTcrpM:b7adp3y) ([raw file](https://github.com/gramineproject/gramine/blob/41f583578288eafd7f1294efe6e8122158cf3bed/LibOS/shim/include/shim_fs.h#L389)):*

:thumbsup: 
</blockquote></details>

You're still blocking, should I change anything here?
___
*[LibOS/shim/src/fs/shim_fs_util.c, line 68 at r2](https://reviewable.io/reviews/gramineproject/gramine/297#-MurtIwC5cuyHNaoadfb:-Muut4ISA7clss6tL9-C:b-sw3bs5) ([raw file](https://github.com/gramineproject/gramine/blob/41f583578288eafd7f1294efe6e8122158cf3bed/LibOS/shim/src/fs/shim_fs_util.c#L68)):*

Jenkins points out an issue with synthetic dentries that actually do *not* have an inode... Let me think of a least-hacky way of handling that.
</blockquote></details>

Okay... I believe the right way is to treat these synthetic nodes as not belonging to the current FS (set `dent->fs` to something else, same as we do with pipes, sockets etc.)

I added a TODO here and I'll try to fix it in a separate PR: this is a problem, but I don't think this PR makes it worse (you already can't interact with such files in various ways, their permissions are 0).

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/297)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

___
*[LibOS/shim/include/shim_fs.h, line 389 at r2](https://reviewable.io/reviews/gramineproject/gramine/297#-MurpUuwEQflWrF6M17-:-Muv6SwX5isj0L2pNNSj:b-p4ybu) ([raw file](https://github.com/gramineproject/gramine/blob/41f583578288eafd7f1294efe6e8122158cf3bed/LibOS/shim/include/shim_fs.h#L389)):*

You're still blocking, should I change anything here?
</blockquote></details>

Oops, fixed.

boryspoplawski (changes requested):
> Reviewed 2 of 7 files at r1, 1 of 3 files at r2, 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/297)* status: all files reviewed, 15 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[LibOS/shim/src/fs/shim_dcache.c, line 392 at r4](https://reviewable.io/reviews/gramineproject/gramine/297#-Muv92JG0Eh6u8FXO3H3:-Muv92JG0Eh6u8FXO3H4:bdthlh8) ([raw file](https://github.com/gramineproject/gramine/blob/6efc0459200967ae77471d750e886a1186c10f23/LibOS/shim/src/fs/shim_dcache.c#L392)):*
> ```C
> void put_inode(struct shim_inode* inode) {
>     if (REF_DEC(inode->ref_count) == 0) {
>         if (inode->fs->d_ops && inode->fs->d_ops->idrop)
> ```

I wonder, is the first part of check needed? When can `fs` not have `d_ops` field?
___
*[LibOS/shim/src/fs/shim_fs_util.c, line 93 at r4](https://reviewable.io/reviews/gramineproject/gramine/297#-MuvE9i6CiOwf87rrVe_:-MuvE9i746EfZWmgNK8o:b864cfo) ([raw file](https://github.com/gramineproject/gramine/blob/6efc0459200967ae77471d750e886a1186c10f23/LibOS/shim/src/fs/shim_fs_util.c#L93)):*
> ```C
>     file_off_t pos = hdl->pos;
> 
>     lock(&hdl->inode->lock);
> ```

Is this lock ordering correct? I was under impression that you should not take any other locks while holding `handle->lock`
Also from what I can see we don't need any lock for accessing `hdl->inode`, so we should be able to move this size acquisition outside of `hdl->lock`
___
*[LibOS/shim/src/fs/shim_fs_util.c, line 109 at r4](https://reviewable.io/reviews/gramineproject/gramine/297#-MuvGIWhDdQOHp7dC09a:-MuvGIWhDdQOHp7dC09b:b-rn1zx9) ([raw file](https://github.com/gramineproject/gramine/blob/6efc0459200967ae77471d750e886a1186c10f23/LibOS/shim/src/fs/shim_fs_util.c#L109)):*
> ```C
>     int ret;
> 
>     lock(&hdl->lock);
> ```

ditto (lock order)
___
*[LibOS/shim/src/fs/shim_fs_util.c, line 116 at r4](https://reviewable.io/reviews/gramineproject/gramine/297#-MuvGNLd8ToNCL2z-Rnq:-MuvGNLd8ToNCL2z-Rnr:b-1vaov3) ([raw file](https://github.com/gramineproject/gramine/blob/6efc0459200967ae77471d750e886a1186c10f23/LibOS/shim/src/fs/shim_fs_util.c#L116)):*
> ```C
>         if (poll_type & FS_POLL_WR)
>             ret |= FS_POLL_WR;
>         if ((poll_type & FS_POLL_RD) && hdl->pos < hdl->inode->size)
> ```

Why this size check? `poll` returns that file is readable even if `pos >= size` (after all `read` will return `0` immediately).
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 84 at r1](https://reviewable.io/reviews/gramineproject/gramine/297#-MruxArTDANZQARJlA5X:-MuvYfjCB2Ag-_GGNV98:ba0r5dt) ([raw file](https://github.com/gramineproject/gramine/blob/3fb9cf5c6069af94c78932fff78692c329788dac/LibOS/shim/src/fs/tmpfs/fs.c#L84)):*
How about this function does not allocate the buffer, but gets it from the caller. It could also be called with ptr=NULL and size=0 and then return the size needed.

> This one would be a simple dispatch, but I'd still rather avoid it.

Actually why? What's the difference of calling a function pointer vs deciding what function to call based on type, in this context?
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 9 at r4](https://reviewable.io/reviews/gramineproject/gramine/297#-MuvPoeQB7Jg05OEEdHz:-MuvPoeQB7Jg05OEEdI-:b-qy1jj1) ([raw file](https://github.com/gramineproject/gramine/blob/66f58d0cc8673ab85e333d68f3c682e8ad20bc4d/LibOS/shim/src/fs/tmpfs/fs.c#L9)):*
> ```C
> /*
>  * This file contains code for implementation of 'tmpfs' filesystem. The tmpfs files are *not*
>  * cloned during fork/clone and cannot be synchronized between processes.
> ```

Why did you drop this comment about synchronization?
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 64 at r4](https://reviewable.io/reviews/gramineproject/gramine/297#-MuvRG_J16QL9-oFt-4T:-MuvRG_J16QL9-oFt-4U:badjtjl) ([raw file](https://github.com/gramineproject/gramine/blob/6efc0459200967ae77471d750e886a1186c10f23/LibOS/shim/src/fs/tmpfs/fs.c#L64)):*
> ```C
> }
> 
> static void tmpfs_idrop(struct shim_inode* inode) {
> ```

Which locks should be here (`inode->lock`)? Please add asserts
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 162 at r4](https://reviewable.io/reviews/gramineproject/gramine/297#-Muv_SjP4JahlnlRUPIe:-Muv_SjP4JahlnlRUPIf:b-nni5bf) ([raw file](https://github.com/gramineproject/gramine/blob/6efc0459200967ae77471d750e886a1186c10f23/LibOS/shim/src/fs/tmpfs/fs.c#L162)):*
> ```C
> 
>     mode_t type = S_IFREG;
>     mode_t perm = mode & ~S_IFMT;
> ```

Previously nothing was cleared from `mode` and it was actually named `perm`. From what I can see it's indeed being passed just permission bits, without file type.
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 176 at r4](https://reviewable.io/reviews/gramineproject/gramine/297#-Muvakro85k0Eial5Az9:-Muvakro85k0Eial5AzA:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/6efc0459200967ae77471d750e886a1186c10f23/LibOS/shim/src/fs/tmpfs/fs.c#L176)):*
> ```C
> 
>     mode_t type = S_IFREG;
>     mode_t perm = mode & ~S_IFDIR;
> ```

ditto
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 222 at r4](https://reviewable.io/reviews/gramineproject/gramine/297#-Muvc0Ow4H6yYvGSsvER:-Muvc0Ow4H6yYvGSsvES:bozpt9t) ([raw file](https://github.com/gramineproject/gramine/blob/6efc0459200967ae77471d750e886a1186c10f23/LibOS/shim/src/fs/tmpfs/fs.c#L222)):*
> ```C
> 
> static int tmpfs_rename(struct shim_dentry* old, struct shim_dentry* new) {
>     assert(locked(&g_dcache_lock));
> ```

Shouldn't we also hold `old->inode->lock`?
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 244 at r4](https://reviewable.io/reviews/gramineproject/gramine/297#-MuvbbwX0aVzkrDiBv_5:-MuvbbwX0aVzkrDiBv_6:bvpkzbh) ([raw file](https://github.com/gramineproject/gramine/blob/6efc0459200967ae77471d750e886a1186c10f23/LibOS/shim/src/fs/tmpfs/fs.c#L244)):*
> ```C
>     old->inode = NULL;
> 
>     new->inode->mtime = time_us / USEC_IN_SEC;
> ```

Shouldn't we copy rest of the `*time` fields?
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 269 at r4](https://reviewable.io/reviews/gramineproject/gramine/297#-MuvcICP7LkzwNlGnVx2:-MuvcICP7LkzwNlGnVx3:b-umsewg) ([raw file](https://github.com/gramineproject/gramine/blob/6efc0459200967ae77471d750e886a1186c10f23/LibOS/shim/src/fs/tmpfs/fs.c#L269)):*
> ```C
> 
>     lock(&hdl->lock);
>     lock(&inode->lock);
> ```

ditto (locks ordering)
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 302 at r4](https://reviewable.io/reviews/gramineproject/gramine/297#-MuvcZavDwofUOEGUHzx:-MuvcZavDwofUOEGUHzy:b-umsewg) ([raw file](https://github.com/gramineproject/gramine/blob/6efc0459200967ae77471d750e886a1186c10f23/LibOS/shim/src/fs/tmpfs/fs.c#L302)):*
> ```C
> 
>     lock(&hdl->lock);
>     lock(&inode->lock);
> ```

ditto (locks ordering)
___
*[LibOS/shim/src/sys/shim_clone.c, line 129 at r4](https://reviewable.io/reviews/gramineproject/gramine/297#-MuveJ0b4l6X71-aU72S:-MuveJ0b4l6X71-aU72T:b-datn1w) ([raw file](https://github.com/gramineproject/gramine/blob/6efc0459200967ae77471d750e886a1186c10f23/LibOS/shim/src/sys/shim_clone.c#L129)):*
> ```C
>                         struct shim_ipc_ids* process_ipc_ids, va_list ap) {
>     __UNUSED(ap);
>     lock(&g_dcache_lock);
> ```

Please add a comment here why we take this lock here.
___
*[LibOS/shim/test/regression/test_libos.py, line 238 at r4](https://reviewable.io/reviews/gramineproject/gramine/297#-MuvepnT2CiMpYM_mmZc:-MuvepnT2CiMpYM_mmZd:b-7ke6p2) ([raw file](https://github.com/gramineproject/gramine/blob/6efc0459200967ae77471d750e886a1186c10f23/LibOS/shim/test/regression/test_libos.py#L238)):*
> ```Python
>             os.unlink(path)
>         try:
>             self._test_send_handle('tmp/pf/send_handle_test')
> ```

`self._test_send_handle(path)`
___
*[LibOS/shim/test/regression/test_libos.py, line 240 at r4](https://reviewable.io/reviews/gramineproject/gramine/297#-Muvez_-BOCf5ArtE-42:-Muvez_-BOCf5ArtE-43:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/6efc0459200967ae77471d750e886a1186c10f23/LibOS/shim/test/regression/test_libos.py#L240)):*
> ```Python
>             self._test_send_handle('tmp/pf/send_handle_test')
>             # TODO: Migrating a protected files handle is not supported when the file is deleted
>             # self._test_send_handle('tmp/pf/send_handle_test', delete=True)
> ```

ditto

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/297)* status: all files reviewed, 15 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @pwmarcz)

___
*[LibOS/shim/src/fs/shim_dcache.c, line 392 at r4](https://reviewable.io/reviews/gramineproject/gramine/297#-Muv92JG0Eh6u8FXO3H3:-MuypgrU1GYwIkH0ojkB:b5uz0gf) ([raw file](https://github.com/gramineproject/gramine/blob/6efc0459200967ae77471d750e886a1186c10f23/LibOS/shim/src/fs/shim_dcache.c#L392)):*

I wonder, is the first part of check needed? When can `fs` not have `d_ops` field?
</blockquote></details>

This happens for pipe, socket etc.

(I agree it would be better to always have both `fs_ops` and `d_ops` in all filesystems, for consistency.)
___
*[LibOS/shim/src/fs/shim_fs_util.c, line 93 at r4](https://reviewable.io/reviews/gramineproject/gramine/297#-MuvE9i6CiOwf87rrVe_:-Muyq7C57EZQHk-VM8lb:b-8sbzrg) ([raw file](https://github.com/gramineproject/gramine/blob/6efc0459200967ae77471d750e886a1186c10f23/LibOS/shim/src/fs/shim_fs_util.c#L93)):*

Is this lock ordering correct? I was under impression that you should not take any other locks while holding `handle->lock`
Also from what I can see we don't need any lock for accessing `hdl->inode`, so we should be able to move this size acquisition outside of `hdl->lock`
</blockquote></details>

I wrote all this assuming exactly this lock ordering; the code in `chroot/fs.c` and `tmpfs/fs.c` also obeys that rule. I added a comment to `shim_handle.inode`.

I think here you actually to hold `hdl->lock` while checking size. Otherwise, you might have a race with `write` on the same handle. Consider concurrent `seek` and `write` operations on a handle:

```c
// thread 1:                        thread 2:
seek(hdl, -1, SEEK_END);            write(hdl, buf, 5);

// hdl->pos = 10, inode->size = 10
size = inode->size;
                                    hdl->pos += 5;
                                    hdl->size += 5;
// hdl->pos = 15, inode->size = 15
hdl->pos = size - 1;
// hdl->pos = 9, inode->size = 15
```

So the two operations end up non-serializable (`seek` couldn't have happened either before or after `write`).
___
*[LibOS/shim/src/fs/shim_fs_util.c, line 116 at r4](https://reviewable.io/reviews/gramineproject/gramine/297#-MuvGNLd8ToNCL2z-Rnq:-MuywqXG4BO_x-6y_i86:b-6uikbf) ([raw file](https://github.com/gramineproject/gramine/blob/6efc0459200967ae77471d750e886a1186c10f23/LibOS/shim/src/fs/shim_fs_util.c#L116)):*

Why this size check? `poll` returns that file is readable even if `pos >= size` (after all `read` will return `0` immediately).
</blockquote></details>

Right. As far as I can tell Gramine's `poll` always worked like this :)

But I'd rather not change the behavior in this PR: I haven't touched the poll implementation yet, and I remember it to be pretty bad, so I expect things might break. I added a TODO.
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 9 at r4](https://reviewable.io/reviews/gramineproject/gramine/297#-MuvPoeQB7Jg05OEEdHz:-MuyxLSHC5yR5YO8YPp_:bokk45c) ([raw file](https://github.com/gramineproject/gramine/blob/66f58d0cc8673ab85e333d68f3c682e8ad20bc4d/LibOS/shim/src/fs/tmpfs/fs.c#L9)):*

Why did you drop this comment about synchronization?
</blockquote></details>

Right, it still applies. I think I thought it was about migration.
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 64 at r4](https://reviewable.io/reviews/gramineproject/gramine/297#-MuvRG_J16QL9-oFt-4T:-MuyxdM5CcF1DxUM9sw3:b2g1jqw) ([raw file](https://github.com/gramineproject/gramine/blob/6efc0459200967ae77471d750e886a1186c10f23/LibOS/shim/src/fs/tmpfs/fs.c#L64)):*

Which locks should be here (`inode->lock`)? Please add asserts
</blockquote></details>

Right, this should be called with `inode->lock`, same as `icheckpoint`. Fixed.
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 162 at r4](https://reviewable.io/reviews/gramineproject/gramine/297#-Muv_SjP4JahlnlRUPIe:-Muyy8VuExqaJOeuQ-Hi:bidrs83) ([raw file](https://github.com/gramineproject/gramine/blob/6efc0459200967ae77471d750e886a1186c10f23/LibOS/shim/src/fs/tmpfs/fs.c#L162)):*

Previously nothing was cleared from `mode` and it was actually named `perm`. From what I can see it's indeed being passed just permission bits, without file type.
</blockquote></details>

Right. Fixed.
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 176 at r4](https://reviewable.io/reviews/gramineproject/gramine/297#-Muvakro85k0Eial5Az9:-MuyyPlj7RSxNVM-rRIx:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/6efc0459200967ae77471d750e886a1186c10f23/LibOS/shim/src/fs/tmpfs/fs.c#L176)):*

ditto
</blockquote></details>

Done.
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 222 at r4](https://reviewable.io/reviews/gramineproject/gramine/297#-Muvc0Ow4H6yYvGSsvER:-Muz-HsvBNFMyWWVk-pR:bn8458z) ([raw file](https://github.com/gramineproject/gramine/blob/6efc0459200967ae77471d750e886a1186c10f23/LibOS/shim/src/fs/tmpfs/fs.c#L222)):*

Shouldn't we also hold `old->inode->lock`?
</blockquote></details>

For modifying time, right. Done.
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 244 at r4](https://reviewable.io/reviews/gramineproject/gramine/297#-MuvbbwX0aVzkrDiBv_5:-Muz-KKX5kRoicO-knYT:b-qlul6y) ([raw file](https://github.com/gramineproject/gramine/blob/6efc0459200967ae77471d750e886a1186c10f23/LibOS/shim/src/fs/tmpfs/fs.c#L244)):*

Shouldn't we copy rest of the `*time` fields?
</blockquote></details>

We don't copy anything, we update the change time[1] after the file has been renamed. The modify time and access time remain the same.

[1] When renaming, apparently we update "change time", not "modify time". Fixed.
___
*[LibOS/shim/src/sys/shim_clone.c, line 129 at r4](https://reviewable.io/reviews/gramineproject/gramine/297#-MuveJ0b4l6X71-aU72S:-Muz0KoO85A6Jjfe1b0X:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/6efc0459200967ae77471d750e886a1186c10f23/LibOS/shim/src/sys/shim_clone.c#L129)):*

Please add a comment here why we take this lock here.
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/test_libos.py, line 238 at r4](https://reviewable.io/reviews/gramineproject/gramine/297#-MuvepnT2CiMpYM_mmZc:-Muz0tbl4_7a6F52v99Q:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/6efc0459200967ae77471d750e886a1186c10f23/LibOS/shim/test/regression/test_libos.py#L238)):*

`self._test_send_handle(path)`
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/test_libos.py, line 240 at r4](https://reviewable.io/reviews/gramineproject/gramine/297#-Muvez_-BOCf5ArtE-42:-Muz0u1p07esSMGwqToc:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/6efc0459200967ae77471d750e886a1186c10f23/LibOS/shim/test/regression/test_libos.py#L240)):*

ditto
</blockquote></details>

Done.

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/297)* status: all files reviewed, 15 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @pwmarcz)

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/297)* status: 4 of 9 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, @mkow, and @pwmarcz)

___
*[LibOS/shim/src/fs/shim_fs_util.c, line 93 at r4](https://reviewable.io/reviews/gramineproject/gramine/297#-MuvE9i6CiOwf87rrVe_:-Mv-MY4bAynaR67BFAnc:b-g1brus) ([raw file](https://github.com/gramineproject/gramine/blob/6efc0459200967ae77471d750e886a1186c10f23/LibOS/shim/src/fs/shim_fs_util.c#L93)):*

The problem is that rest of the code base always assumed that you can take `handle->lock`. Note we need some form of global ordering of locks, usually it's easy, because locks are private for some subsystem and you should not hold any other locks when calling functions exported/globally visible from that subsystem. `shim_handle` is one important exception - it's used across LibOS, so it was the easiest to make it this "last level"/"last taken" lock.
Maybe we can have the current version and just specify that `inode` lock is taken after `hdl`, but it's hard for me to prove that there are no cycles.
Is there any problem with taking these two locks in `inode` -> `hdl` order? From what I understand `hdl->inode` is constant and accessible without any locks.
</blockquote></details>

OK. I don't have a good intuition for what the lock ordering should be. I'm a bit worried about concurrent operations on a single file, but maybe we don't need to deal with that right now.

I reordered these two locks.
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 84 at r1](https://reviewable.io/reviews/gramineproject/gramine/297#-MruxArTDANZQARJlA5X:-Mv-RqKe4fh0uB82YVjo:b-d1ez68) ([raw file](https://github.com/gramineproject/gramine/blob/3fb9cf5c6069af94c78932fff78692c329788dac/LibOS/shim/src/fs/tmpfs/fs.c#L84)):*
>How about this function does not allocate the buffer, but gets it from the caller. It could also be called with ptr=NULL and size=0 and then return the size needed.

So, a callback that you call twice?
```c
ssize_t (*icheckpoint)(struct inode* inode, void* buf, size_t size);

ssize_t size = inode->fs->ops->icheckpoint(inode, NULL, 0);
if (size < 0)
    return size;
...
ret = inode->fs->ops->icheckpoint(inode, buf, size);
```
Looks pretty ugly... Maybe if we split it into two functions: `icheckpoint_size` and `icheckpoint`? IMO, still not worth it, but I can implement it if you want.

Ideally, I would want something like this:

```c
/* This calls some functions on `cp_store` to reserve a buffer, then writes data to it. */
int (*icheckpoint)(struct inode* inode, struct cp_store* store, void** out_buf);
```

We actually have a similar interface for the checkpoint functions (they take a "size" parameter but it's not used for anything). However, everything is hidden behind layers of macros, which is why I don't want to use it in the current state.

>Actually why? What's the difference of calling a function pointer vs deciding what function to call based on type, in this context?

Because this switch would have to be updated for every filesystem that uses this functionality. And I think it sets a bad precedent: it would be too easy to add custom behavior right inside this switch. Finally, I suspect that it would make it harder to rewrite the checkpointing subsystem when we finally get to it.

boryspoplawski (approved):
> Reviewed 5 of 5 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/297)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 84 at r1](https://reviewable.io/reviews/gramineproject/gramine/297#-MruxArTDANZQARJlA5X:-Mv-dp8_5M-FjodstX3i:bsaixbb) ([raw file](https://github.com/gramineproject/gramine/blob/3fb9cf5c6069af94c78932fff78692c329788dac/LibOS/shim/src/fs/tmpfs/fs.c#L84)):*
> IMO, still not worth it, but I can implement it if you want.

No need, was just an idea. I just don't like this additional malloc (because now we have 3 copies of the same file and if that file is big...), but if everything works, we can live with that.

dimakuv (changes requested):
> Reviewed 4 of 7 files at r5, 5 of 5 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/297)* status: all files reviewed, 1 unresolved discussion, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[-- commits, line 18 at r6](https://reviewable.io/reviews/gramineproject/gramine/297#-MvI6p7oBqC7fCuzkHDr):*
`iput` -> `idrop`

Please fix during final rebase.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/297)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

___
*[-- commits, line 18 at r6](https://reviewable.io/reviews/gramineproject/gramine/297#-MvI6p7oBqC7fCuzkHDr):*

`iput` -> `idrop`

Please fix during final rebase.
</blockquote></details>

Done.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/297)* status: :shipit: complete! all files reviewed, all discussions resolved

mkow (commented):
> Reviewed 5 of 5 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/297)* status: all files reviewed, 1 unresolved discussion (waiting on @pwmarcz)

___
*[LibOS/shim/src/bookkeep/shim_handle.c, line 768 at r7](https://reviewable.io/reviews/gramineproject/gramine/297#-MvIuS308KWoicM8kgmg:-MvIuS313tAm28TIsRg3:bep4l9e) ([raw file](https://github.com/gramineproject/gramine/blob/97f8e5db3d5619daed7ee58eb1b2c1c618d197a8/LibOS/shim/src/bookkeep/shim_handle.c#L768)):*
> ```C
>         if (hdl->inode) {
>             /* NOTE: Checkpointing `inode` will take `inode->lock`, so we need to do it after
>              * `hdl->lock` is released. */
> ```

Hmm, is this correct? What something inside `hdl` changes between the code above and checkpointing below, in that window where we don't have the lock?
___
*[LibOS/shim/src/fs/shim_fs_util.c, line 93 at r4](https://reviewable.io/reviews/gramineproject/gramine/297#-MuvE9i6CiOwf87rrVe_:-MvIu9Uu4OszfYcMcmTb:bccvi0c) ([raw file](https://github.com/gramineproject/gramine/blob/6efc0459200967ae77471d750e886a1186c10f23/LibOS/shim/src/fs/shim_fs_util.c#L93)):*

OK. I don't have a good intuition for what the lock ordering should be. I'm a bit worried about concurrent operations on a single file, but maybe we don't need to deal with that right now.

I reordered these two locks.
</blockquote></details>

We should implement a lockdep-like checker at some point...

PR #311, opened 2022-01-04
[LibOS] Fix segfault during async-signal upcalls

dimakuv:
## Description of the changes 

It is possible that an async signal occurs during LibOS thread initialization and is forwarded by PAL to `quit_upcall()` / `interrupted_upcall()`. At this point, `get_cur_thread() == NULL` and `is_internal()` helper function segfaults on `thread->tid`. This commit fixes this bug.

Note that we don't implement a similar change for *synchronous signals* because it is assumed that sync signals can never happen during LibOS code.

## How to test this PR? 

This bug was found while I was experimenting with something different on Gramine. That workload increased the frequency of async signals; some of these signals got delivered while in LibOS thread initialization. This manifested itself as a MEMFAULT in Gramine PAL terminology, and led to an infinite loop of such MEMFAULTs.

On my extreme workload, this bug happened about 1 in 10 runs. I believe it's hard to catch this bug in normal workloads of Gramine, so no (simple) way to test this PR.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/311)

### DIFF
--- a/LibOS/shim/src/bookkeep/shim_signal.c
+++ b/LibOS/shim/src/bookkeep/shim_signal.c
@@ -500,7 +500,9 @@ static void quit_upcall(bool is_in_pal, PAL_NUM addr, PAL_CONTEXT* context) {
         return;
     }
 
-    if (is_internal(get_cur_thread()) || context_is_libos(context) || is_in_pal) {
+    /* "quit" signal may occur during LibOS thread initialization (at which point `cur == NULL`) */
+    struct shim_thread* cur = get_cur_thread();
+    if (!cur || is_internal(cur) || context_is_libos(context) || is_in_pal) {
         return;
     }
     handle_signal(context, /*old_mask_ptr=*/NULL);
@@ -509,7 +511,10 @@ static void quit_upcall(bool is_in_pal, PAL_NUM addr, PAL_CONTEXT* context) {
 static void interrupted_upcall(bool is_in_pal, PAL_NUM addr, PAL_CONTEXT* context) {
     __UNUSED(addr);
 
-    if (is_internal(get_cur_thread()) || context_is_libos(context) || is_in_pal) {
+    /* "interrupted" signal may occur during LibOS thread initialization (at which point
+     * `cur == NULL`) */
+    struct shim_thread* cur = get_cur_thread();
+    if (!cur || is_internal(cur) || context_is_libos(context) || is_in_pal) {
         return;
     }
     handle_signal(context, /*old_mask_ptr=*/NULL);


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/311)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[LibOS/shim/src/bookkeep/shim_signal.c, line 504 at r1](https://reviewable.io/reviews/gramineproject/gramine/311#-MsZZVgZBom6Rp19oHmr:-MsZcYA48GVyUJGdOaLI:btxtiz9) ([raw file](https://github.com/gramineproject/gramine/blob/3706988f61fd5469a82b804ed109039394a6c292/LibOS/shim/src/bookkeep/shim_signal.c#L504)):*

Maybe just swap the order and put `is_internal()` as last part of this disjunction (then `context_is_libos` will take care of this case)? As far as this change is correct, I'm just afraid that someday someone could come and just remove this `!cur` believing that `get_cur_thread()` always returns a non NULL value.
Or maybe just add a comment, idk what's better.
</blockquote></details>

Done. I added a comment because I would like to make this explicit (your proposal to reorder obscures this fix).

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/311)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/311)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/311#-Ms_K9xiCLBo9Ek_l8xw:-Ms_K9xiCLBo9Ek_l8xx:bvf60kq) (no related file):*
> This bug was found while I was experimenting with something different on Gramine. That workload increased the frequency of async signals; some of these signals got delivered while in LibOS thread initialization. This manifested itself as a MEMFAULT in Gramine PAL terminology, and led to an infinite loop of such MEMFAULTs.
> On my extreme workload, this bug happened about 1 in 10 runs. I believe it's hard to catch this bug in normal workloads of Gramine, so no (simple) way to test this PR.

Hmm, is it feasible to implement a "signal spam" test? I.e. a workload/test running, plus an external program spamming async signals at the Graminized process (on the host). We find such issues from time to time, would be good to have a test for this.

CC: @boryspoplawski.

---

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/311)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/311#-Ms_K9xiCLBo9Ek_l8xw:-Ms_NhoY7MG_CoFg1Xz5:beme67t) (no related file):*

> This bug was found while I was experimenting with something different on Gramine. That workload increased the frequency of async signals; some of these signals got delivered while in LibOS thread initialization. This manifested itself as a MEMFAULT in Gramine PAL terminology, and led to an infinite loop of such MEMFAULTs.
> On my extreme workload, this bug happened about 1 in 10 runs. I believe it's hard to catch this bug in normal workloads of Gramine, so no (simple) way to test this PR.

Hmm, is it feasible to implement a "signal spam" test? I.e. a workload/test running, plus an external program spamming async signals at the Graminized process (on the host). We find such issues from time to time, would be good to have a test for this.

CC: @boryspoplawski.
</blockquote></details>

Yeah, it would be nice to have such a spam test. I created an issue for this: #312 

---

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/311)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/311)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #328, opened 2022-01-11
[LibOS] Do not inherit an associated file during VMA split

dimakuv:
## Description of the changes 

Previously, a new VMA that was split from the old VMA would inherit the file that was associated with the old VMA, even if the new VMA would start past the end of the associated file. This led to obscure segfaults during VMA checkpoint-and-restore on the Linux-SGX PAL. Also, it led to worse performance because the checkpoint restore logic would try to reopen and mmap the file associated with the new VMA.

This bug was found on Python's qiskit package (it led to a segfault).

## How to test this PR? 

I didn't come up with a self-contained test. Below is the Python's qiskit (https://qiskit.org/documentation/getting_started.html) test that fails without this PR and succeeds with this PR; you can test manually.

1. Install qiskit on Ubuntu 18.04: `pip3 install qiskit`.

2. Go to `CI-Examples/python` and change the manifest template:
```diff
diff --git a/CI-Examples/python/python.manifest.template b/CI-Examples/python/python.manifest.template
@@ -10,6 +10,9 @@ loader.log_level = "{{ log_level }}"
 loader.env.LD_LIBRARY_PATH = "{{ python.stdlib }}/lib:/lib:{{ arch_libdir }}:/usr/lib:/usr/{{ arch_libdir }}"

 loader.insecure__use_cmdline_argv = true
+loader.insecure__use_host_env     = true
+
+loader.pal_internal_mem_size = "128M"

 sys.enable_sigterm_injection = true

@@ -33,6 +36,10 @@ fs.mount.pydisthome.type = "chroot"
 fs.mount.pydisthome.path = "{{ python.distlib }}"
 fs.mount.pydisthome.uri = "file:{{ python.distlib }}"

+fs.mount.piphome.type = "chroot"
+fs.mount.piphome.path = "{{ env.HOME }}/.local/lib"
+fs.mount.piphome.uri = "file:{{ env.HOME }}/.local/lib"
+
 fs.mount.tmp.type = "chroot"
 fs.mount.tmp.path = "/tmp"
 fs.mount.tmp.uri = "file:/tmp"
@@ -43,7 +50,7 @@ fs.mount.etc.uri = "file:/etc"

 sgx.debug = true
 sgx.nonpie_binary = true
-sgx.enclave_size = "512M"
+sgx.enclave_size = "2G"
 sys.stack.size = "2M"
 sgx.thread_num = 32

@@ -55,6 +62,7 @@ sgx.trusted_files = [
   "file:/usr/{{ arch_libdir }}/",
   "file:{{ python.stdlib }}/",
   "file:{{ python.distlib }}/",
+  "file:{{ env.HOME }}/.local/lib/",
   "file:scripts/",
   "file:/etc/mime.types",
   "file:/etc/default/apport",
```

3. Create a Python script under `CI-Examples/python/scripts/qiskit.py`:
```python
from qiskit import QuantumCircuit
qc = QuantumCircuit(3)
qc.h(0)
print("Done")
```

4. Run under Gramine-SGX: `gramine-sgx python scripts/qiskit.py`
    - There is a `No such file or directory: '/proc/stat'` warning by Python, but this is benign and a separate thing to fix in Gramine (our LibOS FS currently doesn't have `/proc/stat` pseudo-file).

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/328)

### DIFF
--- a/LibOS/shim/src/bookkeep/shim_vma.c
+++ b/LibOS/shim/src/bookkeep/shim_vma.c
@@ -188,11 +188,25 @@ static void split_vma(struct shim_vma* old_vma, struct shim_vma* new_vma, uintpt
 
     copy_vma(old_vma, new_vma);
     new_vma->begin = addr;
+    old_vma->end = addr;
+
     if (new_vma->file) {
         new_vma->offset += new_vma->begin - old_vma->begin;
-    }
 
-    old_vma->end = addr;
+        /* new VMA may start past the end of the associated file: remove the association (to avoid
+         * unnecessary file reopen/mmap during checkpoint restore, see BEGIN_RS_FUNC(vma)) */
+        uint64_t file_size = 0;
+        get_file_size(new_vma->file, &file_size);
+        if (new_vma->offset >= file_size) {
+            put_handle(new_vma->file);
+            new_vma->file   = NULL;
+            new_vma->offset = 0;
+            new_vma->flags &= ~VMA_TAINTED;
+
+            /* TODO: new VMA is useless; maybe just leave this line and remove everything else? */
+            new_vma->flags |= VMA_UNMAPPED;
+        }
+    }
 }
 
 /*
@@ -1303,7 +1317,6 @@ BEGIN_RS_FUNC(vma) {
             get_handle(vma->file);
 
             if (need_mapped < vma->addr + vma->length) {
-                /* first try, use hstat to force it resumes pal handle */
                 if (!fs || !fs->fs_ops || !fs->fs_ops->mmap) {
                     return -EINVAL;
                 }


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/328)* status: 0 of 1 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski)

___
*[LibOS/shim/src/bookkeep/shim_vma.c, line 207 at r2](https://reviewable.io/reviews/gramineproject/gramine/328#-Mt9eFME4OMDq7bfytxo:-Mt9eFME4OMDq7bfytxp:b-1ou6gb) ([raw file](https://github.com/gramineproject/gramine/blob/7b0157a91c6fbe3bdbb483164c4771875c5573e1/LibOS/shim/src/bookkeep/shim_vma.c#L207)):*
> ```C
> 
>             /* TODO: new VMA is useless; maybe just leave this line and remove everything else? */
>             new_vma->flags |= VMA_UNMAPPED;
> ```

FYI: (This is for `gramine-direct` only.)

This is needed because in the `qiskit` example, the new-VMA memory region is sent from the parent process to the child process during checkpoint-and-restore. However, this new-VMA memory region is not even accessible because any address inside this region generates SIGBUS (see ERRORS in `man 2 mmap`). And thus the `send(new_vma->addr, ...)` syscall fails, and the whole checkpointing fails with:
```
[P1:T1:python3.6] error: failed sending checkpoint (ret = -13)
[P1:T1:python3.6] error: process creation failed
```

Note that this doesn't affect `gramine-sgx` because SGX enclave pages are always pre-allocated and there is never a SIGBUS.

@boryspoplawski What should I actually do in this corner-case? Just mark the VMA as `VMA_UNMAPPED`?

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/328)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/328#-MtEXd5yEkVR7MgWVGJF:-MtEXd5yEkVR7MgWVGJG:bqunxp5) (no related file):*
1. I've tested the provided example and it works for me (on the current master).
2. We cannot clear `->file` - it would mean that the new VMA is not file-backed, which is not true. The underlying file might be extended and have a bigger size by the time it's accessed after `mprotect`

---

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/328)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/328#-MtEXd5yEkVR7MgWVGJF:-MtEa2Gm3CxH3eUYnb_8:b-j2hj96) (no related file):*

1. I've tested the provided example and it works for me (on the current master).
2. We cannot clear `->file` - it would mean that the new VMA is not file-backed, which is not true. The underlying file might be extended and have a bigger size by the time it's accessed after `mprotect`
</blockquote></details>

It seems you've used python3.6 whereas I had python3.8 - maybe newer version doesn't have similar mmap+mprotect pattern.

I've managed to reproduce the issue - it only triggers on protected files, allowed + everything on Linux PAL works fine. This seems to be some issue with protected files feature, not the LibOS part (including VMAs).

---

dimakuv:
As @boryspoplawski mentions, this is a wrong way of fixing the problem. This PR is superseded by https://github.com/gramineproject/gramine/pull/336.

PR #336, opened 2022-01-13
[LibOS] Skip mmapping past-file-end VMA contents in checkpoint restore

dimakuv:
## Description of the changes 

There was a bug in Gramine: if a VMA was file-backed and then split into two VMAs, such that the second VMA starts past the end of the corresponding file, this VMA info was sent to the child process (but not the contents of the corresponding memory region). However, this VMA info in the sent checkpoint had the `need_mapped` field that was interpreted by the child process as "mmap from the file at given offset".

This operation is meaningless (because the given offset is past the end of file) but allowed in Linux. However, for trusted/protected files on Linux-SGX this leads to their SGX-specific checks failing, and Gramine
crashed. This PR fixes this bug.

For some details and a wrong attempt at fixing it see #328.

## How to test this PR? 

This bug was found on Python's qiskit package (it led to a segfault). This PR adds a new LibOS regression test to catch this bug.

Also see descriptions in #328.

### DIFF
--- a/LibOS/shim/src/bookkeep/shim_vma.c
+++ b/LibOS/shim/src/bookkeep/shim_vma.c
@@ -1244,8 +1244,7 @@ BEGIN_CP_FUNC(vma) {
             size_t send_size = vma->length;
             if (vma->file) {
                 /*
-                 * Chia-Che 8/13/2017:
-                 * A fix for cloning a private VMA which maps a file to a process.
+                 * Corner case of cloning a private VMA which maps a file to a process.
                  *
                  * (1) Application can access any page backed by the file, wholly
                  *     or partially.
@@ -1263,6 +1262,12 @@ BEGIN_CP_FUNC(vma) {
                         && vma->file_offset + vma->length > file_len) {
                     send_size = file_len > vma->file_offset ? file_len - vma->file_offset : 0;
                     send_size = ALLOC_ALIGN_UP(send_size);
+
+                    if (!send_size) {
+                        /* VMA starts past the last file-backed page. Accessing this VMA will cause
+                         * SIGBUS, so emulate it by skipping mmap during this VMA restore. */
+                        need_mapped = vma->addr + vma->length;
+                    }
                 }
             }
             if (send_size > 0) {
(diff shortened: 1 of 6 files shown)


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/336)* status: 0 of 6 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski)

___
*[LibOS/shim/src/bookkeep/shim_vma.c, line 1306 at r1](https://reviewable.io/reviews/gramineproject/gramine/336#-MtIFHBc6rql9Qc3J73k:-MtIFHBdDYaGjjG1JGFz:bf53cgv) ([raw file](https://github.com/gramineproject/gramine/blob/9544bb1bfa58c4fc198c071b70139d6c32c4e15a/LibOS/shim/src/bookkeep/shim_vma.c#L1306)):*
> ```C
> 
>             if (need_mapped < vma->addr + vma->length) {
>                 /* first try, use hstat to force it resumes pal handle */
> ```

FYI: Just removed this stale and incorrect comment.

boryspoplawski (changes requested):
> Reviewed 5 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/336)* status: 5 of 6 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[-- commits, line 13 at r1](https://reviewable.io/reviews/gramineproject/gramine/336#-MtL6iMg4Tb5pMbFvyWL):*
It's not a bug.
> This operation is meaningless (because the given offset is past the end of  file) but allowed in Linux.

It's not meaningless. We might extend the file before accessing this page, so we **have to** mmap it.

> for trusted/protected files on  Linux-SGX this leads to their SGX-specific checks failing, and Gramine crashed.

This the part we need to fix, not LibOS.

dimakuv:
Superseded by https://github.com/gramineproject/gramine/pull/338.

PR #338, opened 2022-01-14
[LibOS] Do not re-map private VMA contents in checkpoint restore

dimakuv:
## Description of the changes 

There was a bug in Gramine: tainted file-backed memory regions marked as `MAP_PRIVATE` were not sent to the child process sometimes -- instead these memory regions were re-mapped in the child from the corresponding file.

This PR forces Gramine LibOS to always send such memory regions in the checkpoint and disallows them to be re-mapped by child. This is in line with POSIX: "MAP_PRIVATE <...> It is unspecified whether changes made to the file after the mmap() call are visible in the mapped region." Gramine implementation chooses to *not* propagate file changes.

For some details and previous wrong attempts at fixing it see #328 and #336.

## How to test this PR? 

This bug was found on Python's qiskit package (it led to a segfault on SGX PAL). This commit adds new LibOS regression test to catch this bug.

Also see descriptions in #328.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/338)

### DIFF
--- a/LibOS/shim/src/bookkeep/shim_vma.c
+++ b/LibOS/shim/src/bookkeep/shim_vma.c
@@ -1265,17 +1265,16 @@ BEGIN_CP_FUNC(vma) {
                  *
                  *   - Only if VMA is both tainted and private, we must send the memory content.
                  *     This VMA will not be backed by a file in the child (i.e., it will be
-                 *     recreated in the child as anonymous memory). We rely on the fact that private
-                 *     mappings do not reflect updates of the file.
-                 *
-                 * FIXME: If file-backed VMA is not tainted but private, we currently re-map from
-                 *        file during restore in the child. But it is possible that the file was
-                 *        modified by another process, and the child's VMA content will differ from
-                 *        the parent's VMA content. Currently we ignore this possibility.
+                 *     recreated in the child as anonymous memory). We assume that all sane POSIX
+                 *     compliant applications will operate on this VMA as if this mapping does not
+                 *     reflect updates of the file (which is not always true on Linux, since Linux
+                 *     uses COW semantics for file-backed mappings).
                  */
                 if ((vma->flags & (VMA_TAINTED | MAP_PRIVATE)) == (VMA_TAINTED | MAP_PRIVATE)) {
                     uint64_t file_size = 0;
-                    get_file_size(vma->file, &file_size);
+                    int ret = get_file_size(vma->file, &file_size);
+                    if (ret < 0)
+                        BUG();
 
                     if (vma->file_offset < file_size) {
                         /* Access beyond the last file-backed page causes SIGBUS. To simulate this
--- a/LibOS/shim/src/bookkeep/shim_vma.c
+++ b/LibOS/shim/src/bookkeep/shim_vma.c
@@ -1236,45 +1236,67 @@ BEGIN_CP_FUNC(vma) {
         if (vma->file)
             DO_CP(handle, vma->file, &new_vma->file);
 
-        void* need_mapped = vma->addr;
+        /* VMA contents may be checkpointed, sent and restored via shim_mem_entry mechanism below;
+         * in such case `remap_in_child_addr` is updated to skip re-mapping during VMA restore. */
+        void* remap_in_child_addr = vma->addr;
 
-        /* Check whether we need to checkpoint memory this vma bookkeeps. */
-        if ((vma->flags & VMA_TAINTED || !vma->file) && !(vma->flags & VMA_UNMAPPED)) {
-            void* send_addr  = vma->addr;
-            size_t send_size = vma->length;
-            if (vma->file) {
+        if (!(vma->flags & VMA_UNMAPPED)) {
+            if (!vma->file) {
                 /*
-                 * Chia-Che 8/13/2017:
-                 * A fix for cloning a private VMA which maps a file to a process.
+                 * Checkpoint anonymous memory this VMA bookkeeps.
                  *
-                 * (1) Application can access any page backed by the file, wholly
-                 *     or partially.
-                 *
-                 * (2) Access beyond the last file-backed page will cause SIGBUS.
-                 *     For reducing fork latency, the following code truncates the
-                 *     memory size for migrating a process. The memory size is
-                 *     truncated to the file size, round up to pages.
-                 *
-                 * (3) Data in the last file-backed page is valid before or after
-                 *     forking. Has to be included in process migration.
+                 * FIXME: We ignore MAP_SHARED flag. VMA content in parent and child may diverge.
                  */
-                uint64_t file_len = 0;
-                if (!get_file_size(vma->file, &file_len)
-                        && vma->file_offset + vma->length > file_len) {
-                    send_size = file_len > vma->file_offset ? file_len - vma->file_offset : 0;
-                    send_size = ALLOC_ALIGN_UP(send_size);
+                struct shim_mem_entry* mem;
+                DO_CP_SIZE(memory, vma->addr, vma->length, &mem);
+                mem->prot = LINUX_PROT_TO_PAL(vma->prot, /*map_flags=*/0);
+                remap_in_child_addr = vma->addr + vma->length;
+            } else {
+                /*
+                 * Check whether we need to checkpoint file-backed memory this VMA bookkeeps:
+                 *
+                 *   - If VMA is not tainted (doesn't contain updates), then it is faster to re-map
+                 *     from file during VMA restore in the child instead of sending memory content.
+                 *
+                 *   - If VMA is not private (i.e., it is MAP_SHARED), then we must re-map from file
+                 *     during VMA restore in the child instead of from sent memory content. Because
+                 *     updates in the parent are propagated to the file, the child will see the same
+                 *     VMA content after re-map as the parent.
+                 *
+                 *   - Only if VMA is both tainted and private, we must send the memory content.
+                 *     This VMA will not be backed by a file in the child (i.e., it will be
+                 *     recreated in the child as anonymous memory). We rely on the fact that private
+                 *     mappings do not reflect updates of the file.
+                 *
+                 * FIXME: If file-backed VMA is not tainted but private, we currently re-map from
+                 *        file during restore in the child. But it is possible that the file was
+                 *        modified by another process, and the child's VMA content will differ from
+                 *        the parent's VMA content. Currently we ignore this possibility.
+                 */
+                if ((vma->flags & (VMA_TAINTED | MAP_PRIVATE)) == (VMA_TAINTED | MAP_PRIVATE)) {
+                    uint64_t file_size = 0;
+                    get_file_size(vma->file, &file_size);
+
+                    if (vma->file_offset < file_size) {
+                        /* Access beyond the last file-backed page causes SIGBUS. To simulate this
+                         * behavior and to reduce fork latency, we truncate size of memory sent to
+                         * the child to the file size (adjusted by offset in the file), round up to
+                         * pages. */
+                        struct shim_mem_entry* mem;
+                        DO_CP_SIZE(memory, vma->addr,
+                                   MIN(vma->length, ALLOC_ALIGN_UP(file_size - vma->file_offset)),
+                                   &mem);
+                        mem->prot = LINUX_PROT_TO_PAL(vma->prot, /*map_flags=*/0);
+                    }
+
+                    /* private tainted mappings must never be re-mapped */
+                    remap_in_child_addr = vma->addr + vma->length;
                 }
             }
-            if (send_size > 0) {
-                struct shim_mem_entry* mem;
-                DO_CP_SIZE(memory, send_addr, send_size, &mem);
-                mem->prot = LINUX_PROT_TO_PAL(vma->prot, /*map_flags=*/0);
-
-                need_mapped = vma->addr + vma->length;
-            }
         }
+
         ADD_CP_FUNC_ENTRY(off);
-        ADD_CP_ENTRY(ADDR, need_mapped);
+        ADD_CP_ENTRY(ADDR, remap_in_child_addr);
     } else {
         new_vma = (struct shim_vma_info*)(base + off);
     }
@@ -1286,7 +1308,7 @@ END_CP_FUNC(vma)
 
 BEGIN_RS_FUNC(vma) {
     struct shim_vma_info* vma = (void*)(base + GET_CP_FUNC_ENTRY());
-    void* need_mapped = (void*)GET_CP_ENTRY(ADDR);
+    void* remap_in_child_addr = (void*)GET_CP_ENTRY(ADDR);
     CP_REBASE(vma->file);
 
     DEBUG_RS("vma: %p-%p flags 0x%x prot 0x%08x comment \"%s\"", vma->addr,
@@ -1302,39 +1324,41 @@ BEGIN_RS_FUNC(vma) {
             struct shim_fs* fs = vma->file->fs;
             get_handle(vma->file);
 
-            if (need_mapped < vma->addr + vma->length) {
-                /* first try, use hstat to force it resumes pal handle */
+            if (remap_in_child_addr < vma->addr + vma->length) {
                 if (!fs || !fs->fs_ops || !fs->fs_ops->mmap) {
                     return -EINVAL;
                 }
 
-                void* addr = need_mapped;
-                int ret = fs->fs_ops->mmap(vma->file, &addr, vma->addr + vma->length - need_mapped,
+                void* addr = remap_in_child_addr;
+                int ret = fs->fs_ops->mmap(vma->file, &addr,
+                                           vma->addr + vma->length - remap_in_child_addr,
                                            vma->prot, vma->flags | MAP_FIXED,
-                                           vma->file_offset + (need_mapped - vma->addr));
+                                           vma->file_offset + (remap_in_child_addr - vma->addr));
 
                 if (ret < 0)
                     return ret;
                 if (!addr)
                     return -ENOMEM;
-                if (addr != need_mapped)
+                if (addr != remap_in_child_addr)
                     return -EACCES;
 
-                need_mapped += vma->length;
+                remap_in_child_addr += vma->length;
             }
         }
 
-        if (need_mapped < vma->addr + vma->length) {
-            int ret = DkVirtualMemoryAlloc(need_mapped, vma->addr + vma->length - need_mapped,
+        if (remap_in_child_addr < vma->addr + vma->length) {
+            int ret = DkVirtualMemoryAlloc(remap_in_child_addr,
+                                           vma->addr + vma->length - remap_in_child_addr,
                                            /*alloc_type=*/0,
                                            LINUX_PROT_TO_PAL(vma->prot, /*map_flags=*/0));
             if (ret >= 0) {
-                need_mapped += vma->length;
+                remap_in_child_addr += vma->length;
             }
         }
 
-        if (need_mapped < vma->addr + vma->length) {
-            log_error("vma %p-%p cannot be allocated!", need_mapped, vma->addr + vma->length);
+        if (remap_in_child_addr < vma->addr + vma->length) {
+            log_error("vma %p-%p cannot be allocated!", remap_in_child_addr,
+                      vma->addr + vma->length);
             return -ENOMEM;
         }
     }
--- a/LibOS/shim/test/ltp/ltp.cfg
+++ b/LibOS/shim/test/ltp/ltp.cfg
@@ -245,24 +245,9 @@ skip = yes
 [eventfd2_*]
 skip = yes
 
-# BROK: Test haven't reported results
-[execl01]
-must-pass =
-    1
-
 # error while loading libc.so.6
 [execle01]
 skip = yes
-
-# BROK: Test haven't reported results
-[execlp01]
-must-pass =
-    1
-
-# BROK: Test haven't reported results
-[execv01]
-must-pass =
-    1
 
 # error while loading libc.so.6
 [execve01]
@@ -291,11 +276,6 @@ skip = yes
 # copy child
 [execveat03]
 skip = yes
-
-# BROK: Test haven't reported results
-[execvp01]
-must-pass =
-    1
 
 [faccessat01]
 timeout = 80
@@ -1226,12 +1206,6 @@ skip = yes
 [nanosleep01]
 skip = yes
 
-# reports TBROK due to lack of reported results (no shared memory in Gramine)
-[nanosleep02]
-must-pass =
-    1
-    2
-
 [nftw01]
 skip = yes
 
@@ -2075,10 +2049,6 @@ must-pass =
 
 [setrlimit04]
 skip = yes
-
-[setrlimit05]
-must-pass =
-    1
 
 [setrlimit06]
 skip = yes
--- a/LibOS/shim/test/regression/meson.build
+++ b/LibOS/shim/test/regression/meson.build
@@ -56,6 +56,7 @@ tests = {
     'madvise': {},
     'mkfifo': {},
     'mmap_file': {},
+    'mmap_file_backed': {},
     'mprotect_file_fork': {},
     'mprotect_prot_growsdown': {},
     'multi_pthread': {},
--- a/LibOS/shim/test/regression/mmap_file_backed.c
+++ b/LibOS/shim/test/regression/mmap_file_backed.c
@@ -1,6 +1,5 @@
 #define _GNU_SOURCE
 #include <err.h>
-#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <sys/mman.h>
@@ -14,8 +13,8 @@ int main(int argc, char** argv) {
     if (page_size < 0)
         err(1, "sysconf");
 
-    /* open the file (this executable's manifest) for read, find its size, mmap (with read-write
-     * protections) at least one page more than the file size and mprotect the last page */
+    /* open the file (this executable) for read, find its size, mmap (with read-write protections)
+     * at least one page more than the file size and mprotect the last page */
     FILE* fp = fopen(argv[0], "r");
     if (!fp)
         err(1, "fopen");
--- a/LibOS/shim/test/regression/mmap_file_backed.c
+++ b/LibOS/shim/test/regression/mmap_file_backed.c
@@ -1,0 +1,76 @@
+#define _GNU_SOURCE
+#include <err.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <sys/mman.h>
+#include <sys/wait.h>
+#include <unistd.h>
+
+int main(int argc, char** argv) {
+    int ret;
+
+    long page_size = sysconf(_SC_PAGESIZE);
+    if (page_size < 0)
+        err(1, "sysconf");
+
+    /* open the file (this executable's manifest) for read, find its size, mmap (with read-write
+     * protections) at least one page more than the file size and mprotect the last page */
+    FILE* fp = fopen(argv[0], "r");
+    if (!fp)
+        err(1, "fopen");
+
+    ret = fseek(fp, 0, SEEK_END);
+    if (ret < 0)
+        err(1, "fseek");
+
+    long fsize = ftell(fp);
+    if (fsize < 0)
+        err(1, "ftell");
+
+    rewind(fp); /* for sanity */
+
+    size_t mmap_size = fsize + page_size * 2; /* file size plus at least one full aligned page */
+    mmap_size &= ~(page_size - 1);            /* align down */
+
+    void* addr = mmap(NULL, mmap_size, PROT_READ | PROT_WRITE, MAP_PRIVATE, fileno(fp), 0);
+    if (addr == MAP_FAILED)
+        err(1, "mmap");
+
+    ret = mprotect(addr + mmap_size - page_size, page_size, PROT_NONE);
+    if (ret < 0)
+        err(1, "mprotect");
+
+    /* Below fork triggers checkpoint-and-restore logic in Gramine LibOS, which will send all VMAs
+     * info and all corresponding memory contents to the child. These VMAs contain two VMAs that
+     * were split from the file-backed mmap above: the first VMA with the lower part (backed by file
+     * contents) and the second VMA with a single page (not backed by file contents).
+     *
+     * There was a bug in Gramine that forced LibOS to re-map the memory-region content of the
+     * second VMA via mmap(..., <file-fd>, <offset-past-file-end>). This is incorrect because
+     * private writable mappings must inherit memory content from the parent. It led to specific
+     * checks for trusted/protected files on Linux-SGX to fail. So the below fork checks that this
+     * bug was fixed (otherwise Gramine crashes). */
+    int pid = fork();
+    if (pid == -1)
+        err(1, "fork");
+
+    if (pid != 0) {
+        /* parent */
+        int st = 0;
+        ret = wait(&st);
+        if (ret < 0)
+            err(1, "wait");
+
+        if (!WIFEXITED(st) || WEXITSTATUS(st) != 0)
+            errx(1, "abnormal child termination: %d\n", st);
+
+        puts("Parent process done");
+    } else {
+        /* child does nothing interesting */
+        puts("Child process done");
+    }
+
+    fclose(fp);
+    return 0;
+}
--- a/LibOS/shim/test/regression/test_libos.py
+++ b/LibOS/shim/test/regression/test_libos.py
@@ -649,7 +649,12 @@ class TC_30_Syscall(RegressionTestCase):
         self.assertIn('mmap test 5 passed', stdout)
         self.assertIn('mmap test 8 passed', stdout)
 
-    def test_052_large_mmap(self):
+    def test_052_mmap_file_backed(self):
+        stdout, _ = self.run_binary(['mmap_file_backed'], timeout=60)
+        self.assertIn('Child process done', stdout)
+        self.assertIn('Parent process done', stdout)
+
+    def test_053_large_mmap(self):
         try:
             stdout, _ = self.run_binary(['large_mmap'], timeout=480)
 
@@ -663,17 +668,17 @@ class TC_30_Syscall(RegressionTestCase):
             # This test generates a 4 GB file, don't leave it in FS.
             os.remove('testfile')
 
-    def test_053_mprotect_file_fork(self):
+    def test_054_mprotect_file_fork(self):
         stdout, _ = self.run_binary(['mprotect_file_fork'])
 
         self.assertIn('Test successful!', stdout)
 
-    def test_054_mprotect_prot_growsdown(self):
+    def test_055_mprotect_prot_growsdown(self):
         stdout, _ = self.run_binary(['mprotect_prot_growsdown'])
 
         self.assertIn('TEST OK', stdout)
 
-    def test_055_madvise(self):
+    def test_056_madvise(self):
         stdout, _ = self.run_binary(['madvise'])
         self.assertIn('TEST OK', stdout)
 
--- a/LibOS/shim/test/regression/tests.toml
+++ b/LibOS/shim/test/regression/tests.toml
@@ -57,6 +57,7 @@ manifests = [
   "madvise",
   "mkfifo",
   "mmap_file",
+  "mmap_file_backed",
   "mprotect_file_fork",
   "mprotect_prot_growsdown",
   "multi_pthread",
--- a/LibOS/shim/test/regression/tests_musl.toml
+++ b/LibOS/shim/test/regression/tests_musl.toml
@@ -59,6 +59,7 @@ manifests = [
   "madvise",
   "mkfifo",
   "mmap_file",
+  "mmap_file_backed",
   "mprotect_file_fork",
   "mprotect_prot_growsdown",
   "multi_pthread",


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/338)* status: 0 of 6 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

___
*[LibOS/shim/src/bookkeep/shim_vma.c, line 1239 at r1](https://reviewable.io/reviews/gramineproject/gramine/338#-MtNWmGxAvETAoZaT5Oe:-MtNWmGy4jnq-cvR2eeg:b-aztauf) ([raw file](https://github.com/gramineproject/gramine/blob/9544bb1bfa58c4fc198c071b70139d6c32c4e15a/LibOS/shim/src/bookkeep/shim_vma.c#L1239)):*
> ```C
>             DO_CP(handle, vma->file, &new_vma->file);
> 
>         void* need_mapped = vma->addr;
> ```

FYI: I renamed `need_mapped` to `remap_in_child_addr`. Actually, the current logic is binary (need remap or don't need remap), so it could be a boolean. But I kept as an address, to reduce the number of changes.

boryspoplawski (changes requested):
> Reviewed 4 of 6 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/338)* status: all files reviewed, 10 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[LibOS/shim/src/bookkeep/shim_vma.c, line 1239 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtQ1puN3t0bGY9-J6BN:-MtQ1puO62ofcCeUgYiZ:b-b706pn) ([raw file](https://github.com/gramineproject/gramine/blob/cce776e4410006186ee945fde9d34b1df393cbe9/LibOS/shim/src/bookkeep/shim_vma.c#L1239)):*
> ```C
>             DO_CP(handle, vma->file, &new_vma->file);
> 
>         /* VMA contents may be checkpointed, sent and restored via shim_mem_entry mechanism below;
> ```

Which case was changed in this file? The only change I see is handling of `MAP_SHARED`, which should be unrelated to the issue you had with python (which used `MAP_PRIVATE`).
___
*[LibOS/shim/src/bookkeep/shim_vma.c, line 1269 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtP5c39E357jJtGsVfN:-MtP5c39E357jJtGsVfO:b-a074f9) ([raw file](https://github.com/gramineproject/gramine/blob/cce776e4410006186ee945fde9d34b1df393cbe9/LibOS/shim/src/bookkeep/shim_vma.c#L1269)):*
> ```C
>                  *     This VMA will not be backed by a file in the child (i.e., it will be
>                  *     recreated in the child as anonymous memory). We rely on the fact that private
>                  *     mappings do not reflect updates of the file.
> ```

The last sentence is not always true (on Linux). Also, in here we rely on the fact that Linux does COW and we kind of do the same.
The difference is that we mark VMA as tainted if it's writable and Linux does COW on actual writes, so we will have a different behavior on VMAs that are writable, but were yet not COWed.
___
*[LibOS/shim/src/bookkeep/shim_vma.c, line 1274 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtP6FMI71vqBx8alcG5:-MtP6FMI71vqBx8alcG6:b-2uljbh) ([raw file](https://github.com/gramineproject/gramine/blob/cce776e4410006186ee945fde9d34b1df393cbe9/LibOS/shim/src/bookkeep/shim_vma.c#L1274)):*
> ```C
>                  *        file during restore in the child. But it is possible that the file was
>                  *        modified by another process, and the child's VMA content will differ from
>                  *        the parent's VMA content. Currently we ignore this possibility.
> ```

Why differ? If VMA is not tainted, then it cannot have been COWed, so all updates to the file will be visible via that mapping.
___
*[LibOS/shim/src/bookkeep/shim_vma.c, line 1278 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtP6Lxv7jq6uQEQ8aM5:-MtP6Lxw63jf1YUoJK49:b8gpaii) ([raw file](https://github.com/gramineproject/gramine/blob/cce776e4410006186ee945fde9d34b1df393cbe9/LibOS/shim/src/bookkeep/shim_vma.c#L1278)):*
> ```C
>                 if ((vma->flags & (VMA_TAINTED | MAP_PRIVATE)) == (VMA_TAINTED | MAP_PRIVATE)) {
>                     uint64_t file_size = 0;
>                     get_file_size(vma->file, &file_size);
> ```

Needs error checking
___
*[LibOS/shim/src/bookkeep/shim_vma.c, line 1290 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtPIRNM6og6Zbb0ZpF8:-MtPIRNM6og6Zbb0ZpF9:b5y1qsw) ([raw file](https://github.com/gramineproject/gramine/blob/cce776e4410006186ee945fde9d34b1df393cbe9/LibOS/shim/src/bookkeep/shim_vma.c#L1290)):*
> ```C
>                                    &mem);
>                         mem->prot = LINUX_PROT_TO_PAL(vma->prot, /*map_flags=*/0);
>                     }
> ```

What about the rest of the pages (if `vma->length` is greater than the sent size)?
___
*[LibOS/shim/src/bookkeep/shim_vma.c, line 1292 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtQ3fGd4MtGwJ3uH8h_:-MtQ3fGd4MtGwJ3uH8ha:buzoqzx) ([raw file](https://github.com/gramineproject/gramine/blob/cce776e4410006186ee945fde9d34b1df393cbe9/LibOS/shim/src/bookkeep/shim_vma.c#L1292)):*
> ```C
>                     }
> 
>                     /* private tainted mappings must never be re-mapped */
> ```

Actually they should be remapped, if they are bigger than the underlying file. The problem is it's hard to check whether at some point file was big enough and actually memory was touched and COW happened.
___
*[LibOS/shim/test/regression/mmap_file_backed.c, line 3 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtQ6ApS-xjEs-7oOpmZ:-MtQ6ApS-xjEs-7oOpm_:b-s3hmjs) ([raw file](https://github.com/gramineproject/gramine/blob/cce776e4410006186ee945fde9d34b1df393cbe9/LibOS/shim/test/regression/mmap_file_backed.c#L3)):*
> ```C
> #define _GNU_SOURCE
> #include <err.h>
> #include <stdint.h>
> ```

Not needed?
___
*[LibOS/shim/test/regression/mmap_file_backed.c, line 17 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtQ6U8VEcscam4-ugod:-MtQ6U8VEcscam4-ugoe:b-ntl3c) ([raw file](https://github.com/gramineproject/gramine/blob/cce776e4410006186ee945fde9d34b1df393cbe9/LibOS/shim/test/regression/mmap_file_backed.c#L17)):*
> ```C
>         err(1, "sysconf");
> 
>     /* open the file (this executable's manifest) for read, find its size, mmap (with read-write
> ```

Without Gramine this would probably be the program binary itself (not manifest file). Although these tests are designed to check Gramine stuff, they are also valid and runnable directly on host.
___
*[LibOS/shim/test/regression/mmap_file_backed.c, line 49 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtQ9JNR2ZOcXNc_PZdP:-MtQ9JNR2ZOcXNc_PZdQ:b-61406d) ([raw file](https://github.com/gramineproject/gramine/blob/cce776e4410006186ee945fde9d34b1df393cbe9/LibOS/shim/test/regression/mmap_file_backed.c#L49)):*
> ```C
>      * contents) and the second VMA with a single page (not backed by file contents).
>      *
>      * There was a bug in Gramine that forced LibOS to re-map the memory-region content of the
> ```

But such forced remapping of 2nd VMA was correct.
___
*[LibOS/shim/test/regression/mmap_file_backed.c, line 51 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtQ9XzCEJ_lVcgmBnw5:-MtQ9XzCEJ_lVcgmBnw6:b-9gqe8m) ([raw file](https://github.com/gramineproject/gramine/blob/cce776e4410006186ee945fde9d34b1df393cbe9/LibOS/shim/test/regression/mmap_file_backed.c#L51)):*
> ```C
>      * There was a bug in Gramine that forced LibOS to re-map the memory-region content of the
>      * second VMA via mmap(..., <file-fd>, <offset-past-file-end>). This is incorrect because
>      * private writable mappings must inherit memory content from the parent. It led to specific
> ```

How would they inherit content from parent if there is no content?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/338)* status: all files reviewed, 10 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski and @dimakuv)

___
*[LibOS/shim/src/bookkeep/shim_vma.c, line 1239 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtQ1puN3t0bGY9-J6BN:-MtbMkBY8mTnTYjiBXnh:b81bcfu) ([raw file](https://github.com/gramineproject/gramine/blob/cce776e4410006186ee945fde9d34b1df393cbe9/LibOS/shim/src/bookkeep/shim_vma.c#L1239)):*

Which case was changed in this file? The only change I see is handling of `MAP_SHARED`, which should be unrelated to the issue you had with python (which used `MAP_PRIVATE`).
</blockquote></details>

The other change is `remap_in_child_addr = vma->addr + vma->length` unconditionally -- in the old code, it was only if `send_size > 0`. This is what got me rid of the SGX trusted/protected files bug (in the sense that it doesn't manifest now).
___
*[LibOS/shim/src/bookkeep/shim_vma.c, line 1269 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtP5c39E357jJtGsVfN:-MtbSEjKCRe_js3xUDmU:b-nmtgeg) ([raw file](https://github.com/gramineproject/gramine/blob/cce776e4410006186ee945fde9d34b1df393cbe9/LibOS/shim/src/bookkeep/shim_vma.c#L1269)):*

The last sentence is not always true (on Linux). Also, in here we rely on the fact that Linux does COW and we kind of do the same.
The difference is that we mark VMA as tainted if it's writable and Linux does COW on actual writes, so we will have a different behavior on VMAs that are writable, but were yet not COWed.
</blockquote></details>

Yes, I understand your point. We'll have a different behavior to Linux, yes, but I don't think we can do much better. See my other comments as well.

I will rephrase the last sentence. What I meant is more like "a sane POSIX-complying application will operate on this VMA as if this private mapping does not reflect updates of the file (which is not always true on Linux, since it uses COW semantics for file-backed mappings)".
___
*[LibOS/shim/src/bookkeep/shim_vma.c, line 1274 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtP6FMI71vqBx8alcG5:-MtbS0M43gmYh5QiUaGw:bbqy875) ([raw file](https://github.com/gramineproject/gramine/blob/cce776e4410006186ee945fde9d34b1df393cbe9/LibOS/shim/src/bookkeep/shim_vma.c#L1274)):*

Why differ? If VMA is not tainted, then it cannot have been COWed, so all updates to the file will be visible via that mapping.
</blockquote></details>

Good catch, you are right. I will remove this para.
___
*[LibOS/shim/src/bookkeep/shim_vma.c, line 1290 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtPIRNM6og6Zbb0ZpF8:-MtbOQVPEJmJ9U6M6ZCh:br2pj1h) ([raw file](https://github.com/gramineproject/gramine/blob/cce776e4410006186ee945fde9d34b1df393cbe9/LibOS/shim/src/bookkeep/shim_vma.c#L1290)):*

What about the rest of the pages (if `vma->length` is greater than the sent size)?
</blockquote></details>

These pages are not mapped in the child at all. Any access to these pages leads to a `memfault` in LibOS terminology, and there is a special-case for this to generate a SIGBUS: https://github.com/gramineproject/gramine/blob/b2055ab935bd410829b262ef241b13e487fb3382/LibOS/shim/src/bookkeep/shim_signal.c#L370

This is the same as in the old code (see on the left pane). You could argue that this is not exactly what Linux does, but I don't think we can do better. Also, I kinda rely on applications being POSIX-complying, and POSIX says: `MAP_RIVATE [...] It is unspecified whether  changes  made to the file after the mmap() call are visible in the mapped region.` So a sane application won't rely on the fact that the underlying file may be expanded, and then pages become accessible.
___
*[LibOS/shim/src/bookkeep/shim_vma.c, line 1292 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtQ3fGd4MtGwJ3uH8h_:-MtbPzQl3u4ObPJvvdzi:b-tmrpsv) ([raw file](https://github.com/gramineproject/gramine/blob/cce776e4410006186ee945fde9d34b1df393cbe9/LibOS/shim/src/bookkeep/shim_vma.c#L1292)):*

Actually they should be remapped, if they are bigger than the underlying file. The problem is it's hard to check whether at some point file was big enough and actually memory was touched and COW happened.
</blockquote></details>

I didn't understand your scenario. I guess you're saying that `VMA_TAINTED` doesn't really say that the memory was touched (`VMA_TAINTED` is more like "memory may be potentially touched"), so there could be two possibilities:
1.File-backed memory is marked as `VMA_TAINTED` but was not touched; in this case Linux would forward the contents of the underlying file to the child
2. Some pages of file-backed memory were actually touched; in this case Linux would copy these pages from parent to child, but the rest pages will be populated with the contents of the underlying file

Gramine-SGX cannot emulate COW and thus cannot distinguish between these cases. So I chose a conservative (and hopefully POSIX-complying) approach of copying all potentially touched memory from parent to child, without any remapping. I don't know how to emulate COW mechanics of Linux better...
___
*[LibOS/shim/test/regression/mmap_file_backed.c, line 49 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtQ9JNR2ZOcXNc_PZdP:-MtbT-G5ADGuF4M6kw3C:brvbzvd) ([raw file](https://github.com/gramineproject/gramine/blob/cce776e4410006186ee945fde9d34b1df393cbe9/LibOS/shim/test/regression/mmap_file_backed.c#L49)):*

But such forced remapping of 2nd VMA was correct.
</blockquote></details>

Well, I need to expand on this comment... I mean that we do not emulate Linux's COW semantics completely. Instead, we have a simpler logic (which is also POSIX complying): if the file-backed mapping was created with `PROT_WRITE | MAP_PRIVATE`, then it acts as though the whole mapping was modified by the parent process, and so it should not be re-mapped in the child but instead transferred in the checkpoint.
___
*[LibOS/shim/test/regression/mmap_file_backed.c, line 51 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtQ9XzCEJ_lVcgmBnw5:-MtbTyIR1_YKICiIPls5:b3s2689) ([raw file](https://github.com/gramineproject/gramine/blob/cce776e4410006186ee945fde9d34b1df393cbe9/LibOS/shim/test/regression/mmap_file_backed.c#L51)):*

How would they inherit content from parent if there is no content?
</blockquote></details>

In this case, the memory content is "nothing" (no real memory, which leads to SIGSEGV == translated into SIGBUS in LibOS handler).

Hm, my comment on the inner workings of Gramine is only confusing... Should I just remove the comment completely? Or maybe even this whole test file?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/338)* status: 4 of 7 files reviewed, 11 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[-- commits, line 11 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtbeG_D9SK-qp-1htK_):*
Blocking myself: must add that several LTP tests were enabled completely (instead of with `must-pass`).
___
*[LibOS/shim/src/bookkeep/shim_vma.c, line 1269 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtP5c39E357jJtGsVfN:-MtbaXo6BqJAsd90mXO4:bb6un1j) ([raw file](https://github.com/gramineproject/gramine/blob/cce776e4410006186ee945fde9d34b1df393cbe9/LibOS/shim/src/bookkeep/shim_vma.c#L1269)):*

Yes, I understand your point. We'll have a different behavior to Linux, yes, but I don't think we can do much better. See my other comments as well.

I will rephrase the last sentence. What I meant is more like "a sane POSIX-complying application will operate on this VMA as if this private mapping does not reflect updates of the file (which is not always true on Linux, since it uses COW semantics for file-backed mappings)".
</blockquote></details>

Done
___
*[LibOS/shim/src/bookkeep/shim_vma.c, line 1274 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtP6FMI71vqBx8alcG5:-Mtba97IDhPETjcRaemY:bb6un1j) ([raw file](https://github.com/gramineproject/gramine/blob/cce776e4410006186ee945fde9d34b1df393cbe9/LibOS/shim/src/bookkeep/shim_vma.c#L1274)):*

Good catch, you are right. I will remove this para.
</blockquote></details>

Done
___
*[LibOS/shim/src/bookkeep/shim_vma.c, line 1278 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtP6Lxv7jq6uQEQ8aM5:-MtbagDp-56TFbBwiNZL:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/cce776e4410006186ee945fde9d34b1df393cbe9/LibOS/shim/src/bookkeep/shim_vma.c#L1278)):*

Needs error checking
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/mmap_file_backed.c, line 3 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtQ6ApS-xjEs-7oOpmZ:-MtbaoNAEL5fs6N5B3V1:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/cce776e4410006186ee945fde9d34b1df393cbe9/LibOS/shim/test/regression/mmap_file_backed.c#L3)):*

Not needed?
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/mmap_file_backed.c, line 17 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtQ6U8VEcscam4-ugod:-MtbbLWK1p9gtJ5Tq4Iq:b6fzq32) ([raw file](https://github.com/gramineproject/gramine/blob/cce776e4410006186ee945fde9d34b1df393cbe9/LibOS/shim/test/regression/mmap_file_backed.c#L17)):*

Without Gramine this would probably be the program binary itself (not manifest file). Although these tests are designed to check Gramine stuff, they are also valid and runnable directly on host.
</blockquote></details>

Done. Oops, I left the stale comment. I already open the program binary itself, not the manifest file.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/338)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[-- commits, line 11 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtbeG_D9SK-qp-1htK_):*

Blocking myself: must add that several LTP tests were enabled completely (instead of with `must-pass`).
</blockquote></details>

Actually, why did these tests started to work? I thought the issue is that they mmap shared anonymous memory and depend on it being shared between parent and child after fork, which this PR doesnt change.
___
*[LibOS/shim/src/bookkeep/shim_vma.c, line 1290 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtPIRNM6og6Zbb0ZpF8:-Mtc4fMr4CrRycJavsY8:bbkqdhy) ([raw file](https://github.com/gramineproject/gramine/blob/cce776e4410006186ee945fde9d34b1df393cbe9/LibOS/shim/src/bookkeep/shim_vma.c#L1290)):*
> This is the same as in the old code (see on the left pane). 

And I think it was a bug, which now you've made "official" and documented.

> You could argue that this is not exactly what Linux does, but I don't think we can do better.

Why can't we do better? It would be enough to just mmap that file there (with offset > size).

> I kinda rely on applications being POSIX-complying,

But we are emulating Linux not POSIX

Generally speaking I'm not sure if it's worth fixing now, but I don't like that you are making the code depend on this behavior, instead of it being something that can be fixed if need be.
___
*[LibOS/shim/src/bookkeep/shim_vma.c, line 1292 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtQ3fGd4MtGwJ3uH8h_:-Mtc6CMVB5Xu4PX_aQp4:buy23du) ([raw file](https://github.com/gramineproject/gramine/blob/cce776e4410006186ee945fde9d34b1df393cbe9/LibOS/shim/src/bookkeep/shim_vma.c#L1292)):*
> Gramine-SGX cannot emulate COW and thus cannot distinguish between these cases. 

But this is LibOS, not SGX PAL.

>  I guess you're saying that VMA_TAINTED doesn't really say that the memory was touched 

Yes and no. What I meant is that the part that is flle-backed (offset < size) is lost case and we cannot do anything about it (at least at the moment), but the part that extends past the end of file could be easily handled (with just mmaping the file).
What I meant here is: imagine the following scenario:
1) File of size `0x2000` is mmaped RW (hence tainted)
2) It's faulted into memory.
3) File is truncated to size `0x1000`

Now we cannot check if the 2nd page should be mapped or not (at least in the current PAL API, on native Linux it should be doable) - that's what I meant.

Btw, imagine that in the 3rd step file is truncated to `0`. The current version will just silently skip sending all memory and generate SIGBUS in child, even though parent has such memory pages (could have even modified them).
___
*[LibOS/shim/src/bookkeep/shim_vma.c, line 1277 at r3](https://reviewable.io/reviews/gramineproject/gramine/338#-Mtc2ZlUACB1mBubne1p:-Mtc2ZlUACB1mBubne1q:b-tm6dxw) ([raw file](https://github.com/gramineproject/gramine/blob/ba4ae20aaa01486c3923b96f3a5f08486c74802f/LibOS/shim/src/bookkeep/shim_vma.c#L1277)):*
> ```C
>                     int ret = get_file_size(vma->file, &file_size);
>                     if (ret < 0)
>                         BUG();
> ```

But this is not a bug, it may happen (e.g. some host errors, or just malicious host). Why can't we just return the error?
___
*[LibOS/shim/test/regression/mmap_file_backed.c, line 49 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtQ9JNR2ZOcXNc_PZdP:-MtcCLAQDq0mleuA-kvP:btan5yr) ([raw file](https://github.com/gramineproject/gramine/blob/cce776e4410006186ee945fde9d34b1df393cbe9/LibOS/shim/test/regression/mmap_file_backed.c#L49)):*
> I mean that we do not emulate Linux's COW semantics completely.

In this case there was no COW involved - the 2nd VMA is past end of file so it's rather "try to load the file at page fault" than COW.

> which is also POSIX complying

But we are not emulating a libc mmap function, but a syscall (form Linux).

> and so it should not be re-mapped in the child but instead transferred in the checkpoint.

The problem is the above + that we cannot know which file size to use - the one at the time of mapping, at the time of checkpointing, or at the time of faulting (that one we cannot even obtain).

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/338)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/338#-MtcJjJt4zNrqJpy90Ni:-MtcJjJuCQP7tCcfHqLx:blpicvc) (no related file):*
@boryspoplawski I see the general difference in our approaches like this:
1. You want the LibOS component to behave as close to Linux as possible, expecting the PAL component to correctly emulate e.g. COW semantics. You consider PAL limitations as being irrelevant to LibOS implementation.
2. I tried to make LibOS aware of the SGX PAL limitation (impossiblity of COW semantics), but adding restrictions/assumptions on how the application behaves.

I think that my approach is reasonable, because the whole LibOS checkpoint-and-restore design is based on the impossibility of COW (and in general of shared memory among Gramine processes). Shouldn't we add some assumptions/simplifications in the LibOS design about no-host-sharing, no-COW semantics? Why is this PR generally wrong?

---
___
*[-- commits, line 11 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtbeG_D9SK-qp-1htK_):*

Actually, why did these tests started to work? I thought the issue is that they mmap shared anonymous memory and depend on it being shared between parent and child after fork, which this PR doesnt change.
</blockquote></details>

No, it's not *anonymous* memory, it is file-backed memory: https://github.com/gramineproject/gramine/blob/master/LibOS/shim/test/ltp/manifest.template#L17-L19

LTP uses `/dev/shm`. So because I modified the code to re-map shared file-backed memory, it started working (the LTP child process previously did `send_size = <file-size> -> ...send memory contents in checkpoint ... -> DkVirtualMemoryAlloc()` and so that memory wasn't reflected in the file; but now it does).
___
*[LibOS/shim/src/bookkeep/shim_vma.c, line 1290 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtPIRNM6og6Zbb0ZpF8:-MtcLRmS7XS5z1G2zZr3:b-7cu2b6) ([raw file](https://github.com/gramineproject/gramine/blob/cce776e4410006186ee945fde9d34b1df393cbe9/LibOS/shim/src/bookkeep/shim_vma.c#L1290)):*
> Why can't we do better? It would be enough to just mmap that file there (with offset > size).

But this relies on the specific implementation of the host too much. Aren't we shifting too much responsibility to the host (== to the PAL)? The Linux PAL will easily implement this, sure, but we can't even emulate this correctly on Linux-SGX PAL.

I think we need to agree on some limits on what LibOS can realistically expect from PALs. I consider it reasonable to limit the LibOS -- with the benefit that the PAL implementation (in particular, SGX PAL) will become easier.
___
*[LibOS/shim/src/bookkeep/shim_vma.c, line 1277 at r3](https://reviewable.io/reviews/gramineproject/gramine/338#-Mtc2ZlUACB1mBubne1p:-MtcId7E52H1ypKufZPf:b-wwndry) ([raw file](https://github.com/gramineproject/gramine/blob/ba4ae20aaa01486c3923b96f3a5f08486c74802f/LibOS/shim/src/bookkeep/shim_vma.c#L1277)):*

But this is not a bug, it may happen (e.g. some host errors, or just malicious host). Why can't we just return the error?
</blockquote></details>

I just looked at how other `CP()` functions are implemented. At least the ones I checked used `BUG()`, so I assumed our `CP()` macros do not correctly return errors. Let me verify it and fix to `return -EBLABLA` if possible.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/338)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/338#-MtcJjJt4zNrqJpy90Ni:-MtcXEzs5qe3Rvv0DKhS:bhxei58) (no related file):*
> You want the LibOS component to behave as close to Linux as possible, expecting the PAL component to correctly emulate e.g. COW semantics. You consider PAL limitations as being irrelevant to LibOS implementation

No, I do not assume that all PALs must magically handle COW, but I do assume that LibOS should be as generic as possible.

> You consider PAL limitations as being irrelevant to LibOS implementation.

Not irrelevant. I expect LibOS to be able to work with all PALs - these that have such a limitation and those who does not. This can of course influence LibOS code, but ideally not in a way that imposes such limitation on all PALs, even though originally they did not have it.

> because the whole LibOS checkpoint-and-restore design is based on the impossibility of COW (and in general of shared memory among Gramine processes). 

1. I don't see in which way design we have atm is based on COW (or lack of it) and shared memory.
2. I don't see why we should assume sharing memory is impossible on LibOS level. It's just limitation of one of the PALs (and I still hope that it will be fixed someday).

> Shouldn't we add some assumptions/simplifications in the LibOS design about no-host-sharing, no-COW semantics?

I don't think we should. Why would we? But we must consider scenarios/PALs/archs that do not allow for it (e.g. SGX atm), though that means we allow for PAL not to support it, not assume that all PALs do not support it.

---
___
*[-- commits, line 11 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtbeG_D9SK-qp-1htK_):*

No, it's not *anonymous* memory, it is file-backed memory: https://github.com/gramineproject/gramine/blob/master/LibOS/shim/test/ltp/manifest.template#L17-L19

LTP uses `/dev/shm`. So because I modified the code to re-map shared file-backed memory, it started working (the LTP child process previously did `send_size = <file-size> -> ...send memory contents in checkpoint ... -> DkVirtualMemoryAlloc()` and so that memory wasn't reflected in the file; but now it does).
</blockquote></details>

Ah, `/dev/shm`
___
*[LibOS/shim/src/bookkeep/shim_vma.c, line 1290 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtPIRNM6og6Zbb0ZpF8:-MtcZpjhAlDGOTppdk3X:bxzzii8) ([raw file](https://github.com/gramineproject/gramine/blob/cce776e4410006186ee945fde9d34b1df393cbe9/LibOS/shim/src/bookkeep/shim_vma.c#L1290)):*
> The Linux PAL will easily implement this, sure, but we can't even emulate this correctly on Linux-SGX PAL.

But SGX PAL cannot emulate it no matter what, because all memory is always mapped. So even the current solution does not work.

> I think we need to agree on some limits on what LibOS can realistically expect from PALs. I consider it reasonable to limit the LibOS -- with the benefit that the PAL implementation (in particular, SGX PAL) will become easier.

So instead of having one working PAL we have none. Yes, sometimes we might want to limit our expectations, but we should not assume the opposite - that something is always not supported. Like shared memory - it's not supported on SGX, but I don't think it should be assumed in LibOS. Didn't you have some PoC/WIP support for sharing memory with some devices? The fact that it does not support something shouldn't affect the generic parts, especially since the support might come someday, even if only partial or for some specific usecases.
___
*[LibOS/shim/src/bookkeep/shim_vma.c, line 1277 at r3](https://reviewable.io/reviews/gramineproject/gramine/338#-Mtc2ZlUACB1mBubne1p:-MtcZSl_ANpc_kQyCZlQ:btap53x) ([raw file](https://github.com/gramineproject/gramine/blob/ba4ae20aaa01486c3923b96f3a5f08486c74802f/LibOS/shim/src/bookkeep/shim_vma.c#L1277)):*

I just looked at how other `CP()` functions are implemented. At least the ones I checked used `BUG()`, so I assumed our `CP()` macros do not correctly return errors. Let me verify it and fix to `return -EBLABLA` if possible.
</blockquote></details>

Which one does use `BUG()` and on what condition? These I've seen used `return -EXXXX`. `BUG()` should be used only in cases where something impossible (in bug free Gramine) happened.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/338)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/338#-MtcJjJt4zNrqJpy90Ni:-MtgJRk76pGNpNxSBNNX:b-ghxs29) (no related file):*
> I don't see in which way design we have atm is based on COW (or lack of it) and shared memory.

Gramine assumes "no COW mechanics" on the host for fork. Otherwise, Gramine could just have a `DkFork()` instead of all the shenanigans with collecting a proper checkpoint blob. I mean, if Gramine could simply rely on the host-level `fork()`, Gramine could remove the whole subsystem from its implementation.

Anyway, I think we're going astray with this discussion.

> But we must consider scenarios/PALs/archs that do not allow for it (e.g. SGX atm), though that means we allow for PAL not to support it, not assume that all PALs do not support it.

Ok, this is a good argument, I get your point now.

---
___
*[LibOS/shim/src/bookkeep/shim_vma.c, line 1292 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtQ3fGd4MtGwJ3uH8h_:-Mtga5tvB4hmB1zJjsto:b-mjtahl) ([raw file](https://github.com/gramineproject/gramine/blob/cce776e4410006186ee945fde9d34b1df393cbe9/LibOS/shim/src/bookkeep/shim_vma.c#L1292)):*

> Gramine-SGX cannot emulate COW and thus cannot distinguish between these cases. 

But this is LibOS, not SGX PAL.

>  I guess you're saying that VMA_TAINTED doesn't really say that the memory was touched 

Yes and no. What I meant is that the part that is flle-backed (offset < size) is lost case and we cannot do anything about it (at least at the moment), but the part that extends past the end of file could be easily handled (with just mmaping the file).
What I meant here is: imagine the following scenario:
1) File of size `0x2000` is mmaped RW (hence tainted)
2) It's faulted into memory.
3) File is truncated to size `0x1000`

Now we cannot check if the 2nd page should be mapped or not (at least in the current PAL API, on native Linux it should be doable) - that's what I meant.

Btw, imagine that in the 3rd step file is truncated to `0`. The current version will just silently skip sending all memory and generate SIGBUS in child, even though parent has such memory pages (could have even modified them).
</blockquote></details>

Let's decide on what we want to do in your scenario. In the scenario, LibOS knows the following:
- a VMA that is file-backed, tainted and has the size of `0x2000`
- a file that has the size `0x1000`

LibOS doesn't know if the page `0x1000 - 0x2000` actually exists in process memory or not. It may (if your scenario happened) or may not be (if step 2 was omitted from your scenario).

So what do we do with this range `0x1000 - 0x2000`? Three options:
1. Ignore this range completely -- this is what the current code does. But this incorrectly handles your scenario (if the page was faulted into memory).
2. Send this memory in a checkpoint -- but if this page doesn't exist in process memory, then `write_exact()` during checkpointing this memory fails (it tries to access non-existing page).
3. Perform `chroot_mmap()` in the child process -- but then we could lose parent's modifications to this page (if the page was faulted into memory).

So what should we do? Looks like a deadend to me -- any option we choose is incorrect for some scenario. Or am I missing something?

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/338)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/338#-MtcJjJt4zNrqJpy90Ni:-MtgvIwf-2i8ZtM12Gbn:b-mbcq33) (no related file):*
> if Gramine could simply rely on the host-level fork(), Gramine could remove the whole subsystem from its implementation.

But Gramine cannot assume host OS supports fork... (see Windows)

---
___
*[LibOS/shim/src/bookkeep/shim_vma.c, line 1292 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtQ3fGd4MtGwJ3uH8h_:-MtgxGqxDq4ySIJ5-jIs:bc4g2sf) ([raw file](https://github.com/gramineproject/gramine/blob/cce776e4410006186ee945fde9d34b1df393cbe9/LibOS/shim/src/bookkeep/shim_vma.c#L1292)):*

Let's decide on what we want to do in your scenario. In the scenario, LibOS knows the following:
- a VMA that is file-backed, tainted and has the size of `0x2000`
- a file that has the size `0x1000`

LibOS doesn't know if the page `0x1000 - 0x2000` actually exists in process memory or not. It may (if your scenario happened) or may not be (if step 2 was omitted from your scenario).

So what do we do with this range `0x1000 - 0x2000`? Three options:
1. Ignore this range completely -- this is what the current code does. But this incorrectly handles your scenario (if the page was faulted into memory).
2. Send this memory in a checkpoint -- but if this page doesn't exist in process memory, then `write_exact()` during checkpointing this memory fails (it tries to access non-existing page).
3. Perform `chroot_mmap()` in the child process -- but then we could lose parent's modifications to this page (if the page was faulted into memory).

So what should we do? Looks like a deadend to me -- any option we choose is incorrect for some scenario. Or am I missing something?
</blockquote></details>

Yes, seems there is no good solution with the current state of affairs.
But we have also option 4 (idk if doable, but should be): check if memory is readable (we have similar functions for checking user app pointers) and send (do 2.) if it is, or 3. otherwise. But this would require some non trivial code to be added to PALs.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/338)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/338#-MtmUrGf9KAJ492c-zev:-MtmUrGf9KAJ492c-zew:b-bk9bxk) (no related file):*
I created a fix only for Linux-SGX PAL: https://github.com/gramineproject/gramine/pull/346. The LibOS regression test is moved there.

@boryspoplawski Is there anything useful in the current PR, that we can keep? Or should I just close this PR?

---
___
*[LibOS/shim/src/bookkeep/shim_vma.c, line 1292 at r2](https://reviewable.io/reviews/gramineproject/gramine/338#-MtQ3fGd4MtGwJ3uH8h_:-MtmV6b26eoSuv9axrwO:bbwm685) ([raw file](https://github.com/gramineproject/gramine/blob/cce776e4410006186ee945fde9d34b1df393cbe9/LibOS/shim/src/bookkeep/shim_vma.c#L1292)):*
:( Looks like we don't know how to fix it in a simple way. Let's keep it as-is for now. I created another PR #346 that fixes only Linux-SGX PAL stuff. I'm giving up on fixing VMA checkpointing for now.

> check if memory is readable (we have similar functions for checking user app pointers)

I don't think we have. If you mean `test_user_memory()`, it doesn't check *memory accesses*: https://github.com/gramineproject/gramine/blob/master/LibOS/shim/src/bookkeep/shim_signal.c#L393-L409

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/338)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/338#-MtmUrGf9KAJ492c-zev:-Mtmke9K1RAHVGRxn_CJ:b-xtbmx0) (no related file):*

I created a fix only for Linux-SGX PAL: https://github.com/gramineproject/gramine/pull/346. The LibOS regression test is moved there.

@boryspoplawski Is there anything useful in the current PR, that we can keep? Or should I just close this PR?
</blockquote></details>

Maybe handling of shared memory? Sounds like an useful change and allows for `/dev/shm` to work in LTP on Linux PAL.

---

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/338)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/338#-MtmUrGf9KAJ492c-zev:-Mtn38qc2o9Qwd9CQXWP:bvaoytn) (no related file):*

Maybe handling of shared memory? Sounds like an useful change and allows for `/dev/shm` to work in LTP on Linux PAL.
</blockquote></details>

Done in https://github.com/gramineproject/gramine/pull/349

---

dimakuv:
This PR was superseded by https://github.com/gramineproject/gramine/pull/349 and https://github.com/gramineproject/gramine/pull/346

PR #363, opened 2022-01-27
remove launch token dependency for DCAP enclaves

lead4good:
Looking into a C port for `gramine-sgx-get-token` to get rid of the python runtime dependency I discovered that DCAP enclaves do not need a real token and instead just depend on a dummy token. Looking further I found that the dummy token information is generated from the SIGSTRUCT, so instead of using the information from the dummy token struct I took it directly from the SIGSTRUCT.

This removes the dummy launch token of DCAP enclaves and hence the python runtime dependency of DCAP enclaves.

See also:
- https://github.com/gramineproject/gsc/issues/37
- https://github.com/gramineproject/gsc/issues/61
- https://github.com/gramineproject/gsc/issues/13

Related PR in GSC repo: https://github.com/gramineproject/gsc/pull/62

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/363)

### DIFF
--- a/Pal/src/host/Linux-SGX/sgx_framework.c
+++ b/Pal/src/host/Linux-SGX/sgx_framework.c
@@ -110,13 +110,25 @@ bool is_wrfsbase_supported(void) {
     return true;
 }
 
-int create_enclave(sgx_arch_secs_t* secs, sgx_arch_enclave_css_t* sigstruct) {
+int create_enclave(sgx_arch_secs_t* secs, sgx_arch_enclave_css_t* sigstruct,
+                   sgx_arch_token_t* token) {
     assert(secs->size && IS_POWER_OF_2(secs->size));
     assert(IS_ALIGNED(secs->base, secs->size));
 
-    secs->ssa_frame_size = SSA_FRAME_SIZE / g_page_size; /* SECS expects SSA frame size in pages */
-    secs->misc_select    = sigstruct->body.misc_select;
+    /* SECS expects SSA frame size in pages */
+    secs->ssa_frame_size = SSA_FRAME_SIZE / g_page_size;
+
+    /* DCAP platforms do not use SGX tokens so we get enclave attributes from SIGSTRUCT instead,
+     * but EPID platforms require SECS to be populated from the SGX token */
+#ifdef SGX_DCAP
+    __UNUSED(token);
+    secs->misc_select = sigstruct->body.misc_select;
     memcpy(&secs->attributes, &sigstruct->body.attributes, sizeof(sgx_attributes_t));
+#else
+    __UNUSED(sigstruct);
+    secs->misc_select = token->masked_misc_select_le;
+    memcpy(&secs->attributes, &token->body.attributes, sizeof(sgx_attributes_t));
+#endif
 
     /* Do not initialize secs->mr_signer and secs->mr_enclave here as they are
      * not used by ECREATE to populate the internal SECS. SECS's mr_enclave is
--- a/Pal/src/host/Linux-SGX/sgx_framework.c
+++ b/Pal/src/host/Linux-SGX/sgx_framework.c
@@ -110,13 +110,13 @@ bool is_wrfsbase_supported(void) {
     return true;
 }
 
-int create_enclave(sgx_arch_secs_t* secs, sgx_arch_token_t* token) {
+int create_enclave(sgx_arch_secs_t* secs, sgx_arch_enclave_css_t* sigstruct) {
     assert(secs->size && IS_POWER_OF_2(secs->size));
     assert(IS_ALIGNED(secs->base, secs->size));
 
     secs->ssa_frame_size = SSA_FRAME_SIZE / g_page_size; /* SECS expects SSA frame size in pages */
-    secs->misc_select    = token->masked_misc_select_le;
-    memcpy(&secs->attributes, &token->body.attributes, sizeof(sgx_attributes_t));
+    secs->misc_select    = sigstruct->body.misc_select;
+    memcpy(&secs->attributes, &sigstruct->body.attributes, sizeof(sgx_attributes_t));
 
     /* Do not initialize secs->mr_signer and secs->mr_enclave here as they are
      * not used by ECREATE to populate the internal SECS. SECS's mr_enclave is
--- a/Pal/src/host/Linux-SGX/sgx_internal.h
+++ b/Pal/src/host/Linux-SGX/sgx_internal.h
@@ -72,7 +72,8 @@ bool is_wrfsbase_supported(void);
 int read_enclave_token(int token_file, sgx_arch_token_t* token);
 int read_enclave_sigstruct(int sigfile, sgx_arch_enclave_css_t* sig);
 
-int create_enclave(sgx_arch_secs_t* secs, sgx_arch_enclave_css_t* sigstruct);
+int create_enclave(sgx_arch_secs_t* secs, sgx_arch_enclave_css_t* sigstruct,
+                   sgx_arch_token_t* token);
 
 enum sgx_page_type { SGX_PAGE_SECS, SGX_PAGE_TCS, SGX_PAGE_REG };
 int add_pages_to_enclave(sgx_arch_secs_t* secs, void* addr, void* user_addr, unsigned long size,
--- a/Pal/src/host/Linux-SGX/sgx_internal.h
+++ b/Pal/src/host/Linux-SGX/sgx_internal.h
@@ -72,7 +72,7 @@ bool is_wrfsbase_supported(void);
 int read_enclave_token(int token_file, sgx_arch_token_t* token);
 int read_enclave_sigstruct(int sigfile, sgx_arch_enclave_css_t* sig);
 
-int create_enclave(sgx_arch_secs_t* secs, sgx_arch_token_t* token);
+int create_enclave(sgx_arch_secs_t* secs, sgx_arch_enclave_css_t* sigstruct);
 
 enum sgx_page_type { SGX_PAGE_SECS, SGX_PAGE_TCS, SGX_PAGE_REG };
 int add_pages_to_enclave(sgx_arch_secs_t* secs, void* addr, void* user_addr, unsigned long size,
--- a/Pal/src/host/Linux-SGX/sgx_main.c
+++ b/Pal/src/host/Linux-SGX/sgx_main.c
@@ -305,7 +305,7 @@ static int initialize_enclave(struct pal_enclave* enclave, const char* manifest_
     memset(&enclave_secs, 0, sizeof(enclave_secs));
     enclave_secs.base = enclave->baseaddr;
     enclave_secs.size = enclave->size;
-    ret = create_enclave(&enclave_secs, &enclave_sigstruct);
+    ret = create_enclave(&enclave_secs, &enclave_sigstruct, &enclave_token);
     if (ret < 0) {
         log_error("Creating enclave failed: %d", ret);
         goto out;
--- a/Pal/src/host/Linux-SGX/sgx_main.c
+++ b/Pal/src/host/Linux-SGX/sgx_main.c
@@ -14,6 +14,7 @@
 #include "asan.h"
 #include "debug_map.h"
 #include "gdb_integration/sgx_gdb.h"
+#include "gsgx.h"
 #include "linux_utils.h"
 #include "pal_internal-arch.h"
 #include "pal_linux_defs.h"
@@ -200,15 +201,20 @@ out:
 static int initialize_enclave(struct pal_enclave* enclave, const char* manifest_to_measure) {
     int ret = 0;
     int enclave_image = -1;
-    sgx_arch_token_t enclave_token;
-    sgx_arch_enclave_css_t enclave_sigstruct;
+    int enclave_mem = -1;
     sgx_arch_secs_t enclave_secs;
     unsigned long enclave_entry_addr;
     unsigned long enclave_heap_min;
+
+    sgx_arch_enclave_css_t enclave_sigstruct;
     char* sig_path = NULL;
+    int sigfile_fd = -1;
+
+    /* Launch Token (aka EINITTOKEN) is used only on EPID (non-FLC-based) platforms and completely
+     * ignored on DCAP (FLC-based) platforms */
+    sgx_arch_token_t enclave_token;
     char* token_path = NULL;
-    int sigfile_fd = -1, token_fd = -1;
-    int enclave_mem = -1;
+    int token_fd = -1;
 
     /* this array may overflow the stack, so we allocate it in BSS */
     static void* tcs_addrs[MAX_DBG_THREADS];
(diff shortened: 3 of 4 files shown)


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/363)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @lead4good)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/363#-MuQYSDW6Y6SbKnvFDQT:-MuQYSDW6Y6SbKnvFDQU:b-lmbhee) (no related file):*
@lead4good You are absolutely correct: tokens (files containing EINITTOKEN) are only needed for legacy non-FLC (Flexible Launch Control) SGX hardware. In the newer SGX hardware, FLC feature was added (which also coincides with the introduction of the DCAP driver and attestation mode).

Gramine strives to support both old and new SGX hardware, that's why we still rely on tokens (generate them, even if they are dummy).

Why do we always generate and read tokens in Gramine, even though they are not needed in DCAP SGX environments (and thus are mocked via a dummy token)? This is because we want to keep consistency of Gramine usage by higher-level tools and flows. One good example is GSC: https://github.com/gramineproject/gsc. GSC uses Gramine and its tools under the hood, and it would be very cumbersome to have two paths (e.g., two sets of GSC scripts): one for EPID with real tokens, and one for DCAP without tokens.

TLDR: By having dummy tokens, we make our lives easier -- tools/scripts that use Gramine can use the exact same paths/flows for both EPID and DCAP environments.

@lead4good Is there a particular reason you don't want to have this dummy token? Other than "striving for minimality"?

---
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/363#-MuQ_D-W9Ky5NC1pE1Cy:-MuQ_D-XALLopft7ERJW:b-n9h5c4) (no related file):*
The PR is generally correct but it doesn't go far enough: we could also remove some code from our Python tools (`gramine-sgx-get-token` as one example). What this PR is it removes the need to have the token file laying around when Gramine starts a workload. But the build process is unchanged, and thus Gramine will still generate a redundant token file. (Redundant for DCAP environments, I mean.)

---

lead4good (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/363)* status: 3 of 4 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/363#-MuQYSDW6Y6SbKnvFDQT:-MuUh6rpCMP0XpWxXxrf:b-jt8rez) (no related file):*

@lead4good You are absolutely correct: tokens (files containing EINITTOKEN) are only needed for legacy non-FLC (Flexible Launch Control) SGX hardware. In the newer SGX hardware, FLC feature was added (which also coincides with the introduction of the DCAP driver and attestation mode).

Gramine strives to support both old and new SGX hardware, that's why we still rely on tokens (generate them, even if they are dummy).

Why do we always generate and read tokens in Gramine, even though they are not needed in DCAP SGX environments (and thus are mocked via a dummy token)? This is because we want to keep consistency of Gramine usage by higher-level tools and flows. One good example is GSC: https://github.com/gramineproject/gsc. GSC uses Gramine and its tools under the hood, and it would be very cumbersome to have two paths (e.g., two sets of GSC scripts): one for EPID with real tokens, and one for DCAP without tokens.

TLDR: By having dummy tokens, we make our lives easier -- tools/scripts that use Gramine can use the exact same paths/flows for both EPID and DCAP environments.

@lead4good Is there a particular reason you don't want to have this dummy token? Other than "striving for minimality"?
</blockquote></details>

First and foremost the reason I supplied this patch is to remove the python runtime dependency (for DCAP, I only use DCAP, who uses EPID today?). There was a brief discussion here: https://github.com/gramineproject/gsc/issues/37 which also mentions the requirement to create tokens on the executing machine. I see no reason to restrict a user in such ways if it is not necessary. 

---
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/363#-MuQ_D-W9Ky5NC1pE1Cy:-MuUiuU9CR1zWsHXmSlD:bl8a6ev) (no related file):*

The PR is generally correct but it doesn't go far enough: we could also remove some code from our Python tools (`gramine-sgx-get-token` as one example). What this PR is it removes the need to have the token file laying around when Gramine starts a workload. But the build process is unchanged, and thus Gramine will still generate a redundant token file. (Redundant for DCAP environments, I mean.)
</blockquote></details>

I have removed the dummy function and added a hint for anyone using the get token tool.

---

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/363)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @lead4good)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/363#-MuQYSDW6Y6SbKnvFDQT:-MuUrFYR6tjyCsT7mxmT:b-5f5qph) (no related file):*
> for DCAP, I only use DCAP, who uses EPID today?

Yeah, it's generally true. EPID is slowly fading away. Our (Gramine devs) lifes would be much easier if EPID wouldn't exist at all :)

> to remove the python runtime dependency

This is a very good reason. I didn't think of it that way. Do you know how much you win by this, not installing Python on the deployment machine/VM/Docker image? Just curious.

---
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/363#-MuQ_D-W9Ky5NC1pE1Cy:-MuUrv9s8Uzzq0gkFoIu:b82h6ra) (no related file):*
<details><summary><i>Previously, lead4good wrote…</i></summary><blockquote>

I have removed the dummy function and added a hint for anyone using the get token tool.
</blockquote></details>

Yep, this looks good.

---
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/363#-MuUrx1tBRllt_kCMwjO:-MuUrx1tBRllt_kCMwjP:b6owp7m) (no related file):*
Since we're currently busy with other stuff (preparing for the new release), I suggest to put this PR on hold and then discuss what to do with it.

---
___
*[python/graminelibos/sgx_get_token.py, line 139 at r2](https://reviewable.io/reviews/gramineproject/gramine/363#-MuUr08q2EvtszJYgZR9:-MuUr08q2EvtszJYgZRA:bc3efsm) ([raw file](https://github.com/gramineproject/gramine/blob/cf3af818745d74592c177b9c453873cf77eaba5e/python/graminelibos/sgx_get_token.py#L139)):*
> ```Python
>                 'environments. To keep consistency of Gramine usage by higher-level tools and'
>                 'flows this tool will output an empty dummy token which can - but does not need to'
>                 'be - supplied during launch.'  )
> ```

I'll need to come up with a suggestion for a better phrasing. Keeping a blocking comment until I suggest some text.

lead4good (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/363)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @lead4good)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/363#-MuQYSDW6Y6SbKnvFDQT:-MuVApyC1W7EUHXcMtUQ:b-fa94jg) (no related file):*
> Do you know how much you win by this, not installing Python on the deployment machine/VM/Docker image?
No, I haven't checked yet. There are various other benefits though: Pycache files are an annoying source of non determinism, python related files don't clog the manifest file and finally we could now have a look at removing python as a "sign" dependency. If `gramine-sgx-sign` wouldn't depend on python, at least the buildpack that I'm working on, but maybe also GSC could drop python completely which makes them much less complex and much easier to port to different base images.

---
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/363#-MuUrx1tBRllt_kCMwjO:-MuVElD99td4X9r8C5W-:b-896fix) (no related file):*

Since we're currently busy with other stuff (preparing for the new release), I suggest to put this PR on hold and then discuss what to do with it.
</blockquote></details>

Done.

---

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/363)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/363#-MuQYSDW6Y6SbKnvFDQT:-N3T7Ebs4A5kvR73fFR2:b-h8j15w) (no related file):*
<details><summary><i>Previously, lead4good wrote…</i></summary><blockquote>

> Do you know how much you win by this, not installing Python on the deployment machine/VM/Docker image?
No, I haven't checked yet. There are various other benefits though: Pycache files are an annoying source of non determinism, python related files don't clog the manifest file and finally we could now have a look at removing python as a "sign" dependency. If `gramine-sgx-sign` wouldn't depend on python, at least the buildpack that I'm working on, but maybe also GSC could drop python completely which makes them much less complex and much easier to port to different base images.
</blockquote></details>

Thanks @lead4good for a detailed answer.

I like your idea of removing the Python dependency completely from Gramine (from the *enclave building* steps in Gramine), but I think this is hardly possible. Yes, we can remove Python dependencies for *enclave running* in Gramine, which is already a big win. But a lot of *building* utilities/flows are based on Python, and Gramine is actually moving more and more tools to Python.

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/363#-MuUrx1tBRllt_kCMwjO:-N3T71wI7DRaJf840M7P:bq5z825) (no related file):*
<details><summary><i>Previously, lead4good wrote…</i></summary><blockquote>

Done.
</blockquote></details>

The release (v1.2) was done last week, so we can revive this PR. Since @lead4good is not working on Gramine any more (I assume?), I will take over this PR and continue its development.

___

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/363)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/363#-N3TL9xv6ML1C6HYdgfw:-N3TL9xv6ML1C6HYdgfx:br3oirk) (no related file):*
I took over this PR, rebased it to the latest master and fixed some stuff. Please review the whole PR, not the latest diff (because of the rebase, it will be painful).

___
___
*[`Pal/src/host/Linux-SGX/sgx_main.c` line 217 at r3](https://reviewable.io/reviews/gramineproject/gramine/363#-N3TL0Qb4F3zXfuGE3MS:-N3TL0Qb4F3zXfuGE3MT:b-a1jtnk) ([raw file](https://github.com/gramineproject/gramine/blob/e6fe602913f7b57672ebce9dbfcd579c5698f69c/Pal/src/host/Linux-SGX/sgx_main.c#L217)):*
> ```C
>     sgx_arch_token_t enclave_token;
>     char* token_path = NULL;
>     int token_fd = -1;
> ```

FYI: This rearrangement of variable declarations is purely cosmetic. Reads better to me.
___
*[`python/graminelibos/sgx_get_token.py` line 139 at r2](https://reviewable.io/reviews/gramineproject/gramine/363#-MuUr08q2EvtszJYgZR9:-N3TI8Fw87w5PhykdI4v:b-3elygm) ([raw file](https://github.com/gramineproject/gramine/blob/cf3af818745d74592c177b9c453873cf77eaba5e/python/graminelibos/sgx_get_token.py#L139)):*

Can this also be under `if verbose`? For example, `gramine-test --sgx build` will generate hundreds of these warnings.
</blockquote></details>

Done. Also fixed the comment.

dimakuv:
Jenkins, test this please

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/363)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners

___
*[`Pal/src/host/Linux-SGX/sgx_framework.c` line 122 at r4](https://reviewable.io/reviews/gramineproject/gramine/363#-N3T_Xzq1NnX874s8FiT:-N3T_Xzq1NnX874s8FiU:bu3f6mv) ([raw file](https://github.com/gramineproject/gramine/blob/1bb6d14a37ed266d0a8dca8a9edf7fbb2844dfd9/Pal/src/host/Linux-SGX/sgx_framework.c#L122)):*
> ```C
> 
>     /* DCAP platforms do not use SGX tokens so we get enclave attributes from SIGSTRUCT instead,
>      * but EPID platforms require SECS to be populated from the SGX token */
> ```

FYI: On my previous try, the Jenkins CI (which uses EPID) failed all SGX tests with `error: enclave EINIT failed - Invalid enclave attribute`.

To be honest, I don't understand what exactly is bad about taking `attributes`/`misc_select` from SIGSTRUCT instead of the EINITTOKEN, but whatever, I added this distinction in the fixup commit. I don't want to debug this EPID craziness.

dimakuv:
Jenkins, test this please

dimakuv:
Closing this PR as it is considered as a stop-gap solution, and also it is not high priority. We better fix it properly.

Some context. This PR and https://github.com/gramineproject/gsc/pull/62 allow Gramine to ignore `.token` files for FLC/DCAP platforms. This has at least two benefits: (a) allows to run Gramine on read-only file systems, (b) allows to remove Python from run-time dependencies of Gramine.

The solution can be considered partial. Another proposal is more radical but also more correct: add a C/Rust version of `gramine-sgx-get-token` logic inside Gramine and force Gramine to generate the `.token` file when needed (on non-FLC/EPID platforms).

lead4good:
@dimakuv From my side it is fine to close this and work on a proper fix, as you suggest. For us it doesn't have very high priority as there is an easy work around. For DCAP environments, we just run `gramine-sgx-get-token` during build, it generates a platform independent token which subsequently can be used in the runtime environment, hence removing the `gramine-sgx-get-token` python dependency from the runtime.

mythi:
> `gramine-sgx-get-token` logic inside Gramine and force Gramine to generate the .token file when needed (on non-FLC/EPID platforms).

AFAICS, this PR could have re-purposed to that direction already: it implements the "when needed" logic and the DCAP part would have been already covered.

dimakuv:
> AFAICS, this PR could have re-purposed to that direction already: it implements the "when needed" logic and the DCAP part would have been already covered.

@mythi The huge missing piece in this PR is the re-write of the `gramine-sgx-get-token` utility from Python to a compiled language (our classic C or a Rust lib). Without this missing piece, the whole PR is not meaningful.

mythi:
> Without this missing piece, the whole PR is not meaningful.

For the DCAP path it is, no?

dimakuv:
>> Without this missing piece, the whole PR is not meaningful.

> For the DCAP path it is, no?

I see your point now. Yes, for DCAP it is enough. Still, it is better to submit a "full" PR that works both for DCAP and EPID.

PR #368, opened 2022-02-01
[Pal/Linux] Add a workaround for a kernel bug randomizing VDSO

boryspoplawski:
## Description of the changes 
On x64 Linux VDSO is randomized even if ASLR is disabled, but this bug manifests only on systems with 5-level paging. This commit adds a workaround for that bug.

Fixes #369

## How to test this PR? 
Manually run Gramine with Linux PAL on system with 5-level paging.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/368)

PR #388, opened 2022-02-08
[Pal/Linux] Introduce `sys.enable_host_vdso` manifest option

dimakuv:
## Description of the changes 

There is a bug in the Linux kernel: the placement of the vDSO library is randomized even when app is started with `ADDR_NO_RANDOMIZE`. This bug manifests itself on systems with 5-level paging enabled (e.g., newer Icelake Intel CPUs).

The only way to completely circumvent this bug is to not use the vDSO library. This PR changes default Gramine behavior to disable host vDSO and introduces `sys.enable_host_vdso` to re-enable it. This manifest option does not apply to the SGX environment (because it doesn't use host vDSO anyway).

For the initial attempt to fix this bug, see commit e7429e13bb3e9 "[Pal/Linux] Add a workaround for a kernel bug randomizing VDSO". Unfortunately, that bug fix does not eliminate the bug completely but only makes it less frequent.

## How to test this PR? 

CI. I added `sys.enable_host_vdso = true` in a couple tests/examples, just to test it.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/388)

### DIFF
--- a/CI-Examples/redis/redis-server.manifest.template
+++ b/CI-Examples/redis/redis-server.manifest.template
@@ -40,6 +40,15 @@ loader.env.LD_LIBRARY_PATH = "/lib:{{ arch_libdir }}:/usr/{{ arch_libdir }}"
 # Allow for injecting SIGTERM signal from the host.
 sys.enable_sigterm_injection = true
 
+################################ PERFORMANCE ##################################
+
+# The vDSO library is provided by the Linux kernel and may be used by Gramine to
+# improve performance of frequently-used system calls such as `clock_gettime()`.
+# By default, Gramine disables host vDSO because of the bug in Linux that may
+# lead to sporadic failures of multi-process applications. Since Redis is
+# single-process, we can safely enable host vDSO to improve performance.
+sys.enable_host_vdso = true
+
 ################################# MOUNT FS  ###################################
 
 # General notes:
--- a/Documentation/manifest-syntax.rst
+++ b/Documentation/manifest-syntax.rst
@@ -299,8 +299,8 @@ Allowing eventfd
     sys.insecure__allow_eventfd = [true|false]
     (Default: false)
 
-This specifies whether to allow system calls `eventfd()` and `eventfd2()`. Since
-eventfd emulation currently relies on the host, these system calls are
+This specifies whether to allow system calls ``eventfd()`` and ``eventfd2()``.
+Since eventfd emulation currently relies on the host, these system calls are
 disallowed by default due to security concerns.
 
 External SIGTERM injection
@@ -316,6 +316,26 @@ into Gramine. Could be useful to handle graceful shutdown.
 Be careful! In SGX environment, the untrusted host could inject that signal in
 an arbitrary moment. Examine what your application's `SIGTERM` handler does and
 whether it poses any security threat.
+
+Enabling host vDSO
+^^^^^^^^^^^^^^^^^^
+
+::
+
+    sys.enable_host_vdso = [true|false]
+    (Default: false)
+
+This specifies whether Gramine can use the host vDSO library to improve
+performance of frequently-used system calls such as ``clock_gettime()``.
+By default, Gramine does not use the host vDSO library because of the bug in
+Linux that may lead to sporadic failures of multi-process applications. You can
+safely enable this option for single-process applications.
+
+Note that currently this manifest option does *not* apply to the SGX environment
+(because it doesn't use host vDSO anyway).
+
+For technical details on this Linux bug, please see
+https://lore.kernel.org/lkml/9b82908f-eed1-b6b7-62aa-ecbba7bf048b@gmail.com/.
 
 Root FS mount point
 ^^^^^^^^^^^^^^^^^^^
--- a/LibOS/shim/test/regression/multi_pthread.manifest.template
+++ b/LibOS/shim/test/regression/multi_pthread.manifest.template
@@ -12,6 +12,9 @@ fs.mount.entrypoint.type = "chroot"
 fs.mount.entrypoint.path = "{{ entrypoint }}"
 fs.mount.entrypoint.uri = "file:{{ binary_dir }}/{{ entrypoint }}"
 
+# for testing only (multi_pthread is single-process so enabling host vDSO is definitely safe)
+sys.enable_host_vdso = true
+
 # app runs with 4 parallel threads + Gramine has couple internal threads
 sgx.thread_num = 8
 
--- a/Pal/src/host/Linux/db_main.c
+++ b/Pal/src/host/Linux/db_main.c
@@ -388,6 +388,17 @@ noreturn void pal_linux_main(void* initial_rsp, void* fini_callback) {
     }
     g_pal_internal_mem_addr = internal_mem_addr;
 
+    /*
+     * There is a bug in the Linux kernel: the placement of the vDSO library is randomized even
+     * though Gramine is started with `ADDR_NO_RANDOMIZE` in its personality (see the beginning of
+     * this function). This Linux bug manifests itself on systems with 5-level paging enabled (e.g.,
+     * newer Icelake Intel CPUs).
+     *
+     * The only way to completely circumvent this bug is to not use the vDSO library. The below code
+     * prevents Gramine from using host vDSO unless `sys.enable_host_vdso = true` is set in the
+     * manifest (i.e., Gramine ignores host vDSO). Note that the LibOS is still notified about the
+     * vDSO range (mapped by Linux on Gramine startup), see `add_preloaded_range()` above.
+     */
     bool enable_host_vdso;
     ret = toml_bool_in(g_pal_public_state.manifest_root, "sys.enable_host_vdso",
                        /*defaultval=*/false, &enable_host_vdso);
--- a/Pal/src/host/Linux/db_main.c
+++ b/Pal/src/host/Linux/db_main.c
@@ -299,12 +299,6 @@ noreturn void pal_linux_main(void* initial_rsp, void* fini_callback) {
     if (ret < 0)
         INIT_FAIL(unix_to_pal_error(-ret), "pal_thread_init() failed");
 
-    if (sysinfo_ehdr) {
-        ret = setup_vdso(sysinfo_ehdr);
-        if (ret < 0)
-            INIT_FAIL(-ret, "Setup of VDSO failed");
-    }
-
     uintptr_t vdso_start = 0;
     uintptr_t vdso_end = 0;
     uintptr_t vvar_start = 0;
@@ -393,6 +387,20 @@ noreturn void pal_linux_main(void* initial_rsp, void* fini_callback) {
         INIT_FAIL(PAL_ERROR_NOMEM, "Out of memory");
     }
     g_pal_internal_mem_addr = internal_mem_addr;
+
+    bool enable_host_vdso;
+    ret = toml_bool_in(g_pal_public_state.manifest_root, "sys.enable_host_vdso",
+                       /*defaultval=*/false, &enable_host_vdso);
+    if (ret < 0) {
+        INIT_FAIL_MANIFEST(PAL_ERROR_INVAL, "Cannot parse 'sys.enable_host_vdso' "
+                                            "(the value must be `true` or `false`)");
+    }
+
+    if (sysinfo_ehdr && enable_host_vdso) {
+        ret = setup_vdso(sysinfo_ehdr);
+        if (ret < 0)
+            INIT_FAIL(-ret, "Setup of VDSO failed");
+    }
 
     /* call to main function */
     pal_main(instance_id, parent, first_thread, first_process ? argv + 3 : argv + 4, envp);
--- a/Pal/src/host/Linux/db_main.c
+++ b/Pal/src/host/Linux/db_main.c
@@ -356,10 +356,6 @@ noreturn void pal_linux_main(void* initial_rsp, void* fini_callback) {
     }
     assert(manifest);
 
-    /* This depends on `g_vdso_start` and `g_vdso_end`, so it must be called only after they were
-     * initialized. */
-    signal_setup(first_process, g_vdso_start, g_vdso_end);
-
     g_pal_common_state.raw_manifest_data = manifest;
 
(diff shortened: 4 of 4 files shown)


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/388)* status: 3 of 4 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[-- commits, line 4 at r1](https://reviewable.io/reviews/gramineproject/gramine/388#-MvNefdhAZFQzHluEWOL):*

Can you move this explanation (or at least a part of it) to a comment (in `db_main.c`)? If we're working around an issue, I think it's better to have a description next the code that does it. It will be easier to discover what the code does (and why), and harder to inadvertently break something by refactoring.
</blockquote></details>

Done. Decided to duplicate this message in the C code (with minor changes). I prefer to have all explanations in commit body messages, I find it very useful when git blaming.
___
*[Pal/src/host/Linux/db_main.c, line 228 at r1](https://reviewable.io/reviews/gramineproject/gramine/388#-MvNhet502RU8DinOjCQ:-MvNhet502RU8DinOjCR:b-ldnqik) ([raw file](https://github.com/gramineproject/gramine/blob/fac270b6214a5479ca94cd83418f94715ce93d11/Pal/src/host/Linux/db_main.c#L228)):*
> ```C
>              * data, including memory content, to the child process. Disable ASLR to prevent memory
>              * colliding with PAL executable (as it would get a new random address in the child). */
>             ret = DO_SYSCALL(personality, (unsigned int)ret | ADDR_NO_RANDOMIZE);
> ```

FYI: @pwmarcz This is where we set `ADDR_NO_RANDOMIZE` (disable ASLR for all Gramine processes).
___
*[Pal/src/host/Linux/db_main.c, line 318 at r1](https://reviewable.io/reviews/gramineproject/gramine/388#-MvNh_ab2oWrpWMjB3fc:-MvNh_ab2oWrpWMjB3fd:b1ldmky) ([raw file](https://github.com/gramineproject/gramine/blob/fac270b6214a5479ca94cd83418f94715ce93d11/Pal/src/host/Linux/db_main.c#L318)):*
> ```C
> 
>     if (g_vdso_start || g_vdso_end) {
>         ret = add_preloaded_range(g_vdso_start, g_vdso_end, "vdso");
> ```

FYI: @pwmarcz This is where we prevent LibOS from re-using the vDSO memory range.
___
*[Pal/src/host/Linux/db_main.c, line 400 at r1](https://reviewable.io/reviews/gramineproject/gramine/388#-MvNdyymD0B2iVdzmBaq:-MvNhKbqBcsFQjDzzglF:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/fac270b6214a5479ca94cd83418f94715ce93d11/Pal/src/host/Linux/db_main.c#L400)):*

So what happens if we don't call `setup_vdso`? I assume Linux already mapped the vDSO for us. Does that mean we ignore it and possibly map something else over that range later? Will Linux let us do that?

It would be good to have an explanation of what happens somewhere.
</blockquote></details>

Done.

pwmarcz (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/388)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[-- commits, line 4 at r1](https://reviewable.io/reviews/gramineproject/gramine/388#-MvNefdhAZFQzHluEWOL):*

Done. Decided to duplicate this message in the C code (with minor changes). I prefer to have all explanations in commit body messages, I find it very useful when git blaming.
</blockquote></details>

OK, thanks.
___
*[Pal/src/host/Linux/db_main.c, line 400 at r1](https://reviewable.io/reviews/gramineproject/gramine/388#-MvNdyymD0B2iVdzmBaq:-MvNkx0U7LSNzNZxC8rA:b4n0v2g) ([raw file](https://github.com/gramineproject/gramine/blob/fac270b6214a5479ca94cd83418f94715ce93d11/Pal/src/host/Linux/db_main.c#L400)):*
>Note that the LibOS is still notified about the vDSO range (mapped by Linux on Gramine startup), see `add_preloaded_range()` above. 

So we notify LibOS about the range and LibOS takes care not to map over it?

Then I'm still confused. I thought that the problem was because the vDSO range in the child process sometimes appeared at a location where LibOS wanted to restore an application mapping from parent.

But if LibOS is still prevented from mapping over that range, then this problem should still happen, regardless of whether PAL using vDSO or not.

What am I missing?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/388)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[Pal/src/host/Linux/db_main.c, line 400 at r1](https://reviewable.io/reviews/gramineproject/gramine/388#-MvNdyymD0B2iVdzmBaq:-MvNol2iEzob06JFg_xs:b-jdxst1) ([raw file](https://github.com/gramineproject/gramine/blob/fac270b6214a5479ca94cd83418f94715ce93d11/Pal/src/host/Linux/db_main.c#L400)):*
> Then I'm still confused. I thought that the problem was because the vDSO range in the child process sometimes appeared at a location where LibOS wanted to restore an application mapping from parent.

No, this was never a problem. The problem is described here: https://lore.kernel.org/lkml/9b82908f-eed1-b6b7-62aa-ecbba7bf048b@gmail.com/

There is a bug in the Linux kernel that affects Gramine (only Linux PAL) as follows:
1. Gramine (in `gramine-direct` mode) disables ASLR using `personality(ADDR_NO_RANDOMIZE)` syscall: this is required because Gramine needs to re-create the exact same memory mappings in the child process (which is not just forked, but `execve`d and thus Linux could randomize memory mappings if we wouldn't disable ASLR).
2. Gramine (in `gramine-direct` mode) also sets up a seccomp filter to disallow all `syscall` instructions that come from the application. In other words, only PAL and host vDSO code (more precisely, corresponding PAL and vDSO code-segment memory ranges) can issue raw `syscall` instructions, and all other parts of the process receive SIGILL, and then Gramine emulates the syscall inside the LibOS layer.

Now the seccomp filter is set up only in the first (parent) Gramine process. We write the seccomp filter in such a way that only `[pal_code_start, pal_code_end)` and `[vdso_code_start, vdso_code_end)` are allowed to issue `syscall` instructions. All children Gramine processes will inherit this seccomp filter, with the exact same syscall-allowed memory ranges.

Now it turns out that Linux randomizes the vDSO library *even when ADDR_NO_RANDOMIZE is set*. So the child Gramine process may discover that Linux put the vDSO code segment beyond the parent's `[vdso_code_start, vdso_code_end)` region. And boom -- when vDSO code tries to execute, Gramine receives SIGILL and is super-confused.

We thought that we circumvent this scenario in this PR: https://github.com/gramineproject/gramine/pull/368. This PR basically expands the memory range of the vDSO code segment, smth like: `[ALIGN_SIGNIFICANTLY_DOWN(vdso_code_start), ALIGN_SIGNIFICANTLY_UP(vdso_code_end))`.

This workaround PR works in 99% of the cases, but very rarely, Linux still puts the vDSO beyond this expanded memory region. And again boom -- Gramine receives SIGILL on the vDSO execution and fails.

So my PR disables requests to vDSO from Gramine at all. vDSO is still there, it is still mapped, but its code is never called.

pwmarcz (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/388)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[Pal/src/host/Linux/db_main.c, line 400 at r1](https://reviewable.io/reviews/gramineproject/gramine/388#-MvNdyymD0B2iVdzmBaq:-MvNyajTEKx9ExF-ET3E:b7bu79m) ([raw file](https://github.com/gramineproject/gramine/blob/fac270b6214a5479ca94cd83418f94715ce93d11/Pal/src/host/Linux/db_main.c#L400)):*

> Then I'm still confused. I thought that the problem was because the vDSO range in the child process sometimes appeared at a location where LibOS wanted to restore an application mapping from parent.

No, this was never a problem. The problem is described here: https://lore.kernel.org/lkml/9b82908f-eed1-b6b7-62aa-ecbba7bf048b@gmail.com/

There is a bug in the Linux kernel that affects Gramine (only Linux PAL) as follows:
1. Gramine (in `gramine-direct` mode) disables ASLR using `personality(ADDR_NO_RANDOMIZE)` syscall: this is required because Gramine needs to re-create the exact same memory mappings in the child process (which is not just forked, but `execve`d and thus Linux could randomize memory mappings if we wouldn't disable ASLR).
2. Gramine (in `gramine-direct` mode) also sets up a seccomp filter to disallow all `syscall` instructions that come from the application. In other words, only PAL and host vDSO code (more precisely, corresponding PAL and vDSO code-segment memory ranges) can issue raw `syscall` instructions, and all other parts of the process receive SIGILL, and then Gramine emulates the syscall inside the LibOS layer.

Now the seccomp filter is set up only in the first (parent) Gramine process. We write the seccomp filter in such a way that only `[pal_code_start, pal_code_end)` and `[vdso_code_start, vdso_code_end)` are allowed to issue `syscall` instructions. All children Gramine processes will inherit this seccomp filter, with the exact same syscall-allowed memory ranges.

Now it turns out that Linux randomizes the vDSO library *even when ADDR_NO_RANDOMIZE is set*. So the child Gramine process may discover that Linux put the vDSO code segment beyond the parent's `[vdso_code_start, vdso_code_end)` region. And boom -- when vDSO code tries to execute, Gramine receives SIGILL and is super-confused.

We thought that we circumvent this scenario in this PR: https://github.com/gramineproject/gramine/pull/368. This PR basically expands the memory range of the vDSO code segment, smth like: `[ALIGN_SIGNIFICANTLY_DOWN(vdso_code_start), ALIGN_SIGNIFICANTLY_UP(vdso_code_end))`.

This workaround PR works in 99% of the cases, but very rarely, Linux still puts the vDSO beyond this expanded memory region. And again boom -- Gramine receives SIGILL on the vDSO execution and fails.

So my PR disables requests to vDSO from Gramine at all. vDSO is still there, it is still mapped, but its code is never called.
</blockquote></details>

Thanks. I remember us adding the seccomp filter, but somehow I completely missed that this was the problem with vDSO.

Can you add that to the comment? Something like "We cannot use vDSO if its location is different in the child processes, because the first process sets up a seccomp filter that allows inline syscall instructions only from the initial vDSO range".

Also: don't we still have a problem with seccomp filter in this scenario? It sounds like the parent might set a filter with the original vDSO range whitelisted, but in child, that range will be free and might be used for some other code (that potentially contains inline syscalls). So the seccomp filter should not whitelist the original range.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/388)* status: 3 of 4 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[Pal/src/host/Linux/db_main.c, line 400 at r1](https://reviewable.io/reviews/gramineproject/gramine/388#-MvNdyymD0B2iVdzmBaq:-MvOB2nW8yS8lfwGraj2:b-56mgfn) ([raw file](https://github.com/gramineproject/gramine/blob/fac270b6214a5479ca94cd83418f94715ce93d11/Pal/src/host/Linux/db_main.c#L400)):*

Thanks. I remember us adding the seccomp filter, but somehow I completely missed that this was the problem with vDSO.

Can you add that to the comment? Something like "We cannot use vDSO if its location is different in the child processes, because the first process sets up a seccomp filter that allows inline syscall instructions only from the initial vDSO range".

Also: don't we still have a problem with seccomp filter in this scenario? It sounds like the parent might set a filter with the original vDSO range whitelisted, but in child, that range will be free and might be used for some other code (that potentially contains inline syscalls). So the seccomp filter should not whitelist the original range.
</blockquote></details>

Done. You are right, I forgot to remove the vDSO range from seccomp.

pwmarcz (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/388)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

___
*[Pal/src/host/Linux/db_main.c, line 400 at r1](https://reviewable.io/reviews/gramineproject/gramine/388#-MvNdyymD0B2iVdzmBaq:-MvObqOC9bqI21UDJs3Q:b-ucfknh) ([raw file](https://github.com/gramineproject/gramine/blob/fac270b6214a5479ca94cd83418f94715ce93d11/Pal/src/host/Linux/db_main.c#L400)):*

Done. You are right, I forgot to remove the vDSO range from seccomp.
</blockquote></details>

Thanks!

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/388)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/388#-MvQl6ibAAEevRm2NYfQ:-MvQl6ibAAEevRm2NYfR:b-2voe7s) (no related file):*
I've create a PR with an alternative approach: https://github.com/gramineproject/gramine/pull/392
I prefer it because it's simpler, doesn't introduce a new manifest option and does not disable VDSO by default (only where needed).

Let me know what you think about it.

---

dimakuv:
This PR is superseded by #392. Closing.

PR #392, opened 2022-02-08
[Pal/Linux] Disable VDSO support on systems with 5-level paging

boryspoplawski:
## Description of the changes 
Unfortunately workaround from e7429e13bb3e914c5f582ead4ca2e3f81d145a8f was not enough and we are left with no other option.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/392)

### DIFF
--- a/Pal/src/host/Linux/db_exception.c
+++ b/Pal/src/host/Linux/db_exception.c
@@ -158,26 +158,6 @@ static int setup_seccomp(uintptr_t vdso_start, uintptr_t vdso_end) {
         return -1;
     }
 
-#ifdef __x86_64__
-    /*
-     * Hack ahead.
-     * On x64 Linux VDSO is randomized even if ASLR is disabled. This bug does not manifest on
-     * systems with 4-level paging, because stack is located at the highest available user space
-     * address, which does not leave any space for VDSO to be mapped after the stack. Now on systems
-     * with 5-level paging, stack is mapped at the exact same location, but highest available user
-     * space address is much greater, leaving space for VDSO and making the randomizatoin trigger.
-     * Relevant code: https://elixir.bootlin.com/linux/v5.14/source/arch/x86/entry/vdso/vma.c#L312
-     */
-    if (vdso_start >= ((1ul << 47) - PAGE_SIZE)) {
-        /* Maximal randomization for VDSO is the size of page middle directory (PMD).
-         * In theory something could be mapped in this region (besides VDSO), but in practice (i.e.
-         * accounting for how the Linux kernel and Gramine map memory) it's impossible. */
-        const size_t PMD_SIZE = 1ul << 21;
-        vdso_start = ALIGN_DOWN(vdso_start, PMD_SIZE);
-        vdso_end = ALIGN_UP(vdso_end, PMD_SIZE);
-    }
-#endif
-
     uint32_t syscalls_code_begin_low = (uintptr_t)gramine_raw_syscalls_code_begin;
     uint32_t syscalls_code_begin_high = (uintptr_t)gramine_raw_syscalls_code_begin >> 32;
     uint32_t syscalls_code_end_low = (uintptr_t)gramine_raw_syscalls_code_end;
--- a/Pal/src/host/Linux/db_main.c
+++ b/Pal/src/host/Linux/db_main.c
@@ -13,6 +13,7 @@
 
 #include "api.h"
 #include "asan.h"
+#include "cpu.h"
 #include "debug_map.h"
 #include "elf/elf.h"
 #include "init.h"
@@ -299,7 +300,36 @@ noreturn void pal_linux_main(void* initial_rsp, void* fini_callback) {
     if (ret < 0)
         INIT_FAIL(unix_to_pal_error(-ret), "pal_thread_init() failed");
 
-    if (sysinfo_ehdr) {
+    bool disable_vdso = false;
+#ifdef __x86_64__
+    /*
+     * Hack ahead.
+     * On x64 Linux VDSO is randomized even if ASLR is disabled. This bug does not manifest on
+     * systems with 4-level paging, because stack is located at the highest available user space
+     * address, which does not leave any space for VDSO to be mapped after the stack. Now on systems
+     * with 5-level paging, stack is mapped at the exact same location, but highest available user
+     * space address is much greater, leaving space for VDSO and making the randomization trigger.
+     * Relevant code: https://elixir.bootlin.com/linux/v5.14/source/arch/x86/entry/vdso/vma.c#L312
+     * If VDSO location was randomized, we wouldn't be able to use it due to seccomp filter, which
+     * allows us to catch "syscall" instructions. See "db_exception.c" for more details.
+     */
+    uint32_t cpuid_7_0_values[4] = { 0 };
+    cpuid(7, 0, cpuid_7_0_values);
+    if (cpuid_7_0_values[CPUID_WORD_ECX] & (1u << 16)) {
+        /*
+         * `LA57` bit is set - CPU supports 5-level paging - we cannot use VDSO.
+         * Note that we only check CPU support, not that the kernel enabled it. Unfortunately,
+         * the only way to test it is reading `cr4` register, which is a privileged operation that
+         * cannot be done from ring 3. The Linux kernel enabled 5-level paging by default around
+         * version 5.5, so we assume most users either have this turned on or their CPU does not
+         * support it. In theory we could try mmaping something at a high address, but it would be
+         * cumbersome and we didn't bother.
+         */
+        disable_vdso = true;
+    }
+#endif
+
+    if (sysinfo_ehdr && !disable_vdso) {
         ret = setup_vdso(sysinfo_ehdr);
         if (ret < 0)
             INIT_FAIL(-ret, "Setup of VDSO failed");
@@ -364,7 +394,7 @@ noreturn void pal_linux_main(void* initial_rsp, void* fini_callback) {
 
     /* This depends on `g_vdso_start` and `g_vdso_end`, so it must be called only after they were
      * initialized. */
-    signal_setup(first_process, g_vdso_start, g_vdso_end);
+    signal_setup(first_process, disable_vdso ? 0 : g_vdso_start, disable_vdso ? 0 : g_vdso_end);
 
     g_pal_common_state.raw_manifest_data = manifest;
 


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/392)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski)

___
*[-- commits, line 4 at r1](https://reviewable.io/reviews/gramineproject/gramine/392#-MvSNX_T7-YXhUm892S9):*
Could we add a bit more info in the commit message. I suggest this para before your text:
```
  There is a bug in the Linux kernel: the placement of the vDSO library is
  randomized even if ASLR is disabled. This bug manifests itself on systems
  with 5-level paging enabled.
```

Otherwise it's unclear what this commit is about.
___
*[Pal/src/host/Linux/db_main.c, line 311 at r1](https://reviewable.io/reviews/gramineproject/gramine/392#-MvSNryV9iXXyp5Ke4Id:-MvSNryWBgLbf6VBVLA7:b937r4g) ([raw file](https://github.com/gramineproject/gramine/blob/96ca3aab8e70516221dd6bb8a8114e9339cabe5a/Pal/src/host/Linux/db_main.c#L311)):*
> ```C
>      * address, which does not leave any space for VDSO to be mapped after the stack. Now on systems
>      * with 5-level paging, stack is mapped at the exact same location, but highest available user
>      * space address is much greater, leaving space for VDSO and making the randomizatoin trigger.
> ```

typo: `randomization`

pwmarcz (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/392)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski)

___
*[Pal/src/host/Linux/db_main.c, line 316 at r1](https://reviewable.io/reviews/gramineproject/gramine/392#-MvT0Xy979fAbG2g5wb5:-MvT0XyAC5aVWvRM99n4:b-nfn3fa) ([raw file](https://github.com/gramineproject/gramine/blob/96ca3aab8e70516221dd6bb8a8114e9339cabe5a/Pal/src/host/Linux/db_main.c#L316)):*
> ```C
>      * allows us to catch "syscall" instructions. See "db_exception.c" for more details.
>      */
>     uint32_t cpuid_7_0_values[4] = { 0 };
> ```

This checks if 5-level paging is supported, not if it's enabled, right? I think that the kernel will usually be compiled without `CONFIG_X86_5LEVEL` (for instance, it's disabled in my generic Ubuntu 20 kernel).

Can we check if it's enabled somehow? Otherwise, we'll turn vdso off on any Ice Lake machine regardless of the kernel.

Alternatively (e.g. if it's too hard to check, or you think we shouldn't bother), I think the comment should acknowledge that.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/392)* status: 1 of 2 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "squash! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @pwmarcz)

___
*[-- commits, line 4 at r1](https://reviewable.io/reviews/gramineproject/gramine/392#-MvSNX_T7-YXhUm892S9):*

Could we add a bit more info in the commit message. I suggest this para before your text:
```
  There is a bug in the Linux kernel: the placement of the vDSO library is
  randomized even if ASLR is disabled. This bug manifests itself on systems
  with 5-level paging enabled.
```

Otherwise it's unclear what this commit is about.
</blockquote></details>

There is such info in the referred commit, but yeah, it wouldn't hurt to add it also here. Added a squash commit
___
*[Pal/src/host/Linux/db_main.c, line 311 at r1](https://reviewable.io/reviews/gramineproject/gramine/392#-MvSNryV9iXXyp5Ke4Id:-MvT8cE557-rmTTzZM3j:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/96ca3aab8e70516221dd6bb8a8114e9339cabe5a/Pal/src/host/Linux/db_main.c#L311)):*

typo: `randomization`
</blockquote></details>

Done.
___
*[Pal/src/host/Linux/db_main.c, line 316 at r1](https://reviewable.io/reviews/gramineproject/gramine/392#-MvT0Xy979fAbG2g5wb5:-MvT8fjv4tDLeRb4qXQn:b-h5108p) ([raw file](https://github.com/gramineproject/gramine/blob/96ca3aab8e70516221dd6bb8a8114e9339cabe5a/Pal/src/host/Linux/db_main.c#L316)):*
> This checks if 5-level paging is supported, not if it's enabled, right? 

Yes.

> I think that the kernel will usually be compiled without CONFIG_X86_5LEVEL (for instance, it's disabled in my generic Ubuntu 20 kernel).

I assumed the opposite - that on newer kernels it would be enabled even if the cpu does not support it - after all it doesn't hurt to do so, it's being setup dynamically at runtime anyway. It is true for my laptop (not Ubuntu though), and I have Ice Lake CPU (but it does not support 5-level paging? maybe it's just on server machines?)

> Can we check if it's enabled somehow? 

I wanted to do that from beginning, but it turns out the only way (or at least I didn't find any other)  is to read from `cr4`, which we cannot do in userspace.
We could try mmaping something at a high address, that wouldn't work with 4-level paging, but that sounds brittle...

So I think we have two options:
A) Just check cpuid and assume that if CPU supports 5-level paging, then kernel probably does too (currently implemented approach).
B) Try mmaping at high address and see if that works.

cc @dimakuv @mkow

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/392)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "squash! " found in commit messages' one-liners (waiting on @mkow)

___
*[Pal/src/host/Linux/db_main.c, line 316 at r1](https://reviewable.io/reviews/gramineproject/gramine/392#-MvT0Xy979fAbG2g5wb5:-MvTEZjiD5jUVBbiGuGC:bo5q1lv) ([raw file](https://github.com/gramineproject/gramine/blob/96ca3aab8e70516221dd6bb8a8114e9339cabe5a/Pal/src/host/Linux/db_main.c#L316)):*
> I wanted to do that from beginning, but it turns out the only way (or at least I didn't find any other)  is to read from cr4, which we cannot do in userspace.

Yes, this won't work, unfortunately. Maybe the Linux kernel exposes this support via `/proc/cpuinfo` or some other bit (e.g., in auxv)?

> B) Try mmaping at high address and see if that works.

But this doesn't give you the information we need, no?
- If we can mmap at high address, there is 5-level paging, we need to disable vDSO for sure.
- If we cannot mmap at high address, then *probably* there is no 5-level paging, but maybe it's because something else is located at this address? So we don't know whether we can keep vDSO enabled. (Even if we try 10 times at random addresses, where is the guarantee that we were unlucky 10 times?)

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/392)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "squash! " found in commit messages' one-liners (waiting on @mkow)

___
*[Pal/src/host/Linux/db_main.c, line 316 at r1](https://reviewable.io/reviews/gramineproject/gramine/392#-MvT0Xy979fAbG2g5wb5:-MvTHkd91WFc37LWwBi5:bx92b8r) ([raw file](https://github.com/gramineproject/gramine/blob/96ca3aab8e70516221dd6bb8a8114e9339cabe5a/Pal/src/host/Linux/db_main.c#L316)):*
>I have Ice Lake CPU (but it does not support 5-level paging? maybe it's just on server machines?)

Weird, but that would limit the impact of the current code.

And apparently [5-level paging is enabled by default starting from Linux 5.5](https://www.phoronix.com/scan.php?page=news_item&px=Linux-5.5-5-Level-Paging) (the generic kernel I checked is 5.4). I think Gramine users are likely to use a newer kernel because of the SGX driver.

So yeah, the current solution looks fine. It'd be nice to have a comment saying this is an approximation, but I'm not blocking on this.

pwmarcz (dismissed):
> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/392)* status: all files reviewed, 1 unresolved discussion, "squash! " found in commit messages' one-liners (waiting on @mkow)

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/392)* status: all files reviewed, 1 unresolved discussion, "squash! " found in commit messages' one-liners (waiting on @mkow)

___
*[Pal/src/host/Linux/db_main.c, line 316 at r1](https://reviewable.io/reviews/gramineproject/gramine/392#-MvT0Xy979fAbG2g5wb5:-MvTQgZn2kQ-PQ82ubH3:buswyt3) ([raw file](https://github.com/gramineproject/gramine/blob/96ca3aab8e70516221dd6bb8a8114e9339cabe5a/Pal/src/host/Linux/db_main.c#L316)):*
> /proc/cpuinfo

I don't have a machine with 5-level paging at hand so cannot test, but I'm not sure how it behaves when cpu supports something but kernel has it turned off (though it must support it, otherwise it wouldn't print it).

> But this doesn't give you the information we need, no?

It does, if done correctly. You can use `MAP_FIXED_NOREPLACE` (so you know that you hit something). You can also pick an address high enough, that there is enough space for everything (e.g. PAL, initial stack, vdso, etc) in range `[1<<47, picked_address)`, **not** specify `MAP_FIXED*` and assume that Linux works as it does - picks nearest free address.

> It'd be nice to have a comment saying this is an approximation, but I'm not blocking on this.

Yes, if we decide to keep it, I'll add a comment.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/392)* status: all files reviewed, 1 unresolved discussion, "squash! " found in commit messages' one-liners (waiting on @mkow)

___
*[Pal/src/host/Linux/db_main.c, line 316 at r1](https://reviewable.io/reviews/gramineproject/gramine/392#-MvT0Xy979fAbG2g5wb5:-MvTax1k3TZRmqK0-CRs:bvv13d3) ([raw file](https://github.com/gramineproject/gramine/blob/96ca3aab8e70516221dd6bb8a8114e9339cabe5a/Pal/src/host/Linux/db_main.c#L316)):*
> I don't have a machine with 5-level paging at hand so cannot test, but I'm not sure how it behaves when cpu supports something but kernel has it turned off (though it must support it, otherwise it wouldn't print it).

I checked on my machines:
```
# machine without 5-level paging, Linux 4.15
$ grep "address sizes" /proc/cpuinfo
address sizes   : 39 bits physical, 48 bits virtual

# machine with 5-level paging, Linux 5.12
$ grep "address sizes" /proc/cpuinfo
address sizes   : 46 bits physical, 57 bits virtual

# machine with 5-level paging, Linux 5.4
$ grep "address sizes" /proc/cpuinfo
address sizes   : 46 bits physical, 57 bits virtual
```

I verified that machine supports/doesn't support 5-level paging by looking at CPUID: `cpuid -1 -l 7 -s 0 -r`.

So it looks like we can use this `address sizes` string (additionally to the CPUID check) to decide whether to disable vDSO.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/392)* status: all files reviewed, 1 unresolved discussion, "squash! " found in commit messages' one-liners (waiting on @mkow)

___
*[Pal/src/host/Linux/db_main.c, line 316 at r1](https://reviewable.io/reviews/gramineproject/gramine/392#-MvT0Xy979fAbG2g5wb5:-MvTkIwv0Yiyox4twWlr:b-u0fkhn) ([raw file](https://github.com/gramineproject/gramine/blob/96ca3aab8e70516221dd6bb8a8114e9339cabe5a/Pal/src/host/Linux/db_main.c#L316)):*
> So it looks like we can use this address sizes string (additionally to the CPUID check) to decide whether to disable vDSO.

Wait, from what you've said this is exactly the same thing - it probably just comes directly from `cpuid`. If CPU supports 5-level paging it says "57 bits virtual" regardless of kernel support.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/392)* status: all files reviewed, 1 unresolved discussion, "squash! " found in commit messages' one-liners (waiting on @mkow)

___
*[Pal/src/host/Linux/db_main.c, line 316 at r1](https://reviewable.io/reviews/gramineproject/gramine/392#-MvT0Xy979fAbG2g5wb5:-MvTlWD4APpFNOpnb_rz:bvur7b8) ([raw file](https://github.com/gramineproject/gramine/blob/96ca3aab8e70516221dd6bb8a8114e9339cabe5a/Pal/src/host/Linux/db_main.c#L316)):*

> So it looks like we can use this address sizes string (additionally to the CPUID check) to decide whether to disable vDSO.

Wait, from what you've said this is exactly the same thing - it probably just comes directly from `cpuid`. If CPU supports 5-level paging it says "57 bits virtual" regardless of kernel support.
</blockquote></details>

My point is that "57 bits virtual" implies that the Linux kernel supports 5-level paging (not just the hardware which you query directly in this PR). So it is an additional proof point that Linux enables it (well, at least knows about it).

Unfortunately, I don't have some old hardware (without 5-level paging support) but with the new Linux kernel at hand. And I also don't have the hardware that has it the other way around (5-level paging support but very old kernel). Maybe you or Pawel have it and could check what is printed in this case?

But anyway I'm not blocking on this: even if we disable vDSO on most modern CPUs, I don't think this is a huge performance hit to Gramine. Also, it is only on Linux PAL, and we don't care that much about its performance.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/392)* status: all files reviewed, 1 unresolved discussion, "squash! " found in commit messages' one-liners (waiting on @mkow)

___
*[Pal/src/host/Linux/db_main.c, line 316 at r1](https://reviewable.io/reviews/gramineproject/gramine/392#-MvT0Xy979fAbG2g5wb5:-MvToAqt-ZxK279rkjRF:b-m5rwrv) ([raw file](https://github.com/gramineproject/gramine/blob/96ca3aab8e70516221dd6bb8a8114e9339cabe5a/Pal/src/host/Linux/db_main.c#L316)):*
> My point is that "57 bits virtual" implies that the Linux kernel supports 5-level paging (not just the hardware which you query directly in this PR). So it is an additional proof point that Linux enables it (well, at least knows about it).

Is that so? I think it only implies that CPU supports it, not that kernel enabled it. Try grepping kernel config for `CONFIG_PGTABLE_LEVELS` on machine `# machine with 5-level paging, Linux 5.4`.

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/392)* status: all files reviewed, 1 unresolved discussion, "squash! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[Pal/src/host/Linux/db_main.c, line 316 at r1](https://reviewable.io/reviews/gramineproject/gramine/392#-MvT0Xy979fAbG2g5wb5:-Mvavxtj-1NcPCmGpn3B:b-83d4l) ([raw file](https://github.com/gramineproject/gramine/blob/96ca3aab8e70516221dd6bb8a8114e9339cabe5a/Pal/src/host/Linux/db_main.c#L316)):*
> Yes, if we decide to keep it, I'll add a comment.

Blocking on this, otherwise I'm fine with leaving the CPUID version. Worst case some non-SGX Gramine will be slower on some systems where it doesn't need to, but we don't care much about its performance there.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/392)* status: all files reviewed, 1 unresolved discussion, "squash! " found in commit messages' one-liners (waiting on @boryspoplawski)

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/392)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski)

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/392)* status: 1 of 2 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv, @mkow, and @pwmarcz)

___
*[Pal/src/host/Linux/db_main.c, line 316 at r1](https://reviewable.io/reviews/gramineproject/gramine/392#-MvT0Xy979fAbG2g5wb5:-Mvd1maiB78O9uTCCerT:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/96ca3aab8e70516221dd6bb8a8114e9339cabe5a/Pal/src/host/Linux/db_main.c#L316)):*

> Yes, if we decide to keep it, I'll add a comment.

Blocking on this, otherwise I'm fine with leaving the CPUID version. Worst case some non-SGX Gramine will be slower on some systems where it doesn't need to, but we don't care much about its performance there.
</blockquote></details>

Done.

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/392)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

pwmarcz (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/392)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/392)* status: all files reviewed, 1 unresolved discussion, not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski)

___
*[Pal/src/host/Linux/db_main.c, line 323 at r4](https://reviewable.io/reviews/gramineproject/gramine/392#-Mvd47La3HH-c1NB-F3R:-Mvd47Lb-DNAOGkgt92d:b-iy99xa) ([raw file](https://github.com/gramineproject/gramine/blob/4cb5410c92b36b0dc2bda5110f53f60d37956183/Pal/src/host/Linux/db_main.c#L323)):*
> ```C
>          * Note that we only check CPU support, not that the kernel enabled it. Unfortunately,
>          * the only way to test it is reading `cr4` register, which is a privileged operation that
>          * cannot be done from ring 0. The Linux kernel enabled 5-level paging by default around
> ```

`from ring 3`

pwmarcz (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/392)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/392)* status: all files reviewed, 1 unresolved discussion, not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/392)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/392)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #398, opened 2022-02-11
[LibOS,Pal/Linux-SGX] Don't use RBX as operand in inline asm

### LINKED ISSUE #391: Inline assembly produces wrong results in Clang release mode
## Summary

Our inline assembly sometimes uses RBX as an input operand. However, this seems unsupported by LLVM, because LLVM reserves RBX to be a base pointer when code gets more complicated. As a result, Gramine-SGX crashes when compiled in release mode.

## Steps to reproduce

This is a crash reported in #208.

Compile Gramine with `CC=clang`, `--buildtype=debugoptimized`, and `-Dasan=enabled`.

Run any test case in which Gramine forks, e.g. `fork_and_exec`. Gramine will crash with:

```
Unexpected memory fault occurred inside PAL (_DkStreamReportRespond at enclave_framework.c, libpal.so+0x7eaa3)
```

## Gramine commit hash

9e1d1f6ea860323f1f50f054f1cf3fe472a9c12d

## Analysis

Looking more closely, it seems that the ENCLU instruction is called with wrong parameters. This is our code:

```c
static inline int sgx_report(const sgx_target_info_t* targetinfo, const void* reportdata,
                             sgx_report_t* report) {
    __asm__ volatile(
        ENCLU "\n"
        :: "a"(EREPORT), "b"(targetinfo), "c"(reportdata), "d"(report)
        : "memory");
    return 0;
}
```

This is the generated assembly (for `enclave_framework.c`, `_DkStreamReportRespond`):
```
        xorl    %eax, %eax
        movq    %r14, %rbx
        movq    496(%rbx), %rcx
        leaq    2048(%rbx), %rdx
        .byte   15
        .byte   1
        .byte   215
```

The RBX register is used throughout the function as a base pointer. This is a complicated case (because of ASan), but I believe LLVM does this for some functions without ASan too.

Notice how the RBX register is overwritten with the parameter for ENCLU early on, but still used afterwards, and not restored after ENCLU.

I believe that **LLVM does not support using RBX directly as input**. Here is what I found (nothing in official docs, though):

[[CodeGen] emit inline asm clobber list warnings for reserved (cont)](https://reviews.llvm.org/D51165)

> Currently, in line with GCC, when specifying reserved registers like sp or pc on an inline asm() clobber list, we don't always preserve the original value across the statement. And in general, overwriting reserved registers can have surprising results.

This does not state explicitly that RBX is reserved, but points out a similar case.

[Inline assembly](https://doc.rust-lang.org/nightly/reference/inline-assembly.html) (in Rust docs)

> Some registers cannot be used for input or output operands:
> `bx` (x86-64) - This is used internally by LLVM as a "base pointer" for functions with complex stack frames.

## Proposed solutions

Don't use RBX as operand. Instead, save it and restore within the inline assembly block.

```c
    __asm__ volatile(
        "xchgq %%rdi, %%rbx\n"
        ENCLU "\n"
        "movq %%rdi, %%rbx\n"
        :: "a"(EREPORT), "D"(targetinfo), "c"(reportdata), "d"(report)
        : "memory");
```

pwmarcz:
## Description of the changes 

It seems that in LLVM, the RBX register is reserved, and using it as inline assembly operand can produce unexpected results. In our case, Clang emitted incorrect code around the ENCLU instruction when compiling with ASAN and optimizations enabled.

To fix the problem, I'm moving the ENCLU instructions to separate functions defined in assembly files. I'm also updating one instance of RBX use in LibOS.

Fixes #391 (which describes one of the issues in #208).

## How to test this PR? 

Apart from Jenkins, you can compile Gramine with Clang with `--buildtype=debugoptimized` or `release` and ASAN enabled, and verify that the `fork_and_exec` test case passes. See also #391.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/398)


pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/398)* status: 0 of 5 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

___
*[LibOS/shim/include/arch/x86_64/shim_internal-arch.h, line 14 at r1](https://reviewable.io/reviews/gramineproject/gramine/398#-MvcLX_x9PiujMKuY2c8:-MvcLX_yBsbJtmpulJo3:b-96nwdx) ([raw file](https://github.com/gramineproject/gramine/blob/d3c5e21d71b8aaa27e0081b8fa16fe4f66cba726/LibOS/shim/include/arch/x86_64/shim_internal-arch.h#L14)):*
> ```objc
>             "jmp *%%rax\r\n"                \
>             :                               \
>             : "a"(ENTRY), "b"(ARGP), "d"(0) \
> ```

FYI: No idea why this cleared RDX and flags. AFAIK, it's enough to set the stack pointer and jump. I changed the code to reflect what we do in PAL (`db_rtld.c`).
___
*[Pal/src/host/Linux-SGX/enclave_api.S, line 29 at r1](https://reviewable.io/reviews/gramineproject/gramine/398#-MvcKvRM1vp3HbxoiGPp:-MvcKvRM1vp3HbxoiGPq:bnrlgg6) ([raw file](https://github.com/gramineproject/gramine/blob/41667ea3173d09c6a31bfa2b1e6e9e177e735176/Pal/src/host/Linux-SGX/enclave_api.S#L29)):*
> ```
> 	ENCLU
> 
> 	# return 0
> ```

FYI: Intel's documentation is less than clear about the behavior of ENCLU functions. AFAICT, `EREPORT` leaves RAX unchanged (it's already 0, but that wouldn't be clear from the code), and `EGETKEY` sets RAX to either 0 or an error code.

dimakuv (changes requested):
> Reviewed 4 of 5 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/398)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

___
*[-- commits, line 11 at r2](https://reviewable.io/reviews/gramineproject/gramine/398#-MvccvZ04tRjFqfb0quL):*
Can we remove this `I'm moving/updating ...`? We don't use such phrasing in our commit messages. Maybe simply `The ENCLU instructions are now defined ...`

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/398)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "squash! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[-- commits, line 11 at r2](https://reviewable.io/reviews/gramineproject/gramine/398#-MvccvZ04tRjFqfb0quL):*

Can we remove this `I'm moving/updating ...`? We don't use such phrasing in our commit messages. Maybe simply `The ENCLU instructions are now defined ...`
</blockquote></details>

OK. I added a squash commit.

mkow (changes requested):
> Reviewed 4 of 5 files at r1, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/398)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @pwmarcz)

___
*[LibOS/shim/include/arch/x86_64/shim_internal-arch.h, line 14 at r1](https://reviewable.io/reviews/gramineproject/gramine/398#-MvcLX_x9PiujMKuY2c8:-Mvd7cnO0s3maa9ZldPe:b-bysqrp) ([raw file](https://github.com/gramineproject/gramine/blob/d3c5e21d71b8aaa27e0081b8fa16fe4f66cba726/LibOS/shim/include/arch/x86_64/shim_internal-arch.h#L14)):*

Oh... Jenkins failed on `bootstrap_static`, pointing out that RDX is indeed necessary :) I added a comment so that nobody tries to remove it again.
</blockquote></details>

RFLAGS has to be at least partially zeroed, see "3.4.1 Initial Stack and Register State" in Sys-V x64 ABI reference. I think it's a high time to pop "add a test asserting entrypoint conditions" from my TODOlist... ;)

boryspoplawski (changes requested):
> Reviewed 4 of 5 files at r1, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/398)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @pwmarcz)

___
*[-- commits, line 11 at r2](https://reviewable.io/reviews/gramineproject/gramine/398#-MvccvZ04tRjFqfb0quL):*

OK. I added a squash commit.
</blockquote></details>

Did you? I don't see it
___
*[LibOS/shim/include/arch/x86_64/shim_internal-arch.h, line 14 at r1](https://reviewable.io/reviews/gramineproject/gramine/398#-MvcLX_x9PiujMKuY2c8:-MvdCWLy0-sKFBQJQUUp:b-9nu7pl) ([raw file](https://github.com/gramineproject/gramine/blob/d3c5e21d71b8aaa27e0081b8fa16fe4f66cba726/LibOS/shim/include/arch/x86_64/shim_internal-arch.h#L14)):*

RFLAGS has to be at least partially zeroed, see "3.4.1 Initial Stack and Register State" in Sys-V x64 ABI reference. I think it's a high time to pop "add a test asserting entrypoint conditions" from my TODOlist... ;)
</blockquote></details>

Yes, please restore flag clearing
___
*[Pal/src/host/Linux-SGX/enclave_api.S, line 1 at r3](https://reviewable.io/reviews/gramineproject/gramine/398#-MvdD3lu5K4pNpZJF1Pb:-MvdD3lu5K4pNpZJF1Pc:bjtkxzf) ([raw file](https://github.com/gramineproject/gramine/blob/caa2454b24838543ea0a0025e27b1a72958f0aba/Pal/src/host/Linux-SGX/enclave_api.S#L1)):*
> ```
> ```

Maybe move this file to `common/src/arch/x86_64/`? I'm not sure what is the best place to keep it (not blocking, just an idea)...
___
*[Pal/src/host/Linux-SGX/enclave_api.S, line 7 at r3](https://reviewable.io/reviews/gramineproject/gramine/398#-MvdDnke4dNHX7PYEtqk:-MvdDnke4dNHX7PYEtql:b-g1ml3e) ([raw file](https://github.com/gramineproject/gramine/blob/caa2454b24838543ea0a0025e27b1a72958f0aba/Pal/src/host/Linux-SGX/enclave_api.S#L7)):*
> ```
> #include "sgx_arch.h"
> 
> 	.global sgx_report
> ```

Please use spaces not tabs (in the whole file)
___
*[Pal/src/host/Linux-SGX/enclave_api.S, line 11 at r3](https://reviewable.io/reviews/gramineproject/gramine/398#-MvdDxJ2EM9MaRi1d2vs:-MvdDxJ2EM9MaRi1d2vt:b-qztqoe) ([raw file](https://github.com/gramineproject/gramine/blob/caa2454b24838543ea0a0025e27b1a72958f0aba/Pal/src/host/Linux-SGX/enclave_api.S#L11)):*
> ```
> sgx_report:
> 	.cfi_startproc
> 	# Arguments:
> ```

Could you paste C function declaration? It's useful to have types of arguments.
Once we have that it might be not worth to keep the reg - arg list below (but not blocking on that part).
___
*[Pal/src/host/Linux-SGX/enclave_api.S, line 35 at r3](https://reviewable.io/reviews/gramineproject/gramine/398#-MvdEmrn6g_Oci1mVt3w:-MvdEmrn6g_Oci1mVt3x:b-ohafsx) ([raw file](https://github.com/gramineproject/gramine/blob/caa2454b24838543ea0a0025e27b1a72958f0aba/Pal/src/host/Linux-SGX/enclave_api.S#L35)):*
> ```
> 	.cfi_restore %rbx
> 	ret
> ```

Could you remove this empty line? I think it would look better and we do have none around `.cfi_startproc`.
___
*[Pal/src/host/Linux-SGX/enclave_api.S, line 42 at r3](https://reviewable.io/reviews/gramineproject/gramine/398#-MvdExzC0W_RdB41kYwt:-MvdExzC0W_RdB41kYwu:bk73g9l) ([raw file](https://github.com/gramineproject/gramine/blob/caa2454b24838543ea0a0025e27b1a72958f0aba/Pal/src/host/Linux-SGX/enclave_api.S#L42)):*
> ```
> sgx_getkey:
> 	.cfi_startproc
> 	# Arguments:
> ```

ditto (C declaration)
___
*[Pal/src/host/Linux-SGX/enclave_api.S, line 63 at r3](https://reviewable.io/reviews/gramineproject/gramine/398#-MvdF6IXE3nr0kMY67Be:-MvdF6IXE3nr0kMY67Bf:b-n2h2cr) ([raw file](https://github.com/gramineproject/gramine/blob/caa2454b24838543ea0a0025e27b1a72958f0aba/Pal/src/host/Linux-SGX/enclave_api.S#L63)):*
> ```
> 	.cfi_restore %rbx
> 	ret
> ```

ditto (empty line)

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/398)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "squash! " and "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[LibOS/shim/include/arch/x86_64/shim_internal-arch.h, line 14 at r1](https://reviewable.io/reviews/gramineproject/gramine/398#-MvcLX_x9PiujMKuY2c8:-Mvdk6nZBXhKzYG5T4X0:b-udqal8) ([raw file](https://github.com/gramineproject/gramine/blob/d3c5e21d71b8aaa27e0081b8fa16fe4f66cba726/LibOS/shim/include/arch/x86_64/shim_internal-arch.h#L14)):*

Thanks, I didn't know where it was exactly in the ABI PDF. I restored the zeroing and updated the comment.
</blockquote></details>

You didn't restore the removed `"cc"` from clobbers. Not sure if this is needed as we don't return (probably not), but we could add it back for sanity.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/398)* status: 4 of 5 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

___
*[LibOS/shim/include/arch/x86_64/shim_internal-arch.h, line 14 at r1](https://reviewable.io/reviews/gramineproject/gramine/398#-MvcLX_x9PiujMKuY2c8:-Mvdls4k3Cza17ga20RD:bpx32lq) ([raw file](https://github.com/gramineproject/gramine/blob/d3c5e21d71b8aaa27e0081b8fa16fe4f66cba726/LibOS/shim/include/arch/x86_64/shim_internal-arch.h#L14)):*

You didn't restore the removed `"cc"` from clobbers. Not sure if this is needed as we don't return (probably not), but we could add it back for sanity.
</blockquote></details>

Right, it should still be there. Fixed.

boryspoplawski (dismissed):
> Reviewed 1 of 2 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/398)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/398)* status: all files reviewed, 1 unresolved discussion, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/398)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/398)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (approved):
> Reviewed 1 of 2 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/398)* status: :shipit: complete! all files reviewed, all discussions resolved

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/398)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #402, opened 2022-02-14
[LibOS] Convert `pseudo` filesystem to use inodes

pwmarcz:
## Description of the changes 

In addition, `pseudo_node` objects are now tracked by number and checkpointed.

## How to test this PR? 

This should have a pretty good coverage on Jenkins.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/402)

### DIFF
--- a/LibOS/shim/include/shim_fs_pseudo.h
+++ b/LibOS/shim/include/shim_fs_pseudo.h
@@ -66,17 +66,26 @@ struct shim_dev_ops {
 #define PSEUDO_PERM_FILE_R  PERM_r__r__r__  /* default for all other files */
 #define PSEUDO_PERM_FILE_RW PERM_rw_rw_rw_
 
+/* Maximum count of created nodes (`pseudo_add_*` calls) */
+#define PSEUDO_MAX_NODES 128
+
 /*
  * A node of the pseudo filesystem. A single node can describe either a single file, or a family of
  * files (see `name_exists` and `list_names` below).
  *
  * The constructors for `pseudo_node` (`pseudo_add_*`) take arguments for most commonly used fields.
  * The node can then be further customized by directly modifying other fields.
+ *
+ * NOTE: We assume that all Gramine processes within a single instance create exactly the same set
+ * of nodes, in the same order, during initialization. This is because we use the node number (`id`)
+ * for checkpointing.
  */
 DEFINE_LIST(pseudo_node);
 DEFINE_LISTP(pseudo_node);
 struct pseudo_node {
     enum pseudo_type type;
+
+    unsigned int id;
 
     LIST_TYPE(pseudo_node) siblings;
 
--- a/LibOS/shim/src/fs/shim_fs_pseudo.c
+++ b/LibOS/shim/src/fs/shim_fs_pseudo.c
@@ -5,6 +5,9 @@
 
 /*
  * This file contains implementation of the "pseudo" filesystem.
+ *
+ * We store a pointer to a `pseudo_node` object in `inode->data`. Note that since `inode->data`
+ * never changes, we do NOT take `inode->lock` when reading it.
  */
 
 #include "shim_fs_pseudo.h"
@@ -12,6 +15,12 @@
 #include "stat.h"
 
 LISTP_TYPE(pseudo_node) g_pseudo_roots = LISTP_INIT;
+
+/* Array of nodes by ID. Used for restoring a node from checkpoint (we send only node ID). We assume
+ * that all Gramine processes within a single instance create exactly the same set of nodes, in the
+ * same order, during initialization. */
+static struct pseudo_node* g_pseudo_nodes[PSEUDO_MAX_NODES];
+static unsigned int g_pseudo_node_count;
 
 /* Find a root node with given name. */
 static struct pseudo_node* pseudo_find_root(const char* name) {
@@ -26,64 +35,55 @@ static struct pseudo_node* pseudo_find_root(const char* name) {
     return NULL;
 }
 
-/*
- * Find a `pseudo_node` for given dentry. The pointer to retrieved node is cached in the `data`
- * field of the dentry.
- *
- * Note that we call `pseudo_find`, instead of initializing `dent->data` once on dentry lookup,
- * because the dentry might be restored from a checkpoint. Checkpointing clears the `data` field,
- * which is actually what we want, because we are not able to send `pseudo_node` pointers across
- * processes.
- *
- * Instead, we retrieve the node on first access to given dentry, and store it in `dent->data`. At
- * the same time, we also retrieve and cache nodes all the way to the root (see the recursive call
- * below).
- *
- * Note that `pseudo_find` might fail for such a checkpointed dentry: for instance, we might have a
- * dentry for `/proc/<pid>` where the process does not exist anymore. Ideally, we would invalidate
- * such dentries; for now, all operations on them will return -ENOENT.
- */
+/* Find a `pseudo_node` for given dentry. */
 static struct pseudo_node* pseudo_find(struct shim_dentry* dent) {
-    struct pseudo_node* node;
-
-    lock(&dent->lock);
-    node = dent->data;
-    unlock(&dent->lock);
-
-    if (node)
-        return node;
+    assert(locked(&g_dcache_lock));
 
     if (!dent->parent) {
         /* This is the filesystem root */
-        node = pseudo_find_root(dent->mount->uri);
-        goto out;
+        return pseudo_find_root(dent->mount->uri);
     }
 
-    /* Recursive call: find the node for parent */
-    struct pseudo_node* parent_node = pseudo_find(dent->parent);
-    if (!parent_node) {
-        node = NULL;
-        goto out;
-    }
+    assert(dent->parent->inode);
+    struct pseudo_node* parent_node = dent->parent->inode->data;
 
     /* Look for a child node with matching name */
     assert(parent_node->type == PSEUDO_DIR);
+    struct pseudo_node* node;
     LISTP_FOR_EACH_ENTRY(node, &parent_node->dir.children, siblings) {
         if (node->name && strcmp(dent->name, node->name) == 0) {
-            goto out;
+            return node;
         }
         if (node->name_exists && node->name_exists(dent->parent, dent->name)) {
-            goto out;
+            return node;
         }
     }
-    node = NULL;
-out:
-    if (node) {
-        lock(&dent->lock);
-        dent->data = node;
-        unlock(&dent->lock);
+    return NULL;
+}
+
+static int pseudo_icheckpoint(struct shim_inode* inode, void** out_data, size_t* out_size) {
+    unsigned int* id = malloc(sizeof(*id));
+    if (!id)
+        return -ENOMEM;
+
+    struct pseudo_node* node = inode->data;
+    *id = node->id;
+    *out_data = (void*)id;
+    *out_size = sizeof(*id);
+    return 0;
+}
+
+static int pseudo_irestore(struct shim_inode* inode, void* data) {
+    unsigned int* id = data;
+    if (*id >= g_pseudo_node_count) {
+        log_error("Invalid pseudo_node id: %u", *id);
+        return -EINVAL;
     }
-    return node;
+
+    struct pseudo_node* node = g_pseudo_nodes[*id];
+    assert(node);
+    inode->data = node;
+    return 0;
 }
 
 static int pseudo_mount(const char* uri, void** mount_data) {
@@ -93,10 +93,12 @@ static int pseudo_mount(const char* uri, void** mount_data) {
 }
 
 static int pseudo_open(struct shim_handle* hdl, struct shim_dentry* dent, int flags) {
+    assert(locked(&g_dcache_lock));
+    assert(dent->inode);
+
+    struct pseudo_node* node = dent->inode->data;
+
     int ret;
-    struct pseudo_node* node = pseudo_find(dent);
-    if (!node)
-        return -ENOENT;
 
     switch (node->type) {
         case PSEUDO_DIR:
@@ -137,30 +139,47 @@ static int pseudo_open(struct shim_handle* hdl, struct shim_dentry* dent, int fl
             break;
         }
     }
+    hdl->inode = dent->inode;
+    get_inode(dent->inode);
 
     return 0;
 }
 
 static int pseudo_lookup(struct shim_dentry* dent) {
+    assert(locked(&g_dcache_lock));
+    assert(!dent->inode);
+
     struct pseudo_node* node = pseudo_find(dent);
     if (!node)
         return -ENOENT;
 
+    mode_t type;
     switch (node->type) {
         case PSEUDO_DIR:
-            dent->type = S_IFDIR;
+            type = S_IFDIR;
             break;
         case PSEUDO_LINK:
-            dent->type = S_IFLNK;
+            type = S_IFLNK;
             break;
         case PSEUDO_STR:
-            dent->type = S_IFREG;
+            type = S_IFREG;
             break;
         case PSEUDO_DEV:
-            dent->type = S_IFCHR;
+            type = S_IFCHR;
             break;
+        default:
+            BUG();
     }
+
+    struct shim_inode* inode = get_new_inode(dent->mount, type, node->perm);
+    if (!inode)
+        return -ENOMEM;
+
+    inode->data = node;
+
+    dent->type = type;
     dent->perm = node->perm;
+    dent->inode = inode;
     return 0;
 }
 
@@ -171,8 +190,6 @@ static int count_nlink(const char* name, void* arg) {
     return 0;
 }
 
-static int pseudo_readdir(struct shim_dentry* dent, readdir_callback_t callback, void* arg);
-
 static dev_t makedev(unsigned int major, unsigned int minor) {
     dev_t dev;
     dev  = (((dev_t)(major & 0x00000fffu)) <<  8);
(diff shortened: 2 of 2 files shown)


mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/402)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[LibOS/shim/include/shim_fs_pseudo.h, line 79 at r1](https://reviewable.io/reviews/gramineproject/gramine/402#-Mvt5fC104OvWZLiwap0:-Mvup2bk8qQhraUfGMF-:b25iljm) ([raw file](https://github.com/gramineproject/gramine/blob/8564c7312ef31813f90a65519b3bd18045ea62b5/LibOS/shim/include/shim_fs_pseudo.h#L79)):*

FYI: The exact pseudo nodes created depends on filesystems that we initialize. Currently, `/sys` is initialized depending on manifest, and `/dev/attestation` depending on whether we're running under SGX.

A safer way would be to use paths for checkpointing, e.g. `/dev/null` or `/proc/!pid/cmdline`; note that some pseudo-nodes like `!pid` would need an artificial name). But it's more complicated, and IMO not worth it.
</blockquote></details>

Hmm, how does this work with pseudo nodes created dynamically, like e.g. `/proc/<pid>/fd/*`? I think we don't create them right now, but this implementation seems to disallow such a thing to exist?
___
*[LibOS/shim/src/fs/shim_fs_pseudo.c, line 213 at r2](https://reviewable.io/reviews/gramineproject/gramine/402#-Mvv-hk12Xj8wVBvQlzI:-Mvv-hk12Xj8wVBvQlzJ:bnjntux) ([raw file](https://github.com/gramineproject/gramine/blob/b5b6a1328a2ec460d0f3c9f4edd8007f6f1143f7/LibOS/shim/src/fs/shim_fs_pseudo.c#L213)):*
> ```
> Note that we might not be holding `g_dcache_lock` here, so we cannot call
> `pseudo_readdir`.
> ```

Is it still safe to traverse the children list without any locks? I guess you're assuming here that the pseudo directories can't change in runtime?
___
*[LibOS/shim/src/fs/shim_fs_pseudo.c, line 503 at r2](https://reviewable.io/reviews/gramineproject/gramine/402#-Mvv2s12BIJt52Pd9RZ-:-Mvv2s13-apIMyfg6CJl:b-oh1hlt) ([raw file](https://github.com/gramineproject/gramine/blob/b5b6a1328a2ec460d0f3c9f4edd8007f6f1143f7/LibOS/shim/src/fs/shim_fs_pseudo.c#L503)):*
> ```
> g_pseudo_node_count++
> ```

Shouldn't this be atomic? Or you're also assuming here that we create pseudo nodes only during initialization?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/402)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[LibOS/shim/include/shim_fs_pseudo.h, line 79 at r1](https://reviewable.io/reviews/gramineproject/gramine/402#-Mvt5fC104OvWZLiwap0:-MvwV7C6775KjfS6BskV:bty0z63) ([raw file](https://github.com/gramineproject/gramine/blob/8564c7312ef31813f90a65519b3bd18045ea62b5/LibOS/shim/include/shim_fs_pseudo.h#L79)):*

Hmm, how does this work with pseudo nodes created dynamically, like e.g. `/proc/<pid>/fd/*`? I think we don't create them right now, but this implementation seems to disallow such a thing to exist?
</blockquote></details>

That's right. The list of pseudo-nodes is static. This checkpointing scheme also assumes that: otherwise you couldn't easily send a handle to a child process and expect it to find the same node.

However, these nodes can generate a dynamic list of _files_. We do have `/proc/<pid>/fd/<n>`, and it is a single node. The `name_exists` / `list_names` callbacks define the file names (e.g. `0`, `1`, `2`...).
___
*[LibOS/shim/src/fs/shim_fs_pseudo.c, line 213 at r2](https://reviewable.io/reviews/gramineproject/gramine/402#-Mvv-hk12Xj8wVBvQlzI:-MvwVcY36iLnMYcZwwn4:breipu3) ([raw file](https://github.com/gramineproject/gramine/blob/b5b6a1328a2ec460d0f3c9f4edd8007f6f1143f7/LibOS/shim/src/fs/shim_fs_pseudo.c#L213)):*

> ```
> Note that we might not be holding `g_dcache_lock` here, so we cannot call
> `pseudo_readdir`.
> ```

Is it still safe to traverse the children list without any locks? I guess you're assuming here that the pseudo directories can't change in runtime?
</blockquote></details>

Yes, it cannot change in runtime.
___
*[LibOS/shim/src/fs/shim_fs_pseudo.c, line 503 at r2](https://reviewable.io/reviews/gramineproject/gramine/402#-Mvv2s12BIJt52Pd9RZ-:-MvwVio25QyCSwVtkJVi:beo8k6k) ([raw file](https://github.com/gramineproject/gramine/blob/b5b6a1328a2ec460d0f3c9f4edd8007f6f1143f7/LibOS/shim/src/fs/shim_fs_pseudo.c#L503)):*

> ```
> g_pseudo_node_count++
> ```

Shouldn't this be atomic? Or you're also assuming here that we create pseudo nodes only during initialization?
</blockquote></details>

Yes. This is an assumption this module always made. Notice that there are no locks taken for adding the new node to linked lists below.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/402)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/402#-MvwVg7n8KxImlBOvYGf:-MvwVg7n8KxImlBOvYGg:b-pi6iqs) (no related file):*
I'm confused as to what happens in e.g. this case:
1. Parent process creates a child process and opens `/proc/<child-pid>/exe`.
2. Child process terminates.
3. Parent process creates another child process.
4. New child process gets a checkpoint with the inode for `/proc/<child-pid>/exe` and the opened shim handle.
5. Child process reads from the opened shim handle:
    - In the old code, `pseudo_read()` would return `-ENOENT` because it would try to `pseudo_find` and fail to find that `<child-pid>`
    - In your new code, `pseudo_read()` would return the string that was memorized by the parent process in `hdl->info.str.mem`.

Is my understanding correct? Also, is this how the Linux kernel does it?

In other words, this PR fixes the problem of "dangling shim handles" -- the handles' inodes (and corresponding `pseudo_node`s) are now correctly checkpointed and "survive" in the child process.

---
___
*[LibOS/shim/include/shim_fs_pseudo.h, line 79 at r1](https://reviewable.io/reviews/gramineproject/gramine/402#-Mvt5fC104OvWZLiwap0:-MvwXXPd6A7eLG6kPQ5B:bu64h1k) ([raw file](https://github.com/gramineproject/gramine/blob/8564c7312ef31813f90a65519b3bd18045ea62b5/LibOS/shim/include/shim_fs_pseudo.h#L79)):*

That's right. The list of pseudo-nodes is static. This checkpointing scheme also assumes that: otherwise you couldn't easily send a handle to a child process and expect it to find the same node.

However, these nodes can generate a dynamic list of _files_. We do have `/proc/<pid>/fd/<n>`, and it is a single node. The `name_exists` / `list_names` callbacks define the file names (e.g. `0`, `1`, `2`...).
</blockquote></details>

The assumption sounds reasonable. I cannot come up with a hypothetical case when we would diverge a list of pseudo nodes we create in the child process...
___
*[LibOS/shim/src/fs/shim_fs_pseudo.c, line 22 at r2](https://reviewable.io/reviews/gramineproject/gramine/402#-MvwXmM2EqNoRGS7046M:-MvwXmM2EqNoRGS7046N:b-pe7x6q) ([raw file](https://github.com/gramineproject/gramine/blob/b5b6a1328a2ec460d0f3c9f4edd8007f6f1143f7/LibOS/shim/src/fs/shim_fs_pseudo.c#L22)):*
> ```C
> 
> static struct pseudo_node* g_pseudo_nodes[PSEUDO_MAX_NODES];
> static unsigned int g_pseudo_node_count;
> ```

Maybe add a comment here that `g_pseudo_nodes` is used only in two cases:
1. Filled during each-process initialization (mounting of pseudo-FSes), assumed to be deterministic in all processes of the same Gramine instance
2. Queried during checkpoint-restore in the child process, to bind inodes to `pseudo_node`s created in step (1)

Maybe also add a node that in the parent process, this is basically not used at all.
___
*[LibOS/shim/src/fs/shim_fs_pseudo.c, line 46 at r2](https://reviewable.io/reviews/gramineproject/gramine/402#-MvwYZmxD6cmuPf0BKWz:-MvwYZmxD6cmuPf0BKX-:b5eesxt) ([raw file](https://github.com/gramineproject/gramine/blob/329a12f15baf3efa1ce25cf2419a3e5af3c9b156/LibOS/shim/src/fs/shim_fs_pseudo.c#L46)):*
> ```C
>  * Note that `pseudo_find` might fail for such a checkpointed dentry: for instance, we might have a
>  * dentry for `/proc/<pid>` where the process does not exist anymore. Ideally, we would invalidate
>  * such dentries; for now, all operations on them will return -ENOENT.
> ```

"Ideally, we would invalidate such dentries" -- we don't do it even in the new PR, right?
___
*[LibOS/shim/src/fs/shim_fs_pseudo.c, line 64 at r2](https://reviewable.io/reviews/gramineproject/gramine/402#-MvwZ8XG1FFAU-Jy5b6k:-MvwZ8XG1FFAU-Jy5b6l:bhip55r) ([raw file](https://github.com/gramineproject/gramine/blob/b5b6a1328a2ec460d0f3c9f4edd8007f6f1143f7/LibOS/shim/src/fs/shim_fs_pseudo.c#L64)):*
> ```C
> 
> static int pseudo_icheckpoint(struct shim_inode* inode, void** out_data, size_t* out_size) {
>     unsigned int* id = malloc(sizeof(*id));
> ```

Do we just memory leak this object? How is it supposed to be freed / garbage collected?

Ah, I guess it will be freed in the generic checkpoint code (or I guess in the inode-checkpoint code), after it sends all `out_data` to the child process. Is it so?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/402)* status: 0 of 2 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/402#-MvwVg7n8KxImlBOvYGf:-MvwfV4m4UA3JayCc619:b-zi5zjz) (no related file):*
I think that's correct.

>Also, is this how the Linux kernel does it?

I haven't looked too closely. I think there is a similar layer of abstraction to our "str" files, so I would expect the value to be memoized.

---
___
*[LibOS/shim/include/shim_fs_pseudo.h, line 79 at r1](https://reviewable.io/reviews/gramineproject/gramine/402#-Mvt5fC104OvWZLiwap0:-MvwhQjs-JQGvCH49GhQ:b-x2itm1) ([raw file](https://github.com/gramineproject/gramine/blob/8564c7312ef31813f90a65519b3bd18045ea62b5/LibOS/shim/include/shim_fs_pseudo.h#L79)):*

The assumption sounds reasonable. I cannot come up with a hypothetical case when we would diverge a list of pseudo nodes we create in the child process...
</blockquote></details>

Mostly I'd be afraid of doing this by mistake - so far, the initialization has some conditions, but they at most depend on the manifest.
___
*[LibOS/shim/src/fs/shim_fs_pseudo.c, line 22 at r2](https://reviewable.io/reviews/gramineproject/gramine/402#-MvwXmM2EqNoRGS7046M:-Mvwdak5DR3XIaRAI2qz:bwvzx5x) ([raw file](https://github.com/gramineproject/gramine/blob/b5b6a1328a2ec460d0f3c9f4edd8007f6f1143f7/LibOS/shim/src/fs/shim_fs_pseudo.c#L22)):*

Maybe add a comment here that `g_pseudo_nodes` is used only in two cases:
1. Filled during each-process initialization (mounting of pseudo-FSes), assumed to be deterministic in all processes of the same Gramine instance
2. Queried during checkpoint-restore in the child process, to bind inodes to `pseudo_node`s created in step (1)

Maybe also add a node that in the parent process, this is basically not used at all.
</blockquote></details>

I added a comment, explaining what this array contains and how we use it. The assumption is explained in the header file as well, but maybe it doesn't hurt to repeat it here.
___
*[LibOS/shim/src/fs/shim_fs_pseudo.c, line 46 at r2](https://reviewable.io/reviews/gramineproject/gramine/402#-MvwYZmxD6cmuPf0BKWz:-MvweJo_5qBvFhrtCNpr:bxbdo0j) ([raw file](https://github.com/gramineproject/gramine/blob/329a12f15baf3efa1ce25cf2419a3e5af3c9b156/LibOS/shim/src/fs/shim_fs_pseudo.c#L46)):*

"Ideally, we would invalidate such dentries" -- we don't do it even in the new PR, right?
</blockquote></details>

Right. However, the rest of the paragraph does not apply to `pseudo_find` (we don't even call it for dentries restored from a checkpoint). And the fact that we don't invalidate no-longer-existing files is documented in the header file already.

By the way, this is similar to the problem we discussed recently: files disappearing on the host, or modified by another process. The solution is also to mark these files (dentries) as invalid once we notice that. But this should be less problematic after inodes are introduced everywhere, and dentries are just a name-to-inode association.
___
*[LibOS/shim/src/fs/shim_fs_pseudo.c, line 64 at r2](https://reviewable.io/reviews/gramineproject/gramine/402#-MvwZ8XG1FFAU-Jy5b6k:-Mvwed0Z9Rxh10lVzFQe:b-ms5594) ([raw file](https://github.com/gramineproject/gramine/blob/b5b6a1328a2ec460d0f3c9f4edd8007f6f1143f7/LibOS/shim/src/fs/shim_fs_pseudo.c#L64)):*

Do we just memory leak this object? How is it supposed to be freed / garbage collected?

Ah, I guess it will be freed in the generic checkpoint code (or I guess in the inode-checkpoint code), after it sends all `out_data` to the child process. Is it so?
</blockquote></details>

Yes, it will be freed by the caller, and this is explained in the documentation for `icheckpoint`.

This interface is a bit awkward - as explained in the previous PR (#297), it is done this way so that it's self-contained. If/when we rewrite the checkpoint system, it will probably make more sense for `icheckpoint/irestore` to send/receive data directly "over the wire".

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/402)* status: 0 of 2 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/402#-MvwVg7n8KxImlBOvYGf:-MvwjKxB23F60Ekucprc:b-47bso0) (no related file):*

I think that's correct.

>Also, is this how the Linux kernel does it?

I haven't looked too closely. I think there is a similar layer of abstraction to our "str" files, so I would expect the value to be memoized.

</blockquote></details>

And, this is speculation, but I would expect Linux to do what I'm planning to do later (see the comment about invalidating files): keep a handle associated with an inode object, even though that inode object is no longer reachable by other means.

---

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/402)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[LibOS/shim/src/fs/shim_fs_pseudo.c, line 46 at r2](https://reviewable.io/reviews/gramineproject/gramine/402#-MvwYZmxD6cmuPf0BKWz:-MvwqD3wBFVRq93ju8mQ:b-zh0nsv) ([raw file](https://github.com/gramineproject/gramine/blob/329a12f15baf3efa1ce25cf2419a3e5af3c9b156/LibOS/shim/src/fs/shim_fs_pseudo.c#L46)):*

Right. However, the rest of the paragraph does not apply to `pseudo_find` (we don't even call it for dentries restored from a checkpoint). And the fact that we don't invalidate no-longer-existing files is documented in the header file already.

By the way, this is similar to the problem we discussed recently: files disappearing on the host, or modified by another process. The solution is also to mark these files (dentries) as invalid once we notice that. But this should be less problematic after inodes are introduced everywhere, and dentries are just a name-to-inode association.
</blockquote></details>

Agreed, thanks for the explanation.

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/402)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

___
*[LibOS/shim/include/shim_fs_pseudo.h, line 79 at r1](https://reviewable.io/reviews/gramineproject/gramine/402#-Mvt5fC104OvWZLiwap0:-MvxtsmXEhMQZufvw-lq:b-pq2b8w) ([raw file](https://github.com/gramineproject/gramine/blob/8564c7312ef31813f90a65519b3bd18045ea62b5/LibOS/shim/include/shim_fs_pseudo.h#L79)):*

Mostly I'd be afraid of doing this by mistake - so far, the initialization has some conditions, but they at most depend on the manifest.
</blockquote></details>

Fair. I feel a bit uneasy about the fact that we can't assert that the trees are exactly the same in every process, but I don't have an idea for a simple solution to this.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/402)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/402)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #410, opened 2022-02-16
[LibOS] Make sure ASan tests work in optimized builds

pwmarcz:
## Description of the changes 

Previously, `clang -O3` optimized the `asan_stack` and `asan_global` tests to a no-op.

This is reported in #208 (hopefully the last failure from that issue).

## How to test this PR? 

Compile Gramine with `--buildtype=debugoptimized` or `--buildtype=release` and ASan enabled.

The following tests do not trigger ASan on master, and should do so now:

```
regression $ gramine-test run run_test asan_stack
regression $ gramine-test run run_test asan_global
```

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/410)

### DIFF
--- a/LibOS/shim/src/shim_call.c
+++ b/LibOS/shim/src/shim_call.c
@@ -28,33 +28,34 @@ static int run_test_ubsan_int_overflow(void) {
 #endif
 
 #ifdef ASAN
-/* Test: allocate a buffer on heap, write past the end of buffer (ASan only) */
+
+/*
+ * ASan tests: write past a heap/stack/global buffer. We use a volatile pointer in order to bypass
+ * compiler optimizations and warnings.
+ */
+
 __attribute__((no_sanitize("undefined")))
 static int run_test_asan_heap(void) {
-    uint8_t* buf = malloc(30);
-    buf[30] = 1;
+    char* buf = malloc(30);
+    char* volatile c = buf + 30;
+    *c = 1;
     free(buf);
     return 0;
 }
 
-/* Test: write past the end of a stack buffer (ASan only) */
 __attribute__((no_sanitize("undefined")))
 static int run_test_asan_stack(void) {
     char buf[30];
-    /* Take a pointer: direct assignment such as `buf[30] = 1;` triggers a compiler warning */
-    char* c = buf + 30;
+    char* volatile c = buf + 30;
     *c = 1;
-
     return 0;
 }
 
-/* Test: write past the end of a global (static local) buffer (ASan only) */
 __attribute__((no_sanitize("undefined")))
 static int run_test_asan_global(void) {
     static char buf[30];
-    char* c = buf + 30;
+    char* volatile c = buf + 30;
     *c = 1;
-
     return 0;
 }
 


dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/410)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/410)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required) (waiting on @pwmarcz)

___
*[LibOS/shim/src/shim_call.c, line 34 at r1](https://reviewable.io/reviews/gramineproject/gramine/410#-Mw1msupAKQRHvaSOHsN:-Mw1msupAKQRHvaSOHsO:b-uhpm0l) ([raw file](https://github.com/gramineproject/gramine/blob/a400d2f6e082701c9adbb83e6fbcb6026cc0a13a/LibOS/shim/src/shim_call.c#L34)):*
> ```C
> /*
>  * ASan tests: write past a heap/stack/global buffer. We use a volatile pointer in order to bypass
>  * compiler optimizations and warnings.
> ```

We already have a macro for exactly this: `WRITE_ONCE()`.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/410)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required) (waiting on @pwmarcz)

___
*[LibOS/shim/src/shim_call.c, line 34 at r1](https://reviewable.io/reviews/gramineproject/gramine/410#-Mw1msupAKQRHvaSOHsN:-Mw26JI1EOCa2h0aTWK8:bvxje9t) ([raw file](https://github.com/gramineproject/gramine/blob/a400d2f6e082701c9adbb83e6fbcb6026cc0a13a/LibOS/shim/src/shim_call.c#L34)):*

Nope! `WRITE_ONCE` uses `volatile foo*`, not `foo* volatile` (i.e. pointer to a volatile char, not a volatile pointer to char), and this is apparently not enough for the compiler.

This gets compiled to a no-op:

```c
static int run_test_asan_stack(void) {
    char buf[30];
    WRITE_ONCE(buf[30], 1);
    return 0;
}
```

Same if I use `volatile char* c`.
</blockquote></details>

Huh, sounds like a clang bug. OTOH `volatile` is not clearly defined by compilers / standard, no one knows what's the correct semantics of it :/
Maybe it's time to sprinkle more of the magical `volatile` dust on these macros? If they weren't enough here, then they may also be broken in other places...

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/410)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required) (waiting on @mkow)

___
*[LibOS/shim/src/shim_call.c, line 34 at r1](https://reviewable.io/reviews/gramineproject/gramine/410#-Mw1msupAKQRHvaSOHsN:-Mw29zRCDyW8ju--JVyv:b244lu) ([raw file](https://github.com/gramineproject/gramine/blob/a400d2f6e082701c9adbb83e6fbcb6026cc0a13a/LibOS/shim/src/shim_call.c#L34)):*
Experimentally, it looks like that's because we're accessing an illegal location. If I write to `buf[29]` through a `volatile char*`, the write gets emitted (along with ASan instrumentation), but if I write to `buf[30]` through a `volatile char*`, the write gets optimized out (and even `return *c - 1` gets compiled as `return -1`).

That explains why writing through a volatile pointer works: the compiler is prevented from knowing the _location_ of the write. 

> Maybe it's time to sprinkle more of the magical `volatile` dust on these macros? If they weren't enough here, then they may also be broken in other places...

I wouldn't, because (1) Linux seems to have the same definition of `WRITE_ONCE()`, (2) this case is special, as we're doing something clearly forbidden (and our other uses of `WRITE_ONCE()` are normal writes to struct fields).

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/410)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/410)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #371, opened 2022-02-02
RFC: protected files redesign


pwmarcz:
## Summary

Make protected files configurable per mount, and move handling to LibOS.

## Problems

This proposal addresses the following problems:

* Bug: Protected files do not support renaming (#31). This is difficult to fix for the following reasons:
  * Protected files are tracked in PAL by path
  * We store the path in file header (so renaming needs to modify the file)
  * Broken reference counting (see below)
  * It's easy to move a file out of list of protected-files paths, effectively "un-protecting" it. The other way is also possible: you could move a non-protected file to protected path.
  
* Bug: Protected files in PAL use a broken reference counting scheme
  * I'm not sure what the impact is. Probably memory/FD leaks, as we do not close the files when we should.

* Bug: If host returns the same FD for multiple files, Gramine segfaults (https://github.com/gramineproject/graphene/issues/2360)
  * This is difficult to fix because of the way they're tracked by PAL.
  
* Bug: Path normalization for protected files is buggy (`foo` and `./foo` are treated differently)

- Feature request: setting different keys for different protected files (#358)

The following problems are NOT addressed by this proposal:

* Bug: Protected files are not flushed at `exit`, resulting in data corruption (https://github.com/gramineproject/graphene/issues/2663)

* Bug: Unmapping protected files does not work properly (#231)

## Current design

As an overview, here is the current implementation of protected files.

1. Manifest: The user specifies `sgx.protected_files` (as well as `protected_{mrenclave,mrsigner}_files`). These are lists of paths (files AND directories) in the host.

2. LibOS: does not know which files are protected, but (based on specified mounts) translates some I/O operations to PAL operations on host paths.

3. SGX PAL (`db_files.c`): during file operations, checks if the path corresponds to a protected file. If so, forwards the operations
   * :warning: Note that `struct protected_file` is NOT associated with the PAL handle directly, but looked up in a hash table (indexed by path) every time.
   * We need to ensure a single protected file handle is open for a path, even if there are multiple PAL handles for the file
   * :warning: There's a (broken) reference counting scheme that ensures we unload the protected file handle.
  
4. SGX PAL adapter to protected files (`enclave_pf.[ch]`). This is a module responsible for tracking open protected files.
   * `struct protected_file`: describes a protected file path (:warning: either a single file, or a directory!)
   * Stored in a hash map (by path)
   * If we encounter a new file that should be protected (i.e. path is under a protected directory), we create a new `struct protected file`
[...]

Issue #23, opened 2021-03-08
Mountpoint order is not preserved in current manifest syntax


mkow:
## Description of the problem

Currently we don't use proper TOML syntax for declaring mountpoints, instead, we use a syntax which resembles the pre-TOML one used in Graphene. As a result, the entries are not ordered, but Graphene actually relies on the specific mounting order (e.g. you can't mount /lib/asdf first and then /lib, but the other way around works). The problem is, that TOML structure is just a dictionary, so the order of keys is not preserved.

This will be solved by finally migrating to a proper TOML syntax for everything, which will be eventually done (after this, mount points list should be a TOML array). Cleaning up mounting code should also help here - I think that in Graphene mount order shouldn't matter, we only need to ensure that there are no duplicated mounts.

## Steps to reproduce

Swap the order of entries in e.g. manifest.template in LibOS regression, so that `"$(ARCH_LIBDIR)"` is listed before `"/lib"`.

## Expected results

Everything still works.

## Actual results

```
[P6098:T1:bootstrap_cpp] Mount /lib already exists, verify that there are no duplicate mounts in manifest
[P6098:T1:bootstrap_cpp] (note that /proc and /dev are automatically mounted in Graphene).
[P6098:T1:bootstrap_cpp] Mounting file:../../../../Runtime on /lib (type=chroot) failed (17)
[P6098:T1:bootstrap_cpp] Error during shim_init() in init_mount (-17)
```

## Additional information

gramineproject/graphene#2210 contains an ugly workaround, but we need a proper solution.

dimakuv:
This should be fixed now, as we moved to a proper TOML syntax with `fs.mounts` being a TOML array (which preserves the order).

PR #411, opened 2022-02-16
[LibOS] Convert `fs.mount` to an array

pwmarcz:
## Description of the changes 

This converts the `fs.mount` manifest key to an array instead of table. Similar to `sgx.{trusted,allowed,protected}_files`, an array can be parsed more efficiently. It's also more natural, because order matters when specifying mount points.

The old (table) syntax is still accepted, but triggers a deprecation warning.

This commit converts the main manifest for regression tests. Other manifests will be converted later.

In addition, Gramine now allows omitting the `type` key (with "chroot" as default) and the `uri` key (so that it's easier to use the `tmpfs` filesystem).

Issue: #23 (although we might want to close it after the deprecated syntax is gone).

Prerequisite for #371: new syntax for trusted/protected files.

## How to test this PR? 

The `manifest.template` in LibOS regression tests should exercise the new code. You can modify it to check how things fail.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/411)

### DIFF
--- a/Documentation/manifest-syntax.rst
+++ b/Documentation/manifest-syntax.rst
@@ -1,6 +1,11 @@
 Manifest syntax
 ===============
 
+..
+   TODO: We would like to change the below to `.. highlight:: toml`. However,
+   Pygments (as of 2.11.2) fails to parse constructions such as unquoted
+   `[true|false]` (because of the `|` character inside).
+
 .. highlight:: text
 
 A |~| manifest file is an application-specific configuration text file that
@@ -94,11 +99,11 @@ at that path. For example::
 
    libos.entrypoint = "/usr/bin/python3.8"
 
-   fs.mount.python.type = "chroot"
-   fs.mount.python.path = "/usr/bin/python3.8"
-   fs.mount.python.uri = "file:/usr/bin/python3.8"
-   # Or, if using a binary from your local directory:
-   # fs.mount.python.uri = "file:python3.8"
+   fs.mount = [
+     { path = "/usr/bin/python3.8", uri = "file:/usr/bin/python3.8" },
+     # Or, if using a binary from your local directory:
+     # { path = "/usr/bin/python3.8", uri = "file:python3.8" },
+   ]
 
 .. note ::
    Earlier, ``libos.entrypoint`` was a PAL URI. If you used it with a relative
@@ -326,42 +331,67 @@ Root FS mount point
     fs.root.uri  = "[URI]"
 
 This syntax specifies the root file system to be mounted inside the library OS.
-If not specified, then Gramine mounts the current working directory as the
-root. There can be only one root FS mount point specified in the manifest.
+Both parameters are optional. If not specified, then Gramine mounts the current
+working directory as the root.
 
 FS mount points
 ^^^^^^^^^^^^^^^
 
 ::
 
-    fs.mount.[identifier].type = "[chroot|tmpfs]"
-    fs.mount.[identifier].path = "[PATH]"
-    fs.mount.[identifier].uri  = "[URI]"
+    fs.mount = [
+      { type = "[chroot|...]", path = "[PATH]", uri = "[URI]" },
+      { type = "[chroot|...]", path = "[PATH]", uri = "[URI]" },
+    ]
+
+Or, as separate sections:
+
+::
+
+    [[fs.mount]]
+    type = "[chroot|...]"
+    path = "[PATH]"
+    uri  = "[URI]"
+
+    [[fs.mount]]
+    type = "[chroot|...]"
+    path = "[PATH]"
+    uri  = "[URI]"
 
 This syntax specifies how file systems are mounted inside the library OS. For
 dynamically linked binaries, usually at least one `chroot` mount point is
-required in the manifest (the mount point of linked libraries).
+required in the manifest (the mount point of linked libraries). The filesystems
+will be mounted in the order in which they appear in the manifest.
+
+.. note::
+   Keep in mind that TOML does not allow trailing commas in inline tables:
+   ``{ path = "...", uri = "...", }`` is a syntax error.
+
+The ``type`` parameter specifies the mount point type. If omitted, it defaults
+to ``"chroot"``. The ``path`` parameter must be an absolute path (i.e. must
+begin with ``/``).
 
 Gramine currently supports two types of mount points:
 
-* ``chroot``: Host-backed files. All host files and sub-directories found under
-  ``[URI]`` are forwarded to the Gramine instance and placed under ``[PATH]``.
-  For example, with a host-level path specified as
-  ``fs.mount.lib.uri = "file:/one/path/"`` and forwarded to Gramine via
-  ``fs.mount.lib.path = "/another/path"``, a host-level file
-  ``/one/path/file`` is visible to graminized application as
+* ``chroot`` (default): Host-backed files. All host files and sub-directories
+  found under ``[URI]`` are forwarded to the Gramine instance and placed under
+  ``[PATH]``. For example, with a host-level path specified as ``uri =
+  "file:/one/path/"`` and forwarded to Gramine via ``path = "/another/path"``, a
+  host-level file ``/one/path/file`` is visible to graminized application as
   ``/another/path/file``. This concept is similar to FreeBSD's chroot and to
   Docker's named volumes. Files under ``chroot`` mount points support mmap and
   fork/clone.
 
 * ``tmpfs``: Temporary in-memory-only files. These files are *not* backed by
   host-level files. The tmpfs files are created under ``[PATH]`` (this path is
-  empty on Gramine instance startup) and are destroyed when a Gramine
-  instance terminates. The ``[URI]`` parameter is always ignored. ``tmpfs``
-  is especially useful in trusted environments (like Intel SGX) for securely
-  storing temporary files. This concept is similar to Linux's tmpfs. Files
-  under ``tmpfs`` mount points currently do *not* support mmap and each process
-  has its own, non-shared tmpfs (i.e. processes don't see each other's files).
+  empty on Gramine instance startup) and are destroyed when a Gramine instance
+  terminates. The ``[URI]`` parameter is always ignored, and can be omitted.
+
+  ``tmpfs`` is especially useful in trusted environments (like Intel SGX) for
+  securely storing temporary files. This concept is similar to Linux's tmpfs.
+  Files under ``tmpfs`` mount points currently do *not* support mmap and each
+  process has its own, non-shared tmpfs (i.e. processes don't see each other's
+  files).
 
 Start (current working) directory
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
@@ -840,3 +870,15 @@ This usually points to the LibOS library ``libsysdb.so``.
 
 Note that previously this syntax allowed to specify the list of URIs (separated
 by commas). This ability was never used and therefore was removed completely.
+
+FS mount points (deprecated syntax)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+::
+
+   fs.mount.[identifier].type = "[chroot|...]"
+   fs.mount.[identifier].path = "[PATH]"
+   fs.mount.[identifier].uri  = "[URI]"
+
+This syntax used a TOML table schema with keys for each mount. It has been
+replaced with a TOML array.
--- a/LibOS/shim/src/fs/shim_fs.c
+++ b/LibOS/shim/src/fs/shim_fs.c
@@ -86,8 +86,8 @@ static void free_mount(struct shim_mount* mount) {
 
 static bool mount_migrated = false;
 
-static int __mount_root(void) {
-    int ret = 0;
+static int mount_root(void) {
+    int ret;
     char* fs_root_type = NULL;
     char* fs_root_uri  = NULL;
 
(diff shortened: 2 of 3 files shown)


mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/411)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

___
*[Documentation/manifest-syntax.rst, line 335 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw2PvAV7F8Fn-2ddJdT:-Mw2PvAV7F8Fn-2ddJdU:bzgg607) ([raw file](https://github.com/gramineproject/gramine/blob/6987eb795adcedc0539f58e37c7dc4fb41d4a5b1/Documentation/manifest-syntax.rst#L335)):*
> ```reStructuredText
> ^^^^^^^^^^^^^^^
> 
> ::
> ```

It was already like this before, but could you use TOML syntax highlighting here?
___
*[Documentation/manifest-syntax.rst, line 342 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw2Q84r7LLEi0OOI9k7:-Mw2Q84r7LLEi0OOI9k8:bwub6i8) ([raw file](https://github.com/gramineproject/gramine/blob/6987eb795adcedc0539f58e37c7dc4fb41d4a5b1/Documentation/manifest-syntax.rst#L342)):*
> ```reStructuredText
>     ]
> 
>     # or, as separate sections:
> ```

I'd rather move this comment to normal text and split these snippets into two (it will be more visible that these are alternatives, not just one long snippet).
___
*[Documentation/manifest-syntax.rst, line 383 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw2QmhU8NWLNIRSz5bu:-Mw2QmhU8NWLNIRSz5bv:b-j4dup8) ([raw file](https://github.com/gramineproject/gramine/blob/6987eb795adcedc0539f58e37c7dc4fb41d4a5b1/Documentation/manifest-syntax.rst#L383)):*
> ```reStructuredText
>   host-level files. The tmpfs files are created under ``[PATH]`` (this path is
>   empty on Gramine instance startup) and are destroyed when a Gramine instance
>   terminates. The ``[URI]`` parameter is always ignored, and can be omitted.
> ```

Maybe we should actually error out on this? (after some grace period with only a warning)



_Code quote:_
```reStructuredText
and can be omitted
```
___
*[LibOS/shim/src/fs/shim_fs.c, line 309 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw2c-9v22ZKpArm1cxg:-Mw2c-9w7d1ofI9PviW8:b-a2c3rw) ([raw file](https://github.com/gramineproject/gramine/blob/6987eb795adcedc0539f58e37c7dc4fb41d4a5b1/LibOS/shim/src/fs/shim_fs.c#L309)):*
> ```C
> }
> 
> static int mount_others_from_toml_array(void) {
> ```

What does this "others" mean here and in the function above? I was always puzzled by this `mount_one_other`.
___
*[LibOS/shim/src/fs/shim_fs.c, line 332 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw2cXsM3uYBoBubhBSg:-Mw2cXsN1WOraXeGFccs:b-eew34f) ([raw file](https://github.com/gramineproject/gramine/blob/6987eb795adcedc0539f58e37c7dc4fb41d4a5b1/LibOS/shim/src/fs/shim_fs.c#L332)):*
> ```C
>         }
> 
>         /* Prepare a key for `mount_one_other`, so that it in case of errors it will display paths
> ```

Too many "it"s?



_Code quote:_
```C
it in case of errors it will
```

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/411)* status: all files reviewed, 13 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @mkow and @pwmarcz)

___
*[Documentation/manifest-syntax.rst, line 342 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw2Q84r7LLEi0OOI9k7:-Mw62nR43162fHjs0mKh:b32v57) ([raw file](https://github.com/gramineproject/gramine/blob/6987eb795adcedc0539f58e37c7dc4fb41d4a5b1/Documentation/manifest-syntax.rst#L342)):*

I'd rather move this comment to normal text and split these snippets into two (it will be more visible that these are alternatives, not just one long snippet).
</blockquote></details>

+1
___
*[Documentation/manifest-syntax.rst, line 361 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw634Yr5nybnwtMgW3g:-Mw634Yr5nybnwtMgW3h:bo83vof) ([raw file](https://github.com/gramineproject/gramine/blob/6987eb795adcedc0539f58e37c7dc4fb41d4a5b1/Documentation/manifest-syntax.rst#L361)):*
> ```reStructuredText
> The ``fs.mount`` option is a TOML array of tables. As shown in the examples
> above, you can either specify its value using the inline syntax, or use multiple
> ``[[fs.mount]]`` sections.
> ```

Why is this paragraph useful? I think it only duplicates the example above. Developers (who know TOML) won't get anything new from this para, and casual users will only get confused. So I think we should remove this para.
___
*[Documentation/manifest-syntax.rst, line 366 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw63Ic-4N5dnjZDMIkc:-Mw63Ic-4N5dnjZDMIkd:b-ulyjcu) ([raw file](https://github.com/gramineproject/gramine/blob/6987eb795adcedc0539f58e37c7dc4fb41d4a5b1/Documentation/manifest-syntax.rst#L366)):*
> ```reStructuredText
>    If you use the inline table syntax (``{ ... }``), each table should fit in
>    a single line. While multi-line tables are currently accepted by Gramine,
>    they are not valid TOML and Gramine might disallow them in the future.
> ```

I would remove this note as well. This is a controversial feature of TOML, and chances are high that it will be the TOML spec that will allow this, not Gramine that will disallow this. Anyway, the users don't read these fine prints anyway, they just look at the examples under `CI-Examples` :)
___
*[LibOS/shim/src/fs/shim_fs.c, line 120 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw65DDaDnelxvspAFpa:-Mw65DDaDnelxvspAFpb:b-j6csfm) ([raw file](https://github.com/gramineproject/gramine/blob/6987eb795adcedc0539f58e37c7dc4fb41d4a5b1/LibOS/shim/src/fs/shim_fs.c#L120)):*
> ```C
>         ret = mount_fs("chroot", fs_root_uri, "/");
>     } else {
>         ret = mount_fs(fs_root_type, fs_root_uri ?: "", "/");
> ```

What is this `fs_root_uri ?: ""`? This case is like this:
```
fs_root_type = "tmpfs"
fs_root_uri = NULL
```

I guess you're replacing NULL with an empty string to not deal with a NULL corner case?
___
*[LibOS/shim/src/fs/shim_fs.c, line 208 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw6798M8cL5o7TL6EQX:-Mw6798M8cL5o7TL6EQY:b-xwor97) ([raw file](https://github.com/gramineproject/gramine/blob/6987eb795adcedc0539f58e37c7dc4fb41d4a5b1/LibOS/shim/src/fs/shim_fs.c#L208)):*
> ```C
>         ret = mount_fs("chroot", mount_uri, mount_path);
>     } else {
>         ret = mount_fs(mount_type, mount_uri ?: "", mount_path);
> ```

ditto (why empty string instead of NULL?)
___
*[LibOS/shim/src/fs/shim_fs.c, line 218 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw673bTAB3FCdjeGHS8:-Mw673bTAB3FCdjeGHS9:bja6p0k) ([raw file](https://github.com/gramineproject/gramine/blob/b571f2a3efda7d20db958ea1a7880b109acaa9c0/LibOS/shim/src/fs/shim_fs.c#L218)):*
> ```C
>             "Root mount / already exists, verify that there are no duplicate mounts in manifest\n"
>             "(note that root / is automatically mounted in Gramine and can be changed via "
>             "'fs.root' manifest entry).\n");
> ```

Why did you remove this?
___
*[LibOS/shim/src/fs/shim_fs.c, line 224 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw675sq0ECUf_gNs1sv:-Mw675sq0ECUf_gNs1sw:bja6p0k) ([raw file](https://github.com/gramineproject/gramine/blob/b571f2a3efda7d20db958ea1a7880b109acaa9c0/LibOS/shim/src/fs/shim_fs.c#L224)):*
> ```C
> 
>     if (!strcmp(mount_path, ".") || !strcmp(mount_path, "..")) {
>         log_error("Mount points '.' and '..' are not allowed, remove them from manifest.");
> ```

Why did you remove this?
___
*[LibOS/shim/test/regression/manifest.template, line 18 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw64L0CAtZd83T_jvG6:-Mw64L0CAtZd83T_jvG7:b-eozsn3) ([raw file](https://github.com/gramineproject/gramine/blob/6987eb795adcedc0539f58e37c7dc4fb41d4a5b1/LibOS/shim/test/regression/manifest.template#L18)):*
> ```
>   { path = "/usr/{{ arch_libdir }}", uri = "file:/usr/{{ arch_libdir }}" },
>   { path = "/bin", uri = "file:/bin" },
>   { type = "tmpfs", path = "/mnt/tmpfs" },
> ```

Maybe separate this `tmpfs` entry with a newline? For readability.
___
*[LibOS/shim/test/regression/manifest.template, line 19 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw647upA92rbO_35qPA:-Mw647uq0DQ6_xEhSr37:byr4dgo) ([raw file](https://github.com/gramineproject/gramine/blob/6987eb795adcedc0539f58e37c7dc4fb41d4a5b1/LibOS/shim/test/regression/manifest.template#L19)):*
> ```
>   { path = "/bin", uri = "file:/bin" },
>   { type = "tmpfs", path = "/mnt/tmpfs" },
> ]
> ```

Could we indent `uri` fields at the same level, so it's easier to read?

Right now it's messy and very hard to read: https://github.com/gramineproject/gramine/blob/pawel/mount-array-2/LibOS/shim/test/regression/manifest.template

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/411)* status: 0 of 3 files reviewed, 13 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "squash! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @pwmarcz)

___
*[Documentation/manifest-syntax.rst, line 335 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw2PvAV7F8Fn-2ddJdT:-Mw67_geAOVbGeb2-cTi:bjd0a75) ([raw file](https://github.com/gramineproject/gramine/blob/6987eb795adcedc0539f58e37c7dc4fb41d4a5b1/Documentation/manifest-syntax.rst#L335)):*

It was already like this before, but could you use TOML syntax highlighting here?
</blockquote></details>

Done (the `highlight` directive at the top).
___
*[Documentation/manifest-syntax.rst, line 342 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw2Q84r7LLEi0OOI9k7:-Mw6KNp01GtRmcKXhB9V:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/6987eb795adcedc0539f58e37c7dc4fb41d4a5b1/Documentation/manifest-syntax.rst#L342)):*

+1
</blockquote></details>

Done.
___
*[Documentation/manifest-syntax.rst, line 361 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw634Yr5nybnwtMgW3g:-Mw6Hd401lbyPET-XH5a:b-gij58x) ([raw file](https://github.com/gramineproject/gramine/blob/6987eb795adcedc0539f58e37c7dc4fb41d4a5b1/Documentation/manifest-syntax.rst#L361)):*

Why is this paragraph useful? I think it only duplicates the example above. Developers (who know TOML) won't get anything new from this para, and casual users will only get confused. So I think we should remove this para.
</blockquote></details>

OK, I agree it doesn't add much.
___
*[Documentation/manifest-syntax.rst, line 366 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw63Ic-4N5dnjZDMIkc:-Mw6Hi0UCgNSCUhoLPCi:bf6hzq1) ([raw file](https://github.com/gramineproject/gramine/blob/6987eb795adcedc0539f58e37c7dc4fb41d4a5b1/Documentation/manifest-syntax.rst#L366)):*

I would remove this note as well. This is a controversial feature of TOML, and chances are high that it will be the TOML spec that will allow this, not Gramine that will disallow this. Anyway, the users don't read these fine prints anyway, they just look at the examples under `CI-Examples` :)
</blockquote></details>

The actual problem here is that we disallow a trailing comma in tables. This will work in Gramine:

```toml
fs.mount = [
  {
    path = "/bin",
    uri = "file:/usr/bin"
  },
]
```

This will break with a cryptic message from the Python library (when parsed by `gramine-manifest`):

```toml
fs.mount = [
  {
    path = "/bin",
    uri = "file:/usr/bin",
  },
]
```

If I don't discourage multi-line tables here, I think I need to at least add a warning about the traling comma. Do you think that's a better option?
___
*[Documentation/manifest-syntax.rst, line 383 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw2QmhU8NWLNIRSz5bu:-Mw67vJzDDuJ5mbzwGTJ:b-divrjv) ([raw file](https://github.com/gramineproject/gramine/blob/6987eb795adcedc0539f58e37c7dc4fb41d4a5b1/Documentation/manifest-syntax.rst#L383)):*

Maybe we should actually error out on this? (after some grace period with only a warning)
</blockquote></details>

Should we? I don't think it hurts us to allow it, and it only makes the rules/code more complicated.
___
*[LibOS/shim/src/fs/shim_fs.c, line 120 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw65DDaDnelxvspAFpa:-Mw69K40CVwOKkQmHDiG:b-y8c3p0) ([raw file](https://github.com/gramineproject/gramine/blob/6987eb795adcedc0539f58e37c7dc4fb41d4a5b1/LibOS/shim/src/fs/shim_fs.c#L120)):*

What is this `fs_root_uri ?: ""`? This case is like this:
```
fs_root_type = "tmpfs"
fs_root_uri = NULL
```

I guess you're replacing NULL with an empty string to not deal with a NULL corner case?
</blockquote></details>

That's right. This parameter goes to the individual filesystems (chroot, pseudo-FS) and I don't want them all to check against NULL too.
___
*[LibOS/shim/src/fs/shim_fs.c, line 218 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw673bTAB3FCdjeGHS8:-Mw6A_IM8BvpByV76OO8:bz2frhb) ([raw file](https://github.com/gramineproject/gramine/blob/b571f2a3efda7d20db958ea1a7880b109acaa9c0/LibOS/shim/src/fs/shim_fs.c#L218)):*

Why did you remove this?
</blockquote></details>

Right, this should stay (I reworded it though).
___
*[LibOS/shim/src/fs/shim_fs.c, line 224 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw675sq0ECUf_gNs1sv:-Mw6B17FB6UHbmGbfEEh:b70rxma) ([raw file](https://github.com/gramineproject/gramine/blob/b571f2a3efda7d20db958ea1a7880b109acaa9c0/LibOS/shim/src/fs/shim_fs.c#L224)):*

Why did you remove this?
</blockquote></details>

That one was by accident too. I restored it.

Actually, this rule is kind of arbitrary. As discussed on Gitter, I'd like to deprecate all relative paths: they are interpreted relative to root anyway.
___
*[LibOS/shim/src/fs/shim_fs.c, line 309 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw2c-9v22ZKpArm1cxg:-Mw6D2Zo7LOS15OIbJwo:b-6rcgzn) ([raw file](https://github.com/gramineproject/gramine/blob/6987eb795adcedc0539f58e37c7dc4fb41d4a5b1/LibOS/shim/src/fs/shim_fs.c#L309)):*

What does this "others" mean here and in the function above? I was always puzzled by this `mount_one_other`.
</blockquote></details>

Filesystems other than the root one (`fs.root`). I can't think of a better name now, but I can change if you suggest one. :)
___
*[LibOS/shim/src/fs/shim_fs.c, line 332 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw2cXsM3uYBoBubhBSg:-Mw6DQbg3ht-Q5pujwyx:bidrs83) ([raw file](https://github.com/gramineproject/gramine/blob/6987eb795adcedc0539f58e37c7dc4fb41d4a5b1/LibOS/shim/src/fs/shim_fs.c#L332)):*

Too many "it"s?
</blockquote></details>

Right. Fixed.
___
*[LibOS/shim/test/regression/manifest.template, line 18 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw64L0CAtZd83T_jvG6:-Mw6KKMG4NCNATBc2JAK:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/6987eb795adcedc0539f58e37c7dc4fb41d4a5b1/LibOS/shim/test/regression/manifest.template#L18)):*

Maybe separate this `tmpfs` entry with a newline? For readability.
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/manifest.template, line 19 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw647upA92rbO_35qPA:-Mw6JBtO3CgRMi4C9deY:b-6assxz) ([raw file](https://github.com/gramineproject/gramine/blob/6987eb795adcedc0539f58e37c7dc4fb41d4a5b1/LibOS/shim/test/regression/manifest.template#L19)):*

Could we indent `uri` fields at the same level, so it's easier to read?

Right now it's messy and very hard to read: https://github.com/gramineproject/gramine/blob/pawel/mount-array-2/LibOS/shim/test/regression/manifest.template
</blockquote></details>

I don't like this. It might not matter too much for this specific manifest, but I don't think that should be our recommended practice. 

In the case of `fs.mount`, the paths can get quite long, e.g. if you specify individual files. Adding a new path will likely mean having to realign the whole array, and break git diff/blame in the process.

If the inline syntax is hard to read without such special formatting, maybe that means we should not use it? We can instead use the multi-line syntax (but it's non-standard, and annoying because of the trailing comma thing), or TOML sections.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/411)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "squash! " found in commit messages' one-liners (waiting on @mkow and @pwmarcz)

___
*[Documentation/manifest-syntax.rst, line 366 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw63Ic-4N5dnjZDMIkc:-Mw6WKSrAWTEBxHsqEZK:b-rf0xae) ([raw file](https://github.com/gramineproject/gramine/blob/6987eb795adcedc0539f58e37c7dc4fb41d4a5b1/Documentation/manifest-syntax.rst#L366)):*
> If I don't discourage multi-line tables here, I think I need to at least add a warning about the traling comma. Do you think that's a better option?

Yes, I would prefer this warning (about the trailing comma).
___
*[LibOS/shim/test/regression/manifest.template, line 19 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw647upA92rbO_35qPA:-Mw6WxooEwf8RBSU2XeQ:b6mufu3) ([raw file](https://github.com/gramineproject/gramine/blob/6987eb795adcedc0539f58e37c7dc4fb41d4a5b1/LibOS/shim/test/regression/manifest.template#L19)):*
>  Adding a new path will likely mean having to realign the whole array, and break git diff/blame in the process.

That's a good point. Let's not re-align then, but keep your current version.

> If the inline syntax is hard to read without such special formatting, maybe that means we should not use it? We can instead use the multi-line syntax (but it's non-standard, and annoying because of the trailing comma thing), or TOML sections.

No, the current version is definitely better than the verbose TOML sections. I like it very much that this manifest is short and crisp.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/411)* status: 2 of 3 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[Documentation/manifest-syntax.rst, line 366 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw63Ic-4N5dnjZDMIkc:-Mw6fynOAqYPkVSrY0dh:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/6987eb795adcedc0539f58e37c7dc4fb41d4a5b1/Documentation/manifest-syntax.rst#L366)):*

> If I don't discourage multi-line tables here, I think I need to at least add a warning about the traling comma. Do you think that's a better option?

Yes, I would prefer this warning (about the trailing comma).
</blockquote></details>

Done.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/411)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow)

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/411)* status: 2 of 3 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[Documentation/manifest-syntax.rst, line 335 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw2PvAV7F8Fn-2ddJdT:-Mw7Q4DG0IpHJJZwdNAS:ble5wou) ([raw file](https://github.com/gramineproject/gramine/blob/6987eb795adcedc0539f58e37c7dc4fb41d4a5b1/Documentation/manifest-syntax.rst#L335)):*

Done (the `highlight` directive at the top).
</blockquote></details>

Sorry, I actually cannot :) See the TODO I added.

Probably the best solution is to change the "meta-syntax" in this file, but I'm out of ideas, and I don't want to solve it right now...

mkow (commented):
> Reviewed 2 of 3 files at r2, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/411)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[Documentation/manifest-syntax.rst, line 335 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw2PvAV7F8Fn-2ddJdT:-Mw7VlVV9YvUHxVxQCl8:b-sq4802) ([raw file](https://github.com/gramineproject/gramine/blob/6987eb795adcedc0539f58e37c7dc4fb41d4a5b1/Documentation/manifest-syntax.rst#L335)):*

Sorry, I actually cannot :) See the TODO I added.

Probably the best solution is to change the "meta-syntax" in this file, but I'm out of ideas, and I don't want to solve it right now...
</blockquote></details>

Ah, too bad :( Let's leave it for later then.
___
*[Documentation/manifest-syntax.rst, line 383 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw2QmhU8NWLNIRSz5bu:-Mw7WeriElH8XLCFG62J:b-71dndh) ([raw file](https://github.com/gramineproject/gramine/blob/6987eb795adcedc0539f58e37c7dc4fb41d4a5b1/Documentation/manifest-syntax.rst#L383)):*

Should we? I don't think it hurts us to allow it, and it only makes the rules/code more complicated.
</blockquote></details>

My idea was that it could prevent user mistakes / confusion when they misunderstand how it works and set it, but then it's silently ignored. But not blocking.
___
*[LibOS/shim/src/fs/shim_fs.c, line 309 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw2c-9v22ZKpArm1cxg:-Mw7YYnA8DalZwbXhmE_:bcvdwit) ([raw file](https://github.com/gramineproject/gramine/blob/6987eb795adcedc0539f58e37c7dc4fb41d4a5b1/LibOS/shim/src/fs/shim_fs.c#L309)):*

Filesystems other than the root one (`fs.root`). I can't think of a better name now, but I can change if you suggest one. :)
</blockquote></details>

Maybe just `nonroot`?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/411)* status: 2 of 3 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[Documentation/manifest-syntax.rst, line 383 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw2QmhU8NWLNIRSz5bu:-Mw7hGYg71GPh_sNlLd0:b-6m8qwa) ([raw file](https://github.com/gramineproject/gramine/blob/6987eb795adcedc0539f58e37c7dc4fb41d4a5b1/Documentation/manifest-syntax.rst#L383)):*

My idea was that it could prevent user mistakes / confusion when they misunderstand how it works and set it, but then it's silently ignored. But not blocking.
</blockquote></details>

OK. I want to leave it like this for now, because this section of manifest is going to change anyway (with trusted/protected files) and keeping the rules simple will help me refactor. But let's watch out for user problems.
___
*[LibOS/shim/src/fs/shim_fs.c, line 309 at r1](https://reviewable.io/reviews/gramineproject/gramine/411#-Mw2c-9v22ZKpArm1cxg:-Mw7hmdE2TzsVyhFRYRt:b-k1y5pj) ([raw file](https://github.com/gramineproject/gramine/blob/6987eb795adcedc0539f58e37c7dc4fb41d4a5b1/LibOS/shim/src/fs/shim_fs.c#L309)):*

Maybe just `nonroot`?
</blockquote></details>

OK. Done.

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/411)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

dimakuv (dismissed):
> Reviewed 1 of 1 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/411)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/411)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/411)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #416, opened 2022-02-18
[Pal/Linux-SGX] Fix header conflict in `sgx_main.c`

pwmarcz:
## Description of the changes 

The problem with headers was because we imported <sys/auxv.h>, which in turns imports <elf.h>, which clashes with our own "elf.h". That was because we used the `getauxval` function from glibc. Instead, we manually parse the ELF aux vector.

## How to test this PR? 

I used a temporary log line to check if we retrieve the same value of `AT_HWCAP2`.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/416)

### DIFF
--- a/Pal/src/host/Linux-SGX/sgx_main.c
+++ b/Pal/src/host/Linux-SGX/sgx_main.c
@@ -6,11 +6,10 @@
  *                    Michał Kowalczyk <mkow@invisiblethingslab.com>
  */
 
-/* FIXME: Sorting+re-grouping includes here causes tons of
- * "../../../include/sysdeps/generic/ldsodefs.h:30:32: error: unknown type name ‘Elf__ELF_NATIVE_CLASS_Addr’
- *   #define ElfW(type)       _ElfW(Elf, __ELF_NATIVE_CLASS, type)"
- * errors.
- */
+#include <asm/errno.h>
+#include <asm/fcntl.h>
+#include <asm/socket.h>
+#include <linux/fs.h>
 
 #include "asan.h"
 #include "debug_map.h"
@@ -28,12 +27,6 @@
 #include "toml.h"
 #include "toml_utils.h"
 #include "topo_info.h"
-
-#include <asm/errno.h>
-#include <asm/fcntl.h>
-#include <asm/socket.h>
-#include <linux/fs.h>
-#include <sys/auxv.h>
 
 const size_t g_page_size = PRESET_PAGESIZE;
 
@@ -1022,6 +1015,19 @@ noreturn static void print_usage_and_exit(const char* argv_0) {
     die_or_inf_loop();
 }
 
+static int get_aux_value(char** envp, uint64_t type, uint64_t* out_value) {
+    while (*envp)
+        envp++;
+
+    for (elf_auxv_t* auxv = (elf_auxv_t*)(envp + 1); auxv->a_type != AT_NULL; auxv++) {
+        if (auxv->a_type == type) {
+            *out_value = auxv->a_un.a_val;
+            return 0;
+        }
+    }
+    return -1;
+}
+
 #ifdef ASAN
 /*
  * HACK: `setup_asan` is not called inside `main`, but defined as a constructor with a priority of
@@ -1076,7 +1082,8 @@ int main(int argc, char* argv[], char* envp[]) {
     }
 
     /* check whether host kernel supports FSGSBASE feature, otherwise we need the GSGX driver */
-    if (getauxval(AT_HWCAP2) & 0x2) {
+    uint64_t at_hwcap2;
+    if (get_aux_value(envp, AT_HWCAP2, &at_hwcap2) == 0 && (at_hwcap2 & 0x2)) {
         need_gsgx = false;
     }
 


boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/416)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/416#-MwDSCykAXSn_rwxU3t3:-MwDSCyl35mq3cagecrW:b-t76nm0) (no related file):*
Why can't we just declare `getauxval` as extern symbol?
Or even better, fix the original problem, which is importing Gramine elf.h in untrusted PAL?

---

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/416)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/416#-MwDSCykAXSn_rwxU3t3:-MwQn4fa9iHuXc9HAb0f:b-2f0qgc) (no related file):*
>Why can't we just declare `getauxval` as extern symbol?

I assumed that even though we link against glibc, we want to rely on it as little as possible (so that e.g. we might stop linking against it in the future).

We have no written policy for such cases, though. Any opinions, @mkow @dimakuv?

> Or even better, fix the original problem, which is importing Gramine elf.h in untrusted PAL?

This file processes an ELF binary (libsysdb) to prepare the enclave, so it needs ELF struct definitions.

I'd say the "original problem" is including a `<sys/*>` header, which we almost never do because they tend to conflict with Gramine's own.

---

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/416)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required) (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/416#-MwDSCykAXSn_rwxU3t3:-MwQqvId8bkZVo8t9mTl:b-qlddo7) (no related file):*
> I assumed that even though we link against glibc, we want to rely on it as little as possible (so that e.g. we might stop linking against it in the future).

Yes, we definitely should avoid linking against Glibc as little as possible. This way we'll have more shared code between different PALs, and just in general *not* using Glibc is a cleaner approach in my opinion.

I prefer Pawel's solution to the problem (i.e., I prefer this PR).

---

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/416)* status: all files reviewed, 1 unresolved discussion (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/416#-MwDSCykAXSn_rwxU3t3:-MwRBWaM69mBead5CUHz:bk0lqnc) (no related file):*

> I assumed that even though we link against glibc, we want to rely on it as little as possible (so that e.g. we might stop linking against it in the future).

Yes, we definitely should avoid linking against Glibc as little as possible. This way we'll have more shared code between different PALs, and just in general *not* using Glibc is a cleaner approach in my opinion.

I prefer Pawel's solution to the problem (i.e., I prefer this PR).
</blockquote></details>

I'm fine with the current PR, mostly because I don't like any of the solutions, so whatever. The problem is that we use these Gramine-internal common headers in a context with system headers included, while it's designed for the env without them. I'm not sure how to solve it better :/

---

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/416)* status: all files reviewed, 1 unresolved discussion (waiting on @dimakuv, @mkow, and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/416#-MwDSCykAXSn_rwxU3t3:-MwSbujS4zdbjWIRTnLH:b7wgatm) (no related file):*
> I assumed that even though we link against glibc, we want to rely on it as little as possible (so that e.g. we might stop linking against it in the future).

I've never heard that we want to stop linking against glibc (actually, why would we) and I don't think it's possible - we link against protobuf (which itself links against glibc).

> This file processes an ELF binary (libsysdb) to prepare the enclave, so it needs ELF struct definitions.

So it should be fine to include system headers.

> I'd say the "original problem" is including a <sys/*> header, which we almost never do because they tend to conflict with Gramine's own.

I disagree. This is untrusted PAL, which is separate from LibOS and trusted PAL, where we want to use just Gramine headers.

---

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/416)* status: all files reviewed, 1 unresolved discussion (waiting on @dimakuv, @mkow, and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/416#-MwDSCykAXSn_rwxU3t3:-MwSdjhs2072ynMa_djE:blmqc76) (no related file):*

> I assumed that even though we link against glibc, we want to rely on it as little as possible (so that e.g. we might stop linking against it in the future).

I've never heard that we want to stop linking against glibc (actually, why would we) and I don't think it's possible - we link against protobuf (which itself links against glibc).

> This file processes an ELF binary (libsysdb) to prepare the enclave, so it needs ELF struct definitions.

So it should be fine to include system headers.

> I'd say the "original problem" is including a <sys/*> header, which we almost never do because they tend to conflict with Gramine's own.

I disagree. This is untrusted PAL, which is separate from LibOS and trusted PAL, where we want to use just Gramine headers.
</blockquote></details>

Since you all seem to be fine with this PR I'm unblocking, but I disagree with reasoning in the commit message

---

boryspoplawski (approved):

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/416)* status: all files reviewed, 1 unresolved discussion (waiting on @boryspoplawski, @mkow, and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/416#-MwDSCykAXSn_rwxU3t3:-MwVSjwo25JsgfRI78Gy:bq9tl95) (no related file):*
> I've never heard that we want to stop linking against glibc (actually, why would we) and I don't think it's possible - we link against protobuf (which itself links against glibc).

I would prefer to link against a static version of protobuf which in turn links against a static musl. I think we discussed this at some point -- that we could move to static musl and have a completely self-contained untrusted-SGX PAL binary.

---

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/416)* status: all files reviewed, 1 unresolved discussion (waiting on @boryspoplawski, @dimakuv, and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/416#-MwDSCykAXSn_rwxU3t3:-MwVtfsp-NcYSd0ctFgR:b-mfhwrr) (no related file):*
>I disagree. This is untrusted PAL, which is separate from LibOS and trusted PAL, where we want to use just Gramine headers.

It's a separate executable, but it's very much *not* separate in terms of code. Untrusted PAL includes a lot of Gramine's headers, is linked against the "common" library, and so on. If we wanted to enable the use of glibc headers, there would be a lot of work to do in terms of untangling that mess.

---

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/416)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

pwmarcz:
Jenkins, retest Jenkins-18.04 please (`fdatasync01` LTP test failed)

PR #418, opened 2022-02-18
[Pal,LibOS] Add arch_prctl XSTATE component permission options for AMX

dongx1x:
From linux kernel v5.16, Advanced Matrix Extensions (AMX) is in upstream with XFD handled permission control, expose the APIs and add the permission for XFD handling the #NM exception, refer 3.2.6 in
https://software.intel.com/content/dam/develop/external/us/en/documents-tps/architecture-instruction-set-extensions-programming-reference.pdf

Signed-off-by: Dong, Xiaocheng <xiaocheng.dong@intel.com>

## Description of the changes 
Expose more arch_prctl XSTATE component permission APIs for AMX

## How to test this PR? 
Running AMX enabled workloads in a AMX enabled system

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/418)

### DIFF
--- a/LibOS/shim/src/arch/x86_64/shim_arch_prctl.c
+++ b/LibOS/shim/src/arch/x86_64/shim_arch_prctl.c
@@ -12,6 +12,20 @@
 #include "shim_table.h"
 #include "shim_tcb.h"
 
+/* To support AMX in linux kernel v5.16, prctl add XSTATE components permission
+ * control APIs, refer:
+ * https://elixir.bootlin.com/linux/v5.16/source/arch/x86/include/uapi/asm/prctl.h
+ */
+#if !defined(ARCH_GET_XCOMP_SUPP)
+#define ARCH_GET_XCOMP_SUPP	0x1021
+#endif
+#if !defined(ARCH_GET_XCOMP_PERM)
+#define ARCH_GET_XCOMP_PERM	0x1022
+#endif
+#if !defined(ARCH_REQ_XCOMP_PERM)
+#define ARCH_REQ_XCOMP_PERM	0x1023
+#endif
+
 long shim_do_arch_prctl(int code, unsigned long addr) {
     switch (code) {
         case ARCH_SET_FS:
@@ -20,6 +34,12 @@ long shim_do_arch_prctl(int code, unsigned long addr) {
 
         case ARCH_GET_FS:
             return pal_to_unix_errno(DkSegmentBaseGet(PAL_SEGMENT_FS, (unsigned long*)addr));
+
+        case ARCH_GET_XCOMP_SUPP:
+        case ARCH_GET_XCOMP_PERM:
+            return pal_to_unix_errno(DkXCompPerm(code, (unsigned long*)addr));
+        case ARCH_REQ_XCOMP_PERM:
+            return pal_to_unix_errno(DkXCompPerm(code, &addr));
 
         default:
             log_warning("Not supported flag (0x%x) passed to arch_prctl", code);
--- a/Pal/include/pal/pal.h
+++ b/Pal/include/pal/pal.h
@@ -695,6 +695,17 @@ int DkSegmentBaseGet(enum pal_segment_reg reg, uintptr_t* addr);
 int DkSegmentBaseSet(enum pal_segment_reg reg, uintptr_t addr);
 
 /*!
+ * \brief Get/Request XSTATE components permission
+ *
+ * \param[in] code the option of get/requst permission
+ * \param[in,out] addr the bit number in XFD as input 
+ *                     or a filled value with the permssion as output
+ *
+ * \return 0 on success, negative error value on failure
+ */
+int DkXCompPerm(int code, unsigned long* addr);
+
+/*!
  * \brief Return the amount of currently available memory for LibOS/application
  * usage.
  */
--- a/Pal/include/pal_internal.h
+++ b/Pal/include/pal_internal.h
@@ -226,6 +226,7 @@ int _DkRandomBitsRead(void* buffer, size_t size);
 double _DkGetBogomips(void);
 int _DkSegmentBaseGet(enum pal_segment_reg reg, uintptr_t* addr);
 int _DkSegmentBaseSet(enum pal_segment_reg reg, uintptr_t addr);
+int _DkXCompPerm(int code, unsigned long* addr);
 int _DkCpuIdRetrieve(uint32_t leaf, uint32_t subleaf, uint32_t values[4]);
 int _DkAttestationReport(const void* user_report_data, PAL_NUM* user_report_data_size,
                          void* target_info, PAL_NUM* target_info_size, void* report,
--- a/Pal/src/db_misc.c
+++ b/Pal/src/db_misc.c
@@ -24,6 +24,10 @@ int DkSegmentBaseGet(enum pal_segment_reg reg, uintptr_t* addr) {
 int DkSegmentBaseSet(enum pal_segment_reg reg, uintptr_t addr) {
     return _DkSegmentBaseSet(reg, addr);
 }
+
+int DkXCompPerm(int code, unsigned long* addr) {
+    return _DkXCompPerm(code, addr);
+}
 #endif
 
 PAL_NUM DkMemoryAvailableQuota(void) {
--- a/Pal/src/host/Linux-SGX/db_misc.c
+++ b/Pal/src/host/Linux-SGX/db_misc.c
@@ -719,3 +719,9 @@ int _DkSegmentBaseSet(enum pal_segment_reg reg, uintptr_t addr) {
             return -PAL_ERROR_INVAL;
     }
 }
+
+int _DkXCompPerm(int code, unsigned long* addr) {
+    int ret = ocall_xcomp_perm(code, addr);
+
+    return ret;
+}
--- a/Pal/src/host/Linux-SGX/enclave_ocalls.c
+++ b/Pal/src/host/Linux-SGX/enclave_ocalls.c
@@ -1945,3 +1945,29 @@ int ocall_sched_getaffinity(void* tcs, size_t cpumask_size, void* cpu_mask) {
     sgx_reset_ustack(old_ustack);
     return retval;
 }
+
+int ocall_xcomp_perm(int code, unsigned long* addr) {
+    int retval = 0;
+    ms_ocall_xcomp_perm_t* ms;
+    unsigned long data = *addr;
+
+    void* old_ustack = sgx_prepare_ustack();
+    ms = sgx_alloc_on_ustack_aligned(sizeof(*ms), alignof(*ms));
+    if (!ms) {
+        sgx_reset_ustack(old_ustack);
+        return -EPERM;
+    }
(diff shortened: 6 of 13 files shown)


boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/418)* status: 0 of 13 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dongx1x)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/418#-MwDUcNqAeSmC2uF5opJ:-MwDUcNqAeSmC2uF5opK:bw5ncqb) (no related file):*
1. This PR just blindly forwards `arch_prctl` syscall to host OS, with no verification whatsoever.
2. What's the point? From what I understand you cannot enter an enclave which requests bit in XFRM that are not currently enabled. This means that everything that an enclave requires must be enabled prior to first EENTER and enabling anything later on is pointless. Hence there is no need for this syscall at all.

---

dongx1x:
Hi @boryspoplawski , thanks for the review, I submit this PR because latest AMX implementation added [permission controls for dynamic XSTATE components](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/arch/x86/kernel/fpu/xstate.c?h=v5.16&id=db8268df0983adc2bb1fb48c9e5f7bfbb5f617f3), and workloads like oneDNN accelerated by AMX have [this syscall](https://github.com/oneapi-src/oneDNN/commit/bd3f7bca5bcc14275eb1485c475685a4372539ab) now, and would fail in Gramine, so:
1. The syscall would read or change current thead.fpu.perm, and allocate more XSATE memory if needed, what do you think we should verify in Gramine? I am new to develop Gramine, any suggestion is appreciated.
2. Do syscall once before EENTER is because this will add permission to the thread of `loader`, and the child process will [inherit its permission](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/arch/x86/kernel/fpu/core.c?h=v5.16#n518), then all the exception could be handled by XFD handler.

You can try to run latest AI workloads in Gramine in latest kernel (>5.16) in a AMX enabled system to check the failures. 
Thanks.

dimakuv (commented):
@dongx1x Thanks for the PR! I am actually aware on the AMX feature, and I recently got into the same problem of Linux 5.16 requiring additional enabling of this feature (via `arch_prctl(ARCH_REQ_XCOMP_PERM, AMX_TILEDATA)`).

In addition to your links, this one is also very good: https://lwn.net/Articles/874846/

However, as @boryspoplawski mentions, this PR actually does two things:
1. Enables the AMX feature on Linux 5.16 (via `arch_prctl(ARCH_REQ_XCOMP_PERM, AMX_TILEDATA)` during `create_enclave()`). This enables the AMX feature unconditionally for the Linux-SGX PAL of Gramine.
2. Blindly forwards `arch_prctl()` syscall from the app through Gramine to the host, with no verification.

I think that item 1 is meaningful, but item 2 is wrong (as Borys also mentions).

I propose the following scheme:
1. In Linux-SGX PAL: We enable the AMX feature via `arch_prctl(ARCH_REQ_XCOMP_PERM, AMX_TILEDATA)` based on the setting of `secs->attributes.xfrm` AMX TILE bit.
2. In Linux PAL: We enable the AMX feature via `arch_prctl(ARCH_REQ_XCOMP_PERM, AMX_TILEDATA)` *unconditionally*, because we don't have anything to give us a hint whether the app on top of Gramine will use AMX or not. Linux is smart enough to not actually enable AMX immediately -- this only happens on the first AMX instruction (which results in #NM, and the Linux exception handler finally enables AMX at this point.
3. In LibOS: we *emulate* `ARCH_GET_XCOMP_SUPP`, `ARCH_GET_XCOMP_PERM`, `ARCH_REQ_XCOMP_PERM` commands. For now, we can just emulate them by quering the CPUID (`DkCpuIdRetrieve()`) on the AMX leaf, and report only the AMX leaf (these commands are not used for anything else anyway currently).

@dongx1x Do you want to try to implement this, or should I do it (based on your PR)?

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/418#-:-MwQFprU37AeNVCU6WH2:b92tk89)* status: 0 of 13 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dongx1x)

dongx1x:
@dimakuv Please help to do that, in case I did not get your points clearly, I am still new to Gramine, thanks!

dimakuv:
> @dimakuv Please help to do that, in case I did not get your points clearly, I am still new to Gramine, thanks!

@dongx1x Sorry, I didn't understand what exactly you want us to do.

1. Do you want *me* to implement it based on your PR (I will create a new PR).
2. Or do you want to implement my proposed solution in your PR?

If you choose (1), then you don't need to do anything. I will take your PR as a baseline and re-implement it according to my comments.

If you choose (2), then I suggest you to close this PR and create a new PR. In the new PR, please follow my solution I outlined in my comments. If you have questions on how exactly to implement this PR, please ask specific questions.

dongx1x:
I choose option (2), I want to try to do that and I have questions to follow your comments:

- For _3_ in your proposal, do you mean to emulate `ARCH_GET_XCOMP_SUPP` and `ARCH_GET_XCOMP_PERM` by querying CPUID AMX leaf? The `ARCH_REQ_XCOMP_PERM` will go to Linux PAL or Linux-SGX PAL in my understanding.
- For _1_ and _2_, I only need add a condition for Linux-SGX PAL based on this PR, correct?

dimakuv:
> - For _3_ in your proposal, do you mean to emulate `ARCH_GET_XCOMP_SUPP` and `ARCH_GET_XCOMP_PERM` by querying CPUID AMX leaf?

Yes. Actually, not the CPUID AMX leaf, but the `EXTENDED_STATE_LEAF` (0xD) with subleaf 0x0 -- it returns the AMX bits (17 and 18) if AMX is present and enabled in the SGX enclave, so we can simply use this info.

> The `ARCH_REQ_XCOMP_PERM` will go to Linux PAL or Linux-SGX PAL in my understanding.

No. It will be a no-op and simply:
- return success if the call is `arch_prctl(ARCH_REQ_XCOMP_PERM, XFEATURE_XTILEDATA)` *and* CPUID leaf `EXTENDED_STATE_LEAF` (0xD) with subleaf 0x0 returns the AMX bits in EAX register.
- return `-EINVAL` if the call is `arch_prctl(ARCH_REQ_XCOMP_PERM, XFEATURE_XTILEDATA)` *and* CPUID leaf `EXTENDED_STATE_LEAF` (0xD) with subleaf 0x0 does *not* the AMX bits in EAX register.
- return `-EOPNOTSUPP` if the call is `arch_prctl(ARCH_REQ_XCOMP_PERM, something-else)`.

Please note: in my proposal, LibOS *never* forwards `arch_prctl()` to the PAL.

> For _1_ and _2_, I only need add a condition for Linux-SGX PAL based on this PR, correct?

For _1_, you can take the code from this PR, yes.

For _2_, you can take the same code from this PR, but without the `if secs.xfrm ...` check.

P.S. For more context, please see https://github.com/gramineproject/gramine/blob/master/Pal/src/host/Linux-SGX/db_misc.c#L187 and that file in general -- it has a lot of code and explanations on how we treat CPUID leaves and the AMX feature.

dongx1x:
> > * For _3_ in your proposal, do you mean to emulate `ARCH_GET_XCOMP_SUPP` and `ARCH_GET_XCOMP_PERM` by querying CPUID AMX leaf?
> 
> Yes. Actually, not the CPUID AMX leaf, but the `EXTENDED_STATE_LEAF` (0xD) with subleaf 0x0 -- it returns the AMX bits (17 and 18) if AMX is present and enabled in the SGX enclave, so we can simply use this info.
> 
> > The `ARCH_REQ_XCOMP_PERM` will go to Linux PAL or Linux-SGX PAL in my understanding.
> 
> No. It will be a no-op and simply:
> 
> * return success if the call is `arch_prctl(ARCH_REQ_XCOMP_PERM, XFEATURE_XTILEDATA)` _and_ CPUID leaf `EXTENDED_STATE_LEAF` (0xD) with subleaf 0x0 returns the AMX bits in EAX register.
> * return `-EINVAL` if the call is `arch_prctl(ARCH_REQ_XCOMP_PERM, XFEATURE_XTILEDATA)` _and_ CPUID leaf `EXTENDED_STATE_LEAF` (0xD) with subleaf 0x0 does _not_ the AMX bits in EAX register.
> * return `-EOPNOTSUPP` if the call is `arch_prctl(ARCH_REQ_XCOMP_PERM, something-else)`.
> 
> Please note: in my proposal, LibOS _never_ forwards `arch_prctl()` to the PAL.
> 
> > For _1_ and _2_, I only need add a condition for Linux-SGX PAL based on this PR, correct?
> 
> For _1_, you can take the code from this PR, yes.
> 
> For _2_, you can take the same code from this PR, but without the `if secs.xfrm ...` check.
> 
> P.S. For more context, please see https://github.com/gramineproject/gramine/blob/master/Pal/src/host/Linux-SGX/db_misc.c#L187 and that file in general -- it has a lot of code and explanations on how we treat CPUID leaves and the AMX feature.

Thanks @dimakuv , I got your point. In your proposal, LibOS will handle all the syscalls (from SGX or not), without any ocall or real syscall to host, only thing we need to do is enable AMX once in the `"loader"`, if so, I will implement it and create a new PR after some tests.

dimakuv:
> Thanks @dimakuv , I got your point. In your proposal, LibOS will handle all the syscalls (from SGX or not), without any ocall or real syscall to host, only thing we need to do is enable AMX once in the "loader", if so, I will implement it and create a new PR after some tests.

That's exactly correct. Thanks for doing this.

One immediate request, when you will be preparing the new PR: please add error-checking paths everywhere. For example, in your current PR, you added this snippet:
```c
    if (secs->attributes.xfrm & (1 << AMX_TILEDATA))
        DO_SYSCALL(arch_prctl, ARCH_REQ_XCOMP_PERM, AMX_TILEDATA);
```

I would like to see error checking on the result of the syscall here. I understand that you do *not* check for errors on purpose -- because you want to proceed even on older Linux systems that do not implement this syscall. But you need more precise error checking, like this:
```c
    if (secs->attributes.xfrm & (1 << AMX_TILEDATA)) {
        ret = DO_SYSCALL(arch_prctl, ARCH_REQ_XCOMP_PERM, AMX_TILEDATA);
        if (ret < 0 && ret != -EINVAL && ret != -EOPNOTSUPP) {
            log_error("arch_prctl(ARCH_REQ_XCOMP_PERM, AMX_TILEDATA) failed: %d", ret);
            return ret;
        }
        /* if ret is EINVAL or EOPNOTSUPP, then we are on older/patched Linux with no need for
         * arch_prctl-style AMX enablement; simply ignore this syscall */
    }
```

dongx1x:
New PR created #422 , close this PR

PR #420, opened 2022-02-21
[Docs] Improve and unify Doxygen comments, part 1

mkow:
## Description of the changes 

Please check if it looks good, if so then I'll update the rest of the docs (probably in parts).

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)

### DIFF
--- a/Documentation/devel/howto-doc.rst
+++ b/Documentation/devel/howto-doc.rst
@@ -150,19 +150,23 @@ Preferred Doxygen style
 
 #. Prefer Qt-style ``/*!`` and ``\param``:
 
+   .. Note that the snippet below is wrapped to 106 chars per line. This is
+      because it quotes C code (wrapped to 100), and the quote is itself
+      indented in reST.
+
    .. code-block:: c
 
       /*!
-       * \brief An example function
+       * \brief Sum two integers.
        *
-       * This function returns a number augmented by the Answer to the Ultimate
-       * Question of Life, the Universe, and Everything.
+       * \param first   First addend.
+       * \param second  Second addend.
        *
-       * \param n The number to be added
-       * \return A number 42 greater
+       * \returns Sum of the arguments. Sometimes a longer description is needed, then it should be
+       *          wrapped and aligned like this.
        */
-      int foo(int n) {
-          return n + 42;
+      int foo(int first, int second) {
+          return first + second;
       }
 
    ::
--- a/LibOS/shim/include/shim_checkpoint.h
+++ b/LibOS/shim/include/shim_checkpoint.h
@@ -349,14 +349,14 @@ typedef int (*migrate_func_t)(struct shim_cp_store*, struct shim_process*, struc
  *
  * Called in parent process during fork/clone.
  *
- * \param migrate_func          Migration function defined by the caller.
- * \param child_process         Struct bookkeeping the child process, added to the children list.
- * \param process_description   Struct describing the new process (child).
- * \param thread_description    Struct describing main thread of the child process.
+ * \param migrate_func         Migration function defined by the caller.
+ * \param child_process        Struct bookkeeping the child process, added to the children list.
+ * \param process_description  Struct describing the new process (child).
+ * \param thread_description   Struct describing main thread of the child process.
  *
  * The remaining arguments are passed into the migration function.
  *
- * \return  0 on success, negative POSIX error code on failure.
+ * \returns 0 on success, negative POSIX error code on failure.
  */
 int create_process_and_send_checkpoint(migrate_func_t migrate_func,
                                        struct shim_child_process* child_process,
@@ -368,9 +368,9 @@ int create_process_and_send_checkpoint(migrate_func_t migrate_func,
  *
  * Called in child process during initialization.
  *
- * \param[in] hdr  Checkpoint header already received from the parent.
+ * \param hdr  Checkpoint header already received from the parent.
  *
- * \return         0 on success, negative POSIX error code on failure.
+ * \returns 0 on success, negative POSIX error code on failure.
  */
 int receive_checkpoint_and_restore(struct checkpoint_hdr* hdr);
 
--- a/LibOS/shim/include/shim_fs.h
+++ b/LibOS/shim/include/shim_fs.h
@@ -206,9 +206,9 @@ typedef int (*readdir_callback_t)(const char* name, void* arg);
 
 struct shim_d_ops {
     /*
-     * \brief Look up a file
+     * \brief Look up a file.
      *
-     * \param dent dentry, not valid
+     * \param dent  Dentry, not valid.
      *
      * Queries the underlying filesystem for a path described by a dentry (`dent->name` and
      * `dent->parent`). On success, prepares the dentry for use by the filesystem. Always sets the
@@ -220,11 +220,11 @@ struct shim_d_ops {
     int (*lookup)(struct shim_dentry* dent);
 
     /*
-     * \brief Open an existing file
+     * \brief Open an existing file.
      *
-     * \param hdl a newly created handle
-     * \param dent dentry, valid and non-negative
-     * \param flags open flags, including access mode (O_RDONLY / O_WRONLY / O_RDWR)
+     * \param hdl    A newly created handle.
+     * \param dent   Dentry, valid and non-negative.
+     * \param flags  Open flags, including access mode (O_RDONLY / O_WRONLY / O_RDWR).
      *
      * Opens a file, and if successful, prepares the handle for use by the filesystem. Always sets
      * the `type` field.
@@ -235,12 +235,12 @@ struct shim_d_ops {
     int (*open)(struct shim_handle* hdl, struct shim_dentry* dent, int flags);
 
     /*
-     * \brief Create and open a new regular file
+     * \brief Create and open a new regular file.
      *
-     * \param hdl a newly created handle
-     * \param dent dentry, not valid (file to be created)
-     * \param flags open flags, including access mode (O_RDONLY / O_WRONLY / O_RDWR)
-     * \param perm permissions of the new file
+     * \param hdl    A newly created handle.
+     * \param dent   Dentry, not valid (file to be created).
+     * \param flags  Open flags, including access mode (O_RDONLY / O_WRONLY / O_RDWR).
+     * \param perm   Permissions of the new file.
      *
      * Creates and opens a new regular file at path described by `dent`. On success, prepares the
      * handle and the dentry for use by the filesystem (as in `open` and `lookup`).
@@ -251,10 +251,10 @@ struct shim_d_ops {
     int (*creat)(struct shim_handle* hdl, struct shim_dentry* dent, int flags, mode_t perm);
 
     /*
-     * \brief Create a directory
+     * \brief Create a directory.
      *
-     * \param dent dentry, not valid (directory to be created)
-     * \param perm permissions of the new directory
+     * \param dent  dentry, not valid (directory to be created).
+     * \param perm  Permissions of the new directory.
      *
      * Creates a new directory at path described by `dent`. On success, prepares the dentry for use
      * by the filesystem (as in `lookup`).
@@ -265,9 +265,9 @@ struct shim_d_ops {
     int (*mkdir)(struct shim_dentry* dent, mode_t perm);
 
     /*
-     * \brief Unlink a file
+     * \brief Unlink a file.
      *
-     * \param dent dentry, valid and non-negative, must have a parent
+     * \param dent  Dentry, valid and non-negative, must have a parent.
      *
      * Unlinks a file described by `dent`. Note that there might be handles for that file; if
      * possible, they should still work.
@@ -278,10 +278,10 @@ struct shim_d_ops {
     int (*unlink)(struct shim_dentry* dent);
 
     /*
-     * \brief Get file status
+     * \brief Get file status.
      *
-     * \param dent dentry, valid and non-negative
-     * \param buf status buffer to fill
+     * \param dent  Dentry, valid and non-negative.
+     * \param buf   Status buffer to fill.
      *
      * Fills `buf` with information about a file. Omits `st_ino` (which is later filled by the
      * caller).
@@ -291,10 +291,10 @@ struct shim_d_ops {
     int (*stat)(struct shim_dentry* dent, struct stat* buf);
 
     /*
-     * \brief Extract the target of a symbolic link
+     * \brief Extract the target of a symbolic link.
      *
-     * \param dent dentry, valid and non-negative, describing a symlink
-     * \param out_target on success, contains link target
+     * \param dent        Dentry, valid and non-negative, describing a symlink.
+     * \param out_target  On success, contains link target.
      *
      * Determines the target of a symbolic link, and sets `*out_target` to an allocated string.
      *
@@ -303,10 +303,10 @@ struct shim_d_ops {
     int (*follow_link)(struct shim_dentry* dent, char** out_target);
 
     /*
-     * \brief Change file permissions
+     * \brief Change file permissions.
      *
-     * \param dent dentry, valid and non-negative
-     * \param perm new permissions for the file
+     * \param dent  Dentry, valid and non-negative.
+     * \param perm  New permissions for the file.
      *
      * Changes the permissions for a file.
      *
@@ -315,10 +315,10 @@ struct shim_d_ops {
     int (*chmod)(struct shim_dentry* dent, mode_t perm);
 
     /*
-     * \brief Rename a file
+     * \brief Rename a file.
      *
-     * \param old source dentry, valid and non-negative
-     * \param new target dentry, valid, can be negative or non-negative
+     * \param old  Source dentry, valid and non-negative.
+     * \param new  Target dentry, valid, can be negative or non-negative.
      *
      * Moves a file described by `old` to the path described by `new`. Updates the fields of `new`
      * (same as `lookup`).
@@ -332,11 +332,11 @@ struct shim_d_ops {
     int (*rename)(struct shim_dentry* old, struct shim_dentry* new);
 
     /*!
-     * \brief List all files in the directory
+     * \brief List all files in the directory.
      *
-     * \param dent the dentry, must be valid, non-negative and describing a directory
-     * \param callback the callback to invoke on each file name
-     * \param arg argument to pass to the callback
+     * \param dent      The dentry, must be valid, non-negative and describing a directory.
+     * \param callback  The callback to invoke on each file name.
+     * \param arg       Argument to pass to the callback.
      *
      * Calls `callback(name, arg)` for all file names in the directory. `name` is not guaranteed to
      * be valid after callback returns, so the callback should copy it if necessary.
@@ -351,9 +351,9 @@ struct shim_d_ops {
     int (*readdir)(struct shim_dentry* dent, readdir_callback_t callback, void* arg);
 
     /*!
-     * \brief Deallocate inode data
+     * \brief Deallocate inode data.
      *
-     * \param inode the inode about to be freed
+     * \param inode  The inode about to be freed.
      *
      * Deallocates any custom data stored in the inode by filesystem. Called before deleting the
      * inode.
@@ -363,11 +363,11 @@ struct shim_d_ops {
     void (*idrop)(struct shim_inode* inode);
 
     /*!
-     * \brief Checkpoint inode data
+     * \brief Checkpoint inode data.
      *
-     * \param inode the inode to be checkpointed
-     * \param out_data on success, contains a newly allocated buffer with data
-     * \param out_size on success, contains data size
+     * \param inode     The inode to be checkpointed.
+     * \param out_data  On success, contains a newly allocated buffer with data.
+     * \param out_size  On success, contains data size.
      *
      * Prepares any custom data necessary to migrate the inode. Called when checkpointing the inode.
      *
@@ -376,10 +376,10 @@ struct shim_d_ops {
     int (*icheckpoint)(struct shim_inode* inode, void** out_data, size_t* out_size);
 
     /*!
-     * \brief Restore inode data from checkpoint
+     * \brief Restore inode data from checkpoint.
      *
-     * \param inode newly restored inode
-     * \param data checkpoint data (prepared by the `icheckpoint` callback)
+     * \param inode  Newly restored inode.
+     * \param data   Checkpoint data (prepared by the `icheckpoint` callback).
      *
      * Restores custom state of the inode. Called when restoring the inode from checkpoint, after
      * all other fields are set.
@@ -450,11 +450,11 @@ int init_mount(void);
 /* file system operations */
 
 /*!
- * \brief Mount a new filesystem
+ * \brief Mount a new filesystem.
  *
- * \param type Filesystem type (currently defined in `mountable_fs` in `shim_fs.c`)
- * \param uri PAL URI to mount, or NULL if not applicable
- * \param mount_path Path to the mountpoint
+ * \param type        Filesystem type (currently defined in `mountable_fs` in `shim_fs.c`).
+ * \param uri         PAL URI to mount, or NULL if not applicable.
+ * \param mount_path  Path to the mountpoint.
  *
  * Creates a new `shim_mount` structure (mounted filesystem) and attaches to the dentry under
  * `mount_path`. That means (assuming the dentry is called `mount_point`):
@@ -489,9 +489,9 @@ int init_dcache(void);
 extern struct shim_lock g_dcache_lock;
 
 /*!
- * \brief Dump dentry cache
+ * \brief Dump dentry cache.
  *
- * \param dent the starting dentry, or NULL (will default to dentry root)
+ * \param dent  The starting dentry, or NULL (will default to dentry root).
  *
  * Dumps the dentry cache using `log_always`, starting from the provided dentry. Intended for
  * debugging the filesystem - just add it manually to the code.
@@ -499,10 +499,10 @@ extern struct shim_lock g_dcache_lock;
 void dump_dcache(struct shim_dentry* dent);
 
 /*!
- * \brief Check file permissions, similar to Unix access
+ * \brief Check file permissions, similar to Unix access.
  *
- * \param dentry the dentry to check
- * \param mask mask, same as for Unix access
+ * \param dentry  The dentry to check.
+ * \param mask    Mask, same as for Unix access.
  *
  * Checks permissions for a dentry. Because Gramine currently has no notion of users, this will
  * always use the "user" part of file mode.
@@ -533,13 +533,13 @@ int check_permissions(struct shim_dentry* dent, mode_t mask);
 #define MAX_LINK_DEPTH 8
 
 /*!
- * \brief Look up a path, retrieving a dentry
+ * \brief Look up a path, retrieving a dentry.
  *
- * \param start the start dentry for relative paths, or NULL (in which case it will default to
- * process' cwd)
- * \param path the path to look up
- * \param flags lookup flags (see description below)
- * \param[out] found pointer to retrieved dentry
+ * \param      start  The start dentry for relative paths, or NULL (in which case it will default
+ *                    to process' cwd).
+ * \param      path   The path to look up.
+ * \param      flags  Lookup flags (see description below).
+ * \param[out] found  Pointer to retrieved dentry.
  *
  * The caller should hold `g_dcache_lock`. If you do not already hold `g_dcache_lock`, use
  * `path_lookupat` instead.
@@ -591,15 +591,15 @@ int path_lookupat(struct shim_dentry* start, const char* path, int flags,
 int get_dirfd_dentry(int dirfd, struct shim_dentry** dir);
 
 /*!
- * \brief Open a file under a given path, similar to Unix open
+ * \brief Open a file under a given path, similar to Unix open.
  *
- * \param hdl handle to associate with dentry, can be NULL
- * \param start the start dentry for relative paths, or NULL (in which case it will default to
- * process' cwd)
- * \param path the path to look up
- * \param flags Unix open flags (see below)
- * \param mode Unix file mode, used when creating a new file/directory
- * \param[out] found pointer to retrieved dentry, can be NULL
+ * \param      hdl    Handle to associate with dentry, can be NULL.
+ * \param      start  The start dentry for relative paths, or NULL (in which case it will default
+ *                    to process' cwd).
+ * \param      path   The path to look up.
+ * \param      flags  Unix open flags (see below).
+ * \param      mode   Unix file mode, used when creating a new file/directory.
+ * \param[out] found  Pointer to retrieved dentry, can be NULL.
  *
  * If `hdl` is provided, on success it will be associated with the dentry. Otherwise, the file will
  * just be retrieved or created.
@@ -633,11 +633,11 @@ int open_namei(struct shim_handle* hdl, struct shim_dentry* start, const char* p
                int mode, struct shim_dentry** found);
 
 /*!
- * \brief Open an already retrieved dentry, and associate a handle with it
+ * \brief Open an already retrieved dentry, and associates a handle with it.
  *
- * \param hdl handle to associate with dentry
- * \param dent the dentry to open
- * \param flags Unix open flags
+ * \param hdl    Handle to associate with dentry.
+ * \param dent   The dentry to open.
+ * \param flags  Unix open flags.
  *
  * The dentry has to already correspond to a file (i.e. has to be valid and non-negative). The
  * caller has to hold `g_dcache_lock`.
@@ -648,9 +648,9 @@ int open_namei(struct shim_handle* hdl, struct shim_dentry* start, const char* p
 int dentry_open(struct shim_handle* hdl, struct shim_dentry* dent, int flags);
 
 /*!
- * \brief Populate a directory handle with current dentries
+ * \brief Populate a directory handle with current dentries.
  *
- * \param hdl a directory handle
+ * \param hdl  A directory handle.
  *
  * This function populates the `hdl->dir_info` structure with current dentries in a directory, so
  * that the directory can be listed using `getdents/getdents64` syscalls.
@@ -663,9 +663,9 @@ int dentry_open(struct shim_handle* hdl, struct shim_dentry* dent, int flags);
 int populate_directory_handle(struct shim_handle* hdl);
 
 /*!
- * \brief Clear dentries from a directory handle
+ * \brief Clear dentries from a directory handle.
  *
- * \param hdl a directory handle
+ * \param hdl  A directory handle.
  *
  * This function discards an array of dentries previously prepared by `populate_directory_handle`.
  *
@@ -680,9 +680,9 @@ void get_dentry(struct shim_dentry* dent);
 void put_dentry(struct shim_dentry* dent);
 
 /*!
- * \brief Reset dentry state related to a file
+ * \brief Reset dentry state related to a file.
  *
- * \param  dent  the dentry (should be either invalid or negative)
+ * \param dent  The dentry (should be either invalid or negative).
  *
  * Resets the following dentry fields: `state`, `fs`, `type`, `perm`, `inode`. Ensures that there is
  * no leftover state from a file previously associated with the dentry, and the dentry can be used
@@ -699,11 +699,11 @@ void put_dentry(struct shim_dentry* dent);
 void reset_dentry(struct shim_dentry* dent);
 
 /*!
- * \brief Get the dentry one level up
+ * \brief Get the dentry one level up.
  *
- * \param dent the dentry
+ * \param dent  The dentry.
  *
- * \return the dentry one level up, or NULL if one does not exist
+ * \returns The dentry one level up, or NULL if one does not exist.
  *
  * Computes the dentry pointed to by ".." from the current one, unless the current dentry is at
  * global root. Unlike the `parent` field, this traverses mounted filesystems (i.e. works also for a
@@ -712,9 +712,9 @@ void reset_dentry(struct shim_dentry* dent);
 struct shim_dentry* dentry_up(struct shim_dentry* dent);
 
 /*!
- * \brief Garbage-collect a dentry, if possible
+ * \brief Garbage-collect a dentry, if possible.
  *
- * \param dentry the dentry (has to have a parent)
+ * \param dentry  The dentry (has to have a parent).
  *
  * This function checks if a dentry is unused, and deletes it if that's true. The caller must hold
  * `g_dcache_lock`.
@@ -740,13 +740,13 @@ struct shim_dentry* dentry_up(struct shim_dentry* dent);
 void dentry_gc(struct shim_dentry* dent);
 
 /*!
- * \brief Compute an absolute path for dentry, allocating memory for it
+ * \brief Compute an absolute path for dentry, allocating memory for it.
  *
- * \param dent the dentry
- * \param[out] path will be set to computed path
- * \param[out] size if not NULL, will be set to path size, including null terminator
+ * \param      dent  The dentry.
+ * \param[out] path  Will be set to computed path.
+ * \param[out] size  If not NULL, will be set to path size, including null terminator.
  *
- * \return 0 on success, negative error code otherwise
+ * \returns 0 on success, negative error code otherwise.
  *
  * This function computes an absolute path for dentry, allocating a new buffer for it. The path
  * should later be freed using `free`.
@@ -760,13 +760,13 @@ void dentry_gc(struct shim_dentry* dent);
 int dentry_abs_path(struct shim_dentry* dent, char** path, size_t* size);
 
 /*!
- * \brief Compute a relative path for dentry, allocating memory for it
+ * \brief Compute a relative path for dentry, allocating memory for it.
  *
- * \param dent the dentry
- * \param[out] path will be set to computed path
- * \param[out] size if not NULL, will be set to path size, including null terminator
+ * \param      dent  The dentry.
+ * \param[out] path  Will be set to computed path.
+ * \param[out] size  If not NULL, will be set to path size, including null terminator.
  *
- * \return 0 on success, negative error code otherwise
+ * \returns 0 on success, negative error code otherwise.
  *
  * This function computes a relative path for dentry, allocating a new buffer for it. The path
  * should later be freed using `free`.
@@ -782,14 +782,14 @@ int dentry_rel_path(struct shim_dentry* dent, char** path, size_t* size);
 ino_t dentry_ino(struct shim_dentry* dent);
 
 /*!
- * \brief Allocate and initialize a new dentry
+ * \brief Allocate and initializes a new dentry.
  *
- * \param mount the mount the dentry is under
- * \param parent the parent node, or NULL if this is supposed to be the mount root
- * \param name name of the new dentry
- * \param name_len length of the name
+ * \param mount     The mount the dentry is under.
+ * \param parent    The parent node, or NULL if this is supposed to be the mount root.
+ * \param name      Name of the new dentry.
+ * \param name_len  Length of the name.
  *
- * \return the new dentry, or NULL in case of allocation failure
+ * \returns The new dentry, or NULL in case of allocation failure.
  *
  * The caller should hold `g_dcache_lock`.
  *
@@ -806,13 +806,13 @@ struct shim_dentry* get_new_dentry(struct shim_mount* mount, struct shim_dentry*
                                    const char* name, size_t name_len);
 
 /*!
- * \brief Search for a child of a dentry with a given name
+ * \brief Search for a child of a dentry with a given name.
  *
- * \param parent the dentry to search under
- * \param name name of searched dentry
- * \param name_len length of the name
+ * \param parent    The dentry to search under.
+ * \param name      Name of searched dentry.
+ * \param name_len  Length of the name.
  *
- * \return the dentry, or NULL if not found
+ * \returns The dentry, or NULL if not found.
  *
  * The caller should hold `g_dcache_lock`.
  *
@@ -830,11 +830,11 @@ bool dentry_is_ancestor(struct shim_dentry* anc, struct shim_dentry* dent);
  * under space pressure, although for a single app, this may be over-kill. */
 
 /*!
- * \brief Allocate and initialize a new inode
+ * \brief Allocate and initialize a new inode.
  *
- * \param mount the mount the inode is under
- * \param type inode type (S_IFREG, S_IFDIR, etc.)
- * \param perm inode permissions (PERM_rwxrwxrwx, etc.)
+ * \param mount  The mount the inode is under.
+ * \param type   Inode type (S_IFREG, S_IFDIR, etc.).
+ * \param perm   Inode permissions (PERM_rwxrwxrwx, etc.).
  */
 struct shim_inode* get_new_inode(struct shim_mount* mount, mode_t type, mode_t perm);
 
@@ -876,13 +876,13 @@ extern struct shim_fs synthetic_builtin_fs;
 struct shim_fs* find_fs(const char* name);
 
 /*!
- * \brief Compute file position for `seek`
+ * \brief Compute file position for `seek`.
  *
- * \param pos current file position (non-negative)
- * \param size file size (non-negative)
- * \param offset desired offset
- * \param origin `seek` origin parameter (SEEK_SET, SEEK_CUR, SEEK_END)
- * \param[out] out_pos on success, contains new file position
+ * \param      pos      Current file position (non-negative).
+ * \param      size     File size (non-negative).
+ * \param      offset   Desired offset.
+ * \param      origin   `seek` origin parameter (SEEK_SET, SEEK_CUR, SEEK_END).
+ * \param[out] out_pos  On success, contains new file position.
  *
  * Computes new file position according to `seek` semantics. The new position will be non-negative,
  * although it can be larger than file size.
--- a/Pal/src/host/Linux-SGX/enclave_ocalls.c
+++ b/Pal/src/host/Linux-SGX/enclave_ocalls.c
@@ -81,16 +81,16 @@ static long sgx_exitless_ocall(uint64_t code, void* ms) {
     }
 
     /* wait till request processing is finished; try spinlock first */
-    int timedout = spinlock_lock_timeout(&req->lock, RPC_SPINLOCK_TIMEOUT);
+    bool locked = spinlock_lock_timeout(&req->lock, RPC_SPINLOCK_TIMEOUT);
 
     /* at this point:
      * - either RPC thread is done with OCALL and released the request's spinlock,
      *   and our enclave thread grabbed lock but it doesn't matter at this point
-     *   (OCALL is done, timedout = 0, no need to wait on futex)
+     *   (OCALL is done, locked == true, no need to wait on futex)
      * - or OCALL is still pending and the request is still blocked on spinlock
-     *   (OCALL is not done, timedout != 0, let's wait on futex) */
+     *   (OCALL is not done, locked == false, let's wait on futex) */
 
-    if (timedout) {
+    if (!locked) {
         /* OCALL takes a lot of time, so fallback to waiting on a futex; at this point we exit
          * enclave to perform syscall; this code is based on Mutex 2 from Futexes are Tricky */
         uint32_t c = SPINLOCK_UNLOCKED;
--- a/common/include/api.h
+++ b/common/include/api.h
@@ -201,21 +201,18 @@ long strtol(const char* s, char** endptr, int base);
 long long strtoll(const char* s, char** endptr, int base);
 
 /*!
- * \brief Convert a string to number
+ * \brief Convert a string to number.
  *
- * \param str the string
- * \param base digit base, between 2 and 36
- * \param[out] out_value on success, set to the parsed number
- * \param[out] out_end on success, set to the rest of string
+ * \param      str        Input string.
+ * \param      base       Digit base, between 2 and 36.
+ * \param[out] out_value  On success, set to the parsed number.
+ * \param[out] out_end    On success, set to the rest of string.
  *
- * \return 0 on success, -1 on failure
+ * \returns 0 on success, negative on failure.
  *
  * Parses a number from the beginning of a string. The number should be non-empty, consist of digits
  * only (no `+`/`-` signs), and not overflow the `unsigned long` type. For base 16, the "0x" prefix
  * is allowed but not required.
- *
- * On success, returns 0, sets `*out_value` to the value of the number, and `*out_end` to the first
- * byte after the number.
  */
 int str_to_ulong(const char* str, unsigned int base, unsigned long* out_value,
                  const char** out_end);
@@ -368,12 +365,11 @@ int get_base_name(const char* path, char* buf, size_t* inout_size);
 /*!
  * \brief Parse a size (number with optional "G"/"M"/"K" suffix) into an uint64_t.
  *
- * \param str A string containing a non-negative number. The string may end with "G"/"g" suffix
- *            denoting value in GBs, "M"/"m" for MBs, or "K"/"k" for KBs.
- * \param[out] out_val Parsed size (in bytes).
+ * \param      str      A string containing a non-negative, decimal number. The string may end with
+ *                      "G"/"g" suffix denoting value in GBs, "M"/"m" for MBs, or "K"/"k" for KBs.
+ * \param[out] out_val  Parsed size (in bytes).
  *
- * The number should be decimal. Returns -1 if string cannot be parsed into a size
- * (e.g., suffix is wrong).
+ * \returns 0 on success, negative if string cannot be parsed into a size (e.g., suffix is wrong).
  */
 int parse_size_str(const char* str, uint64_t* out_val);
 
--- a/common/include/spinlock.h
+++ b/common/include/spinlock.h
@@ -80,19 +80,6 @@ static inline void spinlock_init(spinlock_t* lock) {
 }
 
 /*!
- * \brief Try to acquire spinlock.
- *
- * \return  0 if acquiring the lock succeeded, 1 if it was already taken.
- */
-static inline int spinlock_trylock(spinlock_t* lock) {
-    if (__atomic_exchange_n(&lock->lock, SPINLOCK_LOCKED, __ATOMIC_ACQUIRE) == SPINLOCK_UNLOCKED) {
-        debug_spinlock_take_ownership(lock);
-        return 0;
-    }
-    return 1;
-}
-
-/*!
  * \brief Acquire spinlock.
  */
 static inline void spinlock_lock(spinlock_t* lock) {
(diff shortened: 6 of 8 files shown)


dimakuv (changes requested):
> Reviewed 8 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @mkow)

___
*[common/include/toml_utils.h, line 33 at r1](https://reviewable.io/reviews/gramineproject/gramine/420#-MwREeqY5tax_ceVo98v:-MwREeqZCBWaeTn0oAHy:b-c917q7) ([raw file](https://github.com/gramineproject/gramine/blob/79e62300407fb1e9f595222bcd7b941cdd162bb6/common/include/toml_utils.h#L33)):*
> ```objc
> 
> /*!
>  * \brief  Find an integer key-value in TOML manifest.
> ```

`Finds`
___
*[common/include/toml_utils.h, line 46 at r1](https://reviewable.io/reviews/gramineproject/gramine/420#-MwREipY27diHARnB6jW:-MwREipY27diHARnB6jX:b-c917q7) ([raw file](https://github.com/gramineproject/gramine/blob/79e62300407fb1e9f595222bcd7b941cdd162bb6/common/include/toml_utils.h#L46)):*
> ```objc
> 
> /*!
>  * \brief  Find a string key-value in TOML manifest.
> ```

`Finds`
___
*[common/include/toml_utils.h, line 58 at r1](https://reviewable.io/reviews/gramineproject/gramine/420#-MwRElr_5TAeT-TssCD0:-MwRElr_5TAeT-TssCD1:b-c917q7) ([raw file](https://github.com/gramineproject/gramine/blob/79e62300407fb1e9f595222bcd7b941cdd162bb6/common/include/toml_utils.h#L58)):*
> ```objc
> 
> /*!
>  * \brief  Find a "size" string key-value in TOML manifest (parsed via `parse_size_str()`).
> ```

`Finds`
___
*[Documentation/devel/howto-doc.rst, line 159 at r1](https://reviewable.io/reviews/gramineproject/gramine/420#-MwREyYe0oRw0zo5VB73:-MwREyYe0oRw0zo5VB74:b-esxlw3) ([raw file](https://github.com/gramineproject/gramine/blob/23bc8ecfe82eef05608423746ad32227bf79913e/Documentation/devel/howto-doc.rst#L159)):*
> ```reStructuredText
>        *
>        * This function returns a number augmented by the Answer to the Ultimate
>        * Question of Life, the Universe, and Everything.
> ```

I'll miss this reference to Adams, but I see why you chose a different example.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: 7 of 8 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[common/include/toml_utils.h, line 33 at r1](https://reviewable.io/reviews/gramineproject/gramine/420#-MwREeqY5tax_ceVo98v:-MwRHCqjA0fULO2NdaL1:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/79e62300407fb1e9f595222bcd7b941cdd162bb6/common/include/toml_utils.h#L33)):*

`Finds`
</blockquote></details>

Done.
___
*[common/include/toml_utils.h, line 46 at r1](https://reviewable.io/reviews/gramineproject/gramine/420#-MwREipY27diHARnB6jW:-MwRHE777KcikVSuM_Qn:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/79e62300407fb1e9f595222bcd7b941cdd162bb6/common/include/toml_utils.h#L46)):*

`Finds`
</blockquote></details>

Done.
___
*[common/include/toml_utils.h, line 58 at r1](https://reviewable.io/reviews/gramineproject/gramine/420#-MwRElr_5TAeT-TssCD0:-MwRHEkz2HFbpxXdixUh:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/79e62300407fb1e9f595222bcd7b941cdd162bb6/common/include/toml_utils.h#L58)):*

`Finds`
</blockquote></details>

Done.
___
*[Documentation/devel/howto-doc.rst, line 159 at r1](https://reviewable.io/reviews/gramineproject/gramine/420#-MwREyYe0oRw0zo5VB73:-MwRHFi55B390ydvHHMx:b-opsqvf) ([raw file](https://github.com/gramineproject/gramine/blob/23bc8ecfe82eef05608423746ad32227bf79913e/Documentation/devel/howto-doc.rst#L159)):*

I'll miss this reference to Adams, but I see why you chose a different example.
</blockquote></details>

Yeah, unfortunately it wasn't really helpful/informative here.

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

pwmarcz (changes requested):
> Reviewed 1 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[common/include/api.h, line 211 at r2](https://reviewable.io/reviews/gramineproject/gramine/420#-MwaLdat8G0qGnZ7Z0w5:-MwaLdat8G0qGnZ7Z0w6:brt7fsv) ([raw file](https://github.com/gramineproject/gramine/blob/cd33828c895fee19a9445a480bcd5244da73e813/common/include/api.h#L211)):*
> ```objc
>  * \param[out]  out_end    On success, set to the rest of string.
>  *
>  * \returns  0 on success, negative on failure.
> ```

I would prefer -1, not "negative". Most of our functions return negative error codes (POSIX or PAL...) on failure, some don't distinguish between errors and return only -1, I think it's better to be explicit and say we return -1 here.
___
*[common/include/spinlock.h, line 10 at r2](https://reviewable.io/reviews/gramineproject/gramine/420#-MwaOpQM9UL4MaU6ctAY:-MwaOpQM9UL4MaU6ctAZ:b5n443i) ([raw file](https://github.com/gramineproject/gramine/blob/cd33828c895fee19a9445a480bcd5244da73e813/common/include/spinlock.h#L10)):*
> ```objc
> #define _SPINLOCK_H
> 
> #include <errno.h>
> ```

`<errno.h>` is from glibc (and defines the `errno` global/macro), I think in no-stdlib setting we should include `<asm/errno.h>`.
___
*[Documentation/devel/howto-doc.rst, line 162 at r2](https://reviewable.io/reviews/gramineproject/gramine/420#-MwaMV8LEe5ga1MbfLMA:-MwaMV8LEe5ga1MbfLMB:b-sozhf9) ([raw file](https://github.com/gramineproject/gramine/blob/cd33828c895fee19a9445a480bcd5244da73e813/Documentation/devel/howto-doc.rst#L162)):*
> ```reStructuredText
>        *
>        * \returns  Sum of the arguments. Sometimes a longer description is needed,
>        *           then is should be wrapped and aligned like this.
> ```

Typo: `is -> it`
___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 93 at r2](https://reviewable.io/reviews/gramineproject/gramine/420#-MwaNBKy40lBEhuonNtH:-MwaNBKy40lBEhuonNtI:b-lzzbs4) ([raw file](https://github.com/gramineproject/gramine/blob/cd33828c895fee19a9445a480bcd5244da73e813/Pal/src/host/Linux-SGX/enclave_ocalls.c#L93)):*
> ```C
>      *   (OCALL is not done, ret == -EAGAIN, let's wait on futex) */
> 
>     if (ret == -EAGAIN) {
> ```

I don't like this, it suggests there might be other errors that we will ignore. Either we keep open the possibility of other errors than `-EAGAIN`, and do something to handle them here, or (IMO better in this case) `spinlock_lock_timeout` should still return only 0 or -1.

(I don't like this `timedout`, but my preferred refactoring would be to keep the -1 return code, and check `if (ret < 0)` here).

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: 5 of 8 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @pwmarcz)

___
*[common/include/api.h, line 211 at r2](https://reviewable.io/reviews/gramineproject/gramine/420#-MwaLdat8G0qGnZ7Z0w5:-MwapDVu4z9WmomEEmok:b-u61wo2) ([raw file](https://github.com/gramineproject/gramine/blob/cd33828c895fee19a9445a480bcd5244da73e813/common/include/api.h#L211)):*

I would prefer -1, not "negative". Most of our functions return negative error codes (POSIX or PAL...) on failure, some don't distinguish between errors and return only -1, I think it's better to be explicit and say we return -1 here.
</blockquote></details>

I was mostly afraid that if we say in _some_ functions that -1 means an error, then it's quite easy to accidentally forward a retval from a subfunction and then do `if (ret == -1)` in another place, resulting in not noticing the failure. So, I'd prefer to have uniform "negative means failure" everywhere.

Ad. POSIX vs PAL vs random error codes: I plan to make them type-safe, I did some experiments and it seems to be possible in C with enough warnings enabled (see https://godbolt.org/z/zMvb3Mqar), so, soon there should be no risk of interpreting -1 as some enum.
___
*[common/include/spinlock.h, line 10 at r2](https://reviewable.io/reviews/gramineproject/gramine/420#-MwaOpQM9UL4MaU6ctAY:-MwaqMXB1Vu2g5fsf9Ge:b-dmz11w) ([raw file](https://github.com/gramineproject/gramine/blob/cd33828c895fee19a9445a480bcd5244da73e813/common/include/spinlock.h#L10)):*

`<errno.h>` is from glibc (and defines the `errno` global/macro), I think in no-stdlib setting we should include `<asm/errno.h>`.
</blockquote></details>

Done (removed because of the other discussion).
___
*[Documentation/devel/howto-doc.rst, line 162 at r2](https://reviewable.io/reviews/gramineproject/gramine/420#-MwaMV8LEe5ga1MbfLMA:-MwaqR9-BXT_AtHpaxCZ:b-ugcbwe) ([raw file](https://github.com/gramineproject/gramine/blob/cd33828c895fee19a9445a480bcd5244da73e813/Documentation/devel/howto-doc.rst#L162)):*

Typo: `is -> it`
</blockquote></details>

Heh, I read this 3 times and it still wasn't enough to spot this :)
___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 93 at r2](https://reviewable.io/reviews/gramineproject/gramine/420#-MwaNBKy40lBEhuonNtH:-MwaqoB-0XzMfve-7zV-:b-nec2kw) ([raw file](https://github.com/gramineproject/gramine/blob/cd33828c895fee19a9445a480bcd5244da73e813/Pal/src/host/Linux-SGX/enclave_ocalls.c#L93)):*

I don't like this, it suggests there might be other errors that we will ignore. Either we keep open the possibility of other errors than `-EAGAIN`, and do something to handle them here, or (IMO better in this case) `spinlock_lock_timeout` should still return only 0 or -1.

(I don't like this `timedout`, but my preferred refactoring would be to keep the -1 return code, and check `if (ret < 0)` here).
</blockquote></details>

Ok, makes sense.

pwmarcz (changes requested):
> Reviewed 3 of 8 files at r1, 1 of 1 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[common/include/api.h, line 211 at r2](https://reviewable.io/reviews/gramineproject/gramine/420#-MwaLdat8G0qGnZ7Z0w5:-MwfsBAl6fbX5vEBRH1Y:b-8e88b1) ([raw file](https://github.com/gramineproject/gramine/blob/cd33828c895fee19a9445a480bcd5244da73e813/common/include/api.h#L211)):*

I was mostly afraid that if we say in _some_ functions that -1 means an error, then it's quite easy to accidentally forward a retval from a subfunction and then do `if (ret == -1)` in another place, resulting in not noticing the failure. So, I'd prefer to have uniform "negative means failure" everywhere.

Ad. POSIX vs PAL vs random error codes: I plan to make them type-safe, I did some experiments and it seems to be possible in C with enough warnings enabled (see https://godbolt.org/z/zMvb3Mqar), so, soon there should be no risk of interpreting -1 as some enum.
</blockquote></details>

OK, I understand why we might want not to promise -1 exactly. But I'm still worried that somebody might forward this value (e.g. `return ret;`) thinking it's a POSIX error code.

So I would prefer something else than "negative": in some places we say "negative error code", so which sounds really similar. But I don't have a better suggestion, so I'm unblocking.
___
*[common/include/spinlock.h, line 85 at r3](https://reviewable.io/reviews/gramineproject/gramine/420#-MwftQxZAO0ZIfGCmYib:-MwftQxZAO0ZIfGCmYic:b-2hd6ip) ([raw file](https://github.com/gramineproject/gramine/blob/00ee8626b0350928bfc62b41387a6d7d3a8d0924/common/include/spinlock.h#L85)):*
> ```objc
>  * \brief Try to acquire spinlock.
>  *
>  * \returns  0 if acquiring the lock succeeded, 1 if it was already taken.
> ```

Can you change this function to return -1 for consistency with the one below?

(This is unused btw, maybe we don't need it?)

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: 7 of 8 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[common/include/spinlock.h, line 85 at r3](https://reviewable.io/reviews/gramineproject/gramine/420#-MwftQxZAO0ZIfGCmYib:-MwfxWJU8fDiL0N8qt2L:b-bopbiu) ([raw file](https://github.com/gramineproject/gramine/blob/00ee8626b0350928bfc62b41387a6d7d3a8d0924/common/include/spinlock.h#L85)):*

Can you change this function to return -1 for consistency with the one below?

(This is unused btw, maybe we don't need it?)
</blockquote></details>

Good catch!
Not sure whether to keep it or remove, it's a library and having such API probably makes sense overall...

pwmarcz (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

boryspoplawski (commented):
> Reviewed 3 of 8 files at r1, 2 of 3 files at r3, 1 of 1 files at r4, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiMnM9DQPYIL12OD1w:-MwiMnM9DQPYIL12OD1x:b-t7up1x) (no related file):*
I personally find two spaces after `\param` redundant (unlike after arg name, where it looks nice) - it does not improve readability and just takes space.

---
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiNM1oD59ROcmuGR-w:-MwiNM1oD59ROcmuGR-x:bpdvc1a) (no related file):*
I've put all comments as non-blocking, since it's late in the review process and already rebased.

---
___
*[common/include/api.h, line 211 at r2](https://reviewable.io/reviews/gramineproject/gramine/420#-MwaLdat8G0qGnZ7Z0w5:-MwiLPryEYJsECF_Yoj-:b8ya8nq) ([raw file](https://github.com/gramineproject/gramine/blob/cd33828c895fee19a9445a480bcd5244da73e813/common/include/api.h#L211)):*
> Ad. POSIX vs PAL vs random error codes: I plan to make them type-safe, I did some experiments and it seems to be possible in C with enough warnings enabled (see https://godbolt.org/z/zMvb3Mqar), so, soon there should be no risk of interpreting -1 as some enum.

It's not so easy. If you do `enum { A=1; B; }`, then we would have to return `-A` and we get no type checking. If you do `enum { A = -1;}` then we get an ugly convention like `return PAL_ERROR_AAAA;` and you don't even know if it's positive or negative (check mbedtls sources for how it looks, they use that convention).

> OK, I understand why we might want not to promise -1 exactly. But I'm still worried that some1body might forward this value (e.g. return ret;) thinking it's a POSIX error code.

Fortunately `1` is `EPERM` so nothing bad should happen (it's not like `EINTR` or something).
___
*[common/include/api.h, line 204 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiNrad7MP4AXJaz5uc:-MwiNrad7MP4AXJaz5ud:bf7m5a7) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/common/include/api.h#L204)):*
> ```objc
> 
> /*!
>  * \brief  Converts a string to number.
> ```

Why did you change all `\brief` to 3rd person? Old version matched our conventions (e.g. commit message, which describes commit content, just like this comment describes a function).
ditto for the whole PR
___
*[common/include/spinlock.h, line 85 at r3](https://reviewable.io/reviews/gramineproject/gramine/420#-MwftQxZAO0ZIfGCmYib:-MwiN93X3Ofv7RrjeKQP:bhulxdi) ([raw file](https://github.com/gramineproject/gramine/blob/00ee8626b0350928bfc62b41387a6d7d3a8d0924/common/include/spinlock.h#L85)):*

Good catch!
Not sure whether to keep it or remove, it's a library and having such API probably makes sense overall...
</blockquote></details>

I would vote for removing it. 
___
*[common/include/spinlock.h, line 126 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiOOMa1QZwISd0beWT:-MwiOOMa1QZwISd0beWU:b-4yh3ez) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/common/include/spinlock.h#L126)):*
> ```objc
>  * \param  iterations  Number of iterations (tries) after which this function times out.
>  *
>  * \returns  0 if acquiring the lock succeeded, negative if timed out.
> ```

Do we treat timeout as an error (since you've changed to a negative value)? I thought of it as just another possible outcome of this function, which is not an error, so `1` (positive value) made sense.

Update: after looking at usages - this should just return `bool`.
___
*[common/src/printf.c, line 59 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiQKmcCmGFdde3zqKg:-MwiQKmcCmGFdde3zqKh:bijxua7) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/common/src/printf.c#L59)):*
> ```C
>  * \brief  Core printf implementation.
>  *
>  * \param       write_callback  Function called on each generated chunk of data.
> ```

I think the old version (no capital letter and no dot at the end) looks better when rendered in docs.
___
*[common/src/printf.c, line 64 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiQAtS5AjaYFWqxrF8:-MwiQAtS5AjaYFWqxrF9:b-vhf770) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/common/src/printf.c#L64)):*
> ```C
>  * \param       ap              List of optional variadic arguments.
>  * \param[out]  out_size        Total size of written data (sum of sizes passed to all \p
>  *                              write_callback invocations).
> ```

I would move `\p` to the new line. It probably renders ok, but is weird to read.
___
*[Documentation/devel/howto-doc.rst, line 161 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiRGf-8H5c1v7Y_v2n:-MwiRGf09A0XZi4D_wCC:ba4l0qp) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/Documentation/devel/howto-doc.rst#L161)):*
> ```reStructuredText
>        * \param  second  Second addend.
>        *
>        * \returns  Sum of the arguments. Sometimes a longer description is needed,
> ```

Is it grammatically correct? I would change `Sometimes` -> `If`/`When`
___
*[Documentation/devel/howto-doc.rst, line 161 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiRWOH2klKlD9hHGDj:-MwiRWOH2klKlD9hHGDk:b3gz93t) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/Documentation/devel/howto-doc.rst#L161)):*
> ```reStructuredText
>        * \param  second  Second addend.
>        *
>        * \returns  Sum of the arguments. Sometimes a longer description is needed,
> ```

This is not correctly wrapped (line length).

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiMnM9DQPYIL12OD1w:-MwkOgOW64du8I87QmEL:bobkxun) (no related file):*

I personally find two spaces after `\param` redundant (unlike after arg name, where it looks nice) - it does not improve readability and just takes space.
</blockquote></details>

I think I asked for it at some point. It looks more readable to me, but I don't care too much.

But keep in mind  it should be consistent with spacing after `\returns` and `\brief`. This PR mostly does that, but I see at least of `\brief` with one space (in `spinlock.h`).

---
___
*[common/include/api.h, line 204 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiNrad7MP4AXJaz5uc:-MwkQbHu6t6sl_MYAfKK:b-43h6qb) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/common/include/api.h#L204)):*

Why did you change all `\brief` to 3rd person? Old version matched our conventions (e.g. commit message, which describes commit content, just like this comment describes a function).
ditto for the whole PR
</blockquote></details>

Actually, yes, +1 to that. The current style (imperative `\brief`, longer description in 3rd person) looks natural to me.

dimakuv (approved):
> Reviewed 2 of 3 files at r3, 1 of 1 files at r4, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

___
*[common/include/api.h, line 204 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiNrad7MP4AXJaz5uc:-MwkTlZv1cALQ_cpZ7B6:b-j11zlp) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/common/include/api.h#L204)):*

Actually, yes, +1 to that. The current style (imperative `\brief`, longer description in 3rd person) looks natural to me.
</blockquote></details>

I don't care, so whatever you decide on.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiMnM9DQPYIL12OD1w:-MwkbpSCDRcXPYsl_SPW:bhh95b4) (no related file):*

I think I asked for it at some point. It looks more readable to me, but I don't care too much.

But keep in mind  it should be consistent with spacing after `\returns` and `\brief`. This PR mostly does that, but I see at least of `\brief` with one space (in `spinlock.h`).
</blockquote></details>

I would remove it from `\ret` and `\brief` too - I believe the original intent was to make `\param arg something` less confusing, so that you don't mistake `arg` as part of the sentence

---

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: 6 of 8 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, @mkow, and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiMnM9DQPYIL12OD1w:-MwlEQAT9jzhH7Ajse0L:bq1n4o3) (no related file):*

I would remove it from `\ret` and `\brief` too - I believe the original intent was to make `\param arg something` less confusing, so that you don't mistake `arg` as part of the sentence
</blockquote></details>

I'm rather neutral on this, but it should be either a double space after all `\tags` or always single (as Borys proposed).

@dimakuv What's your preference?

---
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiNM1oD59ROcmuGR-w:-MwlEiBwBcO5oGltVj-R:bv8x6jz) (no related file):*

I've put all comments as non-blocking, since it's late in the review process and already rebased.
</blockquote></details>

No hurry, let's resolve them first.

---
___
*[common/include/api.h, line 211 at r2](https://reviewable.io/reviews/gramineproject/gramine/420#-MwaLdat8G0qGnZ7Z0w5:-MwlFcfTCUI9AvWX2jUX:b-rxlnz0) ([raw file](https://github.com/gramineproject/gramine/blob/cd33828c895fee19a9445a480bcd5244da73e813/common/include/api.h#L211)):*

> Ad. POSIX vs PAL vs random error codes: I plan to make them type-safe, I did some experiments and it seems to be possible in C with enough warnings enabled (see https://godbolt.org/z/zMvb3Mqar), so, soon there should be no risk of interpreting -1 as some enum.

It's not so easy. If you do `enum { A=1; B; }`, then we would have to return `-A` and we get no type checking. If you do `enum { A = -1;}` then we get an ugly convention like `return PAL_ERROR_AAAA;` and you don't even know if it's positive or negative (check mbedtls sources for how it looks, they use that convention).

> OK, I understand why we might want not to promise -1 exactly. But I'm still worried that some1body might forward this value (e.g. return ret;) thinking it's a POSIX error code.

Fortunately `1` is `EPERM` so nothing bad should happen (it's not like `EINTR` or something).
</blockquote></details>

Hmm, good point, not sure what to do with the negative values...
___
*[common/include/api.h, line 204 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiNrad7MP4AXJaz5uc:-MwlEod4Br4GvEmMjltt:bq1oxgo) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/common/include/api.h#L204)):*

I don't care, so whatever you decide on.
</blockquote></details>

I think for functions it feels more natural to use this form, because the comment describes what a function does "in general", without binding to a specific point in time. With commits it's a bit different, commit history is more like a list of changes/commands applied in order to the source.
___
*[common/include/spinlock.h, line 85 at r3](https://reviewable.io/reviews/gramineproject/gramine/420#-MwftQxZAO0ZIfGCmYib:-MwlFopTADPPHzyx_cwD:bl1h72v) ([raw file](https://github.com/gramineproject/gramine/blob/00ee8626b0350928bfc62b41387a6d7d3a8d0924/common/include/spinlock.h#L85)):*

I would vote for removing it. 
</blockquote></details>

Ok, done.
___
*[common/include/spinlock.h, line 126 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiOOMa1QZwISd0beWT:-MwlKd3xCoZzGjxlHeyp:bblcrog) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/common/include/spinlock.h#L126)):*

Do we treat timeout as an error (since you've changed to a negative value)? I thought of it as just another possible outcome of this function, which is not an error, so `1` (positive value) made sense.

Update: after looking at usages - this should just return `bool`.
</blockquote></details>

Yeah, I consider it to be an equivalent of `-EAGAIN` - function failed to do what was asked, and the reason for failure is a timeout.
Bool also sounds ok for me, but let's ask @pwmarcz what he thinks about it.
___
*[common/src/printf.c, line 59 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiQKmcCmGFdde3zqKg:-MwlLF-G6q_2j6aBlx_S:b-ws59ol) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/common/src/printf.c#L59)):*

I think the old version (no capital letter and no dot at the end) looks better when rendered in docs.
</blockquote></details>

I agree, but tbh I mostly read this directly in the code, where the capital letter help a lot visually.
@pwmarcz, @dimakuv: Any opinions on this?
___
*[common/src/printf.c, line 64 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiQAtS5AjaYFWqxrF8:-MwlMGuD40IeG8PZarND:bfbq99i) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/common/src/printf.c#L64)):*

I would move `\p` to the new line. It probably renders ok, but is weird to read.
</blockquote></details>

Good point, done.
___
*[Documentation/devel/howto-doc.rst, line 161 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiRGf-8H5c1v7Y_v2n:-MwlMQQL51Q3P593MMx3:bd9r2av) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/Documentation/devel/howto-doc.rst#L161)):*

Is it grammatically correct? I would change `Sometimes` -> `If`/`When`
</blockquote></details>

Hmm, I think it's correct, but maybe changing `then`-> `in such a case` would help?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: 6 of 8 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, @mkow, and @pwmarcz)

___
*[common/include/api.h, line 204 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiNrad7MP4AXJaz5uc:-MwlQDcj79Pp0KCn9INo:b-lsl9vn) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/common/include/api.h#L204)):*

I think for functions it feels more natural to use this form, because the comment describes what a function does "in general", without binding to a specific point in time. With commits it's a bit different, commit history is more like a list of changes/commands applied in order to the source.
</blockquote></details>

Looking for precedents:

* I see Doxygen examples with 3rd person
* Linux doesn't seem to use a consistent style (their guide says just to write a "brief description")
* Xen project seems to prefer the current style (brief in 2nd person, description in 3rd person)
* Linux and BSD man pages use the current style (brief: `close - close a file descriptor`, then later `close() closes a file descriptor, so that...`)
* Rust uses 3rd person only (e.g. `pub fn clear(...) - Truncates this String, removing all contents.`), but doesn't seem to have any "brief" section at all, just a paragraph of description)
* Ruby also uses 3rd person only (e.g. `clear -> self - Remove the contents of self`), but also doesn't have a brief section
* Python ([https://www.python.org/dev/peps/pep-0257/](PEP 257)) doesn't talk about the grammatical form, but has examples in the imperative (e.g. `Return the pathname...`, `Do X and return a list.`)
* Javadoc style guide [explicitly says to use 3rd person, not imperative](https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html), but AFAIK also doesn't describe a separate "brief" section

To summarize: I have some examples where a "brief" section uses 2nd person imperative, and description uses 3rd person (Xen, Linux and BSD man pages, Python). Using 3rd person for the "brief" section seems pretty rare (although there are projects that do not have this line at all).

Most relevant/prominent seems to be the example of man pages.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @pwmarcz)

___
*[common/include/api.h, line 204 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiNrad7MP4AXJaz5uc:-MwlPxV539VXmEaO1eLv:bgz4f1n) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/common/include/api.h#L204)):*
> commit history is more like a list of changes/commands applied in order to the source.

Agreed, and commit message is describing these changes, or putting in your words: describes what a commit does "in general".
___
*[common/include/spinlock.h, line 126 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiOOMa1QZwISd0beWT:-MwlRBbO1uRD2RFmSxcf:bxevakx) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/common/include/spinlock.h#L126)):*

Yeah, I consider it to be an equivalent of `-EAGAIN` - function failed to do what was asked, and the reason for failure is a timeout.
Bool also sounds ok for me, but let's ask @pwmarcz what he thinks about it.
</blockquote></details>

In this case it's not really that function failed, we literally asked it to spin for a given timeout...
___
*[common/src/printf.c, line 59 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiQKmcCmGFdde3zqKg:-MwlV_qoDiBKpImx3sJB:bc0c8tq) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/common/src/printf.c#L59)):*

I agree, but tbh I mostly read this directly in the code, where the capital letter help a lot visually.
@pwmarcz, @dimakuv: Any opinions on this?
</blockquote></details>

With double space I don't mind this even in the source

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiMnM9DQPYIL12OD1w:-MwlY0n47Rqeihklf-sr:b2mc9kh) (no related file):*

I'm rather neutral on this, but it should be either a double space after all `\tags` or always single (as Borys proposed).

@dimakuv What's your preference?
</blockquote></details>

I liked a single space for `\ret` and `\brief`. My opinion is the same as Borys's.

---
___
*[common/src/printf.c, line 59 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiQKmcCmGFdde3zqKg:-MwlYPGTDBz5JNAqzgjP:bed91xw) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/common/src/printf.c#L59)):*

With double space I don't mind this even in the source
</blockquote></details>

I like the capital letter.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, and @pwmarcz)

___
*[common/include/spinlock.h, line 126 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiOOMa1QZwISd0beWT:-MwlYOGg0AgfGFsx8jII:buqzh44) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/common/include/spinlock.h#L126)):*

In this case it's not really that function failed, we literally asked it to spin for a given timeout...
</blockquote></details>

I like the idea of using bool (true for success, right?). In other instances where we return -1 (mostly parsing) the failure is more obvious and unrecoverable.

As I said before, using `-EAGAIN` here kind of suggests that this function might fail in other ways which we should handle.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[common/src/printf.c, line 59 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiQKmcCmGFdde3zqKg:-Mwl_wO33DyFo851PULJ:b7da3fv) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/common/src/printf.c#L59)):*

I like the capital letter.
</blockquote></details>

With the capital letter, it's easier to write multi-sentence parameter descriptions (but do we ever do that?) But I don't mind either way.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: 0 of 8 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, @mkow, and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiMnM9DQPYIL12OD1w:-MwlzXfT2V7xq8oma0qY:bo5krl4) (no related file):*

I liked a single space for `\ret` and `\brief`. My opinion is the same as Borys's.
</blockquote></details>

Ok, then I'm changing it to a single space. Please check the new version and see if it looks better.

---
___
*[common/include/api.h, line 204 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiNrad7MP4AXJaz5uc:-Mwln3K26OZMWCgiQBUB:b18rspe) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/common/include/api.h#L204)):*
> > commit history is more like a list of changes/commands applied in order to the source.
> 
> Agreed, and commit message is describing these changes, or putting in your words: describes what a commit does "in general".

Commit message body: yes (but for this I think we don't have a uniform convention). For one-liners it's IMO different, you usually view them as a log of changes done to the codebase, so we use imperative mood.

Anyways, this seem to be mostly a matter of taste, so let's just vote who likes which style more and use the most preferred one. I prefer 3rd person and Borys prefers imperative mood. @dimakuv, @pwmarcz: Please vote :)
___
*[common/include/spinlock.h, line 126 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiOOMa1QZwISd0beWT:-MwltCZS-tonr56Mc-d5:b-5n3e9t) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/common/include/spinlock.h#L126)):*
> true for success, right?

Yup, definitely this way. Done.

> Yeah, and I wouldn't even call it a fail, it's a legitimate and expected timeout.

I consider this as a failure. IMO if it could also fail for other reasons (e.g. OOM), then we'd just return `-EAGAIN` and `-ENOMEM`.
___
*[common/src/printf.c, line 59 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiQKmcCmGFdde3zqKg:-MwlyXNBA_gKbEWvFoSm:bwfc97z) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/common/src/printf.c#L59)):*

I think we should not write multi-sentence descriptions here, they are supposed to be brief
</blockquote></details>

Sometimes we do when function semantic is non-trivial, see e.g. `parse_size_str()` in this very PR.
___
*[Documentation/devel/howto-doc.rst, line 161 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiRWOH2klKlD9hHGDj:-MwlzE2UElMOysvcxGZn:b-4uytkp) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/Documentation/devel/howto-doc.rst#L161)):*

This is not correctly wrapped (line length).
</blockquote></details>

Will fix depending on the resolution of the other discussion on this line.

boryspoplawski (commented):
> Reviewed 8 of 8 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[common/include/api.h, line 204 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiNrad7MP4AXJaz5uc:-MwmRPf1-tLvLe1CXQ-d:b-69r4dq) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/common/include/api.h#L204)):*

> > commit history is more like a list of changes/commands applied in order to the source.
> 
> Agreed, and commit message is describing these changes, or putting in your words: describes what a commit does "in general".

Commit message body: yes (but for this I think we don't have a uniform convention). For one-liners it's IMO different, you usually view them as a log of changes done to the codebase, so we use imperative mood.

Anyways, this seem to be mostly a matter of taste, so let's just vote who likes which style more and use the most preferred one. I prefer 3rd person and Borys prefers imperative mood. @dimakuv, @pwmarcz: Please vote :)
</blockquote></details>

From prev comments seems like @pwmarcz prefers imperative and @dimakuv doesn't care.
___
*[common/src/printf.c, line 59 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiQKmcCmGFdde3zqKg:-MwmSzEz-kTaf29LVc5i:b-5mrvh6) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/common/src/printf.c#L59)):*

Sometimes we do when function semantic is non-trivial, see e.g. `parse_size_str()` in this very PR.
</blockquote></details>

One could argue that the 2nd sentence should be moved to the extended description below.
___
*[Documentation/devel/howto-doc.rst, line 161 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiRGf-8H5c1v7Y_v2n:-MwmTHTc8-qo9JUh8UBa:browqqx) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/Documentation/devel/howto-doc.rst#L161)):*

Hmm, I think it's correct, but maybe changing `then`-> `in such a case` would help?
</blockquote></details>

Idk, it just reads weird to me

pwmarcz (commented):
> Reviewed 8 of 8 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[common/include/api.h, line 204 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiNrad7MP4AXJaz5uc:-MwpJyqy5dEhzuPbh5mK:b-fcopys) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/common/include/api.h#L204)):*

From prev comments seems like @pwmarcz prefers imperative and @dimakuv doesn't care.
</blockquote></details>

Yes, I prefer imperative for brief and 3rd person for description. Both as a matter of taste, and because it seems to be more common (when a project has a "brief" line, they generally use imperative for it).

dimakuv (dismissed):
> Reviewed 8 of 8 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[common/include/api.h, line 204 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiNrad7MP4AXJaz5uc:-MwzpQ6YDG48sCwb5_DU:bsjp30x) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/common/include/api.h#L204)):*

Yes, I prefer imperative for brief and 3rd person for description. Both as a matter of taste, and because it seems to be more common (when a project has a "brief" line, they generally use imperative for it).
</blockquote></details>

I don't care.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: 2 of 8 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, @mkow, and @pwmarcz)

___
*[common/include/api.h, line 204 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiNrad7MP4AXJaz5uc:-Mx6U7kT3WFW8m1KzHCX:bd0ucch) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/common/include/api.h#L204)):*

I don't care.
</blockquote></details>

Ok, changed to imperative then.
___
*[common/src/printf.c, line 59 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiQKmcCmGFdde3zqKg:-Mx5IJVOBiddxaog_ww3:bdo2q0z) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/common/src/printf.c#L59)):*

One could argue that the 2nd sentence should be moved to the extended description below.
</blockquote></details>

I prefer to keep it at the arg if it directly explain this very argument.

boryspoplawski (changes requested):
> Reviewed 6 of 6 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[Documentation/devel/howto-doc.rst, line 161 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiRWOH2klKlD9hHGDj:-MwmTD2l8nFhuwkgz3Td:bjto253) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/Documentation/devel/howto-doc.rst#L161)):*

Will fix depending on the resolution of the other discussion on this line.
</blockquote></details>

Still incorrect?

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[Documentation/devel/howto-doc.rst, line 161 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiRWOH2klKlD9hHGDj:-Mx6pnAr7xGf3i4D_Feo:b-y0skhx) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/Documentation/devel/howto-doc.rst#L161)):*

Still incorrect?
</blockquote></details>

Then what's "correct"? You mean I should wrap like it was a source file and ` * ` was at position 0 in it?

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[Documentation/devel/howto-doc.rst, line 161 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiRWOH2klKlD9hHGDj:-Mx6sm3GAHS1OwybRlXk:b-qjefn) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/Documentation/devel/howto-doc.rst#L161)):*

Then what's "correct"? You mean I should wrap like it was a source file and ` * ` was at position 0 in it?
</blockquote></details>

I meant `(line length)`- to wrap it to 100 chars - it's in `.rst` but it's an example C source.
Idk about position of `*`, but probably yes, assume it's `0`, so like make line limit `108`

pwmarcz (commented):
> Reviewed 5 of 6 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[Documentation/devel/howto-doc.rst, line 161 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiRWOH2klKlD9hHGDj:-Mx8pjSv2FmrmaL2JchY:bswqjwn) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/Documentation/devel/howto-doc.rst#L161)):*

The line limit here is not `108`, and I am very much opposed to this suggestion (if I understood Borys correctly). How am I supposed to set up my Vim to automatically figure out that some lines are 100 chars and some 108?

The current version has a limit of 100 chars, and I prefer to keep like this.
</blockquote></details>

This is C source in our coding style (column limit 100) embedded inside reST (column limit 80), moved 6 characters to the right (notice the position of `/*` and `int`). I think two choices are appropriate:

1. Keep this wrapped at 80 characters, so that it's easy to edit in this file.
2. Keep this wrapped at 106 characters (not 108), so that the code block, which we display to the user, is wrapped to 100 characters.

I think the second choice is more "correct", since this is supposed to show our coding style. But it's also pretty inconvenient to edit, and maybe we need a (reST) comment explaining the column limit:

```
#. Prefer Qt-style ``/*!`` and ``\param``:

   ..
      Please keep the following code block wrapped at 100 characters (plus left margin):
      ....................................................................................................|

   .. code-block:: c

      /*!
       * \brief Sum two integers.

```

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[Documentation/devel/howto-doc.rst, line 161 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiRWOH2klKlD9hHGDj:-Mx8tgVXBElpmgN8ehlM:b-e669uu) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/Documentation/devel/howto-doc.rst#L161)):*

This is C source in our coding style (column limit 100) embedded inside reST (column limit 80), moved 6 characters to the right (notice the position of `/*` and `int`). I think two choices are appropriate:

1. Keep this wrapped at 80 characters, so that it's easy to edit in this file.
2. Keep this wrapped at 106 characters (not 108), so that the code block, which we display to the user, is wrapped to 100 characters.

I think the second choice is more "correct", since this is supposed to show our coding style. But it's also pretty inconvenient to edit, and maybe we need a (reST) comment explaining the column limit:

```
#. Prefer Qt-style ``/*!`` and ``\param``:

   ..
      Please keep the following code block wrapped at 100 characters (plus left margin):
      ....................................................................................................|

   .. code-block:: c

      /*!
       * \brief Sum two integers.

```
</blockquote></details>

Why are we doing this? What will it solve? How do we verify this limit during code reviews?

I have a strong feeling we're doing some bike shedding here. Why not just go with the easiest option (a single char limit for each file type).

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[Documentation/devel/howto-doc.rst, line 161 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiRWOH2klKlD9hHGDj:-Mx8u1Qi0oeZlrpd6UYR:b-gers2u) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/Documentation/devel/howto-doc.rst#L161)):*
> Why are we doing this? What will it solve?

Make the example consistent with other recommendations. We ask contributors to wrap at 100 columns, they should see code wrapped at 100 columns here. (They will probably read the HTML, so they don't care about how the reST source looks like).

>How do we verify this limit during code reviews?

To be clear, I'm not proposing to make a 106 column rule in such cases. I think we can make an exception for this code snippet (because it's supposed to show wrapping) and wrap everything else in our documentation at 80 columns. Hopefully, there won't be many snippets that even need wrapping.

But wrapping this one at 80 columns is fine for me, too.

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

___
*[Documentation/devel/howto-doc.rst, line 161 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiRWOH2klKlD9hHGDj:-Mx8zp-13Frtq8Aiex-q:b1muxfb) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/Documentation/devel/howto-doc.rst#L161)):*

> Why are we doing this? What will it solve?

Make the example consistent with other recommendations. We ask contributors to wrap at 100 columns, they should see code wrapped at 100 columns here. (They will probably read the HTML, so they don't care about how the reST source looks like).

>How do we verify this limit during code reviews?

To be clear, I'm not proposing to make a 106 column rule in such cases. I think we can make an exception for this code snippet (because it's supposed to show wrapping) and wrap everything else in our documentation at 80 columns. Hopefully, there won't be many snippets that even need wrapping.

But wrapping this one at 80 columns is fine for me, too.
</blockquote></details>

Yeah, typo, I meant `106` not `108`.

@dimakuv as @pwmarcz explained the idea is that this is an C code example that users will read in HTML. If we left the old wrapping (`80`), then such example wouldn't go through a review when copy-pasted...

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

___
*[Documentation/devel/howto-doc.rst, line 161 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiRWOH2klKlD9hHGDj:-Mx9DHuo9hcavrfWVa16:b-q5inqz) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/Documentation/devel/howto-doc.rst#L161)):*

Yeah, typo, I meant `106` not `108`.

@dimakuv as @pwmarcz explained the idea is that this is an C code example that users will read in HTML. If we left the old wrapping (`80`), then such example wouldn't go through a review when copy-pasted...
</blockquote></details>

Ah, this discussion is purely for this particular "C code embedded in RST file" case. Thanks and sorry for confusion.

pwmarcz (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

___
*[Documentation/devel/howto-doc.rst, line 161 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiRWOH2klKlD9hHGDj:-MxA6r72DiHuhogXxodj:bpflcp2) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/Documentation/devel/howto-doc.rst#L161)):*

Ah, this discussion is purely for this particular "C code embedded in RST file" case. Thanks and sorry for confusion.
</blockquote></details>

Not blocking, but @mkow please consider leaving a comment for the next person that tries to reformat this code :) (Either the one I suggested above, or something shorter)

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: 7 of 8 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @pwmarcz)

___
*[Documentation/devel/howto-doc.rst, line 161 at r5](https://reviewable.io/reviews/gramineproject/gramine/420#-MwiRWOH2klKlD9hHGDj:-MxC2HjDDdTiHjdvWUJg:bqmjwcz) ([raw file](https://github.com/gramineproject/gramine/blob/032e709a38fb4672dc9048b9f25c96ff8078f457/Documentation/devel/howto-doc.rst#L161)):*

+1 and blocking :)
</blockquote></details>

Added.

boryspoplawski (approved):
> Reviewed 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

pwmarcz (approved):
> Reviewed 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL)

pwmarcz (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420#-:-MxEmoD2CQXUKOGY6p9o:bnfp4nl)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (approved):
> Reviewed 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/420)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #14, opened 2019-08-05
RFC: Reorganize the repository


boryspoplawski:
Currently this repository has a bit messy layout. Proposed changes:
1. Move .h files into "include" directories (now there are symlinks pointing to src/.h)
2. Move shared code (between Pal and LibOS) out of the Pal directory
   ...
Please suggest more

mkow:
I went through the discussion here and almost everything is already done. I extracted two things we still need to do into https://github.com/gramineproject/gramine/issues/1273 and https://github.com/gramineproject/gramine/issues/1274, so that we can close this overloaded issue now.

PR #433, opened 2022-02-28
[Pal] Consolidate db_*.c for Pal/Linux and Pal/Linux-SGX

kailun-qin:
Note that this is just a proposal/WIP. I'd like to have some feedbacks before updating all the related files.

`db_*.c` of Pal/Linux and Pal/Linux_SGX are mostly following the same logic to implement narrow Drawbridge-like ABI interfaces where the only difference is host syscall or ocall. Maintaining them separately may lead to inconsistent behaviors when bugs are fixed in one place but not in the other and may consequently raise the maintenance cost.

This patch proposes to:
* share code of `db_*.c` in Pal/Linux and Pal/Linux_SGX
* have the common code maintained in Pal/Linux-common
* determine the hostcall type thru conditional compilation
* leverage the unified hostcall for syscall or ocall

Partially addresses https://github.com/gramineproject/gramine/issues/14

Related to https://github.com/gramineproject/graphene/issues/443

## How to test this PR? 

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/433)

### DIFF
--- a/Pal/include/host/Linux-common/hostcall.h
+++ b/Pal/include/host/Linux-common/hostcall.h
@@ -1,0 +1,14 @@
+#ifndef HOSTCALL_H_
+#define HOSTCALL_H_
+
+#if defined(DB_LINUX)
+#include "syscall.h"
+
+#define HOSTCALL(name, args...) DO_SYSCALL(name, args)
+#elif defined(DB_LINUX_SGX)
+#include "enclave_ocalls.h"
+
+#define HOSTCALL(name, args...) ocall_##name(args)
+#endif
+
+#endif // HOSTCALL_H_
--- a/Pal/src/host/Linux-SGX/meson.build
+++ b/Pal/src/host/Linux-SGX/meson.build
@@ -17,6 +17,7 @@ sgx_inc = [
 cflags_pal_sgx = [
     cflags_pal_common,
     '-DHOST_TYPE=Linux-SGX',
+    '-DDB_LINUX_SGX',
     # Some of the code uses `alignof` on expressions, which is a GNU extension. Silence Clang - it
     # complains but does support it.
     '-Wno-gnu-alignof-expression',
@@ -53,7 +54,6 @@ pal_sgx_map = custom_target('pal.map',
 pal_sgx_lds = join_paths(meson.current_source_dir(), 'enclave.lds')
 
 libpal_sgx = shared_library('pal',
-    'db_devices.c',
     'db_eventfd.c',
     'db_events.c',
     'db_exception.c',
@@ -82,6 +82,7 @@ libpal_sgx = shared_library('pal',
     'protected-files/protected_files.c',
     pal_sgx_asm_offsets_h,
     pal_common_sources,
+    pal_linux_common_sources_db,
     pal_linux_common_sources_enclave,
     gsgx_h,
 
--- a/Pal/src/host/Linux-common/db_devices.c
+++ b/Pal/src/host/Linux-common/db_devices.c
@@ -11,11 +11,13 @@
  */
 
 #include "api.h"
+#include "hostcall.h"
 #include "pal.h"
 #include "pal_error.h"
 #include "pal_flags_conv.h"
 #include "pal_internal.h"
 #include "pal_linux.h"
+#include "pal_linux_error.h"
 #include "perm.h"
 
 static int dev_open(PAL_HANDLE* handle, const char* type, const char* uri, enum pal_access access,
@@ -54,10 +56,10 @@ static int dev_open(PAL_HANDLE* handle, const char* type, const char* uri, enum 
         /* other devices must be opened through the host */
         hdl->dev.nonblocking = !!(options & PAL_OPTION_NONBLOCK);
 
-        ret = DO_SYSCALL(open, uri, PAL_ACCESS_TO_LINUX_OPEN(access)  |
-                                    PAL_CREATE_TO_LINUX_OPEN(create)  |
-                                    PAL_OPTION_TO_LINUX_OPEN(options),
-                         share);
+        ret = HOSTCALL(open, uri, PAL_ACCESS_TO_LINUX_OPEN(access)  |
+                PAL_CREATE_TO_LINUX_OPEN(create)  |
+                PAL_OPTION_TO_LINUX_OPEN(options),
+                share);
         if (ret < 0) {
             ret = unix_to_pal_error(ret);
             goto fail;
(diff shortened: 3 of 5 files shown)


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/433)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin)

___
*[Pal/include/host/Linux-common/hostcall.h, line 12 at r1](https://reviewable.io/reviews/gramineproject/gramine/433#-Mx-M7xTBQSEamnaQev3:-Mx-M7xTBQSEamnaQev4:b-lphsv6) ([raw file](https://github.com/gramineproject/gramine/blob/572d5c6e888c1c5e2fff40f2460e2fce639f0a2d/Pal/include/host/Linux-common/hostcall.h#L12)):*
> ```objc
> 
> #define HOSTCALL(name, args...) ocall_##name(args)
> #endif
> ```

I'm unsure about this idea. OCALLs and syscalls diverge, especially for "complicated" cases. So in many cases, we'll have to work around this nice header file.
___
*[Pal/src/host/Linux/meson.build, line 54 at r1](https://reviewable.io/reviews/gramineproject/gramine/433#-Mx-MkS21kRsA-czKSli:-Mx-MkS21kRsA-czKSlj:bm8opo5) ([raw file](https://github.com/gramineproject/gramine/blob/572d5c6e888c1c5e2fff40f2460e2fce639f0a2d/Pal/src/host/Linux/meson.build#L54)):*
> ```Meson
>         cflags_custom_stack_protector,
>         '-DHOST_TYPE=Linux',
>         '-DDB_LINUX',
> ```

In my proposal (see other comments), we won't need this macro.
___
*[Pal/src/host/Linux-common/db_devices.c, line 91 at r1](https://reviewable.io/reviews/gramineproject/gramine/433#-Mx-LPXU9fdMJjhZbn3o:-Mx-LPXU9fdMJjhZbn3p:b-msxb9j) ([raw file](https://github.com/gramineproject/gramine/blob/572d5c6e888c1c5e2fff40f2460e2fce639f0a2d/Pal/src/host/Linux-common/db_devices.c#L91)):*
> ```C
> #elif defined(DB_LINUX_SGX)
>     ssize_t bytes;
> #endif
> ```

This is not required. I think we should always use `ssize_t bytes` for all PALs.
___
*[Pal/src/host/Linux-common/db_devices.c, line 111 at r1](https://reviewable.io/reviews/gramineproject/gramine/433#-Mx-Lc3gDTPtqGaCBPR0:-Mx-Lc3gDTPtqGaCBPR1:b-msxb9j) ([raw file](https://github.com/gramineproject/gramine/blob/572d5c6e888c1c5e2fff40f2460e2fce639f0a2d/Pal/src/host/Linux-common/db_devices.c#L111)):*
> ```C
> #elif defined(DB_LINUX_SGX)
>     ssize_t bytes;
> #endif
> ```

This is not required. I think we should always use `ssize_t bytes` for all PALs.
___
*[Pal/src/host/Linux-common/db_devices.c, line 183 at r1](https://reviewable.io/reviews/gramineproject/gramine/433#-Mx-Ld7ACV4gitkJ1Luv:-Mx-Ld7ACV4gitkJ1Luw:b-7rc2n7) ([raw file](https://github.com/gramineproject/gramine/blob/572d5c6e888c1c5e2fff40f2460e2fce639f0a2d/Pal/src/host/Linux-common/db_devices.c#L183)):*
> ```C
> 
>         ocall_close(fd);
> #endif
> ```

I highly dislike this pattern. We don't want to put *any* PAL-specific code in `Linux-common/`.

Instead, we need some form of callbacks. I think ideally we should:
- Define a bunch of generic functions under `Linux-common/` (e.g., like your `dev_open()` above)
- Call these generic functions from `Linux/` and `Linux-SGX/` sources (with reverse "PAL-specific" callbacks if needed)

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/433)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/433#-Mx-clFEBhkizEIyaFcz:-Mx-clFEBhkizEIyaFd-:b-mfuc89) (no related file):*
@kailun-qin As per https://gramine.readthedocs.io/en/latest/devel/contributing.html#architectural-changes such changes are better first discussed with maintainers, best place for this is Gramine weekly meeting.
We already had some internal discussions about this, I even had fixing this in my personal queue (though deep enough that I never started implementing it).

Regarding this PR: I dislike this approach. ocalls and syscalls have different arguments and might have different semantics. The whole point of fixing this issue is to have **no** backend specific code in common code.

---

kailun-qin:
Based on the discussions in the 2.28 community meeting, the general idea/plan is to:
1. Similar to this PR proposal, still move the common/generic parts of `db_*.c` into Pal/Linux-common;
2. Instead of using macros, leverage `call_xxx` on link-time for backend-specific code;
3. Implement back-end specific code (incl. syscall or ocall) in Pal/Linux and Pal/Linux-SGX separately;
4. Hold and to follow up on this one after the trusted/protected fs rewrite completes.

Pls kindly correct me and/or extend if I understand anything wrong. Thanks!

dimakuv:
> 2. Instead of using macros, leverage `call_xxx` on link-time for backend-specific code;

I vote for `hostcall_xxx` prefix.

kailun-qin:
> > 2. Instead of using macros, leverage `call_xxx` on link-time for backend-specific code;
> 
> I vote for `hostcall_xxx` prefix.

Thanks @dimakuv! I used a dummy name to simply illustrate the idea (also to differ from `ocall_xxx`). Personally I also vote for `hostcall_xxx` prefix.

boryspoplawski (commented):
Using `ocall_xxx` directly has some advantages: one less function call on PAL SGX and we do not create another API layer (compared to PAL API + hostcall + ocall). Disadvantage is slightly misleading name on Linux PAL (tho I personally don't mind that much). So I would vote for sticking to `ocall_xxx` names

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/433#-:-Mx3e76hB9xHw4wfdqfD:bf7s637)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin)

dimakuv:
This never went anywhere, and the PR is very outdated at this point. I will just close it.

mkow:
Hmm, but I like the idea behind it. @kailun-qin: Are you interested in finishing this up?

kailun-qin:
> Hmm, but I like the idea behind it. @kailun-qin: Are you interested in finishing this up?

Sure, I can (re-)look into it.

mkow:
I'll reopen it then :)

PR #437, opened 2022-03-03
[common] Fix potential null ptr dereference in "alloc_concat3"

kailun-qin:
## Description of the changes 

The input pointers of `alloc_concat3` could be null. Passing a null
pointer to `memcpy()` would produce possible null pointer dereference or
undefined behaviors.

This patch ensures that null ptr inputs are not passed into `memcpy()`.

## How to test this PR? 

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/437)

### DIFF
--- a/common/src/string/utils.c
+++ b/common/src/string/utils.c
@@ -30,11 +30,11 @@ char* alloc_concat3(const char* a, size_t a_len, const char* b, size_t b_len,
     if (!buf)
         return NULL;
 
-    if (a_len)
+    if (a != NULL && a_len)
         memcpy(buf, a, a_len);
-    if (b_len)
+    if (b != NULL && b_len)
         memcpy(buf + a_len, b, b_len);
-    if (c_len)
+    if (c != NULL && c_len)
         memcpy(buf + a_len + b_len, c, c_len);
 
     buf[a_len + b_len + c_len] = '\0';


boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/437)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin)

___
*[common/src/string/utils.c, line 33 at r1](https://reviewable.io/reviews/gramineproject/gramine/437#-MxEnUXn2alPE_EgHxE9:-MxEnUXn2alPE_EgHxEA:bpwbht4) ([raw file](https://github.com/gramineproject/gramine/blob/f0fb1732f63bc5df6dee9ef6adf0ddadfd9e4289/common/src/string/utils.c#L33)):*
> ```C
>         return NULL;
> 
>     if (a != NULL && a_len)
> ```

The old version was correct. `a` is never `NULL` if `a_len > 0`.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/437)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski)

___
*[common/src/string/utils.c, line 33 at r1](https://reviewable.io/reviews/gramineproject/gramine/437#-MxEnUXn2alPE_EgHxE9:-MxF5Y2Q1bW1FLc2YBGO:brursee) ([raw file](https://github.com/gramineproject/gramine/blob/f0fb1732f63bc5df6dee9ef6adf0ddadfd9e4289/common/src/string/utils.c#L33)):*

The old version was correct. `a` is never `NULL` if `a_len > 0`.
</blockquote></details>

Sorry I'm no C expert but I doubt this.

For below example:
```
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

char* alloc_concat3(const char* a, size_t a_len, const char* b, size_t b_len,
                    const char* c, size_t c_len) {
    a_len = (a_len != (size_t)-1) ? a_len : (a ? strlen(a) : 0);
    b_len = (b_len != (size_t)-1) ? b_len : (b ? strlen(b) : 0);
    c_len = (c_len != (size_t)-1) ? c_len : (c ? strlen(c) : 0);

    char* buf = malloc(a_len + b_len + c_len + 1);
    if (!buf)
        return NULL;

    printf("%p, %ld\n", a, a_len);

    if (a_len)
        memcpy(buf, a, a_len);
    if (b_len)
        memcpy(buf + a_len, b, b_len);
    if (c_len)
        memcpy(buf + a_len + b_len, c, c_len);

    buf[a_len + b_len + c_len] = '\0';
    return buf;
}

int main(void) {
    const char* a = NULL;
    size_t a_len = 3;

    char* buf = alloc_concat3(a, a_len, NULL, 0, NULL, 0);

    printf("%s\n", buf);

    return 0;
}
```
I got the following result:
```
(nil), 3
Segmentation fault (core dumped)
```

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/437)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski)

___
*[common/src/string/utils.c, line 33 at r1](https://reviewable.io/reviews/gramineproject/gramine/437#-MxEnUXn2alPE_EgHxE9:-MxF6ixcDVZpVXMdjp8e:byk0jqu) ([raw file](https://github.com/gramineproject/gramine/blob/f0fb1732f63bc5df6dee9ef6adf0ddadfd9e4289/common/src/string/utils.c#L33)):*

Sorry I'm no C expert but I doubt this.

For below example:
```
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

char* alloc_concat3(const char* a, size_t a_len, const char* b, size_t b_len,
                    const char* c, size_t c_len) {
    a_len = (a_len != (size_t)-1) ? a_len : (a ? strlen(a) : 0);
    b_len = (b_len != (size_t)-1) ? b_len : (b ? strlen(b) : 0);
    c_len = (c_len != (size_t)-1) ? c_len : (c ? strlen(c) : 0);

    char* buf = malloc(a_len + b_len + c_len + 1);
    if (!buf)
        return NULL;

    printf("%p, %ld\n", a, a_len);

    if (a_len)
        memcpy(buf, a, a_len);
    if (b_len)
        memcpy(buf + a_len, b, b_len);
    if (c_len)
        memcpy(buf + a_len + b_len, c, c_len);

    buf[a_len + b_len + c_len] = '\0';
    return buf;
}

int main(void) {
    const char* a = NULL;
    size_t a_len = 3;

    char* buf = alloc_concat3(a, a_len, NULL, 0, NULL, 0);

    printf("%s\n", buf);

    return 0;
}
```
I got the following result:
```
(nil), 3
Segmentation fault (core dumped)
```
</blockquote></details>

The idea is that callers of this function never set `a_len > 0`. If the caller does something like what you do, this is the problem with the caller.

I would close this PR. But if you really want, at most we could add `assert()` statements here instead of `if` checks. But generally I'd just close it.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/437)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski and @dimakuv)

___
*[common/src/string/utils.c, line 33 at r1](https://reviewable.io/reviews/gramineproject/gramine/437#-MxEnUXn2alPE_EgHxE9:-MxF8Uvh9pi9s0eoze8c:buw3omj) ([raw file](https://github.com/gramineproject/gramine/blob/f0fb1732f63bc5df6dee9ef6adf0ddadfd9e4289/common/src/string/utils.c#L33)):*

The idea is that callers of this function never set `a_len > 0`. If the caller does something like what you do, this is the problem with the caller.

I would close this PR. But if you really want, at most we could add `assert()` statements here instead of `if` checks. But generally I'd just close it.
</blockquote></details>

Got it, thanks. If this is something expected, I'm totally fine w/ it.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/437)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski and @dimakuv)

___
*[common/src/string/utils.c, line 33 at r1](https://reviewable.io/reviews/gramineproject/gramine/437#-MxEnUXn2alPE_EgHxE9:-MxFEohjBRsxMPbKZDoM:bu7kwj1) ([raw file](https://github.com/gramineproject/gramine/blob/f0fb1732f63bc5df6dee9ef6adf0ddadfd9e4289/common/src/string/utils.c#L33)):*

Got it, thanks. If this is something expected, I'm totally fine w/ it.
</blockquote></details>

@kailun-qin: `alloc_concat3((void*)0x123, 10, NULL, 0, NULL, 0);` will also crash, and what? ;) There's no way and no point in checking all input arguments in all functions, especially pointers. The interface for this function is "you need to pass a pointer to char arrays and their respective lengths, and the function will concatenate them". If you pass garbage then you'll get garbage ;)

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/437)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv)

___
*[common/src/string/utils.c, line 33 at r1](https://reviewable.io/reviews/gramineproject/gramine/437#-MxEnUXn2alPE_EgHxE9:-MxFN8sW0LxD9BbAQhJP:b-mulc48) ([raw file](https://github.com/gramineproject/gramine/blob/f0fb1732f63bc5df6dee9ef6adf0ddadfd9e4289/common/src/string/utils.c#L33)):*

@kailun-qin: `alloc_concat3((void*)0x123, 10, NULL, 0, NULL, 0);` will also crash, and what? ;) There's no way and no point in checking all input arguments in all functions, especially pointers. The interface for this function is "you need to pass a pointer to char arrays and their respective lengths, and the function will concatenate them". If you pass garbage then you'll get garbage ;)
</blockquote></details>

`a_len` is the length of the array at pointer `a`. If you pass `(NULL, 3)`, then it's your fault as a caller of this function. As @mkow pointed out, you could pass `((void*)0x123, 10)` which is invalid as well.
Note that `(NULL, 0)` is perfectly fine - we are allowed to touch `0` bytes at `NULL` pointer.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/437)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski and @dimakuv)

___
*[common/src/string/utils.c, line 33 at r1](https://reviewable.io/reviews/gramineproject/gramine/437#-MxEnUXn2alPE_EgHxE9:-MxFPYUv4cT6spf7FZew:b-u3ofp) ([raw file](https://github.com/gramineproject/gramine/blob/f0fb1732f63bc5df6dee9ef6adf0ddadfd9e4289/common/src/string/utils.c#L33)):*

`a_len` is the length of the array at pointer `a`. If you pass `(NULL, 3)`, then it's your fault as a caller of this function. As @mkow pointed out, you could pass `((void*)0x123, 10)` which is invalid as well.
Note that `(NULL, 0)` is perfectly fine - we are allowed to touch `0` bytes at `NULL` pointer.
</blockquote></details>

Yep, it's fine for me.

Indeed the input ptr could be invalid but there is no portable way to verify it, however checking for null is a different story. It seems to be general practice to check null before memcpy: https://wiki.sei.cmu.edu/confluence/display/c/EXP34-C.+Do+not+dereference+null+pointers.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/437)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv)

___
*[common/src/string/utils.c, line 33 at r1](https://reviewable.io/reviews/gramineproject/gramine/437#-MxEnUXn2alPE_EgHxE9:-MxFShFP6dFeWFCMi3Fu:b-npecq2) ([raw file](https://github.com/gramineproject/gramine/blob/f0fb1732f63bc5df6dee9ef6adf0ddadfd9e4289/common/src/string/utils.c#L33)):*

@kailun-qin: No, the case you linked is different. It's not about checking for `NULL` on `mempcy()` but about checking the result of `malloc()` which in its specification says that it returns `NULL` on failure.
</blockquote></details>

@mkow it's about both of these cases, `memcpy` too, but the code is different and does not check for `length == 0`.

PR #449, opened 2022-03-09
[LibOS] Pass file offset to `read` and `write` callbacks

pwmarcz:
## Description of the changes 

Instead of using the handle position, `read` and `write` receive file position as a parameter (and might choose to update it). This is a similar design to Linux internals.

As a result, there's no need to emulate operations like `pread` using `seek`, and filesystem callbacks can be simpler.

This is in preparation for adding another filesystem (for protected files).

## How to test this PR? 

CI.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/449)

### DIFF
--- a/LibOS/shim/include/shim_fs.h
+++ b/LibOS/shim/include/shim_fs.h
@@ -33,11 +33,39 @@ struct shim_fs_ops {
     /* close: clean up the file state inside the handle */
     int (*close)(struct shim_handle* hdl);
 
-    /* read: the content from the file opened as handle */
-    ssize_t (*read)(struct shim_handle* hdl, void* buf, size_t count);
+    /*
+     * \brief Read from file.
+     *
+     * \param         hdl    File handle.
+     * \param         buf    Buffer to read into.
+     * \param         count  Size of `buffer`.
+     * \param[in,out] pos    Position at which to start reading. Might be updated on success.
+     *
+     * \returns Number of bytes read on success, negative error code on failure.
+     *
+     * This callback updates `*pos` if the file is seekable (e.g. not a pipe or socket).
+     *
+     * TODO: Callers should make sure that `count` doesn't overflow `ssize_t`, and `*pos + count`
+     * doesn't overflow `file_off_t`.
+     */
+    ssize_t (*read)(struct shim_handle* hdl, void* buf, size_t count, file_off_t* pos);
 
-    /* write: the content from the file opened as handle */
-    ssize_t (*write)(struct shim_handle* hdl, const void* buf, size_t count);
+    /*
+     * \brief Write to file.
+     *
+     * \param         hdl    File handle.
+     * \param         buf    Buffer to write from.
+     * \param         count  Size of `buffer`.
+     * \param[in,out] pos    Position at which to start writing. Might be updated on success.
+     *
+     * \returns Number of bytes written on success, negative error code on failure.
+     *
+     * This callback updates `*pos` if the file is seekable (e.g. not a pipe or socket).
+     *
+     * TODO: Callers should make sure that `count` doesn't overflow `ssize_t`, and `*pos + count`
+     * doesn't overflow `file_off_t`.
+     */
+    ssize_t (*write)(struct shim_handle* hdl, const void* buf, size_t count, file_off_t* pos);
 
     /* mmap: mmap handle to address */
     int (*mmap)(struct shim_handle* hdl, void** addr, size_t size, int prot, int flags,
--- a/LibOS/shim/include/shim_handle.h
+++ b/LibOS/shim/include/shim_handle.h
@@ -186,7 +186,7 @@ struct shim_handle {
      */
     struct shim_inode* inode;
 
-    /* Offset in file */
+    /* Offset in file. Protected by `pos_lock`. */
     file_off_t pos;
 
     /* This list contains `shim_epoll_item` objects this handle is part of. All accesses should be
@@ -229,6 +229,11 @@ struct shim_handle {
     int flags; /* Linux' O_* flags */
     int acc_mode;
     struct shim_lock lock;
+
+    /* Lock for handle position (`pos`). Intended for operations that change the position (e.g.
+     * `read`, `seek` but not `pread`). This lock should be taken *before* `shim_handle.lock` and
+     * `shim_inode.lock`. */
+    struct shim_lock pos_lock;
 };
 
 /* allocating / manage handle */
--- a/LibOS/shim/src/bookkeep/shim_handle.c
+++ b/LibOS/shim/src/bookkeep/shim_handle.c
@@ -342,6 +342,11 @@ struct shim_handle* get_new_handle(void) {
         free_mem_obj_to_mgr(handle_mgr, new_handle);
         return NULL;
     }
+    if (!create_lock(&new_handle->pos_lock)) {
+        destroy_lock(&new_handle->lock);
+        free_mem_obj_to_mgr(handle_mgr, new_handle);
+        return NULL;
+    }
     INIT_LISTP(&new_handle->epoll_items);
     new_handle->epoll_items_count = 0;
     return new_handle;
@@ -470,6 +475,7 @@ void get_handle(struct shim_handle* hdl) {
 
 static void destroy_handle(struct shim_handle* hdl) {
     destroy_lock(&hdl->lock);
+    destroy_lock(&hdl->pos_lock);
 
     free_mem_obj_to_mgr(handle_mgr, hdl);
 }
@@ -712,6 +718,7 @@ BEGIN_CP_FUNC(handle) {
         new_hdl->dentry = NULL;
         REF_SET(new_hdl->ref_count, 0);
         clear_lock(&new_hdl->lock);
+        clear_lock(&new_hdl->pos_lock);
 
         DO_CP(fs, hdl->fs, &new_hdl->fs);
 
@@ -789,6 +796,10 @@ BEGIN_RS_FUNC(handle) {
     CP_REBASE(hdl->epoll_items);
 
     if (!create_lock(&hdl->lock)) {
+        return -ENOMEM;
+    }
+
+    if (!create_lock(&hdl->pos_lock)) {
         return -ENOMEM;
     }
 
--- a/LibOS/shim/src/fs/chroot/fs.c
+++ b/LibOS/shim/src/fs/chroot/fs.c
@@ -11,10 +11,8 @@
  * finished.
  */
 
-#define _POSIX_C_SOURCE 200809L /* for SSIZE_MAX */
 #include <asm/mman.h>
 #include <errno.h>
-#include <limits.h>
 #include <linux/fcntl.h>
 
 #include "pal.h"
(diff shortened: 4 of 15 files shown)


dimakuv (changes requested):
> Reviewed 14 of 14 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/449)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[LibOS/shim/include/shim_handle.h, line 235 at r1](https://reviewable.io/reviews/gramineproject/gramine/449#-MxmoqQf-tqBp9gmwHHN:-MxmoqQf-tqBp9gmwHHO:b-syi38) ([raw file](https://github.com/gramineproject/gramine/blob/df038d82705bef4f3e0b6aa9bf8262cc11c2d250/LibOS/shim/include/shim_handle.h#L235)):*
> ```objc
>     /* Lock for handle position (`pos`). Intended for operations that change the position (e.g.
>      * `read`, `seek` but not `pread`). This lock should be taken *before* `shim_handle.lock` and
>      * `shim_inode.lock`. */
> ```

Why can't we protect `pos` via `shim_handle.lock`? Why this complexity of a new lock? For speed?
___
*[LibOS/shim/src/fs/chroot/fs.c, line 364 at r1](https://reviewable.io/reviews/gramineproject/gramine/449#-MxmrAWZ0ZiD_D2XkwIR:-MxmrAWZ0ZiD_D2XkwIS:b-gi0ozf) ([raw file](https://github.com/gramineproject/gramine/blob/7846eccb5c226e9c6aa8775058d43506c6ad7cfd/LibOS/shim/src/fs/chroot/fs.c#L364)):*
> ```C
>         goto out;
>     }
>     assert(count <= actual_count);
> ```

How did this work before? Is it because we always hit the case `count == actual_count`?
___
*[LibOS/shim/src/sys/shim_file.c, line 509 at r1](https://reviewable.io/reviews/gramineproject/gramine/449#-Mxmsmed3XHmqKlXhUb9:-Mxmsmed3XHmqKlXhUbA:b-b8ii6m) ([raw file](https://github.com/gramineproject/gramine/blob/df038d82705bef4f3e0b6aa9bf8262cc11c2d250/LibOS/shim/src/sys/shim_file.c#L509)):*
> ```C
> 
>     if (offset) {
>         *offset = pos_in;
> ```

Maybe we should keep the comment? The behavior of `sendfile()` is not that obvious. (Maybe a shorter comment would do.)

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/449)* status: 13 of 14 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[LibOS/shim/include/shim_handle.h, line 235 at r1](https://reviewable.io/reviews/gramineproject/gramine/449#-MxmoqQf-tqBp9gmwHHN:-MxnsWgj3vVc6ucmzo7q:b-9nwhnb) ([raw file](https://github.com/gramineproject/gramine/blob/df038d82705bef4f3e0b6aa9bf8262cc11c2d250/LibOS/shim/include/shim_handle.h#L235)):*

Why can't we protect `pos` via `shim_handle.lock`? Why this complexity of a new lock? For speed?
</blockquote></details>

`shim_handle.lock` is not compatible with the current use:

* during `read` syscall, we want exclusive access for the whole `read` callback,
* during `readv` syscall, we want exclusive access *during multiple `read` callbacks* (since the whole `readv` is supposed to be atomic),
* during `pread` syscall, we should allow multiple `read` callbacks at the same time.

To allow simultaneous `pread` syscalls, we cannot take `shim_handle.lock` for the whole `read` callback.

I we don't care about `pread`, we could use `shim_handle.lock`, but to ensure `readv` is atomic, the *caller* (not the callback) would have to take the lock. However, as discussed in one of the earlier PRs (https://github.com/gramineproject/gramine/pull/297#pullrequestreview-870948537), we want `shim_handle.lock` to be the last lock in the hierarchy. That would mean the callbacks cannot take *any* locks (e.g. they cannot lock the inode to update metadata).

AFAICT, Linux does a similar thing: [struct file](https://elixir.bootlin.com/linux/latest/source/include/linux/fs.h#L962) has `f_lock` and `f_pos_lock` fields.

(Basically, I trust that design, because I tried to come up with something simpler and I failed).
___
*[LibOS/shim/src/fs/chroot/fs.c, line 364 at r1](https://reviewable.io/reviews/gramineproject/gramine/449#-MxmrAWZ0ZiD_D2XkwIR:-MxnxmreECF9lH8-LxlJ:b-f30awt) ([raw file](https://github.com/gramineproject/gramine/blob/7846eccb5c226e9c6aa8775058d43506c6ad7cfd/LibOS/shim/src/fs/chroot/fs.c#L364)):*

How did this work before? Is it because we always hit the case `count == actual_count`?
</blockquote></details>

Yeah, I think so... It's easier to hit `actual_count < count` during read (because of EOF), but this is write.
___
*[LibOS/shim/src/sys/shim_file.c, line 509 at r1](https://reviewable.io/reviews/gramineproject/gramine/449#-Mxmsmed3XHmqKlXhUb9:-Mxo-l7l011ROfLtrH7t:b-yj1hik) ([raw file](https://github.com/gramineproject/gramine/blob/df038d82705bef4f3e0b6aa9bf8262cc11c2d250/LibOS/shim/src/sys/shim_file.c#L509)):*

Maybe we should keep the comment? The behavior of `sendfile()` is not that obvious. (Maybe a shorter comment would do.)
</blockquote></details>

Right. It's actually more tricky: I should also update the offset in file handle.

The relevant part of manpage is a bit verbose, but I'd rather add it in full than risk paraphrasing it.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/449)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[LibOS/shim/include/shim_handle.h, line 235 at r1](https://reviewable.io/reviews/gramineproject/gramine/449#-MxmoqQf-tqBp9gmwHHN:-MxrtrBt5knesskfsJtR:b-e1jt32) ([raw file](https://github.com/gramineproject/gramine/blob/df038d82705bef4f3e0b6aa9bf8262cc11c2d250/LibOS/shim/include/shim_handle.h#L235)):*

`shim_handle.lock` is not compatible with the current use:

* during `read` syscall, we want exclusive access for the whole `read` callback,
* during `readv` syscall, we want exclusive access *during multiple `read` callbacks* (since the whole `readv` is supposed to be atomic),
* during `pread` syscall, we should allow multiple `read` callbacks at the same time.

To allow simultaneous `pread` syscalls, we cannot take `shim_handle.lock` for the whole `read` callback.

I we don't care about `pread`, we could use `shim_handle.lock`, but to ensure `readv` is atomic, the *caller* (not the callback) would have to take the lock. However, as discussed in one of the earlier PRs (https://github.com/gramineproject/gramine/pull/297#pullrequestreview-870948537), we want `shim_handle.lock` to be the last lock in the hierarchy. That would mean the callbacks cannot take *any* locks (e.g. they cannot lock the inode to update metadata).

AFAICT, Linux does a similar thing: [struct file](https://elixir.bootlin.com/linux/latest/source/include/linux/fs.h#L962) has `f_lock` and `f_pos_lock` fields.

(Basically, I trust that design, because I tried to come up with something simpler and I failed).
</blockquote></details>

Thanks. I understood the point about " the caller (not the callback) would have to take the lock" and the problem with this ("`shim_handle.lock` to be the last lock"). An interesting problem.
___
*[LibOS/shim/src/sys/shim_file.c, line 474 at r2](https://reviewable.io/reviews/gramineproject/gramine/449#-Mxrulb88ts-KttYPR_Y:-Mxrulb88ts-KttYPR_Z:b-4ihk23) ([raw file](https://github.com/gramineproject/gramine/blob/26a4eeeebd5dcf8bda0b080eb5a12e75ad4fcb67/LibOS/shim/src/sys/shim_file.c#L474)):*
> ```C
>         pos_in = in_hdl->pos;
>         unlock(&in_hdl->pos_lock);
>     }
> ```

Looks like we don't have any test that would use `sendfile()` with `offset = NULL` and then verifies the position in the file? (Otherwise your first iteration of this PR would fail.)

Could you check if maybe we can enable some `sendfile` test in LTP that checks this? And/or write a new LibOS test?

mkow (approved):
> Reviewed 13 of 14 files at r1, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/449)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

___
*[LibOS/shim/src/sys/shim_file.c, line 448 at r3](https://reviewable.io/reviews/gramineproject/gramine/449#-MxuHcFd9HSTJZAMog0t:-MxuHcFd9HSTJZAMog0u:b-qrpih9) ([raw file](https://github.com/gramineproject/gramine/blob/903cf98b3ea2edb2cc98924287df07111679e6f9/LibOS/shim/src/sys/shim_file.c#L448)):*
> ```C
> 
>     /*
>      * `man sendfile` says:
> ```

Not blocking, but not sure if we should quote `man` as a guidance on how syscalls work - it describes the libc wrappers which often have different semantics and often not all the differences/nuances are described in the "C library/kernel differences" section.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/449)* status: :shipit: complete! all files reviewed, all discussions resolved

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/449)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[LibOS/shim/src/sys/shim_file.c, line 448 at r3](https://reviewable.io/reviews/gramineproject/gramine/449#-MxuHcFd9HSTJZAMog0t:-My6d0Kd1D3M1AVC-SDz:b-gno2us) ([raw file](https://github.com/gramineproject/gramine/blob/903cf98b3ea2edb2cc98924287df07111679e6f9/LibOS/shim/src/sys/shim_file.c#L448)):*

Not blocking, but not sure if we should quote `man` as a guidance on how syscalls work - it describes the libc wrappers which often have different semantics and often not all the differences/nuances are described in the "C library/kernel differences" section.
</blockquote></details>

OK. I'm not against quoting the man page in general, because often it's the _only_ piece of documentation we have, and if the explanation is well-written and accurate, it might be better to provide it verbatim.

But here, I think a rephrasing will be much simpler.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/449)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/449)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/449)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/449)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #31, opened 2021-09-13
[Pal/Linux-SGX] Fix renaming issue with protected files

svenkata9:
## Description of the changes 
With certain workloads involving Federated Learning, the frameworks create protected files and rename them as part of the flow. This does not work currently with protected files because every protected file maintains the path name in its metadata (stored when the file was first created) and which is checked during open to make sure the incoming path is the same as what is already stored in the metadata for this file.

This PR is to address this issue by handling rename for protected files. A normal rename call does not need the file to be open in RW mode, but since the metadata needs to be updated with the new path, we pass a special `PAL_OPTION_RENAME` flag to `DkStreamOpen` during rename and open the protected file in RW mode with required permissions, update the metadata with the new path and flush it. Then the flow goes through the one for normal file renaming. If the ocall to rename fails, we restore the metadata back to what it was earlier so the file remains usable still.

Signed-off-by: Sankaranarayanan Venkatasubramanian <sankaranarayanan.venkatasubramanian@intel.com>

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/31)

PR #497, opened 2022-04-05
[LibOS] Support renaming encrypted files

pwmarcz:
## Description of the changes 

This adds rename support for encrypted files. The lowest layer is much like the old implementation in #31, but thanks to encrypted files being in LibOS, the issues mentioned there do not appear:

* it's possible to rename a file that is currently open
* it's not possible to rename a file "into" or "out of" encrypted files (since rename works only within a mounted filesystem)
* this shouldn't introduce problems with reference counting

Part of protected files redesign (#371).

Note that this does *not* affect `sgx.protected_files`, which still uses the old subsystem (we'll switch once the new one has all the same features).

## How to test this PR? 

The `rename_unlink` test is enabled for encrypted files. It tests various scenarios for deleting and renaming a file, possibly overwriting another file, keeping an open handle to the now-unlinked file, etc.

I also performed a manual test using the Python example:

```toml
fs.mounts = [
  { type = "encrypted", path = "/enc", uri = "file:enc" },
]
fs.insecure__keys.default = 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa'
```

```py
from pathlib import Path
import os

Path('enc/a').write_text('Hello world')
os.rename('enc/a', 'enc/b')
print(Path('enc/b').read_text())
```

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/497)

### DIFF
--- a/LibOS/shim/include/shim_fs_encrypted.h
+++ b/LibOS/shim/include/shim_fs_encrypted.h
@@ -137,6 +137,7 @@ int encrypted_file_read(struct shim_encrypted_file* enc, void* buf, size_t buf_s
                         file_off_t offset, size_t* out_count);
 int encrypted_file_write(struct shim_encrypted_file* enc, const void* buf, size_t buf_size,
                          file_off_t offset, size_t* out_count);
+int encrypted_file_rename(struct shim_encrypted_file* enc, const char* new_uri);
 
 int encrypted_file_get_size(struct shim_encrypted_file* enc, file_off_t* out_size);
 int encrypted_file_set_size(struct shim_encrypted_file* enc, file_off_t size);
--- a/LibOS/shim/src/fs/chroot/encrypted.c
+++ b/LibOS/shim/src/fs/chroot/encrypted.c
@@ -28,7 +28,6 @@
  * - mounting with other keys than "default"
  * - mounting with special keys (SGX MRENCLAVE and MRSIGNER)
  * - setting keys through /dev/attestation (and making sure they're copied to child process)
- * - rename (needs support in the `protected_files` module)
  * - mmap
  * - truncate (the current `truncate` operation works only for extending the file, support for
  *   truncation needs to be added to the `protected_files` module)
@@ -272,6 +271,34 @@ out:
     return ret;
 }
 
+static int chroot_encrypted_rename(struct shim_dentry* old, struct shim_dentry* new) {
+    assert(locked(&g_dcache_lock));
+    assert(old->inode);
+    assert(old->inode->type == S_IFREG);
+
+    int ret;
+    char* new_uri = NULL;
+
+    ret = chroot_dentry_uri(new, old->inode->type, &new_uri);
+    if (ret < 0)
+        return ret;
+
+    lock(&old->inode->lock);
+
+    struct shim_encrypted_file* enc = old->inode->data;
+
+    ret = encrypted_file_get(enc);
+    if (ret < 0)
+        goto out;
+
+    ret = encrypted_file_rename(enc, new_uri);
+    encrypted_file_put(enc);
+out:
+    unlock(&old->inode->lock);
+    free(new_uri);
+    return ret;
+}
+
 static int chroot_encrypted_chmod(struct shim_dentry* dent, mode_t perm) {
     assert(locked(&g_dcache_lock));
     assert(!dent->inode);
@@ -411,6 +438,7 @@ struct shim_d_ops chroot_encrypted_d_ops = {
     .stat          = &generic_inode_stat,
     .readdir       = &chroot_readdir, /* same as in `chroot` filesystem */
     .unlink        = &chroot_unlink,  /* same as in `chroot` filesystem */
+    .rename        = &chroot_encrypted_rename,
     .chmod         = &chroot_encrypted_chmod,
     .idrop         = &chroot_encrypted_idrop,
 };
--- a/LibOS/shim/src/fs/shim_fs_encrypted.c
+++ b/LibOS/shim/src/fs/shim_fs_encrypted.c
@@ -515,6 +515,53 @@ int encrypted_file_set_size(struct shim_encrypted_file* enc, file_off_t size) {
     return 0;
 }
 
+int encrypted_file_rename(struct shim_encrypted_file* enc, const char* new_uri) {
+    assert(enc->pf);
+
+    char* new_uri_copy = strdup(new_uri);
+    if (!new_uri_copy)
+        return -ENOMEM;
+
+    int ret;
+
+    assert(strstartswith(enc->uri, URI_PREFIX_FILE));
+    const char* old_path = enc->uri + static_strlen(URI_PREFIX_FILE);
+
+    assert(strstartswith(new_uri, URI_PREFIX_FILE));
+    const char* new_path = new_uri + static_strlen(URI_PREFIX_FILE);
+
+    pf_status_t pfs = pf_rename(enc->pf, new_path);
+    if (PF_FAILURE(pfs)) {
+        log_warning("%s: pf_rename failed: %s", __func__, pf_strerror(pfs));
+        ret = -EACCES;
+        goto out;
+    }
+
+    ret = DkStreamChangeName(enc->pal_handle, new_uri);
+    if (ret < 0) {
+        log_warning("%s: DkStreamChangeName failed: %d", __func__, ret);
+
+        /* We failed to rename the file. Try to restore the name in header. */
+        pfs = pf_rename(enc->pf, old_path);
+        if (PF_FAILURE(pfs)) {
+            log_warning("%s: pf_rename (during cleanup) failed, the file might be unusable: %s",
+                        __func__, pf_strerror(pfs));
+        }
+
+        ret = pal_to_unix_errno(ret);
+        goto out;
+    }
+
+    free(enc->uri);
+    enc->uri = new_uri_copy;
+    new_uri_copy = NULL;
+    ret = 0;
+
+out:
+    free(new_uri_copy);
+    return ret;
+}
+
 BEGIN_CP_FUNC(encrypted_file) {
     __UNUSED(size);
 
--- a/LibOS/shim/test/regression/test_libos.py
+++ b/LibOS/shim/test/regression/test_libos.py
@@ -603,7 +603,7 @@ class TC_30_Syscall(RegressionTestCase):
                     os.unlink(path)
         self.assertIn('TEST OK', stdout)
 
-    @unittest.skip('Protected files do not support renaming yet')
+    @unittest.skip('Protected files (as implemented in Linux-SGX PAL) do not support renaming yet')
     @unittest.skipUnless(HAS_SGX,
         'Protected files are only available with SGX')
     def test_034_rename_unlink_pf(self):
@@ -618,7 +618,23 @@ class TC_30_Syscall(RegressionTestCase):
                     os.unlink(path)
         self.assertIn('TEST OK', stdout)
 
-    def test_035_rename_unlink_tmpfs(self):
+    def test_035_rename_unlink_enc(self):
+        os.makedirs('tmp/enc', exist_ok=True)
+        file1 = 'tmp/enc/file1'
+        file2 = 'tmp/enc/file2'
+        # Delete the files: the test overwrites them anyway, but it may fail if they are malformed.
+        for path in [file1, file2]:
+            if os.path.exists(path):
+                os.unlink(path)
+        try:
+            stdout, _ = self.run_binary(['rename_unlink', file1, file2])
+        finally:
+            for path in [file1, file2]:
+                if os.path.exists(path):
+                    os.unlink(path)
+        self.assertIn('TEST OK', stdout)
+
+    def test_036_rename_unlink_tmpfs(self):
         file1 = '/mnt/tmpfs/file1'
         file2 = '/mnt/tmpfs/file2'
         stdout, _ = self.run_binary(['rename_unlink', file1, file2])
--- a/common/src/protected_files/protected_files.c
+++ b/common/src/protected_files/protected_files.c
@@ -1254,6 +1254,26 @@ pf_status_t pf_set_size(pf_context_t* pf, uint64_t size) {
     return PF_STATUS_NOT_IMPLEMENTED;
 }
 
+pf_status_t pf_rename(pf_context_t* pf, const char* new_path) {
+    if (!g_initialized)
+        return PF_STATUS_UNINITIALIZED;
+
+    if (!(pf->mode & PF_FILE_MODE_WRITE))
+        return PF_STATUS_INVALID_MODE;
+
+    size_t new_path_size = strlen(new_path) + 1;
+    if (new_path_size > PATH_MAX_SIZE)
+        return PF_STATUS_PATH_TOO_LONG;
+
+    memset(pf->encrypted_part_plain.path, 0, sizeof(pf->encrypted_part_plain.path));
+    memcpy(pf->encrypted_part_plain.path, new_path, new_path_size);
+    pf->need_writing = true;
+    if (!ipf_internal_flush(pf))
+        return pf->last_error;
+
+    return PF_STATUS_SUCCESS;
+}
+
 pf_status_t pf_read(pf_context_t* pf, uint64_t offset, size_t size, void* output,
                     size_t* bytes_read) {
     if (!g_initialized)
--- a/common/src/protected_files/protected_files.h
+++ b/common/src/protected_files/protected_files.h
@@ -261,6 +261,17 @@ pf_status_t pf_get_size(pf_context_t* pf, uint64_t* size);
 pf_status_t pf_set_size(pf_context_t* pf, uint64_t size);
 
 /*!
+ * \brief Rename a PF.
+ *
+ * \param pf        PF context.
+ * \param new_path  New file path.
+ *
+ * Updates the path inside protected file header, and flushes all changes. The caller is responsible
+ * for renaming the underlying file.
+ */
+pf_status_t pf_rename(pf_context_t* pf, const char* new_path);
+
+/*!
  * \brief Get underlying handle of a PF
  *
  * \param [in] pf PF context


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/497)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

___
*[-- commits, line 4 at r1](https://reviewable.io/reviews/gramineproject/gramine/497#-Mzy8jEF2qqa2rp23nLL):*
Could you add the co-author + signed-off-by `Sankaranarayanan Venkatasubramanian sankaranarayanan.venkatasubramanian@intel.com`? He did https://github.com/gramineproject/gramine/pull/31/ originally.
___
*[CI-Examples/python/python.manifest.template, line 23 at r1](https://reviewable.io/reviews/gramineproject/gramine/497#-Mzy8doa2stvcI2wO-fm:-Mzy8doa2stvcI2wO-fn:b-33o5f4) ([raw file](https://github.com/gramineproject/gramine/blob/cd1f1d32f010c98e5e2248d3a743badb8d28db18/CI-Examples/python/python.manifest.template#L23)):*
> ```
> 
>   { type = "tmpfs", path = "/tmp" },
>   { type = "encrypted", path = "/enc", uri = "file:enc" },
> ```

Leftover from a debugging session?
___
*[common/src/protected_files/protected_files.c, line 1271 at r1](https://reviewable.io/reviews/gramineproject/gramine/497#-Mzy9WIk6VyVK7kuBuiK:-Mzy9WIk6VyVK7kuBuiL:b6f55i5) ([raw file](https://github.com/gramineproject/gramine/blob/cd1f1d32f010c98e5e2248d3a743badb8d28db18/common/src/protected_files/protected_files.c#L1271)):*
> ```C
>     memcpy(pf->encrypted_part_plain.path, new_path, new_path_size);
>     pf->need_writing = true;
> ```

Hm, what about calling `ipf_internal_flush()`? This was in the original PR https://github.com/gramineproject/gramine/pull/31.

Unless I'm missing something, we should do this flush operation. Otherwise another process/thread trying to open this file will still see the old `path` in the PF metadata and will fail (because the `new_path` of the file doesn't correspond to the old `path` in the file's metadata).
___
*[LibOS/shim/src/fs/shim_fs_encrypted.c, line 540 at r1](https://reviewable.io/reviews/gramineproject/gramine/497#-MzyB_PL8Q-FGoXBmc_K:-MzyB_PL8Q-FGoXBmc_L:b-oqv8p6) ([raw file](https://github.com/gramineproject/gramine/blob/cd1f1d32f010c98e5e2248d3a743badb8d28db18/LibOS/shim/src/fs/shim_fs_encrypted.c#L540)):*
> ```C
>     }
> 
>     ret = DkStreamChangeName(enc->pal_handle, new_uri);
> ```

Is it always guaranteed that we have a `enc->pal_handle`? I thought it was possible to have the `enc` object without the open PAL handle. Couldn't there be a situation when we want to rename an encrypted file that is not actively used?

Compare e.g. with your implementation of a classic `chroot_rename()` -- there you actually open a temporary PAL handle for the rename-operation sake.
___
*[LibOS/shim/src/fs/shim_fs_encrypted.c, line 555 at r1](https://reviewable.io/reviews/gramineproject/gramine/497#-MzyAfwd1Qyyc_t9L3Ai:-MzyAfwd1Qyyc_t9L3Aj:bm5ne88) ([raw file](https://github.com/gramineproject/gramine/blob/cd1f1d32f010c98e5e2248d3a743badb8d28db18/LibOS/shim/src/fs/shim_fs_encrypted.c#L555)):*
> ```C
>     }
> 
>     enc->uri = new_uri_copy;
> ```

What about the old `enc->uri`? You don't free it. Won't you have a memory leak?
___
*[LibOS/shim/src/fs/chroot/encrypted.c, line 284 at r1](https://reviewable.io/reviews/gramineproject/gramine/497#-MzyB6Ba2IvyZxvh0Ln2:-MzyB6Ba2IvyZxvh0Ln3:b-m54axm) ([raw file](https://github.com/gramineproject/gramine/blob/cd1f1d32f010c98e5e2248d3a743badb8d28db18/LibOS/shim/src/fs/chroot/encrypted.c#L284)):*
> ```C
>     ret = chroot_dentry_uri(new, old->inode->type, &new_uri);
>     if (ret < 0)
>         return 0;
> ```

What does this mean? Why do you return success?
___
*[LibOS/shim/test/regression/test_libos.py, line 606 at r1](https://reviewable.io/reviews/gramineproject/gramine/497#-MzyC78A5bTu1fdVrsTw:-MzyC78BEMy-ffbVpte6:bwjfttt) ([raw file](https://github.com/gramineproject/gramine/blob/cd1f1d32f010c98e5e2248d3a743badb8d28db18/LibOS/shim/test/regression/test_libos.py#L606)):*
> ```Python
>         self.assertIn('TEST OK', stdout)
> 
>     @unittest.skip('Protected files do not support renaming yet')
> ```

This comment is confusing now. I understand that here we mean "PFs as implemented in Linux-SGX PAL", and below we mean "PFs (Encrypted Files) as implemented in LibOS". Could you add some explanation to this comment, like `(as implemented in Linux-SGX PAL)`?

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/497)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/497#-MzyEanD5zx0je1APP8_:-MzyEanD5zx0je1APP8a:b-v4vgzt) (no related file):*
> it's not possible to rename a file "into" or "out of" encrypted files (since rename works only within a mounted filesystem)

Actually, isn't this a general problem that our rename works only within a single mount? The issue is that in a normal OS, a single mount is a huge beast covering the whole `/`, but in Gramine, a single mount is typically a sub-dir like `/etc` or `/tmp` or `<current-working-dir>`. So the unmodified application may try to `rename()` from e.g. `/tmp` into a `/etc`, which would work in a normal OS but will fail in Gramine -- and nothing could be done about it...

---

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/497)* status: 1 of 7 files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "squash! " and "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/497#-MzyEanD5zx0je1APP8_:-MzyXJw_2AXhms4Ew1ZP:b8f5c6i) (no related file):*

> it's not possible to rename a file "into" or "out of" encrypted files (since rename works only within a mounted filesystem)

Actually, isn't this a general problem that our rename works only within a single mount? The issue is that in a normal OS, a single mount is a huge beast covering the whole `/`, but in Gramine, a single mount is typically a sub-dir like `/etc` or `/tmp` or `<current-working-dir>`. So the unmodified application may try to `rename()` from e.g. `/tmp` into a `/etc`, which would work in a normal OS but will fail in Gramine -- and nothing could be done about it...
</blockquote></details>

Unfortunately I don't see a non-hacky way of doing such "cross-mount" renames, especially if we want to keep open handles for these files (which we support now), or do something more complicated like renaming directories (which we *don't* support yet).

But I'm not sure if this will be a common use-case: you would need two directories that are read-write for the application, *not* something like `/etc` or `/usr` which will be read-only. `/tmp` is also not a good example, since on normal systems it's usually also a separate mount.

---
___
*[-- commits, line 4 at r1](https://reviewable.io/reviews/gramineproject/gramine/497#-Mzy8jEF2qqa2rp23nLL):*

Could you add the co-author + signed-off-by `Sankaranarayanan Venkatasubramanian sankaranarayanan.venkatasubramanian@intel.com`? He did https://github.com/gramineproject/gramine/pull/31/ originally.
</blockquote></details>

OK. I added a squash commit.
___
*[common/src/protected_files/protected_files.c, line 1271 at r1](https://reviewable.io/reviews/gramineproject/gramine/497#-Mzy9WIk6VyVK7kuBuiK:-MzyZZN76VUkn_jpAuwd:b-36dz68) ([raw file](https://github.com/gramineproject/gramine/blob/cd1f1d32f010c98e5e2248d3a743badb8d28db18/common/src/protected_files/protected_files.c#L1271)):*

Hm, what about calling `ipf_internal_flush()`? This was in the original PR https://github.com/gramineproject/gramine/pull/31.

Unless I'm missing something, we should do this flush operation. Otherwise another process/thread trying to open this file will still see the old `path` in the PF metadata and will fail (because the `new_path` of the file doesn't correspond to the old `path` in the file's metadata).
</blockquote></details>

Yeah, I think it makes sense to flush here. Read/write operations do not flush, but this one is different because it can render the file unopenable.

Unfortunately there is still a window between updating the header and actually renaming the file, I don't think we can get rid of that. But this is generally the problem with protected files, concurrent accesses are not going to work well.

BTW: this only matters for multiple processes. Multiple threads use the same `pf` object, and take a lock.
___
*[LibOS/shim/src/fs/shim_fs_encrypted.c, line 540 at r1](https://reviewable.io/reviews/gramineproject/gramine/497#-MzyB_PL8Q-FGoXBmc_K:-MzyerSOAugYeWTF-wW6:bajidib) ([raw file](https://github.com/gramineproject/gramine/blob/cd1f1d32f010c98e5e2248d3a743badb8d28db18/LibOS/shim/src/fs/shim_fs_encrypted.c#L540)):*

Is it always guaranteed that we have a `enc->pal_handle`? I thought it was possible to have the `enc` object without the open PAL handle. Couldn't there be a situation when we want to rename an encrypted file that is not actively used?

Compare e.g. with your implementation of a classic `chroot_rename()` -- there you actually open a temporary PAL handle for the rename-operation sake.
</blockquote></details>

Yes, `enc` object doesn't have to be open, but all these functions (flush, read, write, rename...) operate on `enc` objects that are open. There are two states:
* open: `enc->pf` and `enc->pal_handle` present, `enc->use_count > 0`,
* closed: both NULL, `enc->use_count == 0`.

The caller (`chroot_encrypted_rename`) ensures the file is open (at least temporarily) by calling `encrypted_file_get/put`. It's done this way because we need to keep exactly 1 handle per file (so, unlike `chroot_rename`, we cannot create a separate one).
___
*[LibOS/shim/src/fs/shim_fs_encrypted.c, line 555 at r1](https://reviewable.io/reviews/gramineproject/gramine/497#-MzyAfwd1Qyyc_t9L3Ai:-MzygIs3BVqEUQlQaNDN:b532zhp) ([raw file](https://github.com/gramineproject/gramine/blob/cd1f1d32f010c98e5e2248d3a743badb8d28db18/LibOS/shim/src/fs/shim_fs_encrypted.c#L555)):*

What about the old `enc->uri`? You don't free it. Won't you have a memory leak?
</blockquote></details>

Right, I must have lost it in refactoring. Fixed.
___
*[LibOS/shim/src/fs/chroot/encrypted.c, line 284 at r1](https://reviewable.io/reviews/gramineproject/gramine/497#-MzyB6Ba2IvyZxvh0Ln2:-MzygPUd64SZsCXUeD6z:b-pp879s) ([raw file](https://github.com/gramineproject/gramine/blob/cd1f1d32f010c98e5e2248d3a743badb8d28db18/LibOS/shim/src/fs/chroot/encrypted.c#L284)):*

What does this mean? Why do you return success?
</blockquote></details>

Oh, that's a dumb typo... Fixed.
___
*[LibOS/shim/test/regression/test_libos.py, line 606 at r1](https://reviewable.io/reviews/gramineproject/gramine/497#-MzyC78A5bTu1fdVrsTw:-MzygcPl1Dg0bY8qJL42:b-ozffqi) ([raw file](https://github.com/gramineproject/gramine/blob/cd1f1d32f010c98e5e2248d3a743badb8d28db18/LibOS/shim/test/regression/test_libos.py#L606)):*

This comment is confusing now. I understand that here we mean "PFs as implemented in Linux-SGX PAL", and below we mean "PFs (Encrypted Files) as implemented in LibOS". Could you add some explanation to this comment, like `(as implemented in Linux-SGX PAL)`?
</blockquote></details>

I intend for the new feature to be consistently called "encrypted files", but right, there's also the library which is still called "protected files". I updated according to your suggestion.
___
*[CI-Examples/python/python.manifest.template, line 23 at r1](https://reviewable.io/reviews/gramineproject/gramine/497#-Mzy8doa2stvcI2wO-fm:-MzyZVvV2nMyOLl-lb-e:bgwt35a) ([raw file](https://github.com/gramineproject/gramine/blob/cd1f1d32f010c98e5e2248d3a743badb8d28db18/CI-Examples/python/python.manifest.template#L23)):*

Leftover from a debugging session?
</blockquote></details>

Yes. Fixed.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/497)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "squash! " and "fixup! " found in commit messages' one-liners

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/497#-MzyEanD5zx0je1APP8_:-MzyjqvZEmhnxEXjZy8g:b9912gm) (no related file):*
> But I'm not sure if this will be a common use-case: you would need two directories that are read-write for the application, not something like /etc or /usr which will be read-only. /tmp is also not a good example, since on normal systems it's usually also a separate mount.

Yes, I know, my examples suck. But you got my point. Anyway, let's finish this discussion, as you think that such support would be hacky.

---
___
*[common/src/protected_files/protected_files.c, line 1271 at r1](https://reviewable.io/reviews/gramineproject/gramine/497#-Mzy9WIk6VyVK7kuBuiK:-MzykGAnBx_SehbdUDIl:bqcr3nz) ([raw file](https://github.com/gramineproject/gramine/blob/cd1f1d32f010c98e5e2248d3a743badb8d28db18/common/src/protected_files/protected_files.c#L1271)):*
> Unfortunately there is still a window between updating the header and actually renaming the file, I don't think we can get rid of that. But this is generally the problem with protected files, concurrent accesses are not going to work well.

Yes, that's true.

NieDzejkob (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/497)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "squash! " and "fixup! " found in commit messages' one-liners

___
*[common/src/protected_files/protected_files.c, line 1271 at r1](https://reviewable.io/reviews/gramineproject/gramine/497#-Mzy9WIk6VyVK7kuBuiK:-N-1vDNU4J0v5b6jQfdA:b-ahbzby) ([raw file](https://github.com/gramineproject/gramine/blob/cd1f1d32f010c98e5e2248d3a743badb8d28db18/common/src/protected_files/protected_files.c#L1271)):*

> Unfortunately there is still a window between updating the header and actually renaming the file, I don't think we can get rid of that. But this is generally the problem with protected files, concurrent accesses are not going to work well.

Yes, that's true.
</blockquote></details>

I suppose we could fix that by starting a dedicated "file server" enclave, microkernel-style. You could even coordinate the enclaves such that a file being accessed by only one process is owned by that process and can be manipulated directly.

Not sure how much that'd affect performance, though. Seems like only open/close would suffer in the uncontended case? Such an implementation probably wouldn't be too straightforward, though.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/497)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "squash! " and "fixup! " found in commit messages' one-liners

___
*[common/src/protected_files/protected_files.c, line 1271 at r1](https://reviewable.io/reviews/gramineproject/gramine/497#-Mzy9WIk6VyVK7kuBuiK:-N-5OJ3dAoy80X5DNWjU:b-j9bnxc) ([raw file](https://github.com/gramineproject/gramine/blob/cd1f1d32f010c98e5e2248d3a743badb8d28db18/common/src/protected_files/protected_files.c#L1271)):*

I suppose we could fix that by starting a dedicated "file server" enclave, microkernel-style. You could even coordinate the enclaves such that a file being accessed by only one process is owned by that process and can be manipulated directly.

Not sure how much that'd affect performance, though. Seems like only open/close would suffer in the uncontended case? Such an implementation probably wouldn't be too straightforward, though.
</blockquote></details>

@niedzejkob It's a hard topic, @pwmarcz already tried to tackle this up, but decided to leave it for later. See https://github.com/gramineproject/graphene/pull/2267. He should be able to elaborate more about the technical problems with this.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/497)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "squash! " and "fixup! " found in commit messages' one-liners

___
*[common/src/protected_files/protected_files.c, line 1271 at r1](https://reviewable.io/reviews/gramineproject/gramine/497#-Mzy9WIk6VyVK7kuBuiK:-N-7j-xzCovLCSSEdEqq:bxno01q) ([raw file](https://github.com/gramineproject/gramine/blob/cd1f1d32f010c98e5e2248d3a743badb8d28db18/common/src/protected_files/protected_files.c#L1271)):*

@niedzejkob It's a hard topic, @pwmarcz already tried to tackle this up, but decided to leave it for later. See https://github.com/gramineproject/graphene/pull/2267. He should be able to elaborate more about the technical problems with this.
</blockquote></details>

Yes. The attempt linked by @mkow is actually more complicated: I wanted to have something that works locally for the uncontested case. See also https://github.com/oscarlab/graphene/issues/2158 for more background.

But the end result is that I'm really not proud of this: it ended up way too complicated, and is basically abandoned for now.

The "server" architecture you mentioned is done for POSIX locks implementation (fcntl); the main process works as a server and other processes always refer to it. It still needs some improvement (see https://github.com/oscarlab/graphene/issues/2517) but I think this is basically the way to go for many "multi-process" features.

Still, I wouldn't touch this until we have a good use case that we want to enable.

mkow (changes requested):
> Reviewed 1 of 7 files at r1, 6 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/497)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "squash! " and "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[common/src/protected_files/protected_files.h, line 264 at r2](https://reviewable.io/reviews/gramineproject/gramine/497#-N-9SfFT53ATFM7GKXvi:-N-9SfFT53ATFM7GKXvj:bcck2kg) ([raw file](https://github.com/gramineproject/gramine/blob/ed56556dd4f8e5d0ae1cd2419fe6dd14e966f7a9/common/src/protected_files/protected_files.h#L264)):*
> ```objc
> 
> /*!
>  * \brief Rename a PF
> ```

missing period
___
*[common/src/protected_files/protected_files.h, line 267 at r2](https://reviewable.io/reviews/gramineproject/gramine/497#-N-9SZyDC6A7EVvoPyEp:-N-9SZyDC6A7EVvoPyEq:bsomu2n) ([raw file](https://github.com/gramineproject/gramine/blob/ed56556dd4f8e5d0ae1cd2419fe6dd14e966f7a9/common/src/protected_files/protected_files.h#L267)):*
> ```objc
>  *
>  * \param [in] pf PF context
>  * \param [in] new_path New path
> ```

no space after `\param` + please align (basically, needs an update to our new, more consistent doxycomment style)

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/497)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "squash! " and "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[common/src/protected_files/protected_files.h, line 264 at r2](https://reviewable.io/reviews/gramineproject/gramine/497#-N-9SfFT53ATFM7GKXvi:-N-Mx7JFA2nBTC33emsi:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/ed56556dd4f8e5d0ae1cd2419fe6dd14e966f7a9/common/src/protected_files/protected_files.h#L264)):*

missing period
</blockquote></details>

Done.
___
*[common/src/protected_files/protected_files.h, line 267 at r2](https://reviewable.io/reviews/gramineproject/gramine/497#-N-9SZyDC6A7EVvoPyEp:-N-Mx7xh-0kQUp5VdH24:b-wccgd7) ([raw file](https://github.com/gramineproject/gramine/blob/ed56556dd4f8e5d0ae1cd2419fe6dd14e966f7a9/common/src/protected_files/protected_files.h#L267)):*

no space after `\param` + please align (basically, needs an update to our new, more consistent doxycomment style)
</blockquote></details>

I updated this whole comment to our new style (and omitted the `\return PF status` as useless, it's obvious from the return type).

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/497)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

dimakuv (approved):
> Reviewed 1 of 1 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/497)* status: :shipit: complete! all files reviewed, all discussions resolved

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/497)* status: :shipit: complete! all files reviewed, all discussions resolved

dimakuv:
@RodgerZhu Only the appended part will be encrypted. Everything shouldn't be re-encrypted. But because Protected Files use Merkle Hash Trees (MHT), some parts of the file will be re-hashed (but *not* re-encrypted).

PR #504, opened 2022-04-07
Support multiple keys in encrypted files

### LINKED ISSUE #502: [LibOS] Errors from writing to a pseudo-file are not propagated to user
## Description of the problem

When writing malformed data to pseudo-files in `/dev/attestation`, all operations will succeed.

## Steps to reproduce

Use the busybox example:

```
gramine-sgx busybox sh -c 'echo foo> /dev/attestation/protected_files_key'
```

## Expected results

`busybox` should complain at some point.

## Actual results

The operation succeeds. Here is the log from Gramine:

```
[P1:T1:busybox] trace: ---- shim_openat(AT_FDCWD, "/dev/attestation/protected_files_key", O_WRONLY|O_CREAT|O_TRUNC, 0666) = 0x3
[P1:T1:busybox] trace: ---- shim_fcntl(1, OP 1030, 0xa) = 0xa
[P1:T1:busybox] trace: ---- shim_dup2(3, 1) = 0x1
[P1:T1:busybox] trace: ---- shim_close(3) = 0x0
[P1:T1:busybox] trace: ---- shim_write(1, 0x16155410, 0x4) ...
[P1:T1:busybox] trace: ---- return from shim_write(...) = 0x4
[P1:T1:busybox] debug: /dev/attestation/protected_files_key: invalid length
[P1:T1:busybox] debug: pseudo_str_flush() failed, proceeding with close
[P1:T1:busybox] trace: ---- shim_dup2(10, 1) = 0x1
[P1:T1:busybox] trace: ---- shim_close(10) = 0x0
```

## Gramine commit hash

c56a3536b182021fa0655d9d55f3076ceced6fff

## Explanation

The pseudo-files are stored as in-memory strings. You can write to them, and flush/close operation will trigger the `save` callback (see `shim_fs_pseudo.c`).

This callback fails ("invalid length"), and the error code is propagated to the `close` callback, which returns it. However, our implementation of the close syscall only decreases the reference count on a handle, and **always returns 0**.

So the call chain is:

```
shim_do_close()  (returns 0)
put_handle()     (returns void, swallows error from pseudo_close())
pseudo_close()   (returns -EINVAL)
pfkey_save()     (returns -EINVAL)
```

## Possible solutions

* ~~Propagate error through the above chain~~: I don't think this works. `shim_do_close` might not actually end up closing the file (e.g. if there are multiple FDs).
* Always flush the file in `shim_do_close`: This will also work, but is pretty hacky, and some applications might disregard errors from `close`.
* **Change semantics of pseudo-files so that `write` will already process the value** (and return an error): I think this is a good choice, but it means that all such files (`protected_files_key`, `user_report_data`, `target_info`) need to be written using a single `write` syscall.

pwmarcz:
## Description of the changes 

This allows mounting encrypted files with `key = "..."` parameter. In addition, the keys can be now updated by application, by writing to `/dev/attestation/keys/<name>`.

## How to test this PR? 

The LibOS regression tests explicitly use `key = "custom"` now. There is also a test that checks if we can set a key, if errors are handled, and if the update persists after a fork.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/504)

### DIFF
--- a/LibOS/shim/include/shim_fs.h
+++ b/LibOS/shim/include/shim_fs.h
@@ -35,6 +35,9 @@ struct shim_mount_params {
 
     /* PAL URI, or NULL if not applicable */
     const char* uri;
+
+    /* Key name (used by `chroot_encrypted` filesystem), or NULL if not applicable */
+    const char* key_name;
 };
 
 struct shim_fs_ops {
--- a/LibOS/shim/include/shim_fs_encrypted.h
+++ b/LibOS/shim/include/shim_fs_encrypted.h
@@ -22,7 +22,7 @@
 
 /*
  * Represents a named key for opening files. The key might not be set yet: value of a key can be
- * specified in the manifest, or set using `set_encrypted_files_key`. Before the key is set,
+ * specified in the manifest, or set using `update_encrypted_files_key`. Before the key is set,
  * operations that use it will fail.
  */
 DEFINE_LIST(shim_encrypted_files_key);
@@ -62,21 +62,50 @@ int init_encrypted_files(void);
 /*
  * \brief Retrieve a key.
  *
- * Sets `*out_key` to a key with a given name. Note that the key might not be set yet (see
- * `struct shim_encrypted_files_key`).
+ * Returns a key with a given name, or NULL if it has not been created yet. Note that even if the
+ * key exists, it might not be set yet (see `struct shim_encrypted_files_key`).
  *
- * This does not pass ownership of `*out_key`: the key objects are still managed by this module.
+ * This does not pass ownership of the key: the key objects are still managed by this module.
  */
-int get_encrypted_files_key(const char* name, struct shim_encrypted_files_key** out_key);
+struct shim_encrypted_files_key* get_encrypted_files_key(const char* name);
+
+/*
+ * \brief List existing keys.
+ *
+ * Calls `callback` on each currently existing key.
+ */
+int list_encrypted_files_keys(int (*callback)(struct shim_encrypted_files_key* key, void* arg),
+                              void* arg);
+
+/*
+ * \brief Retrieve or create a key.
+ *
+ * Sets `*out_key` to a key with given name. If the key has not been created yet, creates a new one.
+ *
+ * Similar to `get_encrypted_files_key`, this does not pass ownership of `*out_key`.
+ */
+int get_or_create_encrypted_files_key(const char* name, struct shim_encrypted_files_key** out_key);
+
+/*
+ * \brief Read value of given key.
+ *
+ * \param      key     The key to read.
+ * \param[out] pf_key  On success, will be set to the current value.
+ *
+ * \returns `true` if the key has a value, `false` otherwise
+ *
+ * If the key has already been set, writes its value to `*pf_key` and returns `true`. Otherwise,
+ * returns `false`.
+ */
+bool read_encrypted_files_key(struct shim_encrypted_files_key* key, pf_key_t* pf_key);
 
 /*
  * \brief Update value of given key.
  *
- * \param key      The key to update.
- * \param key_str  New value for the key. Must be a 32-char null-terminated hex string (AES-GCM
- *                 encryption key).
+ * \param key     The key to update.
+ * \param pf_key  New value for the key.
  */
-int update_encrypted_files_key(struct shim_encrypted_files_key* key, char* key_str);
+void update_encrypted_files_key(struct shim_encrypted_files_key* key, const pf_key_t* pf_key);
 
 /*
  * \brief Open an existing encrypted file.
--- a/LibOS/shim/src/fs/chroot/encrypted.c
+++ b/LibOS/shim/src/fs/chroot/encrypted.c
@@ -26,7 +26,6 @@
  * TODO (most items are needed for feature parity with PAL protected files):
  *
  * - mounting with special keys (SGX MRENCLAVE and MRSIGNER)
- * - setting keys through /dev/attestation (and making sure they're copied to child process)
  * - mmap
  * - truncate (the current `truncate` operation works only for extending the file, support for
  *   truncation needs to be added to the `protected_files` module)
@@ -55,7 +54,7 @@ static int chroot_encrypted_mount(struct shim_mount_params* params, void** mount
     const char* key_name = params->key_name ?: "default";
 
     struct shim_encrypted_files_key* key;
-    int ret = get_encrypted_files_key(key_name, &key);
+    int ret = get_or_create_encrypted_files_key(key_name, &key);
     if (ret < 0)
         return ret;
 
@@ -76,7 +75,7 @@ static int chroot_encrypted_migrate(void* checkpoint, void** mount_data) {
     const char* name = checkpoint;
 
     struct shim_encrypted_files_key* key;
-    int ret = get_encrypted_files_key(name, &key);
+    int ret = get_or_create_encrypted_files_key(name, &key);
     if (ret < 0)
         return ret;
     *mount_data = key;
--- a/LibOS/shim/src/fs/chroot/encrypted.c
+++ b/LibOS/shim/src/fs/chroot/encrypted.c
@@ -25,7 +25,6 @@
  *
  * TODO (most items are needed for feature parity with PAL protected files):
  *
- * - mounting with other keys than "default"
  * - mounting with special keys (SGX MRENCLAVE and MRSIGNER)
  * - setting keys through /dev/attestation (and making sure they're copied to child process)
  * - mmap
@@ -53,8 +52,10 @@ static int chroot_encrypted_mount(struct shim_mount_params* params, void** mount
     if (!params->uri || !strstartswith(params->uri, URI_PREFIX_FILE))
         return -EINVAL;
 
+    const char* key_name = params->key_name ?: "default";
+
     struct shim_encrypted_files_key* key;
-    int ret = get_encrypted_files_key("default", &key);
+    int ret = get_encrypted_files_key(key_name, &key);
     if (ret < 0)
         return ret;
 
--- a/LibOS/shim/src/fs/dev/attestation.c
+++ b/LibOS/shim/src/fs/dev/attestation.c
@@ -15,6 +15,7 @@
  * be generally single-threaded and therefore do not introduce synchronization here.
  */
 
+#include "shim_fs_encrypted.h"
 #include "shim_fs_pseudo.h"
 
 /* user_report_data, target_info and quote are opaque blobs of predefined maximum sizes. Currently
@@ -281,29 +282,104 @@ static int pfkey_save(struct shim_dentry* dent, const char* data, size_t size) {
     return 0;
 }
 
-int init_attestation(struct pseudo_node* dev) {
-    if (strcmp(g_pal_public_state->host_type, "Linux-SGX")) {
-        log_debug("host is not Linux-SGX, skipping /dev/attestation setup");
-        return 0;
+static bool key_name_exists(struct shim_dentry* parent, const char* name) {
+    __UNUSED(parent);
+
+    struct shim_encrypted_files_key* key = get_encrypted_files_key(name);
+    return key != NULL;
+}
+
+struct key_list_names_data {
+    readdir_callback_t callback;
+    void* arg;
+};
+
+static int key_list_names_callback(struct shim_encrypted_files_key* key, void* arg) {
+    struct key_list_names_data* data = arg;
+    return data->callback(key->name, data->arg);
+}
+
+static int key_list_names(struct shim_dentry* parent, readdir_callback_t callback, void* arg) {
+    __UNUSED(parent);
+
+    struct key_list_names_data data = {
+        .callback = callback,
+        .arg = arg,
+    };
+    return list_encrypted_files_keys(&key_list_names_callback, &data);
+}
+
+static int key_load(struct shim_dentry* dent, char** out_data, size_t* out_size) {
+    struct shim_encrypted_files_key* key = get_encrypted_files_key(dent->name);
+    if (!key)
+        return -ENOENT;
+
+    pf_key_t pf_key;
+    bool is_set = read_encrypted_files_key(key, &pf_key);
+
+    if (is_set) {
+        char* buf = malloc(sizeof(pf_key));
+        if (!buf)
+            return -ENOMEM;
+        memcpy(buf, &pf_key, sizeof(pf_key));
+
+        *out_data = buf;
+        *out_size = sizeof(pf_key);
+    } else {
+        *out_data = NULL;
+        *out_size = 0;
+    }
+    return 0;
+}
+
+static int key_save(struct shim_dentry* dent, const char* data, size_t size) {
+    struct shim_encrypted_files_key* key = get_encrypted_files_key(dent->name);
+    if (!key)
+        return -ENOENT;
+
+    pf_key_t pf_key;
+    if (size != sizeof(pf_key)) {
+        log_debug("/dev/attestation/keys: invalid length");
+        return -EACCES;
     }
 
+    memcpy(&pf_key, data, sizeof(pf_key));
+    update_encrypted_files_key(key, &pf_key);
+    return 0;
+}
+
+
+int init_attestation(struct pseudo_node* dev) {
     struct pseudo_node* attestation = pseudo_add_dir(dev, "attestation");
 
-    struct pseudo_node* user_report_data = pseudo_add_str(attestation, "user_report_data", NULL);
-    user_report_data->perm = PSEUDO_PERM_FILE_RW;
-    user_report_data->str.save = &user_report_data_save;
+    if (!strcmp(g_pal_public_state->host_type, "Linux-SGX")) {
+        log_debug("host is Linux-SGX, adding SGX-specific /dev/attestation files: "
+                  "report, quote, etc.");
 
-    struct pseudo_node* target_info = pseudo_add_str(attestation, "target_info", NULL);
-    target_info->perm = PSEUDO_PERM_FILE_RW;
-    target_info->str.save = &target_info_save;
+        struct pseudo_node* user_report_data = pseudo_add_str(attestation, "user_report_data", NULL);
+        user_report_data->perm = PSEUDO_PERM_FILE_RW;
+        user_report_data->str.save = &user_report_data_save;
 
-    pseudo_add_str(attestation, "my_target_info", &my_target_info_load);
-    pseudo_add_str(attestation, "report", &report_load);
-    pseudo_add_str(attestation, "quote", &quote_load);
+        struct pseudo_node* target_info = pseudo_add_str(attestation, "target_info", NULL);
+        target_info->perm = PSEUDO_PERM_FILE_RW;
+        target_info->str.save = &target_info_save;
 
-    struct pseudo_node* pfkey = pseudo_add_str(attestation, "protected_files_key",
-                                               &pfkey_load);
-    pfkey->perm = PSEUDO_PERM_FILE_RW;
-    pfkey->str.save = &pfkey_save;
+        pseudo_add_str(attestation, "my_target_info", &my_target_info_load);
+        pseudo_add_str(attestation, "report", &report_load);
+        pseudo_add_str(attestation, "quote", &quote_load);
+
+        struct pseudo_node* pfkey = pseudo_add_str(attestation, "protected_files_key",
+                                                   &pfkey_load);
+        pfkey->perm = PSEUDO_PERM_FILE_RW;
+        pfkey->str.save = &pfkey_save;
+    }
+
+    struct pseudo_node* keys = pseudo_add_dir(attestation, "keys");
+    struct pseudo_node* key = pseudo_add_str(keys, /*name=*/NULL, &key_load);
+    key->name_exists = &key_name_exists;
+    key->list_names = &key_list_names;
+    key->perm = PSEUDO_PERM_FILE_RW;
+    key->str.save = &key_save;
+
     return 0;
 }
--- a/LibOS/shim/src/fs/shim_fs.c
+++ b/LibOS/shim/src/fs/shim_fs.c
@@ -93,8 +93,9 @@ static bool mount_migrated = false;
 
 static int mount_root(void) {
     int ret;
-    char* fs_root_type = NULL;
-    char* fs_root_uri  = NULL;
+    char* fs_root_type     = NULL;
+    char* fs_root_uri      = NULL;
+    char* fs_root_key_name = NULL;
 
     assert(g_manifest_root);
 
@@ -112,7 +113,17 @@ static int mount_root(void) {
         goto out;
     }
 
-    struct shim_mount_params params = { .path = "/" };
+    ret = toml_string_in(g_manifest_root, "fs.root.key_name", &fs_root_key_name);
+    if (ret < 0) {
+        log_error("Cannot parse 'fs.root.key_name'");
+        ret = -EINVAL;
+        goto out;
+    }
+
+    struct shim_mount_params params = {
+        .path = "/",
+        .key_name = fs_root_key_name,
+    };
 
     if (!fs_root_type && !fs_root_uri) {
         params.type = "chroot";
@@ -182,9 +193,10 @@ static int mount_one_nonroot(toml_table_t* mount, const char* prefix) {
 
     int ret;
 
-    char* mount_type = NULL;
-    char* mount_path = NULL;
-    char* mount_uri  = NULL;
+    char* mount_type     = NULL;
+    char* mount_path     = NULL;
+    char* mount_uri      = NULL;
+    char* mount_key_name = NULL;
 
     ret = toml_string_in(mount, "type", &mount_type);
     if (ret < 0) {
@@ -203,6 +215,13 @@ static int mount_one_nonroot(toml_table_t* mount, const char* prefix) {
     ret = toml_string_in(mount, "uri", &mount_uri);
     if (ret < 0) {
         log_error("Cannot parse '%s.uri'", prefix);
+        ret = -EINVAL;
+        goto out;
+    }
+
+    ret = toml_string_in(mount, "key_name", &mount_key_name);
+    if (ret < 0) {
+        log_error("Cannot parse '%s.key_name'", prefix);
         ret = -EINVAL;
         goto out;
     }
@@ -256,6 +275,7 @@ static int mount_one_nonroot(toml_table_t* mount, const char* prefix) {
         .type = mount_type ?: "chroot",
         .path = mount_path,
         .uri = mount_uri,
+        .key_name = mount_key_name,
     };
     ret = mount_fs(&params);
 
@@ -263,6 +283,7 @@ out:
     free(mount_type);
     free(mount_path);
     free(mount_uri);
+    free(mount_key_name);
     return ret;
 }
 
--- a/LibOS/shim/src/fs/shim_fs_encrypted.c
+++ b/LibOS/shim/src/fs/shim_fs_encrypted.c
@@ -196,6 +196,28 @@ static int encrypted_file_internal_open(struct shim_encrypted_file* enc, PAL_HAN
     return 0;
 }
 
+static int parse_pf_key(const char* key_str, pf_key_t* pf_key) {
+    size_t len = strlen(key_str);
+    if (len != PF_KEY_SIZE * 2) {
+        log_warning("%s: wrong key length (%zu instead of %zu)", __func__, len,
+                    (size_t)(PF_KEY_SIZE * 2));
+        return -1;
+    }
+
+    pf_key_t tmp_pf_key;
+    for (size_t i = 0; i < len; i += 2) {
+        int8_t hi = hex2dec(key_str[i]);
+        int8_t lo = hex2dec(key_str[i+1]);
+        if (hi < 0 || lo < 0) {
+            log_warning("%s: unexpected character encountered", __func__);
+            return -1;
+        }
+        tmp_pf_key[i / 2] = hi * 16 + lo;
+    }
+    memcpy(pf_key, &tmp_pf_key, sizeof(tmp_pf_key));
+    return 0;
+}
+
 static void encrypted_file_internal_close(struct shim_encrypted_file* enc) {
     assert(enc->pf);
 
@@ -244,7 +266,7 @@ int init_encrypted_files(void) {
         assert(name);
 
         struct shim_encrypted_files_key* key;
-        ret = get_encrypted_files_key(name, &key);
+        ret = get_or_create_encrypted_files_key(name, &key);
         if (ret < 0)
             goto out;
 
@@ -256,12 +278,15 @@ int init_encrypted_files(void) {
         }
         assert(key_str);
 
-        ret = update_encrypted_files_key(key, key_str);
+        pf_key_t pf_key;
+        ret = parse_pf_key(key_str, &pf_key);
         if (ret < 0) {
-            log_error("Cannot update key '%s': %d", name, ret);
+            log_error("Cannot parse key 'fs.insecure__keys.%s' as a hex key", name);
+            ret = -EINVAL;
             goto out;
-
         }
+
+        update_encrypted_files_key(key, &pf_key);
         free(key_str);
         key_str = NULL;
     }
@@ -273,7 +298,7 @@ out:
     return ret;
 }
 
-static struct shim_encrypted_files_key* get_or_create_key(const char* name) {
+static struct shim_encrypted_files_key* get_key(const char* name) {
     assert(locked(&g_keys_lock));
 
     struct shim_encrypted_files_key* key;
@@ -282,6 +307,16 @@ static struct shim_encrypted_files_key* get_or_create_key(const char* name) {
             return key;
         }
     }
+
+    return NULL;
+}
+
+static struct shim_encrypted_files_key* get_or_create_key(const char* name) {
+    assert(locked(&g_keys_lock));
+
+    struct shim_encrypted_files_key* key = get_key(name);
+    if (key)
+        return key;
 
     key = calloc(1, sizeof(*key));
     if (!key)
@@ -296,50 +331,66 @@ static struct shim_encrypted_files_key* get_or_create_key(const char* name) {
     return key;
 }
 
-int get_encrypted_files_key(const char* name, struct shim_encrypted_files_key** out_key) {
+struct shim_encrypted_files_key* get_encrypted_files_key(const char* name) {
     lock(&g_keys_lock);
-
-    struct shim_encrypted_files_key* key = get_or_create_key(name);
-    if (!key)
-        return -ENOMEM;
-
-    /* TODO: load special keys (MRENCLAVE, MRSIGNER) here */
-
-    *out_key = key;
-
+    struct shim_encrypted_files_key* key = get_key(name);
     unlock(&g_keys_lock);
-    return 0;
+    return key;
 }
 
-int update_encrypted_files_key(struct shim_encrypted_files_key* key, char* key_str) {
+int list_encrypted_files_keys(int (*callback)(struct shim_encrypted_files_key* key, void* arg),
+                              void* arg) {
+    lock(&g_keys_lock);
+
     int ret;
 
-    lock(&g_keys_lock);
-    size_t len = strlen(key_str);
-    if (len != PF_KEY_SIZE * 2) {
-        log_warning("%s: wrong key length (%zu instead of %zu)", __func__, len,
-                    (size_t)(PF_KEY_SIZE * 2));
-        ret = -EINVAL;
-        goto out;
-    }
-
-    pf_key_t pf_key;
-    for (size_t i = 0; i < len; i += 2) {
-        int8_t hi = hex2dec(key_str[i]);
-        int8_t lo = hex2dec(key_str[i+1]);
-        if (hi < 0 || lo < 0) {
-            log_warning("%s: unexpected character encountered", __func__);
-            ret = -EINVAL;
+    struct shim_encrypted_files_key* key;
+    LISTP_FOR_EACH_ENTRY(key, &g_keys, list) {
+        ret = callback(key, arg);
+        if (ret < 0)
             goto out;
-        }
-        pf_key[i / 2] = hi * 16 + lo;
     }
-    memcpy(key->pf_key, &pf_key, sizeof(pf_key));
-    key->is_set = true;
     ret = 0;
 out:
     unlock(&g_keys_lock);
     return ret;
+}
+
+int get_or_create_encrypted_files_key(const char* name, struct shim_encrypted_files_key** out_key) {
+    lock(&g_keys_lock);
+
+    int ret;
+
+    struct shim_encrypted_files_key* key = get_or_create_key(name);
+    if (!key) {
+        ret = -ENOMEM;
+        goto out;
+    }
+
+    /* TODO: load special keys (MRENCLAVE, MRSIGNER) here */
+
+    *out_key = key;
+    ret = 0;
+out:
+    unlock(&g_keys_lock);
+    return ret;
+}
+
+bool read_encrypted_files_key(struct shim_encrypted_files_key* key, pf_key_t* pf_key) {
+    lock(&g_keys_lock);
+    bool is_set = key->is_set;
+    if (is_set) {
+        memcpy(pf_key, &key->pf_key, sizeof(key->pf_key));
+    }
+    unlock(&g_keys_lock);
+    return is_set;
+}
+
+void update_encrypted_files_key(struct shim_encrypted_files_key* key, const pf_key_t* pf_key) {
+    lock(&g_keys_lock);
+    memcpy(&key->pf_key, pf_key, sizeof(*pf_key));
+    key->is_set = true;
+    unlock(&g_keys_lock);
 }
 
 static int encrypted_file_alloc(const char* uri, struct shim_encrypted_files_key* key,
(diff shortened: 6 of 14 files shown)


mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/504)* status: 0 of 18 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/504#-N-5T7vp7l3ekvSkVAUs:-N-5T7vqCE11UT40uvLc:b5r2e6s) (no related file):*
> This PR also fixes #502 (so that the above update has correct error handling).

Could you split this out into a separate PR? We could merge it quickly, as we already agree on the solution and it's not too much code. And the feature implemented in the rest of this PR is non-trivial, so it may take a while to review.

---

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/504)* status: 0 of 18 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

___
*[-- commits, line 26 at r1](https://reviewable.io/reviews/gramineproject/gramine/504#-N-7NTXbAYnYSdGiVai-):*
Actually, this commit can be also moved to a separate PR (where `struct shim_mount_params` doesn't have the field `key`).

This would be simpler to review. I'm not insisting though.
___
*[LibOS/shim/test/regression/tests.toml, line 53 at r1](https://reviewable.io/reviews/gramineproject/gramine/504#-N-7NCgZ7EmQmx1SidGq:-N-7NCgZ7EmQmx1SidGr:bs8zt4c) ([raw file](https://github.com/gramineproject/gramine/blob/03dae25b06441ccc205243737dab841d2c379345/LibOS/shim/test/regression/tests.toml#L53)):*
> ```TOML
>   "host_root_fs",
>   "init_fail",
>   "keys",
> ```

Forgot about musl-based tests?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/504)* status: 0 of 19 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/504#-N-5T7vp7l3ekvSkVAUs:-N-7Wftk2bcYUcCnvuuE:b8trxsf) (no related file):*

> This PR also fixes #502 (so that the above update has correct error handling).

Could you split this out into a separate PR? We could merge it quickly, as we already agree on the solution and it's not too much code. And the feature implemented in the rest of this PR is non-trivial, so it may take a while to review.
</blockquote></details>

Done: https://github.com/gramineproject/gramine/pull/508

I'll remove this commit when the other one is merged.

---
___
*[-- commits, line 26 at r1](https://reviewable.io/reviews/gramineproject/gramine/504#-N-7NTXbAYnYSdGiVai-):*

Actually, this commit can be also moved to a separate PR (where `struct shim_mount_params` doesn't have the field `key`).

This would be simpler to review. I'm not insisting though.
</blockquote></details>

Done: https://github.com/gramineproject/gramine/pull/508

I'll remove this commit when the other one is merged.
___
*[LibOS/shim/test/regression/tests.toml, line 53 at r1](https://reviewable.io/reviews/gramineproject/gramine/504#-N-7NCgZ7EmQmx1SidGq:-N-7ZsuMEOD8fTbKSgHZ:b-3d4zvp) ([raw file](https://github.com/gramineproject/gramine/blob/03dae25b06441ccc205243737dab841d2c379345/LibOS/shim/test/regression/tests.toml#L53)):*

Forgot about musl-based tests?
</blockquote></details>

Yes, I did :) This is fixed now.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/504)* status: 0 of 19 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[-- commits, line 26 at r1](https://reviewable.io/reviews/gramineproject/gramine/504#-N-7NTXbAYnYSdGiVai-):*

Done: https://github.com/gramineproject/gramine/pull/508

I'll remove this commit when the other one is merged.
</blockquote></details>

I don't see the change. And why do you move it to the already-existing PR #508? Note that I'm talking about a different commit than Michal meant.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/504)* status: 0 of 19 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow and @pwmarcz)

___
*[-- commits, line 26 at r1](https://reviewable.io/reviews/gramineproject/gramine/504#-N-7NTXbAYnYSdGiVai-):*

I don't see the change. And why do you move it to the already-existing PR #508? Note that I'm talking about a different commit than Michal meant.
</blockquote></details>

Oh, sorry, that's a different one.

OK, I'll split this as well :)

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/504)* status: 0 of 19 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow and @pwmarcz)

___
*[-- commits, line 26 at r1](https://reviewable.io/reviews/gramineproject/gramine/504#-N-7NTXbAYnYSdGiVai-):*

Oh, sorry, that's a different one.

OK, I'll split this as well :)
</blockquote></details>

Submitted as #510.

I'll rebase this PR after these two PRs are.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/504)* status: 0 of 19 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow and @pwmarcz)

___
*[-- commits, line 26 at r1](https://reviewable.io/reviews/gramineproject/gramine/504#-N-7NTXbAYnYSdGiVai-):*

Submitted as #510.

I'll rebase this PR after these two PRs are.
</blockquote></details>

*are merged.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/504)* status: 0 of 19 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @mkow and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/504#-N-5T7vp7l3ekvSkVAUs:-N-N_bxb4WXVv6pucJ39:b-ljdtp8) (no related file):*

Done: https://github.com/gramineproject/gramine/pull/508

I'll remove this commit when the other one is merged.
</blockquote></details>

That's merged now, and I removed the commit from this PR.

---
___
*[-- commits, line 26 at r1](https://reviewable.io/reviews/gramineproject/gramine/504#-N-7NTXbAYnYSdGiVai-):*

*are merged.
</blockquote></details>

That PR is merged now, and this commit only adds the `key` parameter now.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/504)* status: 5 of 19 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/504#-N-5T7vp7l3ekvSkVAUs:-N-NwvoaD-2jQFNO0x4F:bdp6ddu) (no related file):*

Please also update the description of this PR.
</blockquote></details>

Done (I removed the part about fixing #502, since the fix was merged separately).

---
___
*[LibOS/shim/test/regression/manifest.template, line 20 at r3](https://reviewable.io/reviews/gramineproject/gramine/504#-N-NbiWi7PfsitEafC8C:-N-Nxiod76rf8miY-At8:b-atoxqx) ([raw file](https://github.com/gramineproject/gramine/blob/0b9d1bd9846d4f27fe97db9398720e139761c6cd/LibOS/shim/test/regression/manifest.template#L20)):*

I'd rather use something like `some_custom_name` to clearly indicate to the users that these are arbitrary, not some magic names with special handling in Gramine.
</blockquote></details>

Renamed to `my_custom_key`, I hope that looks sufficiently non-built-in.

dimakuv (changes requested):
> Reviewed 5 of 18 files at r1, 1 of 1 files at r2, 6 of 11 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/504)* status: all files reviewed, 14 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow and @pwmarcz)

___
*[LibOS/shim/include/shim_fs_encrypted.h, line 67 at r4](https://reviewable.io/reviews/gramineproject/gramine/504#-N-RaL8h14kOe4ZUjlHp:-N-RaL8h14kOe4ZUjlHq:b1izri3) ([raw file](https://github.com/gramineproject/gramine/blob/4ace5ae7b8b37ec43d2b43e8ee4b04583e24687c/LibOS/shim/include/shim_fs_encrypted.h#L67)):*
> ```objc
>  * \brief Retrieve a key.
>  *
>  * Returns a key with a given name, or NULL if it has not been created yet. Note that the even if
> ```

`Note that the even` -> `Note that even`
___
*[LibOS/shim/include/shim_fs_encrypted.h, line 68 at r4](https://reviewable.io/reviews/gramineproject/gramine/504#-N-RbJhu3Ux63xO-dIFM:-N-RbJhu3Ux63xO-dIFN:b-e5i4ri) ([raw file](https://github.com/gramineproject/gramine/blob/e567372e1510a2636c4cbd966b9f81033e0f47af/LibOS/shim/include/shim_fs_encrypted.h#L68)):*
> ```objc
>  * `struct shim_encrypted_files_key`).
>  *
>  * This does not pass ownership of `*out_key`: the key objects are still managed by this module.
> ```

Isn't this comment important to keep?
___
*[LibOS/shim/src/fs/shim_fs_encrypted.c, line 324 at r4](https://reviewable.io/reviews/gramineproject/gramine/504#-N-RcKad6o0Xa_DQp8Vj:-N-RcKad6o0Xa_DQp8Vk:b-lagd3h) ([raw file](https://github.com/gramineproject/gramine/blob/4ace5ae7b8b37ec43d2b43e8ee4b04583e24687c/LibOS/shim/src/fs/shim_fs_encrypted.c#L324)):*
> ```C
>         int ret = callback(key, arg);
>         if (ret < 0)
>             return ret;
> ```

You should unlock first.
___
*[LibOS/shim/src/fs/shim_fs_encrypted.c, line 667 at r4](https://reviewable.io/reviews/gramineproject/gramine/504#-N-Rehf4DkCHk1heGa4E:-N-Rehf4DkCHk1heGa4F:b5qgueq) ([raw file](https://github.com/gramineproject/gramine/blob/4ace5ae7b8b37ec43d2b43e8ee4b04583e24687c/LibOS/shim/src/fs/shim_fs_encrypted.c#L667)):*
> ```C
> 
>     struct shim_encrypted_files_key* key;
>     int ret = get_or_create_encrypted_files_key(migrated_key->name, &key);
> ```

So instead of using the inside-checkpoint-blob `migrated_key` object, you're creating a new `key` object, add it to the list anew, and copy the contents from `migrated_key`. This is on purpose, right?
___
*[LibOS/shim/src/fs/dev/attestation.c, line 16 at r4](https://reviewable.io/reviews/gramineproject/gramine/504#-N-Rh6Gg29tkeW9UjKdW:-N-Rh6Gg29tkeW9UjKdX:b5vhyn5) ([raw file](https://github.com/gramineproject/gramine/blob/4ace5ae7b8b37ec43d2b43e8ee4b04583e24687c/LibOS/shim/src/fs/dev/attestation.c#L16)):*
> ```C
>  * correctly synchronize concurrent accesses to the pseudo-files. We expect attestation flows to
>  * be generally single-threaded and therefore do not introduce synchronization here.
>  */
> ```

Maybe add some TODO to remove:
- `g_pf_key_hex`
- `pfkey_load`
- `pfkey_save`
___
*[LibOS/shim/src/fs/dev/attestation.c, line 299 at r4](https://reviewable.io/reviews/gramineproject/gramine/504#-N-Rg1xz8DaixWxIZv4G:-N-Rg1y-DPqjiR2hEIMn:b4vz34e) ([raw file](https://github.com/gramineproject/gramine/blob/4ace5ae7b8b37ec43d2b43e8ee4b04583e24687c/LibOS/shim/src/fs/dev/attestation.c#L299)):*
> ```C
> static int key_list_names_callback(struct shim_encrypted_files_key* key, void* arg) {
>     struct key_list_names_data* data = arg;
>     return data->callback(key->name, data->arg);
> ```

Why so complicated? You now have a callback that calls another callback. Can't you just specify `list_encrypted_files_keys(readdir_callback_t callback, ...)` as a function signature and use a single indirection?
___
*[LibOS/shim/src/fs/dev/attestation.c, line 339 at r4](https://reviewable.io/reviews/gramineproject/gramine/504#-N-RhPwOAF1_CJOdaHrF:-N-RhPwOAF1_CJOdaHrG:b-g7el06) ([raw file](https://github.com/gramineproject/gramine/blob/4ace5ae7b8b37ec43d2b43e8ee4b04583e24687c/LibOS/shim/src/fs/dev/attestation.c#L339)):*
> ```C
> 
>     if (size != KEY_HEX_LEN) {
>         log_debug("/dev/attestation/protected_files_key: invalid length");
> ```

That's actually `/dev/attestation/keys/`
___
*[LibOS/shim/src/fs/dev/attestation.c, line 360 at r4](https://reviewable.io/reviews/gramineproject/gramine/504#-N-Rk0SQ9DzKssS3Su-v:-N-Rk0SQ9DzKssS3Su-w:bc8lc2x) ([raw file](https://github.com/gramineproject/gramine/blob/4ace5ae7b8b37ec43d2b43e8ee4b04583e24687c/LibOS/shim/src/fs/dev/attestation.c#L360)):*
> ```C
> 
>     if (!strcmp(g_pal_public_state->host_type, "Linux-SGX")) {
>         struct pseudo_node* user_report_data = pseudo_add_str(attestation, "user_report_data", NULL);
> ```

Can we add some `log_debug()` message here? Would be nice to tell something like
```
host is Linux-SGX, adding SGX-specific attestation files: report, quote, etc.
```
___
*[LibOS/shim/test/regression/keys.c, line 18 at r4](https://reviewable.io/reviews/gramineproject/gramine/504#-N-Ris2Y5bxoI6H8wAMA:-N-Ris2Y5bxoI6H8wAMB:bqoy2j5) ([raw file](https://github.com/gramineproject/gramine/blob/4ace5ae7b8b37ec43d2b43e8ee4b04583e24687c/LibOS/shim/test/regression/keys.c#L18)):*
> ```C
> #include "rw_file.h"
> 
> #define KEY_PATH "/dev/attestation/keys/my_custom_key"
> ```

Can you add a comment to look for this key and its original value in `manifest.template`?
___
*[LibOS/shim/test/regression/keys.c, line 36 at r4](https://reviewable.io/reviews/gramineproject/gramine/504#-N-RiaV72mfh9Jdhs2sk:-N-RiaV72mfh9Jdhs2sl:b-z4b53z) ([raw file](https://github.com/gramineproject/gramine/blob/4ace5ae7b8b37ec43d2b43e8ee4b04583e24687c/LibOS/shim/test/regression/keys.c#L36)):*
> ```C
> 
>     if (strcmp(key, expected_key)) {
>         errx(1, "%s: wrong key: expected %s, got %s", desc, key, expected_key);
> ```

Reorder `key` and `expected_key` arguments -- your string wants them in different order
___
*[LibOS/shim/test/regression/manifest.template, line 47 at r4](https://reviewable.io/reviews/gramineproject/gramine/504#-N-RjUAG3yEr5ounfyal:-N-RjUAG3yEr5ounfyam:bj2riut) ([raw file](https://github.com/gramineproject/gramine/blob/4ace5ae7b8b37ec43d2b43e8ee4b04583e24687c/LibOS/shim/test/regression/manifest.template#L47)):*
> ```
> ]
> 
> fs.insecure__keys.my_custom_key = "00112233445566778899aabbccddeeff"
> ```

What about we also specify the `default` key, and augment the test above to check it as well?

This way, we'll additionally test two things:
- The linked list of keys
- The checkpoint-restore of even those keys that didn't have the associated file (`/dev/attestation/keys/default`) opened/written to.
___
*[LibOS/shim/test/regression/manifest.template, line 47 at r4](https://reviewable.io/reviews/gramineproject/gramine/504#-N-Rjr0w8jnE-R__rBUj:-N-Rjr0w8jnE-R__rBUk:bhxk7na) ([raw file](https://github.com/gramineproject/gramine/blob/4ace5ae7b8b37ec43d2b43e8ee4b04583e24687c/LibOS/shim/test/regression/manifest.template#L47)):*
> ```
> ]
> 
> fs.insecure__keys.my_custom_key = "00112233445566778899aabbccddeeff"
> ```

Maybe also add a comment to `see the keys.c test`?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/504)* status: 14 of 19 files reviewed, 14 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[LibOS/shim/include/shim_fs_encrypted.h, line 67 at r4](https://reviewable.io/reviews/gramineproject/gramine/504#-N-RaL8h14kOe4ZUjlHp:-N-RlYEE7VUHoNSK0ph9:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/4ace5ae7b8b37ec43d2b43e8ee4b04583e24687c/LibOS/shim/include/shim_fs_encrypted.h#L67)):*

`Note that the even` -> `Note that even`
</blockquote></details>

Done.
___
*[LibOS/shim/include/shim_fs_encrypted.h, line 68 at r4](https://reviewable.io/reviews/gramineproject/gramine/504#-N-RbJhu3Ux63xO-dIFM:-N-RluJh43Ei2htZMZLy:beep1e7) ([raw file](https://github.com/gramineproject/gramine/blob/e567372e1510a2636c4cbd966b9f81033e0f47af/LibOS/shim/include/shim_fs_encrypted.h#L68)):*

Isn't this comment important to keep?
</blockquote></details>

Right. Restored.
___
*[LibOS/shim/src/fs/shim_fs_encrypted.c, line 324 at r4](https://reviewable.io/reviews/gramineproject/gramine/504#-N-RcKad6o0Xa_DQp8Vj:-N-Rm6Sq1p5oFuK8zJuo:blgeayk) ([raw file](https://github.com/gramineproject/gramine/blob/4ace5ae7b8b37ec43d2b43e8ee4b04583e24687c/LibOS/shim/src/fs/shim_fs_encrypted.c#L324)):*

You should unlock first.
</blockquote></details>

Ouch. Fixed.
___
*[LibOS/shim/src/fs/shim_fs_encrypted.c, line 667 at r4](https://reviewable.io/reviews/gramineproject/gramine/504#-N-Rehf4DkCHk1heGa4E:-N-Rn7mN4ruVnK5QehYN:b-98meve) ([raw file](https://github.com/gramineproject/gramine/blob/4ace5ae7b8b37ec43d2b43e8ee4b04583e24687c/LibOS/shim/src/fs/shim_fs_encrypted.c#L667)):*

So instead of using the inside-checkpoint-blob `migrated_key` object, you're creating a new `key` object, add it to the list anew, and copy the contents from `migrated_key`. This is on purpose, right?
</blockquote></details>

Yes, this is on purpose. I think it deserves a comment - I added one.
___
*[LibOS/shim/src/fs/dev/attestation.c, line 16 at r4](https://reviewable.io/reviews/gramineproject/gramine/504#-N-Rh6Gg29tkeW9UjKdW:-N-RnSSe9Jf9PD99Fdce:bovmuaa) ([raw file](https://github.com/gramineproject/gramine/blob/4ace5ae7b8b37ec43d2b43e8ee4b04583e24687c/LibOS/shim/src/fs/dev/attestation.c#L16)):*

Maybe add some TODO to remove:
- `g_pf_key_hex`
- `pfkey_load`
- `pfkey_save`
</blockquote></details>

Isn't it too early for that? The current system ("protected files") is not even deprecated yet.

I agree that these need to be removed eventually, but first we'll need to switch all these APIs to use encrypted keys (e.g. `pfkey_load` will call `key_load` for the `"default"` key... or even the file could become a symlink), and document `type = "encrypted"` to be the right way of using the feature.

Then, I think, will be the right time to start the deprecation period, and add TODOs to delete all of these compatibility shims.
___
*[LibOS/shim/src/fs/dev/attestation.c, line 299 at r4](https://reviewable.io/reviews/gramineproject/gramine/504#-N-Rg1xz8DaixWxIZv4G:-N-RojZCE9UPJC92ycU6:b-lpfr01) ([raw file](https://github.com/gramineproject/gramine/blob/4ace5ae7b8b37ec43d2b43e8ee4b04583e24687c/LibOS/shim/src/fs/dev/attestation.c#L299)):*

Why so complicated? You now have a callback that calls another callback. Can't you just specify `list_encrypted_files_keys(readdir_callback_t callback, ...)` as a function signature and use a single indirection?
</blockquote></details>

I regret the boilerplate, but I'd rather keep the module (`shim_fs_encrypted.[ch]`) independent from the rest of LibOS, and using the `readdir_callback_t` type breaks that.

I guess I could say that `list_encrypted_files_keys` actually works on *names*, and declare it to take a callback of form `int (*callback)(const char* name, void* arg)`, which happens to be the same type as `readdir_callback_t`. The types would be even enforced by compiler.

But still, I don't think this would be a good situation, the coupling is essentially still there: if we ever change how `readdir` works, now we also have to either reach into `shim_fs_encrypted.[ch]` and change it as well, or rewrite the code here.
___
*[LibOS/shim/src/fs/dev/attestation.c, line 339 at r4](https://reviewable.io/reviews/gramineproject/gramine/504#-N-RhPwOAF1_CJOdaHrF:-N-RrI6ICEfOs6-5wUXX:bidrs83) ([raw file](https://github.com/gramineproject/gramine/blob/4ace5ae7b8b37ec43d2b43e8ee4b04583e24687c/LibOS/shim/src/fs/dev/attestation.c#L339)):*

That's actually `/dev/attestation/keys/`
</blockquote></details>

Right. Fixed.
___
*[LibOS/shim/src/fs/dev/attestation.c, line 360 at r4](https://reviewable.io/reviews/gramineproject/gramine/504#-N-Rk0SQ9DzKssS3Su-v:-N-RrUZa1IddK1fqsaGe:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/4ace5ae7b8b37ec43d2b43e8ee4b04583e24687c/LibOS/shim/src/fs/dev/attestation.c#L360)):*

Can we add some `log_debug()` message here? Would be nice to tell something like
```
host is Linux-SGX, adding SGX-specific attestation files: report, quote, etc.
```
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/keys.c, line 18 at r4](https://reviewable.io/reviews/gramineproject/gramine/504#-N-Ris2Y5bxoI6H8wAMA:-N-RsBVXD96s6ubhj7jW:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/4ace5ae7b8b37ec43d2b43e8ee4b04583e24687c/LibOS/shim/test/regression/keys.c#L18)):*

Can you add a comment to look for this key and its original value in `manifest.template`?
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/keys.c, line 36 at r4](https://reviewable.io/reviews/gramineproject/gramine/504#-N-RiaV72mfh9Jdhs2sk:-N-RsFA08jPcbIlOP83P:bidrs83) ([raw file](https://github.com/gramineproject/gramine/blob/4ace5ae7b8b37ec43d2b43e8ee4b04583e24687c/LibOS/shim/test/regression/keys.c#L36)):*

Reorder `key` and `expected_key` arguments -- your string wants them in different order
</blockquote></details>

Right. Fixed.
___
*[LibOS/shim/test/regression/manifest.template, line 47 at r4](https://reviewable.io/reviews/gramineproject/gramine/504#-N-RjUAG3yEr5ounfyal:-N-RtyZ06aaXxmGLaZeo:bax2gw6) ([raw file](https://github.com/gramineproject/gramine/blob/4ace5ae7b8b37ec43d2b43e8ee4b04583e24687c/LibOS/shim/test/regression/manifest.template#L47)):*

What about we also specify the `default` key, and augment the test above to check it as well?

This way, we'll additionally test two things:
- The linked list of keys
- The checkpoint-restore of even those keys that didn't have the associated file (`/dev/attestation/keys/default`) opened/written to.
</blockquote></details>

OK. Updated.
___
*[LibOS/shim/test/regression/manifest.template, line 47 at r4](https://reviewable.io/reviews/gramineproject/gramine/504#-N-Rjr0w8jnE-R__rBUj:-N-Ru1KEDsjuax6ZyvMP:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/4ace5ae7b8b37ec43d2b43e8ee4b04583e24687c/LibOS/shim/test/regression/manifest.template#L47)):*

Maybe also add a comment to `see the keys.c test`?
</blockquote></details>

Done.

dimakuv (dismissed):
> Reviewed 5 of 5 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/504)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[LibOS/shim/src/fs/dev/attestation.c, line 16 at r4](https://reviewable.io/reviews/gramineproject/gramine/504#-N-Rh6Gg29tkeW9UjKdW:-N-SS64HA9PBmFekEDRb:b-gniufx) ([raw file](https://github.com/gramineproject/gramine/blob/4ace5ae7b8b37ec43d2b43e8ee4b04583e24687c/LibOS/shim/src/fs/dev/attestation.c#L16)):*

Isn't it too early for that? The current system ("protected files") is not even deprecated yet.

I agree that these need to be removed eventually, but first we'll need to switch all these APIs to use encrypted keys (e.g. `pfkey_load` will call `key_load` for the `"default"` key... or even the file could become a symlink), and document `type = "encrypted"` to be the right way of using the feature.

Then, I think, will be the right time to start the deprecation period, and add TODOs to delete all of these compatibility shims.
</blockquote></details>

OK, hope we won't forget to remove all these things.
___
*[LibOS/shim/src/fs/dev/attestation.c, line 299 at r4](https://reviewable.io/reviews/gramineproject/gramine/504#-N-Rg1xz8DaixWxIZv4G:-N-SSAtR0SP7TrtFl-ef:b-xy91cl) ([raw file](https://github.com/gramineproject/gramine/blob/4ace5ae7b8b37ec43d2b43e8ee4b04583e24687c/LibOS/shim/src/fs/dev/attestation.c#L299)):*

I regret the boilerplate, but I'd rather keep the module (`shim_fs_encrypted.[ch]`) independent from the rest of LibOS, and using the `readdir_callback_t` type breaks that.

I guess I could say that `list_encrypted_files_keys` actually works on *names*, and declare it to take a callback of form `int (*callback)(const char* name, void* arg)`, which happens to be the same type as `readdir_callback_t`. The types would be even enforced by compiler.

But still, I don't think this would be a good situation, the coupling is essentially still there: if we ever change how `readdir` works, now we also have to either reach into `shim_fs_encrypted.[ch]` and change it as well, or rewrite the code here.
</blockquote></details>

Ok, I see, you want to keep `shim_fs_encrypted.h` completely separated.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/504)* status: 13 of 19 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "squash! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/504#-N-Y8EWbEluvxPrnOWKK:-N-bxY2FDjX4fKZ7FEg_:b-896fix) (no related file):*

TODO: as discussed in #520, I'll convert `/dev/attestation/keys/*` files to use binary data instead of hex.
</blockquote></details>

Done.

---

dimakuv (changes requested):
> Reviewed 5 of 6 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/504)* status: 18 of 19 files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "squash! " found in commit messages' one-liners (waiting on @mkow and @pwmarcz)

___
*[LibOS/shim/src/fs/shim_fs_encrypted.c, line 204 at r6](https://reviewable.io/reviews/gramineproject/gramine/504#-N-c61AW2_aXv7AaNYIq:-N-c61AXE5nOJxb95SB1:ba15nt9) ([raw file](https://github.com/gramineproject/gramine/blob/7c5550d1a7dd5c5aaf2a3679121de783b8058000/LibOS/shim/src/fs/shim_fs_encrypted.c#L204)):*
> ```C
>         log_warning("%s: wrong key length (%zu instead of %zu)", __func__, len,
>                     (size_t)(PF_KEY_SIZE * 2));
>         return -1;
> ```

Why not just `-EINVAL`?
___
*[LibOS/shim/src/fs/shim_fs_encrypted.c, line 213 at r6](https://reviewable.io/reviews/gramineproject/gramine/504#-N-c63IR4AfGI1auFwJV:-N-c63IR4AfGI1auFwJW:ba15nt9) ([raw file](https://github.com/gramineproject/gramine/blob/7c5550d1a7dd5c5aaf2a3679121de783b8058000/LibOS/shim/src/fs/shim_fs_encrypted.c#L213)):*
> ```C
>         if (hi < 0 || lo < 0) {
>             log_warning("%s: unexpected character encountered", __func__);
>             return -1;
> ```

Why not just `-EINVAL`?
___
*[LibOS/shim/src/fs/shim_fs_encrypted.c, line 285 at r6](https://reviewable.io/reviews/gramineproject/gramine/504#-N-c64cn2m_zYq8DTuR1:-N-c64cn2m_zYq8DTuR2:b-cafcsx) ([raw file](https://github.com/gramineproject/gramine/blob/7c5550d1a7dd5c5aaf2a3679121de783b8058000/LibOS/shim/src/fs/shim_fs_encrypted.c#L285)):*
> ```C
>         if (ret < 0) {
>             log_error("Cannot parse key 'fs.insecure__keys.%s' as a hex key", name);
>             ret = -EINVAL;
> ```

If you change the error code in the function, you won't need this
___
*[LibOS/shim/test/regression/keys.c, line 39 at r6](https://reviewable.io/reviews/gramineproject/gramine/504#-N-c7FpxDU36dzXpjwzS:-N-c7Fpy9ozhv3u-T6QV:bnbe9dj) ([raw file](https://github.com/gramineproject/gramine/blob/7c5550d1a7dd5c5aaf2a3679121de783b8058000/LibOS/shim/test/regression/keys.c#L39)):*
> ```C
>     0x88, 0x99, 0xaa, 0xbb, 0xcc, 0xdd, 0xee, 0xff, 0x00, 0x11, 0x22, 0x33, 0x44, 0x55, 0x66, 0x77};
> 
> static void format_key(char buf[static KEY_STR_SIZE], const pf_key_t* key) {
> ```

What does this mean: `char buf[static KEY_STR_SIZE]`? The first time I see this construct.
___
*[LibOS/shim/test/regression/keys.c, line 82 at r6](https://reviewable.io/reviews/gramineproject/gramine/504#-N-c7tpT8nGTUQugx9bE:-N-c7tpT8nGTUQugx9bF:bdzs92u) ([raw file](https://github.com/gramineproject/gramine/blob/7c5550d1a7dd5c5aaf2a3679121de783b8058000/LibOS/shim/test/regression/keys.c#L82)):*
> ```C
>     expect_key("before writing key", CUSTOM_KEY_PATH, &custom_key);
> 
>     ssize_t n = posix_file_write(CUSTOM_KEY_PATH, (char*)&new_custom_key,
> ```

Can you add a comment that this key-write is invalid because the size is too small?
___
*[LibOS/shim/test/regression/keys.c, line 85 at r6](https://reviewable.io/reviews/gramineproject/gramine/504#-N-c7rNlDKz_kztFsRko:-N-c7rNlDKz_kztFsRkp:b-npvjj2) ([raw file](https://github.com/gramineproject/gramine/blob/7c5550d1a7dd5c5aaf2a3679121de783b8058000/LibOS/shim/test/regression/keys.c#L85)):*
> ```C
>                                  sizeof(new_custom_key) - 1);
>     if (n >= 0 || (n < 0 && errno != EACCES))
>         err(1, "writing invalid key: expected EACCES");
> ```

Can you add an empty line after this?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/504)* status: 14 of 19 files reviewed, 11 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @pwmarcz)

___
*[LibOS/shim/include/shim_fs.h, line 40 at r6](https://reviewable.io/reviews/gramineproject/gramine/504#-N-c-J4e4Jl3io2aTUqK:-N-g_HNJCjkfH1IpA1mR:b-ofdaz5) ([raw file](https://github.com/gramineproject/gramine/blob/7c5550d1a7dd5c5aaf2a3679121de783b8058000/LibOS/shim/include/shim_fs.h#L40)):*

Maybe `key_name` instead? Would be less ambiguous and still not overly long.
</blockquote></details>

Makes sense. Renamed.

We still want the manifest syntax to be `{ ..., key = "foo" }`, not `{ ..., key_name = "foo" }`, right?
___
*[LibOS/shim/include/shim_fs_encrypted.h, line 93 at r6](https://reviewable.io/reviews/gramineproject/gramine/504#-N-c16kC1AFgw30UQ-bY:-N-gXZo_09DEJc5Y0qaJ:bi2vmjc) ([raw file](https://github.com/gramineproject/gramine/blob/7c5550d1a7dd5c5aaf2a3679121de783b8058000/LibOS/shim/include/shim_fs_encrypted.h#L93)):*

new? copy-paste error, I guess?
</blockquote></details>

Right, it's current value. Fixed.
___
*[LibOS/shim/src/fs/shim_fs_encrypted.c, line 204 at r6](https://reviewable.io/reviews/gramineproject/gramine/504#-N-c61AW2_aXv7AaNYIq:-N-gXidxAUGU2cwzs_pJ:bectrkb) ([raw file](https://github.com/gramineproject/gramine/blob/7c5550d1a7dd5c5aaf2a3679121de783b8058000/LibOS/shim/src/fs/shim_fs_encrypted.c#L204)):*

Why not just `-EINVAL`?
</blockquote></details>

This is consistent with other "parse" type functions (there's one in `pseudo`, and some in `api.h`, like `parse_size_str` or `str_to_ulong`) - they don't really return error codes because there's only one type of error possible, so I think it's better to just treat it as success/failure.
___
*[LibOS/shim/src/fs/shim_fs_encrypted.c, line 673 at r6](https://reviewable.io/reviews/gramineproject/gramine/504#-N-c6lLjEdUJMkKGJ35X:-N-gYK24Dez1M-g5zOKM:b-ge9bpt) ([raw file](https://github.com/gramineproject/gramine/blob/7c5550d1a7dd5c5aaf2a3679121de783b8058000/LibOS/shim/src/fs/shim_fs_encrypted.c#L673)):*

I assume we can't use `update_encrypted_files_key()` here because the key may not be set, right?
</blockquote></details>

Yeah... I could do `if (is_set) update_encrypted_files_key()`, and it would probably work (if the key is not set in parent, that means the child won't have initialized it either), but that's unobvious and brittle. I feel better about copying data directly.
___
*[LibOS/shim/test/regression/keys.c, line 39 at r6](https://reviewable.io/reviews/gramineproject/gramine/504#-N-c7FpxDU36dzXpjwzS:-N-gYoMD1SLSmc6BZF0R:b9gugvi) ([raw file](https://github.com/gramineproject/gramine/blob/7c5550d1a7dd5c5aaf2a3679121de783b8058000/LibOS/shim/test/regression/keys.c#L39)):*

In C and C++ you can basically put `static` in any place in the code and it's a valid construct, although it has a different meaning in each of them.
https://hamberg.no/erlend/posts/2013-02-18-static-array-indices.html
</blockquote></details>

As I understand, this means that `buf` is a [valid pointer to at least KEY_STR_SIZE elements](https://stackoverflow.com/questions/18362315/static-size-of-array-in-c99).

Sadly, this doesn't make GCC warn on passing a smaller array or NULL (although when caller is in the same file, it will warn anyway). So basically it's for documentation purposes.
___
*[LibOS/shim/test/regression/keys.c, line 40 at r6](https://reviewable.io/reviews/gramineproject/gramine/504#-N-cCGgB4QVXzr0HEzpB:-N-gYyS77CYdtEeypKaE:b-vmlnd7) ([raw file](https://github.com/gramineproject/gramine/blob/7c5550d1a7dd5c5aaf2a3679121de783b8058000/LibOS/shim/test/regression/keys.c#L40)):*

> 01234567890

:thinking:
</blockquote></details>

Haha :) Good thing it's only a test. Fixed.
___
*[LibOS/shim/test/regression/keys.c, line 53 at r6](https://reviewable.io/reviews/gramineproject/gramine/504#-N-cCoF_4o8vba33vXqL:-N-gZFBfCuYCdqR-0Fbv:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/7c5550d1a7dd5c5aaf2a3679121de783b8058000/LibOS/shim/test/regression/keys.c#L53)):*
> ```C
>     pf_key_t key;
> 
>     ssize_t n = posix_file_read(path, (char*)key, sizeof(key));
> ```

Done.
___
*[LibOS/shim/test/regression/keys.c, line 56 at r6](https://reviewable.io/reviews/gramineproject/gramine/504#-N-cCzNa8gXbv-6LbnqH:-N-gZLLbCJ9zc3xNU_fT:b-9ui3n2) ([raw file](https://github.com/gramineproject/gramine/blob/7c5550d1a7dd5c5aaf2a3679121de783b8058000/LibOS/shim/test/regression/keys.c#L56)):*
> ```C
>     if (n < 0)
>         err(1, "%s: error reading %s", desc, path);
>     if (n < (ssize_t)sizeof(key))
> ```

I  agree. Done.
___
*[LibOS/shim/test/regression/keys.c, line 82 at r6](https://reviewable.io/reviews/gramineproject/gramine/504#-N-c7tpT8nGTUQugx9bE:-N-gZZ1r6fyf1dxEOS_m:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/7c5550d1a7dd5c5aaf2a3679121de783b8058000/LibOS/shim/test/regression/keys.c#L82)):*

Can you add a comment that this key-write is invalid because the size is too small?
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/keys.c, line 85 at r6](https://reviewable.io/reviews/gramineproject/gramine/504#-N-c7rNlDKz_kztFsRko:-N-gZ_N35Vwso3eJn0F5:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/7c5550d1a7dd5c5aaf2a3679121de783b8058000/LibOS/shim/test/regression/keys.c#L85)):*

Can you add an empty line after this?
</blockquote></details>

Done.

mkow (commented):
> Reviewed 5 of 5 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/504)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @pwmarcz)

___
*[LibOS/shim/include/shim_fs.h, line 40 at r6](https://reviewable.io/reviews/gramineproject/gramine/504#-N-c-J4e4Jl3io2aTUqK:-N-u3VTk8eQ2hb4EngAP:b3icr4h) ([raw file](https://github.com/gramineproject/gramine/blob/7c5550d1a7dd5c5aaf2a3679121de783b8058000/LibOS/shim/include/shim_fs.h#L40)):*

Makes sense. Renamed.

We still want the manifest syntax to be `{ ..., key = "foo" }`, not `{ ..., key_name = "foo" }`, right?
</blockquote></details>

Good question, I think I actually prefer the latter, but not super strongly.
@boryspoplawski, @dimakuv ^
___
*[LibOS/shim/test/regression/keys.c, line 40 at r6](https://reviewable.io/reviews/gramineproject/gramine/504#-N-cCGgB4QVXzr0HEzpB:-N-u4ADv80ktU4D2L7D9:buneo1x) ([raw file](https://github.com/gramineproject/gramine/blob/7c5550d1a7dd5c5aaf2a3679121de783b8058000/LibOS/shim/test/regression/keys.c#L40)):*

Haha :) Good thing it's only a test. Fixed.
</blockquote></details>

Yeah, but people sometimes copy code from tests to the main codebase :)

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/504)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @pwmarcz)

___
*[LibOS/shim/include/shim_fs.h, line 40 at r6](https://reviewable.io/reviews/gramineproject/gramine/504#-N-c-J4e4Jl3io2aTUqK:-N-xqrzrCsPcCt3i-jA9:b6b6w67) ([raw file](https://github.com/gramineproject/gramine/blob/7c5550d1a7dd5c5aaf2a3679121de783b8058000/LibOS/shim/include/shim_fs.h#L40)):*

Good question, I think I actually prefer the latter, but not super strongly.
@boryspoplawski, @dimakuv ^
</blockquote></details>

I did not review the whole PR (should I?) so I need more context - that's an identifier for a key, so that we can pick appropriate key from specific `/dev` entry, right? If that's correct, then I would prefer `key_name` or `key_id` - just `key` sounds like an inline key or a path to a file containing it.

dimakuv (dismissed):
> Reviewed 1 of 6 files at r6, 5 of 5 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/504)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[LibOS/shim/include/shim_fs.h, line 40 at r6](https://reviewable.io/reviews/gramineproject/gramine/504#-N-c-J4e4Jl3io2aTUqK:-N0-ze0K6BWoV6OPFyD0:b-4ghb44) ([raw file](https://github.com/gramineproject/gramine/blob/7c5550d1a7dd5c5aaf2a3679121de783b8058000/LibOS/shim/include/shim_fs.h#L40)):*

I did not review the whole PR (should I?) so I need more context - that's an identifier for a key, so that we can pick appropriate key from specific `/dev` entry, right? If that's correct, then I would prefer `key_name` or `key_id` - just `key` sounds like an inline key or a path to a file containing it.
</blockquote></details>

I agree with Michal and Borys that `key_name` is less ambigious. So I vote for this rename in the manifest syntax as well. (I don't like `key_id`, because IDs for me are some random numbers, but here we're talking about human-readable human-defined names.)

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/504)* status: 12 of 19 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "squash! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[LibOS/shim/include/shim_fs.h, line 40 at r6](https://reviewable.io/reviews/gramineproject/gramine/504#-N-c-J4e4Jl3io2aTUqK:-N00Tysu5CU1iA_GqwbM:b-hfgn1z) ([raw file](https://github.com/gramineproject/gramine/blob/7c5550d1a7dd5c5aaf2a3679121de783b8058000/LibOS/shim/include/shim_fs.h#L40)):*

I agree with Michal and Borys that `key_name` is less ambigious. So I vote for this rename in the manifest syntax as well. (I don't like `key_id`, because IDs for me are some random numbers, but here we're talking about human-readable human-defined names.)
</blockquote></details>

OK. I changed the manifest syntax to `key_name` as well.

mkow (dismissed):
> Reviewed 2 of 7 files at r8, 6 of 6 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/504)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

dimakuv (dismissed):
> Reviewed 2 of 7 files at r8, 1 of 6 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/504)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

mkow (approved):
> Reviewed 4 of 4 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/504)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (approved):
> Reviewed 2 of 4 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/504)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #519, opened 2022-04-12
[LibOS] Add support for fallocate

sahason:
Signed-off-by: Sonali Saha <sonali.saha@intel.com>

## Description of the changes 
Support for fallocate syscall is added to enable mariadb server in gramine. This contains support for default operation (i.e., mode is zero) only, other modes are not supported.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/519)


boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/519)* status: 0 of 15 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @sahason)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/519#-N-Su5gt8kKIe2xy5Za2:-N-T98exB5X_B_nc49_f:bp0636t) (no related file):*

I recommend reading https://gramine.readthedocs.io/en/latest/devel/onboarding.html#implementing-a-new-system-call especially point 3.

This implementation does not support Protected Files at all (and this syscall makes no sense on Truested Files). What's the point then?
</blockquote></details>

@sahason please check if #519 works for you.

---

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/519)* status: 0 of 15 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @sahason)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/519#-N-Su5gt8kKIe2xy5Za2:-N-T9IGR2HKJOsnckKuC:b-bypxhu) (no related file):*

@sahason please check if #519 works for you.
</blockquote></details>

Err, I meant #522

---

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/519)* status: 0 of 15 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @sahason)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/519#-N-XaEkv2hmNkbUxD_rm:-N-XaEkwBrVMKTw05kyA:b-7vzj7x) (no related file):*
I'd like to give some explanation on the apparent duplication of this PR and #522. Unfortunately, this was a case of lack of communication between @sahason (the author of this PR) and @boryspoplawski (the author of #522).

This PR suffers from several substantial issues:
1. The implementation of the `fallocate()` syscall is chosen to be pass-through. In other words, Gramine makes no attempts to emulate this syscall inside of the LibOS layer but instead forwards this syscall as-is to the host OS.
    - This unnecesarily extends the attack surface for our Linux-SGX PAL.
    - The `fallocate()` syscall is Linux-specific, thus, implementing its `DkStreamFallocate()` PAL counterpart on a hypothetic Windows PAL would not be feasible (or very complex and unintuitive).
    - This particular syscall may be fully implemented in the LibOS layer; thus no changes to the PAL layer are needed at all.
2. The Linux-SGX PAL implementation is too simplistic and doesn't cover several real-world cases:
    - Protected Files implementation is not supported (explicitly returns an error); however, this is the case where `fallocate()` syscall would make the most sense.
    - `DkStreamFallocate()` on a Trusted File is not explicitly implemented and thus not tested -- Gramine would probably crash or misbehave in such a test case. In reality, `fallocate()` on Trusted Files should have been explicitly prohibited in this implementation.
3. There are no tests created/enabled with this PR that verify implementation of `fallocate()`:
    - No LTP tests were checked and enabled.
    - No LibOS regression/fs tests were created and added to our CI.

Given the above deficiencies, it seems appropriate to close this PR (as it would be too much overhead and complications to modify this PR that significantly) and open a new, improved PR. This is what was done with #522.

Unfortunately, the reasons to close this PR were not articulated, and there was no open discussion with the PR author. Ideally, after the discussion it would be decided whether the original author of this PR would create another PR, with an improved implementation, or yield the implementation to another person.

To ensure such misunderstandings do not happen in the future, I suggest to enforce the following principles:
- There should be no overriding of PRs until a discussion in the original PR and an explicit acknowledgement from the original PR authors (waiting for an acknowledgement with some timeout, like a week)
- There should be some explanation in the original PR description/comments on why there is a need for a duplicate (new) PR.
- The above concerns *not-highly-critical* PRs. (In case of PRs with high priority, speed of execution is more important than managerial discussions.)

---

PR #526, opened 2022-04-14
[Pal/Linux-SGX] Allow READ+WRITE mmap of protected files selectively

### LINKED ISSUE #196: [Pal/SGX] Unable to map executables from protected fs
When trying to execute a binary via execve() or loading .so via dlopen(), if the file is an protected file, a warning is printed from the following line and the mmap() call will fail:

https://github.com/gramineproject/gramine/blob/3db1e28e27622cbf1265650cc92a87281803d31d/Pal/src/host/Linux-SGX/db_files.c#L401

therefore, the binary is not able to run.

The mmap() call actually comes with PAL_PROT_WRITECOPY (aka MAP_PRIVATE), and there's no need to write back to the underlying file. Perhaps we should ignore PAL_PROT_WRITE when PAL_PROT_WRITECOPY is present?

lejunzhu:
Mmap() on protected files only allows PROT_READ or PROT_WRITE, not both.
However, when MAP_PRIVATE flag is present, changes to the file do not need
to be written back to the file, essentially makes the behavior of file access
identical to the read only mmap.

By allowing READ+WRITE mmap with MAP_PRIVATE flag, it is now possible to load
executables from the protected FS.

Fixes #196

Signed-off-by: Lejun Zhu <lejun.zhu@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/526)

### DIFF
--- a/LibOS/shim/test/fs/meson.build
+++ b/LibOS/shim/test/fs/meson.build
@@ -45,6 +45,7 @@ tests = {
     'seek_tell': {},
     'stat': {},
     'truncate': {},
+    'open_mmap': {},
 }
 
 install_dir = join_paths(pkglibdir, 'tests', 'libos', 'fs')
--- a/LibOS/shim/test/fs/open_mmap.c
+++ b/LibOS/shim/test/fs/open_mmap.c
@@ -1,0 +1,61 @@
+#include "common.h"
+
+static void mmap_test(int fd, off_t offset, size_t length,
+                      int prot, const char *prot_str,
+                      int flags, const char *flags_str,
+                      bool expect_success) {
+    void *address = mmap(NULL, length, prot, flags, fd, offset);
+    if (address == MAP_FAILED) {
+        if (expect_success) {
+            fatal_error("mmap %s, %s failed\n", prot_str, flags_str);
+        } else {
+            printf("mmap %s, %s failed as expected\n", prot_str, flags_str);
+        }
+    } else {
+        munmap(address, length);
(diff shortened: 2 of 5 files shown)


pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/526)* status: 0 of 5 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @lejunzhu)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/526#-N0p6YhqBVqvUAZdG40M:-N0p6YhqBVqvUAZdG40N:b9jujh6) (no related file):*
Hi. I'm actually about to replace the whole PAL protected files subsystem with a new version ("encrypted files") that works on all PALs, so unfortunately, the code that you modified will be gone soon. See #371 for background: basically, there were too many design problems to keep developing the current PAL-based design.

The good news is that we just merged support this feature (mapping files, including R+W private and shared mappings), this was done in PR #550.

Could you check if this works for your use case? The new subsystem is not documented yet, but you can take a look at `LibOS/shim/test/regression/manifest.template` for usage details (basically, instead of using `sgx.protected_files`, you need to mount a file/directory using `{ type = "encrypted" }`. Later, `sgx.protected_files` will also use this subsystem).

---

lejunzhu:
I just tried PR #550 and it is working well. So I'll close this PR.

PR #534, opened 2022-04-19
[LibOS] Don't use PAL trusted/protected files in `encrypted` fs

pwmarcz:
## Description of the changes 

The new mount syntax supersedes `sgx.{allowed,trusted,protected}_files`, and the `encrypted` filesystem should always create "passthrough" PAL handles. This is ensured by introducing a special option to `DkStreamOpen`.

(I think that this option is temporary - if we manage to move all of trusted/protected files to LibOS, then all PAL handles will be "passthrough").

## How to test this PR? 

The LibOS regression tests worked only because we mounted `tmp/enc`, and we happened to include `tmp/` in the list of allowed files. Instead, the tests for encrypted files use a separate host directory (`tmp_enc/`).

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/534)

### DIFF
--- a/LibOS/shim/src/fs/chroot/encrypted.c
+++ b/LibOS/shim/src/fs/chroot/encrypted.c
@@ -251,8 +251,8 @@ static int chroot_encrypted_mkdir(struct shim_dentry* dent, mode_t perm) {
 
     /* This opens a "dir:..." URI */
     PAL_HANDLE palhdl;
-    ret = DkStreamOpen(uri, PAL_ACCESS_RDONLY, HOST_PERM(perm), PAL_CREATE_ALWAYS, /*options=*/0,
-                       &palhdl);
+    ret = DkStreamOpen(uri, PAL_ACCESS_RDONLY, HOST_PERM(perm), PAL_CREATE_ALWAYS,
+                       PAL_OPTION_PASSTHROUGH, &palhdl);
     if (ret < 0) {
         ret = pal_to_unix_errno(ret);
         goto out;
@@ -267,6 +267,37 @@ static int chroot_encrypted_mkdir(struct shim_dentry* dent, mode_t perm) {
 
 out:
     put_inode(inode);
+    free(uri);
+    return ret;
+}
+
+/* NOTE: this function is different from generic `chroot_unlink` only to add PAL_OPTION_PASSTHROUGH.
+ * Once that option is removed, we can safely go back to using `chroot_unlink`. */
+static int chroot_encrypted_unlink(struct shim_dentry* dent) {
+    assert(locked(&g_dcache_lock));
+    assert(dent->inode);
+
+    char* uri;
+    int ret = chroot_dentry_uri(dent, dent->inode->type, &uri);
+    if (ret < 0)
+        return ret;
+
+    PAL_HANDLE palhdl;
+    ret = DkStreamOpen(uri, PAL_ACCESS_RDONLY, /*share_flags=*/0, PAL_CREATE_NEVER,
+                       PAL_OPTION_PASSTHROUGH, &palhdl);
+    if (ret < 0) {
+        ret = pal_to_unix_errno(ret);
+        goto out;
+    }
+
+    ret = DkStreamDelete(palhdl, PAL_DELETE_ALL);
+    DkObjectClose(palhdl);
+    if (ret < 0) {
+        ret = pal_to_unix_errno(ret);
+        goto out;
+    }
+    ret = 0;
+out:
     free(uri);
     return ret;
 }
@@ -311,8 +342,8 @@ static int chroot_encrypted_chmod(struct shim_dentry* dent, mode_t perm) {
         goto out;
 
     PAL_HANDLE palhdl;
-    ret = DkStreamOpen(uri, PAL_ACCESS_RDONLY, /*share_flags=*/0, PAL_CREATE_NEVER, /*options=*/0,
-                       &palhdl);
+    ret = DkStreamOpen(uri, PAL_ACCESS_RDONLY, /*share_flags=*/0, PAL_CREATE_NEVER,
+                       PAL_OPTION_PASSTHROUGH, &palhdl);
     if (ret < 0) {
         ret = pal_to_unix_errno(ret);
         goto out;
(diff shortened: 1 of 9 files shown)


mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/534)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

___
*[LibOS/shim/src/fs/chroot/encrypted.c, line 282 at r1](https://reviewable.io/reviews/gramineproject/gramine/534#-N02xNAxCcBEG_eRiDgL:-N02xNAxCcBEG_eRiDgM:biwaiko) ([raw file](https://github.com/gramineproject/gramine/blob/9aee3aeb427413804ec2f4c8caeaba0dae6989d5/LibOS/shim/src/fs/chroot/encrypted.c#L282)):*
> ```C
>     int ret = chroot_dentry_uri(dent, dent->inode->type, &uri);
>     if (ret < 0)
>         goto out;
> ```

I'd feel better if this was `return ret` instead. I don't like having any assumptions about ret-by-arg values in case of failure.
___
*[Pal/include/pal/pal.h, line 261 at r1](https://reviewable.io/reviews/gramineproject/gramine/534#-N02wwNW5q5SsHqZwfNc:-N02wwNXDv44UrovxqA7:bsysufg) ([raw file](https://github.com/gramineproject/gramine/blob/9aee3aeb427413804ec2f4c8caeaba0dae6989d5/Pal/include/pal/pal.h#L261)):*
> ```objc
> #define PAL_OPTION_NONBLOCK        4
> #define PAL_OPTION_DUALSTACK       8 /*!< Create dual-stack socket (opposite of IPV6_V6ONLY) */
> #define PAL_OPTION_PASSTHROUGH    16 /*!< Disregard `sgx.{allowed,trusted,protected}_files` */
> ```

Could you use hex here?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/534)* status: 7 of 10 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[LibOS/shim/src/fs/chroot/encrypted.c, line 255 at r1](https://reviewable.io/reviews/gramineproject/gramine/534#-N04nyhR9KKE96MsFiIl:-N05CkoC2p2xJC1IO6TD:b-o9gx76) ([raw file](https://github.com/gramineproject/gramine/blob/9aee3aeb427413804ec2f4c8caeaba0dae6989d5/LibOS/shim/src/fs/chroot/encrypted.c#L255)):*

Is this needed here (we're opening a dir, so it should end up in `dir_open()` of PAL's file handling)? Or you added it just for consistency?
</blockquote></details>

You're right, it's only for consistency. Would you prefer me to pass 0 here? I'm not very attached to either choice, this one just seemed less surprising.
___
*[LibOS/shim/src/fs/chroot/encrypted.c, line 274 at r1](https://reviewable.io/reviews/gramineproject/gramine/534#-N04ohRHDvBMB4i6yRO4:-N05Deuh3yO6V3GHpkFd:bcvjgzp) ([raw file](https://github.com/gramineproject/gramine/blob/9aee3aeb427413804ec2f4c8caeaba0dae6989d5/LibOS/shim/src/fs/chroot/encrypted.c#L274)):*

The only reason we introduce this function is to specify `PAL_OPTION_PASSTHROUGH`? Can you specify this in the NOTE or TODO comment, that we can safely move back to the generic `chroot_unlink()` callback once this PASSTHROUGH option is removed?
</blockquote></details>

Good idea. Done.
___
*[LibOS/shim/src/fs/chroot/encrypted.c, line 282 at r1](https://reviewable.io/reviews/gramineproject/gramine/534#-N02xNAxCcBEG_eRiDgL:-N05Dn1M8HUrDwG7EtHf:b-k1y5pj) ([raw file](https://github.com/gramineproject/gramine/blob/9aee3aeb427413804ec2f4c8caeaba0dae6989d5/LibOS/shim/src/fs/chroot/encrypted.c#L282)):*

I'd feel better if this was `return ret` instead. I don't like having any assumptions about ret-by-arg values in case of failure.
</blockquote></details>

OK. Done.
___
*[Pal/include/pal/pal.h, line 261 at r1](https://reviewable.io/reviews/gramineproject/gramine/534#-N02wwNW5q5SsHqZwfNc:-N05E81h38hGQW3jkeo1:babrut3) ([raw file](https://github.com/gramineproject/gramine/blob/9aee3aeb427413804ec2f4c8caeaba0dae6989d5/Pal/include/pal/pal.h#L261)):*

Could you use hex here?
</blockquote></details>

Makes sense. I converted all these to hex.

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/534)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[LibOS/shim/src/fs/chroot/encrypted.c, line 255 at r1](https://reviewable.io/reviews/gramineproject/gramine/534#-N04nyhR9KKE96MsFiIl:-N05Kp-h-PjiicEu8ipJ:btrzzuz) ([raw file](https://github.com/gramineproject/gramine/blob/9aee3aeb427413804ec2f4c8caeaba0dae6989d5/LibOS/shim/src/fs/chroot/encrypted.c#L255)):*

You're right, it's only for consistency. Would you prefer me to pass 0 here? I'm not very attached to either choice, this one just seemed less surprising.
</blockquote></details>

No, I'm fine with the current version.

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/534)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/534)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/534)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #540, opened 2022-04-20
[Pal] Allow `arch_prctl(ARCH_REQ_XCOMP_PERM)` to return `-ENOTSUPP`

dimakuv:
## Description of the changes 

This syscall was introduced for CPUs with Intel AMX. On some Linux kernels with the pre-upstreamed AMX patch, this syscall erroneously returns `-ENOTSUPP` instead of `-EOPNOTSUPP`. This commit is to allow
running Gramine on such patched Linux kernels, otherwise Gramine fails with "Requesting AMX permission failed".

I'm not sure if we want this "work around weird patched Linux" PR. But since I found this, I decided to create an immediate PR.

## How to test this PR? 

This was found on a patched Linux 5.12 kernel (with one of the earlier AMX patches that exhibits this wrong errno).

Here is an snippet from the strace of Gramine, when it tries to do this syscall:
```
arch_prctl(0x1023 /* ARCH_??? */, 0x12) = -1 ENOTSUPP (Unknown error 524)
```

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/540)

### DIFF
--- a/Pal/src/host/Linux-SGX/sgx_framework.c
+++ b/Pal/src/host/Linux-SGX/sgx_framework.c
@@ -200,11 +200,12 @@ int create_enclave(sgx_arch_secs_t* secs, sgx_arch_token_t* token) {
      * will inherit the permission, but here for simplicity we call it in every child process as
      * well. Some deployment environments run Linux systems earlier than v5.16 but with
      * an AMX-specific patch; this patch doesn't introduce `arch_prctl(ARCH_REQ_XCOMP_PERM)`
-     * syscall so an attempt to call it may return EINVAL, EOPNOTSUPP or ENOSYS. In this case,
-     * we simply ignore the result of this syscall. */
+     * syscall so an attempt to call it may return EINVAL, ENOTSUPP, EOPNOTSUPP or ENOSYS. In this
+     * case, we simply ignore the result of this syscall. */
     if (secs->attributes.xfrm & (1 << AMX_TILEDATA)) {
         ret = DO_SYSCALL(arch_prctl, ARCH_REQ_XCOMP_PERM, AMX_TILEDATA);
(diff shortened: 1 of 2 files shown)


mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/540)* status: 0 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`-- commits` line 5 at r1](https://reviewable.io/reviews/gramineproject/gramine/540#-N0YHh6x-iBc8W4UMrO7):*
> On some Linux kernels with the pre-upstreamed AMX patch

Why should we care? Why can't the users who reported this use upstream kernel or up-to-date patches?

I really don't like the idea of working around bugs in some development versions of kernel patches.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/540)* status: 0 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

___
*[`-- commits` line 5 at r1](https://reviewable.io/reviews/gramineproject/gramine/540#-N0YHh6x-iBc8W4UMrO7):*

> On some Linux kernels with the pre-upstreamed AMX patch

Why should we care? Why can't the users who reported this use upstream kernel or up-to-date patches?

I really don't like the idea of working around bugs in some development versions of kernel patches.
</blockquote></details>

Yeah. This was a way to quickly fix up the failing workload.

But I agree that this is a hack, and it is not something that affects many users. Let me close this PR.

PR #539, opened 2022-04-20
[LibOS] Support special keys (MRENCLAVE, MRSIGNER) in `encrypted` fs

pwmarcz:
## Description of the changes 

Key names beginning with `_` are considered special, and PAL is queried for these keys. This is used for SGX MRENCLAVE and MRSIGNER keys.

If a key is not supported by the current PAL, mounting will still succeed (but files will not be readable). This way, manifests using these PAL-host-specific keys can still be loaded when using other PAL hosts.

(This design is as discussed in #371).

## How to test this PR? 

The existing tests for MRENCLAVE and MRSIGNER keys now also check the new implementation.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/539)

PR #550, opened 2022-04-26
[LibOS] Implement `mmap` for encrypted and tmpfs files 

pwmarcz:
## Description of the changes 

This adds support for "emulated" memory mappings in LibOS. Memory is written back to the file using a new `msync` callback.

This (and #539) bring us to feature parity with SGX protected files, so the next step will be to replace the old implementation (see issue: #371).

## How to test this PR? 

There is a new LibOS regression test, `mmap_file_emulated`, that checks reading and writing an emulated file. In addition, some tmpfs tests are now unskipped.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/550)

### DIFF
--- a/LibOS/shim/include/shim_fs.h
+++ b/LibOS/shim/include/shim_fs.h
@@ -82,9 +82,43 @@ struct shim_fs_ops {
      */
     ssize_t (*write)(struct shim_handle* hdl, const void* buf, size_t count, file_off_t* pos);
 
-    /* mmap: mmap handle to address */
-    int (*mmap)(struct shim_handle* hdl, void** addr, size_t size, int prot, int flags,
+    /*
+     * \brief Map file at an address.
+     *
+     * \param hdl     File handle.
+     * \param addr    Address of the memory region. Cannot be NULL.
+     * \param size    Size of the memory region.
+     * \param prot    Permissions for the memory region (`PROT_*`).
+     * \param flags   `mmap` flags (`MAP_*`).
+     * \param offset  Offset in file.
+     *
+     * Maps the file at given address. This might involve mapping directly (`DkStreamMap`), or
+     * mapping anonymous memory (`DkVirtualMemoryAlloc`) and writing data.
+     *
+     * `addr`, `offset` and `size` must be alloc-aligned (see `IS_ALLOC_ALIGNED*` macros in
+     * `shim_internal.h`).
+     */
+    int (*mmap)(struct shim_handle* hdl, void* addr, size_t size, int prot, int flags,
                 uint64_t offset);
+
+    /*
+     * \brief Write back mapped memory to file.
+     *
+     * \param hdl     File handle.
+     * \param addr    Address of the memory region. Cannot be NULL.
+     * \param size    Size of the memory region.
+     * \param prot    Permissions for the memory region (`PROT_*`).
+     * \param flags   `mmap` flags (`MAP_*`).
+     * \param offset  Offset in file.
+     *
+     * Writes back any changes made by the user.
+     *
+     * The parameters should describe either a region originally mapped with `mmap` callback, or a
+     * part of that region. This function should only be called for a shared mapping, i.e. `flags`
+     * must contain `MAP_SHARED`.
+     */
+    int (*msync)(struct shim_handle* hdl, void* addr, size_t size, int prot, int flags,
+                 uint64_t offset);
 
     /* flush: flush out user buffer */
     int (*flush)(struct shim_handle* hdl);
@@ -888,6 +922,11 @@ int generic_inode_stat(struct shim_dentry* dent, struct stat* buf);
 int generic_inode_hstat(struct shim_handle* hdl, struct stat* buf);
 file_off_t generic_inode_seek(struct shim_handle* hdl, file_off_t offset, int origin);
 int generic_inode_poll(struct shim_handle* hdl, int poll_type);
+
+int generic_emulated_mmap(struct shim_handle* hdl, void* addr, size_t size, int prot, int flags,
+                          uint64_t offset);
+int generic_emulated_msync(struct shim_handle* hdl, void* addr, size_t size, int prot, int flags,
+                           uint64_t offset);
 
 int synthetic_setup_dentry(struct shim_dentry* dent, mode_t type, mode_t perm);
 
--- a/LibOS/shim/include/shim_vma.h
+++ b/LibOS/shim/include/shim_vma.h
@@ -124,6 +124,12 @@ void free_vma_info_array(struct shim_vma_info* vma_infos, size_t count);
 /* Implementation of madvise(MADV_DONTNEED) syscall */
 int madvise_dontneed_range(uintptr_t begin, uintptr_t end);
 
+/* Call `msync` for file mappings in given range (should be page-aligned) */
+int msync_range(uintptr_t begin, uintptr_t end);
+
+/* Call `msync` for file mappings of `hdl` */
+int msync_handle(struct shim_handle* hdl);
+
 void debug_print_all_vmas(void);
 
 #endif /* _SHIM_VMA_H_ */
--- a/LibOS/shim/src/bookkeep/shim_vma.c
+++ b/LibOS/shim/src/bookkeep/shim_vma.c
@@ -1106,18 +1106,18 @@ bool is_in_adjacent_user_vmas(const void* addr, size_t length, int prot) {
     return is_continuous && ctx.is_ok;
 }
 
-static size_t dump_all_vmas_with_buf(struct shim_vma_info* infos, size_t max_count,
-                                     bool include_unmapped) {
+static size_t dump_vmas_with_buf(struct shim_vma_info* infos, size_t max_count,
+                                 uintptr_t begin, uintptr_t end,
+                                 bool (*vma_filter)(struct shim_vma* vma, void* arg), void* arg) {
     size_t size = 0;
     struct shim_vma_info* vma_info = infos;
 
     spinlock_lock(&vma_tree_lock);
     struct shim_vma* vma;
 
-    for (vma = _get_first_vma(); vma; vma = _get_next_vma(vma)) {
-        if (vma->flags & ((include_unmapped ? 0 : VMA_UNMAPPED) | VMA_INTERNAL)) {
+    for (vma = _lookup_vma(begin); vma && vma->begin < end; vma = _get_next_vma(vma)) {
+        if (!vma_filter(vma, arg))
             continue;
-        }
         if (size < max_count) {
             dump_vma(vma_info, vma);
             vma_info++;
@@ -1130,7 +1130,9 @@ static size_t dump_all_vmas_with_buf(struct shim_vma_info* infos, size_t max_cou
     return size;
 }
 
-int dump_all_vmas(struct shim_vma_info** ret_infos, size_t* ret_count, bool include_unmapped) {
+static int dump_vmas(struct shim_vma_info** ret_infos, size_t* ret_count,
+                     uintptr_t begin, uintptr_t end,
+                     bool (*vma_filter)(struct shim_vma* vma, void* arg), void* arg) {
     size_t count = DEFAULT_VMA_COUNT;
 
     while (true) {
@@ -1139,7 +1141,7 @@ int dump_all_vmas(struct shim_vma_info** ret_infos, size_t* ret_count, bool incl
             return -ENOMEM;
         }
 
-        size_t needed_count = dump_all_vmas_with_buf(vmas, count, include_unmapped);
+        size_t needed_count = dump_vmas_with_buf(vmas, count, begin, end, vma_filter, arg);
         if (needed_count <= count) {
             *ret_infos = vmas;
             *ret_count = needed_count;
@@ -1149,6 +1151,26 @@ int dump_all_vmas(struct shim_vma_info** ret_infos, size_t* ret_count, bool incl
         free_vma_info_array(vmas, count);
         count = needed_count;
     }
+}
+
+static bool vma_filter_all(struct shim_vma* vma, void* arg) {
+    assert(spinlock_is_locked(&vma_tree_lock));
+    __UNUSED(arg);
+
+    return !(vma->flags & VMA_INTERNAL);
+}
+
+static bool vma_filter_exclude_unmapped(struct shim_vma* vma, void* arg) {
+    assert(spinlock_is_locked(&vma_tree_lock));
+    __UNUSED(arg);
+
+    return !(vma->flags & (VMA_INTERNAL | VMA_UNMAPPED));
+}
+
+int dump_all_vmas(struct shim_vma_info** ret_infos, size_t* ret_count, bool include_unmapped) {
+    return dump_vmas(ret_infos, ret_count, /*begin=*/0, /*end=*/UINTPTR_MAX,
+                     include_unmapped ? vma_filter_all : vma_filter_exclude_unmapped,
+                     /*arg=*/NULL);
 }
 
 void free_vma_info_array(struct shim_vma_info* vma_infos, size_t count) {
@@ -1213,6 +1235,77 @@ int madvise_dontneed_range(uintptr_t begin, uintptr_t end) {
     return ctx.error;
 }
 
+static bool vma_filter_needs_msync(struct shim_vma* vma, void* arg) {
+    struct shim_handle* hdl = arg;
+
+    if (vma->flags & (VMA_UNMAPPED | VMA_INTERNAL | MAP_ANONYMOUS | MAP_PRIVATE))
+        return false;
+
+    assert(vma->file);
+
+    if (hdl && vma->file != hdl)
+        return false;
+
+    if (!vma->file->fs || !vma->file->fs->fs_ops || !vma->file->fs->fs_ops->msync)
+        return false;
+
+    /*
+     * XXX: Strictly speaking, reading `vma->file->acc_mode` requires taking `vma->file->lock`,
+     * which we cannot do in this function. However, the `acc_mode` field is only modified for
+     * sockets, and at this point we know that `vma->file` is not a socket (since it implements
+     * `msync`).
+     *
+     * TODO: Remove this comment when the socket code is rewritten.
+     */
+    if (!(vma->file->acc_mode & MAY_WRITE))
+        return false;
+
+    return true;
+}
+
+static int msync_all(uintptr_t begin, uintptr_t end, struct shim_handle* hdl) {
+    assert(IS_ALLOC_ALIGNED(begin));
+    assert(end == UINTPTR_MAX || IS_ALLOC_ALIGNED(end));
+
+    struct shim_vma_info* vma_infos;
+    size_t count;
+
+    int ret = dump_vmas(&vma_infos, &count, begin, end, vma_filter_needs_msync, hdl);
+    if (ret < 0)
+        return ret;
+
+    for (size_t i = 0; i < count; i++) {
+        struct shim_vma_info* vma_info = &vma_infos[i];
+
+        struct shim_handle* file = vma_info->file;
+        assert(file && file->fs && file->fs->fs_ops && file->fs->fs_ops->msync);
+
+        /* NOTE: Unfortunately there's a data race here: the memory can be unmapped, or remapped, by
+         * another thread by the time we get to `msync`. */
+        uintptr_t msync_begin = MAX(begin, (uintptr_t)vma_info->addr);
+        uintptr_t msync_end = MIN(end, (uintptr_t)vma_info->addr + vma_info->length);
+        assert(IS_ALLOC_ALIGNED(msync_begin));
+        assert(IS_ALLOC_ALIGNED(msync_end));
+
+        ret = file->fs->fs_ops->msync(file, (void*)msync_begin, msync_end - msync_begin,
+                                      vma_info->prot, vma_info->flags, vma_info->file_offset);
+        if (ret < 0)
+            goto out;
+    }
+
+    ret = 0;
+out:
+    free_vma_info_array(vma_infos, count);
+    return ret;
+}
+
+int msync_range(uintptr_t begin, uintptr_t end) {
+    return msync_all(begin, end, /*hdl=*/NULL);
+}
+
+int msync_handle(struct shim_handle* hdl) {
+    return msync_all(/*begin=*/0, /*end=*/UINTPTR_MAX, hdl);
+}
 
 BEGIN_CP_FUNC(vma) {
     __UNUSED(size);
@@ -1311,13 +1404,10 @@ BEGIN_RS_FUNC(vma) {
             if (!fs || !fs->fs_ops || !fs->fs_ops->mmap)
                 return -EINVAL;
 
-            void* addr = vma->addr;
-            int ret = fs->fs_ops->mmap(vma->file, &addr, vma->length, vma->prot,
+            int ret = fs->fs_ops->mmap(vma->file, vma->addr, vma->length, vma->prot,
                                        vma->flags | MAP_FIXED, vma->file_offset);
             if (ret < 0)
                 return ret;
-
-            assert(addr == vma->addr);
         }
     }
 }
--- a/LibOS/shim/src/fs/chroot/encrypted.c
+++ b/LibOS/shim/src/fs/chroot/encrypted.c
@@ -23,11 +23,11 @@
  *   `shim_encrypted_file` object for its inode. As a result, multiple handles for a given file
  *   still correspond to one PAL handle.
  *
- * TODO (most items are needed for feature parity with PAL protected files):
+ * TODO:
  *
- * - mmap
  * - truncate (the current `truncate` operation works only for extending the file, support for
  *   truncation needs to be added to the `protected_files` module)
+ * - flush all files on process exit
  */
 
 #define _POSIX_C_SOURCE 200809L /* for SSIZE_MAX */
@@ -36,6 +36,7 @@
 #include "perm.h"
 #include "shim_fs.h"
 #include "shim_fs_encrypted.h"
+#include "shim_vma.h"
 #include "stat.h"
 #include "toml_utils.h"
 
@@ -370,8 +371,14 @@ static int chroot_encrypted_flush(struct shim_handle* hdl) {
 
     struct shim_encrypted_file* enc = hdl->inode->data;
 
+    /* If there are any MAP_SHARED mappings for the file, this will write data to `enc` */
+    int ret = msync_handle(hdl);
+    if (ret < 0)
+        return ret;
+
+    /* This will write changes from `enc` to host file */
     lock(&hdl->inode->lock);
-    int ret = encrypted_file_flush(enc);
+    ret = encrypted_file_flush(enc);
     unlock(&hdl->inode->lock);
     return ret;
 }
@@ -458,6 +465,8 @@ struct shim_fs_ops chroot_encrypted_fs_ops = {
     .hdrop      = &chroot_encrypted_hdrop,
     .checkpoint = &chroot_encrypted_checkpoint,
     .migrate    = &chroot_encrypted_migrate,
+    .mmap       = &generic_emulated_mmap,
+    .msync      = &generic_emulated_msync,
 };
 
 struct shim_d_ops chroot_encrypted_d_ops = {
--- a/LibOS/shim/src/fs/chroot/encrypted.c
+++ b/LibOS/shim/src/fs/chroot/encrypted.c
@@ -396,7 +396,7 @@ static ssize_t chroot_encrypted_read(struct shim_handle* hdl, void* buf, size_t 
     size_t actual_count;
 
     lock(&hdl->inode->lock);
-    int ret = encrypted_file_read(enc, buf, count, hdl->pos, &actual_count);
+    int ret = encrypted_file_read(enc, buf, count, *pos, &actual_count);
     unlock(&hdl->inode->lock);
 
     if (ret < 0)
@@ -416,7 +416,7 @@ static ssize_t chroot_encrypted_write(struct shim_handle* hdl, const void* buf, 
 
     lock(&hdl->inode->lock);
 
-    int ret = encrypted_file_write(enc, buf, count, hdl->pos, &actual_count);
+    int ret = encrypted_file_write(enc, buf, count, *pos, &actual_count);
     if (ret < 0)
         goto out;
 
--- a/LibOS/shim/src/fs/chroot/fs.c
+++ b/LibOS/shim/src/fs/chroot/fs.c
@@ -322,17 +322,23 @@ static ssize_t chroot_write(struct shim_handle* hdl, const void* buf, size_t cou
     return actual_count;
 }
 
-static int chroot_mmap(struct shim_handle* hdl, void** addr, size_t size, int prot, int flags,
+static int chroot_mmap(struct shim_handle* hdl, void* addr, size_t size, int prot, int flags,
                        uint64_t offset) {
     assert(hdl->type == TYPE_CHROOT);
+    assert(addr);
 
     pal_prot_flags_t pal_prot = LINUX_PROT_TO_PAL(prot, flags);
 
     if (flags & MAP_ANONYMOUS)
         return -EINVAL;
 
-    int ret = DkStreamMap(hdl->pal_handle, addr, pal_prot, offset, size);
-    return pal_to_unix_errno(ret);
+    void* actual_addr = addr;
+    int ret = DkStreamMap(hdl->pal_handle, &actual_addr, pal_prot, offset, size);
+    if (ret < 0)
+        return pal_to_unix_errno(ret);
+
+    assert(actual_addr == addr);
+    return 0;
 }
 
 static int chroot_truncate(struct shim_handle* hdl, file_off_t size) {
--- a/LibOS/shim/src/fs/shim_fs_mem.c
+++ b/LibOS/shim/src/fs/shim_fs_mem.c
@@ -7,12 +7,18 @@
 #include "shim_fs.h"
 #include "shim_fs_mem.h"
 
+/* Minimum allocation size in `slabmgr.h` */
+#define MIN_RESIZE ((size_t)16)
+
 static int mem_file_resize(struct shim_mem_file* mem, size_t buf_size) {
+    assert(buf_size > 0);
+
     char* buf = malloc(buf_size);
     if (!buf)
         return -ENOMEM;
 
     memcpy(buf, mem->buf, MIN(buf_size, (size_t)mem->size));
+
     free(mem->buf);
     mem->buf = buf;
     mem->buf_size = buf_size;
@@ -56,8 +62,8 @@ ssize_t mem_file_write(struct shim_mem_file* mem, file_off_t pos_start, const vo
         return -EFBIG;
 
     if (size > 0) {
-        if ((size_t)pos_end > mem->buf_size) {
-            size_t buf_size = MAX(mem->buf_size, 1U);
+        if (mem->buf_size < (size_t)pos_end) {
+            size_t buf_size = MAX(mem->buf_size, MIN_RESIZE);
             while (buf_size < (size_t)pos_end)
                 if (__builtin_mul_overflow(buf_size, 2, &buf_size))
                     return -EFBIG;
@@ -81,9 +87,27 @@ int mem_file_truncate(struct shim_mem_file* mem, file_off_t size) {
     if (OVERFLOWS(size_t, size))
         return -EFBIG;
 
-    int ret = mem_file_resize(mem, size);
-    if (ret < 0)
-        return ret;
+    size_t buf_size = mem->buf_size;
+    if (buf_size < (size_t)size) {
+        buf_size = MAX(buf_size, MIN_RESIZE);
+        while (buf_size < (size_t)size) {
+            if (__builtin_mul_overflow(buf_size, 2, &buf_size))
+                return -EFBIG;
+        }
+    } else if (size > 0) {
+        while (buf_size / 2 >= (size_t)size)
+            buf_size /= 2;
+        buf_size = MAX(buf_size, MIN_RESIZE);
+    } else {
+        buf_size = MIN_RESIZE;
+    }
+
+    assert((size_t)size <= buf_size);
+    if (buf_size != mem->buf_size) {
+        int ret = mem_file_resize(mem, buf_size);
+        if (ret < 0)
+            return ret;
+    }
     mem->size = size;
     return 0;
 }
--- a/LibOS/shim/src/fs/shim_fs_util.c
+++ b/LibOS/shim/src/fs/shim_fs_util.c
@@ -4,6 +4,7 @@
  */
 
 #include "cpu.h"
+#include "shim_flags_conv.h"
 #include "shim_fs.h"
 #include "shim_lock.h"
 #include "stat.h"
@@ -133,5 +134,114 @@ int generic_inode_poll(struct shim_handle* hdl, int poll_type) {
 
     unlock(&hdl->inode->lock);
     unlock(&hdl->pos_lock);
+    return ret;
+}
+
+int generic_emulated_mmap(struct shim_handle* hdl, void* addr, size_t size, int prot, int flags,
+                          uint64_t offset) {
+    assert(addr);
+
+    int ret;
+
+    pal_prot_flags_t pal_prot = LINUX_PROT_TO_PAL(prot, flags);
+    pal_prot_flags_t pal_prot_writable = pal_prot | PAL_PROT_WRITE;
+
+    void* actual_addr = addr;
+    ret = DkVirtualMemoryAlloc(&actual_addr, size, /*alloc_type=*/0, pal_prot_writable);
+    if (ret < 0)
+        return pal_to_unix_errno(ret);
+
+    assert(actual_addr == addr);
+
+    size_t read_size = size;
+    char* read_addr = addr;
+    file_off_t pos = offset;
+    while (read_size > 0) {
+        ssize_t count = hdl->fs->fs_ops->read(hdl, read_addr, read_size, &pos);
+        if (count < 0) {
+            if (count == -EINTR)
+                continue;
+            ret = count;
+            goto err;
+        }
+
+        if (count == 0)
+            break;
+
+        assert((size_t)count <= read_size);
+        read_size -= count;
+        read_addr += count;
+    }
+
+    if (pal_prot != pal_prot_writable) {
+        ret = DkVirtualMemoryProtect(addr, size, pal_prot);
+        if (ret < 0) {
+            ret = pal_to_unix_errno(ret);
+            goto err;
+        }
+    }
+
+    return 0;
+
+err:;
+    int free_ret = DkVirtualMemoryFree(addr, size);
+    if (free_ret < 0) {
+        log_debug("%s: DkVirtualMemoryFree failed on cleanup: %d", __func__, free_ret);
+        BUG();
+    }
+    return ret;
+}
+
+int generic_emulated_msync(struct shim_handle* hdl, void* addr, size_t size, int prot, int flags,
+                           uint64_t offset) {
+    assert(!(flags & MAP_PRIVATE));
+
+    lock(&hdl->inode->lock);
+    file_off_t file_size = hdl->inode->size;
+    unlock(&hdl->inode->lock);
+
+    pal_prot_flags_t pal_prot = LINUX_PROT_TO_PAL(prot, flags);
+    pal_prot_flags_t pal_prot_readable = pal_prot | PAL_PROT_READ;
+
+    int ret;
+    if (pal_prot != pal_prot_readable) {
+        ret = DkVirtualMemoryProtect(addr, size, pal_prot_readable);
+        if (ret < 0)
+            return pal_to_unix_errno(ret);
+    }
+
+    size_t write_size = offset > (uint64_t)file_size ? 0 : MIN(size, (uint64_t)file_size - offset);
+    char* write_addr = addr;
+    file_off_t pos = offset;
+    while (write_size > 0) {
+        ssize_t count = hdl->fs->fs_ops->write(hdl, write_addr, write_size, &pos);
+        if (count < 0) {
+            if (count == -EINTR)
+                continue;
+            ret = count;
+            goto out;
+        }
+
+        if (count == 0) {
+            log_debug("%s: Failed to write back the whole mapping", __func__);
+            ret = -EIO;
+            goto out;
+        }
+
+        assert((size_t)count <= write_size);
+        write_size -= count;
+        write_addr += count;
+    }
+
+    ret = 0;
+
+out:
+    if (pal_prot != pal_prot_readable) {
+        int protect_ret = DkVirtualMemoryProtect(addr, size, pal_prot);
+        if (protect_ret < 0) {
+            log_debug("%s: DkVirtualMemoryProtect failed on cleanup: %d", __func__, protect_ret);
+            BUG();
+        }
+    }
     return ret;
 }
--- a/LibOS/shim/src/fs/tmpfs/fs.c
+++ b/LibOS/shim/src/fs/tmpfs/fs.c
@@ -20,6 +20,7 @@
 #include "shim_fs.h"
 #include "shim_handle.h"
 #include "shim_lock.h"
+#include "shim_vma.h"
 #include "stat.h"
 
 #define USEC_IN_SEC 1000000
@@ -113,8 +114,7 @@ static int tmpfs_mount(struct shim_mount_params* params, void** mount_data) {
 }
 
 static int tmpfs_flush(struct shim_handle* hdl) {
-    __UNUSED(hdl);
-    return 0;
+    return msync_handle(hdl);
 }
 
 static int tmpfs_lookup(struct shim_dentry* dent) {
@@ -289,33 +289,17 @@ out:
     return ret;
 }
 
-/* TODO: tmpfs_mmap() function is not implemented because shim_do_mmap() and shim_do_munmap()
-   are currently not flexible enough for correct tmpfs implementation. In particular, shim_do_mmap()
-   pre-allocates memory region at a specific address (making it impossible to have two mmaps on the
-   same tmpfs file), and shim_do_munmap() doesn't have a callback into tmpfs at all. */
-static int tmpfs_mmap(struct shim_handle* hdl, void** addr, size_t size, int prot, int flags,
-                      uint64_t offset) {
-    __UNUSED(hdl);
-    __UNUSED(addr);
-    __UNUSED(size);
-    __UNUSED(prot);
-    __UNUSED(flags);
-    __UNUSED(offset);
-
-    log_error("tmpfs_mmap(): mmap() function for tmpfs mount type is not implemented.");
-    return -ENOSYS;
-}
-
 struct shim_fs_ops tmp_fs_ops = {
     .mount    = &tmpfs_mount,
     .flush    = &tmpfs_flush,
     .read     = &tmpfs_read,
     .write    = &tmpfs_write,
-    .mmap     = &tmpfs_mmap,
     .seek     = &generic_inode_seek,
     .hstat    = &generic_inode_hstat,
     .truncate = &tmpfs_truncate,
     .poll     = &generic_inode_poll,
+    .mmap     = &generic_emulated_mmap,
+    .msync    = &generic_emulated_msync,
 };
 
 struct shim_d_ops tmp_d_ops = {
--- a/LibOS/shim/src/shim_rtld.c
+++ b/LibOS/shim/src/shim_rtld.c
@@ -261,7 +261,7 @@ static int execute_loadcmd(const struct loadcmd* c, elf_addr_t base_diff,
             return ret;
         }
 
-        if ((ret = file->fs->fs_ops->mmap(file, &map_start, map_size, c->prot, map_flags,
+        if ((ret = file->fs->fs_ops->mmap(file, map_start, map_size, c->prot, map_flags,
                                           c->map_off)) < 0) {
             log_debug("%s: failed to map segment: %d", __func__, ret);
             return ret;
--- a/LibOS/shim/src/sys/shim_mmap.c
+++ b/LibOS/shim/src/sys/shim_mmap.c
@@ -137,6 +137,13 @@ void* shim_do_mmap(void* addr, size_t length, int prot, int flags, int fd, unsig
             ret = -EINVAL;
             goto out_handle;
         }
+        if (!(flags & MAP_FIXED_NOREPLACE)) {
+            /* Flush any file mappings we're about to replace */
+            ret = msync_range((uintptr_t)addr, (uintptr_t)addr + length);
+            if (ret < 0) {
+                goto out_handle;
+            }
+        }
         ret = bkeep_mmap_fixed(addr, length, prot, flags, hdl, offset, NULL);
         if (ret < 0) {
             goto out_handle;
@@ -174,12 +181,7 @@ void* shim_do_mmap(void* addr, size_t length, int prot, int flags, int fd, unsig
             }
         }
     } else {
-        void* ret_addr = addr;
-        ret = hdl->fs->fs_ops->mmap(hdl, &ret_addr, length, prot, flags, offset);
-        if (ret_addr != addr) {
-            log_error("Requested address (%p) differs from allocated (%p)!", addr, ret_addr);
-            BUG();
-        }
+        ret = hdl->fs->fs_ops->mmap(hdl, addr, length, prot, flags, offset);
     }
 
     if (ret < 0) {
@@ -280,8 +282,16 @@ long shim_do_munmap(void* addr, size_t length) {
     if (!IS_ALLOC_ALIGNED(length))
         length = ALLOC_ALIGN_UP(length);
 
+    int ret;
+
+    /* Flush any file mappings we're about to remove */
+    ret = msync_range((uintptr_t)addr, (uintptr_t)addr + length);
+    if (ret < 0) {
+        return ret;
+    }
+
     void* tmp_vma = NULL;
-    int ret = bkeep_munmap(addr, length, /*is_internal=*/false, &tmp_vma);
+    ret = bkeep_munmap(addr, length, /*is_internal=*/false, &tmp_vma);
     if (ret < 0) {
         return ret;
     }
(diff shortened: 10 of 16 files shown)


dimakuv (changes requested):
> Reviewed 16 of 17 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/550)* status: all files reviewed, 18 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[`-- commits` line 9 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0dn46I6D47pt44YEgZ):*
`on an` -> `on`
___
*[`LibOS/shim/include/shim_fs.h` line 113 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0dpfUZ76to_szX7vcx:-N0dpfUZ76to_szX7vcy:ba5r8bq) ([raw file](https://github.com/gramineproject/gramine/blob/9c41ea9c475fc3622cc8ea9bf0aaf7b49e688366/LibOS/shim/include/shim_fs.h#L113)):*
> ```objc
>      */
>     int (*msync)(struct shim_handle* hdl, void* addr, size_t size, int prot, int flags,
>                 uint64_t offset);
> ```

Why do we need `prot` argument here? UPDATE: Ok, because we can have a write-only protection and we want to read the contents of the VMA to write them into the file. So we memorize the original protections, then add R and perform msync, then restore original protections.
___
*[`LibOS/shim/src/bookkeep/shim_vma.c` line 1266 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0dtRzF1VfJgdG8kbgT:-N0dtRzF1VfJgdG8kbgU:bfzd0t4) ([raw file](https://github.com/gramineproject/gramine/blob/9c41ea9c475fc3622cc8ea9bf0aaf7b49e688366/LibOS/shim/src/bookkeep/shim_vma.c#L1266)):*
> ```C
> }
> 
> static int _msync_all(uintptr_t begin, uintptr_t end, struct shim_handle* hdl) {
> ```

Looks like a potentially very slow operation because it iterates through *all* VMAs every time it's called. I guess nothing can be done about it, but do you think it can be a perf bottleneck? I can imagine big applications with thousands of VMAs allocated...
___
*[`LibOS/shim/src/bookkeep/shim_vma.c` line 1285 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0dsdSfC7hDV9OcvCr4:-N0dsdSfC7hDV9OcvCr5:b-tb9y3o) ([raw file](https://github.com/gramineproject/gramine/blob/9c41ea9c475fc3622cc8ea9bf0aaf7b49e688366/LibOS/shim/src/bookkeep/shim_vma.c#L1285)):*
> ```C
>         uintptr_t msync_end = MIN(end, (uintptr_t)vma_info->addr + vma_info->length);
>         ret = file->fs->fs_ops->msync(file, (void*)msync_begin, msync_end - msync_begin,
>                                       vma_info->prot, vma_info->flags, vma_info->file_offset);
> ```

It's unclear to me if any data races are possible here. Say you collected the relevant VMAs into the `vma_infos` array and started calling `msync()` on each of these VMAs. What if another thread unmapped this VMA, mapped another VMA in the same memory range, and now your `msync()` writes wrong values back into the file?

Hm, on the other hand, looks like it's normal behavior of Linux, where no guarantees for such cases are given for `msync`.
___
*[`LibOS/shim/src/fs/shim_fs_mem.c` line 98 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0dod_98Z0z4JdgD5bn:-N0dod_A4U2DYzhJQq3q:ba7c6xj) ([raw file](https://github.com/gramineproject/gramine/blob/9c41ea9c475fc3622cc8ea9bf0aaf7b49e688366/LibOS/shim/src/fs/shim_fs_mem.c#L98)):*
> ```C
>     assert((size_t)size <= buf_size);
>     if (buf_size != mem->buf_size) {
>         int ret = mem_file_resize(mem, buf_size);
> ```

Actually, what happens if the user supplies `size = 0`? Won't we end up with `buf_size = 0` and then `mem_file_resize()` will fail with `-ENOMEM`? Which is not how truncate is supposed to work.
___
*[`LibOS/shim/src/fs/shim_fs_util.c` line 173 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0dw8JE0PoH4ms-b7-x:-N0dw8JE0PoH4ms-b7-y:blxay93) ([raw file](https://github.com/gramineproject/gramine/blob/9c41ea9c475fc3622cc8ea9bf0aaf7b49e688366/LibOS/shim/src/fs/shim_fs_util.c#L173)):*
> ```C
>         assert((size_t)count <= read_size);
>         read_size -= count;
>         read_addr += count;
> ```

Technically we shouldn't use `void*` because it doesn't allow pointer arithmetic. We should technically use `char*`, but I don't insist.
___
*[`LibOS/shim/src/fs/shim_fs_util.c` line 195 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0dwqjr05liiZfmX5nl:-N0dwqjr05liiZfmX5nm:b-k570fe) ([raw file](https://github.com/gramineproject/gramine/blob/9c41ea9c475fc3622cc8ea9bf0aaf7b49e688366/LibOS/shim/src/fs/shim_fs_util.c#L195)):*
> ```C
> int generic_emulated_msync(struct shim_handle* hdl, void* addr, size_t size, int prot, int flags,
>                            uint64_t offset) {
>     assert(!(flags & MAP_PRIVATE));
> ```

Well, you could add more flags here, e.g. `ANONYMOUS`. Why did you decide on this particular assert?
___
*[`LibOS/shim/src/fs/shim_fs_util.c` line 222 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0dxcI8Ep5yHB2ddB9R:-N0dxcI8Ep5yHB2ddB9S:br0ijio) ([raw file](https://github.com/gramineproject/gramine/blob/9c41ea9c475fc3622cc8ea9bf0aaf7b49e688366/LibOS/shim/src/fs/shim_fs_util.c#L222)):*
> ```C
>         }
> 
>         if (count == 0)
> ```

Can this happen?
___
*[`LibOS/shim/src/fs/shim_fs_util.c` line 227 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0e1bdwEKXK3OsTkwwY:-N0e1bdwEKXK3OsTkwwZ:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/9c41ea9c475fc3622cc8ea9bf0aaf7b49e688366/LibOS/shim/src/fs/shim_fs_util.c#L227)):*
> ```C
>         assert((size_t)count <= write_size);
>         write_size -= count;
>         write_addr += count;
> ```

ditto
___
*[`LibOS/shim/src/fs/chroot/encrypted.c` line 31 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0dxxjE1y3c3sln-IS2:-N0dxxjE1y3c3sln-IS3:b-1r30gf) ([raw file](https://github.com/gramineproject/gramine/blob/9c41ea9c475fc3622cc8ea9bf0aaf7b49e688366/LibOS/shim/src/fs/chroot/encrypted.c#L31)):*
> ```C
>  * - truncate (the current `truncate` operation works only for extending the file, support for
>  *   truncation needs to be added to the `protected_files` module)
>  * - flush all files on process exit
> ```

Why didn't you implement it in this PR? It will be a follow-up?
___
*[`LibOS/shim/src/fs/chroot/encrypted.c` line 373 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0dyOon7_KbV8kRvKCt:-N0dyOon7_KbV8kRvKCu:b-6t68v8) ([raw file](https://github.com/gramineproject/gramine/blob/9c41ea9c475fc3622cc8ea9bf0aaf7b49e688366/LibOS/shim/src/fs/chroot/encrypted.c#L373)):*
> ```C
>     assert(hdl->inode->type == S_IFREG);
> 
>     int ret = msync_handle(hdl);
> ```

I'm confused here. Why do we have both `msync_handle()` and `encrypted_file_flush()`? Aren't they doing the same?
___
*[`LibOS/shim/src/fs/chroot/encrypted.c` line 405 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0dmV8g8j6MAb67ggxh:-N0dmV8g8j6MAb67ggxi:b-qwa7gh) ([raw file](https://github.com/gramineproject/gramine/blob/9c41ea9c475fc3622cc8ea9bf0aaf7b49e688366/LibOS/shim/src/fs/chroot/encrypted.c#L405)):*
> ```C
> 
>     lock(&hdl->inode->lock);
>     int ret = encrypted_file_read(enc, buf, count, *pos, &actual_count);
> ```

Ouch. So this bug went unnoticed because the only user of this callback was `shim_do_read()` (which anyhow supplies `&hdl->pos` in the `pos` argument)? And now you added the `mmap` emulation that uses this callback too, but from an arbitrary `pos`, and that's how you exposed the bug?

Same question for the bug in `chroot_encrypted_write()` below.
___
*[`LibOS/shim/src/fs/tmpfs/fs.c` line 295 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0dzAQYCMVlZuV1EX1O:-N0dzAQYCMVlZuV1EX1P:b7h4847) ([raw file](https://github.com/gramineproject/gramine/blob/9774ce1ca8551b868c1ca6c423dfcf1b010115d4/LibOS/shim/src/fs/tmpfs/fs.c#L295)):*
> ```C
>    are currently not flexible enough for correct tmpfs implementation. In particular, shim_do_mmap()
>    pre-allocates memory region at a specific address (making it impossible to have two mmaps on the
>    same tmpfs file), and shim_do_munmap() doesn't have a callback into tmpfs at all. */
> ```

Actually, is it possible now to have two mmaps on the same tmpfs file? Did we change anything that enables this scenario now?

And since we're here, could we add such a case in our tests?
___
*[`LibOS/shim/src/sys/shim_mmap.c` line 459 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0e-asf8UW_QaF5TLIK:-N0e-asgDV5N7Nfl4JIL:b-a6t8x4) ([raw file](https://github.com/gramineproject/gramine/blob/9c41ea9c475fc3622cc8ea9bf0aaf7b49e688366/LibOS/shim/src/sys/shim_mmap.c#L459)):*
> ```C
> 
>     if (flags & MS_SYNC) {
>         return msync_range(start, start + len);
> ```

Nice side-effect!
___
*[`LibOS/shim/test/fs/test_tmpfs.py` line 69 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0e-ht24VfW__MZ824t:-N0e-ht24VfW__MZ824u:b-uqcr67) ([raw file](https://github.com/gramineproject/gramine/blob/9c41ea9c475fc3622cc8ea9bf0aaf7b49e688366/LibOS/shim/test/fs/test_tmpfs.py#L69)):*
> ```Python
>         pass
> 
>     # This overrides parent class to remove @expectedFailureIf(HAS_SGX)
> ```

What's the deal with this `expectedFailureIf`? Why would the parent class have it at all? Isn't it better to change the parent class and remove this attribute completely? After all, now all mmaps should work. And then we won't need to overwrite these methods (well, maybe we'll need to adjust the timeouts, but it may be done in the parent class).
___
*[`LibOS/shim/test/regression/mmap_file_emulated.c` line 7 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0e0JnU-hyLvpDQoBRa:-N0e0JnU-hyLvpDQoBRb:b-boo2a5) ([raw file](https://github.com/gramineproject/gramine/blob/9c41ea9c475fc3622cc8ea9bf0aaf7b49e688366/LibOS/shim/test/regression/mmap_file_emulated.c#L7)):*
> ```C
> 
> /*
>  * Test file mapping emulated by Gramine (encrypted: tmpfs): try reading and writing a file through
> ```

`ecnrypted, tmpfs` (comma instead of colon)
___
*[`LibOS/shim/test/regression/mmap_file_emulated.c` line 35 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0e0_Zp3E8Eaf6yg16b:-N0e0_Zq1aR8_64-bYwi:bqg6ij8) ([raw file](https://github.com/gramineproject/gramine/blob/9c41ea9c475fc3622cc8ea9bf0aaf7b49e688366/LibOS/shim/test/regression/mmap_file_emulated.c#L35)):*
> ```C
>     const char* path = argv[1];
> 
>     setbuf(stdout, NULL);
> ```

Why do we care?
___
*[`LibOS/shim/test/regression/mmap_file_emulated.c` line 67 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0e0mjl25kQMtT17mw4:-N0e0mjl25kQMtT17mw5:b-5zke8k) ([raw file](https://github.com/gramineproject/gramine/blob/9c41ea9c475fc3622cc8ea9bf0aaf7b49e688366/LibOS/shim/test/regression/mmap_file_emulated.c#L67)):*
> ```C
>     if (fd == -1)
>         err(1, "open");
> ```

Please remove one empty line
___
*[`LibOS/shim/test/regression/mmap_file_emulated.c` line 69 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0e0u2vBTZIYe9yquGx:-N0e0u2vBTZIYe9yquGy:b-5wtby4) ([raw file](https://github.com/gramineproject/gramine/blob/9c41ea9c475fc3622cc8ea9bf0aaf7b49e688366/LibOS/shim/test/regression/mmap_file_emulated.c#L69)):*
> ```C
> 
> 
>     void* addr = mmap(NULL, mmap_size, PROT_READ | PROT_WRITE, MAP_SHARED, fd, 0);
> ```

Actually, a test with offset != 0 would be also good.

dimakuv (changes requested):
> Reviewed 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/550)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "squash! " and "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, and @pwmarcz)

___
*[`LibOS/shim/src/bookkeep/shim_vma.c` line 1266 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0dtRzF1VfJgdG8kbgT:-N0f4EotAu6KFvMnn6uG:b-ee5lx6) ([raw file](https://github.com/gramineproject/gramine/blob/9c41ea9c475fc3622cc8ea9bf0aaf7b49e688366/LibOS/shim/src/bookkeep/shim_vma.c#L1266)):*

You're right. I used the function for dumping *all* VMAs, but it can be adapted to dump all VMAs in range (and they're stored in an AVL tree, so lookup will be cheap). Fixed.
</blockquote></details>

Awesome solution!
___
*[`LibOS/shim/src/bookkeep/shim_vma.c` line 1285 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0dsdSfC7hDV9OcvCr4:-N0f4LKC6ooxsq8xdxiO:b-nuglgy) ([raw file](https://github.com/gramineproject/gramine/blob/9c41ea9c475fc3622cc8ea9bf0aaf7b49e688366/LibOS/shim/src/bookkeep/shim_vma.c#L1285)):*

Yes, unfortunately... This `msync` could write wrong values, or even crash if the memory is unmapped by another thread.

I don't see an easy way around it: VMA operations take a lock, but you cannot do anything while holding that lock (since even `malloc` might trigger mapping more memory).
</blockquote></details>

Yep. Anyway, resolved.
___
*[`LibOS/shim/src/fs/chroot/encrypted.c` line 31 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0dxxjE1y3c3sln-IS2:-N0f50wbD4RTnRkbP_SR:b-1hcu2r) ([raw file](https://github.com/gramineproject/gramine/blob/9c41ea9c475fc3622cc8ea9bf0aaf7b49e688366/LibOS/shim/src/fs/chroot/encrypted.c#L31)):*

Actually, I'm not sure how to even do that. I would like to call `flush()` on all the handles in the process, but do we have a list of them? There is the "handle map" which maps FDs to handles, but we might keep some open handles _without_ FDs in the memory mappings (if someone did `mmap()` and then `close()`).
</blockquote></details>

Oh no. @boryspoplawski @mkow Do you know how we can collect all handles, even the ones without an atteched FD?

But anyway, I think this is for a follow-up PR.
___
*[`LibOS/shim/src/fs/chroot/encrypted.c` line 373 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0dyOon7_KbV8kRvKCt:-N0f6DRw-mB6znSCbfwQ:b-tkbjoj) ([raw file](https://github.com/gramineproject/gramine/blob/9c41ea9c475fc3622cc8ea9bf0aaf7b49e688366/LibOS/shim/src/fs/chroot/encrypted.c#L373)):*

`flush` is a callback for the `sync` system call, and it's supposed to also flush file mappings. We could call `chroot_encrypted_msync` here, but we don't know what mappings we should call it on (and I thought it would be better NOT to duplicate this information in handle). So we call `msync_handle`, means basically "find all shared mappings of this file and call `msync` on them".
</blockquote></details>

Ok. Pawel's explanation didn't help me much, but I explained it to myself like this:

- `msync_handle()` ends up in `ipf_write()` -- which performs memcpy from mmapped memory region into the corresponding nodes of the protected/encrypted file (i.e., from the mmap memory region into the file-blocks-in-plaintext memory region)
- `encrypted_file_flush()` ends up in `ipf_internal_flush()` -- which performs writes of file-blocks-in-plaintext into the host-level files.

So we have two levels of indirection here: first sync from mmap region to file-blocks region, then sync from file-blocks region to host-level file. That's why two functions are used together.

Maybe will be helpful for other reviewers or for future.
___
*[`LibOS/shim/src/fs/tmpfs/fs.c` line 295 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0dzAQYCMVlZuV1EX1O:-N0f7Hoi1zN9DHN6zijP:b-f0b3ft) ([raw file](https://github.com/gramineproject/gramine/blob/9774ce1ca8551b868c1ca6c423dfcf1b010115d4/LibOS/shim/src/fs/tmpfs/fs.c#L295)):*

It's possible, but unlike Linux, they will not be synchronized: if you modify one, the other will stay the same.

I added a second mapping to the test.
</blockquote></details>

Wait, but is it Ok?
___
*[`LibOS/shim/test/regression/mmap_file_emulated.c` line 69 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0e0u2vBTZIYe9yquGx:-N0f85Ww9vfaJsT3q9J3:b-z27p04) ([raw file](https://github.com/gramineproject/gramine/blob/9c41ea9c475fc3622cc8ea9bf0aaf7b49e688366/LibOS/shim/test/regression/mmap_file_emulated.c#L69)):*

I added a second mapping with offset != 0. This covers the following cases: two mappings, shared/private, and different offsets.
</blockquote></details>

Thanks, looks awesome!

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/550)* status: 8 of 17 files reviewed, 33 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

___
*[`LibOS/shim/include/shim_fs.h` line 89 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0fdNEU7nvJdXHzWOdA:-N0gXdIS1lehfX6CR4_y:b-dywzp5) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/include/shim_fs.h#L89)):*

Why do we need this artificial disallowance of `NULL`?
Update: is it because it's passed as is to PAL API which still requires it?
</blockquote></details>

Yes. PAL interprets NULL as "use any address".
___
*[`LibOS/shim/include/shim_fs.h` line 91 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0fBPlS7WZ8UK1Hk95z:-N0gXl7d6hB9QLR47O1G:bidrs83) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/include/shim_fs.h#L91)):*

Why these are `PAL_PROT_*`??
Update: from what I can see these are Linux prot, not PAL...
</blockquote></details>

Right. Fixed.
___
*[`LibOS/shim/include/shim_fs.h` line 112 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0evIM64zw9GYqLQqq0:-N0gXtIs8BOyJlnBe0rm:btx8cls) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/include/shim_fs.h#L112)):*

Please comment all args. 
</blockquote></details>

OK, they're the same as `mmap` though. I modified the description to make that clear.
___
*[`LibOS/shim/include/shim_fs.h` line 112 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0fCudh948XAoxrD3a4:-N0gYLcg6SeDKWV-_ebg:b-fntqhz) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/include/shim_fs.h#L112)):*

Why do we need `flags` argument?
</blockquote></details>

See the implementation. We sometimes need to call PAL to change permissions, and that requires both `prot` and `flags`:

```c
pal_prot_flags_t pal_prot = LINUX_PROT_TO_PAL(prot, flags);
```
___
*[`LibOS/shim/include/shim_vma.h` line 131 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0fFZbb0J1LGuaB51RO:-N0gYd28DWpZPKd_UHjj:bnjk5qx) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/include/shim_vma.h#L131)):*

Why do we need this? In which cases we need to call msync based on handle not address range?
</blockquote></details>

When the user requests to sync a specific handle, using the `sync` syscall (our `flush` callback).
___
*[`LibOS/shim/src/bookkeep/shim_vma.c` line 1119 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0fNhHO-pAwfonXdw8X:-N0gYxtV6sGXhtE1NVnB:b-3ga6w5) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/src/bookkeep/shim_vma.c#L1119)):*

How is this case possible?
</blockquote></details>

My mistake, somehow I thought `_lookup_vma` returns the VMA *before* address, not after. Removed.
___
*[`LibOS/shim/src/bookkeep/shim_vma.c` line 1173 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0fOJyc9B66_D4xjAzX:-N0gZfsl4SEvDAbGZ2kw:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/src/bookkeep/shim_vma.c#L1173)):*

Please use `UINTPTR_MAX`.
</blockquote></details>

Done.
___
*[`LibOS/shim/src/bookkeep/shim_vma.c` line 1174 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0fOfNUBcdXgBcvmYLW:-N0g_4Wu5f0ScIEvKHmz:bkmmaxw) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/src/bookkeep/shim_vma.c#L1174)):*

No need for `&` with function pointers.
</blockquote></details>

I thought we use `&` for clarity; I pretty universally used it in my code so far and nobody complained (most notably, all the code where we setup pseudofs). On the other hand, I see @mkow not using `&` in his sysfs code with callbacks.

Removed `&` for now.
___
*[`LibOS/shim/src/bookkeep/shim_vma.c` line 1246 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0fPYRfAnltXnVjnGtU:-N0gaOOh1OJiXdVWd-y-:b-rvgp81) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/src/bookkeep/shim_vma.c#L1246)):*

Is this case possible? We check for `MAP_ANONYMOUS` above.
</blockquote></details>

Yeah, it makes sense that `vma->file` would always be present. I wasn't sure because this module never enforces or documents this, so we depend on callers.

Replaced with assertion.
___
*[`LibOS/shim/src/bookkeep/shim_vma.c` line 1252 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0fPhvmA28yj9oVmC8x:-N0gcSPnES1tewPg3xVs:b-2la1kx) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/src/bookkeep/shim_vma.c#L1252)):*

Don't we need a lock for this?
Conceptually we shouldn't, but on the current master I think it's required, but I'm not sure.
</blockquote></details>

Yeah, we still modify it in socket code. I'll add a lock and a TODO here.
___
*[`LibOS/shim/src/bookkeep/shim_vma.c` line 1261 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0f_2MJ9lAPaP4AHhDt:-N0gdwvkCCTVsYrKmjzQ:b-25y4ak) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/src/bookkeep/shim_vma.c#L1261)):*

Please assert that `begin` and `end` are page aligned.
</blockquote></details>

OK. This is a bit awkward because of `UINTPTR_MAX`, but I don't know how to do it better.
___
*[`LibOS/shim/src/bookkeep/shim_vma.c` line 1261 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0faZJH2mrg32AENycx:-N0ge47x5icsST9VcuqS:b-mwhan6) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/src/bookkeep/shim_vma.c#L1261)):*

We usually use names with leading underscore to highlight functions which must be called with *some* locks taken.
Not blocking - this is rather informal convention.
</blockquote></details>

I called this `_msync_all` back when I actually had a function called `msync_all` (it wasn't about a lock, but about defining a helper function). That's not necessary now, so I'm removing the underscore.
___
*[`LibOS/shim/src/bookkeep/shim_vma.c` line 1265 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0fQ6aa3zSHTkzUcDC5:-N0ge3M-7Y0i3VR4yU4F:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/src/bookkeep/shim_vma.c#L1265)):*

No need for `&` with function pointer
</blockquote></details>

Done.
___
*[`LibOS/shim/src/bookkeep/shim_vma.c` line 1296 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0famf854MStlMK7Pop:-N0geIvrDipluytgsmZj:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/src/bookkeep/shim_vma.c#L1296)):*

`UINTPTR_MAX`
</blockquote></details>

Done.
___
*[`LibOS/shim/src/fs/shim_fs_mem.c` line 91 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0fbuszAUIw2edgBGS4:-N0geTNn2sgCpdl6XpTU:b-pmnlha) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/src/fs/shim_fs_mem.c#L91)):*

Could we have this condition in the same order as the one below?
</blockquote></details>

Done (and same in the function above).
___
*[`LibOS/shim/src/fs/shim_fs_mem.c` line 98 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0fcd9TCBYcW-5hEtky:-N0geThD2Y0X72zOxI1q:b75kcgv) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/src/fs/shim_fs_mem.c#L98)):*

Shouldn't this be `>=`?
</blockquote></details>

Makes sense, but then I have to also add a case for `size == 0`. Done.
___
*[`LibOS/shim/src/fs/shim_fs_util.c` line 238 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0flIaF-4aTbx2fkd8E:-N0goxJD7_bUEf6Qbj_n:bt2vr1b) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/src/fs/shim_fs_util.c#L238)):*

This should be `protect_ret`
</blockquote></details>

Fixed.
___
*[`LibOS/shim/src/fs/shim_fs_util.c` line 240 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0flTsrCS9w2z1mrhmy:-N0gfr9B8sT_K7ChY5mi:bavf66g) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/src/fs/shim_fs_util.c#L240)):*

I think we should kill the app, since the state is inconsistent (and now the app won't even know about it).
</blockquote></details>

Added `BUG()`.
___
*[`LibOS/shim/src/fs/chroot/encrypted.c` line 31 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0dxxjE1y3c3sln-IS2:-N0ggBFWBJy-nkZf9z4a:brxp9sq) ([raw file](https://github.com/gramineproject/gramine/blob/9c41ea9c475fc3622cc8ea9bf0aaf7b49e688366/LibOS/shim/src/fs/chroot/encrypted.c#L31)):*

I think the only way is to get rid of all handle references and have flushing on release. To do that we would probably need to close all fds and unmmap all user memory.
</blockquote></details>

Sounds good.
___
*[`LibOS/shim/src/fs/chroot/encrypted.c` line 373 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0dyOon7_KbV8kRvKCt:-N0ggIO40mO2G1BBZiry:b1pr21u) ([raw file](https://github.com/gramineproject/gramine/blob/9c41ea9c475fc3622cc8ea9bf0aaf7b49e688366/LibOS/shim/src/fs/chroot/encrypted.c#L373)):*

I'm still a bit confused. Why we don't have to call `msync_handle` inside `encrypted_file_flush`?
</blockquote></details>

`encrypted_file_flush` only works with the encrypted-file object. It has no idea that we're *also* keeping emulated mappings (and in fact, does not depend on `shim_handle` or `shim_vma` or anything else in the system; it just takes care of that one object).

I added comments here to explain that better.
___
*[`LibOS/shim/src/fs/tmpfs/fs.c` line 295 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0dzAQYCMVlZuV1EX1O:-N0giW4S4370p_SmYBCj:b-vfowf5) ([raw file](https://github.com/gramineproject/gramine/blob/9774ce1ca8551b868c1ca6c423dfcf1b010115d4/LibOS/shim/src/fs/tmpfs/fs.c#L295)):*

+1 it seems that it's impossible to have correct mmap on tmpfs with SGX PAL.
</blockquote></details>

What do you mean? Emulated mappings should work exactly the same for tmpfs and encrypted fs:

- it is possible to have many mappings of a single file,
- the mappings are *not* updated automatically, we just fill them on mmap and write back to file on mmap.

This is not "correct", but a useful approximation, and what we always had for protected files.
___
*[`LibOS/shim/src/sys/shim_mmap.c` line 140 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0ftXvh6l9PPmFT0JO7:-N0gjq6b-Sj_a7tBZjAo:b6evnhr) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/src/sys/shim_mmap.c#L140)):*

Is this correct? In case of `MAP_FIXED_NOREPLACE` we might fail without doing anything. I guess random `msync` is still correct, because it only affects existing mappings (so we not need to worry if the memory is present), which are shared, so they can be flushed at any point?
</blockquote></details>

Yes. But you're right, we can skip this on `MAP_FIXED_NOREPLACE`.
___
*[`LibOS/shim/src/sys/shim_mmap.c` line 449 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0fwLb5B1C4LZ8exhLj:-N0gm5x32viLV5pHv56D:bh9y2fj) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/src/sys/shim_mmap.c#L449)):*

NP and not blocking: maybe move this to the if below, which kind of iterates through all flags?
</blockquote></details>

Moved below.
___
*[`LibOS/shim/test/regression/mmap_file_emulated.c` line 29 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0fxnXE6b5tXZfvupyC:-N0gncrZ0HESDJRHn7GB:b2nhktp) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/test/regression/mmap_file_emulated.c#L29)):*

Let's move these to lines with first usage (we use c11 so we can :) )
</blockquote></details>

OK.
___
*[`LibOS/shim/test/regression/mmap_file_emulated.c` line 62 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0fycD303u9vKtFsVrV:-N0gnjULEvRUMuC4V74K:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/test/regression/mmap_file_emulated.c#L62)):*

NP: `errx`
</blockquote></details>

Done.
___
*[`LibOS/shim/test/regression/mmap_file_emulated.c` line 74 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0fysur-u14xpW9XQUA:-N0gnjv93O-IAiad6YII:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/test/regression/mmap_file_emulated.c#L74)):*

ditto
</blockquote></details>

Done.
___
*[`LibOS/shim/test/regression/mmap_file_emulated.c` line 80 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0fywOi12hJeyCRLGAi:-N0gnnBQ1Clt8o_z8E6k:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/test/regression/mmap_file_emulated.c#L80)):*

NP: `puts`
</blockquote></details>

Done.
___
*[`LibOS/shim/test/regression/mmap_file_emulated.c` line 85 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0fz8IU-puXwvfjlazt:-N0gnqicCsb_wTthipNO:bt2vr1b) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/test/regression/mmap_file_emulated.c#L85)):*

NP: this check differs from all others (`< 0`)
</blockquote></details>

Fixed.
___
*[`LibOS/shim/test/regression/mmap_file_emulated.c` line 104 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0fzif-E8j5tf0smYh7:-N0gnw6f4XUSiQCQIwv1:bmcbklg) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/test/regression/mmap_file_emulated.c#L104)):*

Can we also check `addr_shared`?
</blockquote></details>

OK, sure.
___
*[`LibOS/shim/test/regression/mmap_file_emulated.c` line 107 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0fzhNiByosRqik3lkW:-N0go-C45uY42VLjjP8d:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/test/regression/mmap_file_emulated.c#L107)):*

NP: `puts`
</blockquote></details>

Done.
___
*[`LibOS/shim/test/regression/mmap_file_emulated.c` line 122 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0fzrw-7itg5ki5pkVO:-N0gom1xAHaZB-85f3Ys:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/test/regression/mmap_file_emulated.c#L122)):*

Could you move `close` to right after mmaps? To make sure these changes are flushed on `munmap`, not `close`.
</blockquote></details>

Done.
___
*[`LibOS/shim/test/regression/mmap_file_emulated.c` line 126 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0g-PEQ4z9eH15lvqty:-N0gop72BtYBq21LJihH:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/test/regression/mmap_file_emulated.c#L126)):*

ditto (`puts`)
</blockquote></details>

Done.
___
*[`LibOS/shim/test/regression/mmap_file_emulated.c` line 132 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0g-d-_Dt1rbPU3YjSn:-N0gouo1BsdYGLwleTX9:bidrs83) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/test/regression/mmap_file_emulated.c#L132)):*

Shouldn't we also check `ret < 0`?
</blockquote></details>

Right. Fixed.
___
*[`LibOS/shim/test/regression/mmap_file_emulated.c` line 142 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0g-gKy1A3dunP4a4Gg:-N0gow2M4RZb-x-DDSY0:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/test/regression/mmap_file_emulated.c#L142)):*

ditto (`puts`)
</blockquote></details>

Done.
___
*[`LibOS/shim/src/sys/shim_exit.c` line 18 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0ftJjHBSwY1adPMp8-:-N0gj_8uCiefgkN9D4Ui:bjrdvad) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/src/sys/shim_exit.c#L18)):*

Why we need this?
</blockquote></details>

Leftover from earlier attempt to flush-on-exit, before I decided I don't want to do it now. Fixed.

boryspoplawski (changes requested):
> Reviewed 9 of 9 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/550)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @pwmarcz)

___
*[`LibOS/shim/include/shim_fs.h` line 89 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0fdNEU7nvJdXHzWOdA:-N0h7P944YsmGwKQnwnS:b-81la7d) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/include/shim_fs.h#L89)):*

Yes. PAL interprets NULL as "use any address".
</blockquote></details>

Right
___
*[`LibOS/shim/include/shim_fs.h` line 112 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0fCudh948XAoxrD3a4:-N0h7lQz58ptc9FK7XAa:b-gr11qm) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/include/shim_fs.h#L112)):*

See the implementation. We sometimes need to call PAL to change permissions, and that requires both `prot` and `flags`:

```c
pal_prot_flags_t pal_prot = LINUX_PROT_TO_PAL(prot, flags);
```
</blockquote></details>

Not sure how much of an implementation details it is, but they only care for `MAP_PRIVATE`, which we know to be false in `msync`.
___
*[`LibOS/shim/src/bookkeep/shim_vma.c` line 1174 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0fOfNUBcdXgBcvmYLW:-N0h8KSvEOrcSr0t8hjC:b3od7gn) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/src/bookkeep/shim_vma.c#L1174)):*

I thought we use `&` for clarity; I pretty universally used it in my code so far and nobody complained (most notably, all the code where we setup pseudofs). On the other hand, I see @mkow not using `&` in his sysfs code with callbacks.

Removed `&` for now.
</blockquote></details>

Hmm, I guess we have no official policy and it's a matter of preference. Should have mentioned it, but not blocked, my bad.
I personally prefer no `&`, because these are function pointers, and that should be clear from the name itself and the `&` is not needed, like, we already know it's a function/pointer.
___
*[`LibOS/shim/src/bookkeep/shim_vma.c` line 1252 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0fPhvmA28yj9oVmC8x:-N0h8yqB8-Ni9W1gPxsZ:brtjszr) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/src/bookkeep/shim_vma.c#L1252)):*

Yeah, we still modify it in socket code. I'll add a lock and a TODO here.
</blockquote></details>

Hmmm, actually, are we allowed to take any locks here? We are inside vmas traversing function, under global vma lock, I think.
In reality `acc_mode` can't change for anything other than sockets, so maybe we can just ignore the lock, leave a `XXX` describing the situation, to be removed when we clear semantics of accessing `acc_mode` (which we should do after sockets rewrite)
wdyt? Above I assumed that only sockets are the issue.
___
*[`LibOS/shim/src/bookkeep/shim_vma.c` line 1261 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0f_2MJ9lAPaP4AHhDt:-N0h9k8N6GTxZYLJhNvI:b-fgak3n) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/src/bookkeep/shim_vma.c#L1261)):*

OK. This is a bit awkward because of `UINTPTR_MAX`, but I don't know how to do it better.
</blockquote></details>

Right. But I guess it's ok
___
*[`LibOS/shim/src/fs/chroot/encrypted.c` line 373 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0dyOon7_KbV8kRvKCt:-N0hAmaX6cKOnBtRHpap:bs7bmwu) ([raw file](https://github.com/gramineproject/gramine/blob/9c41ea9c475fc3622cc8ea9bf0aaf7b49e688366/LibOS/shim/src/fs/chroot/encrypted.c#L373)):*

`encrypted_file_flush` only works with the encrypted-file object. It has no idea that we're *also* keeping emulated mappings (and in fact, does not depend on `shim_handle` or `shim_vma` or anything else in the system; it just takes care of that one object).

I added comments here to explain that better.
</blockquote></details>

Well, I'm not familiar with the new PF implementation, so I would have to check it out. Resolving.
___
*[`LibOS/shim/src/fs/tmpfs/fs.c` line 295 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0dzAQYCMVlZuV1EX1O:-N0hBHkx4VtiXIFAyEmi:b-ddchhy) ([raw file](https://github.com/gramineproject/gramine/blob/9774ce1ca8551b868c1ca6c423dfcf1b010115d4/LibOS/shim/src/fs/tmpfs/fs.c#L295)):*

What do you mean? Emulated mappings should work exactly the same for tmpfs and encrypted fs:

- it is possible to have many mappings of a single file,
- the mappings are *not* updated automatically, we just fill them on mmap and write back to file on mmap.

This is not "correct", but a useful approximation, and what we always had for protected files.
</blockquote></details>

I guess nothing really relies on mapping the same file twice? Maybe the current version won't hurt anybody, idk

dimakuv (dismissed):
> Reviewed 3 of 9 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/550)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @pwmarcz)

___
*[`LibOS/shim/src/bookkeep/shim_vma.c` line 1252 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0fPhvmA28yj9oVmC8x:-N0ioIwN7ZtILnMEFnQH:bl9kz12) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/src/bookkeep/shim_vma.c#L1252)):*

Hmmm, actually, are we allowed to take any locks here? We are inside vmas traversing function, under global vma lock, I think.
In reality `acc_mode` can't change for anything other than sockets, so maybe we can just ignore the lock, leave a `XXX` describing the situation, to be removed when we clear semantics of accessing `acc_mode` (which we should do after sockets rewrite)
wdyt? Above I assumed that only sockets are the issue.
</blockquote></details>

+1. I don't like the idea of having a lock in this function. I think we can safely assume that after the previous check on `msync` callback existence, we can be sure that it's a file handle and then its `acc_mode` is never changed. We can add a comment about this.
___
*[`LibOS/shim/src/fs/tmpfs/fs.c` line 295 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0dzAQYCMVlZuV1EX1O:-N0iq40q7OUcApl_huHs:bd16fb7) ([raw file](https://github.com/gramineproject/gramine/blob/9774ce1ca8551b868c1ca6c423dfcf1b010115d4/LibOS/shim/src/fs/tmpfs/fs.c#L295)):*

I guess nothing really relies on mapping the same file twice? Maybe the current version won't hurt anybody, idk
</blockquote></details>

Yeah, nobody ever complained about this. So I think this is not a realistic scenario. Indeed, why would an app want to map the same file twice. I'm resolving.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/550)* status: 15 of 17 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[`LibOS/shim/include/shim_fs.h` line 112 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0fCudh948XAoxrD3a4:-N0jCk200171GptRssgd:b-6x95oc) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/include/shim_fs.h#L112)):*

Not sure how much of an implementation details it is, but they only care for `MAP_PRIVATE`, which we know to be false in `msync`.
</blockquote></details>

Oooh, actually, the PAL mapping should not be shared. The Linux-SGX PAL ignores that distinction anyway, but we don't want Linux PAL to create a shared region between processes. So right, the `flags` parameter will be unused here.

I still want to leave the parameter, because:
- it makes for a nice symmetry with the `mmap` callback: they have the same parameters, omitting just one of them would be surprising,
- we get to have an `assert` that we're not flushing a Gramine-level `MAP_PRIVATE` mapping.
___
*[`LibOS/shim/src/bookkeep/shim_vma.c` line 1252 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0fPhvmA28yj9oVmC8x:-N0jG_hA2jktoaWgyIg7:b-v5ffb) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/src/bookkeep/shim_vma.c#L1252)):*

+1. I don't like the idea of having a lock in this function. I think we can safely assume that after the previous check on `msync` callback existence, we can be sure that it's a file handle and then its `acc_mode` is never changed. We can add a comment about this.
</blockquote></details>

Oh, right, this is under `vma_tree_lock`, I shouldn't do that.

Changed as suggested.
___
*[`LibOS/shim/src/fs/tmpfs/fs.c` line 295 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0dzAQYCMVlZuV1EX1O:-N0jCQzS2fLp0Nw3anXG:bmdlpfw) ([raw file](https://github.com/gramineproject/gramine/blob/9774ce1ca8551b868c1ca6c423dfcf1b010115d4/LibOS/shim/src/fs/tmpfs/fs.c#L295)):*
>Indeed, why would an app want to map the same file twice.

Well, there's the [magic ring buffer trick](https://abhinavag.medium.com/a-fast-circular-ring-buffer-4d102ef4d4a3) :)

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/550)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`LibOS/shim/src/fs/tmpfs/fs.c` line 295 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0dzAQYCMVlZuV1EX1O:-N0jKb9oA2SPJZFN8uM9:bybvs81) ([raw file](https://github.com/gramineproject/gramine/blob/9774ce1ca8551b868c1ca6c423dfcf1b010115d4/LibOS/shim/src/fs/tmpfs/fs.c#L295)):*

>Indeed, why would an app want to map the same file twice.

Well, there's the [magic ring buffer trick](https://abhinavag.medium.com/a-fast-circular-ring-buffer-4d102ef4d4a3) :)
</blockquote></details>

One `memcpy` instead of two `memcpy`s to allow the compiler to apply vectorization optimizations? Wow, I wonder if people are actually using such tricks to squeze out an additional... 0.001% performance? But maybe I'm wrong, maybe it's useful somewhere.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/550)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[`LibOS/shim/include/shim_fs.h` line 112 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0fCudh948XAoxrD3a4:-N0kmoeFEp_V0QxJlGHz:b-z2bb8f) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/include/shim_fs.h#L112)):*
Ok, let's leave it.

> we don't want Linux PAL to create a shared region between processes

Why is that? Ideally most of the PALs should support shared mappings, so should msync. I guess some of filesystems (e.g. tmpfs) are fully emulated and as such won't create a shared mapping, but that's not part of this interface.
___
*[`LibOS/shim/src/fs/shim_fs_util.c` line 147 at r5](https://reviewable.io/reviews/gramineproject/gramine/550#-N0knqqE46xzA9wMzVLu:-N0knqqE46xzA9wMzVLv:b-u3nel9) ([raw file](https://github.com/gramineproject/gramine/blob/f1cbf1987379aa5ba839385e1baee0fd71df9886/LibOS/shim/src/fs/shim_fs_util.c#L147)):*
> ```C
>     int ret;
> 
>     /* The underlying PAL mapping should always be private. */
> ```

I think this deserves more explanation, i.e. why it should be private
___
*[`LibOS/shim/src/fs/shim_fs_util.c` line 206 at r5](https://reviewable.io/reviews/gramineproject/gramine/550#-N0knyf4BUzJLULBX2Hd:-N0knyf4BUzJLULBX2He:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/f1cbf1987379aa5ba839385e1baee0fd71df9886/LibOS/shim/src/fs/shim_fs_util.c#L206)):*
> ```C
>     unlock(&hdl->inode->lock);
> 
>     /* The underlying PAL mapping should always be private. */
> ```

ditto

mkow (commented):
> Reviewed 1 of 17 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/550)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[`LibOS/shim/src/fs/chroot/encrypted.c` line 31 at r2](https://reviewable.io/reviews/gramineproject/gramine/550#-N0dxxjE1y3c3sln-IS2:-N0lEcpQ2VREWBAedLng:b32v57) ([raw file](https://github.com/gramineproject/gramine/blob/9c41ea9c475fc3622cc8ea9bf0aaf7b49e688366/LibOS/shim/src/fs/chroot/encrypted.c#L31)):*

Sounds good.
</blockquote></details>

+1

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/550)* status: 16 of 17 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @pwmarcz)

___
*[`LibOS/shim/include/shim_fs.h` line 112 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0fCudh948XAoxrD3a4:-N0oCoAq3gZHyZ956CSv:b-yzw8m) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/include/shim_fs.h#L112)):*

Ok, let's leave it.

> we don't want Linux PAL to create a shared region between processes

Why is that? Ideally most of the PALs should support shared mappings, so should msync. I guess some of filesystems (e.g. tmpfs) are fully emulated and as such won't create a shared mapping, but that's not part of this interface.
</blockquote></details>

I agree that this interface (mmap, msync) should include `flags`, and allow shared mappings.

I meant that the current *implementation* of emulated mappings (exactly the case you mention) should probably not share the underlying mappings between processes. It could end up flushing changes made by another process, and I'm not sure if it's a good idea.

However, I just realized that **it actually doesn't matter**: even if we create a shared anonymous mapping, it won't really be shared with another process. LibOS in the child process will just create an identical mapping during checkpoint restore.

So I think both solutions (using `LINUX_PROT_TO_PAL(prot, flags)` and using `LINUX_PROT_TO_PAL(prot, MAP_PRIVATE)`) make some sense. I prefer to use the first one, for simplicity: we do the same elsewhere, and I don't want a special case here that actually has zero effect.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/550)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[`LibOS/shim/include/shim_fs.h` line 112 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0fCudh948XAoxrD3a4:-N0ovqqs2BmJ2s2TBejU:b-sh6pky) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/include/shim_fs.h#L112)):*

I agree that this interface (mmap, msync) should include `flags`, and allow shared mappings.

I meant that the current *implementation* of emulated mappings (exactly the case you mention) should probably not share the underlying mappings between processes. It could end up flushing changes made by another process, and I'm not sure if it's a good idea.

However, I just realized that **it actually doesn't matter**: even if we create a shared anonymous mapping, it won't really be shared with another process. LibOS in the child process will just create an identical mapping during checkpoint restore.

So I think both solutions (using `LINUX_PROT_TO_PAL(prot, flags)` and using `LINUX_PROT_TO_PAL(prot, MAP_PRIVATE)`) make some sense. I prefer to use the first one, for simplicity: we do the same elsewhere, and I don't want a special case here that actually has zero effect.
</blockquote></details>

But these are 2 different versions, because (as per comment) `flags` **must** contain `MAP_SHARED` (not `MAP_PRIVATE`).

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/550)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[`LibOS/shim/include/shim_fs.h` line 112 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0fCudh948XAoxrD3a4:-N0ow8ot0SslHC7Lhl8F:bqeczwp) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/include/shim_fs.h#L112)):*

But these are 2 different versions, because (as per comment) `flags` **must** contain `MAP_SHARED` (not `MAP_PRIVATE`).
</blockquote></details>

I mean, one day, when we have a fully working shared mappings on *some* PALs, how do one find that these places needs to be fixed? Or maybe the  wouldn't need to be?

boryspoplawski (approved):
> Reviewed 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/550)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

___
*[`LibOS/shim/include/shim_fs.h` line 112 at r3](https://reviewable.io/reviews/gramineproject/gramine/550#-N0fCudh948XAoxrD3a4:-N0oworc-bBFQR-uXTS_:bb3pwio) ([raw file](https://github.com/gramineproject/gramine/blob/496c28fccfda5df189b14ac514fab9683e8f8e89/LibOS/shim/include/shim_fs.h#L112)):*

I mean, one day, when we have a fully working shared mappings on *some* PALs, how do one find that these places needs to be fixed? Or maybe the  wouldn't need to be?
</blockquote></details>

Ok, I guess these are implementations of `mmap` and `msync`, so they will need to be revisited then.

dimakuv (approved):
> Reviewed 4 of 9 files at r4, 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/550)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #525, opened 2022-04-13
[LibOS] Warn when app tries using disabled eventfd

NieDzejkob:
## How to test this PR?

See #495 for an example of an application making use of eventfd.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/525)

PR #572, opened 2022-05-09
[common] Create `FIRST_TIME()` macro

NieDzejkob:
## Description of the changes

In #525, @boryspoplawski suggested that a macro like this would be too complicated to be worthwhile. Having actually drafted an implementation, I don't think it looks too bad. In fact, I believe this turns the preferred approach of doing something once per enclave into something you have a chance of remembering.

## How to test this PR?

You may run a program that triggers once of the warnings that make use of `DO_ONCE`, and verify that the warning, indeed, happens once. For example:

```c
#include <stdint.h>
#include <sys/mman.h>

int main(void) {
    unsigned char output[16];
    uintptr_t addr = (uintptr_t) main;
    addr &= ~0xfff;
    mincore((void*)addr, 4096, output);
    mincore((void*)addr, 4096, output);
    mincore((void*)addr, 4096, output);
    return 0;
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/572)

### DIFF
--- a/LibOS/shim/src/sys/shim_exit.c
+++ b/LibOS/shim/src/sys/shim_exit.c
@@ -153,8 +153,7 @@ noreturn void process_exit(int error_code, int term_signal) {
 
     /* If process_exit is invoked multiple times, only a single invocation proceeds past this
      * point. */
-    static int first = 0;
-    if (__atomic_exchange_n(&first, 1, __ATOMIC_RELAXED) != 0) {
+    if (!FIRST_TIME()) {
         /* Just exit current thread. */
         thread_exit(error_code, term_signal);
     }
--- a/LibOS/shim/src/sys/shim_mmap.c
+++ b/LibOS/shim/src/sys/shim_mmap.c
@@ -325,8 +325,7 @@ long shim_do_mincore(void* addr, size_t len, unsigned char* vec) {
     if (!is_user_memory_writable(vec, pages))
         return -EFAULT;
 
-    static unsigned int warned = 0;
-    if (__atomic_exchange_n(&warned, 1, __ATOMIC_RELAXED) == 0) {
+    if (FIRST_TIME()) {
         log_warning("mincore emulation always tells pages are _NOT_ in RAM. This may cause "
                     "issues.");
     }
--- a/LibOS/shim/src/sys/shim_sleep.c
+++ b/LibOS/shim/src/sys/shim_sleep.c
@@ -91,8 +91,7 @@ long shim_do_clock_nanosleep(clockid_t clock_id, int flags, struct __kernel_time
     }
 
     if (clock_id == CLOCK_PROCESS_CPUTIME_ID) {
-        static unsigned int warned = 0;
-        if (__atomic_exchange_n(&warned, 1, __ATOMIC_RELAXED) == 0) {
+        if (FIRST_TIME()) {
             log_warning("Per-process CPU-time clock is not supported in clock_nanosleep(); "
                         "it is replaced with system-wide real-time clock.");
         }
--- a/LibOS/shim/src/sys/shim_time.c
+++ b/LibOS/shim/src/sys/shim_time.c
@@ -62,8 +62,7 @@ long shim_do_clock_gettime(clockid_t which_clock, struct timespec* tp) {
         return -EFAULT;
 
     if (which_clock == CLOCK_PROCESS_CPUTIME_ID || which_clock == CLOCK_THREAD_CPUTIME_ID) {
-        static unsigned int warned = 0;
-        if (__atomic_exchange_n(&warned, 1, __ATOMIC_RELAXED) == 0) {
+        if (FIRST_TIME()) {
             log_warning("Per-process and per-thread CPU-time clocks are not supported in "
                         "clock_gettime(); they are replaced with system-wide real-time clock.");
         }
@@ -86,8 +85,7 @@ long shim_do_clock_getres(clockid_t which_clock, struct timespec* tp) {
         return -EINVAL;
 
     if (which_clock == CLOCK_PROCESS_CPUTIME_ID || which_clock == CLOCK_THREAD_CPUTIME_ID) {
-        static unsigned int warned = 0;
-        if (__atomic_exchange_n(&warned, 1, __ATOMIC_RELAXED) == 0) {
+        if (FIRST_TIME()) {
             log_warning("Per-process and per-thread CPU-time clocks are not supported in "
                         "clock_getres(); they are replaced with system-wide real-time clock.");
         }
--- a/Pal/src/host/Linux-SGX/db_exception.c
+++ b/Pal/src/host/Linux-SGX/db_exception.c
@@ -111,8 +111,7 @@ static void save_pal_context(PAL_CONTEXT* ctx, sgx_cpu_context_t* uc,
 }
 
 static void emulate_rdtsc_and_print_warning(sgx_cpu_context_t* uc) {
-    static int first = 0;
-    if (__atomic_exchange_n(&first, 1, __ATOMIC_RELAXED) == 0) {
+    if (FIRST_TIME()) {
         /* if we end up emulating RDTSC/RDTSCP instruction, we cannot use invariant TSC */
         extern uint64_t g_tsc_hz;
         g_tsc_hz = 0;
@@ -166,8 +165,7 @@ static bool handle_ud(sgx_cpu_context_t* uc) {
         return false;
     } else if (instr[0] == 0x0f && instr[1] == 0x05) {
         /* syscall: LibOS may know how to handle this */
-        static int log_once = 1;
-        if (__atomic_exchange_n(&log_once, 0, __ATOMIC_RELAXED)) {
+        if (FIRST_TIME()) {
             log_always("Emulating a raw syscall instruction. This degrades performance, consider"
                        " patching your application to use Gramine syscall API.");
         }
--- a/Pal/src/host/Linux/db_exception.c
+++ b/Pal/src/host/Linux/db_exception.c
@@ -87,8 +87,7 @@ static void perform_signal_handling(enum pal_event event, bool is_in_pal, PAL_NU
 static void handle_sync_signal(int signum, siginfo_t* info, struct ucontext* uc) {
     if (info->si_signo == SIGSYS && info->si_code == SYS_SECCOMP) {
         ucontext_revert_syscall(uc, info->si_arch, info->si_syscall, info->si_call_addr);
-        static int log_once = 1;
-        if (__atomic_exchange_n(&log_once, 0, __ATOMIC_RELAXED)) {
+        if (FIRST_TIME()) {
             log_always("Emulating a raw system/supervisor call. This degrades performance, consider"
                        " patching your application to use Gramine syscall API.");
         }
--- a/common/include/api.h
+++ b/common/include/api.h
@@ -192,6 +192,13 @@ typedef ptrdiff_t ssize_t;
 
 #define IS_IN_RANGE_INCL(value, start, end) (((value) < (start) || (value) > (end)) ? false : true)
 
+/* Each occurence of this macro in the source code will return `true` only once per process.
+ *
+ * We use __ATOMIC_RELAXED here, as a consistent ordering within the accesses to `first` is enough
+ * for us — FIRST_TIME is not a synchronization primitive, as the macro returning `false` doesn't
+ * actually guarantee that the code path entered when it returned `true` has finished executing. */
+#define FIRST_TIME() ({ static uint8_t first = 0; __atomic_exchange_n(&first, 1, __ATOMIC_RELAXED) == 0; })
+
 /* LibC functions */
 
 /* LibC string functions */


boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/572)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @NieDzejkob)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/572#-N1cxJ4M3ZU6APQuaZJc:-N1cxJ4M3ZU6APQuaZJd:binux35) (no related file):*
> In #525, @boryspoplawski suggested that a macro like this would be too complicated to be worthwhile.

That was about passing a block of code into the macro

---
___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/572#-N1cxsgP3iFSQ3xYKO1S):*
This deserves at least a one sentence explaining what is that and what it does.
Also please append `macro` to the oneliner
___
*[`common/include/api.h` line 195 at r1](https://reviewable.io/reviews/gramineproject/gramine/572#-N1cy6LK9-SBlrkgePzF:-N1cy6LL3j-yg9SvUVtx:bdqoj0o) ([raw file](https://github.com/gramineproject/gramine/blob/79bc4cc056069a07fbaf926231e1ce07ddb1c0f5/common/include/api.h#L195)):*
> ```objc
> #define IS_IN_RANGE_INCL(value, start, end) (((value) < (start) || (value) > (end)) ? false : true)
> 
> /* Each occurence of this macro in the source code will returns `true` only once per enclave.
> ```

What enclave? This is common code.
___
*[`common/include/api.h` line 200 at r1](https://reviewable.io/reviews/gramineproject/gramine/572#-N1cyHWS8yeY3vcrK-Xr:-N1cyHWS8yeY3vcrK-Xs:b-d9gfa9) ([raw file](https://github.com/gramineproject/gramine/blob/79bc4cc056069a07fbaf926231e1ce07ddb1c0f5/common/include/api.h#L200)):*
> ```objc
>  * as memory ordering is concerned. In particular, since there isn't any synchronization to
>  * signal when an `if (DO_ONCE)` block is done executing, returning `false` doesn't constitute
>  * any guarantees about the variables modified within the `if (DO_ONCE)` block. */
> ```

What's the point of this comment? We do not explain on every function/macro that it does guarantee any ordering for the rest of the code.
___
*[`common/include/api.h` line 201 at r1](https://reviewable.io/reviews/gramineproject/gramine/572#-N1cxj9KEHZaFhGGkeCx:-N1cxj9KEHZaFhGGkeCy:bl2chmn) ([raw file](https://github.com/gramineproject/gramine/blob/79bc4cc056069a07fbaf926231e1ce07ddb1c0f5/common/include/api.h#L201)):*
> ```objc
>  * signal when an `if (DO_ONCE)` block is done executing, returning `false` doesn't constitute
>  * any guarantees about the variables modified within the `if (DO_ONCE)` block. */
> #define DO_ONCE ({ static uint8_t first = 0; __atomic_exchange_n(&first, 1, __ATOMIC_RELAXED) == 0; })
> ```

`DO_ONCE()`. This looks like a constant, not a piece of code.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/572)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @NieDzejkob)

___
*[`common/include/api.h` line 200 at r1](https://reviewable.io/reviews/gramineproject/gramine/572#-N1cyHWS8yeY3vcrK-Xr:-N1czQSR6WUQrNamtIQT:bvgahfj) ([raw file](https://github.com/gramineproject/gramine/blob/79bc4cc056069a07fbaf926231e1ce07ddb1c0f5/common/include/api.h#L200)):*

What's the point of this comment? We do not explain on every function/macro that it does guarantee any ordering for the rest of the code.
</blockquote></details>

\***does not**

boryspoplawski (changes requested):
> Reviewed 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/572)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @NieDzejkob)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/572#-N1cxsgP3iFSQ3xYKO1S):*

Sure thing, should I force-push and change that or is there a better approach?
</blockquote></details>

Add a squash commit with something like: `TODO: new commit message:` and new commit message below. Just remember that the part after `squash!` must match the oneliner here

I propose the new oneliner to be: `[common] Create DO_ONCE() macro`
___
*[`common/include/api.h` line 195 at r1](https://reviewable.io/reviews/gramineproject/gramine/572#-N1cy6LK9-SBlrkgePzF:-N1dC9gLEypz0PzVc-3h:bbtww1l) ([raw file](https://github.com/gramineproject/gramine/blob/79bc4cc056069a07fbaf926231e1ce07ddb1c0f5/common/include/api.h#L195)):*

Fair enough. I wanted to stress that nothing like a sync between all the processes. Is `s/enclave/process/` general enough?
</blockquote></details>

Yeah, it's ok
___
*[`common/include/api.h` line 200 at r1](https://reviewable.io/reviews/gramineproject/gramine/572#-N1cyHWS8yeY3vcrK-Xr:-N1dCJ7l3lXw9PHPI6rQ:bpd45ve) ([raw file](https://github.com/gramineproject/gramine/blob/79bc4cc056069a07fbaf926231e1ce07ddb1c0f5/common/include/api.h#L200)):*

Yes, we do not explain it everywhere, but this macro has the rough shape of a synchronization primitive, so I believe it's not exactly unwarranted.

The point of the comment is to (attempt to) convey my understanding as to why `__ATOMIC_RELAXED` is enough here.
</blockquote></details>

But memory ordering of this access has nothing to with synchronization of the whole block. If you used `__ATOMIC_SEQ_CST` or did a full memory barrier, the result would be exactly the same - no synchronization  on stuff used in the code block.

NieDzejkob (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/572)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @NieDzejkob)

___
*[`LibOS/shim/src/sys/shim_exit.c` line 156 at r2](https://reviewable.io/reviews/gramineproject/gramine/572#-N1dS06VAdUjtRikayco:-N1dS06VAdUjtRikaycp:b-mpgmvj) ([raw file](https://github.com/gramineproject/gramine/blob/1e84aaaee8fac12f75af7fc4971aaaeacef713db/LibOS/shim/src/sys/shim_exit.c#L156)):*
> ```C
>     /* If process_exit is invoked multiple times, only a single invocation proceeds past this
>      * point. */
>     if (!DO_ONCE()) {
> ```

Considering this usage, perhaps a name like `FIRST_TIME()` would be better?

NieDzejkob (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/572)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @NieDzejkob)

___
*[`common/include/api.h` line 200 at r1](https://reviewable.io/reviews/gramineproject/gramine/572#-N1cyHWS8yeY3vcrK-Xr:-N1dS8zk0-R3x2XiKKjl:b-m4ilev) ([raw file](https://github.com/gramineproject/gramine/blob/79bc4cc056069a07fbaf926231e1ce07ddb1c0f5/common/include/api.h#L200)):*

But memory ordering of this access has nothing to with synchronization of the whole block. If you used `__ATOMIC_SEQ_CST` or did a full memory barrier, the result would be exactly the same - no synchronization  on stuff used in the code block.
</blockquote></details>

Yes. And because there is no synchronization on stuff used within the block, we don't need any guarantees on how accesses to `first` will be ordered relative to anything else.

Do you think you can write a better explanation?

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/572)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @NieDzejkob)

___
*[`common/include/api.h` line 200 at r1](https://reviewable.io/reviews/gramineproject/gramine/572#-N1cyHWS8yeY3vcrK-Xr:-N1doK-50OrtPGWZCfav:br9v6yv) ([raw file](https://github.com/gramineproject/gramine/blob/79bc4cc056069a07fbaf926231e1ce07ddb1c0f5/common/include/api.h#L200)):*
> And because there is no synchronization on stuff used within the block.

This is false, we cannot assume that. There might be.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/572)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, @NieDzejkob, and @pwmarcz)

___
*[`LibOS/shim/src/sys/shim_exit.c` line 156 at r2](https://reviewable.io/reviews/gramineproject/gramine/572#-N1dS06VAdUjtRikayco:-N1e_qEJ3VdZyw12kGcj:b-2szbj7) ([raw file](https://github.com/gramineproject/gramine/blob/1e84aaaee8fac12f75af7fc4971aaaeacef713db/LibOS/shim/src/sys/shim_exit.c#L156)):*

Considering this usage, perhaps a name like `FIRST_TIME()` would be better?
</blockquote></details>

I'm fine with both names.
@boryspoplawski, @dimakuv, @pwmarcz: ^

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/572)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @NieDzejkob, and @pwmarcz)

___
*[`LibOS/shim/src/sys/shim_exit.c` line 156 at r2](https://reviewable.io/reviews/gramineproject/gramine/572#-N1dS06VAdUjtRikayco:-N1ekegQ9COOuaWlxDt_:be6orxk) ([raw file](https://github.com/gramineproject/gramine/blob/1e84aaaee8fac12f75af7fc4971aaaeacef713db/LibOS/shim/src/sys/shim_exit.c#L156)):*

I'm fine with both names.
@boryspoplawski, @dimakuv, @pwmarcz: ^
</blockquote></details>

I prefer `DO_ONCE`, but very, very slightly.

dimakuv (changes requested):
> Reviewed 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/572)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @NieDzejkob, and @pwmarcz)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/572#-N1cxsgP3iFSQ3xYKO1S):*

Add a squash commit with something like: `TODO: new commit message:` and new commit message below. Just remember that the part after `squash!` must match the oneliner here

I propose the new oneliner to be: `[common] Create DO_ONCE() macro`
</blockquote></details>

Another simpler approach to this is to block yourself here in Reviewable, with a comment like "Change the commit message to `Bla bla bla`". I prefer this way to `squash!` commits.
___
*[`common/include/api.h` line 200 at r1](https://reviewable.io/reviews/gramineproject/gramine/572#-N1cyHWS8yeY3vcrK-Xr:-N1gnjZO7lytYd14zBXl:bifmd3f) ([raw file](https://github.com/gramineproject/gramine/blob/79bc4cc056069a07fbaf926231e1ce07ddb1c0f5/common/include/api.h#L200)):*

> And because there is no synchronization on stuff used within the block.

This is false, we cannot assume that. There might be.
</blockquote></details>

I too think that this explanation is too verbose. I think we can keep the first sentence (`We use ...`). But we can drop the second sentence (`In particular ...`).

Basically, we just want to give a hint to future readers that we indeed intended to use `__ATOMIC_RELAXED`. We don't want to explain each and every detail in our code.
___
*[`common/include/api.h` line 198 at r2](https://reviewable.io/reviews/gramineproject/gramine/572#-N1gnf225GrBHfXFDVvk:-N1gnf225GrBHfXFDVvl:brt3h2z) ([raw file](https://github.com/gramineproject/gramine/blob/1e84aaaee8fac12f75af7fc4971aaaeacef713db/common/include/api.h#L198)):*
> ```objc
>  *
>  * We use __ATOMIC_RELAXED here, as a consistent ordering within the accesses to `first` is enough
>  * for us — if we return `false`, that doesn't actually guarantee anything meaningful as far
> ```

`if we return` -> `if this macro returns`
___
*[`LibOS/shim/src/sys/shim_exit.c` line 156 at r2](https://reviewable.io/reviews/gramineproject/gramine/572#-N1dS06VAdUjtRikayco:-N1goPMk5DXLSu3SNdA0:b-1a9gdv) ([raw file](https://github.com/gramineproject/gramine/blob/1e84aaaee8fac12f75af7fc4971aaaeacef713db/LibOS/shim/src/sys/shim_exit.c#L156)):*

I prefer `DO_ONCE`, but very, very slightly.
</blockquote></details>

I strongly prefer `FIRST_TIME()`, because this particular case of `if (!DO_ONCE())` is not human-readable.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/572)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @NieDzejkob)

___
*[`LibOS/shim/src/sys/shim_exit.c` line 156 at r2](https://reviewable.io/reviews/gramineproject/gramine/572#-N1dS06VAdUjtRikayco:-N1h8Tw5EpCSPMb1nv8N:blm0rwz) ([raw file](https://github.com/gramineproject/gramine/blob/1e84aaaee8fac12f75af7fc4971aaaeacef713db/LibOS/shim/src/sys/shim_exit.c#L156)):*

I strongly prefer `FIRST_TIME()`, because this particular case of `if (!DO_ONCE())` is not human-readable.
</blockquote></details>

It wasn't obvious to when `DO_ONCE` returns true and when false. So yes, `FIRST_TIME` sounds better. Or some other name that suggests "this invocation returns true only the first time around".

NieDzejkob (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/572)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/572#-N1cxsgP3iFSQ3xYKO1S):*

Another simpler approach to this is to block yourself here in Reviewable, with a comment like "Change the commit message to `Bla bla bla`". I prefer this way to `squash!` commits.
</blockquote></details>

TODO: New commit message:

```
[common] Create a FIRST_TIME macro

Create a macro that encapsulates executing a code segment only when it is first reached.
```
___
*[`common/include/api.h` line 200 at r1](https://reviewable.io/reviews/gramineproject/gramine/572#-N1cyHWS8yeY3vcrK-Xr:-N1iNFxCBQ7KYuTXJNuj:b6vcn38) ([raw file](https://github.com/gramineproject/gramine/blob/79bc4cc056069a07fbaf926231e1ce07ddb1c0f5/common/include/api.h#L200)):*

I too think that this explanation is too verbose. I think we can keep the first sentence (`We use ...`). But we can drop the second sentence (`In particular ...`).

Basically, we just want to give a hint to future readers that we indeed intended to use `__ATOMIC_RELAXED`. We don't want to explain each and every detail in our code.
</blockquote></details>

Okay, I've made the explanation more concise.
___
*[`common/include/api.h` line 198 at r2](https://reviewable.io/reviews/gramineproject/gramine/572#-N1gnf225GrBHfXFDVvk:-N1iN7krA7mz7G7-AFJt:b-cjxref) ([raw file](https://github.com/gramineproject/gramine/blob/1e84aaaee8fac12f75af7fc4971aaaeacef713db/common/include/api.h#L198)):*

`if we return` -> `if this macro returns`
</blockquote></details>

Okay, applied the equivalent in the new wording.
___
*[`LibOS/shim/src/sys/shim_exit.c` line 156 at r2](https://reviewable.io/reviews/gramineproject/gramine/572#-N1dS06VAdUjtRikayco:-N1iMb3J9hZj1HG_Utli:bht3ek6) ([raw file](https://github.com/gramineproject/gramine/blob/1e84aaaee8fac12f75af7fc4971aaaeacef713db/LibOS/shim/src/sys/shim_exit.c#L156)):*

It wasn't obvious to when `DO_ONCE` returns true and when false. So yes, `FIRST_TIME` sounds better. Or some other name that suggests "this invocation returns true only the first time around".
</blockquote></details>

Changed to `FIRST_TIME`.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/572)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @NieDzejkob)

NieDzejkob (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/572)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/572#-N1cxsgP3iFSQ3xYKO1S):*

Suggested change: `This macro encapsulates executing a code segment only when it is reached for the first time.`
</blockquote></details>

Okay.
___
*[`common/include/api.h` line 200 at r1](https://reviewable.io/reviews/gramineproject/gramine/572#-N1cyHWS8yeY3vcrK-Xr:-N1jBJRA8_10uC_OLzXM:by8uhw8) ([raw file](https://github.com/gramineproject/gramine/blob/79bc4cc056069a07fbaf926231e1ce07ddb1c0f5/common/include/api.h#L200)):*
I am referring to

> All modifications to any particular atomic variable occur in a total order that is specific to this one atomic variable.

(from https://en.cppreference.com/w/cpp/atomic/memory_order)

`consistent` being "the same for all threads"

> I still dislike having this comment

Memory ordering is a genuinely tricky aspect of the code, and I don't think trying to reduce the comments on it is a good idea.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/572)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`common/include/api.h` line 200 at r1](https://reviewable.io/reviews/gramineproject/gramine/572#-N1cyHWS8yeY3vcrK-Xr:-N1jNG3s3-zhtiH_zA3r:bvcm0g5) ([raw file](https://github.com/gramineproject/gramine/blob/79bc4cc056069a07fbaf926231e1ce07ddb1c0f5/common/include/api.h#L200)):*
> Memory ordering is a genuinely tricky aspect of the code, and I don't think trying to reduce the comments on it is a good idea.

That's the problem here - you start talking about memory ordering in a place that should have nothing to do with it. As you pointed out, this gives 0 guarantees on memory ordering - why even mention it? Seeing a comment like this rings bells, like something fishy is going on here, need to be careful.

But maybe it's just me, I'm unblokcing, I'll leave this to others.
cc @dimakuv @mkow

NieDzejkob (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/572)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`common/include/api.h` line 200 at r1](https://reviewable.io/reviews/gramineproject/gramine/572#-N1cyHWS8yeY3vcrK-Xr:-N1jX_lxC7MwlZ6i2o2x:b-9jwwrb) ([raw file](https://github.com/gramineproject/gramine/blob/79bc4cc056069a07fbaf926231e1ce07ddb1c0f5/common/include/api.h#L200)):*

> Memory ordering is a genuinely tricky aspect of the code, and I don't think trying to reduce the comments on it is a good idea.

That's the problem here - you start talking about memory ordering in a place that should have nothing to do with it. As you pointed out, this gives 0 guarantees on memory ordering - why even mention it? Seeing a comment like this rings bells, like something fishy is going on here, need to be careful.

But maybe it's just me, I'm unblokcing, I'll leave this to others.
cc @dimakuv @mkow 
</blockquote></details>

What do you mean? Atomics are involved, you need to pick the memory ordering they should use, therefore memory ordering is involved.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/572)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`common/include/api.h` line 200 at r1](https://reviewable.io/reviews/gramineproject/gramine/572#-N1cyHWS8yeY3vcrK-Xr:-N1jbsPO71LQT-D5oWPL:b-fwk340) ([raw file](https://github.com/gramineproject/gramine/blob/79bc4cc056069a07fbaf926231e1ce07ddb1c0f5/common/include/api.h#L200)):*

What do you mean? Atomics are involved, you need to pick the memory ordering they should use, therefore memory ordering is involved.
</blockquote></details>

This is ordering w.r.t. other memory accesses, which we do not care about at all.

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/572)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`common/include/api.h` line 200 at r1](https://reviewable.io/reviews/gramineproject/gramine/572#-N1cyHWS8yeY3vcrK-Xr:-N1jp_FP4ExSUkcKpKBn:b3jtet2) ([raw file](https://github.com/gramineproject/gramine/blob/79bc4cc056069a07fbaf926231e1ce07ddb1c0f5/common/include/api.h#L200)):*

This is ordering w.r.t. other memory accesses, which we do not care about at all.
</blockquote></details>

I'm neutral on this. I agree that `__ATOMIC_RELAXED` triggers a red light when reading, but it's also true that here we don't have anything relying on it.

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/572)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`common/include/api.h` line 200 at r1](https://reviewable.io/reviews/gramineproject/gramine/572#-N1cyHWS8yeY3vcrK-Xr:-N1lmeHI4ydxGjrPk8VQ:b-mpn66a) ([raw file](https://github.com/gramineproject/gramine/blob/79bc4cc056069a07fbaf926231e1ce07ddb1c0f5/common/include/api.h#L200)):*

I'm neutral on this. I agree that `__ATOMIC_RELAXED` triggers a red light when reading, but it's also true that here we don't have anything relying on it.
</blockquote></details>

I'm fine with this comment. Basically, I would only highlight the fact that we use `__ATOMIC_RELAXED` to perform the "guarded" block of code only once, but there are no synchronization guarantees on that block of code. The current comment highlights this, though in a slightly more verbose way :) Anyway, this is a nice hint to a future developer, I don't see any problems with this text.

dimakuv:
Jenkins, test this please

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/572)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski and @NieDzejkob)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/572#-N1cxsgP3iFSQ3xYKO1S):*

Okay.
</blockquote></details>

Done.

dimakuv:
Jenkins, test this please

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/572)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required) (waiting on @boryspoplawski and @NieDzejkob)

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/572)* status: all files reviewed, 1 unresolved discussion (waiting on @NieDzejkob)

NieDzejkob (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/572)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/572#-N1cxsgP3iFSQ3xYKO1S):*

Done.
</blockquote></details>

Thanks.

PR #646, opened 2022-06-09
[Pal/Linux-SGX] Fix resource leak in secret_prov

liangintel:
The tcp connection was created every time invoking secret_provision_start(),
but it will not close. So, there will be lots of tcp connection keeping open.
Finally, the total fd that keeping open reached 900, and it cannot open any
more file or socket.

Signed-off-by: Liang Fang <liang.a.fang@intel.com>

## How to test this PR? 

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/646)

### DIFF
--- a/CI-Examples/ra-tls-secret-prov/src/secret_prov_client.c
+++ b/CI-Examples/ra-tls-secret-prov/src/secret_prov_client.c
@@ -79,7 +79,7 @@ int main(int argc, char** argv) {
     printf("--- Received secret1 = '%s', secret2 = '%s' ---\n", secret1, secret2);
     ret = 0;
 out:
-    secret_provision_destroy();
     secret_provision_close(&ctx);
+    secret_provision_destroy(&ctx);
     return ret;
 }
--- a/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov.h
+++ b/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov.h
@@ -22,6 +22,7 @@
 
 struct ra_tls_ctx {
     void* ssl;
+    void* attest_resources;
 };
 
 typedef int (*verify_measurements_cb_t)(const char* mrenclave, const char* mrsigner,
@@ -98,7 +99,7 @@ int secret_provision_get(uint8_t** out_secret, size_t* out_secret_size);
  * This function zeroes out the memory where provisioned secret is stored and frees it.
  */
 __attribute__ ((visibility("default")))
-void secret_provision_destroy(void);
+void secret_provision_destroy(struct ra_tls_ctx* ctx);
 
 /*!
  * \brief Establish an RA-TLS session and retrieve first secret (client-side).
--- a/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c
+++ b/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c
@@ -33,19 +33,21 @@
 #include "secret_prov.h"
 #include "util.h"
 
-/* these are globals because the user may continue using the SSL session even after invoking
- * secret_provision_start() (in the user-supplied callback) */
-static mbedtls_ctr_drbg_context g_ctr_drbg;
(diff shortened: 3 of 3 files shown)


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/646)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @liangintel)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/646#-N4BKv-Q7dnxK-VZBhDZ:-N4BKv-R-RKrZYVJXY-7:b93vzk4) (no related file):*
You're using the `master` branch in your fork of Gramine to submit this PR. This will lead to huge problems.

Please close this PR and open a new one, under a new branch in your fork. See our Contributing guide and especially this part: https://gramine.readthedocs.io/en/latest/devel/contributing.html#branch-names

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/646#-N4BNPu496yMZ-uA8ZYL:-N4BNPu496yMZ-uA8ZYM:b-33zmks) (no related file):*
OK, I understood the problem in SecretProv.

The problem is that we don't free the resources if the client called SecretProv API like this:
```c
secret_provision_start("localhost:4433", CA_CRT_PATH, /*this is important!*/&ctx);
```

In this case, the SecretProv library does *not* free the resources:
https://github.com/gramineproject/gramine/blob/master/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c#L276-L283

But the client then, when it is done with the TLS session, only calls these functions:
```c
    secret_provision_destroy();    /* this only cleans up the secret blob */
    secret_provision_close(&ctx);  /* this only closes the TLS session but doesn't free the resources */
```

And that's why we have the leak of `g_verifier_fd` (among other leaks).

@liangintel Would you mind if I close your PR and submit my own version of the PR? I have an idea how to fix it quickly.

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/646#-N4BSo73Chi9_5iWmzV_:-N4BSo73Chi9_5iWmzVa:b-z4mvf5) (no related file):*
I don't like your PR because it modifies the user-facing API (in particular, `secret_provision_destroy()`). This is a **huge** problem for users -- they will need to rewrite their code for no good reason.

I propose to submit my PR that solves your problem, but doesn't modify the user-facing API.

___
___
*[`Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c` line 45 at r1](https://reviewable.io/reviews/gramineproject/gramine/646#-N4BOejO6tmnbDql2mhK:-N4BOejP9s5dpCKoXSyW:bgqwj02) ([raw file](https://github.com/gramineproject/gramine/blob/6867e81598680cb8dae37e639b392ea8c03e2a83/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c#L45)):*
> ```C
>     mbedtls_pk_context my_ratls_key;
>     mbedtls_x509_crt my_ratls_cert;
> } secret_prov_resources_t;
> ```

Why is there a need to move from global variables to this "resources" object?
___
*[`Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c` line 294 at r1](https://reviewable.io/reviews/gramineproject/gramine/646#-N4BOUDY-qG85TogQmQn:-N4BOUDY-qG85TogQmQo:b-4n9po2) ([raw file](https://github.com/gramineproject/gramine/blob/6867e81598680cb8dae37e639b392ea8c03e2a83/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c#L294)):*
> ```C
>         free(provisioned_secret);
>         provisioned_secret = 0;
>     }
> ```

Ok, this code looks needed. Indeed, we seem to have a memory leak here.

(Also, I should have named the variables with the `g_` prefix here.)

dimakuv:
I created #647 which solves this problem. It will close this PR automatically.

PR #647, opened 2022-06-10
[Pal/Linux-SGX] tools: Fix resource leak in Secret Prov lib

dimakuv:
## Description of the changes 

The resources allocated for mbedTLS functioning on the client side of Secret Prov lib (the one which attests itself to the remote party and runs inside the Gramine-SGX enclave) were not freed if the user of the Secret Prov lib used `secret_provision_start(..., &ctx)` API. This led to memory and resource leaks. In particular, the TLS connection to the remote party was shutdown but the underlying FD was not closed.

Kudos to Liang Fang <liang.a.fang@intel.com> who reported this leak.

Closes #646.

## How to test this PR? 

One can manually try to hit a resource limit using the `secret_prov_client` example:
```diff
diff --git a/CI-Examples/ra-tls-secret-prov/src/secret_prov_client.c

 #define CA_CRT_PATH "ssl/ca.crt"

-int main(int argc, char** argv) {
+int loop(void) {
     int ret;
     int bytes;

@@ -83,3 +83,14 @@ out:
     secret_provision_close(&ctx);
     return ret;
 }
+
+int main(int argc, char** argv) {
+    setbuf(stdout, NULL);
+    for (int i = 0; i < 1024 * 1024; i++) {
+        printf("[%06d] Receiving secret\n", i);
+        int ret = loop();
+        if (ret < 0)
+            return ret;
+    }
+    printf("DONE");
+    return 0;
+}
```

If you want to try out this patch, be careful -- this will try to generate 1,000,000 requests to the IAS web service (if you use EPID), which will throttle your connections and will be slow and painful. Hopefully Intel won't ban me for this stress test :)

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/647)

### DIFF
--- a/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c
+++ b/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c
@@ -57,33 +57,41 @@ void secret_provision_destroy(void) {
     g_provisioned_secret_size = 0;
 }
 
-extern int common_write(mbedtls_ssl_context* ssl, const uint8_t* buf, size_t size);
-extern int common_read(mbedtls_ssl_context* ssl, uint8_t* buf, size_t size);
-extern int common_close(mbedtls_ssl_context* ssl);
+extern int secret_provision_common_write(mbedtls_ssl_context* ssl, const uint8_t* buf, size_t size);
+extern int secret_provision_common_read(mbedtls_ssl_context* ssl, uint8_t* buf, size_t size);
+extern int secret_provision_common_close(mbedtls_ssl_context* ssl);
 
 int secret_provision_write(struct ra_tls_ctx* ctx, const uint8_t* buf, size_t size) {
     mbedtls_ssl_context* ssl = (mbedtls_ssl_context*)ctx->ssl;
-    return common_write(ssl, buf, size);
+    return secret_provision_common_write(ssl, buf, size);
 }
 
 int secret_provision_read(struct ra_tls_ctx* ctx, uint8_t* buf, size_t size) {
     mbedtls_ssl_context* ssl = (mbedtls_ssl_context*)ctx->ssl;
-    return common_read(ssl, buf, size);
+    return secret_provision_common_read(ssl, buf, size);
 }
 
 int secret_provision_close(struct ra_tls_ctx* ctx) {
-    mbedtls_ssl_context* ssl = (mbedtls_ssl_context*)ctx->ssl;
-    mbedtls_ssl_config* conf = (mbedtls_ssl_config*)ctx->conf;
-    mbedtls_net_context* net = (mbedtls_net_context*)ctx->net;
-
-    int ret = common_close(ssl);
-
-    mbedtls_ssl_free(ssl);
-    mbedtls_ssl_config_free(conf);
-    mbedtls_net_free(net);
-    free(ssl);
-    free(conf);
-    free(net);
+    int ret = 0;
+    if (ctx->ssl) {
+        mbedtls_ssl_context* ssl = (mbedtls_ssl_context*)ctx->ssl;
+        ret = secret_provision_common_close(ssl);
+        mbedtls_ssl_free(ssl);
+        free(ssl);
+        ctx->ssl = NULL;
+    }
+    if (ctx->conf) {
+        mbedtls_ssl_config* conf = (mbedtls_ssl_config*)ctx->conf;
+        mbedtls_ssl_config_free(conf);
+        free(conf);
+        ctx->conf = NULL;
+    }
+    if (ctx->net) {
+        mbedtls_net_context* net = (mbedtls_net_context*)ctx->net;
+        mbedtls_net_free(net);
+        free(net);
+        ctx->net = NULL;
+    }
     return ret;
 }
 
@@ -107,6 +115,9 @@ int secret_provision_start(const char* in_servers, const char* in_ca_chain_path,
     mbedtls_ssl_config*  conf        = malloc(sizeof(*conf));
     mbedtls_ssl_context* ssl         = malloc(sizeof(*ssl));
     if (!verifier_fd || !conf || !ssl) {
+        free(verifier_fd);
+        free(conf);
+        free(ssl);
         return -ENOMEM;
     }
 
@@ -247,7 +258,7 @@ int secret_provision_start(const char* in_servers, const char* in_ca_chain_path,
     size = sprintf((char*)buf, SECRET_PROVISION_REQUEST);
     size += 1; /* include null byte */
 
-    ret = common_write(ssl, buf, size);
+    ret = secret_provision_common_write(ssl, buf, size);
     if (ret < 0) {
         goto out;
     }
@@ -258,7 +269,8 @@ int secret_provision_start(const char* in_servers, const char* in_ca_chain_path,
                   "buffer must be sufficiently large to hold SECRET_PROVISION_RESPONSE + int32");
 
     memset(buf, 0, sizeof(buf));
-    ret = common_read(ssl, buf, sizeof(SECRET_PROVISION_RESPONSE) + sizeof(received_secret_size));
+    ret = secret_provision_common_read(ssl, buf, sizeof(SECRET_PROVISION_RESPONSE) +
+                                                 sizeof(received_secret_size));
     if (ret < 0) {
         goto out;
     }
@@ -287,7 +299,7 @@ int secret_provision_start(const char* in_servers, const char* in_ca_chain_path,
     }
     g_provisioned_secret_size = received_secret_size;
 
-    ret = common_read(ssl, g_provisioned_secret, g_provisioned_secret_size);
+    ret = secret_provision_common_read(ssl, g_provisioned_secret, g_provisioned_secret_size);
     if (ret < 0) {
         goto out;
     }
@@ -299,7 +311,7 @@ int secret_provision_start(const char* in_servers, const char* in_ca_chain_path,
         out_ctx->conf = conf;
         out_ctx->net  = verifier_fd;
     } else {
-        common_close(ssl);
+        secret_provision_common_close(ssl);
     }
 
     ret = 0;
--- a/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c
+++ b/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c
@@ -44,28 +44,39 @@ static mbedtls_ssl_context g_ssl;
 static mbedtls_pk_context g_my_ratls_key;
 static mbedtls_x509_crt g_my_ratls_cert;
 
-static uint8_t* provisioned_secret = NULL;
-static size_t provisioned_secret_size = 0;
+static uint8_t* g_provisioned_secret = NULL;
+static size_t g_provisioned_secret_size = 0;
+
+void secret_provision_free_resources(void) {
+    mbedtls_x509_crt_free(&g_my_ratls_cert);
+    mbedtls_pk_free(&g_my_ratls_key);
+    mbedtls_net_free(&g_verifier_fd);
+    mbedtls_ssl_free(&g_ssl);
+    mbedtls_ssl_config_free(&g_conf);
+    mbedtls_x509_crt_free(&g_verifier_ca_chain);
+    mbedtls_ctr_drbg_free(&g_ctr_drbg);
+    mbedtls_entropy_free(&g_entropy);
+}
 
 int secret_provision_get(uint8_t** out_secret, size_t* out_secret_size) {
     if (!out_secret || !out_secret_size)
         return -EINVAL;
 
-    *out_secret      = provisioned_secret;
-    *out_secret_size = provisioned_secret_size;
+    *out_secret      = g_provisioned_secret;
+    *out_secret_size = g_provisioned_secret_size;
     return 0;
 }
 
 void secret_provision_destroy(void) {
-    if (provisioned_secret && provisioned_secret_size)
+    if (g_provisioned_secret && g_provisioned_secret_size)
 #ifdef __STDC_LIB_EXT1__
-        memset_s(provisioned_secret, 0, provisioned_secret_size);
+        memset_s(g_provisioned_secret, 0, g_provisioned_secret_size);
 #else
-        memset(provisioned_secret, 0, provisioned_secret_size);
+        memset(g_provisioned_secret, 0, g_provisioned_secret_size);
 #endif
-    free(provisioned_secret);
-    provisioned_secret      = NULL;
-    provisioned_secret_size = 0;
+    free(g_provisioned_secret);
+    g_provisioned_secret      = NULL;
+    g_provisioned_secret_size = 0;
 }
 
 int secret_provision_start(const char* in_servers, const char* in_ca_chain_path,
@@ -248,14 +259,17 @@ int secret_provision_start(const char* in_servers, const char* in_ca_chain_path,
         goto out;
     }
 
-    provisioned_secret_size = received_secret_size;
-    provisioned_secret = malloc(provisioned_secret_size);
-    if (!provisioned_secret) {
+    /* destroy a previously provisioned secret, if any */
+    secret_provision_destroy();
+
+    g_provisioned_secret_size = received_secret_size;
+    g_provisioned_secret = malloc(g_provisioned_secret_size);
+    if (!g_provisioned_secret) {
         ret = -ENOMEM;
         goto out;
     }
 
-    ret = secret_provision_read(&ctx, provisioned_secret, provisioned_secret_size);
+    ret = secret_provision_read(&ctx, g_provisioned_secret, g_provisioned_secret_size);
     if (ret < 0) {
         goto out;
     }
@@ -270,17 +284,7 @@ int secret_provision_start(const char* in_servers, const char* in_ca_chain_path,
 out:
     if (ret < 0) {
         secret_provision_destroy();
-    }
-
-    if (ret < 0 || !out_ctx) {
-        mbedtls_x509_crt_free(&g_my_ratls_cert);
-        mbedtls_pk_free(&g_my_ratls_key);
-        mbedtls_net_free(&g_verifier_fd);
-        mbedtls_ssl_free(&g_ssl);
-        mbedtls_ssl_config_free(&g_conf);
-        mbedtls_x509_crt_free(&g_verifier_ca_chain);
-        mbedtls_ctr_drbg_free(&g_ctr_drbg);
-        mbedtls_entropy_free(&g_entropy);
+        secret_provision_free_resources();
     }
 
     free(servers);
@@ -348,7 +352,8 @@ __attribute__((constructor)) static void secret_provision_constructor(void) {
 
             size_t total_written = 0;
             while (total_written < sizeof(keydata)) {
-                ssize_t written = write(fd, keydata + total_written, sizeof(keydata) - total_written);
+                ssize_t written = write(fd, keydata + total_written,
+                                        sizeof(keydata) - total_written);
                 if (written > 0) {
                     total_written += written;
                 } else if (written == 0) {
--- a/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c
+++ b/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c
@@ -33,30 +33,8 @@
 #include "secret_prov.h"
 #include "util.h"
 
-/* these are globals because the user may continue using the SSL session even after invoking
- * secret_provision_start() (in the user-supplied callback) */
-static mbedtls_ctr_drbg_context g_ctr_drbg;
-static mbedtls_entropy_context g_entropy;
-static mbedtls_ssl_config g_conf;
-static mbedtls_x509_crt g_verifier_ca_chain;
-static mbedtls_net_context g_verifier_fd;
-static mbedtls_ssl_context g_ssl;
-static mbedtls_pk_context g_my_ratls_key;
-static mbedtls_x509_crt g_my_ratls_cert;
-
 static uint8_t* g_provisioned_secret = NULL;
 static size_t g_provisioned_secret_size = 0;
-
-void secret_provision_free_resources(void) {
-    mbedtls_x509_crt_free(&g_my_ratls_cert);
-    mbedtls_pk_free(&g_my_ratls_key);
-    mbedtls_net_free(&g_verifier_fd);
-    mbedtls_ssl_free(&g_ssl);
-    mbedtls_ssl_config_free(&g_conf);
-    mbedtls_x509_crt_free(&g_verifier_ca_chain);
-    mbedtls_ctr_drbg_free(&g_ctr_drbg);
-    mbedtls_entropy_free(&g_entropy);
-}
 
 int secret_provision_get(uint8_t** out_secret, size_t* out_secret_size) {
     if (!out_secret || !out_secret_size)
@@ -79,6 +57,36 @@ void secret_provision_destroy(void) {
     g_provisioned_secret_size = 0;
 }
 
+extern int common_write(mbedtls_ssl_context* ssl, const uint8_t* buf, size_t size);
+extern int common_read(mbedtls_ssl_context* ssl, uint8_t* buf, size_t size);
+extern int common_close(mbedtls_ssl_context* ssl);
+
+int secret_provision_write(struct ra_tls_ctx* ctx, const uint8_t* buf, size_t size) {
+    mbedtls_ssl_context* ssl = (mbedtls_ssl_context*)ctx->ssl;
+    return common_write(ssl, buf, size);
+}
+
+int secret_provision_read(struct ra_tls_ctx* ctx, uint8_t* buf, size_t size) {
+    mbedtls_ssl_context* ssl = (mbedtls_ssl_context*)ctx->ssl;
+    return common_read(ssl, buf, size);
+}
+
+int secret_provision_close(struct ra_tls_ctx* ctx) {
+    mbedtls_ssl_context* ssl = (mbedtls_ssl_context*)ctx->ssl;
+    mbedtls_ssl_config* conf = (mbedtls_ssl_config*)ctx->conf;
+    mbedtls_net_context* net = (mbedtls_net_context*)ctx->net;
+
+    int ret = common_close(ssl);
+
+    mbedtls_ssl_free(ssl);
+    mbedtls_ssl_config_free(conf);
+    mbedtls_net_free(net);
+    free(ssl);
+    free(conf);
+    free(net);
+    return ret;
+}
+
 int secret_provision_start(const char* in_servers, const char* in_ca_chain_path,
                            struct ra_tls_ctx* out_ctx) {
     int ret;
@@ -89,19 +97,31 @@ int secret_provision_start(const char* in_servers, const char* in_ca_chain_path,
     char* connected_addr = NULL;
     char* connected_port = NULL;
 
-    mbedtls_ctr_drbg_init(&g_ctr_drbg);
-    mbedtls_entropy_init(&g_entropy);
-    mbedtls_x509_crt_init(&g_verifier_ca_chain);
+    mbedtls_ctr_drbg_context ctr_drbg;
+    mbedtls_entropy_context entropy;
+    mbedtls_x509_crt verifier_ca_chain;
+    mbedtls_pk_context my_ratls_key;
+    mbedtls_x509_crt my_ratls_cert;
 
-    mbedtls_pk_init(&g_my_ratls_key);
-    mbedtls_x509_crt_init(&g_my_ratls_cert);
+    mbedtls_net_context* verifier_fd = malloc(sizeof(*verifier_fd));
+    mbedtls_ssl_config*  conf        = malloc(sizeof(*conf));
+    mbedtls_ssl_context* ssl         = malloc(sizeof(*ssl));
+    if (!verifier_fd || !conf || !ssl) {
+        return -ENOMEM;
+    }
 
-    mbedtls_net_init(&g_verifier_fd);
-    mbedtls_ssl_config_init(&g_conf);
-    mbedtls_ssl_init(&g_ssl);
+    mbedtls_ctr_drbg_init(&ctr_drbg);
+    mbedtls_entropy_init(&entropy);
+    mbedtls_x509_crt_init(&verifier_ca_chain);
+    mbedtls_pk_init(&my_ratls_key);
+    mbedtls_x509_crt_init(&my_ratls_cert);
+
+    mbedtls_net_init(verifier_fd);
+    mbedtls_ssl_config_init(conf);
+    mbedtls_ssl_init(ssl);
 
     const char* pers = "secret-provisioning";
-    ret = mbedtls_ctr_drbg_seed(&g_ctr_drbg, mbedtls_entropy_func, &g_entropy,
+    ret = mbedtls_ctr_drbg_seed(&ctr_drbg, mbedtls_entropy_func, &entropy,
                                 (const uint8_t*)pers, strlen(pers));
     if (ret < 0) {
         goto out;
@@ -148,7 +168,7 @@ int secret_provision_start(const char* in_servers, const char* in_ca_chain_path,
         if (!connected_port)
             continue;
 
-        ret = mbedtls_net_connect(&g_verifier_fd, connected_addr, connected_port,
+        ret = mbedtls_net_connect(verifier_fd, connected_addr, connected_port,
                                   MBEDTLS_NET_PROTO_TCP);
         if (!ret)
             break;
@@ -158,53 +178,53 @@ int secret_provision_start(const char* in_servers, const char* in_ca_chain_path,
         goto out;
     }
 
-    ret = mbedtls_ssl_config_defaults(&g_conf, MBEDTLS_SSL_IS_CLIENT, MBEDTLS_SSL_TRANSPORT_STREAM,
+    ret = mbedtls_ssl_config_defaults(conf, MBEDTLS_SSL_IS_CLIENT, MBEDTLS_SSL_TRANSPORT_STREAM,
                                       MBEDTLS_SSL_PRESET_DEFAULT);
     if (ret < 0) {
         goto out;
     }
 
-    ret = mbedtls_x509_crt_parse_file(&g_verifier_ca_chain, ca_chain_path);
+    ret = mbedtls_x509_crt_parse_file(&verifier_ca_chain, ca_chain_path);
     if (ret != 0) {
         goto out;
     }
 
     char crt_issuer[256];
-    ret = mbedtls_x509_dn_gets(crt_issuer, sizeof(crt_issuer), &g_verifier_ca_chain.issuer);
+    ret = mbedtls_x509_dn_gets(crt_issuer, sizeof(crt_issuer), &verifier_ca_chain.issuer);
     if (ret < 0) {
         goto out;
     }
 
-    mbedtls_ssl_conf_authmode(&g_conf, MBEDTLS_SSL_VERIFY_REQUIRED);
-    mbedtls_ssl_conf_ca_chain(&g_conf, &g_verifier_ca_chain, NULL);
+    mbedtls_ssl_conf_authmode(conf, MBEDTLS_SSL_VERIFY_REQUIRED);
+    mbedtls_ssl_conf_ca_chain(conf, &verifier_ca_chain, NULL);
 
-    ret = ra_tls_create_key_and_crt(&g_my_ratls_key, &g_my_ratls_cert);
+    ret = ra_tls_create_key_and_crt(&my_ratls_key, &my_ratls_cert);
     if (ret < 0) {
         goto out;
     }
 
-    mbedtls_ssl_conf_rng(&g_conf, mbedtls_ctr_drbg_random, &g_ctr_drbg);
+    mbedtls_ssl_conf_rng(conf, mbedtls_ctr_drbg_random, &ctr_drbg);
 
-    ret = mbedtls_ssl_conf_own_cert(&g_conf, &g_my_ratls_cert, &g_my_ratls_key);
+    ret = mbedtls_ssl_conf_own_cert(conf, &my_ratls_cert, &my_ratls_key);
     if (ret < 0) {
         goto out;
     }
 
-    ret = mbedtls_ssl_setup(&g_ssl, &g_conf);
+    ret = mbedtls_ssl_setup(ssl, conf);
     if (ret < 0) {
         goto out;
     }
 
-    ret = mbedtls_ssl_set_hostname(&g_ssl, connected_addr);
+    ret = mbedtls_ssl_set_hostname(ssl, connected_addr);
     if (ret < 0) {
         goto out;
     }
 
-    mbedtls_ssl_set_bio(&g_ssl, &g_verifier_fd, mbedtls_net_send, mbedtls_net_recv, NULL);
+    mbedtls_ssl_set_bio(ssl, verifier_fd, mbedtls_net_send, mbedtls_net_recv, NULL);
 
     ret = -1;
     while (ret < 0) {
-        ret = mbedtls_ssl_handshake(&g_ssl);
+        ret = mbedtls_ssl_handshake(ssl);
         if (ret == MBEDTLS_ERR_SSL_WANT_READ || ret == MBEDTLS_ERR_SSL_WANT_WRITE) {
             continue;
         }
@@ -213,13 +233,12 @@ int secret_provision_start(const char* in_servers, const char* in_ca_chain_path,
         }
     }
 
-    uint32_t flags = mbedtls_ssl_get_verify_result(&g_ssl);
+    uint32_t flags = mbedtls_ssl_get_verify_result(ssl);
     if (flags != 0) {
         ret = MBEDTLS_ERR_X509_CERT_VERIFY_FAILED;
         goto out;
     }
 
-    struct ra_tls_ctx ctx = {.ssl = &g_ssl};
     uint8_t buf[128] = {0};
     size_t size;
 
@@ -228,7 +247,7 @@ int secret_provision_start(const char* in_servers, const char* in_ca_chain_path,
     size = sprintf((char*)buf, SECRET_PROVISION_REQUEST);
     size += 1; /* include null byte */
 
-    ret = secret_provision_write(&ctx, buf, size);
+    ret = common_write(ssl, buf, size);
     if (ret < 0) {
         goto out;
     }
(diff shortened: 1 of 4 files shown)


Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c
```diff
@@ -248,14 +259,17 @@ int secret_provision_start(const char* in_servers, const char* in_ca_chain_path,
         goto out;
     }
 
-    provisioned_secret_size = received_secret_size;
-    provisioned_secret = malloc(provisioned_secret_size);
-    if (!provisioned_secret) {
+    /* destroy a previously provisioned secret, if any */
+    secret_provision_destroy();
+
+    g_provisioned_secret_size = received_secret_size;
```
liangintel, line 265: seems this line should be moved to line 271 where malloc have succeeded.

Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c
```diff
@@ -44,28 +44,39 @@ static mbedtls_ssl_context g_ssl;
...
-        memset_s(provisioned_secret, 0, provisioned_secret_size);
+        memset_s(g_provisioned_secret, 0, g_provisioned_secret_size);
 #else
-        memset(provisioned_secret, 0, provisioned_secret_size);
+        memset(g_provisioned_secret, 0, g_provisioned_secret_size);
 #endif
-    free(provisioned_secret);
-    provisioned_secret      = NULL;
-    provisioned_secret_size = 0;
+    free(g_provisioned_secret);
```
liangintel, line 77: seems need to add"if (g_provisioned_secret)", otherwise free() may lead to exception when calling secret_provision_destroy() twice

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/647)* status: 0 of 4 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @liangintel)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/647#-N4CGcnEBl9wigdoLrT8:-N4CGcnF-imSyPaIN3ff:b3891mk) (no related file):*
UPDATE: In my experiments, I went to `[001026] Receiving secret` in the modified `secret_prov_client`, and I stopped there. Seems that no resource leaks are there (I examined opened FDs of the process, looked good).

___
___
*[`Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c` line 77 at r1](https://reviewable.io/reviews/gramineproject/gramine/647#-N4BsDJzFIGmxVwhIaGj-r1-77:-N4CGscx1LPOo3hkN_gL:bwda768) ([raw file](https://github.com/gramineproject/gramine/blob/91b764bb67f985c54425e112551bb8e6efe0921d/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c#L77)):*
<details><summary><i>Previously, liangintel (Liang Fang) wrote…</i></summary><blockquote>

seems need to add"if (g_provisioned_secret)", otherwise free() may lead to exception when calling secret_provision_destroy() twice
</blockquote></details>

No, `free(NULL)` does nothing bad. So there is no need for an IF statement.
___
*[`Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c` line 265 at r1](https://reviewable.io/reviews/gramineproject/gramine/647#-N4BsDJzFIGmxVwhIaGj-r1-265:-N4CH9ET9t9mlMKztxVM:b-376vyb) ([raw file](https://github.com/gramineproject/gramine/blob/91b764bb67f985c54425e112551bb8e6efe0921d/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c#L265)):*
<details><summary><i>Previously, liangintel (Liang Fang) wrote…</i></summary><blockquote>

seems this line should be moved to line 271 where malloc have succeeded.
</blockquote></details>

Done. Good catch.

boryspoplawski (changes requested):
> Reviewed 3 of 4 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/647)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @liangintel)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/647#-N4D3wOMB7GvhpJ3ZfPe:-N4D3wOMB7GvhpJ3ZfPf:b-kuug7w) (no related file):*
@liangintel please use reviewable for commenting / review

___
___
*[`Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c` line 287 at r2](https://reviewable.io/reviews/gramineproject/gramine/647#-N4D57O4E3vrt0zli1Lv:-N4D57O4E3vrt0zli1Lw:bi0re3h) ([raw file](https://github.com/gramineproject/gramine/blob/73a14b59bacf7f1c079618f5b3349e042653ce91/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c#L287)):*
> ```C
>     if (ret < 0) {
>         secret_provision_destroy();
>         secret_provision_free_resources();
> ```

What about case of `!out_ctx` which was handled before?
Update: dont we just miss `secret_provision_close` here in some cases?
This API is confusing with `close`, `destroy` and ` free`...
___
*[`Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_common.c` line 86 at r2](https://reviewable.io/reviews/gramineproject/gramine/647#-N4D6IdR51p9_DyEmGnz:-N4D6IdR51p9_DyEmGo-:b4j7w4e) ([raw file](https://github.com/gramineproject/gramine/blob/73a14b59bacf7f1c079618f5b3349e042653ce91/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_common.c#L86)):*
> ```C
>             /* use well-known error code for a typical case when remote party closes connection */
>             if (ret == MBEDTLS_ERR_SSL_PEER_CLOSE_NOTIFY)
>                 ret = -ECONNRESET;
> ```

ughh, we mix mbedtls error codes with unix ones?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/647)* status: 3 of 4 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @liangintel)

___
*[`Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c` line 287 at r2](https://reviewable.io/reviews/gramineproject/gramine/647#-N4D57O4E3vrt0zli1Lv:-N4Qs0Mw2VR-5nC1Cjoy:bmcgwbr) ([raw file](https://github.com/gramineproject/gramine/blob/73a14b59bacf7f1c079618f5b3349e042653ce91/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c#L287)):*
Yes, I agree that the naming got confusing. A quick explanation:
- `close` terminates the TLS connection gracefully and frees the used resources, if any (by calling `free_resources`).
- `destroy` deallocates the secret blob (nothing more than that).
- `free_resources` is a helper func to free the resources used by the TLS connection.

`close` and `destroy` are user-facing APIs of the Secret Prov library, and so we cannot modify them. We can modify `free_resources` though, if you wish so -- we can change its name to something better.

> Update: dont we just miss `secret_provision_close` here in some cases?

Not sure what you mean. If we are in this error handling, then we don't care about graceful termination of the TLS session, so no need to call `close()`. It is enough to call `free_resources()` to kill the connection.
___
*[`Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_common.c` line 86 at r2](https://reviewable.io/reviews/gramineproject/gramine/647#-N4D6IdR51p9_DyEmGnz:-N4Qtf4k6pQFBQelQx5X:bivq0l5) ([raw file](https://github.com/gramineproject/gramine/blob/73a14b59bacf7f1c079618f5b3349e042653ce91/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_common.c#L86)):*

ughh, we mix mbedtls error codes with unix ones?
</blockquote></details>

Done, you're right. Hopefully this is good enough for this case.

But I looked at other functions in the Secret Prov library, we have this mixing of mbedTLS vs UNIX error codes in other funcs. Do we want to fix it (not in this PR)? If yes, then how -- convert all mbedTLS error codes to "close enough" UNIX error codes?

liangintel (commented):
I ran my workload(https://github.com/intel/confidential-computing-zoo/tree/branch-dev/cross_lang_framework/cczoo/cross_lang_framework), and it's fine now.

> Reviewed 2 of 4 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/647#-:-N4UpMzK1bGqcgCaSFvz:blg2jyt)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @liangintel)

liangintel (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/647#-:-N4Upzr_AbFt8a9HxEph:bnfp4nl)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c` line 287 at r2](https://reviewable.io/reviews/gramineproject/gramine/647#-N4D57O4E3vrt0zli1Lv:-N4UqSJo9_ytW_8pPpXq:b-uxucji) ([raw file](https://github.com/gramineproject/gramine/blob/73a14b59bacf7f1c079618f5b3349e042653ce91/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c#L287)):*

>  are user-facing APIs of the Secret Prov library, and so we cannot modify them. 

Why we cannot modify them?

> Not sure what you mean. If we are in this error handling, then we don't care about graceful termination of the TLS session, so no need to call close(). It is enough to call free_resources() to kill the connection.

The old version called `free` if `!out_ctx`, the new version does not
</blockquote></details>

from my point of view, the function secret_provision_destroy() did too few things, and destroy should be the last step (open->close->destroy). Meanwhile, because no bug has been reported previously, I guess this API is not widely used. So we may can change this API here.

liangintel (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/647)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c` line 287 at r2](https://reviewable.io/reviews/gramineproject/gramine/647#-N4D57O4E3vrt0zli1Lv:-N4UxHo3A6uxKpzshha1:bpsg5z6) ([raw file](https://github.com/gramineproject/gramine/blob/73a14b59bacf7f1c079618f5b3349e042653ce91/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c#L287)):*
<details><summary><i>Previously, liangintel (Liang Fang) wrote…</i></summary><blockquote>

from my point of view, the function secret_provision_destroy() did too few things, and destroy should be the last step (open->close->destroy). Meanwhile, because no bug has been reported previously, I guess this API is not widely used. So we may can change this API here.
</blockquote></details>

I think we can let the bug fix go first (a customer is evaluating my cross-language framework, and the bug fix is urgent), and code refine can be in another PR

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/647)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c` line 287 at r2](https://reviewable.io/reviews/gramineproject/gramine/647#-N4D57O4E3vrt0zli1Lv:-N4W1NYhEGLaDIehGZLy:b-x8ulux) ([raw file](https://github.com/gramineproject/gramine/blob/73a14b59bacf7f1c079618f5b3349e042653ce91/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c#L287)):*
> The old version called `free` if `!out_ctx`, the new version does not

This is because when `!out_ctx`, then we do `secret_provision_close(&ctx)`; see my other comment for the exact line of code.

And in this PR, I add the call to `secret_provision_free_resources()` in this function `secret_provision_close()`. So I shifted the old "resource freeing" code there.

> > are user-facing APIs of the Secret Prov library, and so we cannot modify them.

> Why we cannot modify them?

Because this is a Secret Provisioning library that is used by users -- I know at least a couple users (including production systems) that use it. We will need some kind of a versioning/deprecation strategy. But this goes out of scope of this PR, I think. We can surely discuss this.

That said, I regret I haven't introduced a proper opaque object as an argument to all these functions (like the `context`). And the names could be better.
___
*[`Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c` line 280 at r3](https://reviewable.io/reviews/gramineproject/gramine/647#-N4W1XxDCJeMZIkm144q:-N4W1XxDCJeMZIkm144r:bhp8j7y) ([raw file](https://github.com/gramineproject/gramine/blob/fc7f1d2969908cb0be19bc57c74d9b371c9dd3d0/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c#L280)):*
> ```C
>         out_ctx->ssl = ctx.ssl;
>     } else {
>         secret_provision_close(&ctx);
> ```

FYI: @boryspoplawski , here we handle the case `!out_ctx`.
___
*[`Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_common.c` line 86 at r2](https://reviewable.io/reviews/gramineproject/gramine/647#-N4D6IdR51p9_DyEmGnz:-N4W1u-U6YAOLga4oOvc:bwgwb9l) ([raw file](https://github.com/gramineproject/gramine/blob/73a14b59bacf7f1c079618f5b3349e042653ce91/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_common.c#L86)):*
> Maybe mbedtls has some option to embed unix errno inside their errors?

No, I looked into this. Unfortunately, no such thing. We'll have to introduce such a function ourselves.

I'm actually deeply in this RA-TLS and Secret Provisioning code now, and I see that we mix error codes in many places there. This will need a proper PR to untangle, and a discussion beforehand -- do we want to always return mbedTLS error codes, or UNIX error codes? Can we mix them sometimes (but looks like not -- mbedTLS codes can overlap with UNIX ones, and the whole mbedTLS codebase is very accurate to use only mbedTLS codes)?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/647)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c` line 287 at r2](https://reviewable.io/reviews/gramineproject/gramine/647#-N4D57O4E3vrt0zli1Lv:-N4Wtcky6baoTcBQnUzp:b2wdh6s) ([raw file](https://github.com/gramineproject/gramine/blob/73a14b59bacf7f1c079618f5b3349e042653ce91/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c#L287)):*
> That said, I regret I haven't introduced a proper opaque object as an argument to all these functions (like the context). And the names could be better.

Wait, sorry, this was my mental lapse. We actually have the opaque `struct ra_tls_ctx* context` in every SecretProv function. So I could stash those global variables there, if needed. (This is actually how @liangintel did it in his original PR.)

mkow (changes requested):
> Reviewed 1 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/647)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[`Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c` line 37 at r3](https://reviewable.io/reviews/gramineproject/gramine/647#-N4X6qkC5FzgsGAJWGaH:-N4X6qkDBJrO-LJJi2KT:bk8385k) ([raw file](https://github.com/gramineproject/gramine/blob/fc7f1d2969908cb0be19bc57c74d9b371c9dd3d0/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c#L37)):*
> ```C
> 
> /* these are globals because the user may continue using the SSL session even after invoking
>  * secret_provision_start() (in the user-supplied callback) */
> ```

Why can't they reside in the context? Can the user use them after the context was destroyed? If so, then this PR will break this case?
I mean, this comment seem to stand in the opposition of what this PR does. Am I missing something?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/647)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[`Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c` line 37 at r3](https://reviewable.io/reviews/gramineproject/gramine/647#-N4X6qkC5FzgsGAJWGaH:-N4XTgj-2eGLyFWZ2Wsn:bqaakm0) ([raw file](https://github.com/gramineproject/gramine/blob/fc7f1d2969908cb0be19bc57c74d9b371c9dd3d0/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c#L37)):*
> Why can't they reside in the context?

They could, I just didn't implement it at the time (and I didn't implement it in this PR as well, for the sake of minimality of changes).

> Can the user use them after the context was destroyed?

No, the user can't use them after he calls `secret_provision_close()` -- at this point the context is destroyed (via calls like `mbedtls_x509_crt_free(&g_my_ratls_cert)`).

> I mean, this comment seem to stand in the opposition of what this PR does.

No, why. The user calls `secret_provision_start()`, then does some `secret_provision_read`/`secret_provision_write`, and then -- when the user is done with secret provisioning -- closes the secret-provisioning session with `secret_provision_close()`. See the example: https://github.com/gramineproject/gramine/blob/master/CI-Examples/ra-tls-secret-prov/src/secret_prov_client.c

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/647)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[`Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c` line 37 at r3](https://reviewable.io/reviews/gramineproject/gramine/647#-N4X6qkC5FzgsGAJWGaH:-N4XgkKD72OJ0G801-BW:b-norwn) ([raw file](https://github.com/gramineproject/gramine/blob/fc7f1d2969908cb0be19bc57c74d9b371c9dd3d0/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c#L37)):*

> Why can't they reside in the context?

They could, I just didn't implement it at the time (and I didn't implement it in this PR as well, for the sake of minimality of changes).

> Can the user use them after the context was destroyed?

No, the user can't use them after he calls `secret_provision_close()` -- at this point the context is destroyed (via calls like `mbedtls_x509_crt_free(&g_my_ratls_cert)`).

> I mean, this comment seem to stand in the opposition of what this PR does.

No, why. The user calls `secret_provision_start()`, then does some `secret_provision_read`/`secret_provision_write`, and then -- when the user is done with secret provisioning -- closes the secret-provisioning session with `secret_provision_close()`. See the example: https://github.com/gramineproject/gramine/blob/master/CI-Examples/ra-tls-secret-prov/src/secret_prov_client.c
</blockquote></details>

Ok, then I think we really should move them to the context if they die together with it. Right now this library is a weird singleton object with some of its fields taken out as globals, but still destroying them in its destructor.
It should also make this PR more logical, IMO (and the whole library).

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/647)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @liangintel)

___
*[`Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c` line 287 at r2](https://reviewable.io/reviews/gramineproject/gramine/647#-N4D57O4E3vrt0zli1Lv:-N4Ye81PBOVh0Od9tMcD:b-2t1xwk) ([raw file](https://github.com/gramineproject/gramine/blob/73a14b59bacf7f1c079618f5b3349e042653ce91/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c#L287)):*

> That said, I regret I haven't introduced a proper opaque object as an argument to all these functions (like the context). And the names could be better.

Wait, sorry, this was my mental lapse. We actually have the opaque `struct ra_tls_ctx* context` in every SecretProv function. So I could stash those global variables there, if needed. (This is actually how @liangintel did it in his original PR.)
</blockquote></details>

Okay, but then why don't we call close on errors here?
Sorry, I still cannot wrap my head around this.
___
*[`Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_common.c` line 86 at r2](https://reviewable.io/reviews/gramineproject/gramine/647#-N4D6IdR51p9_DyEmGnz:-N4YeVYdBK4zsrq3_zV1:bc3iifv) ([raw file](https://github.com/gramineproject/gramine/blob/73a14b59bacf7f1c079618f5b3349e042653ce91/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_common.c#L86)):*

> Maybe mbedtls has some option to embed unix errno inside their errors?

No, I looked into this. Unfortunately, no such thing. We'll have to introduce such a function ourselves.

I'm actually deeply in this RA-TLS and Secret Provisioning code now, and I see that we mix error codes in many places there. This will need a proper PR to untangle, and a discussion beforehand -- do we want to always return mbedTLS error codes, or UNIX error codes? Can we mix them sometimes (but looks like not -- mbedTLS codes can overlap with UNIX ones, and the whole mbedTLS codebase is very accurate to use only mbedTLS codes)?
</blockquote></details>

If they overlap then we cannot. Please add it to your TODO list or create an issue :)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/647)* status: 0 of 4 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @liangintel, and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/647#-N4afGs_4htyQhB-fCHV:-N4afGs_4htyQhB-fCHW:b-ngz9rq) (no related file):*
FYI: I updated the PR significantly; it's best to compare against the base revision, not the last one.

___
___
*[`Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov.h` line 133 at r4](https://reviewable.io/reviews/gramineproject/gramine/647#-N4agUzgCs6_QzHrpIxh:-N4agUzgCs6_QzHrpIxi:bn7sw0c) ([raw file](https://github.com/gramineproject/gramine/blob/f000e8491499c33f6b3c2492a0c510386b7b6e7c/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov.h#L133)):*
> ```objc
> __attribute__ ((visibility("default")))
> int secret_provision_start(const char* in_servers, const char* in_ca_chain_path,
>                            struct ra_tls_ctx* out_ctx);
> ```

FYI: I came up with this wrong function signature. It should have been `struct ra_tls_ctx** out_ctx`, so that this function would *allocate* the context instead of stuffing the user-provided context with values.

Because of this blunder, I can't make `struct ra_tls_ctx` truly opaque... And I don't want to change this function's signature because it is public API.

Well, on the other hand, not a big deal, this Secret Prov library is anyhow just a *reference* implementation. Power users are supposed to write their own, ad-hoc Secret Prov flows, not blindly use this library.
___
*[`Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c` line 287 at r2](https://reviewable.io/reviews/gramineproject/gramine/647#-N4D57O4E3vrt0zli1Lv:-N4agEuZCh-riM0-dbfD:bx58oqz) ([raw file](https://github.com/gramineproject/gramine/blob/73a14b59bacf7f1c079618f5b3349e042653ce91/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c#L287)):*

Okay, but then why don't we call close on errors here?
Sorry, I still cannot wrap my head around this.
</blockquote></details>

Done. The new logic should be more straight-forward.
___
*[`Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c` line 37 at r3](https://reviewable.io/reviews/gramineproject/gramine/647#-N4X6qkC5FzgsGAJWGaH:-N4aem-4BPM87Bd0gIUN:bv4bog4) ([raw file](https://github.com/gramineproject/gramine/blob/fc7f1d2969908cb0be19bc57c74d9b371c9dd3d0/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c#L37)):*

Ok, then I think we really should move them to the context if they die together with it. Right now this library is a weird singleton object with some of its fields taken out as globals, but still destroying them in its destructor.
It should also make this PR more logical, IMO (and the whole library).
</blockquote></details>

Done. Now the global variables are replaced with these two sets of variables:
- Local stack-allocated vars (e.g. `ctr_drbg`, `entropy`): they are used only inside the `secret_provision_start()` function and do *not* constitute the context of the TLS session after it is established;
- Long-lived heap-allocated vars (`ssl`, `net`, `conf`): they can be used outside of the `secret_provision_start()` function and constitute the context of the TLS session.
___
*[`Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c` line 37 at r4](https://reviewable.io/reviews/gramineproject/gramine/647#-N4afUDbB3h3z_tUDQiG:-N4afUDbB3h3z_tUDQiH:b-q1cevd) ([raw file](https://github.com/gramineproject/gramine/blob/f000e8491499c33f6b3c2492a0c510386b7b6e7c/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c#L37)):*
> ```C
> 
> static uint8_t* g_provisioned_secret = NULL;
> static size_t g_provisioned_secret_size = 0;
> ```

FYI: I'm a bit unhappy that the "provisioned secret" object is a global variable, and not part of the context of the TLS session. But we already have the public API `secret_provision_get()` and `secret_provision_destroy()` that don't have a context arg, so this has to stay global.
___
*[`Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c` line 62 at r4](https://reviewable.io/reviews/gramineproject/gramine/647#-N4afmjd40txTU2yff3F:-N4afmje7iuArOJ_JCP7:b-javgmt) ([raw file](https://github.com/gramineproject/gramine/blob/f000e8491499c33f6b3c2492a0c510386b7b6e7c/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c#L62)):*
> ```C
> extern int common_write(mbedtls_ssl_context* ssl, const uint8_t* buf, size_t size);
> extern int common_read(mbedtls_ssl_context* ssl, uint8_t* buf, size_t size);
> extern int common_close(mbedtls_ssl_context* ssl);
> ```

FYI: These `extern`s are a bit ugly, but I didn't want to introduce a new header `secret_prov_internal.h` just for these three funcs. If anyone insists, I can put these declarations in the internal header.
___
*[`Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_common.c` line 86 at r2](https://reviewable.io/reviews/gramineproject/gramine/647#-N4D6IdR51p9_DyEmGnz:-N4a52Xv6zCrmHfLNPXn:b-jaimxg) ([raw file](https://github.com/gramineproject/gramine/blob/73a14b59bacf7f1c079618f5b3349e042653ce91/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_common.c#L86)):*

If they overlap then we cannot. Please add it to your TODO list or create an issue :)
</blockquote></details>

Done. See item 3 in https://github.com/gramineproject/gramine/issues/654.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/647)* status: 0 of 4 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, @liangintel, and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/647#-N4bOS-T6jdPobzr4JHm:-N4bOS-T6jdPobzr4JHn:b-nbhg81) (no related file):*
I think it's the time we tackle the problem of backward compatibility here - we need some way of updating the code with braking changes. We probably can:
A) Not be compatible - the users will need to adjust
B) Be fully compatible (I believe this is the option atm) - I don't think we can keep up with this, at some point (even if not now) we will have to change stuff
C) Create another version of library? This would probably mean having two copies of the same code (because I would expect that some/most of the code would be the same). The older version would have some deprecation time (like 1-2 releases, 3 months, whatever)
or something else I didn't come up with.
I would like to decide it now (before we review this PR), because this might change how we proceed (e.g. option C would most likely mean introducing small bugfix in the old version and moving the rest of the changes to the new version)

___

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/647)* status: 0 of 4 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @liangintel, and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/647#-N4bOS-T6jdPobzr4JHm:-N4bQPWa21RMQLxcb1vC:bkt8uyn) (no related file):*

I think it's the time we tackle the problem of backward compatibility here - we need some way of updating the code with braking changes. We probably can:
A) Not be compatible - the users will need to adjust
B) Be fully compatible (I believe this is the option atm) - I don't think we can keep up with this, at some point (even if not now) we will have to change stuff
C) Create another version of library? This would probably mean having two copies of the same code (because I would expect that some/most of the code would be the same). The older version would have some deprecation time (like 1-2 releases, 3 months, whatever)
or something else I didn't come up with.
I would like to decide it now (before we review this PR), because this might change how we proceed (e.g. option C would most likely mean introducing small bugfix in the old version and moving the rest of the changes to the new version)
</blockquote></details>

I think we're still in a young and unstable phase of the project and we should go with A in cases like this one. To mitigate the problems we could help users adjust to the changes (by describing nicely all the needed changes in release notes under "breaking changes" and by directly helping them to adjust their codebase).

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/647)* status: 0 of 4 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @liangintel, and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/647#-N4bOS-T6jdPobzr4JHm:-N4bixzb4jjvw0q5q_bQ:bxbke7p) (no related file):*

I think we're still in a young and unstable phase of the project and we should go with A in cases like this one. To mitigate the problems we could help users adjust to the changes (by describing nicely all the needed changes in release notes under "breaking changes" and by directly helping them to adjust their codebase).
</blockquote></details>

We had an offline discussion, and the agreement is like this:
1. Use option A -- we don't have many users, so they can adjust,
2. Mention this breaking change in the public API of the affected funcs in this PR description and in the commit message (for history).
3. Describe this breaking change in the Release Notes of the next stable version of Gramine.

___

Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c
```diff
@@ -78,19 +97,31 @@ int secret_provision_start(const char* in_servers, const char* in_ca_chain_path,
...
+    mbedtls_ctr_drbg_context ctr_drbg;
+    mbedtls_entropy_context entropy;
+    mbedtls_x509_crt verifier_ca_chain;
+    mbedtls_pk_context my_ratls_key;
+    mbedtls_x509_crt my_ratls_cert;
+
+    mbedtls_net_context* verifier_fd = malloc(sizeof(*verifier_fd));
+    mbedtls_ssl_config*  conf        = malloc(sizeof(*conf));
+    mbedtls_ssl_context* ssl         = malloc(sizeof(*ssl));
+    if (!verifier_fd || !conf || !ssl) {
```
liangintel, line 109: if verifier_fd was allocated successfully, but ssl failed to be allocated, then return -ENOMEM here will lead to memory leak.

Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c
```diff
@@ -33,39 +33,58 @@
...
-static mbedtls_ssl_config g_conf;
-static mbedtls_x509_crt g_verifier_ca_chain;
-static mbedtls_net_context g_verifier_fd;
-static mbedtls_ssl_context g_ssl;
-static mbedtls_pk_context g_my_ratls_key;
-static mbedtls_x509_crt g_my_ratls_cert;
-
-static uint8_t* provisioned_secret = NULL;
-static size_t provisioned_secret_size = 0;
+static uint8_t* g_provisioned_secret = NULL;
```
liangintel, line 36: global variables means cannot be used in multi-threads environment.

Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c
```diff
@@ -33,39 +33,58 @@
...
+    return common_write(ssl, buf, size);
+}
+
+int secret_provision_read(struct ra_tls_ctx* ctx, uint8_t* buf, size_t size) {
+    mbedtls_ssl_context* ssl = (mbedtls_ssl_context*)ctx->ssl;
+    return common_read(ssl, buf, size);
+}
+
+int secret_provision_close(struct ra_tls_ctx* ctx) {
+    mbedtls_ssl_context* ssl = (mbedtls_ssl_context*)ctx->ssl;
```
liangintel, line 75: secret_provision_close may be called twice by user. Consider to add parameter check here? e,g,
if(!ctx)
if(!ssl)

Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c
```diff
@@ -33,39 +33,58 @@
...
+
+    int ret = common_close(ssl);
+
+    mbedtls_ssl_free(ssl);
+    mbedtls_ssl_config_free(conf);
+    mbedtls_net_free(net);
+    free(ssl);
+    free(conf);
+    free(net);
+    return ret;
```
liangintel, line 87: consider to zero the variables before return.
ctx->net = 0;
ctx->ssl = 0;
...

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/647)* status: 0 of 4 files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @liangintel, and @mkow)

___
*[`Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c` line 36 at r5](https://reviewable.io/reviews/gramineproject/gramine/647#-N4BsDJzFIGmxVwhIaGj-r5-36:-N4fwCeA-FQDbzRdgwr7:b-np6ugl) ([raw file](https://github.com/gramineproject/gramine/blob/3649d7a0990455c0750e62c2ef868e5792f03b18/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c#L36)):*
<details><summary><i>Previously, liangintel (Liang Fang) wrote…</i></summary><blockquote>

global variables means cannot be used in multi-threads environment.
</blockquote></details>

Yes, that's a problem.

And that's why we're discussing to modify the public API of this library to use the context. In particular, for secrets we'll have:
```
int secret_provision_get(struct ra_tls_ctx* ctx, uint8_t** out_secret, size_t* out_secret_size);
void secret_provision_destroy(struct ra_tls_ctx* ctx);
```

And the secret variable will be kept inside the context `ctx`. And it will be thread-safe then.
___
*[`Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c` line 75 at r5](https://reviewable.io/reviews/gramineproject/gramine/647#-N4BsDJzFIGmxVwhIaGj-r5-75:-N4gBiv2-cthRKO1V3Gs:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/3649d7a0990455c0750e62c2ef868e5792f03b18/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c#L75)):*
<details><summary><i>Previously, liangintel (Liang Fang) wrote…</i></summary><blockquote>

secret_provision_close may be called twice by user. Consider to add parameter check here? e,g,
if(!ctx)
if(!ssl)

</blockquote></details>

Done.
___
*[`Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c` line 87 at r5](https://reviewable.io/reviews/gramineproject/gramine/647#-N4BsDJzFIGmxVwhIaGj-r5-87:-N4gBjlzEVuuLnUH-Slg:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/3649d7a0990455c0750e62c2ef868e5792f03b18/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c#L87)):*
<details><summary><i>Previously, liangintel (Liang Fang) wrote…</i></summary><blockquote>

consider to zero the variables before return.
ctx->net = 0;
ctx->ssl = 0;
...
</blockquote></details>

Done.
___
*[`Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c` line 109 at r5](https://reviewable.io/reviews/gramineproject/gramine/647#-N4BsDJzFIGmxVwhIaGj-r5-109:-N4gBrr739DMFqll7B67:b-376vyb) ([raw file](https://github.com/gramineproject/gramine/blob/3649d7a0990455c0750e62c2ef868e5792f03b18/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_attest.c#L109)):*
<details><summary><i>Previously, liangintel (Liang Fang) wrote…</i></summary><blockquote>

if verifier_fd was allocated successfully, but ssl failed to be allocated, then return -ENOMEM here will lead to memory leak.
</blockquote></details>

Done. Good catch.
___
*[`Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_common.c` line 19 at r5](https://reviewable.io/reviews/gramineproject/gramine/647#-N4BsDJzFIGmxVwhIaGk-r5-19:-N4gDKB2BeCh7wWuLOju:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/3649d7a0990455c0750e62c2ef868e5792f03b18/Pal/src/host/Linux-SGX/tools/ra-tls/secret_prov_common.c#L19)):*
<details><summary><i>Previously, liangintel (Liang Fang) wrote…</i></summary><blockquote>

it's too common to use prefix common_ , may harm other function definition in Gramine modules
actually it's tls_
</blockquote></details>

Done.

liangintel (commented):
> Reviewed 1 of 4 files at r4, 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/647)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @liangintel, and @mkow)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/647)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @liangintel, and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/647#-N4bOS-T6jdPobzr4JHm:-N4gyU780r581iz6ebY9:b-tatjdn) (no related file):*

We had an offline discussion, and the agreement is like this:
1. Use option A -- we don't have many users, so they can adjust,
2. Mention this breaking change in the public API of the affected funcs in this PR description and in the commit message (for history).
3. Describe this breaking change in the Release Notes of the next stable version of Gramine.
</blockquote></details>

One of the users who is using the Secret Prov library told me this (paraphrasing): "Our code does not use the public functions of the SecretProv library. We have written our code *based* on the SecretProv code, but we didn't use this library directly."

So we are good to use option A and modify the public API of SecretProv as we see fit.

I will submit a new fixup commit that tweaks the public-function signatures for SecretProv tomorrow, and we can continue this review.

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/647)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @liangintel, and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/647#-N8NUyuh4A1Ggybviz-o:-N8NUyui79IEWGXIOAkL:b-pufvj0) (no related file):*
A reminder to myself: I need to revive this PR as soon as we merge #796 (which should happen today/tomorrow).

___

dimakuv:
I created another PR https://github.com/gramineproject/gramine/pull/813 that supersedes this one. I tried to address all outstanding comments in that new PR. Closing this one.

PR #650, opened 2022-06-14
[LibOS] Add `O_PATH` support to `open` syscall

boryspoplawski:
## Description of the changes 
I'm not sure if this doesn't break something - couldn't wrap my head around semantics of some fs operations in Gramine. Hopefully it's all fine...

## How to test this PR? 
Added a new test.

### DIFF
--- a/LibOS/shim/include/shim_fs.h
+++ b/LibOS/shim/include/shim_fs.h
@@ -126,10 +126,6 @@ struct shim_fs_ops {
     /* seek: the content from the file opened as handle */
     file_off_t (*seek)(struct shim_handle* hdl, file_off_t offset, int whence);
 
-    /* move, copy: rename or duplicate the file */
-    int (*move)(const char* trim_old_name, const char* trim_new_name);
-    int (*copy)(const char* trim_old_name, const char* trim_new_name);
-
     /* Returns 0 on success, -errno on error */
     int (*truncate)(struct shim_handle* hdl, file_off_t len);
 
@@ -138,14 +134,6 @@ struct shim_fs_ops {
 
     /* setflags: set flags of the file */
     int (*setflags)(struct shim_handle* hdl, int flags);
-
-    /*
-     * \brief Deallocate handle data.
-     *
-     * Deallocates any filesystem-specific resources associated with the handle. Called just before
-     * destroying the handle.
-     */
-    void (*hdrop)(struct shim_handle* hdl);
 
     /* lock and unlock the file */
     int (*lock)(const char* trim_name);
@@ -898,6 +886,7 @@ extern struct shim_fs socket_builtin_fs;
 extern struct shim_fs epoll_builtin_fs;
 extern struct shim_fs eventfd_builtin_fs;
 extern struct shim_fs synthetic_builtin_fs;
+extern struct shim_fs path_builtin_fs;
 
 struct shim_fs* find_fs(const char* name);
 
--- a/LibOS/shim/include/shim_handle.h
+++ b/LibOS/shim/include/shim_handle.h
@@ -39,6 +39,7 @@ enum shim_handle_type {
     TYPE_PSEUDO,     /* pseudo nodes (currently directories), handled by `pseudo_*` functions */
     TYPE_TMPFS,      /* string-based files (with data inside dentry), used by `tmpfs` filesystem */
     TYPE_SYNTHETIC,  /* synthetic files, used by `synthetic` filesystem */
+    TYPE_PATH,       /* path to a file (the file is not actually opened) */
 
     /* Pipes and sockets: */
     TYPE_PIPE,       /* pipes, used by `pipe` filesystem */
--- a/LibOS/shim/include/shim_types.h
+++ b/LibOS/shim/include/shim_types.h
@@ -243,4 +243,8 @@ struct shim_lock {
 #define EPOLLNVAL ((uint32_t)0x00000020)
 #endif
 
+#ifndef O_ASYNC
+#define O_ASYNC 020000
+#endif
+
 #endif /* _SHIM_TYPES_H_ */
--- a/LibOS/shim/src/bookkeep/shim_handle.c
+++ b/LibOS/shim/src/bookkeep/shim_handle.c
@@ -503,9 +503,6 @@ void put_handle(struct shim_handle* hdl) {
             hdl->info.sock.peek_buffer = NULL;
         }
 
-        if (hdl->fs && hdl->fs->fs_ops && hdl->fs->fs_ops->hdrop)
-            hdl->fs->fs_ops->hdrop(hdl);
-
         free(hdl->uri);
 
         if (hdl->pal_handle) {
--- a/LibOS/shim/src/fs/chroot/encrypted.c
+++ b/LibOS/shim/src/fs/chroot/encrypted.c
@@ -384,16 +384,18 @@ static int chroot_encrypted_flush(struct shim_handle* hdl) {
     return ret;
 }
 
-static void chroot_encrypted_hdrop(struct shim_handle* hdl) {
+static int chroot_encrypted_close(struct shim_handle* hdl) {
     assert(hdl->type == TYPE_CHROOT_ENCRYPTED);
     if (hdl->inode->type != S_IFREG)
-        return;
+        return 0;
 
     struct shim_encrypted_file* enc = hdl->inode->data;
 
     lock(&hdl->inode->lock);
     encrypted_file_put(enc);
     unlock(&hdl->inode->lock);
+
+    return 0;
 }
 
 static ssize_t chroot_encrypted_read(struct shim_handle* hdl, void* buf, size_t count,
@@ -473,7 +475,7 @@ struct shim_fs_ops chroot_encrypted_fs_ops = {
     .hstat      = &generic_inode_hstat,
     .truncate   = &chroot_encrypted_truncate,
     .poll       = &generic_inode_poll,
-    .hdrop      = &chroot_encrypted_hdrop,
+    .close      = &chroot_encrypted_close,
     .checkpoint = &chroot_encrypted_checkpoint,
     .migrate    = &chroot_encrypted_migrate,
     .mmap       = &generic_emulated_mmap,
--- a/LibOS/shim/src/fs/pipe/fs.c
+++ b/LibOS/shim/src/fs/pipe/fs.c
@@ -131,7 +131,7 @@ static ssize_t pipe_write(struct shim_handle* hdl, const void* buf, size_t count
 }
 
 static int pipe_hstat(struct shim_handle* hdl, struct stat* stat) {
-    /* XXX: Is any of this right?
+    /* XXX: DOES THIS WORK LOL
      * Shouldn't we be using hdl to figure something out?
      * if stat is NULL, should we not return -EFAULT?
      */
--- a/LibOS/shim/src/fs/shim_fs.c
+++ b/LibOS/shim/src/fs/shim_fs.c
@@ -30,6 +30,7 @@ struct shim_fs* builtin_fs[] = {
     &eventfd_builtin_fs,
     &pseudo_builtin_fs,
     &synthetic_builtin_fs,
+    &path_builtin_fs,
 };
 
 static struct shim_lock mount_mgr_lock;
--- a/LibOS/shim/src/fs/shim_fs_path.c
+++ b/LibOS/shim/src/fs/shim_fs_path.c
@@ -1,0 +1,25 @@
+/* SPDX-License-Identifier: LGPL-3.0-or-later */
+/* Copyright (C) 2022 Intel Corporation
+ *                    Borys Popławski <borysp@invisiblethingslab.com>
+ */
+
+#include "shim_fs.h"
+
+static int hstat(struct shim_handle* handle, struct stat* buf) {
+    if (!handle->inode) {
+        return -EINVAL;
+    }
+    return generic_inode_hstat(handle, buf);
+}
+
+static struct shim_fs_ops path_fs_ops = {
+    .hstat = hstat,
+};
+
+static struct shim_d_ops path_d_ops = {};
+
+struct shim_fs path_builtin_fs = {
+    .name = "path",
+    .fs_ops = &path_fs_ops,
+    .d_ops = &path_d_ops,
+};
--- a/LibOS/shim/src/fs/shim_namei.c
+++ b/LibOS/shim/src/fs/shim_namei.c
@@ -445,13 +445,31 @@ int open_namei(struct shim_handle* hdl, struct shim_dentry* start, const char* p
 
     ret = path_lookupat(start, path, lookup_flags, &dent);
     if (ret < 0)
-        goto err;
+        goto out;
+
+    if (flags & O_PATH) {
+        if (!dent->inode) {
+            ret = -ENOENT;
+            goto out;
+        }
+
+        assoc_handle_with_dentry(hdl, dent, flags);
+        if (dent->inode->type == S_IFDIR) {
+            hdl->is_dir = true;
+            hdl->dir_info.dents = NULL;
+        }
+
+        hdl->type = TYPE_PATH;
+        hdl->fs = &path_builtin_fs;
+        ret = 0;
+        goto out;
+    }
 
     if (dent->inode && dent->inode->type == S_IFDIR) {
         if (flags & O_WRONLY || flags & O_RDWR ||
                 ((flags & O_CREAT) && !(flags & O_DIRECTORY) && !(flags & O_EXCL))) {
             ret = -EISDIR;
-            goto err;
+            goto out;
         }
     }
 
@@ -459,18 +477,16 @@ int open_namei(struct shim_handle* hdl, struct shim_dentry* start, const char* p
         /*
          * Can happen if user specified O_NOFOLLOW, or O_TRUNC | O_EXCL. Posix requires us to fail
          * with -ELOOP when trying to open a symlink.
-         *
-         * (Linux allows opening a symlink with O_PATH, but Gramine does not support it yet).
          */
         ret = -ELOOP;
-        goto err;
+        goto out;
     }
 
     bool need_open = true;
     if (!dent->inode) {
         if (!(flags & O_CREAT)) {
             ret = -ENOENT;
-            goto err;
+            goto out;
         }
 
         /* Check the parent permission first */
@@ -478,7 +494,7 @@ int open_namei(struct shim_handle* hdl, struct shim_dentry* start, const char* p
         if (dir) {
             ret = check_permissions(dir, MAY_WRITE | MAY_EXEC);
             if (ret < 0)
-                goto err;
+                goto out;
         }
 
         struct shim_fs* fs = dent->mount->fs;
@@ -487,19 +503,19 @@ int open_namei(struct shim_handle* hdl, struct shim_dentry* start, const char* p
         if (flags & O_DIRECTORY) {
             if (!fs->d_ops->mkdir) {
                 ret = -EINVAL;
-                goto err;
+                goto out;
             }
             ret = fs->d_ops->mkdir(dent, mode & ~S_IFMT);
             if (ret < 0)
-                goto err;
+                goto out;
         } else {
             if (!fs->d_ops->creat) {
                 ret = -EINVAL;
-                goto err;
+                goto out;
             }
             ret = fs->d_ops->creat(hdl, dent, flags, mode & ~S_IFMT);
             if (ret < 0)
-                goto err;
+                goto out;
             assoc_handle_with_dentry(hdl, dent, flags);
             need_open = false;
         }
@@ -507,36 +523,37 @@ int open_namei(struct shim_handle* hdl, struct shim_dentry* start, const char* p
         /* The file exists. This is not permitted if both O_CREAT and O_EXCL are set. */
         if ((flags & O_CREAT) && (flags & O_EXCL)) {
             ret = -EEXIST;
-            goto err;
+            goto out;
         }
 
         /* Check permissions. Note that we do it only if the file already exists: a newly created
          * file is allowed to have a mode that's incompatible with `acc_mode`. */
         ret = check_permissions(dent, acc_mode);
         if (ret < 0)
-            goto err;
+            goto out;
     }
 
     if (hdl && need_open) {
         ret = dentry_open(hdl, dent, flags);
         if (ret < 0)
-            goto err;
+            goto out;
     }
 
+    ret = 0;
+
+out:
     if (found) {
-        *found = dent;
-    } else {
-        put_dentry(dent);
+        if (ret == 0) {
+            assert(dent);
+            get_dentry(dent);
+            *found = dent;
+        } else {
+            *found = NULL;
+        }
     }
-    unlock(&g_dcache_lock);
-    return 0;
 
-err:
     if (dent)
         put_dentry(dent);
-
-    if (found)
-        *found = NULL;
 
     unlock(&g_dcache_lock);
     return ret;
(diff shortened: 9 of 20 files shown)


boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/650)* status: 0 of 20 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

___
*[`LibOS/shim/include/shim_fs.h` line 129 at r1](https://reviewable.io/reviews/gramineproject/gramine/650#-N4UWwp27fCBI9v6qbFO:-N4UWwp27fCBI9v6qbFP:bkhf9hz) ([raw file](https://github.com/gramineproject/gramine/blob/aa4486b3357b943e8f131e7473ffb098a46b9195/LibOS/shim/include/shim_fs.h#L129)):*
> ```objc
>     file_off_t (*seek)(struct shim_handle* hdl, file_off_t offset, int whence);
> 
>     /* move, copy: rename or duplicate the file */
> ```

These 3 removals are kinda unrelated, but the change is small enough I decided not to create another PR. I could create another commit though.
___
*[`LibOS/shim/src/sys/shim_open.c` line 113 at r1](https://reviewable.io/reviews/gramineproject/gramine/650#-N4UXIa832SwZQuuHPoN:-N4UXIa9ElhhB7LDa6_y:b-flu3em) ([raw file](https://github.com/gramineproject/gramine/blob/c6dfd5db6aef114d147e9c7a38d9dd9d7e421c5b/LibOS/shim/src/sys/shim_open.c#L113)):*
> ```C
>              | O_EXCL | O_LARGEFILE | O_NOATIME | O_NOCTTY | O_NOFOLLOW | O_NONBLOCK | O_PATH
>              | O_SYNC | O_TMPFILE | O_TRUNC;
> ```

We might want to explicitly fail on unsupported (but valid) flags. Do we? If yes, which ones?

dimakuv (changes requested):
> Reviewed 20 of 20 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/650)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski)

___
*[`LibOS/shim/include/shim_fs.h` line 129 at r1](https://reviewable.io/reviews/gramineproject/gramine/650#-N4UWwp27fCBI9v6qbFO:-N4fiQ00CAHW_Yj5UjQq:b8pq8g9) ([raw file](https://github.com/gramineproject/gramine/blob/aa4486b3357b943e8f131e7473ffb098a46b9195/LibOS/shim/include/shim_fs.h#L129)):*

These 3 removals are kinda unrelated, but the change is small enough I decided not to create another PR. I could create another commit though.
</blockquote></details>

So `move` and `copy` functions were never used at all? And `hdrop` was exactly the same as `close`?
___
*[`LibOS/shim/include/shim_types.h` line 248 at r1](https://reviewable.io/reviews/gramineproject/gramine/650#-N4finu2Ec0iqaSP8LPh:-N4finu2Ec0iqaSP8LPi:bkiez82) ([raw file](https://github.com/gramineproject/gramine/blob/c6dfd5db6aef114d147e9c7a38d9dd9d7e421c5b/LibOS/shim/include/shim_types.h#L248)):*
> ```objc
> #ifndef O_ASYNC
> #define O_ASYNC 020000
> #endif
> ```

Why did you add it here? It was introduced in some newer Linux, so you just want to be sure? But from what I google, this was always in Linux...
___
*[`LibOS/shim/src/fs/shim_namei.c` line 457 at r1](https://reviewable.io/reviews/gramineproject/gramine/650#-N4frzGL74q_D4LYU8ax:-N4frzGL74q_D4LYU8ay:bd9b5wc) ([raw file](https://github.com/gramineproject/gramine/blob/c6dfd5db6aef114d147e9c7a38d9dd9d7e421c5b/LibOS/shim/src/fs/shim_namei.c#L457)):*
> ```C
> 
>         assoc_handle_with_dentry(hdl, dent, flags);
>         if (dent->inode->type == S_IFDIR) {
> ```

What if it's a symlink (`S_IFLNK`), and the flags do not contain `O_NOFOLLOW`? Then we get the `hdl` which is a symlink. Is this what Linux does, is this how it should work?
___
*[`LibOS/shim/src/fs/shim_namei.c` line 463 at r1](https://reviewable.io/reviews/gramineproject/gramine/650#-N4fkGf2EnV2OVFbBLBw:-N4fkGf2EnV2OVFbBLBx:bvbpmhn) ([raw file](https://github.com/gramineproject/gramine/blob/aa4486b3357b943e8f131e7473ffb098a46b9195/LibOS/shim/src/fs/shim_namei.c#L463)):*
> ```C
>          * with -ELOOP when trying to open a symlink.
>          *
>          * (Linux allows opening a symlink with O_PATH, but Gramine does not support it yet).
> ```

Actually, what's the point of this comment? I mean, why is it exactly here? Why did you remove it? Do we support it now (the combination of symlinks + `O_PATH`)?

I see that the relevant man page snippet seems to be this:
```
       ELOOP  pathname was a symbolic link, and flags specified
              O_NOFOLLOW but not O_PATH.
```

But I can't map this text to your current implementation.

UPDATE: Ok, you handle `O_PATH` above so we can't ever get to this place if flags contained `O_PATH`. Resolving.
___
*[`LibOS/shim/src/fs/shim_namei.c` line 548 at r1](https://reviewable.io/reviews/gramineproject/gramine/650#-N4fkZFB4uOfkJiI5_eT:-N4fkZFB4uOfkJiI5_eU:bsaal1j) ([raw file](https://github.com/gramineproject/gramine/blob/c6dfd5db6aef114d147e9c7a38d9dd9d7e421c5b/LibOS/shim/src/fs/shim_namei.c#L548)):*
> ```C
>         if (ret == 0) {
>             assert(dent);
>             get_dentry(dent);
> ```

Was that a bug previously? I don't see that the old code got a reference to `dent`.

UPDATE: Ah, that's because you later do `put_dentry(dent)`. Got it.
___
*[`LibOS/shim/src/fs/shim_namei.c` line 747 at r1](https://reviewable.io/reviews/gramineproject/gramine/650#-N4flI-EEL_pmbTvCFk4:-N4flI-EEL_pmbTvCFk5:b6if93a) ([raw file](https://github.com/gramineproject/gramine/blob/c6dfd5db6aef114d147e9c7a38d9dd9d7e421c5b/LibOS/shim/src/fs/shim_namei.c#L747)):*
> ```C
>     }
> 
>     if (!hdl->is_dir && hdl->type != TYPE_PATH) {
> ```

The logic "is this handle a directory" got complicated now... Now we have two "directory" handles:
- old `hdl->type == TYPE_CHROOT` / `TYPE_TMPFS` / etc. + `hdl->is_dir`, and
- new `hdl->type == TYPE_PATH`

Is this maintainable in the long run? Why do you only add this check in this one place, and not any other places? Is it because only this function is used for functionality enabled by `O_PATH`?

UPDATE: I see why you added it here. Because the man page directly talks about this case:
```
    openat()
       ...

       The dirfd argument is used in conjunction with the pathname
       argument as follows:

       *  If the pathname given in pathname is relative, then it is
          interpreted relative to the directory referred to by the file
          descriptor dirfd (rather than relative to the current working
          directory of the calling process, as is done by open() for a
          relative pathname).  In this case, dirfd must be a directory
          that was opened for reading (O_RDONLY) or using the O_PATH
          flag.
```
___
*[`LibOS/shim/src/fs/pipe/fs.c` line 134 at r1](https://reviewable.io/reviews/gramineproject/gramine/650#-N4fluM9Dk6gw-_PnKvz:-N4fluM9Dk6gw-_PnKw-:b-g6u7ns) ([raw file](https://github.com/gramineproject/gramine/blob/c6dfd5db6aef114d147e9c7a38d9dd9d7e421c5b/LibOS/shim/src/fs/pipe/fs.c#L134)):*
> ```C
> 
> static int pipe_hstat(struct shim_handle* hdl, struct stat* stat) {
>     /* XXX: DOES THIS WORK LOL
> ```

Can we revert this? Having a whole diff on a file just for the sake of a more visible comment is not nice.
___
*[`LibOS/shim/src/sys/shim_open.c` line 113 at r1](https://reviewable.io/reviews/gramineproject/gramine/650#-N4UXIa832SwZQuuHPoN:-N4fmhRk1hoZ8w1Rrg8P:b-1gwir4) ([raw file](https://github.com/gramineproject/gramine/blob/c6dfd5db6aef114d147e9c7a38d9dd9d7e421c5b/LibOS/shim/src/sys/shim_open.c#L113)):*

We might want to explicitly fail on unsupported (but valid) flags. Do we? If yes, which ones?
</blockquote></details>

This is orthogonal to this PR. Ideally we'd analyze all the available flags and their consequences. I think some flags we may silently ignore (they have no side effects for our Gramine purposes), and on some we should loudly fail. But I don't think we need to decide now. So maybe mark as a TODO or add a GitHub issue.
___
*[`LibOS/shim/test/ltp/ltp.cfg` line 268 at r1](https://reviewable.io/reviews/gramineproject/gramine/650#-N4fi9BW3uPXi3YzQTNO:-N4fi9BX-zQT1ms3dnrE:b-in4u4r) ([raw file](https://github.com/gramineproject/gramine/blob/c6dfd5db6aef114d147e9c7a38d9dd9d7e421c5b/LibOS/shim/test/ltp/ltp.cfg#L268)):*
> ```INI
> skip = yes
> 
> # execveat not implemented
> ```

Are these changes something completely unrelated, but you stashed them here? I don't mind, just want to understand the rationale.

I guess you were looking through this file in an effort to find `O_PATH` tests, and accidentally saw these wrong comments?

Also, it looks like you didn't find any `O_PATH` test in LTP? Weird.
___
*[`LibOS/shim/test/regression/common.h` line 9 at r1](https://reviewable.io/reviews/gramineproject/gramine/650#-N4fouSe2zwLQP237WsT:-N4fouSe2zwLQP237WsU:b-omx8kp) ([raw file](https://github.com/gramineproject/gramine/blob/c6dfd5db6aef114d147e9c7a38d9dd9d7e421c5b/LibOS/shim/test/regression/common.h#L9)):*
> ```objc
> #define COMMON_H_
> 
> #include "err.h"
> ```

Why not `<>`?
___
*[`LibOS/shim/test/regression/epoll_test.c` line 15 at r1](https://reviewable.io/reviews/gramineproject/gramine/650#-N4fp0td39QWgPK6ftJJ:-N4fp0td39QWgPK6ftJK:b5gzfch) ([raw file](https://github.com/gramineproject/gramine/blob/c6dfd5db6aef114d147e9c7a38d9dd9d7e421c5b/LibOS/shim/test/regression/epoll_test.c#L15)):*
> ```C
>     errx(1, "%d: " msg, __LINE__, ##args)
> 
> #define ARRAY_LEN(arr) (sizeof(arr) / sizeof(arr[0]))
> ```

It would make sense to move these two macros in `common.h` as well.
___
*[`LibOS/shim/test/regression/open_opath.c` line 50 at r1](https://reviewable.io/reviews/gramineproject/gramine/650#-N4fqdU_1DzAeoMg5b9K:-N4fqdU_1DzAeoMg5b9L:b-xyfbmx) ([raw file](https://github.com/gramineproject/gramine/blob/c6dfd5db6aef114d147e9c7a38d9dd9d7e421c5b/LibOS/shim/test/regression/open_opath.c#L50)):*
> ```C
> 
>     CHECK(unlinkat(mnt_dir_fd, "../tmp/B", 0));
> ```

Can we also add an execve test as described in the man page:
```
                  char buf[PATH_MAX];
                  fd = open("some_prog", O_PATH);
                  snprintf(buf, PATH_MAX, "/proc/self/fd/%d", fd);
                  execl(buf, "some_prog", (char *) NULL);
```

Though I'm not sure that Gramine already implements `/proc/self/fd/` functionality? If not, then ignore this.

But I guess this would be important to test because this is one of the most common usages of `O_PATH`.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/650)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

___
*[`LibOS/shim/include/shim_fs.h` line 129 at r1](https://reviewable.io/reviews/gramineproject/gramine/650#-N4UWwp27fCBI9v6qbFO:-N4hS3fz1NTV-GAEnb2Z:btqqu58) ([raw file](https://github.com/gramineproject/gramine/blob/aa4486b3357b943e8f131e7473ffb098a46b9195/LibOS/shim/include/shim_fs.h#L129)):*

So `move` and `copy` functions were never used at all? And `hdrop` was exactly the same as `close`?
</blockquote></details>

Correct!
___
*[`LibOS/shim/include/shim_types.h` line 248 at r1](https://reviewable.io/reviews/gramineproject/gramine/650#-N4finu2Ec0iqaSP8LPh:-N4hS8l5DLWJprLG3sUZ:bianb4s) ([raw file](https://github.com/gramineproject/gramine/blob/c6dfd5db6aef114d147e9c7a38d9dd9d7e421c5b/LibOS/shim/include/shim_types.h#L248)):*

Why did you add it here? It was introduced in some newer Linux, so you just want to be sure? But from what I google, this was always in Linux...
</blockquote></details>

No, but it's defined in a header which we cannot include (`bits/fcntl.h`, which can be only included by `fcntl.h`, but we cannot use the latter).
Fun fact (checked just now): it's not defined by the kernel at all - which means it's not used as well.
Which actually means that we should not have it as well. Updated.
___
*[`LibOS/shim/src/fs/shim_namei.c` line 457 at r1](https://reviewable.io/reviews/gramineproject/gramine/650#-N4frzGL74q_D4LYU8ax:-N4hUedw9fjOC5jvmxpp:b5ot293) ([raw file](https://github.com/gramineproject/gramine/blob/c6dfd5db6aef114d147e9c7a38d9dd9d7e421c5b/LibOS/shim/src/fs/shim_namei.c#L457)):*

What if it's a symlink (`S_IFLNK`), and the flags do not contain `O_NOFOLLOW`? Then we get the `hdl` which is a symlink. Is this what Linux does, is this how it should work?
</blockquote></details>

`O_PATH` is used to get a `fd` to a "item" in the filesystem, whatever that is. So yes.
___
*[`LibOS/shim/src/fs/shim_namei.c` line 548 at r1](https://reviewable.io/reviews/gramineproject/gramine/650#-N4fkZFB4uOfkJiI5_eT:-N4hV-oB7N9WB1epxWAy:b-wfpmoo) ([raw file](https://github.com/gramineproject/gramine/blob/c6dfd5db6aef114d147e9c7a38d9dd9d7e421c5b/LibOS/shim/src/fs/shim_namei.c#L548)):*

Was that a bug previously? I don't see that the old code got a reference to `dent`.

UPDATE: Ah, that's because you later do `put_dentry(dent)`. Got it.
</blockquote></details>

Yes, I just didn't want to duplicate the cleanup code and I needed anther exit path for opath handling, so just unified it.
___
*[`LibOS/shim/src/fs/shim_namei.c` line 747 at r1](https://reviewable.io/reviews/gramineproject/gramine/650#-N4flI-EEL_pmbTvCFk4:-N4hVOTMExM3KF14l_zK:bio2ksi) ([raw file](https://github.com/gramineproject/gramine/blob/c6dfd5db6aef114d147e9c7a38d9dd9d7e421c5b/LibOS/shim/src/fs/shim_namei.c#L747)):*
It's all a bit complicated in Gramine right now. We have a `shim_handle` which has both `dentry` and `inode`, but the latter is not to be used by the common code (at least that's what comment in the header says). I'll try to remove `dentry` and see how it goes (imo it should not be there).

> Is this maintainable in the long run? 

No idea. I tried removing `is_dir` completely, but it's impossible currently.


But now I think this change is wrong - we set `is_dir` for path handles as well. I've removed it, it all still works, with small quire we return different error code in one case (`openat(fd_to_file_not_dir, "", O_RDONLY)`), but that should not be a problem.
___
*[`LibOS/shim/src/fs/pipe/fs.c` line 134 at r1](https://reviewable.io/reviews/gramineproject/gramine/650#-N4fluM9Dk6gw-_PnKvz:-N4hbo4w0yzO8p5xTspk:b-9dbw8q) ([raw file](https://github.com/gramineproject/gramine/blob/c6dfd5db6aef114d147e9c7a38d9dd9d7e421c5b/LibOS/shim/src/fs/pipe/fs.c#L134)):*
Sorry, couldn't stop myself from doing this :)
> more visible comment 

this was hardly the reason
___
*[`LibOS/shim/src/sys/shim_open.c` line 113 at r1](https://reviewable.io/reviews/gramineproject/gramine/650#-N4UXIa832SwZQuuHPoN:-N4hdCVYB0yrrKRt__ZT:baw5np6) ([raw file](https://github.com/gramineproject/gramine/blob/c6dfd5db6aef114d147e9c7a38d9dd9d7e421c5b/LibOS/shim/src/sys/shim_open.c#L113)):*

This is orthogonal to this PR. Ideally we'd analyze all the available flags and their consequences. I think some flags we may silently ignore (they have no side effects for our Gramine purposes), and on some we should loudly fail. But I don't think we need to decide now. So maybe mark as a TODO or add a GitHub issue.
</blockquote></details>

Ok, added a TODO
___
*[`LibOS/shim/test/ltp/ltp.cfg` line 268 at r1](https://reviewable.io/reviews/gramineproject/gramine/650#-N4fi9BW3uPXi3YzQTNO:-N4hdHv85kq3_uNC8op-:b4i3c0l) ([raw file](https://github.com/gramineproject/gramine/blob/c6dfd5db6aef114d147e9c7a38d9dd9d7e421c5b/LibOS/shim/test/ltp/ltp.cfg#L268)):*

Are these changes something completely unrelated, but you stashed them here? I don't mind, just want to understand the rationale.

I guess you were looking through this file in an effort to find `O_PATH` tests, and accidentally saw these wrong comments?

Also, it looks like you didn't find any `O_PATH` test in LTP? Weird.
</blockquote></details>

All of these (with changed comments) tests use `O_PATH` flag, but fail for unrelated reasons.
___
*[`LibOS/shim/test/regression/common.h` line 9 at r1](https://reviewable.io/reviews/gramineproject/gramine/650#-N4fouSe2zwLQP237WsT:-N4hdRuI6-0-0HHR9qj8:b-i8ctrt) ([raw file](https://github.com/gramineproject/gramine/blob/c6dfd5db6aef114d147e9c7a38d9dd9d7e421c5b/LibOS/shim/test/regression/common.h#L9)):*

Why not `<>`?
</blockquote></details>

Erm, no idea
___
*[`LibOS/shim/test/regression/epoll_test.c` line 15 at r1](https://reviewable.io/reviews/gramineproject/gramine/650#-N4fp0td39QWgPK6ftJJ:-N4hdfwXDIVL9uyD54sV:bk2amoz) ([raw file](https://github.com/gramineproject/gramine/blob/c6dfd5db6aef114d147e9c7a38d9dd9d7e421c5b/LibOS/shim/test/regression/epoll_test.c#L15)):*

It would make sense to move these two macros in `common.h` as well.
</blockquote></details>

`ERR` doesn't seem that useful, but `ARRAY_LEN` is, so moved it.
___
*[`LibOS/shim/test/regression/open_opath.c` line 50 at r1](https://reviewable.io/reviews/gramineproject/gramine/650#-N4fqdU_1DzAeoMg5b9K:-N4hgwTl8l0VN9QnrhrF:b-cpx24m) ([raw file](https://github.com/gramineproject/gramine/blob/c6dfd5db6aef114d147e9c7a38d9dd9d7e421c5b/LibOS/shim/test/regression/open_opath.c#L50)):*
> this is one of the most common usages of O_PATH.

Is it? This doesn't sound useful at all and I've never seen such pattern. Why do you think it is?

But sure, I can add this.

dimakuv (approved):
> Reviewed 9 of 9 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/650)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`LibOS/shim/src/fs/shim_namei.c` line 747 at r1](https://reviewable.io/reviews/gramineproject/gramine/650#-N4flI-EEL_pmbTvCFk4:-N4l3Xyi1I2zSjwCdQK5:b7niv0m) ([raw file](https://github.com/gramineproject/gramine/blob/c6dfd5db6aef114d147e9c7a38d9dd9d7e421c5b/LibOS/shim/src/fs/shim_namei.c#L747)):*
> I'll try to remove dentry and see how it goes (imo it should not be there).

I didn't get this part. You mean you want to remove `shim_handle::dentry` field? But why?

My question is unrelated to this PR, so I unblocked.
___
*[`LibOS/shim/test/regression/open_opath.c` line 50 at r1](https://reviewable.io/reviews/gramineproject/gramine/650#-N4fqdU_1DzAeoMg5b9K:-N4l43g33uVg2OFif9qb:bo6bvfs) ([raw file](https://github.com/gramineproject/gramine/blob/c6dfd5db6aef114d147e9c7a38d9dd9d7e421c5b/LibOS/shim/test/regression/open_opath.c#L50)):*

> this is one of the most common usages of O_PATH.

Is it? This doesn't sound useful at all and I've never seen such pattern. Why do you think it is?

But sure, I can add this.
</blockquote></details>

The man page gives this example, and my googliing also gave this example (probably just a copy-paste from the man page for illustrative purposes). So I thought it's one of the primary uses. Whatever.
___
*[`LibOS/shim/src/fs/pipe/fs.c` line 134 at r1](https://reviewable.io/reviews/gramineproject/gramine/650#-N4fluM9Dk6gw-_PnKvz:-N4l4ZP6-7o_C5-p0pia:b-uetvek) ([raw file](https://github.com/gramineproject/gramine/blob/c6dfd5db6aef114d147e9c7a38d9dd9d7e421c5b/LibOS/shim/src/fs/pipe/fs.c#L134)):*
> this was hardly the reason

I know :)

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/650)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`LibOS/shim/src/fs/shim_namei.c` line 747 at r1](https://reviewable.io/reviews/gramineproject/gramine/650#-N4flI-EEL_pmbTvCFk4:-N4lWEo10EPWlNmVFwXT:bof7ogp) ([raw file](https://github.com/gramineproject/gramine/blob/c6dfd5db6aef114d147e9c7a38d9dd9d7e421c5b/LibOS/shim/src/fs/shim_namei.c#L747)):*

> I'll try to remove dentry and see how it goes (imo it should not be there).

I didn't get this part. You mean you want to remove `shim_handle::dentry` field? But why?

My question is unrelated to this PR, so I unblocked.
</blockquote></details>

`dentry` is (as the name says) `directory entry` - an entry in file system layout. You open a handle to the actual "item" (which is generally called `inode`), not specific place in filesystem. Handle does not need to know about specific fs location of the file/`inode` (actually it can change).

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/650)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

___
*[`LibOS/shim/src/fs/shim_namei.c` line 747 at r1](https://reviewable.io/reviews/gramineproject/gramine/650#-N4flI-EEL_pmbTvCFk4:-N5-30ro4XO4jirca_2j:bv1f8fy) ([raw file](https://github.com/gramineproject/gramine/blob/c6dfd5db6aef114d147e9c7a38d9dd9d7e421c5b/LibOS/shim/src/fs/shim_namei.c#L747)):*

`dentry` is (as the name says) `directory entry` - an entry in file system layout. You open a handle to the actual "item" (which is generally called `inode`), not specific place in filesystem. Handle does not need to know about specific fs location of the file/`inode` (actually it can change).
</blockquote></details>

Yes, you're right. We even have a comment about this:

https://github.com/gramineproject/gramine/blob/97d3f69d89af6f1507942f57c2927f95e76d3fa9/LibOS/shim/include/shim_handle.h#L154

Unfortunately, this was not finished by Pawel.

dimakuv (approved):
> Reviewed 9 of 9 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/650)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #641, opened 2022-06-08
Issue in creating Protected file


sivashax:
## Description of the problem
With Gramine 1.2 TAG seeing issue with Protected file creation. When the protected file / plain txt file is already available, opening the file again in *write* mode is failing. FYI, this was working in previous version of Gramine.

## Steps to reproduce
Use attached sample program. Place it in gramine source directory and compile. Create "prot_dir" in /tmp for creating protected file.
Run as below.
gramine-sgx protect_file /tmp/prot_dir/sample.txt -> Protected file would be created.

Run the same command again.
sample.txt should be created again in /tmp/prot_dir. However, its failing in this release.

## Expected results

Able to open the protected file / plain text file if already available for creating as protected file.

## Actual results

File open is failing, incase if file is already available in protected file path.

## Gramine commit hash
commit 132ebc14fd339232a95538b920efa85e6119693c

sivashax:
[protect_file.tar.gz](https://github.com/gramineproject/gramine/files/8860542/protect_file.tar.gz)

dimakuv:
@mkow Could we assign someone to it?

oshogbo:
I can take a look.

oshogbo:
It seems that the issue is the file truncation functionality in protected files.
```
[P1:T1:protect_file] debug: ipf_open: handle: 0, path: '/tmp/prot_dir/sample.txt', real size: 4096, mode: 0x3
[P1:T1:protect_file] debug: ipf_init_existing_file: data size 37
[P1:T1:protect_file] debug: ipf_open: OK (data size 37)
[P1:T1:protect_file] warning: encrypted_file_set_size: pf_set_size failed: Functionality not implemented
```
The `fopen(..., "w")` from GLIBC uses `openat(2)` with `O_TRUNC` flag. This ends up calling the `pf_set_size` function with size equals to 0. 
```
// TODO: file truncation
pf_status_t pf_set_size(pf_context_t* pf, uint64_t size) {
```

I will add at least 0 size truncation ASAP.

oshogbo:
> FYI, this was working in previous version of Gramine.

[sivashax](https://github.com/sivashax) Could I ask you for a code that was working for you on 1.1? Probably you have a different manifest file. The example you sent works on 1.2 (and I was able to reproduce the bug) but doesn't work on 1.1 at all.

sivashax:
[...]

PR #648, opened 2022-06-13
Add testcase for protected file

anjalirai-intel:
Writing to already existing protected file fails
https://github.com/gramineproject/gramine/issues/641

Signed-off-by: Anjali Rai <anjali.rai@intel.com>

## How to test this PR? 

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/648)

PR #651, opened 2022-06-14
[LibOS] Add truncate to zero support

oshogbo:
## Description of the changes 

This adds a truncation to zero. In this case, we are issuing normal truncation and refreshing the internal buffer.
We also drop all pending data.
This should fix the issue: https://github.com/gramineproject/gramine/issues/641

## How to test this PR? 
New tests are added.


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/651)* status: all files reviewed, 18 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @oshogbo)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fPSXWEKiWhJ-xXb3a:-N4fPSXWEKiWhJ-xXb3b:bq1hirl) (no related file):*
Please also fix comments here:
- https://github.com/gramineproject/gramine/blob/425e845086d018103034b59849f04b2798302d73/LibOS/shim/src/fs/chroot/encrypted.c#L28
- https://github.com/gramineproject/gramine/blob/425e845086d018103034b59849f04b2798302d73/common/src/protected_files/README.rst#todo
- https://github.com/gramineproject/gramine/blob/425e845086d018103034b59849f04b2798302d73/common/src/protected_files/protected_files.h#L259

These are all occurences I could find, the rest you already fixed in this PR. But please check as well (I grepped for `truncate`, `truncation`, `set_size`).

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fbhiE57ooq6tbVJqh:-N4fbhiE57ooq6tbVJqi:b-612h63) (no related file):*
What do we do about https://github.com/gramineproject/gramine/pull/648, since it is a duplicate of your enabled tests in this PR?

I wouldn't mind actually a separate test for PFs in LibOS regression tests, that will operate on the file, then truncate it to zero, then continue operations. From what I understand, the current `truncate` tests under `test/fs/` do not have such a "complex" flow of operations, and I'm afraid this first revision of the PR doesn't really work in such complex cases.

___
___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fNUc789EBdNRnF2S9):*
It's not the LibOS component, and it doesn't mention Protected Files at all.

I suggest:
```
[common] Protected Files: Add support for `ftruncate(0)`

Generic support for `ftruncate(<smaller size>)` is hard to implement in Protected Files,
but support for the ftruncate(0)` corner case is easy and important: `fopen(..., "w")`
from GLIBC uses `openat(.., O_TRUNC)`. This ends up calling `pf_set_size(0)`.

Thus, this commit enables a common scenario of opening an existing PF file in write
mode. New FS tests are enabled to test it.
```
___
*[`common/src/protected_files/protected_files.c` line 1270 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4f_NF8657w5mISkRrK:-N4f_NF8657w5mISkRrL:b-c85xy5) ([raw file](https://github.com/gramineproject/gramine/blob/43c02b741a3a7670ad3485d99d9602c5c24c2e8a/common/src/protected_files/protected_files.c#L1270)):*
> ```C
>        path_len = strlen(pf->encrypted_part_plain.path);
> 
>        memset(&pf->file_metadata, 0, sizeof(pf->file_metadata));
> ```

This looks brutal, erasing the whole metadata part. Are you sure the file is "workable" after that? Do we have a test that does something like:
```
... write/read with protected file ...
ftruncate(pf, 0);
... write/read with protected file ...
close(pf);
pf = open(<pf path>, "r");
... read from pf and verify it contains what we've written to it
```

`pf->file_metadata` contains such important fields as `metadata_key_id`, and you just nullified it. So how can the PF continue read/write operations if there is no key ID associated with it anymore? Or maybe the reconstruction of `file_metadata` is kicked somewhere after calling this function?
 (https://github.com/gramineproject/gramine/blob/e780fa7c434655b6f36a57c06a12cc3076fc8a93/common/src/protected_files/protected_files_format.h#L36)
___
*[`common/src/protected_files/protected_files.c` line 1270 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fa2oVBSNbr2CqAYML:-N4fa2oWAHilCzJRg_eB:b-x8lrs1) ([raw file](https://github.com/gramineproject/gramine/blob/43c02b741a3a7670ad3485d99d9602c5c24c2e8a/common/src/protected_files/protected_files.c#L1270)):*
> ```C
>        path_len = strlen(pf->encrypted_part_plain.path);
> 
>        memset(&pf->file_metadata, 0, sizeof(pf->file_metadata));
> ```

Why do you use `memset` instead of `erase_memory`? I think you should also use the latter in PFs.
___
*[`common/src/protected_files/protected_files.c` line 1274 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4faCAq7sNePydRLmaI:-N4faCAq7sNePydRLmaJ:bw170v2) ([raw file](https://github.com/gramineproject/gramine/blob/43c02b741a3a7670ad3485d99d9602c5c24c2e8a/common/src/protected_files/protected_files.c#L1274)):*
> ```C
>        memcpy(path, pf->encrypted_part_plain.path, path_len);
>        erase_memory(&pf->encrypted_part_plain, sizeof(pf->encrypted_part_plain));
>        memcpy(pf->encrypted_part_plain.path, path, path_len);
> ```

Could you group the variables and operations on each part of the `pf` object? Right now it's messy: you start with extracting the `path` from `encrypted_part_plain`, then you switch to erasing `file_metadata`, and then continue with `encrypted_part_plain`.

It's hard to read.
___
*[`common/src/protected_files/protected_files.c` line 1274 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fb39V7TaLcj0AHuyj:-N4fb39V7TaLcj0AHuyk:b-yqo0ed) ([raw file](https://github.com/gramineproject/gramine/blob/43c02b741a3a7670ad3485d99d9602c5c24c2e8a/common/src/protected_files/protected_files.c#L1274)):*
> ```C
>        memcpy(path, pf->encrypted_part_plain.path, path_len);
>        erase_memory(&pf->encrypted_part_plain, sizeof(pf->encrypted_part_plain));
>        memcpy(pf->encrypted_part_plain.path, path, path_len);
> ```

Again, this looks brutal. You only keep the `encrypted_part_plain.path` value intact. But what about e.g. `mht_key`? https://github.com/gramineproject/gramine/blob/e780fa7c434655b6f36a57c06a12cc3076fc8a93/common/src/protected_files/protected_files_format.h#L50
___
*[`common/src/protected_files/protected_files.c` line 1291 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fbEjkAoYEXRtc0XE_:-N4fbEjkAoYEXRtc0XEa:b8pn42o) ([raw file](https://github.com/gramineproject/gramine/blob/43c02b741a3a7670ad3485d99d9602c5c24c2e8a/common/src/protected_files/protected_files.c#L1291)):*
> ```C
>                free(file_node);
>                lruc_remove_last(pf->cache);
>        }
> ```

Ok, these operations on `need_writing` ... and `cache` look reasonable.
___
*[`LibOS/shim/test/fs/test_enc.py` line 84 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fS16l-mNLLt7n9S6F:-N4fS16l-mNLLt7n9S6G:bjhwq9x) ([raw file](https://github.com/gramineproject/gramine/blob/aa4486b3357b943e8f131e7473ffb098a46b9195/LibOS/shim/test/fs/test_enc.py#L84)):*
> ```Python
>             self.assertTrue(filecmp.cmp(self.INPUT_FILES[i], dec_path, shallow=False))
> 
>     # overrides TC_00_FileSystem to change input dir (from plaintext to encrypted)
> ```

Why did you remove this comment? This comment is correct -- the first line here is indeed changed (from `INPUT_FILES` to `ENCRYPTED_FILES`). Please restore this comment.
___
*[`LibOS/shim/test/fs/test_enc.py` line 93 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fSN38DUr2NwnPjnmN:-N4fSN38DUr2NwnPjnmO:b7pwtue) ([raw file](https://github.com/gramineproject/gramine/blob/43c02b741a3a7670ad3485d99d9602c5c24c2e8a/LibOS/shim/test/fs/test_enc.py#L93)):*
> ```Python
>         self.assertTrue(os.path.isfile(output_path))
> 
>     # overrides TC_00_FileSystem to change input dir (from plaintext to encrypted)
> ```

This comment is wrong, it doesn't overwrite anything. Just remove this comment.
___
*[`LibOS/shim/test/fs/test_enc.py` line 98 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fSQlrEnvvavfYnXIR:-N4fSQlrEnvvavfYnXIS:b-p0wsgh) ([raw file](https://github.com/gramineproject/gramine/blob/43c02b741a3a7670ad3485d99d9602c5c24c2e8a/LibOS/shim/test/fs/test_enc.py#L98)):*
> ```Python
>         file_path = os.path.join(self.OUTPUT_DIR, 'test_101') # new file
>         stdout, stderr = self.run_binary(['open_flags', file_path])
>         self.verify_open_flags(stdout, stderr)
> ```

Actually, this whole test (`101`) is exactly the same as in the base class `TC_00_FileSystem`, so you don't need to overwrite it. Thus, we can simply remove this whole test from here, and then the base test will be used instead.

I think the only reason why this test was copy-pasted in here was to add that `expectedFailure` decorator. Now that you don't have it, you don't need to re-write this test at all.
___
*[`LibOS/shim/test/fs/test_enc.py` line 127 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fUfWPE18MzsD1ayhx:-N4fUfWPE18MzsD1ayhy:b-vjbp1o) ([raw file](https://github.com/gramineproject/gramine/blob/43c02b741a3a7670ad3485d99d9602c5c24c2e8a/LibOS/shim/test/fs/test_enc.py#L127)):*
> ```Python
> 
>     # overrides TC_00_FileSystem to decrypt output
>     def verify_size(self, file, size):
> ```

This function is not even used?
___
*[`LibOS/shim/test/fs/test_enc.py` line 135 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fUqfO5mOKgHEk24tT:-N4fUqfO5mOKgHEk24tU:byw0hdh) ([raw file](https://github.com/gramineproject/gramine/blob/43c02b741a3a7670ad3485d99d9602c5c24c2e8a/LibOS/shim/test/fs/test_enc.py#L135)):*
> ```Python
>         stdout, stderr = self.run_binary(['truncate', path_1, path_2, str(size_out)])
> 
>         stdout, stderr = self.run_binary(['truncate', path_1, path_2, str(size_out)])
> ```

You run the same thing *twice*?
___
*[`LibOS/shim/test/fs/test_enc.py` line 140 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fVdqP2WACJRfEmtVG:-N4fVdqP2WACJRfEmtVH:bnsx7az) ([raw file](https://github.com/gramineproject/gramine/blob/43c02b741a3a7670ad3485d99d9602c5c24c2e8a/LibOS/shim/test/fs/test_enc.py#L140)):*
> ```Python
>         self.assertIn('open(' + path_2 + ') output OK', stdout)
>         self.assertIn('ftruncate(' + path_2 + ') to ' + str(size_out) + ' OK', stdout)
>         self.assertIn('close(' + path_2 + ') output OK', stdout)
> ```

You don't call `self.verify_size(path_1, path_2)` here at all. What's the reason?
___
*[`LibOS/shim/test/fs/test_enc.py` line 142 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fThMcCsVPftviId6S:-N4fThMdEbjmR-Q_kcEB:b5yysml) ([raw file](https://github.com/gramineproject/gramine/blob/43c02b741a3a7670ad3485d99d9602c5c24c2e8a/LibOS/shim/test/fs/test_enc.py#L142)):*
> ```Python
>         self.assertIn('close(' + path_2 + ') output OK', stdout)
> 
>     # TODO: File truncation to arbitrary size.
> ```

You may also add a comment like this:
```
    # overrides TC_00_FileSystem to change input dir (from plaintext to encrypted) and
    # because file truncation from greater to small size is not yet implemented
```

But not blocking, as it is pretty obvious.
___
*[`LibOS/shim/test/fs/test_enc.py` line 156 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fUz1qAKhji4km388o:-N4fUz1qAKhji4km388p:bu4q5y1) ([raw file](https://github.com/gramineproject/gramine/blob/43c02b741a3a7670ad3485d99d9602c5c24c2e8a/LibOS/shim/test/fs/test_enc.py#L156)):*
> ```Python
>         self.verify_truncate_test(path_1, path_2, 1000)
>         self.verify_truncate_test(path_1, path_2, 0)
>         self.verify_truncate_test(path_1, path_2, 0)
> ```

I don't like that your logic for this test is completely different from the logic in `test_fs.py`. Could you use the same logic? There is an unnecessary divergence here, and ideally (when PFs support proper truncation), the only change will be `INPUT_FILES` -> `ENCRYPTED_FILES`.

In particular, the `test_fs.py` version does this:
```
    def test_140_file_truncate(self):
        self.do_truncate_test(0, 1)
        self.do_truncate_test(0, 16)
```

I.e., it gets an input file of size 0 and expands it to size 1, then gets an input file of size 0 and expands it to size 16, and so on. So each `do_truncate_test()` is stateless and doesn't depend on the previous runs, unlike your current solution.
___
*[`LibOS/shim/test/fs/test_fs.py` line 62 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fSAFOEDTH7svF3qEa:-N4fSAFOEDTH7svF3qEb:b-86rq5t) ([raw file](https://github.com/gramineproject/gramine/blob/43c02b741a3a7670ad3485d99d9602c5c24c2e8a/LibOS/shim/test/fs/test_fs.py#L62)):*
> ```Python
>         self.assertIn('fclose(' + path + ') ' + mode + ' 2 OK', stdout)
> 
>     # overrides TC_00_FileSystem to change input dir (from plaintext to encrypted)
> ```

Please revert this change; this comment doesn't belong to this file but to the `test_enc.py`. I guess you confused the two files when doing this change.
___
*[`LibOS/shim/test/fs/test_fs.py` line 297 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fT473-bzFA9d1y7RX:-N4fT473-bzFA9d1y7RY:bh25rwf) ([raw file](https://github.com/gramineproject/gramine/blob/43c02b741a3a7670ad3485d99d9602c5c24c2e8a/LibOS/shim/test/fs/test_fs.py#L297)):*
> ```Python
>         self.do_copy_test('copy_sendfile', 60)
> 
>     @expectedFailureIf(HAS_SGX)
> ```

Unrelated, but why do the following mmap tests fail on SGX? It would be good to examine why this happens, and a comment with explanations here.

oshogbo (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/651)* status: 0 of 3 files reviewed, 17 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "squash! " and "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @oshogbo)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fbhiE57ooq6tbVJqh:-N4gz4ij-ekNpb6Jp5_x:bn8oi3m) (no related file):*

What do we do about https://github.com/gramineproject/gramine/pull/648, since it is a duplicate of your enabled tests in this PR?

I wouldn't mind actually a separate test for PFs in LibOS regression tests, that will operate on the file, then truncate it to zero, then continue operations. From what I understand, the current `truncate` tests under `test/fs/` do not have such a "complex" flow of operations, and I'm afraid this first revision of the PR doesn't really work in such complex cases.
</blockquote></details>

Actually it has in the `open_flags.c` but maybe we want to have a separate one, let's think about it afterwords.

___
___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fNUc789EBdNRnF2S9):*

It's not the LibOS component, and it doesn't mention Protected Files at all.

I suggest:
```
[common] Protected Files: Add support for `ftruncate(0)`

Generic support for `ftruncate(<smaller size>)` is hard to implement in Protected Files,
but support for the ftruncate(0)` corner case is easy and important: `fopen(..., "w")`
from GLIBC uses `openat(.., O_TRUNC)`. This ends up calling `pf_set_size(0)`.

Thus, this commit enables a common scenario of opening an existing PF file in write
mode. New FS tests are enabled to test it.
```
</blockquote></details>

Done.
___
*[`common/src/protected_files/protected_files.c` line 1270 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4f_NF8657w5mISkRrK:-N4gmUdQ40umcuP6SuCx:b-itzymp) ([raw file](https://github.com/gramineproject/gramine/blob/43c02b741a3a7670ad3485d99d9602c5c24c2e8a/common/src/protected_files/protected_files.c#L1270)):*

This looks brutal, erasing the whole metadata part. Are you sure the file is "workable" after that? Do we have a test that does something like:
```
... write/read with protected file ...
ftruncate(pf, 0);
... write/read with protected file ...
close(pf);
pf = open(<pf path>, "r");
... read from pf and verify it contains what we've written to it
```

`pf->file_metadata` contains such important fields as `metadata_key_id`, and you just nullified it. So how can the PF continue read/write operations if there is no key ID associated with it anymore? Or maybe the reconstruction of `file_metadata` is kicked somewhere after calling this function?
 (https://github.com/gramineproject/gramine/blob/e780fa7c434655b6f36a57c06a12cc3076fc8a93/common/src/protected_files/protected_files_format.h#L36)
</blockquote></details>

Yes, this is done by "open flags" test, with O_TRUNC.
This is brutal. But my idea was to truncate the original file to 0 and reinitialize the internal status to the same as the `ipf_open` is leaving it when creating a new file.
___
*[`common/src/protected_files/protected_files.c` line 1270 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fa2oVBSNbr2CqAYML:-N4gzhMa1sMjtRYA5s2b:b-f4lgzu) ([raw file](https://github.com/gramineproject/gramine/blob/43c02b741a3a7670ad3485d99d9602c5c24c2e8a/common/src/protected_files/protected_files.c#L1270)):*
> erase_memory
I wasn't sure when exactly should we use erase_memory.
I will fix that.
___
*[`common/src/protected_files/protected_files.c` line 1274 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fb39V7TaLcj0AHuyj:-N4gmwECCxncBwVAc443:bydajcs) ([raw file](https://github.com/gramineproject/gramine/blob/43c02b741a3a7670ad3485d99d9602c5c24c2e8a/common/src/protected_files/protected_files.c#L1274)):*

Again, this looks brutal. You only keep the `encrypted_part_plain.path` value intact. But what about e.g. `mht_key`? https://github.com/gramineproject/gramine/blob/e780fa7c434655b6f36a57c06a12cc3076fc8a93/common/src/protected_files/protected_files_format.h#L50
</blockquote></details>

Answered in `metadata_key_id` thread.
___
*[`common/src/protected_files/protected_files.c` line 1274 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4faCAq7sNePydRLmaI:-N4h-XoJ1pL2KB28c5s4:b1u23q7) ([raw file](https://github.com/gramineproject/gramine/blob/43c02b741a3a7670ad3485d99d9602c5c24c2e8a/common/src/protected_files/protected_files.c#L1274)):*

Could you group the variables and operations on each part of the `pf` object? Right now it's messy: you start with extracting the `path` from `encrypted_part_plain`, then you switch to erasing `file_metadata`, and then continue with `encrypted_part_plain`.

It's hard to read.
</blockquote></details>

Sure, make sense.
Thanks.
___
*[`LibOS/shim/test/fs/test_enc.py` line 84 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fS16l-mNLLt7n9S6F:-N4h-dNq9FVbp_ZhS0j-:bsi8v3l) ([raw file](https://github.com/gramineproject/gramine/blob/aa4486b3357b943e8f131e7473ffb098a46b9195/LibOS/shim/test/fs/test_enc.py#L84)):*

Why did you remove this comment? This comment is correct -- the first line here is indeed changed (from `INPUT_FILES` to `ENCRYPTED_FILES`). Please restore this comment.
</blockquote></details>

Sorry, my bad.
___
*[`LibOS/shim/test/fs/test_enc.py` line 93 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fSN38DUr2NwnPjnmN:-N4h1k8rDUyHQHj10CoJ:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/43c02b741a3a7670ad3485d99d9602c5c24c2e8a/LibOS/shim/test/fs/test_enc.py#L93)):*

This comment is wrong, it doesn't overwrite anything. Just remove this comment.
</blockquote></details>

Done.
___
*[`LibOS/shim/test/fs/test_enc.py` line 98 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fSQlrEnvvavfYnXIR:-N4h1jBQ1bhhF4j5lcHz:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/43c02b741a3a7670ad3485d99d9602c5c24c2e8a/LibOS/shim/test/fs/test_enc.py#L98)):*

Actually, this whole test (`101`) is exactly the same as in the base class `TC_00_FileSystem`, so you don't need to overwrite it. Thus, we can simply remove this whole test from here, and then the base test will be used instead.

I think the only reason why this test was copy-pasted in here was to add that `expectedFailure` decorator. Now that you don't have it, you don't need to re-write this test at all.
</blockquote></details>

Done.



_Code quote:_
```Python
test_101_open_flags
```
___
*[`LibOS/shim/test/fs/test_enc.py` line 127 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fUfWPE18MzsD1ayhx:-N4h2zUWCONLuhl0Y6uW:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/43c02b741a3a7670ad3485d99d9602c5c24c2e8a/LibOS/shim/test/fs/test_enc.py#L127)):*

This function is not even used?
</blockquote></details>

Done.
___
*[`LibOS/shim/test/fs/test_enc.py` line 135 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fUqfO5mOKgHEk24tT:-N4h2CYW5bS-KYoWhGl3:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/43c02b741a3a7670ad3485d99d9602c5c24c2e8a/LibOS/shim/test/fs/test_enc.py#L135)):*

You run the same thing *twice*?
</blockquote></details>

Done.
___
*[`LibOS/shim/test/fs/test_enc.py` line 140 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fVdqP2WACJRfEmtVG:-N4h1r_J74GTqVNefTvH:b-ki9hso) ([raw file](https://github.com/gramineproject/gramine/blob/43c02b741a3a7670ad3485d99d9602c5c24c2e8a/LibOS/shim/test/fs/test_enc.py#L140)):*

You don't call `self.verify_size(path_1, path_2)` here at all. What's the reason?
</blockquote></details>

Because the size of file is not the same as size of the content, I wasn't sure what to do with it.
But you have pointed out to me the `verify_size` which make now sense.



_Code quote:_
```Python
verify_size
```
___
*[`LibOS/shim/test/fs/test_enc.py` line 156 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fUz1qAKhji4km388o:-N4h2Mi3BoEU9vq6LGnn:bytg0gn) ([raw file](https://github.com/gramineproject/gramine/blob/43c02b741a3a7670ad3485d99d9602c5c24c2e8a/LibOS/shim/test/fs/test_enc.py#L156)):*

I don't like that your logic for this test is completely different from the logic in `test_fs.py`. Could you use the same logic? There is an unnecessary divergence here, and ideally (when PFs support proper truncation), the only change will be `INPUT_FILES` -> `ENCRYPTED_FILES`.

In particular, the `test_fs.py` version does this:
```
    def test_140_file_truncate(self):
        self.do_truncate_test(0, 1)
        self.do_truncate_test(0, 16)
```

I.e., it gets an input file of size 0 and expands it to size 1, then gets an input file of size 0 and expands it to size 16, and so on. So each `do_truncate_test()` is stateless and doesn't depend on the previous runs, unlike your current solution.
</blockquote></details>

Yes, I wanted to have an option to have a dependable test because I wanted for example to check that we are still able to read file after the truncation.
___
*[`LibOS/shim/test/fs/test_fs.py` line 62 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fSAFOEDTH7svF3qEa:-N4h2ezi7-PO-2LgGudD:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/43c02b741a3a7670ad3485d99d9602c5c24c2e8a/LibOS/shim/test/fs/test_fs.py#L62)):*

Please revert this change; this comment doesn't belong to this file but to the `test_enc.py`. I guess you confused the two files when doing this change.
</blockquote></details>

Done.
___
*[`LibOS/shim/test/fs/test_fs.py` line 297 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fT473-bzFA9d1y7RX:-N4h2tw2D-4AX2ifIEJV:btn9tnr) ([raw file](https://github.com/gramineproject/gramine/blob/43c02b741a3a7670ad3485d99d9602c5c24c2e8a/LibOS/shim/test/fs/test_fs.py#L297)):*

Unrelated, but why do the following mmap tests fail on SGX? It would be good to examine why this happens, and a comment with explanations here.
</blockquote></details>

I guess why should create an issue about it?
I can work on this afterwords.



_Code quote:_
```Python
# overrides TC_00_FileSystem to change input dir (from plaintext to encrypted)
```

oshogbo (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/651)* status: 0 of 6 files reviewed, 17 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @oshogbo)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fPSXWEKiWhJ-xXb3a:-N4hDmmlEe_IGPE5eDix:b-896fix) (no related file):*

Please also fix comments here:
- https://github.com/gramineproject/gramine/blob/425e845086d018103034b59849f04b2798302d73/LibOS/shim/src/fs/chroot/encrypted.c#L28
- https://github.com/gramineproject/gramine/blob/425e845086d018103034b59849f04b2798302d73/common/src/protected_files/README.rst#todo
- https://github.com/gramineproject/gramine/blob/425e845086d018103034b59849f04b2798302d73/common/src/protected_files/protected_files.h#L259

These are all occurences I could find, the rest you already fixed in this PR. But please check as well (I grepped for `truncate`, `truncation`, `set_size`).
</blockquote></details>

Done.

___

dimakuv (changes requested):
> Reviewed 3 of 3 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/651)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @oshogbo)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/651#-N4l1JJ_CeoNHkzFya7l:-N4l1JJ_CeoNHkzFya7m:bu7v7ur) (no related file):*
When looking into this PR, I found a potential bug: https://github.com/gramineproject/gramine/blob/fa7b91304f555946956422c706efa4ca0ad847f2/common/src/protected_files/protected_files.c#L816

Here the code assumes that `pf->encrypted_part_plain.size >= pf->offset`. But this may be wrong, especially now that we have `ftruncate(0)` -- this function does *not* change the offset of the file but nullifies the size. So this line of code may have an integer overflow. We should add an explicit check for this (and maybe come up with a C test that exposes this bug).

Or maybe this case can never be hit... IIUC, the Protected File can only be opened for read or only for write. In the for-read case, `ftruncate(0)` is not allowed. In the for-write case, we never hit this line (because it is inside `ipf_read()`). Anyway, need to analyze a bit more and at least add an `assert` here.

@oshogbo Could you look into this? I think an assert would be enough, with a comment explaining why offset can never be greater than size in the current code.

___
___
*[`common/src/protected_files/protected_files.c` line 1270 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4f_NF8657w5mISkRrK:-N4kp6Rv8mUP2eq82p9n:b-ht5w9p) ([raw file](https://github.com/gramineproject/gramine/blob/43c02b741a3a7670ad3485d99d9602c5c24c2e8a/common/src/protected_files/protected_files.c#L1270)):*

Yes, this is done by "open flags" test, with O_TRUNC.
This is brutal. But my idea was to truncate the original file to 0 and reinitialize the internal status to the same as the `ipf_open` is leaving it when creating a new file.
</blockquote></details>

But `ipf_open` also calls `ipf_init_existing_file` for the existing file (the case I care about here!) or `ipf_init_new_file` for the non-existing file (the case which you modelled your PR on).

UPDATE: I did a quick GDB session, with the following change:
```diff
diff --git a/LibOS/shim/test/regression/rw_file.c b/LibOS/shim/test/regression/rw_file.c
@@ -115,6 +115,13 @@ static ssize_t stdio_file_rw(const char* path, char* buf, size_t count, bool do_
         return -1;
     }

+    /* for testing of Protected/Encrypted files (used in sealed_file.c test) */
+    if (do_write) {
+        fflush(f);
+        ftruncate(fileno(f), 0);
+        rewind(f);
+        stdio_fd_rw(f, buf, count, do_write);
+    }
+
     int close_ret = fclose(f);
     if (close_ret < 0) {
         int ret_errno = errno;
```

Here I wanted to test the case: a protected file is opened, it is written to (we make sure it's not cached in the user-space buffer by doing an explicit fflush), then truncated to zero, then written to again. Everything works.

I went a bit deeper into the current PF code (https://github.com/gramineproject/gramine/blob/fa7b91304f555946956422c706efa4ca0ad847f2/common/src/protected_files/protected_files.c), and I can confirm that all the metadata fields are updated on demand. The function to look at is `ipf_get_data_node()`, and from then check the logic of called funcs.
___
*[`LibOS/shim/test/fs/test_enc.py` line 156 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fUz1qAKhji4km388o:-N4kqO6rDyyZlevOXwvY:b32w0x) ([raw file](https://github.com/gramineproject/gramine/blob/43c02b741a3a7670ad3485d99d9602c5c24c2e8a/LibOS/shim/test/fs/test_enc.py#L156)):*

Yes, I wanted to have an option to have a dependable test because I wanted for example to check that we are still able to read file after the truncation.
</blockquote></details>

OK
___
*[`LibOS/shim/test/fs/test_fs.py` line 297 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fT473-bzFA9d1y7RX:-N4kqSa98gqluJjZwuS7:bdkuwr6) ([raw file](https://github.com/gramineproject/gramine/blob/43c02b741a3a7670ad3485d99d9602c5c24c2e8a/LibOS/shim/test/fs/test_fs.py#L297)):*

I guess why should create an issue about it?
I can work on this afterwords.
</blockquote></details>

Looks like you had a typo: `why` -> `we`.

Sure, let's create an issue.

oshogbo (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/651)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/651#-N4l1JJ_CeoNHkzFya7l:-N4rr3n62lBsJA23X3EQ:b791f8z) (no related file):*

When looking into this PR, I found a potential bug: https://github.com/gramineproject/gramine/blob/fa7b91304f555946956422c706efa4ca0ad847f2/common/src/protected_files/protected_files.c#L816

Here the code assumes that `pf->encrypted_part_plain.size >= pf->offset`. But this may be wrong, especially now that we have `ftruncate(0)` -- this function does *not* change the offset of the file but nullifies the size. So this line of code may have an integer overflow. We should add an explicit check for this (and maybe come up with a C test that exposes this bug).

Or maybe this case can never be hit... IIUC, the Protected File can only be opened for read or only for write. In the for-read case, `ftruncate(0)` is not allowed. In the for-write case, we never hit this line (because it is inside `ipf_read()`). Anyway, need to analyze a bit more and at least add an `assert` here.

@oshogbo Could you look into this? I think an assert would be enough, with a comment explaining why offset can never be greater than size in the current code.
</blockquote></details>

Sure, let me look into this outside this PR.

___
___
*[`common/src/protected_files/protected_files.c` line 1270 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4f_NF8657w5mISkRrK:-N4rrBwf975VBzfmvO4Q:b-wqn3fs) ([raw file](https://github.com/gramineproject/gramine/blob/43c02b741a3a7670ad3485d99d9602c5c24c2e8a/common/src/protected_files/protected_files.c#L1270)):*

But `ipf_open` also calls `ipf_init_existing_file` for the existing file (the case I care about here!) or `ipf_init_new_file` for the non-existing file (the case which you modelled your PR on).

UPDATE: I did a quick GDB session, with the following change:
```diff
diff --git a/LibOS/shim/test/regression/rw_file.c b/LibOS/shim/test/regression/rw_file.c
@@ -115,6 +115,13 @@ static ssize_t stdio_file_rw(const char* path, char* buf, size_t count, bool do_
         return -1;
     }

+    /* for testing of Protected/Encrypted files (used in sealed_file.c test) */
+    if (do_write) {
+        fflush(f);
+        ftruncate(fileno(f), 0);
+        rewind(f);
+        stdio_fd_rw(f, buf, count, do_write);
+    }
+
     int close_ret = fclose(f);
     if (close_ret < 0) {
         int ret_errno = errno;
```

Here I wanted to test the case: a protected file is opened, it is written to (we make sure it's not cached in the user-space buffer by doing an explicit fflush), then truncated to zero, then written to again. Everything works.

I went a bit deeper into the current PF code (https://github.com/gramineproject/gramine/blob/fa7b91304f555946956422c706efa4ca0ad847f2/common/src/protected_files/protected_files.c), and I can confirm that all the metadata fields are updated on demand. The function to look at is `ipf_get_data_node()`, and from then check the logic of called funcs.
</blockquote></details>

I'm glad we agree that this works fine.

My general idea was that if we just running real truncate(0) on this file. This almost similar to removing and creating new file. So we can treat this file as a newly created, so all metadata should be cleared as well, and we should follow the `ifp_open` path for a new file.
___
*[`LibOS/shim/test/fs/test_fs.py` line 297 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fT473-bzFA9d1y7RX:-N4rrdyh0b6nNWW99EMz:b-x0paxf) ([raw file](https://github.com/gramineproject/gramine/blob/43c02b741a3a7670ad3485d99d9602c5c24c2e8a/LibOS/shim/test/fs/test_fs.py#L297)):*

Looks like you had a typo: `why` -> `we`.

Sure, let's create an issue.
</blockquote></details>

Yes :)

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/651)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @oshogbo)

___
*[`LibOS/shim/test/fs/test_fs.py` line 297 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fT473-bzFA9d1y7RX:-N5-4Jkh1HSBQdx9KMN0:b-ypbv0k) ([raw file](https://github.com/gramineproject/gramine/blob/43c02b741a3a7670ad3485d99d9602c5c24c2e8a/LibOS/shim/test/fs/test_fs.py#L297)):*

Yes :)
</blockquote></details>

I don't think this was done (`let's create an issue`)?

oshogbo (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/651)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`LibOS/shim/test/fs/test_fs.py` line 297 at r1](https://reviewable.io/reviews/gramineproject/gramine/651#-N4fT473-bzFA9d1y7RX:-N5-cnpGE-KSIWpLcGDq:b-ww2boa) ([raw file](https://github.com/gramineproject/gramine/blob/43c02b741a3a7670ad3485d99d9602c5c24c2e8a/LibOS/shim/test/fs/test_fs.py#L297)):*

I don't think this was done (`let's create an issue`)?
</blockquote></details>

Done: https://github.com/gramineproject/gramine/issues/663

boryspoplawski:
Jenkins, test this please

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/651)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners

pwmarcz (dismissed):
> Reviewed 2 of 3 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/651)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @oshogbo)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/651#-N50XBlf-m8qaY0rldXR:-N50XBlf-m8qaY0rldXS:b-he1u9d) (no related file):*
Nice! This was a painful corner case, since even `open(O_TRUNC)` failed without it.

FYI: I don't see any problems with this code, but I don't know much about Protected Files internals.

___
___
*[`common/src/protected_files/protected_files.c` line 1285 at r3](https://reviewable.io/reviews/gramineproject/gramine/651#-N50WscZBsSl_hEr1_qX:-N50WscZBsSl_hEr1_qY:bhz55f2) ([raw file](https://github.com/gramineproject/gramine/blob/c1053f17f5b75d65c12a55ff13cc232916d82a5e/common/src/protected_files/protected_files.c#L1285)):*
> ```C
> 
>        while ((data = lruc_get_last(pf->cache)) != NULL) {
>                file_node_t* file_node = (file_node_t*)data;
> ```

Reindent to 4g spaces?

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/651)* status: all files reviewed, 3 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @mkow and @oshogbo)

___
*[`LibOS/shim/test/fs/test_enc.py` line 96 at r3](https://reviewable.io/reviews/gramineproject/gramine/651#-N55444M8YuVgycDGcla:-N555e2a3lnxcPX6oxhl:bqdqqmc) ([raw file](https://github.com/gramineproject/gramine/blob/aa4486b3357b943e8f131e7473ffb098a46b9195/LibOS/shim/test/fs/test_enc.py#L96)):*

Why have you deleted this tests' invocation? Sounds like it should work now?
</blockquote></details>

Because this test (`test_101_open_flags`) is exactly the same as the one in the base-class `test_fs.py`: https://github.com/gramineproject/gramine/blob/5faa9eb4f11592e379fdc4319a043def557f2422/LibOS/test/fs/test_fs.py#L82

So by deleting this test, we remove the overwrite, and this test will be taken from the base class. It **will** be executed.

Also see one of our previous (and resolved) discussions in this PR.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/651)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @oshogbo)

___
*[`LibOS/shim/test/fs/test_enc.py` line 96 at r3](https://reviewable.io/reviews/gramineproject/gramine/651#-N55444M8YuVgycDGcla:-N55BCxk2mfce7qGyMJp:b2s39jb) ([raw file](https://github.com/gramineproject/gramine/blob/aa4486b3357b943e8f131e7473ffb098a46b9195/LibOS/shim/test/fs/test_enc.py#L96)):*

Because this test (`test_101_open_flags`) is exactly the same as the one in the base-class `test_fs.py`: https://github.com/gramineproject/gramine/blob/5faa9eb4f11592e379fdc4319a043def557f2422/LibOS/test/fs/test_fs.py#L82

So by deleting this test, we remove the overwrite, and this test will be taken from the base class. It **will** be executed.

Also see one of our previous (and resolved) discussions in this PR.
</blockquote></details>

Oh, I see, didn't notice that these are just overrides for _some_ tests and everything is actually run.

mkow:
Jenkins, retest this please

dimakuv (approved):
> Reviewed 7 of 7 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/651)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

mkow (changes requested):
> Reviewed 4 of 7 files at r4, 6 of 6 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/651)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @oshogbo)

___
*[`common/src/protected_files/protected_files.c` line 1273 at r5](https://reviewable.io/reviews/gramineproject/gramine/651#-N5BIEZF3KFA2C_mpCZw:-N5BIEZGD256yh0xoAtr:bwbhs7r) ([raw file](https://github.com/gramineproject/gramine/blob/a996189510a0b1b864b1d4cdf3899342797286f1/common/src/protected_files/protected_files.c#L1273)):*
> ```C
>         memcpy(pf->encrypted_part_plain.path, path, path_len);
> 
>         erase_memory(&pf->file_metadata, sizeof(pf->file_metadata));
> ```

No need to erase this, it resides on the host disk in plaintext. Can be just a `memset`.
___
*[`libos/test/fs/test_enc.py` line 95 at r5](https://reviewable.io/reviews/gramineproject/gramine/651#-N5AiM5pBXf44mR4qHit:-N5AiM5qBfIABBLSMJhK:b-9bji0h) ([raw file](https://github.com/gramineproject/gramine/blob/3d788aee1f7c0a245c686e397742c24bd7f27ebc/libos/test/fs/test_enc.py#L95)):*
> ```Python
> 
>     # overrides TC_00_FileSystem to change input dir (from plaintext to encrypted)
>     # doesn't work because encrypted files do not support truncation (and the test opens an
> ```

How does this work? Why don't we need to override the input dir for `test_101_open_flags`, but we do for other tests?

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/651)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @oshogbo)

___
*[`common/src/protected_files/protected_files.c` line 1273 at r5](https://reviewable.io/reviews/gramineproject/gramine/651#-N5BIEZF3KFA2C_mpCZw:-N5CEISYEFasnL0XnhPf:b-3a88ok) ([raw file](https://github.com/gramineproject/gramine/blob/a996189510a0b1b864b1d4cdf3899342797286f1/common/src/protected_files/protected_files.c#L1273)):*

Hym, I think @dimakuv proposed using `erase_memory` on this one.
https://reviewable.io/reviews/gramineproject/gramine/651#-N4fa2oVBSNbr2CqAYML

If I understand correctly @dimakuv, we just want to be sure that the zeroing really happens and none of the flags will be set.


</blockquote></details>

Yeah, but this buffer is not secret. It's literally the first chunk from the disk cached :)
___
*[`libos/test/fs/test_enc.py` line 95 at r5](https://reviewable.io/reviews/gramineproject/gramine/651#-N5AiM5pBXf44mR4qHit:-N5CEhNKAoK4ddk45hJR:b28d8zg) ([raw file](https://github.com/gramineproject/gramine/blob/3d788aee1f7c0a245c686e397742c24bd7f27ebc/libos/test/fs/test_enc.py#L95)):*

We actually do override the `INPUT_DIR`/`OUTPUT_DIR`  in `setUpClass`. 
So it should override the one from the original test, right?
Or you meant something else?
</blockquote></details>

Ok, seems we do, so it looks like it works.
But this looks weird - why all the tests have that comment, if it's not actually where the dir is overwritten? I originally pasted this comment because I saw that you deleted the overriding method with that comment on it (`overrides TC_00_FileSystem to change input dir (from plaintext to encrypted)`), and it seemed like it should be missing after that removal? I.e. the comment says that the change happens in the overridden method, but now you say that it actually happens in `setUpClass`?

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/651)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv and @oshogbo)

___
*[`libos/test/fs/test_enc.py` line 95 at r5](https://reviewable.io/reviews/gramineproject/gramine/651#-N5AiM5pBXf44mR4qHit:-N5F6qXZBUfHY4nSGdZ5:bxa3gco) ([raw file](https://github.com/gramineproject/gramine/blob/3d788aee1f7c0a245c686e397742c24bd7f27ebc/libos/test/fs/test_enc.py#L95)):*

Wait, people, let's backtrack.

> We actually do override the `INPUT_DIR`/`OUTPUT_DIR` in setUpClass.

This is not true. We only override `OUTPUT_DIR` in setUpClass. We do *not* override `INPUT_DIR`.

> Why don't we need to override the input dir for `test_101_open_flags`, but we do for other tests?

Because if you look at this test (https://github.com/gramineproject/gramine/blob/master/libos/test/fs/test_fs.py#L82-L85), you'll notice that there is no `INPUT_DIR` involved at all. Only the `OUTPUT_DIR` is needed. If you look at the test file (https://github.com/gramineproject/gramine/blob/master/libos/test/fs/open_flags.c), you'll see that this test *creates* a new output file. So no input file is needed at all.

So if we combine the two facts -- `OUTPUT_DIR` is set up in setUpClass in this `test_enc.py` class, and `test_101_open_flags` does *not* use the input dir/files at all -- we see that there is no need to override this method (`def test_101_open_flags()`) from the base class at all.
</blockquote></details>

I see, thanks for the explanation! So, the comment on the overridden `test_101_open_flags` was wrong before?

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/651)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow and @oshogbo)

___
*[`libos/test/fs/test_enc.py` line 95 at r5](https://reviewable.io/reviews/gramineproject/gramine/651#-N5AiM5pBXf44mR4qHit:-N5FBU67EXdFd6wsczSb:bup2t3z) ([raw file](https://github.com/gramineproject/gramine/blob/3d788aee1f7c0a245c686e397742c24bd7f27ebc/libos/test/fs/test_enc.py#L95)):*

I see, thanks for the explanation! So, the comment on the overridden `test_101_open_flags` was wrong before?
</blockquote></details>

Yes, the comment was wrong before.

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/651)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/651)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (approved):
> Reviewed 4 of 6 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/651)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #666, opened 2022-06-20
[LibOS,Pal] Do not pass error codes to `DkProcessExit`

boryspoplawski:
## Description of the changes 
`DkProcessExit` returns the error code to the host, which might expect just {0, 1} values. Also in most cases there is an error message printed before it, explaining what exactly happened.


dimakuv (approved):
> Reviewed 12 of 12 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/666)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

PR #670, opened 2022-06-21
[Pal/Linux-SGX] Allow `CPUID(EAX=0x7, ECX=<any value>)`

### LINKED ISSUE #668: Cannot render with blender example on J4105 NUC
## Description of the problem

Blender example not working on NUC running with J4105 CPU. 
Helloworld example runs without problem.

## Steps to reproduce

Ubuntu 18.04 with 5.15 kernel. 
Run blender in CI-Example. 

## Expected results

Renders the frame.

## Actual results

Got error: 
```
-----------------------------------------------------------------------------------------------------------------------
Gramine detected the following insecure configurations:

  - loader.insecure__use_cmdline_argv = true   (forwarding command-line args from untrusted host to the app)
  - sgx.allowed_files = [ ... ]                (some files are passed through from untrusted host without verification)

Gramine will continue application execution, but this configuration must not be used in production!
-----------------------------------------------------------------------------------------------------------------------

Blender 2.82 (sub 7) (hash 77d23b0bd76f built 2020-02-12 17:14:50)
ALSA lib conf.c:4120:(snd_config_update_r) Cannot access file /usr/share/alsa/alsa.conf
ALSA lib pcm.c:2642:(snd_pcm_open_noupdate) Unknown PCM default
AL lib: (EE) ALCplaybackAlsa_open: Could not open playback device 'default': No such file or directory
WARN (bke.appdir): /home/sources/buildbot-x86_64-slave/linux_glibc217_x86_64_cmake/blender.git/source/blender/blenkernel/intern/appdir.c:976 where_is_temp: Could not generate a temp file name for '/tmp/blender_JFkCwC', falling back to '/tmp/'
could not get a list of mounted filesystems
found bundled python: /blender/2.82/python
WARN (bke.appdir): /home/sources/buildbot-x86_64-slave/linux_glibc217_x86_64_cmake/blender.git/source/blender/blenkernel/intern/appdir.c:976 where_is_temp: Could not generate a temp file name for '/tmp/blender_lLN50q', falling back to '/tmp/'
Read blend: ./data/scenes/simple_scene.blend
error: Unrecognized leaf/subleaf in CPUID (EAX=0x7, ECX=0x4ff8ebbf). Exiting...
```

## Gramine commit hash

Installed using repo in https://gramine.readthedocs.io/en/latest/quickstart.html
version: gramine/stable,now 1.2-1 amd64

dimakuv:
## Description of the changes 

Previously, `_DkCpuIdRetrieve()` was too restrictive on the leaf 0x7 (Structured Extended Feature Flags) -- it only allowed sub-leaves 0 and 1. In reality, Intel SDM states the following:

    If ECX contains an invalid sub-leaf index, EAX/EBX/ECX/EDX return 0.
    Sub-leaf index n is invalid if n exceeds the value that sub-leaf 0
    returns in EAX.

This could lead to Gramine erroneously failing with `Unrecognized leaf/subleaf in CPUID`.

Fixes #668.

## How to test this PR? 

Added a test case in the `cpuid` LibOS test.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/670)

### DIFF
--- a/libos/test/regression/cpuid.c
+++ b/libos/test/regression/cpuid.c
@@ -107,6 +107,16 @@ static void test_cpuid_leaf_reserved(void) {
     struct regs r;
     set_dummy_regs(&r);
 
+    cpuid(0x7, 0x2, &r); /* leaf 0x7 returns all-zeros on sub-leaves > 1 */
+    if (r.eax || r.ebx || r.ecx || r.edx)
+        abort();
+    set_dummy_regs(&r);
+
+    cpuid(0x7, 0xFFFF, &r); /* leaf 0x7 returns all-zeros on sub-leaves > 1 */
+    if (r.eax || r.ebx || r.ecx || r.edx)
+        abort();
+    set_dummy_regs(&r);
+
     cpuid(0x8, 0x0, &r); /* subleaf value doesn't matter */
     if (r.eax || r.ebx || r.ecx || r.edx)
         abort();
--- a/pal/src/host/linux-sgx/pal_linux.h
+++ b/pal/src/host/linux-sgx/pal_linux.h
@@ -107,6 +107,8 @@ bool is_tsc_usable(void);
 uint64_t get_tsc_hz(void);
 void init_tsc(void);
 
+int init_cpuid(void);
+
 int init_enclave(void);
 void init_untrusted_slab_mgr(void);
 
--- a/pal/src/host/linux-sgx/pal_main.c
+++ b/pal/src/host/linux-sgx/pal_main.c
@@ -511,6 +511,12 @@ noreturn void pal_linux_main(char* uptr_libpal_uri, size_t libpal_uri_len, char*
     init_tsc();
     (void)get_tsc(); /* must be after `ready_for_exceptions=1` since it may generate SIGILL */
 
+    ret = init_cpuid();
+    if (ret < 0) {
+        log_error("init_cpuid failed: %d", ret);
+        ocall_exit(1, /*is_exitgroup=*/true);
+    }
+
     /* initialize master key (used for pipes' encryption for all enclaves of an application); it
      * will be overwritten below in init_child_process() with inherited-from-parent master key if
      * this enclave is child */
--- a/pal/src/host/linux-sgx/pal_misc.c
+++ b/pal/src/host/linux-sgx/pal_misc.c
@@ -29,9 +29,10 @@
  */
 #define TSC_REFINE_INIT_TIMEOUT_USECS 50000
 
-#define CPU_VENDOR_LEAF     0x0
-#define AMX_TILE_INFO_LEAF  0x1D
-#define AMX_TMUL_INFO_LEAF  0x1E
+#define CPU_VENDOR_LEAF             0x0
+#define EXTENDED_FEATURE_FLAGS_LEAF 0x7
+#define AMX_TILE_INFO_LEAF          0x1D
+#define AMX_TMUL_INFO_LEAF          0x1E
 
 uint64_t g_tsc_hz = 0; /* TSC frequency for fast and accurate time ("invariant TSC" HW feature) */
 static uint64_t g_start_tsc = 0;
@@ -114,6 +115,8 @@ int _PalSystemTimeQuery(uint64_t* out_usec) {
     return 0;
 }
 
+static uint32_t g_extended_feature_flags_max_supported_sub_leaves = 0;
+
 #define CPUID_CACHE_SIZE 64 /* cache only 64 distinct CPUID entries; sufficient for most apps */
 static struct pal_cpuid {
     unsigned int leaf, subleaf;
@@ -191,6 +194,12 @@ static void sanitize_cpuid(uint32_t leaf, uint32_t subleaf, uint32_t values[4]) 
         values[CPUID_WORD_EBX] = 0x756e6547; /* 'Genu' */
         values[CPUID_WORD_EDX] = 0x49656e69; /* 'ineI' */
         values[CPUID_WORD_ECX] = 0x6c65746e; /* 'ntel' */
+    } else if (leaf == EXTENDED_FEATURE_FLAGS_LEAF) {
+        if (subleaf == 0x0) {
+            values[CPUID_WORD_EAX] = g_extended_feature_flags_max_supported_sub_leaves;
+            values[CPUID_WORD_EBX] |= 1U << 0; /* SGX-enabled CPUs always support FSGSBASE */
+            values[CPUID_WORD_EBX] |= 1U << 2; /* SGX-enabled CPUs always report the SGX bit */
+        }
     } else if (leaf == EXTENDED_STATE_LEAF) {
         switch (subleaf) {
             case X87:
@@ -409,10 +418,21 @@ int _PalCpuIdRetrieve(uint32_t leaf, uint32_t subleaf, uint32_t values[4]) {
      *       all-zeros on them (as if these leaves are reserved). It is unclear why this discrepancy
      *       exists, but we decided to emulate how actual CPUs behave. */
     if (leaf == 0x08 || leaf == 0x0C || leaf == 0x0E || leaf == 0x11 || leaf == 0x13) {
-        values[0] = 0;
-        values[1] = 0;
-        values[2] = 0;
-        values[3] = 0;
+        values[CPUID_WORD_EAX] = 0;
+        values[CPUID_WORD_EBX] = 0;
+        values[CPUID_WORD_ECX] = 0;
+        values[CPUID_WORD_EDX] = 0;
+        return 0;
+    }
+
+    /* leaf 0x7 (Structured Extended Feature Flags) must return all-zeros if the subleaf contains an
+     * invalid index (larger than max supported) */
+    if (leaf == EXTENDED_FEATURE_FLAGS_LEAF &&
+            subleaf > g_extended_feature_flags_max_supported_sub_leaves) {
+        values[CPUID_WORD_EAX] = 0;
+        values[CPUID_WORD_EBX] = 0;
+        values[CPUID_WORD_ECX] = 0;
+        values[CPUID_WORD_EDX] = 0;
         return 0;
     }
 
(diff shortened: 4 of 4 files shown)


boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/670)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`Pal/src/host/Linux-SGX/db_misc.c` line 197 at r1](https://reviewable.io/reviews/gramineproject/gramine/670#-N5A-wQSBcj7j7osPdbY:-N5A-wQSBcj7j7osPdbZ:b256vqy) ([raw file](https://github.com/gramineproject/gramine/blob/2e68f7a51341391d5f725382ab22bd4d3856291b/Pal/src/host/Linux-SGX/db_misc.c#L197)):*
> ```C
>     } else if (leaf == EXTENDED_FEATURE_FLAGS_LEAF) {
>         if (subleaf == 0x0) {
>             values[CPUID_WORD_EAX] = 1; /* report max input value for supported sub-leaves */
> ```

Shouldn't we also somehow sanitize rest of the regs? They seem not to contain any sizes or stuff like that (so it's not that bad), but one (albeit convoluted) issue comes to my mind: what if software uses `cpuid(7, 0)->ebx` to check e.g. for `AVX*` support? It could be required in manifest, enabled, enclave starts correctly, but malicious host spoofs this cpuid and user app now does not use it.
Or there is no overlap between xfrm + stuff that is checked on enclave creation and this cpuid leaf?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/670)* status: 1 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`Pal/src/host/Linux-SGX/db_misc.c` line 197 at r1](https://reviewable.io/reviews/gramineproject/gramine/670#-N5A-wQSBcj7j7osPdbY:-N5EgQqQ7FkGo47mAzSd:b31wibg) ([raw file](https://github.com/gramineproject/gramine/blob/2e68f7a51341391d5f725382ab22bd4d3856291b/Pal/src/host/Linux-SGX/db_misc.c#L197)):*

Shouldn't we also somehow sanitize rest of the regs? They seem not to contain any sizes or stuff like that (so it's not that bad), but one (albeit convoluted) issue comes to my mind: what if software uses `cpuid(7, 0)->ebx` to check e.g. for `AVX*` support? It could be required in manifest, enabled, enclave starts correctly, but malicious host spoofs this cpuid and user app now does not use it.
Or there is no overlap between xfrm + stuff that is checked on enclave creation and this cpuid leaf?
</blockquote></details>

I went through the CPUID 0x07 leaf description in the Intel SDM (Table 3-8), and could reliably sanitize only two bits.

Sanitizing `AVX*` support is a hell and looks like straight impossible:
1. We have the secure `XFRM` (XCR0) bits as our "source of truth". Recall that XCR0 is used for the XSAVE area data/offsets. XCR0 does *not* reflect exact instructions/features available on the CPU, it only reflects "how much memory is needed for feature X in the XSAVE area".
2. The bits in CPUID 0x07 are much more detailed -- they reflect the exact instructions/features available on the CPU.

For example, CPUID 0x07 leaf has `AVX2`. But XFRM does **not** have `AVX2`, but only `AVX` (this is because AVX2 feature does not expand the XSAVE area but only introduces new instructions). So based on `XFRM[AVX bit] = 1`, we can't reliably say whether AVX2 is available or not on the CPU.

Similarly, CPUID 0x07 leaf has multiple bits for `AVX512`: AVX512F, AVX512DQ, AVX512_IFMA, AVX512PF, AVX512ER, AVX512CD, AVX512BW, AVX512VL, AVX512_VBMI, AVX512_VBMI2, and so on. At the same time, XFRM has only three bits: opmask, ZMM_hi256, Hi16_ZMM. Now look at this table and tell me what maps to what and what is available on which CPU: https://en.wikipedia.org/wiki/Advanced_Vector_Extensions#CPUs_with_AVX-512

TLDR: I have zero confidence we can meaningfully sanitize CPUID bits based on XFRM bits.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/670)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`Pal/src/host/Linux-SGX/db_misc.c` line 197 at r1](https://reviewable.io/reviews/gramineproject/gramine/670#-N5A-wQSBcj7j7osPdbY:-N5JRb5n8LARQC-8Zdmx:b7zwjma) ([raw file](https://github.com/gramineproject/gramine/blob/2e68f7a51341391d5f725382ab22bd4d3856291b/Pal/src/host/Linux-SGX/db_misc.c#L197)):*

> is a hell

Indeed it is.
I guess it's fine in most cases, since they are just runtime configurations and they do not affect Gramine, only user app, which can detect and act accordingly.

My main problem here is that we claim to sanitize CPUID, while when I look at it we do it only for a very small subset of it. I didn't look at everything, but the first thing: `cpuid(1)->ebx[8:16]` contains cache line size and can definitely be sanitized. There are probably many more things that are not "innocent" and cannot be left unchecked and passed from untrusted host as is.
</blockquote></details>

Yes, I agree, we sanitize only a couple things here. There is *a lot of room for improvement*. But I definitely don't want to start this effort in this PR.
___
*[`Pal/src/host/Linux-SGX/db_misc.c` line 197 at r2](https://reviewable.io/reviews/gramineproject/gramine/670#-N5G57F27zg5uVdbJ3i1:-N5JRp8s0jnHh9w7-tXK:b-s26bzw) ([raw file](https://github.com/gramineproject/gramine/blob/35af94129860e8e9d86efee259172d1863da8370/Pal/src/host/Linux-SGX/db_misc.c#L197)):*

Is this even correct? My cpu returns `0` here.
</blockquote></details>

Hm, good point, older CPUs only have 1 sub-leaf. Should I then do something like this:
```
if (values[CPUID_WORD_EAX] > 1) {
    /* report max input value for supported sub-leaves known to Gramine */
    values[CPUID_WORD_EAX] = 1;
}
```

Alternatively, I can remove this sanitization of EAX completely... This will support future CPUs, but then the malicious OS can lie about the number of supported sub-leaves. Not sure it has any potential of breaking apps.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/670)* status: 1 of 4 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`Pal/src/host/Linux-SGX/db_misc.c` line 197 at r2](https://reviewable.io/reviews/gramineproject/gramine/670#-N5G57F27zg5uVdbJ3i1:-N5Yrh-b20B5ELFz500x:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/35af94129860e8e9d86efee259172d1863da8370/Pal/src/host/Linux-SGX/db_misc.c#L197)):*

IMO just cap at `1`,
</blockquote></details>

Done.
___
*[`Pal/src/host/Linux-SGX/db_misc.c` line 429 at r2](https://reviewable.io/reviews/gramineproject/gramine/670#-N5L294z6Ere-qHp6gca:-N5YxN6n1aBgxdr_JBuH:bu9892i) ([raw file](https://github.com/gramineproject/gramine/blob/35af94129860e8e9d86efee259172d1863da8370/Pal/src/host/Linux-SGX/db_misc.c#L429)):*

Also `1` is hardcoded here. This should be changed to the value acquired form `0` subleaf (but the app might not have queried it; do we support nested cpuid calling here?)
</blockquote></details>

Done. I introduced `init_cpuid` to collect such things at startup.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/670)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`Pal/src/host/Linux-SGX/db_misc.c` line 119 at r3](https://reviewable.io/reviews/gramineproject/gramine/670#-N5ZjPS10NRPB146VlRA:-N5ZjPS10NRPB146VlRB:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/fc9c6833bde1727acd8611df256c773683fa1158/Pal/src/host/Linux-SGX/db_misc.c#L119)):*
> ```C
> 
> uint32_t g_extended_feature_flags_max_supported_sub_leaves = 0;
> ```





_Suggestion:_
```C
static uint32_t
```

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/670)* status: 3 of 4 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`Pal/src/host/Linux-SGX/db_misc.c` line 119 at r3](https://reviewable.io/reviews/gramineproject/gramine/670#-N5ZjPS10NRPB146VlRA:-N5Zmp3o2Fbf0Mige1ma:bak290) ([raw file](https://github.com/gramineproject/gramine/blob/fc9c6833bde1727acd8611df256c773683fa1158/Pal/src/host/Linux-SGX/db_misc.c#L119)):*
> ```C
> 
> uint32_t g_extended_feature_flags_max_supported_sub_leaves = 0;
> ```

Done. My bad.

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/670)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

boryspoplawski (approved):
> Reviewed 8 of 8 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/670)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #677, opened 2022-06-22
[LibOS] Allow setting `O_NONBLOCK` on not connected UNIX sockets

boryspoplawski:
## Description of the changes 
And some code clean ups.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/677)

### DIFF
--- a/libos/include/shim_fs.h
+++ b/libos/include/shim_fs.h
@@ -157,8 +157,15 @@ struct shim_fs_ops {
     /* hstat: get status of the file; `st_ino` will be taken from dentry, if there's one */
     int (*hstat)(struct shim_handle* hdl, struct stat* buf);
 
-    /* setflags: set flags of the file */
-    int (*setflags)(struct shim_handle* hdl, int flags);
+    /*!
+     * \brief Set flags on the handle.
+     *
+     * \param handle  The handle to set flags on.
+     * \param flags   Flags to set.
+     * \param mask    Indicates which flags to change. Only bits set in \p mask will be changed
+     *                (to values taken from \p flags). Must be non zero.
+     */
+    int (*setflags)(struct shim_handle* handle, unsigned int flags, unsigned int mask);
 
     /* lock and unlock the file */
     int (*lock)(const char* trim_name);
--- a/libos/include/shim_handle.h
+++ b/libos/include/shim_handle.h
@@ -81,13 +81,15 @@ enum shim_sock_state {
  * `ops`, `domain`, `type` and `protocol` are read-only and do not need any locking.
  * Access to `peek` struct is protected by `recv_lock`. This lock also ensures proper ordering of
  * stream reads (see the comment in `do_recvmsg` in "libos/src/sys/shim_socket.c").
- * `pal_handle` should be accessed using atomic operations. It can be NULL. Once it's set, it cannot
- * change anymore.
+ * `pal_handle` should be accessed using atomic operations.
  * If you need to take both `recv_lock` and `lock`, take the former first.
  */
 struct shim_sock_handle {
     struct shim_lock lock;
     struct shim_sock_ops* ops;
+    /* `pal_handle` can be NULL. Once it's set, it cannot change anymore. All implementations must
+     * take into account all necessary settings when instantiating this field, e.g. `handle->flags`,
+     * of handle wrapping this struct. */
     PAL_HANDLE pal_handle;
     int domain;
     int type;
--- a/libos/src/fs/pipe/fs.c
+++ b/libos/src/fs/pipe/fs.c
@@ -158,35 +158,43 @@ static int pipe_hstat(struct shim_handle* hdl, struct stat* stat) {
     return 0;
 }
 
-static int pipe_setflags(struct shim_handle* hdl, int flags) {
-    if (!hdl->pal_handle)
+static int pipe_setflags(struct shim_handle* handle, unsigned int flags, unsigned int mask) {
+    assert(mask != 0);
+    assert((flags & ~mask) == 0);
+
+    /* TODO: what is this check? Why it has no locking? */
+    if (!handle->pal_handle)
         return 0;
 
+    if (!WITHIN_MASK(flags, O_NONBLOCK)) {
+        return -EINVAL;
+    }
+
+    lock(&handle->lock);
+
     PAL_STREAM_ATTR attr;
-
-    int ret = DkStreamAttributesQueryByHandle(hdl->pal_handle, &attr);
+    int ret = DkStreamAttributesQueryByHandle(handle->pal_handle, &attr);
     if (ret < 0) {
-        return pal_to_unix_errno(ret);
+        ret = pal_to_unix_errno(ret);
+        goto out;
     }
 
-    if (attr.nonblocking) {
-        if (flags & O_NONBLOCK)
-            return 0;
-
-        attr.nonblocking = false;
-    } else {
-        if (!(flags & O_NONBLOCK))
-            return 0;
-
-        attr.nonblocking = true;
+    bool nonblocking = flags & O_NONBLOCK;
+    if (attr.nonblocking != nonblocking) {
+        attr.nonblocking = nonblocking;
+        ret = DkStreamAttributesSetByHandle(handle->pal_handle, &attr);
+        if (ret < 0) {
+            ret = pal_to_unix_errno(ret);
+            goto out;
+        }
     }
 
-    ret = DkStreamAttributesSetByHandle(hdl->pal_handle, &attr);
-    if (ret < 0) {
-        return pal_to_unix_errno(ret);
-    }
+    handle->flags = (handle->flags & ~mask) | flags;
+    ret = 0;
 
-    return 0;
+out:
+    unlock(&handle->lock);
+    return ret;
 }
 
 static int fifo_open(struct shim_handle* hdl, struct shim_dentry* dent, int flags) {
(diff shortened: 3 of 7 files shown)


boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/677)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`-- commits` line 5 at r2](https://reviewable.io/reviews/gramineproject/gramine/677#-N5EYbLBBiwyQe5p0J01):*

Could you remind again why we talk only about UNIX sockets, even though the change is in a generic function `src/fs/socket/fs.c: set_flags()`?

Is it simply because `handle->pal_handle == NULL` can *only* happen on UNIX sockets?
</blockquote></details>

Yes, excatly. Maybe a comment is due, explaining that all implementations allowing `pal_handle` to be `NULL` should set the currently enabled flags when they instantiate `pal_handle`?
___
*[`libos/include/shim_fs.h` line 168 at r2](https://reviewable.io/reviews/gramineproject/gramine/677#-N5EX8kKAmK4juc0pj6J:-N5Fpl9s2142i1QCw9SM:bakd4q3) ([raw file](https://github.com/gramineproject/gramine/blob/fd0c88e23973f27ea72ca3e363943e4bcf36104f/libos/include/shim_fs.h#L168)):*

I don't like that now we have `handle` name here, but `hdl` everywhere else in this file (and associated files with callback implementations). Maybe we should revert to `hdl` for uniformity, and have a proper separate PR will do this name change?

UPDATE: I see that you already have callback implementations with `handle`. So we already have a non-uniformity. Resolving.
</blockquote></details>

I just dislike `hdl` so I use `handle` in the new code I write. I wouldn't create a PR just for this rename, it would touch half of the LibOS code...
___
*[`libos/src/net/unix.c` line 136 at r2](https://reviewable.io/reviews/gramineproject/gramine/677#-N5EZhUi0h32FHv9fCPA:-N5Fqjz08yD0GyUY9cGw:bo1s579) ([raw file](https://github.com/gramineproject/gramine/blob/fd0c88e23973f27ea72ca3e363943e4bcf36104f/libos/src/net/unix.c#L136)):*

I always have to pause and think about this comment. Maybe

`so this is race-free` -> ```so using `options` when creating the socket below is race-free ```
</blockquote></details>

Slightly reworded, let me know it it looks ok
___
*[`libos/src/net/unix.c` line 245 at r2](https://reviewable.io/reviews/gramineproject/gramine/677#-N5EZzw83L3v-rest_8d:-N5FqnkaCg6EL3-OggvK:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/fd0c88e23973f27ea72ca3e363943e4bcf36104f/libos/src/net/unix.c#L245)):*

I always have to pause and think about this comment. Maybe

`so this is race-free` -> ```so using `options` when creating the socket below is race-free ```
</blockquote></details>

ditto
___
*[`libos/src/sys/shim_fcntl.c` line 35 at r2](https://reviewable.io/reviews/gramineproject/gramine/677#-N5Ea5fWDkJmN0jJjPlf:-N5FqvS69mCWTJHcjRMT:bk7ncr4) ([raw file](https://github.com/gramineproject/gramine/blob/fd0c88e23973f27ea72ca3e363943e4bcf36104f/libos/src/sys/shim_fcntl.c#L35)):*

Incorrect in the sense that this has no "handle-specific" logic (like updating PAL handle's attributes and doing additional locking)?

I agree that this looks incorrect and hopefully is never used (instead the handle-specific callback is always used).
</blockquote></details>

Just that it only saves the value, doesn't actually **do** anything with it. For some handles and some flags it's probably fine (e.g. because code using them always check them), but generally sounds incorrect. IMO each handle type should have it's own callback and deal with these options as it wishes (then just memorizing them would be fine)>
___
*[`libos/src/sys/shim_pipe.c` line 58 at r2](https://reviewable.io/reviews/gramineproject/gramine/677#-N5EYMqH2W_vDi7Y3sfz:-N5FrGG4AtLCnSkY2Uzr:b-fmsmep) ([raw file](https://github.com/gramineproject/gramine/blob/fd0c88e23973f27ea72ca3e363943e4bcf36104f/libos/src/sys/shim_pipe.c#L58)):*

You just simplified it because who cares about modifications to `srv` and `cli` out arguments if this func anyway returns an error?
</blockquote></details>

This function is always used on newly created handles. I'm not sure why it doesn't create them itself, why does it take them as arguments (and populate).
Generally I think that having a function which initializes part of an already half initialized object is a bad smell (unless the initial object was fully initialized, the function just changes parts of it and all failures are fatal, i.e. the object is destroyed)>

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/677)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

mkow (changes requested):
> Reviewed 3 of 6 files at r1, 2 of 2 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/677)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`libos/src/fs/pipe/fs.c` line 216 at r1](https://reviewable.io/reviews/gramineproject/gramine/677#-N5B6kIX3Irka1UcluP2:-N5Fx183AbBCwpKoowgW:b-rpahry) ([raw file](https://github.com/gramineproject/gramine/blob/f0e82493b00facf778a82f23d987da93c8e60d9b/libos/src/fs/pipe/fs.c#L216)):*

FYI: this was a bug.
</blockquote></details>

:grimacing: 
___
*[`libos/src/fs/pipe/fs.c` line 192 at r3](https://reviewable.io/reviews/gramineproject/gramine/677#-N5FwicU7X_lXtghULeQ:-N5FwicU7X_lXtghULeR:b-sobsqn) ([raw file](https://github.com/gramineproject/gramine/blob/42629f613d31cc2d00351049137cefb6d1d38e01/libos/src/fs/pipe/fs.c#L192)):*
> ```C
>     }
> 
>     handle->flags = (handle->flags & ~mask) | flags;
> ```

Not sure if it matters, but `handle->flags` is still `int` (i.e. signed, and different than the types you used here).
___
*[`libos/src/sys/shim_fcntl.c` line 35 at r2](https://reviewable.io/reviews/gramineproject/gramine/677#-N5Ea5fWDkJmN0jJjPlf:-N5FzGOY5dDrEao4Ilbe:bntf3an) ([raw file](https://github.com/gramineproject/gramine/blob/fd0c88e23973f27ea72ca3e363943e4bcf36104f/libos/src/sys/shim_fcntl.c#L35)):*

Just that it only saves the value, doesn't actually **do** anything with it. For some handles and some flags it's probably fine (e.g. because code using them always check them), but generally sounds incorrect. IMO each handle type should have it's own callback and deal with these options as it wishes (then just memorizing them would be fine)>
</blockquote></details>

Could you add this explanation to the comment?

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/677)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`libos/src/fs/pipe/fs.c` line 192 at r3](https://reviewable.io/reviews/gramineproject/gramine/677#-N5FwicU7X_lXtghULeQ:-N5G9_BX9ciMlWGFPvaA:bhcpoay) ([raw file](https://github.com/gramineproject/gramine/blob/42629f613d31cc2d00351049137cefb6d1d38e01/libos/src/fs/pipe/fs.c#L192)):*

Not sure if it matters, but `handle->flags` is still `int` (i.e. signed, and different than the types you used here).
</blockquote></details>

The highest bit cannot be ever set, so it does not. We should change it to `unsigned int`, but not in this PR.
___
*[`libos/src/sys/shim_fcntl.c` line 35 at r2](https://reviewable.io/reviews/gramineproject/gramine/677#-N5Ea5fWDkJmN0jJjPlf:-N5GAebY93hBS6b2e89m:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/fd0c88e23973f27ea72ca3e363943e4bcf36104f/libos/src/sys/shim_fcntl.c#L35)):*

Could you add this explanation to the comment?
</blockquote></details>

Done.

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/677)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

___
*[`libos/src/fs/pipe/fs.c` line 192 at r3](https://reviewable.io/reviews/gramineproject/gramine/677#-N5FwicU7X_lXtghULeQ:-N5GC3ZsDfoXOCxAitlG:b7h8ijt) ([raw file](https://github.com/gramineproject/gramine/blob/42629f613d31cc2d00351049137cefb6d1d38e01/libos/src/fs/pipe/fs.c#L192)):*

The highest bit cannot be ever set, so it does not. We should change it to `unsigned int`, but not in this PR.
</blockquote></details>

yeah, I know, it's more like a stylistic issue ;)

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/677)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

PR #678, opened 2022-06-22
[LibOS] Handle `IP_RECVERR` socket option as a no-op

### LINKED ISSUE #673: Domain name resolution stopped working
## Description of the problem
Domain name resolution stopped working on the latest version (5a39aee217955701b412d4ff807c32da8e53ab94).
Previos version (57b2cb30e7def6e1f34f8875496952b9c1d310ae) didn't have that problem.

## Steps to reproduce
Some of my applications stopped recognizing domain names. I reproduced the problem by writing simple code:
```
int main(int argc, char *argv[])
{
    if (argc != 2) {
        printf("usage: %s <hostname>\n", argv[0]);
        exit(1);
    }

    hostent * record = gethostbyname(argv[1]);
    if(record == NULL) {
        printf("%s is unavailable\n", argv[1]);
        exit(1);
    }

    in_addr * address = (in_addr * )record->h_addr;
    string ip_address = inet_ntoa(* address);

    cout << argv[1] << " (" << ip_address << ")\n";

    return 0;
}
```
The required files are allowed in the manifest
```
...
sgx.allowed_files = [
  "file:/etc/nsswitch.conf",
  "file:/etc/ethers",
  "file:/etc/hosts",
  "file:/etc/group",
  "file:/etc/passwd",
  "file:/etc/localtime",
  "file:/etc/host.conf",
  "file:/etc/resolv.conf",
]
...
```

## Expected results
IP address of the passed name is determined
## Actual results
Getting an error: "hostname is unavailable"
## Additional information
 This error in the log:
`[P1:T1:gethostbyname] trace: ---- shim_setsockopt(3, 0, 11, 0x70b56191860c, 4) = -92`

I think it's related to [this](https://github.com/gramineproject/gramine/pull/579) commit

## Gramine commit hash
5a39aee217955701b412d4ff807c32da8e53ab94

boryspoplawski:
## Description of the changes 
This socket option allows for later retrieval of detailed error descriptions that happened on the socket. We do not support such retrieval, so allowing to set this option basically defers the moment the app notices that it's not supported from the call to `setsockopt`, to the moment the error condition actually happens (which might be never).

Fixes #673

## How to test this PR? 
Try running repro code from https://github.com/gramineproject/gramine/issues/673

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/678)

### DIFF
--- a/libos/src/net/ip.c
+++ b/libos/src/net/ip.c
@@ -256,6 +256,25 @@ static int set_tcp_option(struct shim_handle* handle, int optname, void* optval,
     return pal_to_unix_errno(ret);
 }
 
+static int set_ipv4_option(struct shim_handle* handle, int optname, void* optval, size_t len) {
+    __UNUSED(handle);
+    __UNUSED(optval);
+    if (optname == IP_RECVERR) {
+        if (len < sizeof(int)) {
+            return -EINVAL;
+        }
+        /* We ignore this option. Full support would require handling `MSG_ERRQUEUE` in `recvmsg`
+         * syscall (which now fails with `-EOPNOTSUPP` if this flag is passed), which would be hard
+         * to implement. This basically defers the moment the app notices this error reporting
+         * mechanism is not supported from `setsockopt` call to when actual error condition on
+         * the socket happens (which might be never). */
+        return 0;
+    }
+
+    /* No other option supported at the moment. */
+    return -ENOPROTOOPT;
+}
+
 static int set_ipv6_option(struct shim_handle* handle, int optname, void* optval, size_t len) {
     PAL_STREAM_ATTR attr;
     int ret = DkStreamAttributesQueryByHandle(handle->info.sock.pal_handle, &attr);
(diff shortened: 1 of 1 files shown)


dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/678)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required) (waiting on @boryspoplawski)

___
*[`libos/src/net/ip.c` line 285 at r1](https://reviewable.io/reviews/gramineproject/gramine/678#-N5EbJQ2BcXWvPuHTFMC:-N5EbJQ2BcXWvPuHTFMD:b-m4u565) ([raw file](https://github.com/gramineproject/gramine/blob/2542b4a0f490e46752013b96bd3622ecba2dbd13/libos/src/net/ip.c#L285)):*
> ```C
>         /* See the comment in `set_ipv4_option` for why we handle it this way. */
>         return 0;
>     }
> ```

Actually, you could just move this under the `switch (optname)` -- this will basically be just 2 lines of code.

I guess you don't want to call `DkStreamAttributesQueryByHandle()` unnecessarily, but on the other hand -- this `setsockopt()` is a rare operation, so we don't lose any performance on this. Not blocking though.

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/678)* status: all files reviewed, 1 unresolved discussion (waiting on @boryspoplawski)

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/678)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #704, opened 2022-06-30
[common,PAL] Introduce constant-time memcmp and use it for MAC check

kailun-qin:
The MAC check in `sgx_verify_report()` should be done in a constant-time manner
to avoid side-channel leaks via timing attacks which could potentially allow an
attacker to guess a valid value and thereby bypass security controls. Similar
hardening is also applied in SGX SDK.

Signed-off-by: Kailun Qin <kailun.qin@intel.com>

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)

### DIFF
--- a/common/include/api.h
+++ b/common/include/api.h
@@ -248,6 +248,21 @@ void* memmove(void* dest, const void* src, size_t count);
 void* memset(void* dest, int ch, size_t count);
 int memcmp(const void* lhs, const void* rhs, size_t count);
 
+/*!
+ * \brief Constant-time buffer comparison.
+ *
+ * \param lhs    Pointer to the first buffer.
+ * \param rhs    Pointer to the second buffer.
+ * \param count  The number of bytes to compare in the buffer.
+ *
+ * \returns true if the content of the two buffers is the same, otherwise false.
+ *
+ * The time taken by this function depends on `count`, but not on the data at `lhs` or `rhs`.
+ * Hence, it can be used for comparing cryptographic secrets, hashes, message authentication codes
+ * etc. without timing side-channel leaks.
+ */
+bool ct_memequal(const void* lhs, const void* rhs, size_t count);
+
 /* Used by _FORTIFY_SOURCE */
 void* __memcpy_chk(void* restrict dest, const void* restrict src, size_t count, size_t dest_count);
 void* __memmove_chk(void* dest, const void* src, size_t count, size_t dest_count);
--- a/common/src/arch/meson.build
+++ b/common/src/arch/meson.build
@@ -1,0 +1,1 @@
+subdir(host_machine.cpu_family())
--- a/common/src/arch/x86_64/ct_memequal.nasm
+++ b/common/src/arch/x86_64/ct_memequal.nasm
@@ -1,0 +1,33 @@
+; SPDX-License-Identifier: LGPL-3.0-or-later
+; Copyright (C) 2022 Intel Corporation
+
+; Constant-time buffer comparison
+
+global ct_memequal
+
+section .text
+
+; bool ct_memequal(const void* lhs, const void* rhs, size_t count)
+; Arguments:
+; - RDI: Pointer to the first buffer (lhs)
+; - RSI: Pointer to the second buffer (rhs)
+; - RDX: The number of bytes to compare in the buffer (count)
+ct_memequal:
+    mov     ecx, 0
+    test    rdx, rdx
+    jz      .return
+    add     rdx, rdi
+.loop:
+    movzx   eax, byte [rdi]
+    movzx   r8d, byte [rsi]
+    add     rdi, 1
+    add     rsi, 1
+    xor     eax, r8d        ; using bitwise operation rather than a branch
+    or      ecx, eax
+    cmp     rdi, rdx
+    jnz     .loop
+.return:
+    mov     eax, ecx
+    test    al, al
+    sete    al
+    ret
--- a/common/src/arch/x86_64/meson.build
+++ b/common/src/arch/x86_64/meson.build
@@ -1,0 +1,10 @@
+common_src_arch_nasm = nasm_gen.process(
+    'ct_memequal.nasm',
+)
+
+common_src_arch_c = files()
+
+common_src_arch = [
+    common_src_arch_nasm,
+    common_src_arch_c,
+]
--- a/common/src/meson.build
+++ b/common/src/meson.build
@@ -21,6 +21,12 @@ common_src = files(
     'string/toml_utils.c',
     'string/utils.c',
 )
+
+# Arch-specific meson.build must define the following Meson variables:
+#   - `common_src_arch` - a list of arch-specific sources.
+subdir('arch')
+common_src += common_src_arch
+
 if asan
     common_src += files('asan.c')
 endif
--- a/pal/src/host/linux-sgx/enclave_framework.c
+++ b/pal/src/host/linux-sgx/enclave_framework.c
@@ -230,7 +230,7 @@ int sgx_verify_report(sgx_report_t* report) {
     log_debug("Verify report:");
     print_report(report);
 
-    if (memcmp(&check_mac, &report->mac, sizeof(check_mac))) {
+    if (!ct_memequal(&check_mac, &report->mac, sizeof(check_mac))) {
         log_error("Report verification failed");
         return -PAL_ERROR_DENIED;
     }


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/704#-N5ntKPo1fpmwimS5rdZ:-N5ntKPo1fpmwimS5rd_:b-c2bqwp) (no related file):*
If we keep using this "adapted from mbedTLS" function (I propose to re-write ourselves to asm), then we need to update https://github.com/gramineproject/gramine/blob/master/LICENSE.addendum.txt.

___
___
*[`common/include/api.h` line 264 at r1](https://reviewable.io/reviews/gramineproject/gramine/704#-N5nrE7aDgl5O7oH_z1z:-N5nrE7aDgl5O7oH_z2-:b65nyix) ([raw file](https://github.com/gramineproject/gramine/blob/205f0dbd7782456fa9c96f3e33f2031574a64953/common/include/api.h#L264)):*
> ```objc
>  *
>  * \return        Zero if the content of the two buffer is the same,
>  *                otherwise non-zero.
> ```

1. Please use `/*!` in the very beginning of the comment.
2. Use `\brief` for the first sentence.
3. Then follow with params.
4. Then `\returns` (not `\return`). Also, `Zero` -> `0`, and no intendation.
5. Only then put the long text.
6. Re-wrap all lines to 100-char-limit

See an example of how we format Doxygen comments e.g. here: https://github.com/gramineproject/gramine/blob/bb7769a1c48cd6c39c009abc695f3db2f6985aba/pal/include/pal/pal.h#L275
___
*[`common/src/string/ct_memcmp.c` line 4 at r1](https://reviewable.io/reviews/gramineproject/gramine/704#-N5nronq3sLb13pCICQ2:-N5nronq3sLb13pCICQ3:b-nqtqih) ([raw file](https://github.com/gramineproject/gramine/blob/205f0dbd7782456fa9c96f3e33f2031574a64953/common/src/string/ct_memcmp.c#L4)):*
> ```C
> /* Adapted from Mbed TLS v3.1.0.
>  * Copyright (C) 2021, The Mbed TLS Contributors
>  *               2022, Intel Corporation
> ```

We don't use `,` after the year. Please fix two lines.

Please move the `Adapted ...` line to after the copyrights.

A good example: https://github.com/gramineproject/gramine/blob/master/libos/src/sys/shim_epoll.c
___
*[`common/src/string/ct_memcmp.c` line 12 at r1](https://reviewable.io/reviews/gramineproject/gramine/704#-N5nsFCW2Dl4lCdtIw0f:-N5nsFCW2Dl4lCdtIw0g:b-vb1lrn) ([raw file](https://github.com/gramineproject/gramine/blob/205f0dbd7782456fa9c96f3e33f2031574a64953/common/src/string/ct_memcmp.c#L12)):*
> ```C
>     size_t i;
>     volatile const unsigned char *l = (volatile const unsigned char *) lhs;
>     volatile const unsigned char *r = (volatile const unsigned char *) rhs;
> ```

Please fix these lines like this:
```
volatile const unsigned char* r = (volatile const unsigned char*)rhs;
```
___
*[`common/src/string/ct_memcmp.c` line 13 at r1](https://reviewable.io/reviews/gramineproject/gramine/704#-N5nsbR3EAHFItpQPzRc:-N5nsbR3EAHFItpQPzRd:b9go7gk) ([raw file](https://github.com/gramineproject/gramine/blob/205f0dbd7782456fa9c96f3e33f2031574a64953/common/src/string/ct_memcmp.c#L13)):*
> ```C
>     volatile const unsigned char *l = (volatile const unsigned char *) lhs;
>     volatile const unsigned char *r = (volatile const unsigned char *) rhs;
>     volatile unsigned char diff = 0;
> ```

I feel like this is not really preventing the compiler to optimize the hack out of this code. @boryspoplawski and @mkow may give better opinions here.

But in general, I think we should just write this function in pure assembly.
___
*[`common/src/string/ct_memcmp.c` line 18 at r1](https://reviewable.io/reviews/gramineproject/gramine/704#-N5nsP0B7_JhLZV7_Y6k:-N5nsP0B7_JhLZV7_Y6l:boyn84i) ([raw file](https://github.com/gramineproject/gramine/blob/205f0dbd7782456fa9c96f3e33f2031574a64953/common/src/string/ct_memcmp.c#L18)):*
> ```C
>         /* Read volatile data in order before computing diff to avoid the warning:
>          * 'the order of volatile accesses is undefined ..' */
>         unsigned char x = l[i], y = r[i];
> ```

Please split into two lines. I find this hard to read.
___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 233 at r1](https://reviewable.io/reviews/gramineproject/gramine/704#-N5nsu7KDWjnizCR3Hs7:-N5nsu7KDWjnizCR3Hs8:b-9fnvbm) ([raw file](https://github.com/gramineproject/gramine/blob/205f0dbd7782456fa9c96f3e33f2031574a64953/pal/src/host/linux-sgx/enclave_framework.c#L233)):*
> ```C
>     print_report(report);
> 
>     if (ct_memcmp(&check_mac, &report->mac, sizeof(check_mac))) {
> ```

Did you check if there are more `memcmp()` usages like this?

I believe the previous `memcmp()` usage indeed could leak some info on the "first diverging byte", so looks like this is a legitimate change.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/704#-N5ntKPo1fpmwimS5rdZ:-N5oAvky2n4Vbm1AimJl:bxjs1y1) (no related file):*

If we keep using this "adapted from mbedTLS" function (I propose to re-write ourselves to asm), then we need to update https://github.com/gramineproject/gramine/blob/master/LICENSE.addendum.txt.
</blockquote></details>

I don't think it's needed/good to rewrite it ourselves as there are already mature constant-time memcmp C implementations available there (e.g., another option is to use consttime_memequal by Netbsd: https://github.com/NetBSD/src/blob/05082e19134c05f2f4b6eca73223cdc6b5ab09bf/common/lib/libc/string/consttime_memequal.c#L19, picked by SGX SDK). But I'm open to it. :)

___
___
*[`common/include/api.h` line 264 at r1](https://reviewable.io/reviews/gramineproject/gramine/704#-N5nrE7aDgl5O7oH_z1z:-N5oCuDC4kbUAgTpNTBH:bt1x36r) ([raw file](https://github.com/gramineproject/gramine/blob/205f0dbd7782456fa9c96f3e33f2031574a64953/common/include/api.h#L264)):*

1. Please use `/*!` in the very beginning of the comment.
2. Use `\brief` for the first sentence.
3. Then follow with params.
4. Then `\returns` (not `\return`). Also, `Zero` -> `0`, and no intendation.
5. Only then put the long text.
6. Re-wrap all lines to 100-char-limit

See an example of how we format Doxygen comments e.g. here: https://github.com/gramineproject/gramine/blob/bb7769a1c48cd6c39c009abc695f3db2f6985aba/pal/include/pal/pal.h#L275
</blockquote></details>

Sure, will do.
___
*[`common/src/string/ct_memcmp.c` line 4 at r1](https://reviewable.io/reviews/gramineproject/gramine/704#-N5nronq3sLb13pCICQ2:-N5oCBNA4Lc9s05u9d_4:b-pcoct) ([raw file](https://github.com/gramineproject/gramine/blob/205f0dbd7782456fa9c96f3e33f2031574a64953/common/src/string/ct_memcmp.c#L4)):*

We don't use `,` after the year. Please fix two lines.

Please move the `Adapted ...` line to after the copyrights.

A good example: https://github.com/gramineproject/gramine/blob/master/libos/src/sys/shim_epoll.c
</blockquote></details>

Gotcha, thanks. But I also noticed there are multiple places that are still doing this way, e.g., https://github.com/gramineproject/gramine/blob/master/pal/src/host/linux-sgx/sgx_platform.c#L2. Well, we can fix them in another thread.
___
*[`common/src/string/ct_memcmp.c` line 12 at r1](https://reviewable.io/reviews/gramineproject/gramine/704#-N5nsFCW2Dl4lCdtIw0f:-N5oCp7H4UYbuKg5uIpj:bt1x36r) ([raw file](https://github.com/gramineproject/gramine/blob/205f0dbd7782456fa9c96f3e33f2031574a64953/common/src/string/ct_memcmp.c#L12)):*

Please fix these lines like this:
```
volatile const unsigned char* r = (volatile const unsigned char*)rhs;
```
</blockquote></details>

Sure, will do.
___
*[`common/src/string/ct_memcmp.c` line 18 at r1](https://reviewable.io/reviews/gramineproject/gramine/704#-N5nsP0B7_JhLZV7_Y6k:-N5oClAPDzcYtuk9Tm2Y:bt1x36r) ([raw file](https://github.com/gramineproject/gramine/blob/205f0dbd7782456fa9c96f3e33f2031574a64953/common/src/string/ct_memcmp.c#L18)):*

Please split into two lines. I find this hard to read.
</blockquote></details>

Sure, will do.
___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 233 at r1](https://reviewable.io/reviews/gramineproject/gramine/704#-N5nsu7KDWjnizCR3Hs7:-N5oCxGr8Lz7x-TKDcQw:b-2434sh) ([raw file](https://github.com/gramineproject/gramine/blob/205f0dbd7782456fa9c96f3e33f2031574a64953/pal/src/host/linux-sgx/enclave_framework.c#L233)):*

Did you check if there are more `memcmp()` usages like this?

I believe the previous `memcmp()` usage indeed could leak some info on the "first diverging byte", so looks like this is a legitimate change.
</blockquote></details>

Yes, I do. Generally, the risk is applicable to all the other comparing of integrity or authenticity data. But this is the one I'm certain of - maybe we can do a more thorough analysis on the rest, e.g., comparing file hashes.

boryspoplawski (changes requested):
> Reviewed 2 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/704#-N5ntKPo1fpmwimS5rdZ:-N5oL1_aAMRByJd3VICK:bdzfdt4) (no related file):*

I don't think it's needed/good to rewrite it ourselves as there are already mature constant-time memcmp C implementations available there (e.g., another option is to use consttime_memequal by Netbsd: https://github.com/NetBSD/src/blob/05082e19134c05f2f4b6eca73223cdc6b5ab09bf/common/lib/libc/string/consttime_memequal.c#L19, picked by SGX SDK). But I'm open to it. :)
</blockquote></details>

The problem is that we don't trust code copied from "the internet" and we need to analyze it anyway, so usually it's not much of a difference in copy vs write ourselves.

___
___
*[`common/src/string/ct_memcmp.c` line 4 at r1](https://reviewable.io/reviews/gramineproject/gramine/704#-N5nronq3sLb13pCICQ2:-N5oQcdH0as3iJ5xUdjg:bex5xgc) ([raw file](https://github.com/gramineproject/gramine/blob/205f0dbd7782456fa9c96f3e33f2031574a64953/common/src/string/ct_memcmp.c#L4)):*

Gotcha, thanks. But I also noticed there are multiple places that are still doing this way, e.g., https://github.com/gramineproject/gramine/blob/master/pal/src/host/linux-sgx/sgx_platform.c#L2. Well, we can fix them in another thread.
</blockquote></details>

I wouldn't bother creating a PR just for the sake of removing `,` there, just don't introduce it in new PRs.
___
*[`common/src/string/ct_memcmp.c` line 13 at r1](https://reviewable.io/reviews/gramineproject/gramine/704#-N5nsbR3EAHFItpQPzRc:-N5oRFbp4AiAvKEBhBV5:bdq9mqz) ([raw file](https://github.com/gramineproject/gramine/blob/205f0dbd7782456fa9c96f3e33f2031574a64953/common/src/string/ct_memcmp.c#L13)):*

I feel like this is not really preventing the compiler to optimize the hack out of this code. @boryspoplawski and @mkow may give better opinions here.

But in general, I think we should just write this function in pure assembly.
</blockquote></details>

Yes we should. Albeit the risk is rather theoretical, I could imagine a scenario where this is inlined, both arguments are pointers to local stack variables (so volatile semantics are not clear), gcc recognizes this as a memcmp pattern and replaces with a builtin.
___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 233 at r1](https://reviewable.io/reviews/gramineproject/gramine/704#-N5nsu7KDWjnizCR3Hs7:-N5oMKYK31IIoYaiFkBR:be2j03a) ([raw file](https://github.com/gramineproject/gramine/blob/205f0dbd7782456fa9c96f3e33f2031574a64953/pal/src/host/linux-sgx/enclave_framework.c#L233)):*

Yes, I do. Generally, the risk is applicable to all the other comparing of integrity or authenticity data. But this is the one I'm certain of - maybe we can do a more thorough analysis on the rest, e.g., comparing file hashes.
</blockquote></details>

@dimakuv  If I understand correctly `report` comes from untrusted source, so the adversary could obtain a MAC with `report_key` for arbitrary data, by submitting the report and then timing `memcmp` on the actual MAC vs submitted, fully controlled `report->mac`.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @mkow)

___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 233 at r1](https://reviewable.io/reviews/gramineproject/gramine/704#-N5nsu7KDWjnizCR3Hs7:-N5oTE-D3QeUQVmt9ivN:b-ylvisk) ([raw file](https://github.com/gramineproject/gramine/blob/205f0dbd7782456fa9c96f3e33f2031574a64953/pal/src/host/linux-sgx/enclave_framework.c#L233)):*

@dimakuv  If I understand correctly `report` comes from untrusted source, so the adversary could obtain a MAC with `report_key` for arbitrary data, by submitting the report and then timing `memcmp` on the actual MAC vs submitted, fully controlled `report->mac`.
</blockquote></details>

@boryspoplawski Yes, I agree with this attack.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @kailun-qin, and @mkow)

___
*[`common/src/string/ct_memcmp.S` line 1 at r2](https://reviewable.io/reviews/gramineproject/gramine/704#-N5pCIat7Une2xhSxGNJ:-N5pCIauDlpfIzwcU2Wr:b-oxz8ss) ([raw file](https://github.com/gramineproject/gramine/blob/561b2363d5f003d1e36c7fff67a8bb52a6e6793d/common/src/string/ct_memcmp.S#L1)):*
> ```
> ```

Hmm, we use `nasm` for new assembly files, could you rewrite?
___
*[`common/src/string/ct_memcmp.S` line 10 at r2](https://reviewable.io/reviews/gramineproject/gramine/704#-N5pCYaz7qgiUGeLhbSP:-N5pCYaz7qgiUGeLhbSQ:buhr7zp) ([raw file](https://github.com/gramineproject/gramine/blob/561b2363d5f003d1e36c7fff67a8bb52a6e6793d/common/src/string/ct_memcmp.S#L10)):*
> ```
>     .cfi_startproc
> 
>     # Arguments:
> ```

I find it useful to also but C level function signature. Please add it (either replace this commend or just extend it)
___
*[`common/src/string/ct_memcmp.S` line 15 at r2](https://reviewable.io/reviews/gramineproject/gramine/704#-N5pCva-6ux6zJ22pL-0:-N5pCva-6ux6zJ22pL-1:b-m2gwkk) ([raw file](https://github.com/gramineproject/gramine/blob/561b2363d5f003d1e36c7fff67a8bb52a6e6793d/common/src/string/ct_memcmp.S#L15)):*
> ```
>     #   RDX: The number of bytes to compare in the buffer (count)
> 
>     movb   $0x0, -0x1(%rsp)
> ```

Why do you keep this variable on the stack? It can easily be just a register.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

___
*[`common/src/string/ct_memcmp.S` line 1 at r2](https://reviewable.io/reviews/gramineproject/gramine/704#-N5pCIat7Une2xhSxGNJ:-N5pS3y59VdFVROUfu8v:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/561b2363d5f003d1e36c7fff67a8bb52a6e6793d/common/src/string/ct_memcmp.S#L1)):*

Hmm, we use `nasm` for new assembly files, could you rewrite?
</blockquote></details>

Done.
___
*[`common/src/string/ct_memcmp.S` line 10 at r2](https://reviewable.io/reviews/gramineproject/gramine/704#-N5pCYaz7qgiUGeLhbSP:-N5pS7ZKDgBnDwQ46I2N:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/561b2363d5f003d1e36c7fff67a8bb52a6e6793d/common/src/string/ct_memcmp.S#L10)):*

I find it useful to also but C level function signature. Please add it (either replace this commend or just extend it)
</blockquote></details>

Done.
___
*[`common/src/string/ct_memcmp.S` line 15 at r2](https://reviewable.io/reviews/gramineproject/gramine/704#-N5pCva-6ux6zJ22pL-0:-N5pSFQK8AzwGjoyz4eL:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/561b2363d5f003d1e36c7fff67a8bb52a6e6793d/common/src/string/ct_memcmp.S#L15)):*

Why do you keep this variable on the stack? It can easily be just a register.
</blockquote></details>

Done.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: 3 of 6 files reviewed, 10 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, @kailun-qin, and @mkow)

___
*[`common/src/string/ct_memcmp.nasm` line 29 at r3](https://reviewable.io/reviews/gramineproject/gramine/704#-N5pV2ch1iLxM9dcy_fg:-N5pV2ch1iLxM9dcy_fh:b-supewy) ([raw file](https://github.com/gramineproject/gramine/blob/b3c0419974923bebbe56f14b8828a353e926a0e6/common/src/string/ct_memcmp.nasm#L29)):*
> ```Assembly
>         jnz     L1                 ;
> L2:     movzx   eax, r8b           ;
>         ret                        ;
> ```

Do we want to do this? We don't have a nasm styleguide yet, so it's a good moment to discuss what style we want for it.
@boryspoplawski, @dimakuv: any preferences?



_Code quote:_
```Assembly
                        ;
```

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: all files reviewed, 11 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @kailun-qin, and @mkow)

___
*[`common/src/meson.build` line 24 at r3](https://reviewable.io/reviews/gramineproject/gramine/704#-N5pZANj6b9-WjbfEz-s:-N5pZANj6b9-WjbfEz-t:b-35wqeq) ([raw file](https://github.com/gramineproject/gramine/blob/b3c0419974923bebbe56f14b8828a353e926a0e6/common/src/meson.build#L24)):*
> ```Meson
>     'string/utils.c': {},
>     'string/ct_memcmp.nasm': { 'type': 'nasm' },
> }
> ```

Please just split these into two lists, one for `.c` and one for `.nasm` files
Update: the second one should be in arch specific dir anyway
___
*[`common/src/string/ct_memcmp.nasm` line 1 at r3](https://reviewable.io/reviews/gramineproject/gramine/704#-N5p_2jO616Dzu-p2Ttn:-N5p_2jO616Dzu-p2Tto:b-b29y9f) ([raw file](https://github.com/gramineproject/gramine/blob/b3c0419974923bebbe56f14b8828a353e926a0e6/common/src/string/ct_memcmp.nasm#L1)):*
> ```Assembly
> ; SPDX-License-Identifier: LGPL-3.0-or-later
> ```

Actually, this file must go into arch specific directory, like `common/src/arch/x86_64/`
___
*[`common/src/string/ct_memcmp.nasm` line 16 at r3](https://reviewable.io/reviews/gramineproject/gramine/704#-N5p_Jf_-KzL_4JMN8P4:-N5p_Jf_-KzL_4JMN8P5:b-8o7t8l) ([raw file](https://github.com/gramineproject/gramine/blob/b3c0419974923bebbe56f14b8828a353e926a0e6/common/src/string/ct_memcmp.nasm#L16)):*
> ```Assembly
>         mov     r8b, 0             ;
>         test    rdx, rdx           ;
>         jz      L2                 ;
> ```

Please add `.` to local labels
___
*[`common/src/string/ct_memcmp.nasm` line 25 at r3](https://reviewable.io/reviews/gramineproject/gramine/704#-N5p_X3KBat5M_NHB3Se:-N5p_X3KBat5M_NHB3Sf:b-xlkc4w) ([raw file](https://github.com/gramineproject/gramine/blob/b3c0419974923bebbe56f14b8828a353e926a0e6/common/src/string/ct_memcmp.nasm#L25)):*
> ```Assembly
>         xor     eax, r9d           ;
>         or      eax, ecx           ;
>         mov     r8b, al            ;
> ```

Why not simpler:
```
add rsi, 1
xor eax, r9d
or ecx, eax
cmp ...
```
and just keep the bit mask in `ecx` (need to zero it instead of `r8d` at begining)
___
*[`common/src/string/ct_memcmp.nasm` line 29 at r3](https://reviewable.io/reviews/gramineproject/gramine/704#-N5pV2ch1iLxM9dcy_fg:-N5pZgm64cFWapTtpeA7:b-juhfia) ([raw file](https://github.com/gramineproject/gramine/blob/b3c0419974923bebbe56f14b8828a353e926a0e6/common/src/string/ct_memcmp.nasm#L29)):*

Do we want to do this? We don't have a nasm styleguide yet, so it's a good moment to discuss what style we want for it.
@boryspoplawski, @dimakuv: any preferences?
</blockquote></details>

Why would we want to have it? What purpose does it serve?
I dislike it.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: all files reviewed, 11 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

___
*[`common/src/meson.build` line 24 at r3](https://reviewable.io/reviews/gramineproject/gramine/704#-N5pZANj6b9-WjbfEz-s:-N5r3KXnCoCHyPVpOZ-7:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/b3c0419974923bebbe56f14b8828a353e926a0e6/common/src/meson.build#L24)):*

Please just split these into two lists, one for `.c` and one for `.nasm` files
Update: the second one should be in arch specific dir anyway
</blockquote></details>

Done.
___
*[`common/src/string/ct_memcmp.nasm` line 1 at r3](https://reviewable.io/reviews/gramineproject/gramine/704#-N5p_2jO616Dzu-p2Ttn:-N5r3I6lAquHM-JGVMSi:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/b3c0419974923bebbe56f14b8828a353e926a0e6/common/src/string/ct_memcmp.nasm#L1)):*

Actually, this file must go into arch specific directory, like `common/src/arch/x86_64/`
</blockquote></details>

Done.
___
*[`common/src/string/ct_memcmp.nasm` line 16 at r3](https://reviewable.io/reviews/gramineproject/gramine/704#-N5p_Jf_-KzL_4JMN8P4:-N5r3cSI0rgLgYVm4m3Z:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/b3c0419974923bebbe56f14b8828a353e926a0e6/common/src/string/ct_memcmp.nasm#L16)):*

Please add `.` to local labels
</blockquote></details>

Done.
___
*[`common/src/string/ct_memcmp.nasm` line 25 at r3](https://reviewable.io/reviews/gramineproject/gramine/704#-N5p_X3KBat5M_NHB3Se:-N5r7oW-7CnlQ9n0Lhpg:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/b3c0419974923bebbe56f14b8828a353e926a0e6/common/src/string/ct_memcmp.nasm#L25)):*

Why not simpler:
```
add rsi, 1
xor eax, r9d
or ecx, eax
cmp ...
```
and just keep the bit mask in `ecx` (need to zero it instead of `r8d` at begining)
</blockquote></details>

Done.
___
*[`common/src/string/ct_memcmp.nasm` line 29 at r3](https://reviewable.io/reviews/gramineproject/gramine/704#-N5pV2ch1iLxM9dcy_fg:-N5r4HvbEKBnzUb5xFS7:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/b3c0419974923bebbe56f14b8828a353e926a0e6/common/src/string/ct_memcmp.nasm#L29)):*

Why would we want to have it? What purpose does it serve?
I dislike it.
</blockquote></details>

Done.

dimakuv (approved):
> Reviewed 1 of 4 files at r2, 5 of 6 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: 9 of 10 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @kailun-qin, and @mkow)

___
*[`common/include/api.h` line 258 at r4](https://reviewable.io/reviews/gramineproject/gramine/704#-N5ssUEA3ykzDfrzzrej:-N5ssUEB2P1Qk7Gji2gU:bt8sr7d) ([raw file](https://github.com/gramineproject/gramine/blob/662b5063c4a0d8074bc417f6e5e95e5325f67053/common/include/api.h#L258)):*
> ```objc
>  * \param count  The number of bytes to compare in the buffer.
>  *
>  * \returns      Zero if the content of the two buffer is the same, otherwise non-zero.
> ```

We typicall write like this:
```
\returns 0 if the content of the two buffers is the same, otherwise non-zero.
```

But not blocking as the current version looks fine to me too.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: 9 of 10 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @kailun-qin, and @mkow)

___
*[`common/src/meson.build` line 13 at r4](https://reviewable.io/reviews/gramineproject/gramine/704#-N5t2xFm2WfzPIRPSZn8:-N5t2xFm2WfzPIRPSZn9:bipjfao) ([raw file](https://github.com/gramineproject/gramine/blob/662b5063c4a0d8074bc417f6e5e95e5325f67053/common/src/meson.build#L13)):*
> ```Meson
> )
> 
> subdir('string')
> ```

Wait, what happened here? Why did you move `string` dir?
___
*[`common/src/string/meson.build` line 18 at r4](https://reviewable.io/reviews/gramineproject/gramine/704#-N5t3Ho05f7aoJVIMKFK:-N5t3Ho05f7aoJVIMKFL:b-7y1l0j) ([raw file](https://github.com/gramineproject/gramine/blob/662b5063c4a0d8074bc417f6e5e95e5325f67053/common/src/string/meson.build#L18)):*
> ```Meson
> # Arch-specific meson.build must define the following Meson variables:
> #   - `string_src_arch` - a list of arch-specific sources.
> subdir('arch')
> ```

`arch` should be under` src` directory, not in some nested one.
Just revert all these moves, create `common/src/arch/x86_64/` and put `ct_memcmp.nasm` there

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: 9 of 10 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @kailun-qin, and @mkow)

___
*[`common/include/api.h` line 258 at r4](https://reviewable.io/reviews/gramineproject/gramine/704#-N5ssUEA3ykzDfrzzrej:-N5t9cwdEphWfKFAHSK9:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/662b5063c4a0d8074bc417f6e5e95e5325f67053/common/include/api.h#L258)):*

We typicall write like this:
```
\returns 0 if the content of the two buffers is the same, otherwise non-zero.
```

But not blocking as the current version looks fine to me too.
</blockquote></details>

Done.
___
*[`common/src/string/meson.build` line 18 at r4](https://reviewable.io/reviews/gramineproject/gramine/704#-N5t3Ho05f7aoJVIMKFK:-N5t7KxkBFrBSIrZVjZT:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/662b5063c4a0d8074bc417f6e5e95e5325f67053/common/src/string/meson.build#L18)):*

`arch` should be under` src` directory, not in some nested one.
Just revert all these moves, create `common/src/arch/x86_64/` and put `ct_memcmp.nasm` there
</blockquote></details>

Done.

dimakuv (dismissed):
> Reviewed 5 of 6 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin and @mkow)

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[`common/src/arch/x86_64/meson.build` line 13 at r5](https://reviewable.io/reviews/gramineproject/gramine/704#-N5tBKGj0k2HTcF39M6s:-N5tFg7J86_AQQN1QREL:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/4e931a627439dd53d36845db5dde39974706ab17/common/src/arch/x86_64/meson.build#L13)):*

Please either just:
1. `common_src_arch = nasm_gen.process('ct_memcmp.nasm')`
2. or split this list into two, one for nasm and one for C (empty atm)
(I know we have something similar in some other place, but it was done for other reasons and might change in the future)
I just don't want to use this unnecessarily complicated parameters list, we don't need so much configuration
</blockquote></details>

Done.

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow)

mkow (changes requested):
> Reviewed 1 of 4 files at r1, 1 of 4 files at r2, 1 of 3 files at r3, 1 of 6 files at r4, 5 of 6 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: all files reviewed, 7 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

___
*[`-- commits` line 7 at r6](https://reviewable.io/reviews/gramineproject/gramine/704#-N6EFvie02jqq56CSwWx):*
"applied" sounds like you just fixed this both here and in SDK (which I think isn't the case)



_Suggestion:_
```
is also used in SGX SDK
```
___
*[`common/include/api.h` line 252 at r6](https://reviewable.io/reviews/gramineproject/gramine/704#-N6EG9YO3_rBDsld5iKs:-N6EG9YO3_rBDsld5iKt:b-3lrqj) ([raw file](https://github.com/gramineproject/gramine/blob/8438324e09212850519174addb24f0f347f3f9db/common/include/api.h#L252)):*
> ```objc
> 
> /*!
>  * \brief Constant-time buffer comparison without branches.
> ```

But it actually has branches?



_Code quote (i):_
```objc
without branches
```



_Code quote (ii) (from common/src/arch/x86_64/ct_memcmp.nasm):_
```Assembly
jnz     .L1
```
___
*[`common/include/api.h` line 258 at r6](https://reviewable.io/reviews/gramineproject/gramine/704#-N6EGSwH5D9Ve7-9_19F:-N6EGSwH5D9Ve7-9_19G:b2l4s6o) ([raw file](https://github.com/gramineproject/gramine/blob/8438324e09212850519174addb24f0f347f3f9db/common/include/api.h#L258)):*
> ```objc
>  * \param count  The number of bytes to compare in the buffer.
>  *
>  * \returns 0 if the content of the two buffers is the same, otherwise non-zero.
> ```

If we call it "memcmp" then we need to preserve the original `memcmp()` semantics (which you don't). I'd rather change the name, it's not obvious to me how to implement original `memcmp()` semantics in constant time.



_Code quote:_
```objc
otherwise non-zero.
```
___
*[`common/include/api.h` line 260 at r6](https://reviewable.io/reviews/gramineproject/gramine/704#-N6EH9Fg9TuAPl2Ib9Jl:-N6EH9Fg9TuAPl2Ib9Jm:bnecf6g) ([raw file](https://github.com/gramineproject/gramine/blob/8438324e09212850519174addb24f0f347f3f9db/common/include/api.h#L260)):*
> ```objc
>  * \returns 0 if the content of the two buffers is the same, otherwise non-zero.
>  *
>  * This is equivalent to the standard memcmp function, but using bitwise operation rather than a
> ```

This is an implementation detail, not useful for the users of this function - so, it should be placed in .nasm file right where you do this trick.



_Code quote:_
```objc
using bitwise operation
```
___
*[`common/include/api.h` line 262 at r6](https://reviewable.io/reviews/gramineproject/gramine/704#-N6EGkaC0XpGRv8IZW9H:-N6EGkaC0XpGRv8IZW9I:b-hw6axe) ([raw file](https://github.com/gramineproject/gramine/blob/8438324e09212850519174addb24f0f347f3f9db/common/include/api.h#L262)):*
> It can be used to write constant-time code by replacing branches with bit operations using masks.

I don't understand this sentence. `ct_memcmp()` can be used to write [...] code by replacing branches? Maybe you wanted to say "they" instead of "it"?
___
*[`common/src/arch/x86_64/ct_memcmp.nasm` line 18 at r6](https://reviewable.io/reviews/gramineproject/gramine/704#-N6EHs56BNX3nUod-9BS:-N6EHs56BNX3nUod-9BT:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/8438324e09212850519174addb24f0f347f3f9db/common/src/arch/x86_64/ct_memcmp.nasm#L18)):*
> ```Assembly
>         jz      .L2
>         add     rdx, rdi
> .L1:    movzx   eax, byte [rdi]
> ```





_Suggestion:_
```Assembly
.loop:
```
___
*[`common/src/arch/x86_64/ct_memcmp.nasm` line 26 at r6](https://reviewable.io/reviews/gramineproject/gramine/704#-N6EHlY3Az0Ty1jddluy:-N6EHlY3Az0Ty1jddluz:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/8438324e09212850519174addb24f0f347f3f9db/common/src/arch/x86_64/ct_memcmp.nasm#L26)):*
> ```Assembly
>         cmp     rdi, rdx
>         jnz     .L1
> .L2:    mov     eax, ecx
> ```





_Suggestion:_
```Assembly
.return:
```

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: 6 of 10 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "squash! " and "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

___
*[`-- commits` line 7 at r6](https://reviewable.io/reviews/gramineproject/gramine/704#-N6EFvie02jqq56CSwWx):*

"applied" sounds like you just fixed this both here and in SDK (which I think isn't the case)
</blockquote></details>

Done, added a squash commit.
___
*[`common/include/api.h` line 252 at r6](https://reviewable.io/reviews/gramineproject/gramine/704#-N6EG9YO3_rBDsld5iKs:-N6HIFld7e9ZqoUVEtnh:b-qjlmb) ([raw file](https://github.com/gramineproject/gramine/blob/8438324e09212850519174addb24f0f347f3f9db/common/include/api.h#L252)):*

But it actually has branches?
</blockquote></details>

Indeed, there is still (asm) branch instruction. But I suppose the intention is that there is no `if`-like branch within the `loop`. Anyway, I remove it to avoid confusions.
___
*[`common/include/api.h` line 258 at r6](https://reviewable.io/reviews/gramineproject/gramine/704#-N6EGSwH5D9Ve7-9_19F:-N6HFjYe4-Ln21tWW-n1:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/8438324e09212850519174addb24f0f347f3f9db/common/include/api.h#L258)):*

If we call it "memcmp" then we need to preserve the original `memcmp()` semantics (which you don't). I'd rather change the name, it's not obvious to me how to implement original `memcmp()` semantics in constant time.
</blockquote></details>

Done.
___
*[`common/include/api.h` line 260 at r6](https://reviewable.io/reviews/gramineproject/gramine/704#-N6EH9Fg9TuAPl2Ib9Jl:-N6HLmoj1jmIi5K1DkyN:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/8438324e09212850519174addb24f0f347f3f9db/common/include/api.h#L260)):*

This is an implementation detail, not useful for the users of this function - so, it should be placed in .nasm file right where you do this trick.
</blockquote></details>

Done.
___
*[`common/include/api.h` line 262 at r6](https://reviewable.io/reviews/gramineproject/gramine/704#-N6EGkaC0XpGRv8IZW9H:-N6HMIyz5_IO34C_rFgt:bdmuoyc) ([raw file](https://github.com/gramineproject/gramine/blob/8438324e09212850519174addb24f0f347f3f9db/common/include/api.h#L262)):*

> It can be used to write constant-time code by replacing branches with bit operations using masks.

I don't understand this sentence. `ct_memcmp()` can be used to write [...] code by replacing branches? Maybe you wanted to say "they" instead of "it"?
</blockquote></details>

The original description was from mbedtls - I did not change it. I agree, this makes no sense to me neither. Updated.
___
*[`common/src/arch/x86_64/ct_memcmp.nasm` line 18 at r6](https://reviewable.io/reviews/gramineproject/gramine/704#-N6EHs56BNX3nUod-9BS:-N6HMlahCpSEJDswCUkv:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/8438324e09212850519174addb24f0f347f3f9db/common/src/arch/x86_64/ct_memcmp.nasm#L18)):*
> ```Assembly
>         jz      .L2
>         add     rdx, rdi
> .L1:    movzx   eax, byte [rdi]
> ```

Done.
___
*[`common/src/arch/x86_64/ct_memcmp.nasm` line 26 at r6](https://reviewable.io/reviews/gramineproject/gramine/704#-N6EHlY3Az0Ty1jddluy:-N6HMjjM95CCj40nmIf6:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/8438324e09212850519174addb24f0f347f3f9db/common/src/arch/x86_64/ct_memcmp.nasm#L26)):*
> ```Assembly
>         cmp     rdi, rdx
>         jnz     .L1
> .L2:    mov     eax, ecx
> ```

Done.

dimakuv (approved):
> Reviewed 4 of 4 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "squash! " and "fixup! " found in commit messages' one-liners (waiting on @mkow)

mkow (changes requested):
> Reviewed 4 of 4 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "squash! " and "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

___
*[`common/include/api.h` line 252 at r6](https://reviewable.io/reviews/gramineproject/gramine/704#-N6EG9YO3_rBDsld5iKs:-N6IY-Sx464aeQ_Im6qJ:bc1mwzo) ([raw file](https://github.com/gramineproject/gramine/blob/8438324e09212850519174addb24f0f347f3f9db/common/include/api.h#L252)):*

Indeed, there is still (asm) branch instruction. But I suppose the intention is that there is no `if`-like branch within the `loop`. Anyway, I remove it to avoid confusions.
</blockquote></details>

I think it was more confusing than helping :)
___
*[`common/include/api.h` line 258 at r6](https://reviewable.io/reviews/gramineproject/gramine/704#-N6EGSwH5D9Ve7-9_19F:-N6IYV7L5YdOyvPYqB9w:b-nlv66j) ([raw file](https://github.com/gramineproject/gramine/blob/8438324e09212850519174addb24f0f347f3f9db/common/include/api.h#L258)):*

Done.
</blockquote></details>

Maybe we could change the return type to `bool` now? It would have a more intuitive signature (`bool ct_memequal(...)`).
___
*[`common/include/api.h` line 262 at r6](https://reviewable.io/reviews/gramineproject/gramine/704#-N6EGkaC0XpGRv8IZW9H:-N6IYOPyCBhpx-aUwRpj:b-et3ed1) ([raw file](https://github.com/gramineproject/gramine/blob/8438324e09212850519174addb24f0f347f3f9db/common/include/api.h#L262)):*

The original description was from mbedtls - I did not change it. I agree, this makes no sense to me neither. Updated.
</blockquote></details>

Thanks, it reads much better now.

mkow (changes requested):
> Reviewed 3 of 3 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

___
*[`common/src/arch/x86_64/ct_memequal.nasm` line 30 at r8](https://reviewable.io/reviews/gramineproject/gramine/704#-N6J5ezg7SXHezHZlSCh:-N6J5ezg7SXHezHZlSCi:b-931dmy) ([raw file](https://github.com/gramineproject/gramine/blob/47c9fb7dd2fba60410f211da2faea6d43e807284/common/src/arch/x86_64/ct_memequal.nasm#L30)):*
> ```Assembly
>     mov     eax, ecx
>     test    al, al
>     sete    al
> ```

I think you can't leave the upper bytes of `rax` like this (see Sys-V x64 ABI).

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

___
*[`common/src/arch/x86_64/ct_memequal.nasm` line 30 at r8](https://reviewable.io/reviews/gramineproject/gramine/704#-N6J5ezg7SXHezHZlSCh:-N6JD2JH4BXeJEb71tGn:b2p5pq0) ([raw file](https://github.com/gramineproject/gramine/blob/47c9fb7dd2fba60410f211da2faea6d43e807284/common/src/arch/x86_64/ct_memequal.nasm#L30)):*

Oh, right, but that's pretty obscure. @kailun-qin: Could you explicitly state in a comment here that there's no need to zero the upper part of RAX, because ECX has already zeros there?
</blockquote></details>

But `rcx` could have anything in there and it wouldn't matter. It's `mov eax, xxx` which cleans upper part of `rax`. `rcx` is not changes (and can have garbage in upper bits).

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @kailun-qin)

___
*[`common/src/arch/x86_64/ct_memequal.nasm` line 30 at r8](https://reviewable.io/reviews/gramineproject/gramine/704#-N6J5ezg7SXHezHZlSCh:-N6JKFZX8S2tbM1iNrpH:b-su3j3e) ([raw file](https://github.com/gramineproject/gramine/blob/47c9fb7dd2fba60410f211da2faea6d43e807284/common/src/arch/x86_64/ct_memequal.nasm#L30)):*
@boryspoplawski: What I meant are the bytes of RAX above the lowest one. For them both the `mov` and `ecx` value matters.

> Yes, `mov` operates on a 32-bit operand automatically zeroes the upper 32 bits of the dest register.

Yes, of course I know this :P

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

___
*[`common/src/arch/x86_64/ct_memequal.nasm` line 30 at r8](https://reviewable.io/reviews/gramineproject/gramine/704#-N6J5ezg7SXHezHZlSCh:-N6JaQPN4vSi9sy86LyG:bjkjo7) ([raw file](https://github.com/gramineproject/gramine/blob/47c9fb7dd2fba60410f211da2faea6d43e807284/common/src/arch/x86_64/ct_memequal.nasm#L30)):*
Hmm, interesting. "3.1.2 Data Representation" says:

> Booleans, when stored in a memory object, are stored as single byte objects the
value of which is always 0 (`false`) or 1 (`true`). When stored in integer registers
(except for passing as arguments), all 8 bytes of the register are significant; any
nonzero value is considered `true`.

I guess that other section overrides this, but sounds like a really badly written documentation... Anyways, resolving.

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

___
*[`common/src/arch/x86_64/ct_memequal.nasm` line 30 at r8](https://reviewable.io/reviews/gramineproject/gramine/704#-N6J5ezg7SXHezHZlSCh:-N6JdV8364Pyf_F1hEn2:b-23fkdi) ([raw file](https://github.com/gramineproject/gramine/blob/47c9fb7dd2fba60410f211da2faea6d43e807284/common/src/arch/x86_64/ct_memequal.nasm#L30)):*

Hmm, interesting. "3.1.2 Data Representation" says:

> Booleans, when stored in a memory object, are stored as single byte objects the
value of which is always 0 (`false`) or 1 (`true`). When stored in integer registers
(except for passing as arguments), all 8 bytes of the register are significant; any
nonzero value is considered `true`.

I guess that other section overrides this, but sounds like a really badly written documentation... Anyways, resolving.
</blockquote></details>

The key here is "except for passing as arguments".

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

___
*[`common/src/arch/x86_64/ct_memequal.nasm` line 30 at r8](https://reviewable.io/reviews/gramineproject/gramine/704#-N6J5ezg7SXHezHZlSCh:-N6K91aB24XUnYuV9qBc:b-381qtk) ([raw file](https://github.com/gramineproject/gramine/blob/47c9fb7dd2fba60410f211da2faea6d43e807284/common/src/arch/x86_64/ct_memequal.nasm#L30)):*

The key here is "except for passing as arguments".
</blockquote></details>

This is a return value, not an argument...

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

___
*[`common/src/arch/x86_64/ct_memequal.nasm` line 30 at r8](https://reviewable.io/reviews/gramineproject/gramine/704#-N6J5ezg7SXHezHZlSCh:-N6KAUxZ8PV7EEQLSIaP:b-la5f4j) ([raw file](https://github.com/gramineproject/gramine/blob/47c9fb7dd2fba60410f211da2faea6d43e807284/common/src/arch/x86_64/ct_memequal.nasm#L30)):*

This is a return value, not an argument...
</blockquote></details>

Well, return value is just an out argument :)
But yes, that document is definitely imprecise in this sentence.

boryspoplawski:
Jenkins, test this please

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

___
*[`common/src/arch/x86_64/ct_memequal.nasm` line 30 at r8](https://reviewable.io/reviews/gramineproject/gramine/704#-N6J5ezg7SXHezHZlSCh:-N6KDyIT5hsztNFfQv8K:b21vxe9) ([raw file](https://github.com/gramineproject/gramine/blob/47c9fb7dd2fba60410f211da2faea6d43e807284/common/src/arch/x86_64/ct_memequal.nasm#L30)):*
> Would be if it was a ret-via-pointer, but not if it's just a register.

Why is that? If we had C# like `out` arguments and ABI saying first 2 are returned via `rax`,`rdx` I wouldn't see an issue with that.

dimakuv (approved):
> Reviewed 2 of 3 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

dimakuv:
Jenkins, retest this please (for some reason, all non-SGX pipelines failed with `error: PAL failed at ../pal/src/host/linux/db_main.c:214: setting personality failed: -6`)

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

dimakuv:
Jenkins, test this please

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: :shipit: complete! all files reviewed, all discussions resolved

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/704#-N6Npkj09ObHciTkLhi1:-N6Npkj09ObHciTkLhi2:b-g7rbzx) (no related file):*
We need to merge https://github.com/gramineproject/gramine/pull/736 first.

___

boryspoplawski (commented):
> Reviewed 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

dimakuv (dismissed):
> Reviewed 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

boryspoplawski:
Jenkins, test this please

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: :shipit: complete! all files reviewed, all discussions resolved

mkow (approved):
> Reviewed 2 of 2 files at r10, 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/704)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #614, opened 2022-05-27
[Feature request] Interface to get peak memory usage


kvinwang:
The memory is precious for a program running in SGX. We used to build our App with intel `sgx-sdk`, and use the `g_peak_heap_used` from the SDK to track and see if our app would be about to run out of memory.

It would be great if Gramine provide the equivalent functionality via adding an item in the `/proc/meminfo`:
```
MemPeak:         1187612 kB
```

boryspoplawski:
There is no such entry in `/proc/meminfo` (talking about normal Linux system)

kvinwang:
> There is no such entry in `/proc/meminfo` (talking about normal Linux system)

It can be put somewhere else which doesn't matter to me. Hmm, I am looking forward to the feature itself.

dimakuv:
@kvinwang I like this idea.

We have `Committed_AS` line in `/proc/meminfo` -- this line shows the *current* enclave-memory usage. Some references from the code, if you want to look into this deeper:
- https://github.com/gramineproject/gramine/blob/f17ca4525c38b880530597fae28fc43a0840b8a0/LibOS/shim/src/fs/proc/info.c#L49?plain=1
- https://github.com/gramineproject/gramine/blob/0affeefe376143614dd1122096ef8e79c782cc58/Pal/src/host/Linux-SGX/db_memory.c#L95?plain=1

However, you are right, there is no information about the *peak* enclave-memory usage. It could be easily added though, similarly to the already-existing `g_allocated_pages` variable.

@boryspoplawski @mkow Is this something you think is useful generally? Or is it a too obscure feature to add and promote to users?

mkow:
I don't like the idea of adding custom (i.e. not present on Linux) stuff to `/proc/meminfo`. Also, in this version the app itself would need to read it and reply to the developer? Why not just print this in the debug logs when Gramine exits?

dimakuv:
> Also, in this version the app itself would need to read it and reply to the developer?

That's how I understood @kvinwang's request -- every time the app reads `/proc/meminfo` (well, more specifically, when the app opens it), the `MemPeak` value is updated to reflect the last seen peak enclave memory. So the app must be "Gramine-aware".

> Why not just print this in the debug logs when Gramine exits?

Because the app wants to adjust itself based on this `MemPeak` value, e.g., perform some garbage collection or flush its caches. Something like this. Which needs a dynamically updated and accessible value.

boryspoplawski:
[...]

PR #714, opened 2022-07-01
[LibOS,PAL,Docs] Add peak memory usage support in procfs

kailun-qin:
The initial requirement was from https://github.com/gramineproject/gramine/issues/614.

An application may want to adjust itself based on the the peak (enclave) memory
usage in some scenarios, e.g., to perform some garbage collection or flush its
caches. This patch adds `VmPeak` stats into the pseudo-file `/proc/<pid>/status`
to support the usage.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/714)

### DIFF
--- a/Documentation/pal/host-abi.rst
+++ b/Documentation/pal/host-abi.rst
@@ -364,6 +364,9 @@ and to obtain an attestation report and quote.
 .. doxygenfunction:: PalMemoryAvailableQuota
    :project: pal
 
+.. doxygenfunction:: PalPeakMemoryUsage
+   :project: pal
+
 .. doxygenfunction:: PalCpuIdRetrieve
    :project: pal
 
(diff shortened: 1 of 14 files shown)


boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/714)* status: 0 of 14 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/714#-N5sz_EFEPQXGsmbl9o1:-N5tCvxJ2mwMgkXDR2rt:b-ber5m6) (no related file):*

I oppose the idea of introducing a new PAL function. Technically, the LibOS layer should bookkeep all memory (including PAL-internal memory) in its VMAs. So we don't need to "pierce through" to the PAL layer to calculate peak memory.

So, we should add the logic on memorizing and updating peak memory somewhere here, in the VMA subsystem: https://github.com/gramineproject/gramine/blob/master/libos/src/bookkeep/libos_vma.c

And remove *all* changes that you made to PALs in this PR.

@boryspoplawski What do you think?
</blockquote></details>

Yes, this is definitely job for LibOS. We might need to traverse whole vmas tree each time, but whatever.
Since that solution would be completely different from this PR, I suggest closing this one (instead of stashing changes here).

___

PR #723, opened 2022-07-04
Support `MSG_DONTWAIT` flag in send syscall

boryspoplawski:
## Description of the changes 
Additionally, this PR changes the socket level options are set.

### DIFF
--- a/libos/include/libos_socket.h
+++ b/libos/include/libos_socket.h
@@ -77,16 +77,19 @@ struct libos_sock_ops {
     /*!
      * \brief Send an array of buffers as continuous data.
      *
-     * \param      handle    A handle.
-     * \param      iov       An array of buffers to write from.
-     * \param      iov_len   The length of \p iov.
-     * \param[out] out_size  On success contains the number of bytes sent.
-     * \param      addr      An address to send to. May be NULL. It's up to the implementation to
-     *                       decide what to do with it (which might mean completely ignoring it).
-     * \param      addrlen   The length of \p addr.
+     * \param      handle             A handle.
+     * \param      iov                An array of buffers to write from.
+     * \param      iov_len            The length of \p iov.
+     * \param[out] out_size           On success contains the number of bytes sent.
+     * \param      addr               An address to send to. May be NULL. It's up to
+     *                                the implementation to decide what to do with it (which might
+     *                                mean completely ignoring it).
+     * \param      addrlen            The length of \p addr.
+     * \param      force_nonblocking  If `true` this request should not block. Otherwise just use
+     *                                whatever mode the handle is in.
      */
     int (*send)(struct libos_handle* handle, struct iovec* iov, size_t iov_len, size_t* out_size,
-                void* addr, size_t addrlen);
+                void* addr, size_t addrlen, bool force_nonblocking);
 
     /*!
      * \brief Receive continuous data into an array of buffers.
--- a/libos/src/net/ip.c
+++ b/libos/src/net/ip.c
@@ -303,27 +303,104 @@ static int set_socket_option(struct libos_handle* handle, int optname, void* opt
     }
     assert(attr.handle_type == PAL_TYPE_SOCKET);
 
-    /* All currently supported options use `int`. */
-    int val;
-    if (len < sizeof(val)) {
+    size_t required_len;
+    switch (optname) {
+        case SO_KEEPALIVE:
+            required_len = sizeof(int);
+            break;
+        case SO_LINGER:
+            required_len = sizeof(struct linger);
+            break;
+        case SO_RCVBUF:
+            required_len = sizeof(int);
+            break;
+        case SO_SNDBUF:
+            required_len = sizeof(int);
+            break;
+        case SO_RCVTIMEO:
+            required_len = sizeof(struct timeval);
+            break;
+        case SO_SNDTIMEO:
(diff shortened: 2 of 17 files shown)


boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/723)* status: 0 of 17 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

___
*[`libos/src/sys/libos_socket.c` line 1253 at r1](https://reviewable.io/reviews/gramineproject/gramine/723#-N68zhta1LfrAjTe8xGc:-N68zhta1LfrAjTe8xGd:b-clvhdk) ([raw file](https://github.com/gramineproject/gramine/blob/839ba4eef309350c41005e670562193a5f6d0767/libos/src/sys/libos_socket.c#L1253)):*
> ```C
>         return pal_to_unix_errno(ret);
>     }
>     assert(attr.handle_type == PAL_TYPE_SOCKET);
> ```

FYI: this was triggering on UNIX domain sockets (because they are implemented as PAL pipes).

dimakuv (changes requested):
> Reviewed 16 of 17 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/723)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski)

___
*[`libos/src/net/ip.c` line 413 at r2](https://reviewable.io/reviews/gramineproject/gramine/723#-N6CWqDvBBSw_0u6N7Ws:-N6CWqDvBBSw_0u6N7Wt:b-jdrtla) ([raw file](https://github.com/gramineproject/gramine/blob/db3d70adb61c7343cab692bbba88663c4c5ce8c6/libos/src/net/ip.c#L413)):*
> ```C
>     }
> 
>     /* Cache values in LibOS. */
> ```

Actually, what's the story here? Why do we cache only some values, but do not cache other values (e.g., we don't cache `SO_RCVBUF` but instead query it from the PAL)?
___
*[`libos/src/net/unix.c` line 334 at r2](https://reviewable.io/reviews/gramineproject/gramine/723#-N6CVZr81s3E1jS7M8oc:-N6CVZr95rpHqf-Jy7qe:b12gli1) ([raw file](https://github.com/gramineproject/gramine/blob/db3d70adb61c7343cab692bbba88663c4c5ce8c6/libos/src/net/unix.c#L334)):*
> ```C
>             /* `PalStreamWrite` has no way of making one-time nonblocking write, so we have no other
>              * option but to fail. */
>             return -EINVAL;
> ```

Oops. But this is a legit scenario, right? The app can have blocking UNIX domain socket, and then request `send(... MSG_DONTWAIT)` on it.

Linux will oblidge, but Gramine will fail in this case... Maybe the idea to emulate UNIX domain sockets over pipes wasn't that smart?

Anyway, could you put an XXX in the comment? This is unimplemented functionality, not just a comment.
___
*[`libos/src/sys/libos_socket.c` line 1253 at r1](https://reviewable.io/reviews/gramineproject/gramine/723#-N68zhta1LfrAjTe8xGc:-N6CXZRL9TzgtKZPWcGz:b-wjjm6w) ([raw file](https://github.com/gramineproject/gramine/blob/839ba4eef309350c41005e670562193a5f6d0767/libos/src/sys/libos_socket.c#L1253)):*

FYI: this was triggering on UNIX domain sockets (because they are implemented as PAL pipes).
</blockquote></details>

You mean if the app would do `setsockopt(unix_domain_sock_fd, SO_KEEPALIVE)`, then Gramine would fail? Bad. Looks like there are no such tests (I would expect LTP to have a test like this).

How did you trigger this bug then? Should we augment some of our tests to check that a snippet like I mentioned returns a proper error?
___
*[`libos/src/sys/libos_socket.c` line 621 at r2](https://reviewable.io/reviews/gramineproject/gramine/723#-N6CUGkN949zI0mnLh4W:-N6CUGkN949zI0mnLh4X:bxllsuu) ([raw file](https://github.com/gramineproject/gramine/blob/db3d70adb61c7343cab692bbba88663c4c5ce8c6/libos/src/sys/libos_socket.c#L621)):*
> ```C
> 
>     /* Note this only indicates whether this operation was requested to be nonblocking. If it's
>      * `false`, but the handle is in nonblocking mode, this read won't block. */
> ```

`read` -> `send`
___
*[`pal/src/host/linux-sgx/pal_pipes.c` line 464 at r2](https://reviewable.io/reviews/gramineproject/gramine/723#-N6CUs-C8vs3cUGRo8mb:-N6CUs-C8vs3cUGRo8mc:b-o9yv12) ([raw file](https://github.com/gramineproject/gramine/blob/db3d70adb61c7343cab692bbba88663c4c5ce8c6/pal/src/host/linux-sgx/pal_pipes.c#L464)):*
> ```C
>         return -PAL_ERROR_BADHANDLE;
> 
>     memset(attr, 0, sizeof(*attr));
> ```

Is it just for sanity?

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/723)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #456, opened 2022-03-14
[Pal,meson] Link Gramine patched mbedTLS statically

kailun-qin:
## How to test this PR? 

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/456)

PR #730, opened 2022-07-06
[mbedtls,CI-Examples,common] Upgrade Mbed TLS to version 2.28.0

kailun-qin:
Mbed TLS should be (periodically) updated to include security (also bug) fixes. The initial PR and discussions are from https://github.com/gramineproject/gramine/pull/426 and this patch proposes to upgrade it to 2.28.0 (LTS version) instead of 3.1.0 for smaller concern/controversy.


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/730)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/730#-N6IN9yF0oDpwit1Yes_:-N6IN9yH-_Y6rJnFned5:b-r2cjto) (no related file):*
IIRC, @woju was against a major update (from 2.26 to 3.1). @woju, should we go with this PR then and close #426?

___
___
*[`common/src/crypto/config.h` line 1 at r1](https://reviewable.io/reviews/gramineproject/gramine/730#-N6IOKRf9wUeJq_Ul588:-N6IOKRf9wUeJq_Ul589:bscm8t0) ([raw file](https://github.com/gramineproject/gramine/blob/363783115dcd9e74e7a3edb9b1ba9db3ca879286/common/src/crypto/config.h#L1)):*
> ```objc
> ```

This was a completely redundant file? I see that we already have an identical `subprojects/packagefiles/mbedtls/include/mbedtls/config-pal.h`
___
*[`subprojects/mbedtls-mbedtls-2.28.0.wrap` line 4 at r1](https://reviewable.io/reviews/gramineproject/gramine/730#-N6INcT12XL-HYz74tiB:-N6INcT12XL-HYz74tiC:b-9py6g4) ([raw file](https://github.com/gramineproject/gramine/blob/59e90caea87fb3466264cad770c5beeb0dfa6dcd/subprojects/mbedtls-mbedtls-2.28.0.wrap#L4)):*
> ```
> directory = mbedtls-mbedtls-2.28.0
> source_url = https://github.com/ARMmbed/mbedtls/archive/mbedtls-2.28.0.tar.gz
> source_fallback_url = https://packages.gramineproject.io/distfiles/mbedtls-2.28.0.tar.gz
> ```

@woju We'll need to upload this file to gramineproject.io
___
*[`subprojects/packagefiles/mbedtls/gramine.patch` line 138 at r1](https://reviewable.io/reviews/gramineproject/gramine/730#-N6INwWh938UKh4ipZE-:-N6INwWh938UKh4ipZE0:bpkz9kx) ([raw file](https://github.com/gramineproject/gramine/blob/59e90caea87fb3466264cad770c5beeb0dfa6dcd/subprojects/packagefiles/mbedtls/gramine.patch#L138)):*
> ```Diff
>  	echo "  LD    $@"
> -	$(CC) -shared -Wl,-soname,$@ -o $@ $(OBJS_TLS) -L. -lmbedx509 -lmbedcrypto $(LOCAL_LDFLAGS) $(LDFLAGS)
> +	$(CC) -shared -Wl,-soname,$@ -L. -lmbedcrypto$(SUFFIX) -lmbedx509$(SUFFIX) $(LOCAL_LDFLAGS) $(LDFLAGS) -o $@ $(OBJS_TLS)
> ```

Could you move `-o $@ $(OBJS_TLS)` to the same place as in the "replaced line"? This way we'll have uniformity.
___
*[`subprojects/packagefiles/mbedtls/gramine.patch` line 162 at r1](https://reviewable.io/reviews/gramineproject/gramine/730#-N6IO7T6E7r1e4ZTKiZc:-N6IO7T74CgWtb00dxfJ:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/59e90caea87fb3466264cad770c5beeb0dfa6dcd/subprojects/packagefiles/mbedtls/gramine.patch#L162)):*
> ```Diff
>  	echo "  LD    $@"
> -	$(CC) -shared -Wl,-soname,$@ -o $@ $(OBJS_X509) -L. -lmbedcrypto $(LOCAL_LDFLAGS) $(LDFLAGS)
> +	$(CC) -shared -Wl,-soname,$@ -L. -lmbedcrypto$(SUFFIX) $(LOCAL_LDFLAGS) $(LDFLAGS) -o $@ $(OBJS_X509)
> ```

ditto

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/730)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @kailun-qin)

___
*[`common/src/crypto/config.h` line 1 at r1](https://reviewable.io/reviews/gramineproject/gramine/730#-N6IOKRf9wUeJq_Ul588:-N6Iz89h9cisNkug2CXu:b-96wfu2) ([raw file](https://github.com/gramineproject/gramine/blob/363783115dcd9e74e7a3edb9b1ba9db3ca879286/common/src/crypto/config.h#L1)):*

This was a completely redundant file? I see that we already have an identical `subprojects/packagefiles/mbedtls/include/mbedtls/config-pal.h`
</blockquote></details>

It looks so.
___
*[`subprojects/packagefiles/mbedtls/gramine.patch` line 138 at r1](https://reviewable.io/reviews/gramineproject/gramine/730#-N6INwWh938UKh4ipZE-:-N6IzDZaCpAGDd0-kVGL:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/59e90caea87fb3466264cad770c5beeb0dfa6dcd/subprojects/packagefiles/mbedtls/gramine.patch#L138)):*

Could you move `-o $@ $(OBJS_TLS)` to the same place as in the "replaced line"? This way we'll have uniformity.
</blockquote></details>

Done.
___
*[`subprojects/packagefiles/mbedtls/gramine.patch` line 162 at r1](https://reviewable.io/reviews/gramineproject/gramine/730#-N6IO7T6E7r1e4ZTKiZc:-N6IzE1z9A3FzOAKtE1c:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/59e90caea87fb3466264cad770c5beeb0dfa6dcd/subprojects/packagefiles/mbedtls/gramine.patch#L162)):*

ditto
</blockquote></details>

Done.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/730)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/730#-N6IN9yF0oDpwit1Yes_:-N6JB3Gh9X3pYytr5N59:bxddfqh) (no related file):*

Yes, #426 should be delayed until 3.x version trickles down to distros. 2.28 is currently in most distros and it's LTS so it's OK to upgrade.
</blockquote></details>

What's the problem with upgrading now? We will never use distro versions anyway.

___

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/730)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/730#-N6IN9yF0oDpwit1Yes_:-N6KmTmq3pwkxxKfQX5_:b-e7r4mp) (no related file):*

What's the problem with upgrading now? We will never use distro versions anyway.
</blockquote></details>

For RATLS, we will use distro version. The problem is CMAC which is disabled by default and distros didn't enable it, but upstream enabled it in default config in 3.0, so it will be in distros from some 3.x version (probably 3.1, the first 3.x LTS).

___

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/730)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/730#-N6IN9yF0oDpwit1Yes_:-N6MbDRE0IJRJQq0JGDD:bpw8s2z) (no related file):*
> For RATLS, we will use distro version. 

Why? https://github.com/gramineproject/gramine/pull/456

___

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/730)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/730#-N6IN9yF0oDpwit1Yes_:-N6N3BC25cj2XYlGOghR:bsoilu0) (no related file):*

> For RATLS, we will use distro version. 

Why? https://github.com/gramineproject/gramine/pull/456
</blockquote></details>

I think @woju confused a bit: RA-TLS uses the static (our) version of the mbedTLS library.

But there are other things in Gramine that link against the shared version of mbedTLS:
1. Tools like `ias-request`, `pf_crypt`, etc -- they all use the "kitchen-sink" `libsgx_util.so` library (https://github.com/gramineproject/gramine/tree/master/tools/sgx/common) which in turns links against the shared mbedTLS.
2. Our CI-Examples: `ra-tls-mbedtls` and `ra-tls-secret-prov`. See e.g. https://github.com/gramineproject/gramine/blob/8c03dd46c5eb3c200fd0fc3ca3bc76ccd46f70aa/CI-Examples/ra-tls-secret-prov/Makefile#L54

So, the former (Gramine tools) can be rewritten to use a static (our) version of mbedTLS. In fact, I wanted to do it for a long time now.

However, the latter (CI-examples) should use the shared version of mbedTLS. Currently, we use our Gramine-built version of mbedTLS because distros don't have the CMAC-enabled version (which is required). So for these examples we'll want to use distro versions, which will become available in not-too-distant future.

___

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/730)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/730#-N6IN9yF0oDpwit1Yes_:-N6N8_118SF-zCgtbNHv:b-qh5t4l) (no related file):*

I think @woju confused a bit: RA-TLS uses the static (our) version of the mbedTLS library.

But there are other things in Gramine that link against the shared version of mbedTLS:
1. Tools like `ias-request`, `pf_crypt`, etc -- they all use the "kitchen-sink" `libsgx_util.so` library (https://github.com/gramineproject/gramine/tree/master/tools/sgx/common) which in turns links against the shared mbedTLS.
2. Our CI-Examples: `ra-tls-mbedtls` and `ra-tls-secret-prov`. See e.g. https://github.com/gramineproject/gramine/blob/8c03dd46c5eb3c200fd0fc3ca3bc76ccd46f70aa/CI-Examples/ra-tls-secret-prov/Makefile#L54

So, the former (Gramine tools) can be rewritten to use a static (our) version of mbedTLS. In fact, I wanted to do it for a long time now.

However, the latter (CI-examples) should use the shared version of mbedTLS. Currently, we use our Gramine-built version of mbedTLS because distros don't have the CMAC-enabled version (which is required). So for these examples we'll want to use distro versions, which will become available in not-too-distant future.
</blockquote></details>

Makes more sense.
Still, why we cannot upgrade to the newest version internally? When we migrate to distro version then either:
A) They do not support features we use - we cannot upgrade then.
B) They do support features we want - no problem, even if this would be version downgrade.

___

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/730)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/730#-N6IN9yF0oDpwit1Yes_:-N6NNtvV97GQeZW5sY2h:b-u5vpk7) (no related file):*
> I think @woju confused a bit: RA-TLS uses the static (our) version of the mbedTLS library.

It kinda does both, as seen in `CI-Examples/ra-tls-mbedtls`. The problem is, it has public API which references mbedtls structs (https://github.com/gramineproject/gramine/blob/8c03dd46c5eb3c200fd0fc3ca3bc76ccd46f70aa/tools/sgx/ra-tls/ra_tls.h#L94), so if user's app is dynamically linked against system's mbedtls, the versions need to match. So we need to either remove this API and leave only `_der` version, or dynamically link to system's version.

> why we cannot upgrade to the newest version internally?

We can. We would have to carry two mbedtls versions, one internal and one for ra-tls, but they have very different configs and patches, so it won't be as bad as it sounds.

___

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/730)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/730#-N6IN9yF0oDpwit1Yes_:-N6NSNDVEfBen-kWRljb:b5eo49b) (no related file):*

> We would have to carry two mbedtls versions, one internal and one for ra-tls

What, why? Why cannot `ra-tls` use our internal version until we migrate it to distro versions?
</blockquote></details>

At the time of migration we need to have a version that is not newer that what distro carries, otherwise users would have to downgrade mbedtls or build it themselves. So if we upgrade now, we'll need to wait until distros would pick whatever version we choose now. I think it's better to now move to 2.28 (which is currently in most distros), and wait until distros pick 3.1, then build mbedtls and libsgx_util against distro version.

___

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/730)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/730#-N6IN9yF0oDpwit1Yes_:-N6NTmRe0UXQmONhtN0i:b-hw3nb9) (no related file):*
> At the time of migration we need to have a version that is not newer that what distro carries, otherwise users would have to downgrade mbedtls or build it themselves. 

Why would they have to? They will have to recompile the example/ra-tls tool anyway

___

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/730)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/730#-N6IN9yF0oDpwit1Yes_:-N6NUx4-39irpQ-abf7i:bjjt5wg) (no related file):*

@woju raises a very good point. I regret having that `ra_tls_verify_callback(..., mbedtls_x509_crt* ...)` API.

@boryspoplawski Imagine that `libra-tls` uses our internal version and is linked to the user app. The user app then calls this function `ra_tls_verify_callback(..., mbedtls_x509_crt* app_x509_crt_ptr)`. If mbedTLS versions mismatch, then `libra-tls` fills the object with its own layout of `mbedtls_x509_crt*`, and the app reads this object with another layout of `mbedtls_x509_crt*`.
</blockquote></details>

As I said user will have to recompile their app.
How's the situation different if we use 2.28 or any other version? We will never have the exact same version with exact same config; there will always be difference. I don't see how user upgrading a version differs from downgrading here. They will have to recompile their app against the new version anyway.

___

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/730)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/730#-N6IN9yF0oDpwit1Yes_:-N6NYTNP7Rr0A4It5sw9:bjcrz01) (no related file):*

As I said user will have to recompile their app.
How's the situation different if we use 2.28 or any other version? We will never have the exact same version with exact same config; there will always be difference. I don't see how user upgrading a version differs from downgrading here. They will have to recompile their app against the new version anyway.
</blockquote></details>

I am confused by Borys's reply. We're probably talking about slightly different problems now.

My point is: I just realized that we cannot move our RA-TLS examples away from `mbedtls_gramine` version of the mbedTLS library because then there is no guarantee that the `mbedtls_x509_crt*` layout (used in a couple RA-TLS APIs) is compatible. Until we remove these ill-conceived RA-TLS APIs, we must dictate the exact mbedTLS version to be used with the apps that use these RA-TLS APIs (e.g. `CI-Examples/ra-tls-mbedtls`). Currently we dictate this exact mbedTLS version by linking against `mbedtls_gramine`.

___

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/730)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/730#-N6IN9yF0oDpwit1Yes_:-N6N_QHrCdtp97xOTeUU:b-43h3ta) (no related file):*

I am confused by Borys's reply. We're probably talking about slightly different problems now.

My point is: I just realized that we cannot move our RA-TLS examples away from `mbedtls_gramine` version of the mbedTLS library because then there is no guarantee that the `mbedtls_x509_crt*` layout (used in a couple RA-TLS APIs) is compatible. Until we remove these ill-conceived RA-TLS APIs, we must dictate the exact mbedTLS version to be used with the apps that use these RA-TLS APIs (e.g. `CI-Examples/ra-tls-mbedtls`). Currently we dictate this exact mbedTLS version by linking against `mbedtls_gramine`.
</blockquote></details>

So you are saying the user is supposed to link agains `ra-tls` library that we provide as a `.so`? Then from what you are saying it means user will always have to link against our library. Exact matching version of mbedtls might not be enough (e.g. config can probably influence this as well).
Also even if we somehow picked the exact same version and config as one of the distros, then other will faill (e.g. ubuntu 20.04 vs 22.04).
In any case, I don't see how not updating to the newest version (i.e. a version newer than what is/will be on some of supported distros) helps with anything.

___

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/730)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/730#-N6IN9yF0oDpwit1Yes_:-N6NnwaTC8WxU382Fwl0:bu1it7t) (no related file):*

> t seems that after we

Why after? At no point our newer version makes it impossible to migrate to distro version (after or before). It's always different from (some) distro version (+ possibly different configs), so it doesn't matter whether is newer or older.
</blockquote></details>

We can't migrate `CI-Examples/ra-tls-mbedtls` code to the distro version because the example's code uses `ra_tls_verify_callback(..., mbedtls_x509_crt* ...)`:

https://github.com/gramineproject/gramine/blob/2108aa8a35c1051620d350a60e79d731b6019a40/CI-Examples/ra-tls-mbedtls/src/client.c#L49

___

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/730)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/730#-N6IN9yF0oDpwit1Yes_:-N6NtxpIBxBM4EqOMimh:b-i049nx) (no related file):*

Looks like we are stuck in this conversation...

1. The example code of the client app in `ra-tls-mbedtls/` currently links against `mbedtls-gramine`, which is exactly the same version+config as Gramine's RA-TLS library uses.
2. If the example code of the client app in `ra-tls-mbedtls/` will link against a distro mbedTLS, then the version+config will differ from the one Gramine's RA-TLS library uses. Which may result in a subtle mismatch of object `mbedtls_x509_crt*`.
</blockquote></details>

Exactly. So it **does NOT** matter which version we use for `mbedtls-gramine`.

___

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/730)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/730#-N6IN9yF0oDpwit1Yes_:-N6NvT532EJhnjOlxMnb:brghvak) (no related file):*

Exactly. So it **does NOT** matter which version we use for `mbedtls-gramine`.
</blockquote></details>

Yup, I think Borys is right here, this whole idea of exposing `mbedtls_x509_crt*` from our .so (which requires custom mbedtls) is broken. So, there's no argument against updating mbedtls to the newest release, and moreover, in the meantime we should fix ra-tls API.

___

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/730)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/730#-N6IN9yF0oDpwit1Yes_:-N6PG4vi92_afb0z2Aa4:b-7kqodz) (no related file):*
> We can't migrate CI-Examples/ra-tls-mbedtls code to the distro version because the example's code uses `ra_tls_verify_callback(..., mbedtls_x509_crt* ...)`

We will be able to do it when distro's mbedtls ships CMAC. It will work if all the parties (distro, we and our users) use distro-provided compiler and the same `mbed*.so`, which is not that hard. If people use different compiler, they need to recompile both ratls and mbedtls anyway, so this case is not important here.

There's a viable migration path, but it depends on us not jumping the gun and not upgrading for the sake of upgrading:

The plan is roughly this:
1. we maintain our own, patched version of mbedtls_gramine.so; distro version doesn't have CMAC so it's not viable
2. new mbedtls version (3.x) appears in distros, it has CMAC so we could use it
3. we change to distro version
4. people recompile their apps against distro's `libmbed*.so`

There's an important requirement that at step 3. we have no later version that distros have. It's true that our users need to link against `libmbed*_gramine` (2.26 or 2.28 after this PR), so they will use whatever features are provided by our version. If we change to an older version, people will need to not just recompile, but polyfill their apps to make up for missing features while jumping to an older version, unless they want to recompile ratls themselves against newer mbedtls.

So the time between 2. and 3. is the time that it takes distros to ship as new version as we provide. The timelines are on the order of 2 years, if we support LTS, so we risk needing to support this version for two years more than needed. I want to avoid that, so I want to keep our version no more recent than distros have, to keep our options open. If we start upgrading like crazy, we condemn ourselves to support this mbedtls fork forever.

___

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/730)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/730#-N6IN9yF0oDpwit1Yes_:-N6POCHGBfypy0CiZiOJ:b-38djit) (no related file):*

> We can't migrate CI-Examples/ra-tls-mbedtls code to the distro version because the example's code uses `ra_tls_verify_callback(..., mbedtls_x509_crt* ...)`

We will be able to do it when distro's mbedtls ships CMAC. It will work if all the parties (distro, we and our users) use distro-provided compiler and the same `mbed*.so`, which is not that hard. If people use different compiler, they need to recompile both ratls and mbedtls anyway, so this case is not important here.

There's a viable migration path, but it depends on us not jumping the gun and not upgrading for the sake of upgrading:

The plan is roughly this:
1. we maintain our own, patched version of mbedtls_gramine.so; distro version doesn't have CMAC so it's not viable
2. new mbedtls version (3.x) appears in distros, it has CMAC so we could use it
3. we change to distro version
4. people recompile their apps against distro's `libmbed*.so`

There's an important requirement that at step 3. we have no later version that distros have. It's true that our users need to link against `libmbed*_gramine` (2.26 or 2.28 after this PR), so they will use whatever features are provided by our version. If we change to an older version, people will need to not just recompile, but polyfill their apps to make up for missing features while jumping to an older version, unless they want to recompile ratls themselves against newer mbedtls.

So the time between 2. and 3. is the time that it takes distros to ship as new version as we provide. The timelines are on the order of 2 years, if we support LTS, so we risk needing to support this version for two years more than needed. I want to avoid that, so I want to keep our version no more recent than distros have, to keep our options open. If we start upgrading like crazy, we condemn ourselves to support this mbedtls fork forever.
</blockquote></details>

Unless, of course, we get rid of `mbedtls_x509_crt*`, then we can statically link whatever we want and there's no problem.

___

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/730)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/730#-N6IN9yF0oDpwit1Yes_:-N6PSzQs4hK2rmhXvBPT:bkobtzr) (no related file):*

Unless, of course, we get rid of `mbedtls_x509_crt*`, then we can statically link whatever we want and there's no problem.
</blockquote></details>

On what you base this assumption that using the same or older version than distro is any different than using a higher version, w.r.t. binary compatibility?

___

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/730)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow)

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/730)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/730#-N6IN9yF0oDpwit1Yes_:-N6S8YxrDymqpx0w8u6i:b-d566la) (no related file):*

On what you base this assumption that using the same or older version than distro is any different than using a higher version, w.r.t. binary compatibility?
</blockquote></details>

I'm not talking about binary compatibility, I'm talking about new APIs, which become available in new version. If we roll back the versions, those APIs become unavailable.

___

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/730)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/730#-N6IN9yF0oDpwit1Yes_:-N6SGHZs32uWpQ-igPTa:b89vy1o) (no related file):*

I'm not talking about binary compatibility, I'm talking about new APIs, which become available in new version. If we roll back the versions, those APIs become unavailable.
</blockquote></details>

But everyone else is talking about binary compatibility, which is the core problem here? It makes the current ra-tls API broken regardless of specific versions we use anywhere (so, we can equally well use the newest release).

___

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/730)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/730#-N6IN9yF0oDpwit1Yes_:-N6XhXNOBxGUmCV3hzmG:b-4l75t1) (no related file):*

But everyone else is talking about binary compatibility, which is the core problem here? It makes the current ra-tls API broken regardless of specific versions we use anywhere (so, we can equally well use the newest release).
</blockquote></details>

I'm talking about source compatibility, which is the reason why we shouldn't push to the absolute newest version.

___

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/730)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/730#-N6IN9yF0oDpwit1Yes_:-N6XiwmrB1DB1t15uY1I:b-bzkw7w) (no related file):*

I'm talking about source compatibility, which is the reason why we shouldn't push to the absolute newest version.
</blockquote></details>

Where do we have source compatibility in our RA-TLS interface? We have a binary interface exposed from it, which erroneously exposes a structure which depends on mbedtls specific version and shouldn't be part of the public API. And this problem exists regardless of the mbedtls version we use in Gramine (becase the host/user can have a different one in their distro, breaking binary compatibility).

___

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/730)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/730#-N6IN9yF0oDpwit1Yes_:-N6XnmQV-YQnzZs0hblv:b-4w91hw) (no related file):*

Where do we have source compatibility in our RA-TLS interface? We have a binary interface exposed from it, which erroneously exposes a structure which depends on mbedtls specific version and shouldn't be part of the public API. And this problem exists regardless of the mbedtls version we use in Gramine (becase the host/user can have a different one in their distro, breaking binary compatibility).
</blockquote></details>

Yes, so because binary compatibility is broken between our and distro's version, then if you're using mbedtls not just for ratls, but also for something other (like maybe TLS to other endpoint? downloading models over HTTPS?), then for all your needs you need to target our version of mbedtls. IOW, people are dependent of the version we ship and use whatever APIs it provides. If we upgrade, new APIs might be available and people will be able to use them (IIUC this was motivation for original Kailun's PR -- the rationale given was a link to release notes with new features available in 3.x).

But, after 3.x gets available in distros, it doesn't need to be this way. Instead of providing mbedtls, we will be able to link to distro's version and users' apps will do the same. Yes, they will have to recompile. But, critically, if at the time of this transition, distro version is older than what we already shipped, then this transition will cause the very new APIs to become unavailable again. So they won't be able to just recompile if in the meantime they already rewritten their apps to use new APIs. This will turn into expectation that we continue to maintain our mbedtls to not break their apps, then to ship new version for security fixes and/or more features etc., forever.

So if we don't want to maintain mbedtls (and we don't want to do it, because it won't be needed), the solution is to prevent downstream from rewriting their apps against new mbedtls APIs. IOW, we keep shipping older, but still supported version (like 2.28 LTS) until we migrate to distro's version (then it becomes not our problem). Note this doesn't *break existing user's apps, just prevent them from upgrading to newer mbedtls, but the assumption is, if they exist, they are already working, so there's no problem. (If they don't exist, all this is moot).

___

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/730)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/730)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/730#-N6IN9yF0oDpwit1Yes_:-N6h35R39mvaA5Z1o8sh:bowj9qf) (no related file):*

> Yes, so because binary compatibility is broken between our and distro's version, then if you're using mbedtls not just for ratls, but also for something other (like maybe TLS to other endpoint? downloading models over HTTPS?), then for all your needs you need to target our version of mbedtls. IOW, people are dependent of the version we ship and use whatever APIs it provides. If we upgrade, new APIs might be available and people will be able to use them (IIUC this was motivation for original Kailun's PR -- the rationale given was a link to release notes with new features available in 3.x).

No. This mbedtls is not exposed to user apps, it's only used internally, that's the whole idea. And right now we're leaking a bit of it via this broken API, but it's only one structure and it's broken regardless of whether we use mbedtls 2 or 3 (and we plan to fix it).
</blockquote></details>

https://github.com/gramineproject/gramine/pull/740 -- and here is the PR that removes the broken APIs (two functions that use two mbedTLS structs).

___

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/730)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/730#-N6IN9yF0oDpwit1Yes_:-N7uI_775f6ZUbB-nVeK:b-irwwcf) (no related file):*

https://github.com/gramineproject/gramine/pull/740 -- and here is the PR that removes the broken APIs (two functions that use two mbedTLS structs).
</blockquote></details>

So #740 was merged.

I think we can rebase this PR to the latest master, and continue reviewing. Or if there are any other pre-requisites for this PR, please tell. @mkow @boryspoplawski @woju @kailun-qin 

___

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/730)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/730#-N6IN9yF0oDpwit1Yes_:-N7uK5Ka9H53Sboy8Mdn:bjeccs3) (no related file):*

So #740 was merged.

I think we can rebase this PR to the latest master, and continue reviewing. Or if there are any other pre-requisites for this PR, please tell. @mkow @boryspoplawski @woju @kailun-qin 
</blockquote></details>

Shouldn't we proceed with https://github.com/gramineproject/gramine/pull/426 instead? Wasn't #740 the only blocker for it?

___

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/730)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/730#-N6IN9yF0oDpwit1Yes_:-N7uLxUtEAaKXBzYed0o:b-8k8kw) (no related file):*

Shouldn't we proceed with https://github.com/gramineproject/gramine/pull/426 instead? Wasn't #740 the only blocker for it?
</blockquote></details>

Fair enough. I'd prefer #426, yep.

___

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/730)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/730#-N6IN9yF0oDpwit1Yes_:-N7vB_OG1F2ujrcb207V:b-qzp1i0) (no related file):*

As mbedtls released 3.2.1 recently, should we move forward w/ 3.2.1 instead of 3.1.0? If so, I'll make a new PR to proceed.
</blockquote></details>

The PR for upgrading to v3.2.1 is drafted at https://github.com/gramineproject/gramine/pull/796. PTAL, thanks!

___

dimakuv:
This PR was superseded by https://github.com/gramineproject/gramine/pull/796, which was merged now. Closing.

PR #738, opened 2022-07-11
[common] Don't extend files while reading data from protected files

oshogbo:
## Description of the changes 

The file should grow only during write and truncation. The protected file was extended when the application opened it for reading and writing and read from an offset larger than the current size. Previous logic, in the `ipf_update_after_seek` function, assumed that if the file is writable, the file should be extended. This function was called in the read and write routine of protected files.

New tests verify that the file is not extended after seek/tell routine, and they are not extended after the read syscall.

## How to test this PR? 

```
gramine-test -C libos/test/fs pytest
gramine-test --sgx -C libos/test/fs pytest
```

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/738)

### DIFF
--- a/common/src/protected_files/protected_files.c
+++ b/common/src/protected_files/protected_files.c
@@ -161,13 +161,28 @@ static bool ipf_init_fields(pf_context_t* pf) {
     ipf_init_root_mht(&pf->root_mht);
 
     pf->file           = NULL;
-    pf->end_of_file    = false;
     pf->need_writing   = false;
     pf->file_status    = PF_STATUS_UNINITIALIZED;
     pf->last_error     = PF_STATUS_SUCCESS;
     pf->real_file_size = 0;
 
     pf->cache = lruc_create();
+    return true;
+}
+
+static bool ipf_check_writable(pf_context_t* pf) {
+    if (PF_FAILURE(pf->file_status)) {
+        pf->last_error = pf->file_status;
+        DEBUG_PF("bad file status %d", pf->last_error);
+        return false;
+    }
+
+    if (!(pf->mode & PF_FILE_MODE_WRITE)) {
+        pf->last_error = PF_STATUS_INVALID_MODE;
+        DEBUG_PF("File is read-only");
+        return false;
+    }
+
     return true;
 }
 
@@ -654,32 +669,6 @@ static bool ipf_write_all_changes_to_disk(pf_context_t* pf) {
     return true;
 }
 
-// verify the seek value, and set an internal state of the structure accordingly
-// seek beyond the current size is supported if the file is writable,
-// the file is then extended with zeros (Intel SGX SDK implementation didn't support extending)
-static bool ipf_update_after_seek(pf_context_t* pf, uint64_t new_offset) {
-    if (PF_FAILURE(pf->file_status)) {
-        pf->last_error = pf->file_status;
-        return false;
-    }
-
-    bool result = false;
-
-    if (new_offset <= pf->encrypted_part_plain.size) {
-        result = true;
-    } else if (pf->mode & PF_FILE_MODE_WRITE) {
-        // need to extend the file
-        result = PF_SUCCESS(pf_set_size(pf, new_offset));
-    }
-
-    if (result)
-        pf->end_of_file = false;
-    else
-        pf->last_error = PF_STATUS_INVALID_PARAMETER;
-
-    return result;
-}
-
 static void ipf_try_clear_error(pf_context_t* pf) {
     if (pf->file_status == PF_STATUS_UNINITIALIZED ||
         pf->file_status == PF_STATUS_CRYPTO_ERROR ||
@@ -703,7 +692,6 @@ static void ipf_try_clear_error(pf_context_t* pf) {
 
     if (pf->file_status == PF_STATUS_SUCCESS) {
         pf->last_error = PF_STATUS_SUCCESS;
-        pf->end_of_file = false;
     }
 }
 
@@ -724,15 +712,7 @@ static size_t ipf_write(pf_context_t* pf, const void* ptr, uint64_t offset, size
 
     size_t data_left_to_write = size;
 
-    if (PF_FAILURE(pf->file_status)) {
-        pf->last_error = pf->file_status;
-        DEBUG_PF("bad file status %d", pf->last_error);
-        return 0;
-    }
-
-    if (!(pf->mode & PF_FILE_MODE_WRITE)) {
-        pf->last_error = PF_STATUS_INVALID_MODE;
-        DEBUG_PF("File is read-only");
+    if (!ipf_check_writable(pf)) {
         return 0;
     }
 
@@ -854,12 +834,6 @@ static size_t ipf_read(pf_context_t* pf, void* ptr, uint64_t offset, size_t size
         offset += size_to_read;
         out_buffer += size_to_read;
         data_left_to_read -= size_to_read;
-    }
-
-    if (data_left_to_read == 0 && data_attempted_to_read != size) {
-        // user wanted to read more and we had to shrink the request
-        assert(offset == pf->encrypted_part_plain.size);
-        pf->end_of_file = true;
     }
 
     return data_attempted_to_read - data_left_to_read;
@@ -1276,7 +1250,6 @@ pf_status_t pf_set_size(pf_context_t* pf, uint64_t size) {
         ipf_init_root_mht(&pf->root_mht);
 
         pf->need_writing = true;
-        pf->end_of_file  = false;
         pf->real_file_size = 0;
 
         while ((data = lruc_get_last(pf->cache)) != NULL) {
@@ -1322,11 +1295,12 @@ pf_status_t pf_read(pf_context_t* pf, uint64_t offset, size_t size, void* output
         return PF_STATUS_SUCCESS;
     }
 
-    if (!ipf_update_after_seek(pf, offset))
+    if (PF_FAILURE(pf->file_status)) {
+        pf->last_error = pf->file_status;
         return pf->last_error;
+    }
 
-    if (pf->end_of_file || offset == pf->encrypted_part_plain.size) {
-        pf->end_of_file = true;
+    if (offset >= pf->encrypted_part_plain.size) {
         *bytes_read = 0;
         return PF_STATUS_SUCCESS;
     }
@@ -1343,11 +1317,20 @@ pf_status_t pf_write(pf_context_t* pf, uint64_t offset, size_t size, const void*
     if (!g_initialized)
         return PF_STATUS_UNINITIALIZED;
 
-    if (!ipf_update_after_seek(pf, offset))
+    if (!ipf_check_writable(pf))
         return pf->last_error;
+
+    if (offset > pf->encrypted_part_plain.size) {
+        pf_status_t ret = pf_set_size(pf, offset);
+        if (!PF_SUCCESS(ret)) {
+            pf->last_error = ret;
+            return pf->last_error;
+        }
+    }
 
     if (ipf_write(pf, input, offset, size) != size)
         return pf->last_error;
+
     return PF_STATUS_SUCCESS;
 }
 
--- a/common/src/protected_files/protected_files_internal.h
+++ b/common/src/protected_files/protected_files_internal.h
@@ -20,7 +20,6 @@ struct pf_context {
     file_node_t root_mht; // the root of the mht is always needed (for files bigger than 3KB)
     pf_handle_t file;
     pf_file_mode_t mode;
-    bool end_of_file;
     uint64_t real_file_size;
     bool need_writing;
     pf_status_t file_status;
@@ -62,5 +61,4 @@ static pf_context_t* ipf_open(const char* path, pf_file_mode_t mode, bool create
 static bool ipf_close(pf_context_t* pf);
 static size_t ipf_read(pf_context_t* pf, void* ptr, uint64_t offset, size_t size);
 static size_t ipf_write(pf_context_t* pf, const void* ptr, uint64_t offset, size_t size);
-static bool ipf_update_after_seek(pf_context_t* pf, uint64_t new_offset);
 static void ipf_try_clear_error(pf_context_t* pf);
--- a/libos/test/fs/seek_tell.c
+++ b/libos/test/fs/seek_tell.c
@@ -63,6 +63,16 @@ static void seek_output_fd(const char* path) {
     }
     pos = tell_fd(path, f);
     printf("tell(%s) output end 2 OK: %zd\n", path, pos);
+    seek_fd(path, f, pos * 2, SEEK_SET);
+    printf("seek(%s) output beyond end OK\n", path);
+    pos = tell_fd(path, f);
+    printf("tell(%s) output beyond end OK: %zd\n", path, pos);
+    ssize_t ret = read(f, &buf, 1);
+    printf("read(%s) output OK: %zd\n", path, ret);
+    seek_fd(path, f, 0, SEEK_END);
+    printf("seek(%s) output end 4 OK\n", path);
+    pos = tell_fd(path, f);
+    printf("tell(%s) output end 3 OK: %zd\n", path, pos);
     close_fd(path, f);
     printf("close(%s) output OK\n", path);
 }
@@ -94,6 +104,16 @@ static void seek_output_stdio(const char* path) {
     }
     pos = tell_stdio(path, f);
     printf("ftell(%s) output end 2 OK: %zd\n", path, pos);
+    seek_stdio(path, f, pos * 2, SEEK_SET);
+    printf("fseek(%s) output beyond end OK\n", path);
+    pos = tell_stdio(path, f);
+    printf("ftell(%s) output beyond end OK: %zd\n", path, pos);
+    ssize_t ret = fread(&buf, 1, 1, f);
+    printf("fread(%s) output OK: %zd\n", path, ret);
+    seek_stdio(path, f, 0, SEEK_END);
+    printf("fseek(%s) output end 4 OK\n", path);
+    pos = tell_stdio(path, f);
+    printf("ftell(%s) output end 3 OK: %zd\n", path, pos);
     close_stdio(path, f);
     printf("fclose(%s) output OK\n", path);
 }
--- a/libos/test/fs/test_fs.py
+++ b/libos/test/fs/test_fs.py
@@ -99,6 +99,8 @@ class TC_00_FileSystem(RegressionTestCase):
     # pylint: disable=too-many-arguments
     def verify_seek_tell(self, stdout, stderr, input_path, output_path_1, output_path_2, size):
         self.assertNotIn('ERROR: ', stderr)
+
+        # seek_input_fd:
         self.assertIn('open(' + input_path + ') input OK', stdout)
         self.assertIn('seek(' + input_path + ') input start OK', stdout)
         self.assertIn('seek(' + input_path + ') input end OK', stdout)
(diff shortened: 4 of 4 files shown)


mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/738)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @oshogbo)

___
*[`-- commits` line 13 at r1](https://reviewable.io/reviews/gramineproject/gramine/738#-N6h1IiX-QNursYOZbRZ):*




_Suggestion:_
```
and that they are not
```
___
*[`common/src/protected_files/protected_files.c` line 1337 at r1](https://reviewable.io/reviews/gramineproject/gramine/738#-N6h4zOOAXlmSPwThaw0:-N6h4zOOAXlmSPwThaw1:b-dfuv92) ([raw file](https://github.com/gramineproject/gramine/blob/3e60bcbba22f7469baf8b5e7bd20c2fdc05fa6c6/common/src/protected_files/protected_files.c#L1337)):*
> ```C
>             return pf->last_error;
>         }
>         pf->end_of_file = false;
> ```

Shouldn't this be `true`?
Actually, do we even still need `end_of_file`? There's only one place reading it, and it seems that it's redundant there.
___
*[`libos/test/fs/seek_tell.c` line 69 at r1](https://reviewable.io/reviews/gramineproject/gramine/738#-N6h1r7yB6qbGpkcxnl7:-N6h1r7yB6qbGpkcxnl8:bj90qn) ([raw file](https://github.com/gramineproject/gramine/blob/3e60bcbba22f7469baf8b5e7bd20c2fdc05fa6c6/libos/test/fs/seek_tell.c#L69)):*
> ```C
>     printf("seek(%s) output end 4 OK\n", path);
>     pos = tell_fd(path, f);
>     printf("tell(%s) output end 3 OK: %zd\n", path, pos);
> ```

Why is it 4 and then 3? Also, what do you mean by "output end"? You're seeking beyond it, not just to the end.



_Code quote:_
```C
    printf("seek(%s) output end 4 OK\n", path);
    pos = tell_fd(path, f);
    printf("tell(%s) output end 3 OK: %zd\n", path, pos);
```
___
*[`libos/test/fs/seek_tell.c` line 75 at r1](https://reviewable.io/reviews/gramineproject/gramine/738#-N6h2Tgb8YXOivt2rTcT:-N6h2Tgb8YXOivt2rTcU:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/3e60bcbba22f7469baf8b5e7bd20c2fdc05fa6c6/libos/test/fs/seek_tell.c#L75)):*
> ```C
>     printf("seek(%s) output end 5 OK\n", path);
>     pos = tell_fd(path, f);
>     printf("tell(%s) output end 4 OK: %zd\n", path, pos);
> ```

ditto

dimakuv (changes requested):
> Reviewed 3 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/738)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @mkow and @oshogbo)

___
*[`common/src/protected_files/protected_files.c` line 1334 at r1](https://reviewable.io/reviews/gramineproject/gramine/738#-N6h6KHC5QbhCzF82cq2:-N6h6KHC5QbhCzF82cq3:b-bf543s) ([raw file](https://github.com/gramineproject/gramine/blob/3e60bcbba22f7469baf8b5e7bd20c2fdc05fa6c6/common/src/protected_files/protected_files.c#L1334)):*
> ```C
>     if (offset > pf->encrypted_part_plain.size) {
>         if (!PF_SUCCESS(pf_set_size(pf, offset))) {
>             pf->last_error = PF_STATUS_INVALID_PARAMETER;
> ```

Wouldn't it make more sense to memorize the return value of `pf_set_size()` and assign `pf->last_error` to this return value?
___
*[`common/src/protected_files/protected_files.c` line 1337 at r1](https://reviewable.io/reviews/gramineproject/gramine/738#-N6h4zOOAXlmSPwThaw0:-N6h6XpLDb1AACV6dk-a:b-wzyzjx) ([raw file](https://github.com/gramineproject/gramine/blob/3e60bcbba22f7469baf8b5e7bd20c2fdc05fa6c6/common/src/protected_files/protected_files.c#L1337)):*

Shouldn't this be `true`?
Actually, do we even still need `end_of_file`? There's only one place reading it, and it seems that it's redundant there.
</blockquote></details>

+1, `end_of_file` seems to be pointless.
___
*[`libos/test/fs/seek_tell.c` line 116 at r1](https://reviewable.io/reviews/gramineproject/gramine/738#-N6h8d_B1w7yV7To9A9J:-N6h8d_B1w7yV7To9A9K:b-wvx4gt) ([raw file](https://github.com/gramineproject/gramine/blob/3e60bcbba22f7469baf8b5e7bd20c2fdc05fa6c6/libos/test/fs/seek_tell.c#L116)):*
> ```C
>     printf("fseek(%s) output end 5 OK\n", path);
>     pos = tell_stdio(path, f);
>     printf("ftell(%s) output end 4 OK: %zd\n", path, pos);
> ```

Hm, probably @boryspoplawski will want you to rewrite this test to be self-checking. But maybe not -- this is the old test being modifying, not a brand new test...
___
*[`libos/test/fs/test_fs.py` line 133 at r1](https://reviewable.io/reviews/gramineproject/gramine/738#-N6h7nim5QsTXwvMyawQ:-N6h7nim5QsTXwvMyawR:b-7kkb3z) ([raw file](https://github.com/gramineproject/gramine/blob/3e60bcbba22f7469baf8b5e7bd20c2fdc05fa6c6/libos/test/fs/test_fs.py#L133)):*
> ```Python
>         self.assertIn('read(' + output_path_2 + ') output OK: 0', stdout)
>         self.assertIn('seek(' + output_path_2 + ') output end 5 OK', stdout)
>         self.assertIn('tell(' + output_path_2 + ') output end 4 OK: ' + str(size + 4098), stdout)
> ```

What does it mean? Why does it work this way?

You first perform a `seek((size + 4098) * 2)` -- which works fine. Then you perform a `seek(SEEK_END)`, and it returns `size + 4098`. So the update to max size of the file is lost -- is this how POSIX dictates it to work? Like, if there was no write, then no matter how far you seeked, the real end of the file is at the point of the last write?

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/738)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @oshogbo)

___
*[`libos/test/fs/seek_tell.c` line 116 at r1](https://reviewable.io/reviews/gramineproject/gramine/738#-N6h8d_B1w7yV7To9A9J:-N6h9xvTA-13VYhYxPXH:b-5i1k42) ([raw file](https://github.com/gramineproject/gramine/blob/3e60bcbba22f7469baf8b5e7bd20c2fdc05fa6c6/libos/test/fs/seek_tell.c#L116)):*

Hm, probably @boryspoplawski will want you to rewrite this test to be self-checking. But maybe not -- this is the old test being modifying, not a brand new test...
</blockquote></details>

I would prefer that, but won't insist since, as you noted, this is old test being extended (and it does not require a complete rewrite).

oshogbo (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/738)* status: 0 of 4 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "squash! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`-- commits` line 13 at r1](https://reviewable.io/reviews/gramineproject/gramine/738#-N6h1IiX-QNursYOZbRZ):*
Done.
___
*[`common/src/protected_files/protected_files.c` line 1334 at r1](https://reviewable.io/reviews/gramineproject/gramine/738#-N6h6KHC5QbhCzF82cq2:-N6hO_J-BwODWxc2UixP:b-pfhe0c) ([raw file](https://github.com/gramineproject/gramine/blob/3e60bcbba22f7469baf8b5e7bd20c2fdc05fa6c6/common/src/protected_files/protected_files.c#L1334)):*

Wouldn't it make more sense to memorize the return value of `pf_set_size()` and assign `pf->last_error` to this return value?
</blockquote></details>

So that was my initial idea, but I'm wasn't sure about that because the previous code was setting this value.

So if there isn't any downsides I will change that.
___
*[`common/src/protected_files/protected_files.c` line 1337 at r1](https://reviewable.io/reviews/gramineproject/gramine/738#-N6h4zOOAXlmSPwThaw0:-N6hOBxD1QT083XZtYB4:b2htuvm) ([raw file](https://github.com/gramineproject/gramine/blob/3e60bcbba22f7469baf8b5e7bd20c2fdc05fa6c6/common/src/protected_files/protected_files.c#L1337)):*

+1, `end_of_file` seems to be pointless.
</blockquote></details>

In the previous code:
```
result = PF_SUCCESS(pf_set_size(pf, new_offset));
```
then
```
if (result)
    pf->end_of_file = false;
```

But now when I think about it you are right, it doesn't make sense.
I will remove `end_of_file` completely.
___
*[`libos/test/fs/seek_tell.c` line 69 at r1](https://reviewable.io/reviews/gramineproject/gramine/738#-N6h1r7yB6qbGpkcxnl7:-N6hP6iN5Cs7aQgCyZ4e:bffcmgf) ([raw file](https://github.com/gramineproject/gramine/blob/3e60bcbba22f7469baf8b5e7bd20c2fdc05fa6c6/libos/test/fs/seek_tell.c#L69)):*

Why is it 4 and then 3? Also, what do you mean by "output end"? You're seeking beyond it, not just to the end.
</blockquote></details>

It's the number of `tell`, and then numbers of `seek`.
So this is 4th time that the `seek` was called, in this test,and 3rd time `tell`.
As pointed out by @dimakuv in other PR we have to distinguish the operation otherwise the ` self.assertIn` will check only the first occurrence.
This is done in this test already for example:
```
    printf("seek(%s) output end 2 OK\n", path);
    write_fd(path, f, buf, 1);
    seek_fd(path, f, -EXTEND_SIZE - 1, SEEK_CUR); // rewind to former end
    printf("seek(%s) output end 3 OK\n", path);
```
I also think it's messy. So I will create another PR which changes this in whole file to notation:
```
1st tell...
2nd seek...
```
And so on.

The `output`, if I understand correctly is the type of file. You can see that there are `input` tests and `output` tests here.
Ou, in case of "end" I assumed this is just saying "end OK", not that it seeking to "end" I will fix that.
___
*[`libos/test/fs/seek_tell.c` line 75 at r1](https://reviewable.io/reviews/gramineproject/gramine/738#-N6h2Tgb8YXOivt2rTcT:-N6haBLaC4n17V9WyUz8:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/3e60bcbba22f7469baf8b5e7bd20c2fdc05fa6c6/libos/test/fs/seek_tell.c#L75)):*

ditto
</blockquote></details>

Done.
___
*[`libos/test/fs/seek_tell.c` line 116 at r1](https://reviewable.io/reviews/gramineproject/gramine/738#-N6h8d_B1w7yV7To9A9J:-N6hP-A09dHrRykuG2W6:bs1nd7p) ([raw file](https://github.com/gramineproject/gramine/blob/3e60bcbba22f7469baf8b5e7bd20c2fdc05fa6c6/libos/test/fs/seek_tell.c#L116)):*

I would prefer that, but won't insist since, as you noted, this is old test being extended (and it does not require a complete rewrite).
</blockquote></details>

ACK. :)
___
*[`libos/test/fs/test_fs.py` line 133 at r1](https://reviewable.io/reviews/gramineproject/gramine/738#-N6h7nim5QsTXwvMyawQ:-N6hQ6M84SsgotxJ0qR6:b-br1ipi) ([raw file](https://github.com/gramineproject/gramine/blob/3e60bcbba22f7469baf8b5e7bd20c2fdc05fa6c6/libos/test/fs/test_fs.py#L133)):*

What does it mean? Why does it work this way?

You first perform a `seek((size + 4098) * 2)` -- which works fine. Then you perform a `seek(SEEK_END)`, and it returns `size + 4098`. So the update to max size of the file is lost -- is this how POSIX dictates it to work? Like, if there was no write, then no matter how far you seeked, the real end of the file is at the point of the last write?
</blockquote></details>

I'm not sure what you mean the "the update to max size of the file is lost" means.

The seek itself doesn't change the size of the file.
Dunno if this is POSIX, this is how Linux works (tested), and this is documented in man page:
```
lseek() allows the file offset to be set beyond the end of the file (but this does not change the size of the file). 
```

oshogbo (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/738)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "squash! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`libos/test/fs/seek_tell.c` line 75 at r2](https://reviewable.io/reviews/gramineproject/gramine/738#-N6i0RwqA-BW8OPdsvfV:-N6iMls96V-lwv2Hpb5H:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/d48c62b72d1fc6f8ded97639c7ac63224f52f79f/libos/test/fs/seek_tell.c#L75)):*

off-by-one? (you already used "2" above)
</blockquote></details>

Done.
___
*[`libos/test/fs/seek_tell.c` line 116 at r2](https://reviewable.io/reviews/gramineproject/gramine/738#-N6i0Y-P5-lL3XYUPB9l:-N6iMlb8BtVEsbsMjVlH:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/d48c62b72d1fc6f8ded97639c7ac63224f52f79f/libos/test/fs/seek_tell.c#L116)):*

ditto
</blockquote></details>

Done.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/738)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/738)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

oshogbo (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/738)* status: 3 of 4 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "squash! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

___
*[`-- commits` line 2 at r4](https://reviewable.io/reviews/gramineproject/gramine/738#-N6mxYiS-aE_MVrRIHVD):*

One liner lacks info that this is about PF
</blockquote></details>

Done.
___
*[`common/src/protected_files/protected_files.c` line 1299 at r4](https://reviewable.io/reviews/gramineproject/gramine/738#-N6n-Pa73vhRN4Uv4TIC:-N6r2pLDBor0T_qr0GyA:bcjjxqc) ([raw file](https://github.com/gramineproject/gramine/blob/b5eac6892db82717af3a9e502a5f1fcf13782915/common/src/protected_files/protected_files.c#L1299)):*
> pf->last_error = pf->file_status
Nice catch.
___
*[`common/src/protected_files/protected_files_internal.h` line 64 at r4](https://reviewable.io/reviews/gramineproject/gramine/738#-N6n01ni1lJbGA2r_m8_:-N6r2C2H5xw0iv8EpKQh:bfw5h6h) ([raw file](https://github.com/gramineproject/gramine/blob/b5eac6892db82717af3a9e502a5f1fcf13782915/common/src/protected_files/protected_files_internal.h#L64)):*

Wait, what is this? Why do we have static noinline functions in header? How is this supposed to work?
Also this one is used only in one file.
</blockquote></details>

I guess the idea was to define all the static functions in one place, include them and don't care about the order of definition. 

I'm not sure why you suggesting inlinie functions here.

But yea this file is weird, and kinda useless.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/738)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "squash! " found in commit messages' one-liners (waiting on @oshogbo)

___
*[`common/src/protected_files/protected_files_internal.h` line 64 at r4](https://reviewable.io/reviews/gramineproject/gramine/738#-N6n01ni1lJbGA2r_m8_:-N6rMSdz6pky_8wPdmQ3:bfqi67h) ([raw file](https://github.com/gramineproject/gramine/blob/b5eac6892db82717af3a9e502a5f1fcf13782915/common/src/protected_files/protected_files_internal.h#L64)):*

I guess the idea was to define all the static functions in one place, include them and don't care about the order of definition. 

I'm not sure why you suggesting inlinie functions here.

But yea this file is weird, and kinda useless.
</blockquote></details>

Even if you don't remove all of this, then please don't add new functions here, it makes no sense.

oshogbo (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/738)* status: 2 of 4 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

___
*[`common/src/protected_files/protected_files_internal.h` line 64 at r4](https://reviewable.io/reviews/gramineproject/gramine/738#-N6n01ni1lJbGA2r_m8_:-N6s7YHj22NPewoTPjwi:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/b5eac6892db82717af3a9e502a5f1fcf13782915/common/src/protected_files/protected_files_internal.h#L64)):*

Even if you don't remove all of this, then please don't add new functions here, it makes no sense.
</blockquote></details>

Done.

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/738)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/738)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/738)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/738)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/738)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #744, opened 2022-07-11
[LibOS,PAL] Remove `struct atomic_int`

boryspoplawski:
## Description of the changes 
We don't need custom atomics.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/744)


dimakuv (dismissed):
> Reviewed 11 of 11 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/744)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

mkow (changes requested):
> Reviewed 11 of 11 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/744)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required) (waiting on @boryspoplawski)

___
*[`libos/include/libos_internal.h` line 169 at r1](https://reviewable.io/reviews/gramineproject/gramine/744#-N6lxlUA8x5EHMClJDxL:-N6lxlUB1jknDQZrD71M:byivuk7) ([raw file](https://github.com/gramineproject/gramine/blob/1d1011c66d5effca97666ab940bc4a5a62f4e468/libos/include/libos_internal.h#L169)):*
> ```objc
>         _c = __atomic_load_n(ref, __ATOMIC_SEQ_CST);
>         assert(_c >= 0);
>     } while (!__atomic_compare_exchange_n(ref, &_c, _c + 1, /*weak=*/false,
> ```

You'll have memory corruption here if `REFTYPE` diverges from `int64_t`. I think `_c` and return type should be `REFTYPE` to prevent this?
___
*[`libos/include/libos_internal.h` line 176 at r1](https://reviewable.io/reviews/gramineproject/gramine/744#-N6m-AeW-mG-QCBroVu0:-N6m-AeW-mG-QCBroVu1:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/1d1011c66d5effca97666ab940bc4a5a62f4e468/libos/include/libos_internal.h#L176)):*
> ```objc
> #define REF_INC(ref) __ref_inc(&(ref))
> 
> static inline int64_t __ref_dec(REFTYPE* ref) {
> ```

ditto
___
*[`pal/src/host/linux-sgx/enclave_pages.c` line 10 at r1](https://reviewable.io/reviews/gramineproject/gramine/744#-N6mYT5pCicf5HoZyNMR:-N6mYT5pCicf5HoZyNMS:b-1742k8) ([raw file](https://github.com/gramineproject/gramine/blob/1d1011c66d5effca97666ab940bc4a5a62f4e468/pal/src/host/linux-sgx/enclave_pages.c#L10)):*
> ```C
> #include "spinlock.h"
> 
> int64_t g_allocated_pages;
> ```

I guess this should actually be `size_t`? (and was `int64_t` only because of old atomics' inflexibility?)
___
*[`pal/src/host/linux-sgx/pal_memory.c` line 16 at r1](https://reviewable.io/reviews/gramineproject/gramine/744#-N6mYbAcEiG8Od6jy3C3:-N6mYbAcEiG8Od6jy3C4:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/1d1011c66d5effca97666ab940bc4a5a62f4e468/pal/src/host/linux-sgx/pal_memory.c#L16)):*
> ```C
> #include "pal_linux.h"
> 
> extern int64_t g_allocated_pages;
> ```

ditto
___
*[`pal/src/host/linux-sgx/pal_memory.c` line 83 at r1](https://reviewable.io/reviews/gramineproject/gramine/744#-N6mZi2R-1jiAgwJB04e:-N6mZi2R-1jiAgwJB04f:bpjkl5p) ([raw file](https://github.com/gramineproject/gramine/blob/1d1011c66d5effca97666ab940bc4a5a62f4e468/pal/src/host/linux-sgx/pal_memory.c#L83)):*
> ```C
> #endif
> 
>     static int64_t at_cnt = 0;
> ```

Just `int`?
___
*[`pal/src/host/linux-sgx/pal_threading.c` line 38 at r1](https://reviewable.io/reviews/gramineproject/gramine/744#-N6mZoqI2Oj0LIJhQFxu:-N6mZoqI2Oj0LIJhQFxv:b-mox4zf) ([raw file](https://github.com/gramineproject/gramine/blob/1d1011c66d5effca97666ab940bc4a5a62f4e468/pal/src/host/linux-sgx/pal_threading.c#L38)):*
> ```C
> static PAL_IDX pal_assign_tid(void) {
>     /* tid 1 is assigned to the first thread; see pal_linux_main() */
>     static int64_t tid = 1;
> ```

unsigned?

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/744)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required) (waiting on @boryspoplawski and @mkow)

___
*[`libos/include/libos_internal.h` line 169 at r1](https://reviewable.io/reviews/gramineproject/gramine/744#-N6lxlUA8x5EHMClJDxL:-N6mwvfy1fW0W6XLkpqp:b-zef9uj) ([raw file](https://github.com/gramineproject/gramine/blob/1d1011c66d5effca97666ab940bc4a5a62f4e468/libos/include/libos_internal.h#L169)):*

You'll have memory corruption here if `REFTYPE` diverges from `int64_t`. I think `_c` and return type should be `REFTYPE` to prevent this?
</blockquote></details>

1. Why? Looks correct: https://godbolt.org/z/e85hKcbzh
2. I didn't want to change anything in this PR, just remove this struct.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/744)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required) (waiting on @boryspoplawski)

___
*[`libos/include/libos_internal.h` line 169 at r1](https://reviewable.io/reviews/gramineproject/gramine/744#-N6lxlUA8x5EHMClJDxL:-N6nZBZm95IxYTxG_Fym:b-90pqxd) ([raw file](https://github.com/gramineproject/gramine/blob/1d1011c66d5effca97666ab940bc4a5a62f4e468/libos/include/libos_internal.h#L169)):*

1. Why? Looks correct: https://godbolt.org/z/e85hKcbzh
2. I didn't want to change anything in this PR, just remove this struct.
</blockquote></details>

1. Hmm, maybe it was another way around... Anyways, these atomic builtins doesn't check types, so it looks risky (and you remember that repro case we had in the past for GCC builtins not checking argument sizes). I don't like such implicit assumption that two types must be equal.
2. But this change is trivial? And it has been made a bit worse in this PR, because now `REFTYPE` is just a typedef to an integer type, which suggests that other types than currently used may also be valid for it (otherwise you'd just use the only appropriate type directly).

So, overall, I'd prefer to fix such places now.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/744)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required) (waiting on @boryspoplawski and @mkow)

___
*[`libos/include/libos_internal.h` line 169 at r1](https://reviewable.io/reviews/gramineproject/gramine/744#-N6lxlUA8x5EHMClJDxL:-N6n_aBC4XFj2JOxuZPw:bx3okbf) ([raw file](https://github.com/gramineproject/gramine/blob/1d1011c66d5effca97666ab940bc4a5a62f4e468/libos/include/libos_internal.h#L169)):*

1. No, actually I've never seen a repro, just heard you talking about it.

2a. TBH I had plans to clean this up over all, but this was never a priority.
2b.
> because now REFTYPE is just a typedef to an integer type, which suggests that other types than currently used may also be valid for it 

And how that differs from changing `int64_t counter` embedded in struct previously? You could change the type of it as well.
2c.
> otherwise you'd just use the only appropriate type directly

So maybe we should just do that?

3. I hate this smart list feature.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/744)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required) (waiting on @boryspoplawski)

___
*[`libos/include/libos_internal.h` line 169 at r1](https://reviewable.io/reviews/gramineproject/gramine/744#-N6lxlUA8x5EHMClJDxL:-N6qtNQH0dfxGXvcggOf:bwjbzup) ([raw file](https://github.com/gramineproject/gramine/blob/1d1011c66d5effca97666ab940bc4a5a62f4e468/libos/include/libos_internal.h#L169)):*
> I hate this smart list feature.

Same here. It shouldn't override the numbers, only format them.

> No, actually I've never seen a repro, just heard you talking about it.

I'm pretty sure I linked to a godbolt snippet when I originally found it in some review.
update: Just found another place where I pasted a repro: https://github.com/gramineproject/graphene/issues/1593#issuecomment-648826006.

> And how that differs from changing `int64_t counter` embedded in struct previously? You could change the type of it as well.

It wasn't great before, but previously at least you could think that the typedef is to hide the struct name or something weird. But now it's not the case, it's just a typedefed integer type, but a lot of code like the one here both use the typedef _and_ assume that it has to be `int64_t`. This makes no sense and looks like an easy way to introduce a bug later.

> So maybe we should just do that?

Fine with me.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/744)* status: 3 of 25 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`libos/include/libos_internal.h` line 169 at r1](https://reviewable.io/reviews/gramineproject/gramine/744#-N6lxlUA8x5EHMClJDxL:-N6rMm6g-E03dwtfb4J0:bs4yecl) ([raw file](https://github.com/gramineproject/gramine/blob/1d1011c66d5effca97666ab940bc4a5a62f4e468/libos/include/libos_internal.h#L169)):*
> update: Just found another place where I pasted a repro: gramineproject/graphene#1593 (comment).

All right, this is the other case, but yeah, this sucks. Actually, did you report this upstream?

> Fine with me.

I've actually reworked this completely now.
___
*[`libos/include/libos_internal.h` line 176 at r1](https://reviewable.io/reviews/gramineproject/gramine/744#-N6m-AeW-mG-QCBroVu0:-N6rwOWe8AqSlI5LcWOW:b6vt3s0) ([raw file](https://github.com/gramineproject/gramine/blob/1d1011c66d5effca97666ab940bc4a5a62f4e468/libos/include/libos_internal.h#L176)):*

ditto
</blockquote></details>

does not apply anymore
___
*[`pal/src/host/linux-sgx/enclave_pages.c` line 10 at r1](https://reviewable.io/reviews/gramineproject/gramine/744#-N6mYT5pCicf5HoZyNMR:-N6rPXLF1URAaPKf7S-b:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/1d1011c66d5effca97666ab940bc4a5a62f4e468/pal/src/host/linux-sgx/enclave_pages.c#L10)):*

I guess this should actually be `size_t`? (and was `int64_t` only because of old atomics' inflexibility?)
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/pal_memory.c` line 16 at r1](https://reviewable.io/reviews/gramineproject/gramine/744#-N6mYbAcEiG8Od6jy3C3:-N6rPYjpDicozYgfkUkf:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/1d1011c66d5effca97666ab940bc4a5a62f4e468/pal/src/host/linux-sgx/pal_memory.c#L16)):*

ditto
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/pal_memory.c` line 83 at r1](https://reviewable.io/reviews/gramineproject/gramine/744#-N6mZi2R-1jiAgwJB04e:-N6rQ9Mh9ZWLbtbdxOB3:b-7u9v5e) ([raw file](https://github.com/gramineproject/gramine/blob/1d1011c66d5effca97666ab940bc4a5a62f4e468/pal/src/host/linux-sgx/pal_memory.c#L83)):*

Just `int`?
</blockquote></details>

Or even better this.
___
*[`pal/src/host/linux-sgx/pal_threading.c` line 38 at r1](https://reviewable.io/reviews/gramineproject/gramine/744#-N6mZoqI2Oj0LIJhQFxu:-N6rNNdU5-jSEM90x2ud:b-65x02o) ([raw file](https://github.com/gramineproject/gramine/blob/1d1011c66d5effca97666ab940bc4a5a62f4e468/pal/src/host/linux-sgx/pal_threading.c#L38)):*

unsigned?
</blockquote></details>

I went with `PAL_IDX`. I dislike this, but since we've yet not removed it, it makes more sense to use it here.

Update: and removed this in a next commit.

dimakuv (changes requested):
> Reviewed 22 of 22 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/744)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[`libos/src/bookkeep/libos_thread.c` line 23 at r2](https://reviewable.io/reviews/gramineproject/gramine/744#-N6vVmGADoVPn_liu4op:-N6vVmGBA39IQCb7j-mZ:b-o3xhlk) ([raw file](https://github.com/gramineproject/gramine/blob/8f892c78fe230cad1b16951360658e92ab29b9b0/libos/src/bookkeep/libos_thread.c#L23)):*
> ```C
> #include "libos_process.h"
> #include "libos_signal.h"
> #include "libos_thread.h"
> ```

Shouldn't you add an explicit `include libos_refcount.h` here? Actually, in all C files that use `refcount` stuff.

Or what is our policy? I understand that this file already uses e.g. `libos_thread.h` which implies the include of `libos_refcount.h`. I thought that we want to be explicit always in our include lists.
___
*[`libos/src/bookkeep/libos_thread.c` line 39 at r2](https://reviewable.io/reviews/gramineproject/gramine/744#-N6vVVBc-Sxja_Rkf4d8:-N6vVVBc-Sxja_Rkf4d9:b-yer7fs) ([raw file](https://github.com/gramineproject/gramine/blob/8f892c78fe230cad1b16951360658e92ab29b9b0/libos/src/bookkeep/libos_thread.c#L39)):*
> ```C
> #else
> #define DEBUG_PRINT_REF_COUNT(rc) __UNUSED(rc)
> #endif
> ```

Don't we want this snippet to be in `libos_refcount.h`? And use these macros everywhere instead of those ad-hoc
```
#ifdef DEBUG_REF
    log_debug("put handle %p(%s) (ref_count = %ld)", hdl, __handle_name(hdl), ref_count);
#endif
```

code snippets?

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/744)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`libos/src/bookkeep/libos_thread.c` line 23 at r2](https://reviewable.io/reviews/gramineproject/gramine/744#-N6vVmGADoVPn_liu4op:-N6wVI-58aq0MJn-qHZh:b5igzgi) ([raw file](https://github.com/gramineproject/gramine/blob/8f892c78fe230cad1b16951360658e92ab29b9b0/libos/src/bookkeep/libos_thread.c#L23)):*

Shouldn't you add an explicit `include libos_refcount.h` here? Actually, in all C files that use `refcount` stuff.

Or what is our policy? I understand that this file already uses e.g. `libos_thread.h` which implies the include of `libos_refcount.h`. I thought that we want to be explicit always in our include lists.
</blockquote></details>

My thought process was: if we have a definition of a struct with a `refcount_t` inside it, then it must have been already included. Basically if we have any reference to `refcount_t` then appropriate files must have been included and we don't have to do it.
Note this is different from case of accidentally included file by another header
___
*[`libos/src/bookkeep/libos_thread.c` line 39 at r2](https://reviewable.io/reviews/gramineproject/gramine/744#-N6vVVBc-Sxja_Rkf4d8:-N6wVlma9x6waAF2NNBW:b-cajdbg) ([raw file](https://github.com/gramineproject/gramine/blob/8f892c78fe230cad1b16951360658e92ab29b9b0/libos/src/bookkeep/libos_thread.c#L39)):*

Don't we want this snippet to be in `libos_refcount.h`? And use these macros everywhere instead of those ad-hoc
```
#ifdef DEBUG_REF
    log_debug("put handle %p(%s) (ref_count = %ld)", hdl, __handle_name(hdl), ref_count);
#endif
```

code snippets?
</blockquote></details>

No, I don't think it's useful. We got rid of issues with refcounting already (or at least nothing popped up for a long time), so I would rather remove those.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/744)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`libos/src/bookkeep/libos_thread.c` line 23 at r2](https://reviewable.io/reviews/gramineproject/gramine/744#-N6vVmGADoVPn_liu4op:-N6wYM6TE7IuWU5r1FPH:bi5xjla) ([raw file](https://github.com/gramineproject/gramine/blob/8f892c78fe230cad1b16951360658e92ab29b9b0/libos/src/bookkeep/libos_thread.c#L23)):*

My thought process was: if we have a definition of a struct with a `refcount_t` inside it, then it must have been already included. Basically if we have any reference to `refcount_t` then appropriate files must have been included and we don't have to do it.
Note this is different from case of accidentally included file by another header
</blockquote></details>

+1 to Borys, I use exactly the same rule.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/744)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`libos/src/bookkeep/libos_thread.c` line 23 at r2](https://reviewable.io/reviews/gramineproject/gramine/744#-N6vVmGADoVPn_liu4op:-N6w_tysCFuzQViltaFc:b-25i0l0) ([raw file](https://github.com/gramineproject/gramine/blob/8f892c78fe230cad1b16951360658e92ab29b9b0/libos/src/bookkeep/libos_thread.c#L23)):*

+1 to Borys, I use exactly the same rule.
</blockquote></details>

Got it. I'll use the same rule.
___
*[`libos/src/bookkeep/libos_thread.c` line 39 at r2](https://reviewable.io/reviews/gramineproject/gramine/744#-N6vVVBc-Sxja_Rkf4d8:-N6wa--W0mB4Wrva6t2Q:b-dhxouc) ([raw file](https://github.com/gramineproject/gramine/blob/8f892c78fe230cad1b16951360658e92ab29b9b0/libos/src/bookkeep/libos_thread.c#L39)):*

No, I don't think it's useful. We got rid of issues with refcounting already (or at least nothing popped up for a long time), so I would rather remove those.
</blockquote></details>

Fair enough. Removing these is also fine with me.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/744)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`libos/src/bookkeep/libos_thread.c` line 39 at r2](https://reviewable.io/reviews/gramineproject/gramine/744#-N6vVVBc-Sxja_Rkf4d8:-N6wiBGa95ETjNAvwo2j:bwmyj78) ([raw file](https://github.com/gramineproject/gramine/blob/8f892c78fe230cad1b16951360658e92ab29b9b0/libos/src/bookkeep/libos_thread.c#L39)):*

Fair enough. Removing these is also fine with me.
</blockquote></details>

So should I leave them as is, or remove?

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/744)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`libos/src/bookkeep/libos_thread.c` line 39 at r2](https://reviewable.io/reviews/gramineproject/gramine/744#-N6vVVBc-Sxja_Rkf4d8:-N6wifeE29_lxA3Yd-3N:bjkvi6q) ([raw file](https://github.com/gramineproject/gramine/blob/8f892c78fe230cad1b16951360658e92ab29b9b0/libos/src/bookkeep/libos_thread.c#L39)):*

So should I leave them as is, or remove?
</blockquote></details>

I understood your previous comment "I will remove them in a follow up PR".

But you can also just remove them here, in a separate commit.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/744)* status: 23 of 25 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @mkow)

___
*[`libos/src/bookkeep/libos_thread.c` line 39 at r2](https://reviewable.io/reviews/gramineproject/gramine/744#-N6vVVBc-Sxja_Rkf4d8:-N6wmbdRBVz79AVSZow5:b-prrf5v) ([raw file](https://github.com/gramineproject/gramine/blob/8f892c78fe230cad1b16951360658e92ab29b9b0/libos/src/bookkeep/libos_thread.c#L39)):*

I understood your previous comment "I will remove them in a follow up PR".

But you can also just remove them here, in a separate commit.
</blockquote></details>

My stance was I wouldn't touch them, but if you want to move them, then better to move to `/dev/null`.
Anyway, added a commit removing them (left dcache, not sure if it's useful or not, so left just in case).

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/744)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required) (waiting on @mkow)

mkow (approved):
> Reviewed 20 of 22 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/744)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`libos/include/libos_internal.h` line 169 at r1](https://reviewable.io/reviews/gramineproject/gramine/744#-N6lxlUA8x5EHMClJDxL:-N6yLvJ80LF5LbFTMK_2:by8owhf) ([raw file](https://github.com/gramineproject/gramine/blob/1d1011c66d5effca97666ab940bc4a5a62f4e468/libos/include/libos_internal.h#L169)):*
> Actually, did you report this upstream?

As you could have easily guessed, this bug was reported by someone else many years ago (2016) and is still not fixed. See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=69404.
___
*[`pal/src/host/linux-sgx/pal_memory.c` line 83 at r1](https://reviewable.io/reviews/gramineproject/gramine/744#-N6mZi2R-1jiAgwJB04e:-N6yRKhr9MdLD7199YMG:bw7xciz) ([raw file](https://github.com/gramineproject/gramine/blob/1d1011c66d5effca97666ab940bc4a5a62f4e468/pal/src/host/linux-sgx/pal_memory.c#L83)):*

Or even better this.
</blockquote></details>

Oh, right! Forgot about this nice macro :)

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/744)* status: 24 of 25 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @mkow)

___
*[`libos/include/libos_internal.h` line 169 at r1](https://reviewable.io/reviews/gramineproject/gramine/744#-N6lxlUA8x5EHMClJDxL:-N6yTACcDit2s36gLvN5:b-znnvi) ([raw file](https://github.com/gramineproject/gramine/blob/1d1011c66d5effca97666ab940bc4a5a62f4e468/libos/include/libos_internal.h#L169)):*

> Actually, did you report this upstream?

As you could have easily guessed, this bug was reported by someone else many years ago (2016) and is still not fixed. See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=69404.
</blockquote></details>

This is a different issue. It only talks about incompatible pointers but to objects of the same size. This is different bug.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/744)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/744)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`libos/include/libos_internal.h` line 169 at r1](https://reviewable.io/reviews/gramineproject/gramine/744#-N6lxlUA8x5EHMClJDxL:-N6yUYmT8um-DhzEGs2V:baz4dd9) ([raw file](https://github.com/gramineproject/gramine/blob/1d1011c66d5effca97666ab940bc4a5a62f4e468/libos/include/libos_internal.h#L169)):*

This is a different issue. It only talks about incompatible pointers but to objects of the same size. This is different bug.
</blockquote></details>

Ah, right, that one is a less dangerous subclass of this problem.

I just checked and they already fixed it! In GCC 12.1, which was released 2 months ago, lol :) (you can verify by changing GCC versions in that godbolt repro)

Issue #592, opened 2022-05-18
[LibOS] `/proc/cpuinfo` doesn't have the `flags` line


dimakuv:
## Description of the problem

Some workloads parse the `flags` line in the pseudo-file `/proc/cpuinfo`. Case in point: https://github.com/intel/intel-extension-for-pytorch/blob/306aa1ea0baf51ed700c3abfead5e46c4e076427/intel_extension_for_pytorch/utils/_cpu_isa.py

This PyTorch extension from Intel looks for the `AVX2` word (upper-cased) in the `flags` line of `/proc/cpuinfo`.

However, currently Gramine doesn't provide the `flags` line at all, see our emulation code of `/proc/cpuinfo`: https://github.com/gramineproject/gramine/blob/e6b697521a35ae24809d79237ad814ca497b3470/LibOS/shim/src/fs/proc/info.c#L132

## Steps to reproduce

Just to verify, I ran:
```
gramine/CI-Examples/busybox$ gramine-sgx busybox sh

cat /proc/cpuinfo
processor       : 0
vendor_id       : GenuineIntel
cpu family      : 6
model           : 158
model name      : Intel(R) Xeon(R) E-2176G CPU @ 3.70GHz
stepping        : 10
physical id     : 0
core id         : 0
cpu cores       : 6
bogomips        : 7399.70

... same for other CPU cores ...
```

As you can see, no flags printed.

## How to fix

- A workaround just to test this particular workload would be something like this diff:
```diff
diff --git a/LibOS/shim/src/fs/proc/info.c b/LibOS/shim/src/fs/proc/info.c
@@ -167,6 +167,7 @@ int proc_cpuinfo_load(struct shim_dentry* dent, char** out_data, size_t* out_siz
                 cores_in_socket++;
         }
         ADD_INFO("cpu cores\t: %zu\n", cores_in_socket);
+        ADD_INFO("flags\t\t: avx2\n");  /* FIXME: hard-coding is wrong, just for testing! */
         double bogomips = cpu->cpu_bogomips;
         // Apparently Gramine snprintf cannot into floats.
         ADD_INFO("bogomips\t: %lu.%02lu\n", (unsigned long)bogomips,
```

- A real fix would be to perform a bunch of `DkCpuIdRetrieve()` calls, find the currently enabled HW features, and add the corresponding flag-descriptions.

## Gramine commit hash

e6b697521a35ae24809d79237ad814ca497b3470

dimakuv:
[...]

PR #756, opened 2022-07-14
[LibOS,PAL] Add `flags` support to `/proc/cpuinfo`

kailun-qin:
Resolves https://github.com/gramineproject/gramine/issues/592.

Some workloads may require to parse the flags line in the pseudo-file `/proc/cpuinfo`. This patch adds support for `/proc/cpuinfo` by performing a bunch of `_PalCpuIdRetrieve()` calls, finding the currently enabled HW features, and adding the corresponding flag-descriptions.

### DIFF
--- a/libos/src/arch/x86_64/fs_proc/info.c
+++ b/libos/src/arch/x86_64/fs_proc/info.c
@@ -41,6 +41,9 @@ int proc_cpuinfo_display_cpu(char** str, size_t* size, size_t* max,
             cores_in_socket++;
     }
     ADD_INFO("cpu cores\t: %zu\n", cores_in_socket);
+
+    ADD_INFO("flags\t\t: %s\n", cpu->cpu_flags);
+
     double bogomips = cpu->cpu_bogomips;
     // Apparently Gramine snprintf cannot into floats.
     ADD_INFO("bogomips\t: %lu.%02lu\n", (unsigned long)bogomips,
--- a/pal/src/arch/x86_64/pal_cpuid.c
+++ b/pal/src/arch/x86_64/pal_cpuid.c
@@ -20,40 +20,262 @@
     (s)[2] = ((w) >> 16) & 0xff;    \
     (s)[3] = ((w) >> 24) & 0xff;
 
-static const char* const g_cpu_flags[] = {
-    "fpu",    // "x87 FPU on chip"
-    "vme",    // "virtual-8086 mode enhancement"
-    "de",     // "debugging extensions"
-    "pse",    // "page size extensions"
-    "tsc",    // "time stamp counter"
-    "msr",    // "RDMSR and WRMSR support"
-    "pae",    // "physical address extensions"
-    "mce",    // "machine check exception"
-    "cx8",    // "CMPXCHG8B inst."
-    "apic",   // "APIC on chip"
-    NULL,
-    "sep",    // "SYSENTER and SYSEXIT"
-    "mtrr",   // "memory type range registers"
-    "pge",    // "PTE global bit"
-    "mca",    // "machine check architecture"
-    "cmov",   // "conditional move/compare instruction"
-    "pat",    // "page attribute table"
-    "pse36",  // "page size extension"
-    "pn",     // "processor serial number"
-    "clflush",    // "CLFLUSH instruction"
-    NULL,
-    "dts",    // "debug store"
-    "acpi",   // "Onboard thermal control"
-    "mmx",    // "MMX Technology"
-    "fxsr",   // "FXSAVE/FXRSTOR"
-    "sse",    // "SSE extensions"
-    "sse2",   // "SSE2 extensions"
-    "ss",     // "self snoop"
-    "ht",     // "hyper-threading / multi-core supported"
-    "tm",     // "therm. monitor"
-    "ia64",   // "IA64"
-    "pbe",    // "pending break event"
+static const char* const g_cpu_flags_cpuid_1_ecx[] = {
+    [0]  = "pni",                /* "pni" SSE-3 */
+    [1]  = "pclmulqdq",          /* PCLMULQDQ instruction */
+    [2]  = "dtes64",             /* 64-bit Debug Store */
+    [3]  = "monitor",            /* "monitor" MONITOR/MWAIT support */
+    [4]  = "ds_cpl",             /* "ds_cpl" CPL-qualified (filtered) Debug Store */
+    [5]  = "vmx",                /* Hardware virtualization */
+    [6]  = "smx",                /* Safer Mode eXtensions */
+    [7]  = "est",                /* Enhanced SpeedStep */
+    [8]  = "tm2",                /* Thermal Monitor 2 */
+    [9]  = "ssse3",              /* Supplemental SSE-3 */
+    [10] = "cid",                /* Context ID */
+    [11] = "sdbg",               /* Silicon Debug */
+    [12] = "fma",                /* Fused multiply-add */
+    [13] = "cx16",               /* CMPXCHG16B instruction */
+    [14] = "xtpr",               /* Send Task Priority Messages */
+    [15] = "pdcm",               /* Perf/Debug Capabilities MSR */
+    [17] = "pcid",               /* Process Context Identifiers */
+    [18] = "dca",                /* Direct Cache Access */
+    [19] = "sse4_1",             /* "sse4_1" SSE-4.1 */
+    [20] = "sse4_2",             /* "sse4_2" SSE-4.2 */
+    [21] = "x2apic",             /* X2APIC */
+    [22] = "movbe",              /* MOVBE instruction */
+    [23] = "popcnt",             /* POPCNT instruction */
+    [24] = "tsc_deadline_timer", /* TSC deadline timer */
+    [25] = "aes",                /* AES instructions */
+    [26] = "xsave",              /* XSAVE/XRSTOR/XSETBV/XGETBV instructions */
+    [27] = "osxsave",            /* "" XSAVE instruction enabled in the OS */
+    [28] = "avx",                /* Advanced Vector Extensions */
+    [29] = "f16c",               /* 16-bit FP conversions */
+    [30] = "rdrand",             /* RDRAND instruction */
+    [31] = "hypervisor",         /* Running on a hypervisor */
 };
+
+static const char* const g_cpu_flags_cpuid_1_edx[] = {
+    [0]  = "fpu",     /* Onboard FPU */
+    [1]  = "vme",     /* Virtual Mode Extensions */
+    [2]  = "de",      /* Debugging Extensions */
+    [3]  = "pse",     /* Page Size Extensions */
+    [4]  = "tsc",     /* Time Stamp Counter */
+    [5]  = "msr",     /* Model-Specific Registers */
+    [6]  = "pae",     /* Physical Address Extensions */
+    [7]  = "mce",     /* Machine Check Exception */
+    [8]  = "cx8",     /* CMPXCHG8 instruction */
+    [9]  = "apic",    /* Onboard APIC */
+    [11] = "sep",     /* SYSENTER/SYSEXIT */
+    [12] = "mtrr",    /* Memory Type Range Registers */
+    [13] = "pge",     /* Page Global Enable */
+    [14] = "mca",     /* Machine Check Architecture */
+    [15] = "cmov",    /* CMOV instructions (plus FCMOVcc, FCOMI with FPU) */
+    [16] = "pat",     /* Page Attribute Table */
+    [17] = "pse36",   /* 36-bit PSEs */
+    [18] = "pn",      /* Processor serial number */
+    [19] = "clflush", /* CLFLUSH instruction */
+    [21] = "dts",     /* "dts" Debug Store */
+    [22] = "acpi",    /* ACPI via MSR */
+    [23] = "mmx",     /* Multimedia Extensions */
+    [24] = "fxsr",    /* FXSAVE/FXRSTOR, CR4.OSFXSR */
+    [25] = "sse",     /* "sse" */
+    [26] = "sse2",    /* "sse2" */
+    [27] = "ss",      /* "ss" CPU self snoop */
+    [28] = "ht",      /* Hyper-Threading */
+    [29] = "tm",      /* "tm" Automatic clock control */
+    [30] = "ia64",    /* IA-64 processor */
+    [31] = "pbe",     /* Pending Break Enable */
+};
+
+static const char* const g_cpu_flags_cpuid_6_eax[] = {
+    [0]  = "dtherm",         /* Digital Thermal Sensor */
+    [1]  = "ida",            /* Intel Dynamic Acceleration */
+    [2]  = "arat",           /* Always Running APIC Timer */
+    [4]  = "pln",            /* Intel Power Limit Notification */
+    [6]  = "pts",            /* Intel Package Thermal Status */
+    [7]  = "hwp",            /* Intel Hardware P-states */
+    [8]  = "hwp_notify",     /* HWP Notification */
+    [9]  = "hwp_act_window", /* HWP Activity Window */
+    [10] = "hwp_epp",        /* HWP Energy Perf. Preference */
+    [11] = "hwp_pkg_req",    /* HWP Package Level Request */
+    [19] = "hfi",            /* Hardware Feedback Interface */
+};
+
+static const char* const g_cpu_flags_cpuid_7_0_ebx[] = {
+    [0]  = "fsgsbase",        /* RDFSBASE, WRFSBASE, RDGSBASE, WRGSBASE instructions*/
+    [1]  = "tsc_adjust",      /* TSC adjustment MSR 0x3B */
+    [2]  = "sgx",             /* Software Guard Extensions */
+    [3]  = "bmi1",            /* 1st group bit manipulation extensions */
+    [4]  = "hle",             /* Hardware Lock Elision */
+    [5]  = "avx2",            /* AVX2 instructions */
+    [6]  = "fdp_excptn_only", /* "" FPU data pointer updated only on x87 exceptions */
+    [7]  = "smep",            /* Supervisor Mode Execution Protection */
+    [8]  = "bmi2",            /* 2nd group bit manipulation extensions */
+    [9]  = "erms",            /* Enhanced REP MOVSB/STOSB instructions */
+    [10] = "invpcid",         /* Invalidate Processor Context ID */
+    [11] = "rtm",             /* Restricted Transactional Memory */
+    [12] = "cqm",             /* Cache QoS Monitoring */
+    [13] = "zero_fcs_fds",    /* "" Zero out FPU CS and FPU DS */
+    [14] = "mpx",             /* Memory Protection Extension */
+    [15] = "rdt_a",           /* Resource Director Technology Allocation */
+    [16] = "avx512f",         /* AVX-512 Foundation */
+    [17] = "avx512dq",        /* AVX-512 DQ (Double/Quad granular) Instructions */
+    [18] = "rdseed",          /* RDSEED instruction */
+    [19] = "adx",             /* ADCX and ADOX instructions */
+    [20] = "smap",            /* Supervisor Mode Access Prevention */
+    [21] = "avx512ifma",      /* AVX-512 Integer Fused Multiply-Add instructions */
+    [23] = "clflushopt",      /* CLFLUSHOPT instruction */
+    [24] = "clwb",            /* CLWB instruction */
+    [25] = "intel_pt",        /* Intel Processor Trace */
+    [26] = "avx512pf",        /* AVX-512 Prefetch */
+    [27] = "avx512er",        /* AVX-512 Exponential and Reciprocal */
+    [28] = "avx512cd",        /* AVX-512 Conflict Detection */
+    [29] = "sha_ni",          /* SHA1/SHA256 Instruction Extensions */
+    [30] = "avx512bw",        /* AVX-512 BW (Byte/Word granular) Instructions */
+    [31] = "avx512vl",        /* AVX-512 VL (128/256 Vector Length) Extensions */
+};
+
+static const char* const g_cpu_flags_cpuid_7_0_ecx[] = {
+    [1]  = "avx512vbmi",       /* AVX512 Vector Bit Manipulation instructions*/
+    [2]  = "umip",             /* User Mode Instruction Protection */
+    [3]  = "pku",              /* Protection Keys for Userspace */
+    [4]  = "ospke",            /* OS Protection Keys Enable */
+    [5]  = "waitpkg",          /* UMONITOR/UMWAIT/TPAUSE Instructions */
+    [6]  = "avx512_vbmi2",     /* Additional AVX512 Vector Bit Manipulation Instructions */
+    [8]  = "gfni",             /* Galois Field New Instructions */
+    [9]  = "vaes",             /* Vector AES */
+    [10] = "vpclmulqdq",       /* Carry-Less Multiplication Double Quadword */
+    [11] = "avx512_vnni",      /* Vector Neural Network Instructions */
+    [12] = "avx512_bitalg",    /* Support for VPOPCNT[B,W] and VPSHUF-BITQMB instructions */
+    [13] = "tme",              /* Intel Total Memory Encryption */
+    [14] = "avx512_vpopcntdq", /* POPCNT for vectors of DW/QW */
+    [16] = "la57",             /* 5-level page tables */
+    [22] = "rdpid",            /* RDPID instruction */
+    [24] = "bus_lock_detect",  /* Bus Lock detect */
+    [25] = "cldemote",         /* CLDEMOTE instruction */
+    [27] = "movdiri",          /* MOVDIRI instruction */
+    [28] = "movdir64b",        /* MOVDIR64B instruction */
+    [29] = "enqcmd",           /* ENQCMD and ENQCMDS instructions */
+    [30] = "sgx_lc",           /* Software Guard Extensions Launch Control */
+};
+
+static const char* const g_cpu_flags_cpuid_7_0_edx[] = {
+    [2]  = "avx512_4vnniw",       /* AVX-512 Neural Network Instructions */
+    [3]  = "avx512_4fmaps",       /* AVX-512 Multiply Accumulation Single precision */
+    [4]  = "fsrm",                /* Fast Short Rep Mov */
+    [8]  = "avx512_vp2intersect", /* AVX-512 Intersect for D/Q */
+    [9]  = "srbds_ctrl",          /* "" SRBDS mitigation MSR available */
+    [10] = "md_clear",            /* VERW clears CPU buffers */
+    [11] = "rtm_always_abort",    /* "" RTM transaction always aborts */
+    [13] = "tsx_force_abort",     /* "" TSX_FORCE_ABORT */
+    [14] = "serialize",           /* SERIALIZE instruction */
+    [15] = "hybrid_cpu",          /* "" This part has CPUs of more than one type */
+    [16] = "tsxldtrk",            /* TSX Suspend Load Address Tracking */
+    [18] = "pconfig",             /* Intel PCONFIG */
+    [19] = "arch_lbr",            /* Intel ARCH LBR */
+    [20] = "ibt",                 /* Indirect Branch Tracking */
+    [22] = "amx_bf16",            /* AMX bf16 Support */
+    [23] = "avx512_fp16",         /* AVX512 FP16 */
+    [24] = "amx_tile",            /* AMX tile Support */
+    [25] = "amx_int8",            /* AMX int8 Support */
+    [26] = "spec_ctrl",           /* "" Speculation Control (IBRS + IBPB) */
+    [27] = "intel_stibp",         /* "" Single Thread Indirect Branch Predictors */
+    [28] = "flush_l1d",           /* Flush L1D cache */
+    [29] = "arch_capabilities",   /* IA32_ARCH_CAPABILITIES MSR (Intel) */
+    [30] = "core_capabilities",   /* "" IA32_CORE_CAPABILITIES MSR */
+    [31] = "spec_ctrl_ssbd",      /* "" Speculative Store Bypass Disable */
+};
+
+static const char* const g_cpu_flags_cpuid_7_1_eax[] = {
+    [4] = "avx_vnni",    /* AVX VNNI instructions */
+    [5] = "avx512_bf16", /* AVX512 BFLOAT16 instructions */
+};
+
+static const char* const g_cpu_flags_cpuid_d_1_eax[] = {
+    [0] = "xsaveopt", /* XSAVEOPT instruction */
+    [1] = "xsavec",   /* XSAVEC instruction */
+    [2] = "xgetbv1",  /* XGETBV with ECX = 1 instruction */
+    [3] = "xsaves",   /* XSAVES/XRSTORS instructions */
+    [4] = "xfd",      /* "" eXtended Feature Disabling */
+};
+
+static const char* const g_cpu_flags_cpuid_8000_0001_ecx[] = {
+    [0]  = "lahf_lm",       /* LAHF/SAHF in long mode */
+    [1]  = "cmp_legacy",    /* If yes HyperThreading not valid */
+    [2]  = "svm",           /* Secure Virtual Machine */
+    [3]  = "extapic",       /* Extended APIC space */
+    [4]  = "cr8_legacy",    /* CR8 in 32-bit mode */
+    [5]  = "abm",           /* Advanced bit manipulation */
+    [6]  = "sse4a",         /* SSE-4A */
+    [7]  = "misalignsse",   /* Misaligned SSE mode */
+    [8]  = "3dnowprefetch", /* 3DNow prefetch instructions */
+    [9]  = "osvw",          /* OS Visible Workaround */
+    [10] = "ibs",           /* Instruction Based Sampling */
+    [11] = "xop",           /* extended AVX instructions */
+    [12] = "skinit",        /* SKINIT/STGI instructions */
+    [13] = "wdt",           /* Watchdog timer */
+    [15] = "lwp",           /* Light Weight Profiling */
+    [16] = "fma4",          /* 4 operands MAC instructions */
+    [17] = "tce",           /* Translation Cache Extension */
+    [19] = "nodeid_msr",    /* NodeId MSR */
+    [21] = "tbm",           /* Trailing Bit Manipulations */
+    [22] = "topoext",       /* Topology extensions CPUID leafs */
+    [23] = "perfctr_core",  /* Core performance counter extensions */
+    [24] = "perfctr_nb",    /* NB performance counter extensions */
+    [26] = "bpext",         /* Data breakpoint extension */
+    [27] = "ptsc",          /* Performance time-stamp counter */
+    [28] = "perfctr_llc",   /* Last Level Cache performance counter extensions */
+    [29] = "mwaitx",        /* MWAIT extension (MONITORX/MWAITX instructions) */
+};
+
+static const char* const g_cpu_flags_cpuid_8000_0001_edx[] = {
+    [11] = "syscall",  /* SYSCALL/SYSRET */
+    [19] = "mp",       /* MP Capable */
+    [20] = "nx",       /* Execute Disable */
+    [22] = "mmxext",   /* AMD MMX extensions */
+    [25] = "fxsr_opt", /* FXSAVE/FXRSTOR optimizations */
+    [26] = "pdpe1gb",  /* "pdpe1gb" GB pages */
+    [27] = "rdtscp",   /* RDTSCP */
+    [29] = "lm",       /* Long Mode (x86-64, 64-bit support) */
+    [30] = "3dnowext", /* AMD 3DNow extensions */
+    [31] = "3dnow",    /* 3DNow */
+};
+
+static int extend_cap_flags(const char* const cpu_flags[], const unsigned int* words,
+                            enum CPUID_WORD w, char** flags, size_t* flen, size_t* fmax) {
+    assert(*flags != NULL);
+
+    int rv = 0;
+
+    for (int i = 0; i < 32; i++) {
+        if (!cpu_flags[i])
+            continue;
+
+        if (BIT_EXTRACT_LE(words[w], i, i + 1)) {
+            size_t len = strlen(cpu_flags[i]);
+            if (*flen + len + 1 > *fmax) {
+                char* new_flags = malloc(*fmax * 2);
+                if (!new_flags) {
+                    rv = -PAL_ERROR_NOMEM;
+                    goto out_err;
+                }
+                memcpy(new_flags, *flags, *flen);
+                free(*flags);
+                *fmax *= 2;
+                *flags = new_flags;
+            }
+            memcpy(*flags + *flen, cpu_flags[i], len);
+            *flen += len;
+            (*flags)[(*flen)++] = ' ';
+        }
+    }
+
+    return 0;
+
+out_err:
+    free(*flags);
+    return rv;
+}
 
 int _PalGetCPUInfo(struct pal_cpu_info* ci) {
     unsigned int words[CPUID_WORD_NUM];
@@ -68,6 +290,7 @@ int _PalGetCPUInfo(struct pal_cpu_info* ci) {
         return -PAL_ERROR_NOMEM;
 
     _PalCpuIdRetrieve(0, 0, words);
+    unsigned int cpuid_level = words[CPUID_WORD_EAX];
     FOUR_CHARS_VALUE(&vendor_id[0], words[CPUID_WORD_EBX]);
     FOUR_CHARS_VALUE(&vendor_id[4], words[CPUID_WORD_EDX]);
     FOUR_CHARS_VALUE(&vendor_id[8], words[CPUID_WORD_ECX]);
@@ -96,7 +319,7 @@ int _PalGetCPUInfo(struct pal_cpu_info* ci) {
 
     if (!memcmp(vendor_id, "GenuineIntel", 12) || !memcmp(vendor_id, "AuthenticAMD", 12)) {
         ci->cpu_family += BIT_EXTRACT_LE(words[CPUID_WORD_EAX], 20, 28);
-        ci->cpu_model  += BIT_EXTRACT_LE(words[CPUID_WORD_EAX], 16, 20) << 4;
+        ci->cpu_model += BIT_EXTRACT_LE(words[CPUID_WORD_EAX], 16, 20) << 4;
     }
 
     size_t flen = 0;
@@ -107,30 +330,57 @@ int _PalGetCPUInfo(struct pal_cpu_info* ci) {
         goto out_err;
     }
 
-    for (int i = 0; i < 32; i++) {
-        if (!g_cpu_flags[i])
-            continue;
+    /* Intel-defined flags: level 0x00000001 */
+    extend_cap_flags(g_cpu_flags_cpuid_1_ecx, words, CPUID_WORD_ECX, &flags, &flen, &fmax);
+    extend_cap_flags(g_cpu_flags_cpuid_1_edx, words, CPUID_WORD_EDX, &flags, &flen, &fmax);
 
-        if (BIT_EXTRACT_LE(words[CPUID_WORD_EDX], i, i + 1)) {
-            size_t len = strlen(g_cpu_flags[i]);
-            if (flen + len + 1 > fmax) {
-                char* new_flags = malloc(fmax * 2);
-                if (!new_flags) {
-                    rv = -PAL_ERROR_NOMEM;
-                    goto out_err;
-                }
-                memcpy(new_flags, flags, flen);
-                free(flags);
-                fmax *= 2;
-                flags = new_flags;
-            }
-            memcpy(flags + flen, g_cpu_flags[i], len);
-            flen += len;
-            flags[flen++] = ' ';
+    /* Thermal and Power Management Leaf: level 0x00000006 (eax) */
+    if (cpuid_level >= 0x00000006) {
+        _PalCpuIdRetrieve(0x00000006, 0, words);
+
+        extend_cap_flags(g_cpu_flags_cpuid_6_eax, words, CPUID_WORD_EAX, &flags, &flen, &fmax);
+    }
+
+    /* Additional Intel-defined flags: level 0x00000007 */
+    if (cpuid_level >= 0x00000007) {
+        _PalCpuIdRetrieve(0x00000007, 0, words);
+
+        unsigned int eax = words[CPUID_WORD_EAX];
+        extend_cap_flags(g_cpu_flags_cpuid_7_0_ebx, words, CPUID_WORD_EBX, &flags, &flen, &fmax);
+        extend_cap_flags(g_cpu_flags_cpuid_7_0_ecx, words, CPUID_WORD_ECX, &flags, &flen, &fmax);
+        extend_cap_flags(g_cpu_flags_cpuid_7_0_edx, words, CPUID_WORD_EDX, &flags, &flen, &fmax);
+
+        /* Check valid sub-leaf index before accessing it */
+        if (eax >= 1) {
+            _PalCpuIdRetrieve(0x00000007, 1, words);
+
+            extend_cap_flags(g_cpu_flags_cpuid_7_1_eax, words, CPUID_WORD_EAX, &flags, &flen, &fmax);
         }
     }
 
+    /* Extended state features: level 0x0000000d */
+    if (cpuid_level >= 0x0000000d) {
+        _PalCpuIdRetrieve(0x0000000d, 1, words);
+
+        extend_cap_flags(g_cpu_flags_cpuid_d_1_eax, words, CPUID_WORD_EAX, &flags, &flen, &fmax);
+    }
+
+    /* AMD-defined flags: level 0x80000001 */
+    _PalCpuIdRetrieve(0x80000000, 0, words);
+    unsigned int eax = words[CPUID_WORD_EAX];
+
+    if ((eax & 0xffff0000) == 0x80000000) {
+        if (eax >= 0x80000001) {
+            _PalCpuIdRetrieve(0x80000001, 0, words);
+
+            extend_cap_flags(g_cpu_flags_cpuid_8000_0001_ecx, words, CPUID_WORD_ECX, &flags, &flen, &fmax);
+            extend_cap_flags(g_cpu_flags_cpuid_8000_0001_edx, words, CPUID_WORD_EDX, &flags, &flen, &fmax);
+        }
+    }
+
+    /* End the capability flags extension */
     flags[flen ? flen - 1 : 0] = 0;
+
     ci->cpu_flags = flags;
 
     ci->cpu_bogomips = _PalGetBogomips();


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/756)* status: all files reviewed, 20 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vlNPw8rr91zO082gS:-N6vlNPw8rr91zO082gT:b15qszh) (no related file):*
Relevant links to reviewing this PR:
- https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/Documentation/x86/cpuinfo.rst
- https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/arch/x86/include/asm/cpufeatures.h

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vvIty0PLgFjdP9446:-N6vvIty0PLgFjdP9447:b-6vb34t) (no related file):*
Awesome PR!

I tried to be very thorough in reviewing the PR, manually checking all flags against the Linux source code.

___
___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vileg34XvE_fKLEdJ):*
`support` -> `line`
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 22 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vmUl0DosiM4xPxUNR:-N6vmUl0DosiM4xPxUNS:bl835vt) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L22)):*
> ```C
>     (s)[2] = ((w) >> 16) & 0xff;    \
>     (s)[3] = ((w) >> 24) & 0xff;
> ```

Please add a comment with these two links:
```
- https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/Documentation/x86/cpuinfo.rst
- https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/arch/x86/include/asm/cpufeatures.h
```

And mention that this list was last revised in July 2022.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 22 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vu49_2WykZXUlAE4N:-N6vu49_2WykZXUlAE4O:bpmx79p) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L22)):*
> ```C
>     (s)[2] = ((w) >> 16) & 0xff;    \
>     (s)[3] = ((w) >> 24) & 0xff;
> ```

You are only listing common x86-64 flags and Intel-specific flags here. You **ignore** the AMD-specific flags.

But technically, in the future, we may want to also make Gramine useful on AMD machines. So we'll need to add AMD-specific flags there. Please add a TODO comment about it.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 50 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vmHG0BIaOrK1pFZxA:-N6vmHG0BIaOrK1pFZxB:b94wroa) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L50)):*
> ```C
>     [25] = "aes",                /* AES instructions */
>     [26] = "xsave",              /* XSAVE/XRSTOR/XSETBV/XGETBV instructions */
>     [27] = "osxsave",            /* "" XSAVE instruction enabled in the OS */
> ```

This is wrong, it should be `""` (meaning it will not appear in the cpuinfo output). See https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/Documentation/x86/cpuinfo.rst#n104 for explanation.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 111 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vnzzk7SoR0G15DtvE:-N6vnzzk7SoR0G15DtvF:b94wroa) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L111)):*
> ```C
>     [4]  = "hle",             /* Hardware Lock Elision */
>     [5]  = "avx2",            /* AVX2 instructions */
>     [6]  = "fdp_excptn_only", /* "" FPU data pointer updated only on x87 exceptions */
> ```

This is wrong, it should be `""` (meaning it will not appear in the cpuinfo output). See https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/Documentation/x86/cpuinfo.rst#n104 for explanation.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 118 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vo4UZ7CyWdlYlaQcL:-N6vo4UZ7CyWdlYlaQcM:b94wroa) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L118)):*
> ```C
>     [11] = "rtm",             /* Restricted Transactional Memory */
>     [12] = "cqm",             /* Cache QoS Monitoring */
>     [13] = "zero_fcs_fds",    /* "" Zero out FPU CS and FPU DS */
> ```

This is wrong, it should be `""` (meaning it will not appear in the cpuinfo output). See https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/Documentation/x86/cpuinfo.rst#n104 for explanation.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 167 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vohrj0tOUm0ISlVmO:-N6vohrj0tOUm0ISlVmP:b94wroa) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L167)):*
> ```C
>     [4]  = "fsrm",                /* Fast Short Rep Mov */
>     [8]  = "avx512_vp2intersect", /* AVX-512 Intersect for D/Q */
>     [9]  = "srbds_ctrl",          /* "" SRBDS mitigation MSR available */
> ```

This is wrong, it should be `""` (meaning it will not appear in the cpuinfo output). See https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/Documentation/x86/cpuinfo.rst#n104 for explanation.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 169 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vokgi91TaFmdWmArn:-N6vokgi91TaFmdWmAro:b94wroa) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L169)):*
> ```C
>     [9]  = "srbds_ctrl",          /* "" SRBDS mitigation MSR available */
>     [10] = "md_clear",            /* VERW clears CPU buffers */
>     [11] = "rtm_always_abort",    /* "" RTM transaction always aborts */
> ```

This is wrong, it should be `""` (meaning it will not appear in the cpuinfo output). See https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/Documentation/x86/cpuinfo.rst#n104 for explanation.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 170 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6voltO1qaKpyMb2jqw:-N6voltO1qaKpyMb2jqx:b94wroa) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L170)):*
> ```C
>     [10] = "md_clear",            /* VERW clears CPU buffers */
>     [11] = "rtm_always_abort",    /* "" RTM transaction always aborts */
>     [13] = "tsx_force_abort",     /* "" TSX_FORCE_ABORT */
> ```

This is wrong, it should be `""` (meaning it will not appear in the cpuinfo output). See https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/Documentation/x86/cpuinfo.rst#n104 for explanation.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 172 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vop-i81jcvcOprDNX:-N6vop-i81jcvcOprDNY:b94wroa) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L172)):*
> ```C
>     [13] = "tsx_force_abort",     /* "" TSX_FORCE_ABORT */
>     [14] = "serialize",           /* SERIALIZE instruction */
>     [15] = "hybrid_cpu",          /* "" This part has CPUs of more than one type */
> ```

This is wrong, it should be `""` (meaning it will not appear in the cpuinfo output). See https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/Documentation/x86/cpuinfo.rst#n104 for explanation.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 181 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vowLN0-58tUFbAmzU:-N6vowLN0-58tUFbAmzV:b94wroa) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L181)):*
> ```C
>     [24] = "amx_tile",            /* AMX tile Support */
>     [25] = "amx_int8",            /* AMX int8 Support */
>     [26] = "spec_ctrl",           /* "" Speculation Control (IBRS + IBPB) */
> ```

This is wrong, it should be `""` (meaning it will not appear in the cpuinfo output). See https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/Documentation/x86/cpuinfo.rst#n104 for explanation.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 182 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vowreCvk2PbC-Rv_j:-N6vowreCvk2PbC-Rv_k:b94wroa) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L182)):*
> ```C
>     [25] = "amx_int8",            /* AMX int8 Support */
>     [26] = "spec_ctrl",           /* "" Speculation Control (IBRS + IBPB) */
>     [27] = "intel_stibp",         /* "" Single Thread Indirect Branch Predictors */
> ```

This is wrong, it should be `""` (meaning it will not appear in the cpuinfo output). See https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/Documentation/x86/cpuinfo.rst#n104 for explanation.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 185 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vp-8bECs119fWLZBe:-N6vp-8bECs119fWLZBf:b94wroa) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L185)):*
> ```C
>     [28] = "flush_l1d",           /* Flush L1D cache */
>     [29] = "arch_capabilities",   /* IA32_ARCH_CAPABILITIES MSR (Intel) */
>     [30] = "core_capabilities",   /* "" IA32_CORE_CAPABILITIES MSR */
> ```

This is wrong, it should be `""` (meaning it will not appear in the cpuinfo output). See https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/Documentation/x86/cpuinfo.rst#n104 for explanation.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 186 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vp-dTE_UfAIYVnGom:-N6vp-dTE_UfAIYVnGon:b94wroa) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L186)):*
> ```C
>     [29] = "arch_capabilities",   /* IA32_ARCH_CAPABILITIES MSR (Intel) */
>     [30] = "core_capabilities",   /* "" IA32_CORE_CAPABILITIES MSR */
>     [31] = "spec_ctrl_ssbd",      /* "" Speculative Store Bypass Disable */
> ```

This is wrong, it should be `""` (meaning it will not appear in the cpuinfo output). See https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/Documentation/x86/cpuinfo.rst#n104 for explanation.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 199 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vpE-02nF3QDaL5MA-:-N6vpE-02nF3QDaL5MA0:b94wroa) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L199)):*
> ```C
>     [2] = "xgetbv1",  /* XGETBV with ECX = 1 instruction */
>     [3] = "xsaves",   /* XSAVES/XRSTORS instructions */
>     [4] = "xfd",      /* "" eXtended Feature Disabling */
> ```

This is wrong, it should be `""` (meaning it will not appear in the cpuinfo output). See https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/Documentation/x86/cpuinfo.rst#n104 for explanation.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 306 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vuaDW4v-_lvsPbBvB:-N6vuaDX-ORctlCNtog8:b4odzyx) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L306)):*
> ```C
>         goto out_err;
>     }
>     _PalCpuIdRetrieve(0x80000002, 0, words);
> ```

Here you're accessing high leaves without checking `eax` from `0x80000000`. Could you please add the check?

Actually, since we require this, I would even do a check that `if 0x80000000[eax] < 0x80000004 then goto out_err`.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 322 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vs6E1-tKfaZbmK2K0:-N6vs6E1-tKfaZbmK2K1:b-1csib2) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L322)):*
> ```C
>     if (!memcmp(vendor_id, "GenuineIntel", 12) || !memcmp(vendor_id, "AuthenticAMD", 12)) {
>         ci->cpu_family += BIT_EXTRACT_LE(words[CPUID_WORD_EAX], 20, 28);
>         ci->cpu_model += BIT_EXTRACT_LE(words[CPUID_WORD_EAX], 16, 20) << 4;
> ```

This is not really needed, I suggest to revert this change (it only distracts from the main changes)
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 334 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vsDQO1-CNGtZD1InE:-N6vsDQP6UrU2MP5aFzj:b-qf5xj5) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L334)):*
> ```C
> 
>     /* Intel-defined flags: level 0x00000001 */
>     extend_cap_flags(g_cpu_flags_cpuid_1_ecx, words, CPUID_WORD_ECX, &flags, &flen, &fmax);
> ```

You're not checking for errors. I suggest to wrap `extend_cap_flags()` calls into a `EXTEND_CAP_FLAGS()` macro that will call this func and `return ret` on errors. This way, you can also hide always-the-same `words, ..., &flags, &flen, &fmax` arguments under the macro.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 372 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vtZHM7dFiOV9xdNUm:-N6vtZHM7dFiOV9xdNUn:bfr1od4) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L372)):*
> ```C
>     unsigned int eax = words[CPUID_WORD_EAX];
> 
>     if ((eax & 0xffff0000) == 0x80000000) {
> ```

What is this check? Where did you get it from? I don't see such a description in the CPUID instruction description.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 373 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vv0li3HUC7S6WQOZm:-N6vv0lj5u0VfX02CGzr:bq0phu4) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L373)):*
> ```C
> 
>     if ((eax & 0xffff0000) == 0x80000000) {
>         if (eax >= 0x80000001) {
> ```

We already access `0x80000002`, `0x80000003`, ... above, so this check is redundant.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/756)* status: 1 of 2 files reviewed, 20 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " and "squash! " found in commit messages' one-liners (waiting on @dimakuv and @kailun-qin)

___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 22 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vmUl0DosiM4xPxUNR:-N6vzfWL0IiQMYNtWxiF:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L22)):*

Please add a comment with these two links:
```
- https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/Documentation/x86/cpuinfo.rst
- https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/arch/x86/include/asm/cpufeatures.h
```

And mention that this list was last revised in July 2022.
</blockquote></details>

Done.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 22 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vu49_2WykZXUlAE4N:-N6vzwo4AY6vYjJvxfyH:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L22)):*

You are only listing common x86-64 flags and Intel-specific flags here. You **ignore** the AMD-specific flags.

But technically, in the future, we may want to also make Gramine useful on AMD machines. So we'll need to add AMD-specific flags there. Please add a TODO comment about it.
</blockquote></details>

Done.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 50 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vmHG0BIaOrK1pFZxA:-N6vzxVsEfBKC3YBn53R:b-ubb87g) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L50)):*

This is wrong, it should be `""` (meaning it will not appear in the cpuinfo output). See https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/Documentation/x86/cpuinfo.rst#n104 for explanation.
</blockquote></details>

Good catch, thanks.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 111 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vnzzk7SoR0G15DtvE:-N6w-1X-9Q99AHMmHv09:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L111)):*

This is wrong, it should be `""` (meaning it will not appear in the cpuinfo output). See https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/Documentation/x86/cpuinfo.rst#n104 for explanation.
</blockquote></details>

Done.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 118 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vo4UZ7CyWdlYlaQcL:-N6w-2C43c2XH7jb9cet:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L118)):*

This is wrong, it should be `""` (meaning it will not appear in the cpuinfo output). See https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/Documentation/x86/cpuinfo.rst#n104 for explanation.
</blockquote></details>

Done.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 167 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vohrj0tOUm0ISlVmO:-N6w-4ki3XsCYxf0tdXG:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L167)):*

This is wrong, it should be `""` (meaning it will not appear in the cpuinfo output). See https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/Documentation/x86/cpuinfo.rst#n104 for explanation.
</blockquote></details>

Done.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 169 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vokgi91TaFmdWmArn:-N6w-8w55Lag61Pm49r3:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L169)):*

This is wrong, it should be `""` (meaning it will not appear in the cpuinfo output). See https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/Documentation/x86/cpuinfo.rst#n104 for explanation.
</blockquote></details>

Done.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 170 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6voltO1qaKpyMb2jqw:-N6w-9UC6l_q-f1LBOC8:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L170)):*

This is wrong, it should be `""` (meaning it will not appear in the cpuinfo output). See https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/Documentation/x86/cpuinfo.rst#n104 for explanation.
</blockquote></details>

Done.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 172 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vop-i81jcvcOprDNX:-N6w-AAZ0f7NxOvWtW_G:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L172)):*

This is wrong, it should be `""` (meaning it will not appear in the cpuinfo output). See https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/Documentation/x86/cpuinfo.rst#n104 for explanation.
</blockquote></details>

Done.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 181 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vowLN0-58tUFbAmzU:-N6w-DfT-Bhr-uBuk-Ke:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L181)):*

This is wrong, it should be `""` (meaning it will not appear in the cpuinfo output). See https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/Documentation/x86/cpuinfo.rst#n104 for explanation.
</blockquote></details>

Done.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 182 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vowreCvk2PbC-Rv_j:-N6w-fCcCEh6lVeep_-X:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L182)):*

This is wrong, it should be `""` (meaning it will not appear in the cpuinfo output). See https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/Documentation/x86/cpuinfo.rst#n104 for explanation.
</blockquote></details>

Done.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 185 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vp-8bECs119fWLZBe:-N6w-hOP2C4R4IeSQuZt:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L185)):*

This is wrong, it should be `""` (meaning it will not appear in the cpuinfo output). See https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/Documentation/x86/cpuinfo.rst#n104 for explanation.
</blockquote></details>

Done.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 186 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vp-dTE_UfAIYVnGom:-N6w-hlgEUj5G29HTPrR:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L186)):*

This is wrong, it should be `""` (meaning it will not appear in the cpuinfo output). See https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/Documentation/x86/cpuinfo.rst#n104 for explanation.
</blockquote></details>

Done.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 199 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vpE-02nF3QDaL5MA-:-N6w-j6-1Wco6OjGbEu5:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L199)):*

This is wrong, it should be `""` (meaning it will not appear in the cpuinfo output). See https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/Documentation/x86/cpuinfo.rst#n104 for explanation.
</blockquote></details>

Done.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 306 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vuaDW4v-_lvsPbBvB:-N6w2gj553nVENjRX9nR:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L306)):*

Here you're accessing high leaves without checking `eax` from `0x80000000`. Could you please add the check?

Actually, since we require this, I would even do a check that `if 0x80000000[eax] < 0x80000004 then goto out_err`.
</blockquote></details>

Done.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 322 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vs6E1-tKfaZbmK2K0:-N6w4nSbDHawaMeGH5ed:bp3aow6) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L322)):*

This is not really needed, I suggest to revert this change (it only distracts from the main changes)
</blockquote></details>

Reverted.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 334 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vsDQO1-CNGtZD1InE:-N6w0DFA4rFEfsa1A4ZK:b-5y2gm0) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L334)):*

You're not checking for errors. I suggest to wrap `extend_cap_flags()` calls into a `EXTEND_CAP_FLAGS()` macro that will call this func and `return ret` on errors. This way, you can also hide always-the-same `words, ..., &flags, &flen, &fmax` arguments under the macro.
</blockquote></details>

Great suggestion, thanks!
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 372 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vtZHM7dFiOV9xdNUm:-N6w6kPC9RBBeU3mEznj:b-gh363i) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L372)):*

What is this check? Where did you get it from? I don't see such a description in the CPUID instruction description.
</blockquote></details>

It's actually from https://github.com/torvalds/linux/blob/master/arch/x86/kernel/cpu/common.c#L1048. I checked the manual but found nothing relevant to the check. Remove it for now.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 373 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vv0li3HUC7S6WQOZm:-N6w6jw27q95F1jz3DhR:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L373)):*

We already access `0x80000002`, `0x80000003`, ... above, so this check is redundant.
</blockquote></details>

Done.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/756)* status: 1 of 2 files reviewed, 20 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " and "squash! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vileg34XvE_fKLEdJ):*

`support` -> `line`
</blockquote></details>

Done.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/756)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " and "squash! " found in commit messages' one-liners (waiting on @kailun-qin)

___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 322 at r1](https://reviewable.io/reviews/gramineproject/gramine/756#-N6vs6E1-tKfaZbmK2K0:-N6wTu8g6pehiwdx3YKt:bfhqmln) ([raw file](https://github.com/gramineproject/gramine/blob/ac40193352cb68d7a2ebfd8120b1444d6bb1c4c1/pal/src/arch/x86_64/pal_cpuid.c#L322)):*

Reverted.
</blockquote></details>

Not done?

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/756)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/756)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners

dimakuv:
Jenkins, test this please

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/756)* status: all files reviewed, 15 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @kailun-qin)

___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 392 at r4](https://reviewable.io/reviews/gramineproject/gramine/756#-N71v9cT023E5-ltQ18v:-N77SJSw-iPP9ORo_egL:bv52dhb) ([raw file](https://github.com/gramineproject/gramine/blob/8892d2ea3cf99480209c7e77a7e662a691e34ac2/pal/src/arch/x86_64/pal_cpuid.c#L392)):*

Hmm, I don't like that we keep such big strings in per CPU struct. This string is same for all CPUs currently, so we could just have a global variable.
Actually, how does Gramine handle NUMA? Is it supported? Does all CPU require to have exactly the same feature set? cc @dimakuv @mkow 
</blockquote></details>

Yes, it's supported. But there's still a TODO there: https://github.com/gramineproject/gramine/blob/85078de3222367da4d0aaa14793c218648a94659/pal/include/arch/x86_64/pal_topology.h#L72
Maybe it should actually be fixed before merging this PR? (just asking, didn't read this PR yet)

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/756)* status: all files reviewed, 15 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @kailun-qin)

___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 392 at r4](https://reviewable.io/reviews/gramineproject/gramine/756#-N71v9cT023E5-ltQ18v:-N7CRp_B0oEouZcvuOH1:b8u5oqh) ([raw file](https://github.com/gramineproject/gramine/blob/8892d2ea3cf99480209c7e77a7e662a691e34ac2/pal/src/arch/x86_64/pal_cpuid.c#L392)):*

Yes, it's supported. But there's still a TODO there: https://github.com/gramineproject/gramine/blob/85078de3222367da4d0aaa14793c218648a94659/pal/include/arch/x86_64/pal_topology.h#L72
Maybe it should actually be fixed before merging this PR? (just asking, didn't read this PR yet)
</blockquote></details>

What if two CPUs are different and support different features?

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/756)* status: all files reviewed, 15 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 26 at r4](https://reviewable.io/reviews/gramineproject/gramine/756#-N71hrpsDpgLqnAo9aQ1:-N7EDJL_B0evJtUFOb7w:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/8892d2ea3cf99480209c7e77a7e662a691e34ac2/pal/src/arch/x86_64/pal_cpuid.c#L26)):*

Since this is a link to web interface (not in-git file), please change to `https://elixir.bootlin.com/`. We use it because of more user-friendly interface. Also there you can link to a specific version this was taken from (as opposed to `stable`, which can change).
Alternatively, you can just point users to a specific commit and change these links to file paths.
</blockquote></details>

Done.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 36 at r4](https://reviewable.io/reviews/gramineproject/gramine/756#-N71isPZ5Vc9pSODn6uG:-N7EDLZHAvGRJzkXFtio:bffttkq) ([raw file](https://github.com/gramineproject/gramine/blob/8892d2ea3cf99480209c7e77a7e662a691e34ac2/pal/src/arch/x86_64/pal_cpuid.c#L36)):*

I understand these comments are taken directly from kernel source, but do we want to keep the same format? Since we do not parse this like kernel does, we do not need the `"monitor"` here, since we use that string directly in the table.
(ditto for the whole PR)
</blockquote></details>

I think it's reasonable to remove them. Done for all.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 59 at r4](https://reviewable.io/reviews/gramineproject/gramine/756#-N71o9Zr6KRePv1FD4UF:-N7EEdEG0BpuSl-5fKeM:b-214ijf) ([raw file](https://github.com/gramineproject/gramine/blob/8892d2ea3cf99480209c7e77a7e662a691e34ac2/pal/src/arch/x86_64/pal_cpuid.c#L59)):*

Why `""` instead of `NULL`? We wouldn't have to call `strcmp` below
</blockquote></details>

Thanks, actually I also thought about it. The intention was to distinguish a bit from those flags not listed in the spec. I think it's fine to make it NULL (so that we don't need strcmp) but leave the description comment there to differ from those not-listed flags.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 253 at r4](https://reviewable.io/reviews/gramineproject/gramine/756#-N71pRfxEMQcK2ghjyAi:-N7EGOg3AJtJ5cpoAkzi:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/8892d2ea3cf99480209c7e77a7e662a691e34ac2/pal/src/arch/x86_64/pal_cpuid.c#L253)):*

Please add a comment warning that this function does not keep `*flags` null terminated.
</blockquote></details>

Done.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 255 at r4](https://reviewable.io/reviews/gramineproject/gramine/756#-N71p-8HD60LpQfcWRlU:-N7EGPZO6hK6aUMCFAGx:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/8892d2ea3cf99480209c7e77a7e662a691e34ac2/pal/src/arch/x86_64/pal_cpuid.c#L255)):*

What's the point of this assert? The code below works well in such case. `*fmax == 0` case is not handled though, so this could be asserted.
</blockquote></details>

Done.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 257 at r4](https://reviewable.io/reviews/gramineproject/gramine/756#-N71o01y8rdCvTV5NK_Z:-N7EGTw_7pb29TEnxRwc:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/8892d2ea3cf99480209c7e77a7e662a691e34ac2/pal/src/arch/x86_64/pal_cpuid.c#L257)):*
> ```C
>     assert(*flags != NULL);
> 
>     for (int i = 0; i < 32; i++) {
> ```

Done.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 301 at r4](https://reviewable.io/reviews/gramineproject/gramine/756#-N71ril2DxDAUngBybO1:-N7EGtjtBehPIHF64L5j:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/8892d2ea3cf99480209c7e77a7e662a691e34ac2/pal/src/arch/x86_64/pal_cpuid.c#L301)):*

What is `cpuid_level`? Please comment or more preferably change the name to be more descriptive (maybe `max_supported_cpuid`? idk)
</blockquote></details>

Done.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 316 at r4](https://reviewable.io/reviews/gramineproject/gramine/756#-N71rwLa1GyddcGc9o_D:-N7EGuXD1A_GQcsVKhpL:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/8892d2ea3cf99480209c7e77a7e662a691e34ac2/pal/src/arch/x86_64/pal_cpuid.c#L316)):*

ditto (better name)
</blockquote></details>

Done.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 353 at r4](https://reviewable.io/reviews/gramineproject/gramine/756#-N71sagQ4wmhYg8w9_r-:-N7EGvbyBetviqyhOmQj:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/8892d2ea3cf99480209c7e77a7e662a691e34ac2/pal/src/arch/x86_64/pal_cpuid.c#L353)):*

Please remove the leading zeroes, they are just confusing and hard to read (ditto below)
</blockquote></details>

Done.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 360 at r4](https://reviewable.io/reviews/gramineproject/gramine/756#-N71sex4EEWL6e6Wlqda:-N7EHdlqD0OshfUH-HQR:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/8892d2ea3cf99480209c7e77a7e662a691e34ac2/pal/src/arch/x86_64/pal_cpuid.c#L360)):*

ditto
</blockquote></details>

Done.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 368 at r4](https://reviewable.io/reviews/gramineproject/gramine/756#-N71snVB55MRClhATmJM:-N7EHim_-r2fkTqBrWAo:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/8892d2ea3cf99480209c7e77a7e662a691e34ac2/pal/src/arch/x86_64/pal_cpuid.c#L368)):*

You don't need to restate what code does, better explain the part that does not follow from the code, e.g. ``` `words[CPUID_WORD_EAX]` holds the max supported subleaf```
I would also remove the intermediate variable (`eax`) it doesn't make it more readable.
</blockquote></details>

Done.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 377 at r4](https://reviewable.io/reviews/gramineproject/gramine/756#-N71tJafDjX-K5XLuRBa:-N7EHja-7b19RNYejQAX:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/8892d2ea3cf99480209c7e77a7e662a691e34ac2/pal/src/arch/x86_64/pal_cpuid.c#L377)):*

ditto (leading zeros)
</blockquote></details>

Done.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 383 at r4](https://reviewable.io/reviews/gramineproject/gramine/756#-N71uBWF03fmC0AZnpbL:-N7EIG5A1ikFTEYGWTDD:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/8892d2ea3cf99480209c7e77a7e662a691e34ac2/pal/src/arch/x86_64/pal_cpuid.c#L383)):*

Why `AMD-defined`? Intel has these as well.
</blockquote></details>

Done.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 390 at r4](https://reviewable.io/reviews/gramineproject/gramine/756#-N71v47B-7GrJwBg949I:-N7EIIUnBUXdcoLN_D9K:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/8892d2ea3cf99480209c7e77a7e662a691e34ac2/pal/src/arch/x86_64/pal_cpuid.c#L390)):*

Please add a comment that this string is always ended with a space, so we can overwrite it.
</blockquote></details>

Done.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 392 at r4](https://reviewable.io/reviews/gramineproject/gramine/756#-N71v9cT023E5-ltQ18v:-N7EIz9y8Q-tK9QdSbTi:bt68emv) ([raw file](https://github.com/gramineproject/gramine/blob/8892d2ea3cf99480209c7e77a7e662a691e34ac2/pal/src/arch/x86_64/pal_cpuid.c#L392)):*

What if two CPUs are different and support different features?
</blockquote></details>

From HW perspective, it's technically possible as long as the two CPUs meet specific requirements (e.g. same QPI, TDP). From system SW perspective, there might be several optioins (e.g.,  to support only the intersection/union/minium set of features or simply make assumptions that both CPUs have identical feature sets).

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/756)* status: all files reviewed, 15 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 59 at r4](https://reviewable.io/reviews/gramineproject/gramine/756#-N71o9Zr6KRePv1FD4UF:-N7FBs_135lE5lVeea59:bbsy7em) ([raw file](https://github.com/gramineproject/gramine/blob/8892d2ea3cf99480209c7e77a7e662a691e34ac2/pal/src/arch/x86_64/pal_cpuid.c#L59)):*

Thanks, actually I also thought about it. The intention was to distinguish a bit from those flags not listed in the spec. I think it's fine to make it NULL (so that we don't need strcmp) but leave the description comment there to differ from those not-listed flags.
</blockquote></details>

I agree with Kailun, we should definitely keep the line itself. I'm fine with the current implementation.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 392 at r4](https://reviewable.io/reviews/gramineproject/gramine/756#-N71v9cT023E5-ltQ18v:-N7FDSPc9tiVqbLeaEsZ:byxy0dz) ([raw file](https://github.com/gramineproject/gramine/blob/8892d2ea3cf99480209c7e77a7e662a691e34ac2/pal/src/arch/x86_64/pal_cpuid.c#L392)):*
I agree that we should have these `cpu_flags` per NUMA node.

> This string is same for all CPUs currently, so we could just have a global variable.

I would prefer to keep it per-numa. Fortunately, we already have `struct pal_socket_info` to place all those things (as @mkow pointed out in his comment).

> Maybe it should actually be fixed before merging this PR?

I would do it the other way around: first merge this PR, and then create a follow-up PR that moves everything from `struct pal_cpu_info` to `struct pal_socket_info`.

Btw, I checked the length of the string on a new-ish Intel CPU (with a lot of the possible features enumerated), the string is 1,000 chars. So spending a 1KB per thread is indeed not optimal, but also won't kill Gramine. Loosing 1MB for 1,000 threads is acceptible for now, and with a follow-up PR we'll decrease this size again.

TLDR: I wouldn't bother and just merge this PR, and have a follow up that moves it all to per-NUMA.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/756)* status: all files reviewed, 15 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 392 at r4](https://reviewable.io/reviews/gramineproject/gramine/756#-N71v9cT023E5-ltQ18v:-N7FoJfw4GbqvXRcNZAl:b-7pxtlt) ([raw file](https://github.com/gramineproject/gramine/blob/8892d2ea3cf99480209c7e77a7e662a691e34ac2/pal/src/arch/x86_64/pal_cpuid.c#L392)):*
> What if two CPUs are different and support different features?

In the same socket? Look at where that comment in `pal_topology.h` is placed ;) So, @boryspoplawski, I'm not sure what are your doubts here.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/756)* status: all files reviewed, 16 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @kailun-qin)

___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 394 at r5](https://reviewable.io/reviews/gramineproject/gramine/756#-N7FpoF26I_1BrJXcRov:-N7FpoF26I_1BrJXcRow:b-tlsuum) ([raw file](https://github.com/gramineproject/gramine/blob/e0968803e706526395e5d4e01a0304e4ae773e14/pal/src/arch/x86_64/pal_cpuid.c#L394)):*
> ```C
>     flags[flen ? flen - 1 : 0] = 0;
> 
>     ci->cpu_flags = flags;
> ```

I don't like this design. In this PR it's PAL which formats the string which will be then literally printed by LibOS in `/proc/cpuinfo`. I.e. you now emulate part of Linux-specific stuff in PAL, while PAL should be a generic API. It's the LibOS where we should emulate Linux internals.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/756)* status: all files reviewed, 16 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 394 at r5](https://reviewable.io/reviews/gramineproject/gramine/756#-N7FpoF26I_1BrJXcRov:-N7FxLkO66XnKmJjT8lf:b-lcs5n6) ([raw file](https://github.com/gramineproject/gramine/blob/e0968803e706526395e5d4e01a0304e4ae773e14/pal/src/arch/x86_64/pal_cpuid.c#L394)):*

I don't like this design. In this PR it's PAL which formats the string which will be then literally printed by LibOS in `/proc/cpuinfo`. I.e. you now emulate part of Linux-specific stuff in PAL, while PAL should be a generic API. It's the LibOS where we should emulate Linux internals.
</blockquote></details>

Hm, good point. This is the "old" design, where PAL populates all this stuff at initialization. But I see no reason to *not* to move all this logic to LibOS (maybe only `cpu_bogomips` should be kept in PAL?).

But the question is also: do we want to do it in this PR, or in a follow-up PR? Or do we want to close this PR completely and create a new PR that will remove `ci->cpu_flags` from PAL and instead do it somewhere during `libos_init()`?

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/756)* status: all files reviewed, 16 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 394 at r5](https://reviewable.io/reviews/gramineproject/gramine/756#-N7FpoF26I_1BrJXcRov:-N7G393K-zK8XmXoKEET:b-z22tlc) ([raw file](https://github.com/gramineproject/gramine/blob/e0968803e706526395e5d4e01a0304e4ae773e14/pal/src/arch/x86_64/pal_cpuid.c#L394)):*

Hm, good point. This is the "old" design, where PAL populates all this stuff at initialization. But I see no reason to *not* to move all this logic to LibOS (maybe only `cpu_bogomips` should be kept in PAL?).

But the question is also: do we want to do it in this PR, or in a follow-up PR? Or do we want to close this PR completely and create a new PR that will remove `ci->cpu_flags` from PAL and instead do it somewhere during `libos_init()`?
</blockquote></details>

I don't like adding so much code in this PR just to move it somewhere else in the next one. So, I'd change this PR into two - first one moves the flags to LibOS, second one does what the current one does, but in LibOS.
Also, notice that it's _this_ PR which introduces the problem - previously `pal_cpu_info::cpu_flags` were not intended to be used for Linux procfs emulation, it was just some info exposed from PAL.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/756)* status: all files reviewed, 16 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @kailun-qin)

___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 394 at r5](https://reviewable.io/reviews/gramineproject/gramine/756#-N7FpoF26I_1BrJXcRov:-N7G4IJV5uH49VhjfXZy:bisu76m) ([raw file](https://github.com/gramineproject/gramine/blob/e0968803e706526395e5d4e01a0304e4ae773e14/pal/src/arch/x86_64/pal_cpuid.c#L394)):*

I don't like adding so much code in this PR just to move it somewhere else in the next one. So, I'd change this PR into two - first one moves the flags to LibOS, second one does what the current one does, but in LibOS.
Also, notice that it's _this_ PR which introduces the problem - previously `pal_cpu_info::cpu_flags` were not intended to be used for Linux procfs emulation, it was just some info exposed from PAL.
</blockquote></details>

Agreed, fair enough. @kailun-qin Could you do this?

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/756)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @kailun-qin, and @mkow)

___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 392 at r4](https://reviewable.io/reviews/gramineproject/gramine/756#-N71v9cT023E5-ltQ18v:-N7GUfRt6ZjbaFuqkApR:b-5x2lv) ([raw file](https://github.com/gramineproject/gramine/blob/8892d2ea3cf99480209c7e77a7e662a691e34ac2/pal/src/arch/x86_64/pal_cpuid.c#L392)):*
> In the same socket? Look at where that comment in pal_topology.h is placed ;) So, @boryspoplawski, I'm not sure what are your doubts here.

Why does it matter where the comment is places? I asked what if there are 2 CPUs with different features.

> I agree that we should have these cpu_flags per NUMA node.

And then how do we populate them? I guess we have to pin to specific CPU, run all initialization and then unpin? `cpuid` output is untrusted anyway, so I guess tihs is our best bet...
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 354 at r5](https://reviewable.io/reviews/gramineproject/gramine/756#-N7GT_ru8K3gtzQWaA43:-N7GT_ru8K3gtzQWaA44:b-8jwc4j) ([raw file](https://github.com/gramineproject/gramine/blob/e0968803e706526395e5d4e01a0304e4ae773e14/pal/src/arch/x86_64/pal_cpuid.c#L354)):*
> ```C
>     EXTEND_CAP_FLAGS(g_cpu_flags_cpuid_1_edx, CPUID_WORD_EDX);
> 
>     /* Thermal and Power Management Leaf: level 0x00000006 (eax) */
> ```

What about this comment (leading `0`)?
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 361 at r5](https://reviewable.io/reviews/gramineproject/gramine/756#-N7GTcId54dJ-5-f7CNV:-N7GTcIeEM3U-OlI5hM4:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/e0968803e706526395e5d4e01a0304e4ae773e14/pal/src/arch/x86_64/pal_cpuid.c#L361)):*
> ```C
>     }
> 
>     /* Additional Intel-defined flags: level 0x00000007 */
> ```

ditto
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 377 at r5](https://reviewable.io/reviews/gramineproject/gramine/756#-N7GTm-97Uuv7K5vS3jI:-N7GTm-97Uuv7K5vS3jJ:bpfaeva) ([raw file](https://github.com/gramineproject/gramine/blob/e0968803e706526395e5d4e01a0304e4ae773e14/pal/src/arch/x86_64/pal_cpuid.c#L377)):*
> ```C
>     }
> 
>     /* Extended state features: level 0x0000000d */
> ```

ditto (zeroes)

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/756)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @kailun-qin, and @mkow)

___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 394 at r5](https://reviewable.io/reviews/gramineproject/gramine/756#-N7FpoF26I_1BrJXcRov:-N7Gw2W5APikPBJVlYdo:bfyqn0s) ([raw file](https://github.com/gramineproject/gramine/blob/e0968803e706526395e5d4e01a0304e4ae773e14/pal/src/arch/x86_64/pal_cpuid.c#L394)):*

Agreed, fair enough. @kailun-qin Could you do this?
</blockquote></details>

Yes, let me give a shot.

Just to double check - the ask for the 1st PR is to only move the init logic of `ci->cpu_flags` into LibOS (i.e., leaving the inits of other `ci->xxx` as they are in PAL)? Thanks!

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/756)* status: 1 of 2 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, @kailun-qin, and @mkow)

___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 354 at r5](https://reviewable.io/reviews/gramineproject/gramine/756#-N7GT_ru8K3gtzQWaA43:-N7GzrJT04YzhZtDuT96:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/e0968803e706526395e5d4e01a0304e4ae773e14/pal/src/arch/x86_64/pal_cpuid.c#L354)):*

What about this comment (leading `0`)?
</blockquote></details>

Done.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 361 at r5](https://reviewable.io/reviews/gramineproject/gramine/756#-N7GTcId54dJ-5-f7CNV:-N7Gzs-mAzGOujc_tQLY:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/e0968803e706526395e5d4e01a0304e4ae773e14/pal/src/arch/x86_64/pal_cpuid.c#L361)):*

ditto
</blockquote></details>

Done.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 377 at r5](https://reviewable.io/reviews/gramineproject/gramine/756#-N7GTm-97Uuv7K5vS3jI:-N7Gzskf0jwy8eEANkm7:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/e0968803e706526395e5d4e01a0304e4ae773e14/pal/src/arch/x86_64/pal_cpuid.c#L377)):*

ditto (zeroes)
</blockquote></details>

Done.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/756)* status: 1 of 2 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 392 at r4](https://reviewable.io/reviews/gramineproject/gramine/756#-N71v9cT023E5-ltQ18v:-N7H0eOT-gNO1X89z1Ew:b-fl9w85) ([raw file](https://github.com/gramineproject/gramine/blob/8892d2ea3cf99480209c7e77a7e662a691e34ac2/pal/src/arch/x86_64/pal_cpuid.c#L392)):*
> Why does it matter where the comment is places? I asked what if there are 2 CPUs with different features.

Because it's placed in `pal_socket_info` - i.e., this information is planned to be per-socket, which should answer your original question?

> And then how do we populate them? I guess we have to pin to specific CPU, run all initialization and then unpin? `cpuid` output is untrusted anyway, so I guess tihs is our best bet...

Yeah, something like that probably.
___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 394 at r5](https://reviewable.io/reviews/gramineproject/gramine/756#-N7FpoF26I_1BrJXcRov:-N7H03HPEXr8vAImFoxr:b-k49ohv) ([raw file](https://github.com/gramineproject/gramine/blob/e0968803e706526395e5d4e01a0304e4ae773e14/pal/src/arch/x86_64/pal_cpuid.c#L394)):*
> Just to double check - the ask for the 1st PR is to only move the init logic of `ci->cpu_flags` into LibOS (i.e., leaving the inits of other `ci->xxx` as they are in PAL)? Thanks!

Yes and no. If you move this to LibOS, then it shouldn't be stored in `pal_cpu_info`. Maybe just build it when printing `cpuinfo`? (in a helper function) Printing will be a bit slower, but we save global memory, so it's a reasonable deal.

> BTW, do you want me to just close this PR (then to create 2 follow-up PRs as mkow suggested)?

Yup.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/756)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @kailun-qin, and @mkow)

___
*[`pal/src/arch/x86_64/pal_cpuid.c` line 392 at r4](https://reviewable.io/reviews/gramineproject/gramine/756#-N71v9cT023E5-ltQ18v:-N7H3-Sk2Nnr4BplppTe:b-z4k7sv) ([raw file](https://github.com/gramineproject/gramine/blob/8892d2ea3cf99480209c7e77a7e662a691e34ac2/pal/src/arch/x86_64/pal_cpuid.c#L392)):*
> Because it's placed in pal_socket_info - i.e., this information is planned to be per-socket, which should answer your original question?

No it does not. I asked what if 2 CPUs are different - in different sockets (but that's actually irrelevant).

I guess I can answer this myself now: it's basically unsupported atm.

PR #773, opened 2022-07-20
[tools/sgx,curl,Docs] Link all dependencies of `tools/sgx` statically

kailun-qin:
Previously, we built `libsgx_util` as a shared library which is linked into RA-TLS, Secret Provisioning libraries and some other tools.

This patch
- builds `libsgx_util` into a static library
- builds and minimizes `curl` into a static library (leveraging `mbedtls` as the TLS backend)
- links all dependencies of `tools/sgx` statically to avoid pulling in a lot of dependencies.

See https://github.com/gramineproject/gramine/pull/740 and https://github.com/gramineproject/gramine/pull/730 for some previous discussions.

**BEFORE**
```
kailun@kailun-NUC:~/workspace/intel/upstream/gramine/build/tools/sgx/ias-request$ ldd gramine-sgx-ias-request
        linux-vdso.so.1 (0x00007ffd21e84000)
        libsgx_util.so => /home/kailun/workspace/intel/upstream/gramine/build/tools/sgx/ias-request/./../common/libsgx_util.so (0x00007ff62a17d000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007ff629f77000)
        libcurl.so.4 => /lib/x86_64-linux-gnu/libcurl.so.4 (0x00007ff629ee5000)
        /lib64/ld-linux-x86-64.so.2 (0x00007ff62a1f8000)
        libnghttp2.so.14 => /lib/x86_64-linux-gnu/libnghttp2.so.14 (0x00007ff629ebc000)
        libidn2.so.0 => /lib/x86_64-linux-gnu/libidn2.so.0 (0x00007ff629e9b000)
        librtmp.so.1 => /lib/x86_64-linux-gnu/librtmp.so.1 (0x00007ff629e79000)
        libssh.so.4 => /lib/x86_64-linux-gnu/libssh.so.4 (0x00007ff629e0b000)
        libpsl.so.5 => /lib/x86_64-linux-gnu/libpsl.so.5 (0x00007ff629df8000)
        libssl.so.1.1 => /lib/x86_64-linux-gnu/libssl.so.1.1 (0x00007ff629d65000)
        libcrypto.so.1.1 => /lib/x86_64-linux-gnu/libcrypto.so.1.1 (0x00007ff629a8f000)
        libgssapi_krb5.so.2 => /lib/x86_64-linux-gnu/libgssapi_krb5.so.2 (0x00007ff629a42000)
[...]

PR #778, opened 2022-07-22
[LibOS] Emulate size field for fstat

vijaydhanraj:
## Description of the changes 

This is useful when users want to encrypt and decrypt files within the enclave. Without this commit, functions like
`pf_encrypt_files()` and `pf_decrypt_files()` that are part of `libsgx_util.so` will fail when running inside the enclave.

## How to test this PR? 
Please run `LibOS`, `PAL`, `fs` and LTP tests.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/778)

### DIFF
--- a/libos/src/fs/libos_fs_pseudo.c
+++ b/libos/src/fs/libos_fs_pseudo.c
@@ -206,7 +206,35 @@ static dev_t makedev(unsigned int major, unsigned int minor) {
     return dev;
 }
 
-static int pseudo_istat(struct libos_dentry* dent, struct libos_inode* inode, struct stat* buf) {
+static file_off_t pseudo_seek(struct libos_handle* hdl, file_off_t offset, int whence) {
+    file_off_t ret;
+
+    struct pseudo_node* node = hdl->inode->data;
+    switch (node->type) {
+        case PSEUDO_STR: {
+            lock(&hdl->lock);
+            file_off_t pos = hdl->pos;
+            ret = generic_seek(pos, hdl->info.str.mem.size, offset, whence, &pos);
+            if (ret == 0) {
+                hdl->pos = pos;
+                ret = pos;
+            }
+            unlock(&hdl->lock);
+            return ret;
+        }
+
+        case PSEUDO_DEV:
+            if (!node->dev.dev_ops.seek)
+                return -EACCES;
+            return node->dev.dev_ops.seek(hdl, offset, whence);
+
+        default:
+            return -ENOSYS;
+    }
+}
+
+static int pseudo_istat(struct libos_handle* handle, struct libos_dentry* dent,
+                        struct libos_inode* inode, struct stat* buf) {
     memset(buf, 0, sizeof(*buf));
     buf->st_dev = 1;
     buf->st_mode = inode->type | inode->perm;
@@ -241,7 +269,20 @@ static int pseudo_istat(struct libos_dentry* dent, struct libos_inode* inode, st
             buf->st_rdev = makedev(node->dev.major, node->dev.minor);
             buf->st_nlink = 1;
             break;
-        default:
+        case PSEUDO_STR:
+            buf->st_nlink = 1;
+            if (handle) {
+                int ret = pseudo_seek(handle, 0, SEEK_END);
+                if (ret < 0)
+                    return ret;
+                buf->st_size = ret;
+
+                ret = pseudo_seek(handle, 0, SEEK_SET);
+                if (ret < 0)
+                    return ret;
+            }
+            break;
+        default: /* Symbolic link */
             buf->st_nlink = 1;
             break;
     }
@@ -252,11 +293,11 @@ static int pseudo_stat(struct libos_dentry* dent, struct stat* buf) {
     assert(locked(&g_dcache_lock));
     assert(dent->inode);
 
-    return pseudo_istat(dent, dent->inode, buf);
+    return pseudo_istat(NULL/*handle*/, dent, dent->inode, buf);
 }
 
 static int pseudo_hstat(struct libos_handle* handle, struct stat* buf) {
-    return pseudo_istat(handle->dentry, handle->inode, buf);
+    return pseudo_istat(handle, handle->dentry, handle->inode, buf);
 }
 
 static int pseudo_follow_link(struct libos_dentry* dent, char** out_target) {
@@ -379,33 +420,6 @@ static ssize_t pseudo_write(struct libos_handle* hdl, const void* buf, size_t si
             if (!node->dev.dev_ops.write)
                 return -EACCES;
             return node->dev.dev_ops.write(hdl, buf, size);
-
-        default:
-            return -ENOSYS;
-    }
-}
-
-static file_off_t pseudo_seek(struct libos_handle* hdl, file_off_t offset, int whence) {
-    file_off_t ret;
-
-    struct pseudo_node* node = hdl->inode->data;
-    switch (node->type) {
-        case PSEUDO_STR: {
-            lock(&hdl->lock);
-            file_off_t pos = hdl->pos;
-            ret = generic_seek(pos, hdl->info.str.mem.size, offset, whence, &pos);
-            if (ret == 0) {
-                hdl->pos = pos;
-                ret = pos;
-            }
-            unlock(&hdl->lock);
-            return ret;
-        }
-
-        case PSEUDO_DEV:
-            if (!node->dev.dev_ops.seek)
-                return -EACCES;
-            return node->dev.dev_ops.seek(hdl, offset, whence);
 
         default:
             return -ENOSYS;


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/778)* status: 0 of 1 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @vijaydhanraj)

___
*[`-- commits` line 7 at r1](https://reviewable.io/reviews/gramineproject/gramine/778#-N7_-xj10amdYy0sbafI):*
We should just call it `libsgx_util`, without the `.so` suffix. This is because we're modifying it to be a static lib, see https://github.com/gramineproject/gramine/pull/773

We'll fix the commit message during final rebase.
___
*[`libos/src/fs/libos_fs_pseudo.c` line 275 at r1](https://reviewable.io/reviews/gramineproject/gramine/778#-N7_2Kdv4QowyGz8perJ:-N7_2Kdv4QowyGz8perK:b-iqh7gd) ([raw file](https://github.com/gramineproject/gramine/blob/6c9d7b30e21acf99afb5a41cabde7709a653bd02/libos/src/fs/libos_fs_pseudo.c#L275)):*
> ```C
>             buf->st_nlink = 1;
>             if (handle) {
>                 int ret = pseudo_seek(handle, 0, SEEK_END);
> ```

Usage of `pseudo_seek()` here is wrong for many reasons.

First of all, you could replace all this with a simple:
```
if (handle)
    buf->st_size = hdl->info.str.mem.size;
```

Second, there is still a problem that this only works for cases `handle != NULL`, in other words, it works only for *opened* pseudo-files. If a file was opened, written to, and then closed, then this won't work at all and return `st_size = 0`. Which makes no sense.

The root problem seems to be that pseudo-files do *not* save the file size into `inode->size`, as other file systems do (like chroot, encrypted, tmpfs...). For example, here's how tmpfs updates `inode->size`:
- https://github.com/gramineproject/gramine/blob/d0f64b5417cc86267a320e4b72646e48c4fedfa0/libos/src/fs/tmpfs/fs.c#L260
- https://github.com/gramineproject/gramine/blob/d0f64b5417cc86267a320e4b72646e48c4fedfa0/libos/src/fs/tmpfs/fs.c#L286

We need to do something similar in this file. @vijaydhanraj Could you try my suggestion?

@pwmarcz Hopefully you could help us here, since you implemented this functionality. Is my analysis correct?

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/778)* status: 0 of 1 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @pwmarcz)

___
*[`-- commits` line 7 at r1](https://reviewable.io/reviews/gramineproject/gramine/778#-N7_-xj10amdYy0sbafI):*

We should just call it `libsgx_util`, without the `.so` suffix. This is because we're modifying it to be a static lib, see https://github.com/gramineproject/gramine/pull/773

We'll fix the commit message during final rebase.
</blockquote></details>

Yes, thanks.
___
*[`libos/src/fs/libos_fs_pseudo.c` line 275 at r1](https://reviewable.io/reviews/gramineproject/gramine/778#-N7_2Kdv4QowyGz8perJ:-N7aWMvJ6LGXFBciLSoy:b-fkz920) ([raw file](https://github.com/gramineproject/gramine/blob/6c9d7b30e21acf99afb5a41cabde7709a653bd02/libos/src/fs/libos_fs_pseudo.c#L275)):*

Usage of `pseudo_seek()` here is wrong for many reasons.

First of all, you could replace all this with a simple:
```
if (handle)
    buf->st_size = hdl->info.str.mem.size;
```

Second, there is still a problem that this only works for cases `handle != NULL`, in other words, it works only for *opened* pseudo-files. If a file was opened, written to, and then closed, then this won't work at all and return `st_size = 0`. Which makes no sense.

The root problem seems to be that pseudo-files do *not* save the file size into `inode->size`, as other file systems do (like chroot, encrypted, tmpfs...). For example, here's how tmpfs updates `inode->size`:
- https://github.com/gramineproject/gramine/blob/d0f64b5417cc86267a320e4b72646e48c4fedfa0/libos/src/fs/tmpfs/fs.c#L260
- https://github.com/gramineproject/gramine/blob/d0f64b5417cc86267a320e4b72646e48c4fedfa0/libos/src/fs/tmpfs/fs.c#L286

We need to do something similar in this file. @vijaydhanraj Could you try my suggestion?

@pwmarcz Hopefully you could help us here, since you implemented this functionality. Is my analysis correct?
</blockquote></details>

Thanks @dimakuv, will try the suggestions.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/778)* status: 0 of 1 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv)

___
*[`libos/src/fs/libos_fs_pseudo.c` line 275 at r1](https://reviewable.io/reviews/gramineproject/gramine/778#-N7_2Kdv4QowyGz8perJ:-N7bGuSN3jU_RUdav-5a:b5ew9lj) ([raw file](https://github.com/gramineproject/gramine/blob/6c9d7b30e21acf99afb5a41cabde7709a653bd02/libos/src/fs/libos_fs_pseudo.c#L275)):*
First of all, I don't understand the the justification:

>pf_encrypt_files() and pf_decrypt_files() that are part of libsgx_util.so will fail when running inside the enclave.

Does that mean you want to encrypt and decrypt files that are in a `pseudo` type FS? And if so, why not fix these functions? They are part of Gramine.

Also, this change will cause `/proc` in Gramine to behave differently from real Linux `/proc` (which, AFAIK, returns size 0 in both `stat` and `fstat`). Do you want that?

As for setting `inode->size` - we cannot do that, because the content of a pseudo-file is only generated when you open it (and each time you open it, it might generate a different content with different size). This is also the reason why `/proc/` returns 0.

Finally, if you do want to return size from `fstat()`, then @dimakuv's suggestion seems right, and you could ultimately look at `hdl->info.str.mem.size` (but make sure you use the right lock!)

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/778)* status: 0 of 1 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @vijaydhanraj)

___
*[`-- commits` line 4 at r1](https://reviewable.io/reviews/gramineproject/gramine/778#-N7bNb1b9YIpilUkp0Zz):*
This explanation makes no sense to me. Why do we need this PR exactly?

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/778)* status: 0 of 1 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski and @dimakuv)

___
*[`-- commits` line 4 at r1](https://reviewable.io/reviews/gramineproject/gramine/778#-N7bNb1b9YIpilUkp0Zz):*

This explanation makes no sense to me. Why do we need this PR exactly?
</blockquote></details>

@boryspoplawski can you please see my other comment and see if it makes things a little clear?
___
*[`libos/src/fs/libos_fs_pseudo.c` line 275 at r1](https://reviewable.io/reviews/gramineproject/gramine/778#-N7_2Kdv4QowyGz8perJ:-N7bW7wp62et-DRh8BN1:b-tw76gc) ([raw file](https://github.com/gramineproject/gramine/blob/6c9d7b30e21acf99afb5a41cabde7709a653bd02/libos/src/fs/libos_fs_pseudo.c#L275)):*
> Does that mean you want to encrypt and decrypt files that are in a pseudo type FS? 

The idea is an application running inside the enclave may want to encrpt/decrypt files using `libsgx_util` library. But the issue is when trying to load the wrap key, it fails as the size is 0. Please see this,
1. https://github.com/gramineproject/gramine/blob/master/tools/sgx/common/pf_util.c#L243
2. https://github.com/gramineproject/gramine/blob/master/tools/sgx/common/util.c#L66
3. https://github.com/gramineproject/gramine/blob/master/tools/sgx/common/util.c#L57

Since the size returned is 0, `fread` here, https://github.com/gramineproject/gramine/blob/master/tools/sgx/common/util.c#L100 fails and we stop.

> Also, this change will cause /proc in Gramine to behave differently from real Linux /proc (which, AFAIK, returns size 0 in both stat and fstat). Do you want that?

No, but thanks for pointing this out. I'm not sure how we want to tackle this.

> (and each time you open it, it might generate a different content with different size)

Is this true even for `/dev/attesation/keys/default`? I thought this is a pseudo file but the size would be constant.

> Finally, if you do want to return size from fstat(), then @dimakuv's suggestion seems right, and you could ultimately look at hdl->info.str.mem.size (but make sure you use the right lock!)

Got it, thanks.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/778)* status: 0 of 1 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski and @dimakuv)

___
*[`libos/src/fs/libos_fs_pseudo.c` line 275 at r1](https://reviewable.io/reviews/gramineproject/gramine/778#-N7_2Kdv4QowyGz8perJ:-N7be58z2O7w4iyHDOOL:bt1s7cd) ([raw file](https://github.com/gramineproject/gramine/blob/6c9d7b30e21acf99afb5a41cabde7709a653bd02/libos/src/fs/libos_fs_pseudo.c#L275)):*

> Does that mean you want to encrypt and decrypt files that are in a pseudo type FS? 

The idea is an application running inside the enclave may want to encrpt/decrypt files using `libsgx_util` library. But the issue is when trying to load the wrap key, it fails as the size is 0. Please see this,
1. https://github.com/gramineproject/gramine/blob/master/tools/sgx/common/pf_util.c#L243
2. https://github.com/gramineproject/gramine/blob/master/tools/sgx/common/util.c#L66
3. https://github.com/gramineproject/gramine/blob/master/tools/sgx/common/util.c#L57

Since the size returned is 0, `fread` here, https://github.com/gramineproject/gramine/blob/master/tools/sgx/common/util.c#L100 fails and we stop.

> Also, this change will cause /proc in Gramine to behave differently from real Linux /proc (which, AFAIK, returns size 0 in both stat and fstat). Do you want that?

No, but thanks for pointing this out. I'm not sure how we want to tackle this.

> (and each time you open it, it might generate a different content with different size)

Is this true even for `/dev/attesation/keys/default`? I thought this is a pseudo file but the size would be constant.

> Finally, if you do want to return size from fstat(), then @dimakuv's suggestion seems right, and you could ultimately look at hdl->info.str.mem.size (but make sure you use the right lock!)

Got it, thanks.
</blockquote></details>

We should probably change `tools/sgx/pf_util.c` not to use fstat, but hardcode the key length (assuming it's really constant length). This PR definitely does things we don't want to do

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/778)* status: 0 of 1 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski and @dimakuv)

___
*[`libos/src/fs/libos_fs_pseudo.c` line 275 at r1](https://reviewable.io/reviews/gramineproject/gramine/778#-N7_2Kdv4QowyGz8perJ:-N7hOAHd13wOw557t2wY:bkqor9u) ([raw file](https://github.com/gramineproject/gramine/blob/6c9d7b30e21acf99afb5a41cabde7709a653bd02/libos/src/fs/libos_fs_pseudo.c#L275)):*

We should probably change `tools/sgx/pf_util.c` not to use fstat, but hardcode the key length (assuming it's really constant length). This PR definitely does things we don't want to do
</blockquote></details>

Yes, for the case mentioned above if we modify the tool, it should be good enough. And for the case when a directory with many files and other subdirectories is passed, `readdir` and `seek` set of APIs be enough in Gramine? Is my understanding correct?

Do we want to close this PR and open another PR fixing the `pf_util` or change this PR itself?

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/778)* status: 0 of 1 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski and @dimakuv)

___
*[`libos/src/fs/libos_fs_pseudo.c` line 275 at r1](https://reviewable.io/reviews/gramineproject/gramine/778#-N7_2Kdv4QowyGz8perJ:-N7jpvopEGKRRkvQ0FtZ:b-v53zse) ([raw file](https://github.com/gramineproject/gramine/blob/6c9d7b30e21acf99afb5a41cabde7709a653bd02/libos/src/fs/libos_fs_pseudo.c#L275)):*
>  And for the case when a directory with many files and other subdirectories is passed, readdir and seek set of APIs be enough in Gramine?

I did not understand this. What are you talking about?

> Do we want to close this PR and open another PR fixing the pf_util or change this PR itself?

New PR

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/778)* status: 0 of 1 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski and @dimakuv)

___
*[`libos/src/fs/libos_fs_pseudo.c` line 275 at r1](https://reviewable.io/reviews/gramineproject/gramine/778#-N7_2Kdv4QowyGz8perJ:-N7oBRLODMGJdvzMXEPV:be0cf9n) ([raw file](https://github.com/gramineproject/gramine/blob/6c9d7b30e21acf99afb5a41cabde7709a653bd02/libos/src/fs/libos_fs_pseudo.c#L275)):*

>  And for the case when a directory with many files and other subdirectories is passed, readdir and seek set of APIs be enough in Gramine?

I did not understand this. What are you talking about?

> Do we want to close this PR and open another PR fixing the pf_util or change this PR itself?

New PR
</blockquote></details>

Wow, looks like the situation is much more complicated, and this PR is incorrect then...

I suggest to close this PR and open a new one -- there, as Borys suggests, you should modify `tools/sgx/pf_util.c` and just hard-code the size of the key (and drop the `fstat()` calling).

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/778)* status: 0 of 1 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski and @dimakuv)

___
*[`libos/src/fs/libos_fs_pseudo.c` line 275 at r1](https://reviewable.io/reviews/gramineproject/gramine/778#-N7_2Kdv4QowyGz8perJ:-N7pbsfx1fBuOq-xh7D_:b-4k65vu) ([raw file](https://github.com/gramineproject/gramine/blob/6c9d7b30e21acf99afb5a41cabde7709a653bd02/libos/src/fs/libos_fs_pseudo.c#L275)):*
> I did not understand this. What are you talking about?

The above discussion was about loading the wrap key. But once this is done, the application could try to encrypt/decrypt more files, directories, and subdirectories that were passed as input. So, wanted to see if there would be some gaps in getting `stat` information for these.

But it looks like there shouldn't be any issues using `readdir` and `seek` APIs to get this info.

Yes, I will close this one and create a new PR to fix the tool.

> just hard-code the size of the key (and drop the fstat() calling).

@dimakuv why hard code, can't we use `fseek`, yes this would need some more modification but would be more flexible.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/778)* status: 0 of 1 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski and @dimakuv)

___
*[`libos/src/fs/libos_fs_pseudo.c` line 275 at r1](https://reviewable.io/reviews/gramineproject/gramine/778#-N7_2Kdv4QowyGz8perJ:-N7pejjS6eCRhB3ZZ7_U:b5t1apv) ([raw file](https://github.com/gramineproject/gramine/blob/6c9d7b30e21acf99afb5a41cabde7709a653bd02/libos/src/fs/libos_fs_pseudo.c#L275)):*

> I did not understand this. What are you talking about?

The above discussion was about loading the wrap key. But once this is done, the application could try to encrypt/decrypt more files, directories, and subdirectories that were passed as input. So, wanted to see if there would be some gaps in getting `stat` information for these.

But it looks like there shouldn't be any issues using `readdir` and `seek` APIs to get this info.

Yes, I will close this one and create a new PR to fix the tool.

> just hard-code the size of the key (and drop the fstat() calling).

@dimakuv why hard code, can't we use `fseek`, yes this would need some more modification but would be more flexible. 
</blockquote></details>

Yes, Vijay, you're right -- no need to hard-code, you can use `fseek`.

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/778)* status: 0 of 1 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski and @dimakuv)

___
*[`libos/src/fs/libos_fs_pseudo.c` line 275 at r1](https://reviewable.io/reviews/gramineproject/gramine/778#-N7_2Kdv4QowyGz8perJ:-N7ranrX0dC5cGKSwHsG:bnr155c) ([raw file](https://github.com/gramineproject/gramine/blob/6c9d7b30e21acf99afb5a41cabde7709a653bd02/libos/src/fs/libos_fs_pseudo.c#L275)):*

Yes, Vijay, you're right -- no need to hard-code, you can use `fseek`.
</blockquote></details>

It looks like we can't use `fseek` as the `glibc` library implements this using `fstat`. Please see here, https://code.woboq.org/userspace/glibc/libio/fileops.c.html#970.

So, is hardcoding the only option? Any thoughts?

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/778)* status: 0 of 1 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski and @dimakuv)

___
*[`libos/src/fs/libos_fs_pseudo.c` line 275 at r1](https://reviewable.io/reviews/gramineproject/gramine/778#-N7_2Kdv4QowyGz8perJ:-N7rf9z3DN-n5vfMVHQD:b5w0muv) ([raw file](https://github.com/gramineproject/gramine/blob/6c9d7b30e21acf99afb5a41cabde7709a653bd02/libos/src/fs/libos_fs_pseudo.c#L275)):*

It looks like we can't use `fseek` as the `glibc` library implements this using `fstat`. Please see here, https://code.woboq.org/userspace/glibc/libio/fileops.c.html#970.

So, is hardcoding the only option? Any thoughts?
</blockquote></details>

`lseek`

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/778)* status: 0 of 1 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski and @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/778#-N7rx4JQ9QJBffMu-TwK:-N7rx4JQ9QJBffMu-TwL:bux5dfy) (no related file):*
Closing this PR as discussed.
Created https://github.com/gramineproject/gramine/pull/787 to address this issue.

___
___
*[`libos/src/fs/libos_fs_pseudo.c` line 275 at r1](https://reviewable.io/reviews/gramineproject/gramine/778#-N7_2Kdv4QowyGz8perJ:-N7rwsxQ3MbTOtrZaKnV:bwh3jm0) ([raw file](https://github.com/gramineproject/gramine/blob/6c9d7b30e21acf99afb5a41cabde7709a653bd02/libos/src/fs/libos_fs_pseudo.c#L275)):*

`lseek`
</blockquote></details>

Thanks `lseek` worked.

PR #798, opened 2022-07-26
[LibOS,PAL] Emulate CPU affinity in LibOS

boryspoplawski:
## Description of the changes 
New version of #631

Closes #631

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/798)

### DIFF
--- a/common/include/api.h
+++ b/common/include/api.h
@@ -87,8 +87,7 @@ typedef ptrdiff_t ssize_t;
 #define BITS_IN_TYPE(type)              (sizeof(type) * BITS_IN_BYTE)
 #define BITS_TO_UINT32S(nr)             UDIV_ROUND_UP(nr, BITS_IN_TYPE(uint32_t))
 #define BITS_TO_LONGS(nr)               UDIV_ROUND_UP(nr, BITS_IN_TYPE(long))
-/* Note: This macro is not intended for use when nbits == BITS_IN_TYPE(type) */
-#define SET_HIGHEST_N_BITS(type, nbits) (~(((uint64_t)1 << (BITS_IN_TYPE(type) - (nbits))) - 1))
+#define SET_HIGHEST_N_BITS(type, nbits) (nbits ? ~(((type)1 << (BITS_IN_TYPE(type) - (nbits))) - 1) : 0)
 #define WITHIN_MASK(val, mask) (((val) | (mask)) == (mask))
 
 #define IS_ALIGNED(val, alignment)     ((val) % (alignment) == 0)
--- a/libos/include/libos_table.h
+++ b/libos/include/libos_table.h
@@ -154,9 +154,9 @@ long libos_syscall_gettid(void);
 long libos_syscall_tkill(int pid, int sig);
 long libos_syscall_time(time_t* tloc);
 long libos_syscall_futex(int* uaddr, int op, int val, void* utime, int* uaddr2, int val3);
-long libos_syscall_sched_setaffinity(pid_t pid, unsigned int cpumask_size,
+long libos_syscall_sched_setaffinity(pid_t pid, unsigned int user_mask_size,
                                      unsigned long* user_mask_ptr);
-long libos_syscall_sched_getaffinity(pid_t pid, unsigned int cpumask_size,
+long libos_syscall_sched_getaffinity(pid_t pid, unsigned int user_mask_size,
                                      unsigned long* user_mask_ptr);
 long libos_syscall_set_tid_address(int* tidptr);
 long libos_syscall_fadvise64(int fd, loff_t offset, size_t len, int advice);
--- a/libos/include/libos_thread.h
+++ b/libos/include/libos_thread.h
@@ -57,6 +57,8 @@ struct libos_signal_queue {
     struct libos_rt_signal_queue rt_signal_queues[SIGS_CNT - SIGRTMIN + 1];
 };
 
+#define GET_CPU_MASK_LEN() (BITS_TO_LONGS(g_pal_public_state->topo_info.threads_cnt))
+
 DEFINE_LIST(libos_thread);
 DEFINE_LISTP(libos_thread);
 struct libos_thread {
@@ -128,6 +130,8 @@ struct libos_thread {
     void* stack_red;
     libos_tcb_t* libos_tcb;
     void* frameptr;
+
+    unsigned long* cpu_affinity_mask;
 
     refcount_t ref_count;
     struct libos_lock lock;
--- a/libos/src/bookkeep/libos_thread.c
+++ b/libos/src/bookkeep/libos_thread.c
@@ -199,7 +199,7 @@ static int init_main_thread(void) {
                          /*auto_clear=*/true);
     if (ret < 0) {
         put_thread(cur_thread);
-        return pal_to_unix_errno(ret);;
+        return pal_to_unix_errno(ret);
     }
 
     /* TODO: I believe there is some PAL allocated initial stack which could be reused by the first
@@ -211,6 +211,20 @@ static int init_main_thread(void) {
     }
 
     cur_thread->pal_handle = g_pal_public_state->first_thread;
+
+    cur_thread->cpu_affinity_mask = calloc(GET_CPU_MASK_LEN(),
+                                           sizeof(*cur_thread->cpu_affinity_mask));
+    if (!cur_thread->cpu_affinity_mask) {
+        put_thread(cur_thread);
+        return -ENOMEM;
+    }
+    ret = PalThreadGetCpuAffinity(cur_thread->pal_handle, cur_thread->cpu_affinity_mask,
+                                  GET_CPU_MASK_LEN());
+    if (ret < 0) {
+        ret = pal_to_unix_errno(ret);
+        log_error("Failed to get thread CPU affinity mask: %d", ret);
+        return ret;
+    }
 
     set_cur_thread(cur_thread);
     add_thread(cur_thread);
@@ -266,6 +280,12 @@ struct libos_thread* get_new_thread(void) {
         memcpy(thread->groups_info.groups, cur_thread->groups_info.groups, groups_size);
     }
 
+    thread->cpu_affinity_mask = malloc(GET_CPU_MASK_LEN() * sizeof(*thread->cpu_affinity_mask));
+    if (!thread->cpu_affinity_mask) {
+        put_thread(thread);
+        return NULL;
+    }
+
     lock(&cur_thread->lock);
 
     thread->uid       = cur_thread->uid;
@@ -290,6 +310,9 @@ struct libos_thread* get_new_thread(void) {
     struct libos_handle_map* map = get_thread_handle_map(cur_thread);
     assert(map);
     set_handle_map(thread, map);
+
+    memcpy(thread->cpu_affinity_mask, cur_thread->cpu_affinity_mask,
+           GET_CPU_MASK_LEN() * sizeof(*thread->cpu_affinity_mask));
 
     unlock(&cur_thread->lock);
 
@@ -379,6 +402,8 @@ void put_thread(struct libos_thread* thread) {
         if (thread->tid && !is_internal(thread)) {
             release_id(thread->tid);
         }
+
+        free(thread->cpu_affinity_mask);
 
         destroy_pollable_event(&thread->pollable_event);
 
@@ -566,6 +591,11 @@ BEGIN_CP_FUNC(thread) {
 
         new_thread->pal_handle = NULL;
 
+        size_t mask_off = ADD_CP_OFFSET(GET_CPU_MASK_LEN() * sizeof(*thread->cpu_affinity_mask));
+        new_thread->cpu_affinity_mask = (void*)(base + mask_off);
+        memcpy(new_thread->cpu_affinity_mask, thread->cpu_affinity_mask,
+               GET_CPU_MASK_LEN() * sizeof(*thread->cpu_affinity_mask));
+
         memset(&new_thread->pollable_event, 0, sizeof(new_thread->pollable_event));
 
         new_thread->handle_map = NULL;
@@ -643,6 +673,8 @@ BEGIN_RS_FUNC(thread) {
         *thread->set_child_tid = thread->tid;
         thread->set_child_tid = NULL;
     }
+
+    CP_REBASE(thread->cpu_affinity_mask);
 
     assert(!get_cur_thread());
 
--- a/libos/src/sys/libos_sched.c
+++ b/libos/src/sys/libos_sched.c
@@ -14,6 +14,7 @@
 
 #include "api.h"
 #include "libos_internal.h"
+#include "libos_lock.h"
 #include "libos_table.h"
 #include "libos_thread.h"
 #include "pal.h"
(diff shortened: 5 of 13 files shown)


vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/798)* status: 0 of 13 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski)

___
*[`libos/src/sys/libos_sched.c` line 243 at r1](https://reviewable.io/reviews/gramineproject/gramine/798#-N7yWeJn1HoNT3wSAlxV:-N7yWeJoEWslcjTVRTNc:bnb74mj) ([raw file](https://github.com/gramineproject/gramine/blob/8b1866d95a486fbea57e057a4869de3c3a9e5ed9/libos/src/sys/libos_sched.c#L243)):*
> ```C
> 
>     lock(&thread->lock);
>     memcpy(user_mask_ptr, thread->cpu_affinity_mask, GET_CPU_MASK_LEN() * sizeof(unsigned long));
> ```

Will this be an issue if `user_mask_ptr` doesn't have write access beyond `user_mask_size`?
___
*[`pal/src/host/linux-sgx/pal_threading.c` line 157 at r1](https://reviewable.io/reviews/gramineproject/gramine/798#-N7yXhT785jfyx0_-KSP:-N7yXhT785jfyx0_-KSQ:b-991ye9) ([raw file](https://github.com/gramineproject/gramine/blob/8b1866d95a486fbea57e057a4869de3c3a9e5ed9/pal/src/host/linux-sgx/pal_threading.c#L157)):*
> ```C
> 
> int _PalThreadGetCpuAffinity(PAL_HANDLE thread, unsigned long* cpu_mask, size_t cpu_mask_len) {
>     int ret = ocall_sched_getaffinity(thread->thread.tcs, cpu_mask, cpu_mask_len);
> ```

For uniformity between Linux PAL and Linux SGX PAL, can we do `cpu_mask_len -> cpu_mask_len * sizeof(*cpu_mask);` here rather than in the ocall?

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/798)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @vijaydhanraj)

___
*[`libos/src/sys/libos_sched.c` line 174 at r1](https://reviewable.io/reviews/gramineproject/gramine/798#-N7yfLX96gJgF2DTB1jv:-N7zBYC9C-rh_kjpN_2Z:bdhf8p4) ([raw file](https://github.com/gramineproject/gramine/blob/8b1866d95a486fbea57e057a4869de3c3a9e5ed9/libos/src/sys/libos_sched.c#L174)):*
> I don't like that we overwrite the function arg user_mask_size

What's wrong with it? We only shrink it to the size we will actually ever touch.

> Why not simply:

I guess micro optimization.
___
*[`pal/src/host/linux/pal_threading.c` line 259 at r1](https://reviewable.io/reviews/gramineproject/gramine/798#-N7yh1qTEunyqkdGtoeu:-N7zDY8S9BsoFPZjH1hq:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/8b1866d95a486fbea57e057a4869de3c3a9e5ed9/pal/src/host/linux/pal_threading.c#L259)):*

Shouldn't you add this comment to the function above as well? Or just remove here. (You also don't have this comment in SGX version.)
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/pal_threading.c` line 157 at r1](https://reviewable.io/reviews/gramineproject/gramine/798#-N7yXhT785jfyx0_-KSP:-N7zE73p5yJnUhaIO_WN:bpl0g96) ([raw file](https://github.com/gramineproject/gramine/blob/8b1866d95a486fbea57e057a4869de3c3a9e5ed9/pal/src/host/linux-sgx/pal_threading.c#L157)):*

I wouldn't care about these cosmetics.
</blockquote></details>

What's the point?
Also if we ever used this ocall in multiple places, we would have to do the multiplication everywhere
___
*[`pal/src/host/linux-sgx/pal_threading.c` line 175 at r1](https://reviewable.io/reviews/gramineproject/gramine/798#-N7yjWfQ8_UK87UZD0re:-N7zDsIA9HrgabgcnzxD:bc6g9cd) ([raw file](https://github.com/gramineproject/gramine/blob/8b1866d95a486fbea57e057a4869de3c3a9e5ed9/pal/src/host/linux-sgx/pal_threading.c#L175)):*

I wonder if we could move this check in LibOS itself, after `PalThreadGetCpuAffinity`. This check is generic enough, so it can be done in LibOS. Right now it doesn't make sense, but in the future when we have more secure PALs (e.g., Windows SGX) we'll have the exact duplicate of this code snippet in `windows-sgx/pal_threading.c`.

Not blocking though.
</blockquote></details>

Technically we could, but I always assume that whatever comes from PAL API is already trusted (well, to the extent it can be trusted) and removing this from here would violate this assumption.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/798)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @vijaydhanraj)

___
*[`pal/src/host/linux-sgx/pal_threading.c` line 175 at r1](https://reviewable.io/reviews/gramineproject/gramine/798#-N7yjWfQ8_UK87UZD0re:-N7zOYoJBY5Mhw2i5rY5:b5ifuui) ([raw file](https://github.com/gramineproject/gramine/blob/8b1866d95a486fbea57e057a4869de3c3a9e5ed9/pal/src/host/linux-sgx/pal_threading.c#L175)):*

Technically we could, but I always assume that whatever comes from PAL API is already trusted (well, to the extent it can be trusted) and removing this from here would violate this assumption.
</blockquote></details>

Yeah, right. Probably this will go in some PAL-common code. Anyway, this is for the (distant) future.

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/798)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`libos/src/sys/libos_sched.c` line 243 at r1](https://reviewable.io/reviews/gramineproject/gramine/798#-N7yWeJn1HoNT3wSAlxV:-N7zur1R2CxcN5sh_Omp:b71lubi) ([raw file](https://github.com/gramineproject/gramine/blob/8b1866d95a486fbea57e057a4869de3c3a9e5ed9/libos/src/sys/libos_sched.c#L243)):*

@vijaydhanraj I think it is impossible because above we have the check `user_mask_size < GET_CPU_MASK_LEN() * sizeof(unsigned long)`. So at this point we are guaranteed that `user_mask_size` is equal or greater than `GET_CPU_MASK_LEN() * sizeof(unsigned long)`. So we're good here.
</blockquote></details>

ah yes, this shouldn't be an issue.
___
*[`pal/src/host/linux-sgx/enclave_ocalls.c` line 2151 at r2](https://reviewable.io/reviews/gramineproject/gramine/798#-N8-JlqtCngJ_pjzCK8f:-N8-JlqtCngJ_pjzCK8g:b8uyg1o) ([raw file](https://github.com/gramineproject/gramine/blob/05b0a3d80dae21f925b677e611a6509405d683a8/pal/src/host/linux-sgx/enclave_ocalls.c#L2151)):*
> ```C
>     }
> 
>     if (retval % sizeof(*cpu_mask)) {
> ```

Looks like the sched_getaffinity always returns a non-zero positive value on success, https://elixir.bootlin.com/linux/v5.11/source/kernel/sched/core.c#L6628. Should we return failure if `retval` is `zero`?
___
*[`pal/src/host/linux-sgx/enclave_ocalls.c` line 2161 at r2](https://reviewable.io/reviews/gramineproject/gramine/798#-N8-IRXICr3CcgxqlY_v:-N8-IRXKCT-xvasufHLo:b-47kghw) ([raw file](https://github.com/gramineproject/gramine/blob/05b0a3d80dae21f925b677e611a6509405d683a8/pal/src/host/linux-sgx/enclave_ocalls.c#L2161)):*
> ```C
>     }
> 
>     assert((size_t)retval <= cpu_mask_size);
> ```

Is this assert needed? `sgx_copy_to_enclave` seems to have this check already, https://github.com/gramineproject/gramine/blob/master/pal/src/host/linux-sgx/enclave_framework.c#L137
___
*[`pal/src/host/linux-sgx/pal_threading.c` line 157 at r1](https://reviewable.io/reviews/gramineproject/gramine/798#-N7yXhT785jfyx0_-KSP:-N7zvC85AXP6Vg1Hqmwv:b-af8r25) ([raw file](https://github.com/gramineproject/gramine/blob/8b1866d95a486fbea57e057a4869de3c3a9e5ed9/pal/src/host/linux-sgx/pal_threading.c#L157)):*

What's the point?
Also if we ever used this ocall in multiple places, we would have to do the multiplication everywhere
</blockquote></details>

While reading the code, it made me wonder why we don't do this in Linux SGX PAL counterpart but only after checking the `OCALL` it was clear. So, this is just for readability but not blocking.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/798)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @vijaydhanraj)

___
*[`pal/src/host/linux-sgx/enclave_ocalls.c` line 2151 at r2](https://reviewable.io/reviews/gramineproject/gramine/798#-N8-JlqtCngJ_pjzCK8f:-N8-cdCX02qTiUqKnM77:bpfcchq) ([raw file](https://github.com/gramineproject/gramine/blob/05b0a3d80dae21f925b677e611a6509405d683a8/pal/src/host/linux-sgx/enclave_ocalls.c#L2151)):*

Looks like the sched_getaffinity always returns a non-zero positive value on success, https://elixir.bootlin.com/linux/v5.11/source/kernel/sched/core.c#L6628. Should we return failure if `retval` is `zero`?
</blockquote></details>

I thought about this, but I don't think there is any point - the code below handles `0` correctly and malicious host can always lie anyway (i.e. return some arbitrary affinity).
___
*[`pal/src/host/linux-sgx/enclave_ocalls.c` line 2161 at r2](https://reviewable.io/reviews/gramineproject/gramine/798#-N8-IRXICr3CcgxqlY_v:-N8-cpvTAC4ZybvrOevh:b-8gbf31) ([raw file](https://github.com/gramineproject/gramine/blob/05b0a3d80dae21f925b677e611a6509405d683a8/pal/src/host/linux-sgx/enclave_ocalls.c#L2161)):*

Is this assert needed? `sgx_copy_to_enclave` seems to have this check already, https://github.com/gramineproject/gramine/blob/master/pal/src/host/linux-sgx/enclave_framework.c#L137
</blockquote></details>

??? `assert` is not a check. We just state some property of the code. Think about it as a form of a comment.
___
*[`pal/src/host/linux-sgx/pal_threading.c` line 157 at r1](https://reviewable.io/reviews/gramineproject/gramine/798#-N7yXhT785jfyx0_-KSP:-N8-dOoy7PewwYY3nBVp:b-guvfh) ([raw file](https://github.com/gramineproject/gramine/blob/8b1866d95a486fbea57e057a4869de3c3a9e5ed9/pal/src/host/linux-sgx/pal_threading.c#L157)):*
> why we don't do this in Linux SGX PAL counterpart

This is Linux SGX PAL.
If you meant non-sgx PAL, then there are no ocalls there, so no idea what you mean. `_PalThreadGetCpuAffinity` has the same signature in both PALs

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/798)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

___
*[`pal/src/host/linux-sgx/enclave_ocalls.c` line 2151 at r2](https://reviewable.io/reviews/gramineproject/gramine/798#-N8-JlqtCngJ_pjzCK8f:-N8-pvB5-Usj_v33hsW0:bc2d5b5) ([raw file](https://github.com/gramineproject/gramine/blob/05b0a3d80dae21f925b677e611a6509405d683a8/pal/src/host/linux-sgx/enclave_ocalls.c#L2151)):*

I thought about this, but I don't think there is any point - the code below handles `0` correctly and malicious host can always lie anyway (i.e. return some arbitrary affinity).
</blockquote></details>

Yes true, the code handles `0` correctly just that we could avoid some empty copy and `cpumask` check. Not blocking though.
___
*[`pal/src/host/linux-sgx/enclave_ocalls.c` line 2161 at r2](https://reviewable.io/reviews/gramineproject/gramine/798#-N8-IRXICr3CcgxqlY_v:-N8-qGDD7Tvrgab-sJNp:bbismi4) ([raw file](https://github.com/gramineproject/gramine/blob/05b0a3d80dae21f925b677e611a6509405d683a8/pal/src/host/linux-sgx/enclave_ocalls.c#L2161)):*

??? `assert` is not a check. We just state some property of the code. Think about it as a form of a comment.
</blockquote></details>

OK, you want to make it explicit.
___
*[`pal/src/host/linux-sgx/pal_threading.c` line 157 at r1](https://reviewable.io/reviews/gramineproject/gramine/798#-N7yXhT785jfyx0_-KSP:-N8-qQf39NjZZasRrTjX:b99scin) ([raw file](https://github.com/gramineproject/gramine/blob/8b1866d95a486fbea57e057a4869de3c3a9e5ed9/pal/src/host/linux-sgx/pal_threading.c#L157)):*

> why we don't do this in Linux SGX PAL counterpart

This is Linux SGX PAL.
If you meant non-sgx PAL, then there are no ocalls there, so no idea what you mean. `_PalThreadGetCpuAffinity` has the same signature in both PALs
</blockquote></details>

I meant implementation of this function `_PalThreadGetCpuAffinity ` in Linux SGX PAL vs Linux PAL. Linux PAL implementation does the multiplication but in case of Linux SGX PAL, it is done in `OCALL`. Just a thought, please proceed as you feel right.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/798)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

___
*[`pal/src/host/linux-sgx/enclave_ocalls.c` line 2151 at r2](https://reviewable.io/reviews/gramineproject/gramine/798#-N8-JlqtCngJ_pjzCK8f:-N8-uq83AiSIrZL-jnwG:b4q4ha9) ([raw file](https://github.com/gramineproject/gramine/blob/05b0a3d80dae21f925b677e611a6509405d683a8/pal/src/host/linux-sgx/enclave_ocalls.c#L2151)):*

Yes true, the code handles `0` correctly just that we could avoid some empty copy and `cpumask` check. Not blocking though.
</blockquote></details>

No point in optimizing impossible cases.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/798)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

mkow (changes requested):
> Reviewed 11 of 13 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/798)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`common/include/api.h` line 90 at r2](https://reviewable.io/reviews/gramineproject/gramine/798#-N84Md44A8wGHSxtlT8k:-N84Md44A8wGHSxtlT8l:b-8e1cb4) ([raw file](https://github.com/gramineproject/gramine/blob/05b0a3d80dae21f925b677e611a6509405d683a8/common/include/api.h#L90)):*
> ```objc
> #define BITS_TO_UINT32S(nr)             UDIV_ROUND_UP(nr, BITS_IN_TYPE(uint32_t))
> #define BITS_TO_LONGS(nr)               UDIV_ROUND_UP(nr, BITS_IN_TYPE(long))
> /* Note: This macro is not intended for use when nbits == BITS_IN_TYPE(type) */
> ```

Huh, this old comment is wrong, it should originally be about `nbits == 0`. `nbits == BITS_IN_TYPE(type)` already worked correctly...
___
*[`libos/src/sys/libos_sched.c` line 176 at r2](https://reviewable.io/reviews/gramineproject/gramine/798#-N84YrY8DGq0edBiAYPi:-N84YrY8DGq0edBiAYPj:bh920kc) ([raw file](https://github.com/gramineproject/gramine/blob/05b0a3d80dae21f925b677e611a6509405d683a8/libos/src/sys/libos_sched.c#L176)):*
> ```C
>         for (size_t j = 0; j < BITS_IN_TYPE(__typeof__(*cpu_mask)); j++) {
>             size_t thread_idx = i * BITS_IN_TYPE(__typeof__(*cpu_mask)) + j;
>             if (thread_idx >= threads_count) {
> ```

What does Linux do in case there are some higher bits set? Maybe we should zero them before passing to PAL?
___
*[`libos/src/sys/libos_sched.c` line 180 at r2](https://reviewable.io/reviews/gramineproject/gramine/798#-N84ZTEi0-_pqnI_biWM:-N84ZTEj-3LY09nrUwyM:b-9vwfu2) ([raw file](https://github.com/gramineproject/gramine/blob/05b0a3d80dae21f925b677e611a6509405d683a8/libos/src/sys/libos_sched.c#L180)):*
> ```C
>             }
>             if (!g_pal_public_state->topo_info.threads[thread_idx].is_online) {
>                 cpu_mask[i] &= ~(1ul << j);
> ```

I'd rather cast to the proper type, if we desynchronize them then `1ul` won't sign-extend and we'll mask out some high bits.
OTOH, we have many similar places in our codebase, so maybe whatever?



_Code quote:_
```C
1ul
```
___
*[`libos/src/sys/libos_sched.c` line 182 at r2](https://reviewable.io/reviews/gramineproject/gramine/798#-N84ZhlR9iS4iOnG6Frc:-N84ZhlR9iS4iOnG6Frd:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/05b0a3d80dae21f925b677e611a6509405d683a8/libos/src/sys/libos_sched.c#L182)):*
> ```C
>                 cpu_mask[i] &= ~(1ul << j);
>             }
>             if (cpu_mask[i] & (1ul << j)) {
> ```

ditto



_Code quote:_
```C
1ul
```
___
*[`pal/include/pal/pal.h` line 644 at r2](https://reviewable.io/reviews/gramineproject/gramine/798#-N84b91n0HyZ3Ot5I5E3:-N84b91n0HyZ3Ot5I5E4:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/05b0a3d80dae21f925b677e611a6509405d683a8/pal/include/pal/pal.h#L644)):*
> ```objc
>  *
>  * All bit positions exceeding the count of host CPUs are ignored. \p cpu_mask should select at
>  * least one CPU.
> ```





_Suggestion:_
```objc
one online CPU
```
___
*[`pal/include/pal/pal.h` line 657 at r2](https://reviewable.io/reviews/gramineproject/gramine/798#-N84bCrHESRdXpeI8Wsk:-N84bCrHESRdXpeI8Wsl:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/05b0a3d80dae21f925b677e611a6509405d683a8/pal/include/pal/pal.h#L657)):*
> ```objc
>  * \returns 0 on success, negative error code on failure.
>  *
>  * \p cpu_mask must be able to fit all the processors in the host. This function may overwrite less
> ```





_Suggestion:_
```objc
on the host
```
___
*[`pal/include/pal/pal.h` line 658 at r2](https://reviewable.io/reviews/gramineproject/gramine/798#-N85ALdV5WOU710AOK9o:-N85ALdV5WOU710AOK9p:bmspmko) ([raw file](https://github.com/gramineproject/gramine/blob/05b0a3d80dae21f925b677e611a6509405d683a8/pal/include/pal/pal.h#L658)):*
> This function may overwrite less than \p cpu_mask_len entries, so \p cpu_mask should be zero-initialized.

Maybe it would be safer if this very API did this? I know it was like this before, but seems dangerous - if the caller won't zero the buffer (which isn't intuitive that it's required for this API), then on Linux-SGX PAL we'll leak enclave memory.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/798)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[`common/include/api.h` line 90 at r2](https://reviewable.io/reviews/gramineproject/gramine/798#-N84Md44A8wGHSxtlT8k:-N861kSaDQvZkQB_Q_pG:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/05b0a3d80dae21f925b677e611a6509405d683a8/common/include/api.h#L90)):*

Huh, this old comment is wrong, it should originally be about `nbits == 0`. `nbits == BITS_IN_TYPE(type)` already worked correctly...
</blockquote></details>

Done.
___
*[`libos/src/sys/libos_sched.c` line 176 at r2](https://reviewable.io/reviews/gramineproject/gramine/798#-N84YrY8DGq0edBiAYPi:-N8620FV3dRguH9rddnm:b-xvox2m) ([raw file](https://github.com/gramineproject/gramine/blob/05b0a3d80dae21f925b677e611a6509405d683a8/libos/src/sys/libos_sched.c#L176)):*

What does Linux do in case there are some higher bits set? Maybe we should zero them before passing to PAL?
</blockquote></details>

Yeah, I think they are ignored/cleared.
___
*[`libos/src/sys/libos_sched.c` line 180 at r2](https://reviewable.io/reviews/gramineproject/gramine/798#-N84ZTEi0-_pqnI_biWM:-N8630bbCq5uxda0wzLQ:byrx3pw) ([raw file](https://github.com/gramineproject/gramine/blob/05b0a3d80dae21f925b677e611a6509405d683a8/libos/src/sys/libos_sched.c#L180)):*

I'd rather cast to the proper type, if we desynchronize them then `1ul` won't sign-extend and we'll mask out some high bits.
OTOH, we have many similar places in our codebase, so maybe whatever?
</blockquote></details>

Not sure what you mean. Sign-extending unsigned type? Doesnt make sense...

Oh, you mean if `cpu_mask` type becomes larger than `unsigned long` and right site here will have less bits on? Yeah, I don't think we should bother:
1) we probably will never change the type, since it's tied to syscall interface
2) It would be ugly: `cpu_mask[i] &= ~((__typeof__(cpu_mask[i]))1 << j);`
___
*[`pal/include/pal/pal.h` line 644 at r2](https://reviewable.io/reviews/gramineproject/gramine/798#-N84b91n0HyZ3Ot5I5E3:-N864QU7B6AS2gnKcztM:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/05b0a3d80dae21f925b677e611a6509405d683a8/pal/include/pal/pal.h#L644)):*
> ```objc
>  *
>  * All bit positions exceeding the count of host CPUs are ignored. \p cpu_mask should select at
>  * least one CPU.
> ```

Done.
___
*[`pal/include/pal/pal.h` line 657 at r2](https://reviewable.io/reviews/gramineproject/gramine/798#-N84bCrHESRdXpeI8Wsk:-N864V6o7sxha6H1cHWV:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/05b0a3d80dae21f925b677e611a6509405d683a8/pal/include/pal/pal.h#L657)):*
> ```objc
>  * \returns 0 on success, negative error code on failure.
>  *
>  * \p cpu_mask must be able to fit all the processors in the host. This function may overwrite less
> ```

Done.
___
*[`pal/include/pal/pal.h` line 658 at r2](https://reviewable.io/reviews/gramineproject/gramine/798#-N85ALdV5WOU710AOK9o:-N864Zr5CGGUN3RVfr5R:b6rn7pp) ([raw file](https://github.com/gramineproject/gramine/blob/05b0a3d80dae21f925b677e611a6509405d683a8/pal/include/pal/pal.h#L658)):*
> then on Linux-SGX PAL we'll leak enclave memory.

What? Leak to whom and how?

Anyway, done.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/798)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/798)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

___
*[`libos/src/sys/libos_sched.c` line 180 at r2](https://reviewable.io/reviews/gramineproject/gramine/798#-N84ZTEi0-_pqnI_biWM:-N88JTt37A7Rid1qHv6h:b-9swdbc) ([raw file](https://github.com/gramineproject/gramine/blob/05b0a3d80dae21f925b677e611a6509405d683a8/libos/src/sys/libos_sched.c#L180)):*
> Oh, you mean if `cpu_mask` type becomes larger than `unsigned long`

yes
___
*[`pal/include/pal/pal.h` line 658 at r2](https://reviewable.io/reviews/gramineproject/gramine/798#-N85ALdV5WOU710AOK9o:-N88JhRD25GhEIIqLjaJ:blvq4c0) ([raw file](https://github.com/gramineproject/gramine/blob/05b0a3d80dae21f925b677e611a6509405d683a8/pal/include/pal/pal.h#L658)):*
> What? Leak to whom and how?

`ocall_sched_getaffinity()` copies the mask to untrusted memory before doing the OCALL.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/798)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/798)* status: 12 of 13 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @mkow)

___
*[`pal/include/pal/pal.h` line 658 at r2](https://reviewable.io/reviews/gramineproject/gramine/798#-N85ALdV5WOU710AOK9o:-N88wJJaEKkt7mOU9KTL:b9tot2v) ([raw file](https://github.com/gramineproject/gramine/blob/05b0a3d80dae21f925b677e611a6509405d683a8/pal/include/pal/pal.h#L658)):*
> ocall_sched_getaffinity() copies the mask to untrusted memory before doing the OCALL.

What? No it does not, why would it?

Update: wtf, it does copy it. So many reviews it slipped through... I've revisited both ocalls to check if there are no more f***-ups.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/798)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

___
*[`pal/include/pal/pal.h` line 658 at r2](https://reviewable.io/reviews/gramineproject/gramine/798#-N85ALdV5WOU710AOK9o:-N893KttCwRGLDxxLLZv:beaa0cm) ([raw file](https://github.com/gramineproject/gramine/blob/05b0a3d80dae21f925b677e611a6509405d683a8/pal/include/pal/pal.h#L658)):*

> ocall_sched_getaffinity() copies the mask to untrusted memory before doing the OCALL.

What? No it does not, why would it?

Update: wtf, it does copy it. So many reviews it slipped through... I've revisited both ocalls to check if there are no more f***-ups.
</blockquote></details>

Yeah, it was weird and I wasn't sure what was the point of it (although it wasn't directly doing anything incorrect, just weird and risky). Thx for cleaning it up.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/798)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #801, opened 2022-07-27
[LibOS,PAL,Docs] Remove `loader.argv0_override` manifest option

### LINKED ISSUE #795: [LibOS] Remove `loader.argv0_override` and use instead `argv = [libos.entrypoint]`
## Description of the problem

We have a redundant `loader.argv0_override` manifest option.

We should remove it and replace the LibOS-initialization logic to use `argv = [ libos.entrypoint ]` if there are no explicit args provided. In other words, LibOS should start the process with `argv` that contains a single argument, and it is a string taken from `libos.entrypoint` manifest option.

See #722 for discussions.

dimakuv:
## Description of the changes 

This manifest option was always set to `libos.entrypoint` anyway (or abused in one of the LibOS ABI stack tests).

Instead, when there is no argv-handling manifest option specified, Gramine sets `argv = [ libos.entrypoint ]` (or `argv = [ loader.entrypoint ]` if there is no LibOS, as happens in PAL tests).

Note that this only happens in the first process; child processes get argv from the parent as-is.

Fixes #795.

## How to test this PR? 

CI is enough.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/801)

### DIFF
--- a/CI-Examples/busybox/Makefile
+++ b/CI-Examples/busybox/Makefile
@@ -79,7 +79,7 @@ endif
 
 .PHONY: check
 check: all
-	$(GRAMINE) ./busybox echo "Hello" > OUTPUT
+	$(GRAMINE) busybox echo "Hello" > OUTPUT
 	@grep -q "Hello" OUTPUT && echo "[ Success 1/1 ]"
 	@rm OUTPUT
 
--- a/CI-Examples/busybox/busybox.manifest.template
+++ b/CI-Examples/busybox/busybox.manifest.template
@@ -1,7 +1,7 @@
 # Busybox manifest file example
 
 loader.entrypoint = "file:{{ gramine.libos }}"
-libos.entrypoint = "busybox"
+libos.entrypoint = "/busybox"
 
 loader.log_level = "{{ log_level }}"
 
@@ -18,6 +18,7 @@ fs.mounts = [
   { path = "/lib", uri = "file:{{ gramine.runtimedir() }}" },
   { path = "{{ arch_libdir }}", uri = "file:{{ arch_libdir }}" },
   { path = "/etc", uri = "file:/etc" },
+  { path = "/busybox", uri = "file:busybox" },
 ]
 
 sgx.debug = true
--- a/CI-Examples/helloworld/helloworld.manifest.template
+++ b/CI-Examples/helloworld/helloworld.manifest.template
@@ -1,14 +1,14 @@
 # Hello World manifest file example
 
 loader.entrypoint = "file:{{ gramine.libos }}"
-libos.entrypoint = "helloworld"
+libos.entrypoint = "/helloworld"
 loader.log_level = "{{ log_level }}"
-loader.argv0_override = "helloworld"
 
 loader.env.LD_LIBRARY_PATH = "/lib"
 
 fs.mounts = [
   { path = "/lib", uri = "file:{{ gramine.runtimedir() }}" },
+  { path = "/helloworld", uri = "file:helloworld" },
 ]
 
 sgx.debug = true
--- a/CI-Examples/memcached/memcached.manifest.template
+++ b/CI-Examples/memcached/memcached.manifest.template
@@ -1,7 +1,7 @@
 # Memcached manifest file example
 
 loader.entrypoint = "file:{{ gramine.libos }}"
-libos.entrypoint = "memcached"
+libos.entrypoint = "/memcached"
 
 loader.log_level = "{{ log_level }}"
 
@@ -16,6 +16,7 @@ fs.mounts = [
   { path = "{{ arch_libdir }}", uri = "file:{{ arch_libdir }}" },
   { path = "/usr/{{ arch_libdir }}", uri = "file:/usr/{{ arch_libdir }}" },
   { path = "/etc", uri = "file:/etc" },
+  { path = "/memcached", uri = "file:memcached" },
 ]
 
 sgx.debug = true
--- a/CI-Examples/ra-tls-mbedtls/client.manifest.template
+++ b/CI-Examples/ra-tls-mbedtls/client.manifest.template
@@ -1,7 +1,7 @@
 # Client manifest file
 
 loader.entrypoint = "file:{{ gramine.libos }}"
-libos.entrypoint = "client"
+libos.entrypoint = "/client"
 
 loader.log_level = "{{ log_level }}"
 
@@ -19,6 +19,7 @@ fs.mounts = [
   { path = "{{ arch_libdir }}", uri = "file:{{ arch_libdir }}" },
   { path = "/usr{{ arch_libdir }}", uri = "file:/usr{{ arch_libdir }}" },
   { path = "/etc", uri = "file:/etc" },
+  { path = "/client", uri = "file:client" },
 ]
 
 sgx.debug = true
--- a/CI-Examples/ra-tls-mbedtls/server.manifest.template
+++ b/CI-Examples/ra-tls-mbedtls/server.manifest.template
@@ -1,7 +1,7 @@
 # RA-TLS manifest file example
 
 loader.entrypoint = "file:{{ gramine.libos }}"
-libos.entrypoint = "server"
+libos.entrypoint = "/server"
 
 loader.log_level = "{{ log_level }}"
 
@@ -16,6 +16,7 @@ fs.mounts = [
   { path = "{{ arch_libdir }}", uri = "file:{{ arch_libdir }}" },
   { path = "/usr{{ arch_libdir }}", uri = "file:/usr{{ arch_libdir }}" },
   { path = "/etc", uri = "file:/etc" },
+  { path = "/server", uri = "file:server" },
 ]
 
 sgx.debug = true
--- a/CI-Examples/ra-tls-secret-prov/Makefile
+++ b/CI-Examples/ra-tls-secret-prov/Makefile
@@ -153,9 +153,9 @@ secret_prov_pf_client.token: secret_prov_pf_client.sig
 check_epid: app epid files/input.txt
 	./secret_prov_server_epid >/dev/null & SERVER_ID=$$!; \
 	sleep 3; \
-	gramine-sgx ./secret_prov_min_client > OUTPUT; \
-	gramine-sgx ./secret_prov_client >> OUTPUT; \
-	gramine-sgx ./secret_prov_pf_client >> OUTPUT; \
+	gramine-sgx secret_prov_min_client > OUTPUT; \
+	gramine-sgx secret_prov_client >> OUTPUT; \
+	gramine-sgx secret_prov_pf_client >> OUTPUT; \
 	kill -9 $$SERVER_ID;
 	@grep "Received secret = 'ffeeddccbbaa99887766554433221100'" OUTPUT && echo "[ Success 1/4 ]"
 	@grep "Received secret1 = 'ffeeddccbbaa99887766554433221100', secret2 = '42'" OUTPUT && echo "[ Success 2/4 ]"
@@ -167,9 +167,9 @@ check_epid: app epid files/input.txt
 check_dcap: app dcap files/input.txt
 	./secret_prov_server_dcap >/dev/null & SERVER_ID=$$!; \
 	sleep 3; \
-	gramine-sgx ./secret_prov_min_client > OUTPUT; \
-	gramine-sgx ./secret_prov_client >> OUTPUT; \
-	gramine-sgx ./secret_prov_pf_client >> OUTPUT; \
+	gramine-sgx secret_prov_min_client > OUTPUT; \
+	gramine-sgx secret_prov_client >> OUTPUT; \
+	gramine-sgx secret_prov_pf_client >> OUTPUT; \
 	kill -9 $$SERVER_ID;
 	@grep "Received secret = 'ffeeddccbbaa99887766554433221100'" OUTPUT && echo "[ Success 1/4 ]"
 	@grep "Received secret1 = 'ffeeddccbbaa99887766554433221100', secret2 = '42'" OUTPUT && echo "[ Success 2/4 ]"
--- a/CI-Examples/ra-tls-secret-prov/secret_prov_client.manifest.template
+++ b/CI-Examples/ra-tls-secret-prov/secret_prov_client.manifest.template
@@ -1,7 +1,7 @@
 # Secret Provisioning manifest file example (client)
 
 loader.entrypoint = "file:{{ gramine.libos }}"
-libos.entrypoint = "secret_prov_client"
+libos.entrypoint = "/secret_prov_client"
 
 loader.log_level = "{{ log_level }}"
 
@@ -14,6 +14,7 @@ fs.mounts = [
   { path = "{{ arch_libdir }}", uri = "file:{{ arch_libdir }}" },
   { path = "/usr{{ arch_libdir }}", uri = "file:/usr{{ arch_libdir }}" },
   { path = "/etc", uri = "file:/etc" },
+  { path = "/secret_prov_client", uri = "file:secret_prov_client" },
 ]
 
 sgx.enclave_size = "512M"
--- a/CI-Examples/ra-tls-secret-prov/secret_prov_min_client.manifest.template
+++ b/CI-Examples/ra-tls-secret-prov/secret_prov_min_client.manifest.template
@@ -1,7 +1,7 @@
 # Secret Provisioning manifest file example (minimal client)
 
 loader.entrypoint = "file:{{ gramine.libos }}"
-libos.entrypoint = "secret_prov_min_client"
+libos.entrypoint = "/secret_prov_min_client"
 
 loader.log_level = "{{ log_level }}"
 
@@ -18,6 +18,7 @@ fs.mounts = [
   { path = "{{ arch_libdir }}", uri = "file:{{ arch_libdir }}" },
   { path = "/usr{{ arch_libdir }}", uri = "file:/usr{{ arch_libdir }}" },
   { path = "/etc", uri = "file:/etc" },
+  { path = "/secret_prov_min_client", uri = "file:secret_prov_min_client" },
 ]
 
 sgx.enclave_size = "512M"
--- a/CI-Examples/ra-tls-secret-prov/secret_prov_pf_client.manifest.template
+++ b/CI-Examples/ra-tls-secret-prov/secret_prov_pf_client.manifest.template
@@ -1,7 +1,7 @@
 # Secret Provisioning manifest file example (Protected Files client)
 
 loader.entrypoint = "file:{{ gramine.libos }}"
-libos.entrypoint = "secret_prov_pf_client"
+libos.entrypoint = "/secret_prov_pf_client"
 
 loader.log_level = "{{ log_level }}"
 
@@ -20,6 +20,7 @@ fs.mounts = [
   { path = "/usr{{ arch_libdir }}", uri = "file:/usr{{ arch_libdir }}" },
   { path = "/etc", uri = "file:/etc" },
   { path = "/files/input.txt", uri = "file:files/input.txt", type = "encrypted" },
+  { path = "/secret_prov_pf_client", uri = "file:secret_prov_pf_client" },
 ]
 
 sgx.enclave_size = "512M"
--- a/CI-Examples/redis/redis-server.manifest.template
+++ b/CI-Examples/redis/redis-server.manifest.template
@@ -7,7 +7,7 @@
 loader.entrypoint = "file:{{ gramine.libos }}"
 
 # Entrypoint binary which Gramine invokes.
-libos.entrypoint = "redis-server"
+libos.entrypoint = "/redis-server"
 
 # Verbosity of Gramine debug log (none/error/warning/debug/trace/all). Note
 # that GRAMINE_LOG_LEVEL macro is expanded in the Makefile as part of the
@@ -59,6 +59,10 @@ fs.mounts = [
   # Mount host-OS directory to NSS files required by Glibc + NSS libs (in 'uri')
   # into in-Gramine visible directory /etc (in 'path').
   { path = "/etc", uri = "file:/etc" },
+
+  # Mount redis-server executable (located in the current directory) under the
+  # in-Gramine visible root directory.
+  { path = "/redis-server", uri = "file:redis-server" },
 ]
 
 ############################### SGX: GENERAL ##################################
--- a/CI-Examples/rust/README.md
+++ b/CI-Examples/rust/README.md
@@ -15,7 +15,7 @@ make SGX=1
 
 # run the server in Gramine-SGX against an HTTP benchmark
 make SGX=1 start-gramine-server &
-../common_tools/benchmark-http.sh 127.0.0.1:3000
+../common_tools/benchmark-http.sh http://127.0.0.1:3000
 kill -SIGINT %%
 ```
 
--- a/CI-Examples/rust/rust-hyper-http-server.manifest.template
+++ b/CI-Examples/rust/rust-hyper-http-server.manifest.template
@@ -3,7 +3,6 @@
 loader.entrypoint = "file:{{ gramine.libos }}"
 libos.entrypoint = "{{ self_exe }}"
 loader.log_level = "{{ log_level }}"
-loader.argv0_override = "{{ self_exe }}"
 
 loader.env.LD_LIBRARY_PATH = "/lib:{{ arch_libdir }}"
 
--- a/Documentation/manifest-syntax.rst
+++ b/Documentation/manifest-syntax.rst
@@ -110,17 +110,6 @@ Command-line arguments
 
 ::
 
-   loader.argv0_override = "[STRING]"
-
-This syntax specifies an arbitrary string (typically the executable name) that
-will be passed as the first argument (``argv[0]``) to the executable.
-
-If the string is not specified in the manifest, the application will get
-``argv[0]`` from :program:`gramine-direct` or :program:`gramine-sgx`
-invocation.
-
-::
-
    loader.insecure__use_cmdline_argv = true
 
 or
@@ -143,6 +132,9 @@ arguments to be provided at runtime from an external (trusted) source.
    Pointing to an encrypted file is currently not supported, due to the fact
    that encryption key provisioning currently happens after setting up
    arguments.
+
+If none of the above arguments-handling manifest options is specified in the
+manifest, the application will get ``argv = [ <libos.entrypoint value> ]``.
 
 Environment variables
 ^^^^^^^^^^^^^^^^^^^^^
--- a/libos/test/abi/x86_64/stack_arg.manifest.template
+++ b/libos/test/abi/x86_64/stack_arg.manifest.template
@@ -1,6 +1,11 @@
 loader.entrypoint = "file:{{ gramine.libos }}"
 libos.entrypoint = "{{ entrypoint }}"
 
+# The argv source file must set argv[0] as the basename of the entrypoint. This is crucial for
+# stack tests because if argv[0] contains an absolute path to the binary, then tests become
+# unreliable.
+loader.argv_src_file = "file:stack_arg_argv_input"
+
 fs.mounts = [
   { path = "/lib", uri = "file:{{ gramine.runtimedir() }}" },
   { path = "/{{ entrypoint }}", uri = "file:{{ binary_dir }}/{{ entrypoint }}" },
@@ -14,4 +19,8 @@ sgx.trusted_files = [
   "file:{{ gramine.libos }}",
   "file:{{ binary_dir }}/{{ entrypoint }}",
   "file:{{ gramine.runtimedir() }}/",
+]
+
+sgx.allowed_files = [
+  "file:stack_arg_argv_input",
 ]
--- a/libos/test/abi/x86_64/test_entrypoint.py
+++ b/libos/test/abi/x86_64/test_entrypoint.py
@@ -1,5 +1,8 @@
 #!/usr/bin/env python3
 
+import os
+import subprocess
+
 from graminelibos.regression import RegressionTestCase
 
 class TC_00_Entrypoint(RegressionTestCase):
@@ -19,7 +22,14 @@ class TC_00_Entrypoint(RegressionTestCase):
         self.run_binary(['stack'])
 
     def test_060_arg(self):
-        self.run_binary(['stack_arg', 'foo', 'bar'])
+        args = ['stack_arg', 'foo', 'bar']
+        result = subprocess.check_output(['gramine-argv-serializer'] + args)
+        with open('stack_arg_argv_input', 'wb') as f:
+            f.write(result)
+        try:
+            self.run_binary(['stack_arg'])
+        finally:
+            os.remove('stack_arg_argv_input')
 
     def test_070_env(self):
         self.run_binary(['stack_env'])
--- a/libos/test/regression/argv_from_file.manifest.template
+++ b/libos/test/regression/argv_from_file.manifest.template
@@ -3,7 +3,6 @@
 loader.entrypoint = "file:{{ gramine.libos }}"
 libos.entrypoint = "{{ entrypoint }}"
 
-loader.argv0_override = "{{ entrypoint }}"
 loader.env.LD_LIBRARY_PATH = "/lib"
 loader.argv_src_file = "file:argv_test_input"
 
--- a/libos/test/regression/attestation.manifest.template
+++ b/libos/test/regression/attestation.manifest.template
@@ -1,7 +1,6 @@
 loader.entrypoint = "file:{{ gramine.libos }}"
 libos.entrypoint = "{{ entrypoint }}"
 
-loader.argv0_override = "{{ entrypoint }}"
 loader.env.LD_LIBRARY_PATH = "/lib"
 loader.insecure__use_cmdline_argv = true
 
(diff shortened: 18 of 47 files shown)


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/801)* status: 0 of 38 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

___
*[`CI-Examples/rust/README.md` line 18 at r1](https://reviewable.io/reviews/gramineproject/gramine/801#-N7zsbBS745B_Jm4cdht:-N7zsbBS745B_Jm4cdhu:b-62a4kg) ([raw file](https://github.com/gramineproject/gramine/blob/e4eaa4cca061c1cde5f95e02ad21de9e5f1d6003/CI-Examples/rust/README.md#L18)):*
> ```Markdown
> # run the server in Gramine-SGX against an HTTP benchmark
> make SGX=1 start-gramine-server &
> ../common_tools/benchmark-http.sh http://127.0.0.1:3000
> ```

FYI: Just a small bug in the documentation. Doesn't work without this prefix.
___
*[`libos/test/abi/x86_64/manifest.template` line 7 at r1](https://reviewable.io/reviews/gramineproject/gramine/801#-N7zsmVb8C0RhFmjQWam:-N7zsmVb8C0RhFmjQWan:b-aevbfw) ([raw file](https://github.com/gramineproject/gramine/blob/4f8d7446d25711a4983cdf6695f44e4c8a8e01d9/libos/test/abi/x86_64/manifest.template#L7)):*
> ```
> # This is crucial for stack tests as otherwise argv[0] contains an absolute path
> # to the binary, and tests become unreliable.
> loader.argv0_override = "{{ entrypoint }}"
> ```

FYI: This was a hack. After removing `argv0_override`, I had to use a proper `argv_src_file` manifest option, and thus had to create a new manifest file.
___
*[`libos/test/regression/test_libos.py` line 183 at r1](https://reviewable.io/reviews/gramineproject/gramine/801#-N7zswigBnGcNvZtn-w5:-N7zswihAVivZwSWrbfl:b-l5fdz7) ([raw file](https://github.com/gramineproject/gramine/blob/e4eaa4cca061c1cde5f95e02ad21de9e5f1d6003/libos/test/regression/test_libos.py#L183)):*
> ```Python
>             stderr = e.stderr.decode()
>             self.assertIn('argv handling wasn\'t configured in the manifest, but cmdline arguments '
>                           'were specified', stderr)
> ```

FYI: Just added this negative test because we didn't have such a test before. Not really related to this PR.
___
*[`pal/regression/test_pal.py` line 59 at r1](https://reviewable.io/reviews/gramineproject/gramine/801#-N7zt3Cw9RhcTL92oGHf:-N7zt3Cw9RhcTL92oGHg:b-awos5k) ([raw file](https://github.com/gramineproject/gramine/blob/e4eaa4cca061c1cde5f95e02ad21de9e5f1d6003/pal/regression/test_pal.py#L59)):*
> ```Python
>     def test_Process4(self):
>         _, stderr = self.run_binary(['Process4'], timeout=5)
>         self.assertRegex(stderr, r'In process: (.*?)Process4', stderr)
> ```

FYI: Here `argv[0]` is replaced with the `loader.entrypoint` which is the absolute path. So must use Regex.
___
*[`pal/src/pal_main.c` line 433 at r1](https://reviewable.io/reviews/gramineproject/gramine/801#-N7ztBv14sYEVTgLEloi:-N7ztBv14sYEVTgLEloj:b-8rww90) ([raw file](https://github.com/gramineproject/gramine/blob/e4eaa4cca061c1cde5f95e02ad21de9e5f1d6003/pal/src/pal_main.c#L433)):*
> ```C
>     assert(argv0_override);
> 
>     if (use_cmdline_argv && !parent_process) {
> ```

FYI: Note that we use `argv0_override` only in the first process -- child processes are expected to just use args from the parent.
___
*[`pal/src/pal_main.c` line 462 at r1](https://reviewable.io/reviews/gramineproject/gramine/801#-N7ztIVsCIcqh7IxF-r6:-N7ztIVsCIcqh7IxF-r7:b-8rww90) ([raw file](https://github.com/gramineproject/gramine/blob/e4eaa4cca061c1cde5f95e02ad21de9e5f1d6003/pal/src/pal_main.c#L462)):*
> ```C
>             INIT_FAIL("argv handling wasn't configured in the manifest, but cmdline arguments were "
>                       "specified.");
>         } else if (!parent_process) {
> ```

FYI: Note that we use `argv0_override` only in the first process -- child processes are expected to just use args from the parent.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/801)* status: 0 of 48 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners

___
*[`CI-Examples/ra-tls-secret-prov/Makefile` line 158 at r2](https://reviewable.io/reviews/gramineproject/gramine/801#-N8-5PHr04bp3RiQUPs4:-N8-5PHr04bp3RiQUPs5:bfzd0ga) ([raw file](https://github.com/gramineproject/gramine/blob/4f8d7446d25711a4983cdf6695f44e4c8a8e01d9/CI-Examples/ra-tls-secret-prov/Makefile#L158)):*
> ```Makefile
> 	gramine-sgx ./secret_prov_min_client > OUTPUT; \
> 	gramine-sgx ./secret_prov_client >> OUTPUT; \
> 	gramine-sgx ./secret_prov_pf_client >> OUTPUT; \
> ```

FYI: There was a fun bug that we had here. This `secret_prov_pf_client` spawns a child via: `execvp(argv[0], ...)`.

Previously, Gramine logic didn't find an `loader.argv0_override` manifest option, so `argv[0] = ./secret_prov_pf_client`. And `execvp()` worked by accident.

Now, Gramine logic always performs an `argv[0]` override with a `libos.entrypoint` -- which in this example was relative path `secret_prov_pf_client`. And `execvp("secret_prov_pf_client")` failed because `exec` doesn't understand relative paths (tried to find under `/bin/` and `/usr/bin/` but not under current dir).

So to prevent these accidental weirdness in the future, I modified all CI-examples to contain an absolute path in the `libos.entrypoint`. I think it looks much cleaner and more meaningful. And is not prone to whether we add `./` or not during Gramine invocation.

boryspoplawski (changes requested):
> Reviewed 38 of 38 files at r1, 11 of 11 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/801)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`CI-Examples/ra-tls-secret-prov/Makefile` line 158 at r2](https://reviewable.io/reviews/gramineproject/gramine/801#-N8-5PHr04bp3RiQUPs4:-N8-tD0ZBOM6De0VL5vF:bnphdb4) ([raw file](https://github.com/gramineproject/gramine/blob/4f8d7446d25711a4983cdf6695f44e4c8a8e01d9/CI-Examples/ra-tls-secret-prov/Makefile#L158)):*

FYI: There was a fun bug that we had here. This `secret_prov_pf_client` spawns a child via: `execvp(argv[0], ...)`.

Previously, Gramine logic didn't find an `loader.argv0_override` manifest option, so `argv[0] = ./secret_prov_pf_client`. And `execvp()` worked by accident.

Now, Gramine logic always performs an `argv[0]` override with a `libos.entrypoint` -- which in this example was relative path `secret_prov_pf_client`. And `execvp("secret_prov_pf_client")` failed because `exec` doesn't understand relative paths (tried to find under `/bin/` and `/usr/bin/` but not under current dir).

So to prevent these accidental weirdness in the future, I modified all CI-examples to contain an absolute path in the `libos.entrypoint`. I think it looks much cleaner and more meaningful. And is not prone to whether we add `./` or not during Gramine invocation.
</blockquote></details>

I'm confused. `execve` does understand relative paths. Or are you talking about different `exec`?
___
*[`libos/test/abi/x86_64/test_entrypoint.py` line 26 at r1](https://reviewable.io/reviews/gramineproject/gramine/801#-N7zuc7jDZS4ChJUim3-:-N7zuc7k5IgzdiywL5fx:b-c6x3r2) ([raw file](https://github.com/gramineproject/gramine/blob/e4eaa4cca061c1cde5f95e02ad21de9e5f1d6003/libos/test/abi/x86_64/test_entrypoint.py#L26)):*
> ```Python
>     def test_060_arg(self):
>         args = ['stack_arg', 'foo', 'bar']
>         result = subprocess.run(['gramine-argv-serializer'] + args,
> ```

`subprocess.check_output` does what you want
___
*[`libos/test/regression/test_libos.py` line 175 at r1](https://reviewable.io/reviews/gramineproject/gramine/801#-N7zvY-E9wHUQbwo0q_i:-N7zvY-E9wHUQbwo0q_j:b-nm6hsu) ([raw file](https://github.com/gramineproject/gramine/blob/e4eaa4cca061c1cde5f95e02ad21de9e5f1d6003/libos/test/regression/test_libos.py#L175)):*
> ```Python
>     @unittest.skipUnless(ON_X86, 'x86-specific')
>     @unittest.skipIf(USES_MUSL, 'C++ is not supported with musl')
>     def test_111_basic_bootstrapping_cpp_unexpected_arg(self):
> ```

Why this example? Why not something that is not C++ and x86 specific?
___
*[`pal/regression/test_pal.py` line 59 at r1](https://reviewable.io/reviews/gramineproject/gramine/801#-N7zt3Cw9RhcTL92oGHf:-N7zvywo03NJv412kFnN:b-guy6k2) ([raw file](https://github.com/gramineproject/gramine/blob/e4eaa4cca061c1cde5f95e02ad21de9e5f1d6003/pal/regression/test_pal.py#L59)):*

FYI: Here `argv[0]` is replaced with the `loader.entrypoint` which is the absolute path. So must use Regex.
</blockquote></details>

Why the regex is so weird, why not `.*`?
Also do you pass `stderr` at the end?
___
*[`pal/regression/test_pal.py` line 79 at r1](https://reviewable.io/reviews/gramineproject/gramine/801#-N7zwaINELHSgjYkUxuq:-N7zwaINELHSgjYkUxur:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/e4eaa4cca061c1cde5f95e02ad21de9e5f1d6003/pal/regression/test_pal.py#L79)):*
> ```Python
>         # One Argument Given
>         self.assertIn('# of Arguments: 1', stderr)
>         self.assertRegex(stderr, r'argv\[0\] = (.*?)Bootstrap', stderr)
> ```

ditto
___
*[`pal/src/pal_main.c` line 433 at r1](https://reviewable.io/reviews/gramineproject/gramine/801#-N7ztBv14sYEVTgLEloi:-N8-ogR61QIJFOWlvgCc:bi6bf31) ([raw file](https://github.com/gramineproject/gramine/blob/e4eaa4cca061c1cde5f95e02ad21de9e5f1d6003/pal/src/pal_main.c#L433)):*

FYI: Note that we use `argv0_override` only in the first process -- child processes are expected to just use args from the parent.
</blockquote></details>

Hmm, but is it worth special casing? Since child processes are expected to ignore these, we could always overwrite? This could be just an `else` in the `if` below.
___
*[`pal/src/pal_main.c` line 462 at r1](https://reviewable.io/reviews/gramineproject/gramine/801#-N7ztIVsCIcqh7IxF-r6:-N8-qzjs20p3qqm3_CG4:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/e4eaa4cca061c1cde5f95e02ad21de9e5f1d6003/pal/src/pal_main.c#L462)):*

FYI: Note that we use `argv0_override` only in the first process -- child processes are expected to just use args from the parent.
</blockquote></details>

ditto
___
*[`pal/src/pal_main.c` line 464 at r1](https://reviewable.io/reviews/gramineproject/gramine/801#-N8-r9d22LKfA95QT0BZ:-N8-r9d22LKfA95QT0B_:b-ecr659) ([raw file](https://github.com/gramineproject/gramine/blob/e4eaa4cca061c1cde5f95e02ad21de9e5f1d6003/pal/src/pal_main.c#L464)):*
> ```C
>         } else if (!parent_process) {
>             /* no argv handling and no cmdline args -- override argv[0] with entrypoint */
>             arguments = malloc(sizeof(const char*) * 2);
> ```

@dimakuv Weren't you the one disliking overwriting function arguments?
___
*[`pal/src/pal_main.c` line 464 at r1](https://reviewable.io/reviews/gramineproject/gramine/801#-N8-r_uQ7UGNV3VKgQJV:-N8-r_uQ7UGNV3VKgQJW:b17po3x) ([raw file](https://github.com/gramineproject/gramine/blob/e4eaa4cca061c1cde5f95e02ad21de9e5f1d6003/pal/src/pal_main.c#L464)):*
> ```C
>         } else if (!parent_process) {
>             /* no argv handling and no cmdline args -- override argv[0] with entrypoint */
>             arguments = malloc(sizeof(const char*) * 2);
> ```

We will never free this memory. Is that fine? Maybe just allocate it on the stack?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/801)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[`CI-Examples/ra-tls-secret-prov/Makefile` line 158 at r2](https://reviewable.io/reviews/gramineproject/gramine/801#-N8-5PHr04bp3RiQUPs4:-N800KQ2D5WtuH9se2mL:b-ijub96) ([raw file](https://github.com/gramineproject/gramine/blob/4f8d7446d25711a4983cdf6695f44e4c8a8e01d9/CI-Examples/ra-tls-secret-prov/Makefile#L158)):*

I'm confused. `execve` does understand relative paths. Or are you talking about different `exec`?
</blockquote></details>

I'm talking about `execvp`. The man page has this:
```
       The  default search path (used when the environment does not contain the variable PATH) shows some variation across
       systems.  It generally includes /bin and /usr/bin (in that order) and may also include the current  working  direc‐
       tory.  On some other systems, the current working is included after /bin and /usr/bin, as an anti-Trojan-horse mea‐
       sure.  The glibc implementation long followed the traditional default where the current working  directory  is  in‐
       cluded at the start of the search path.  However, some code refactoring during the development of glibc 2.24 caused
       the current working directory to be dropped altogether from the default  search  path.   This  accidental  behavior
       change is considered mildly beneficial, and won't be reverted.
```

Note the sentence: *However, some code refactoring during the development of glibc 2.24 caused the current working directory to be dropped altogether from the default  search  path.*.
___
*[`libos/test/regression/test_libos.py` line 175 at r1](https://reviewable.io/reviews/gramineproject/gramine/801#-N7zvY-E9wHUQbwo0q_i:-N800_V_26DdKCVV8cv0:b-ryle2v) ([raw file](https://github.com/gramineproject/gramine/blob/e4eaa4cca061c1cde5f95e02ad21de9e5f1d6003/libos/test/regression/test_libos.py#L175)):*

Why this example? Why not something that is not C++ and x86 specific?
</blockquote></details>

This one had several characteristics:
- It has its own `.manifest.template` file -- I need this because the generic `manifest.template` uses `loader.insecure__forward_host_args`. 
- It has "bootstrap" in its name, so sounded appropriate.

Nothing else fit the above... What would you recommend? Create a separate `.manifest.template` file?
___
*[`pal/src/pal_main.c` line 433 at r1](https://reviewable.io/reviews/gramineproject/gramine/801#-N7ztBv14sYEVTgLEloi:-N800ynC-Y3Io7yhp6lg:br6ufnb) ([raw file](https://github.com/gramineproject/gramine/blob/e4eaa4cca061c1cde5f95e02ad21de9e5f1d6003/pal/src/pal_main.c#L433)):*
> Since child processes are expected to ignore these, we could always overwrite? 

But we have the infamous PAL regression tests -- they do not ignore the args. (I actually hit a bug when I removed `&& !parent_process` in one of the PAL regression tests that does child creation.)

In general, if someone wants to write an application on top of PAL, then this overriding will be unexpected. That's why I'm special casing here.
___
*[`pal/src/pal_main.c` line 462 at r1](https://reviewable.io/reviews/gramineproject/gramine/801#-N7ztIVsCIcqh7IxF-r6:-N801JpK8_7ETVHvjiKP:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/e4eaa4cca061c1cde5f95e02ad21de9e5f1d6003/pal/src/pal_main.c#L462)):*

ditto
</blockquote></details>

ditto
___
*[`pal/src/pal_main.c` line 464 at r1](https://reviewable.io/reviews/gramineproject/gramine/801#-N8-r9d22LKfA95QT0BZ:-N801KRREcI6T3KFvlZv:bkkzda9) ([raw file](https://github.com/gramineproject/gramine/blob/e4eaa4cca061c1cde5f95e02ad21de9e5f1d6003/pal/src/pal_main.c#L464)):*

@dimakuv Weren't you the one disliking overwriting function arguments?
</blockquote></details>

Yeah, you got me here. I dislike it only in others' PRs :)

So what do you think we should do? Create a copy, or keep it as-is (maybe with a comment) and then in general allow such overwriting of function args?
___
*[`pal/src/pal_main.c` line 464 at r1](https://reviewable.io/reviews/gramineproject/gramine/801#-N8-r_uQ7UGNV3VKgQJV:-N801Y1xEC52BlW6iLB9:bsc6zk7) ([raw file](https://github.com/gramineproject/gramine/blob/e4eaa4cca061c1cde5f95e02ad21de9e5f1d6003/pal/src/pal_main.c#L464)):*

We will never free this memory. Is that fine? Maybe just allocate it on the stack?
</blockquote></details>

The previous code didn't care, and I also don't really care. We can allocate on the stack, sure. What do you prefer?

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/801)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`CI-Examples/ra-tls-secret-prov/Makefile` line 158 at r2](https://reviewable.io/reviews/gramineproject/gramine/801#-N8-5PHr04bp3RiQUPs4:-N80PUIpD6NK6iMH0jKY:b1vldje) ([raw file](https://github.com/gramineproject/gramine/blob/4f8d7446d25711a4983cdf6695f44e4c8a8e01d9/CI-Examples/ra-tls-secret-prov/Makefile#L158)):*

I'm talking about `execvp`. The man page has this:
```
       The  default search path (used when the environment does not contain the variable PATH) shows some variation across
       systems.  It generally includes /bin and /usr/bin (in that order) and may also include the current  working  direc‐
       tory.  On some other systems, the current working is included after /bin and /usr/bin, as an anti-Trojan-horse mea‐
       sure.  The glibc implementation long followed the traditional default where the current working  directory  is  in‐
       cluded at the start of the search path.  However, some code refactoring during the development of glibc 2.24 caused
       the current working directory to be dropped altogether from the default  search  path.   This  accidental  behavior
       change is considered mildly beneficial, and won't be reverted.
```

Note the sentence: *However, some code refactoring during the development of glibc 2.24 caused the current working directory to be dropped altogether from the default  search  path.*.
</blockquote></details>

Well, maybe we should use `execv` then?
___
*[`libos/test/regression/test_libos.py` line 175 at r1](https://reviewable.io/reviews/gramineproject/gramine/801#-N7zvY-E9wHUQbwo0q_i:-N80PYcoADYB-vsupYlQ:b-zf5c7s) ([raw file](https://github.com/gramineproject/gramine/blob/e4eaa4cca061c1cde5f95e02ad21de9e5f1d6003/libos/test/regression/test_libos.py#L175)):*

This one had several characteristics:
- It has its own `.manifest.template` file -- I need this because the generic `manifest.template` uses `loader.insecure__forward_host_args`. 
- It has "bootstrap" in its name, so sounded appropriate.

Nothing else fit the above... What would you recommend? Create a separate `.manifest.template` file?
</blockquote></details>

There are multiple files witch match point 1 e.g. `libos/test/regression/debug_log_file.manifest.template`. We could reuse one of them (which is not x86 specific and is not c++)
___
*[`pal/src/pal_main.c` line 464 at r1](https://reviewable.io/reviews/gramineproject/gramine/801#-N8-r_uQ7UGNV3VKgQJV:-N80QcLQ9H51NWlHnbD6:b-k5aw99) ([raw file](https://github.com/gramineproject/gramine/blob/e4eaa4cca061c1cde5f95e02ad21de9e5f1d6003/pal/src/pal_main.c#L464)):*

The previous code didn't care, and I also don't really care. We can allocate on the stack, sure. What do you prefer?
</blockquote></details>

The previous code did not allocate this dynamically, did it?
___
*[`pal/src/pal_main.c` line 464 at r1](https://reviewable.io/reviews/gramineproject/gramine/801#-N8-r9d22LKfA95QT0BZ:-N80QgcbEqSlSFrQHFuU:b2cycwf) ([raw file](https://github.com/gramineproject/gramine/blob/e4eaa4cca061c1cde5f95e02ad21de9e5f1d6003/pal/src/pal_main.c#L464)):*

Yeah, you got me here. I dislike it only in others' PRs :)

So what do you think we should do? Create a copy, or keep it as-is (maybe with a comment) and then in general allow such overwriting of function args?
</blockquote></details>

Generally I don't mind, depends on the discussion below (if it's originally nor `malloc`ated, then I dislike overwriting it with dynamically allocated)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/801)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[`CI-Examples/ra-tls-secret-prov/Makefile` line 158 at r2](https://reviewable.io/reviews/gramineproject/gramine/801#-N8-5PHr04bp3RiQUPs4:-N80X1-xE57MtuMt8V-k:b-mn0eqr) ([raw file](https://github.com/gramineproject/gramine/blob/4f8d7446d25711a4983cdf6695f44e4c8a8e01d9/CI-Examples/ra-tls-secret-prov/Makefile#L158)):*

Well, maybe we should use `execv` then?
</blockquote></details>

But why are you against the current Makefile change? Or you're not against this change, and you also propose to change `execvp()` to `execv()` in the `secret_prov_pf_client.c`? I mean here: https://github.com/gramineproject/gramine/blob/7e9d6925c9b5ed2a912fda6ad0b261a97cd1b1a3/CI-Examples/ra-tls-secret-prov/src/secret_prov_pf_client.c#L65
___
*[`pal/src/pal_main.c` line 420 at r2](https://reviewable.io/reviews/gramineproject/gramine/801#-N80XPRqEqbVnBD1q-s4:-N80XPRqEqbVnBD1q-s5:b-hzwad0) ([raw file](https://github.com/gramineproject/gramine/blob/4f8d7446d25711a4983cdf6695f44e4c8a8e01d9/pal/src/pal_main.c#L420)):*
> ```C
>         argv0_overridden = true;
>         if (!arguments[0]) {
>             arguments = malloc(sizeof(const char*) * 2);
> ```

FYI: The previous code allocated dynamically here.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/801)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`CI-Examples/ra-tls-secret-prov/Makefile` line 158 at r2](https://reviewable.io/reviews/gramineproject/gramine/801#-N8-5PHr04bp3RiQUPs4:-N80aIwt0tNH-SUvK2R8:b-jedu4d) ([raw file](https://github.com/gramineproject/gramine/blob/4f8d7446d25711a4983cdf6695f44e4c8a8e01d9/CI-Examples/ra-tls-secret-prov/Makefile#L158)):*

But why are you against the current Makefile change? Or you're not against this change, and you also propose to change `execvp()` to `execv()` in the `secret_prov_pf_client.c`? I mean here: https://github.com/gramineproject/gramine/blob/7e9d6925c9b5ed2a912fda6ad0b261a97cd1b1a3/CI-Examples/ra-tls-secret-prov/src/secret_prov_pf_client.c#L65
</blockquote></details>

I just don't like that we add some hidden assumption that `argv[0]` will always be absolute path
___
*[`pal/src/pal_main.c` line 464 at r1](https://reviewable.io/reviews/gramineproject/gramine/801#-N8-r_uQ7UGNV3VKgQJV:-N80_R6y0CWKhhe0bntv:b-dtp37b) ([raw file](https://github.com/gramineproject/gramine/blob/e4eaa4cca061c1cde5f95e02ad21de9e5f1d6003/pal/src/pal_main.c#L464)):*

The previous code did not allocate this dynamically, did it?
</blockquote></details>

Ok, I've checked `arguments` point to stack memory, so I definitely dislike overwriting it with malloced memory.
___
*[`pal/src/pal_main.c` line 464 at r1](https://reviewable.io/reviews/gramineproject/gramine/801#-N8-r9d22LKfA95QT0BZ:-N80_ZzRDeAYbyJEj-7N:bk6m610) ([raw file](https://github.com/gramineproject/gramine/blob/e4eaa4cca061c1cde5f95e02ad21de9e5f1d6003/pal/src/pal_main.c#L464)):*
For this case see the discussion below.

> then in general allow such overwriting of function args?

I just consider function argument a normal variable (well, with some exceptions like out-arguments), so reusing it is fine, as long as it makes sense in the specific case.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/801)* status: 44 of 48 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`CI-Examples/ra-tls-secret-prov/Makefile` line 158 at r2](https://reviewable.io/reviews/gramineproject/gramine/801#-N8-5PHr04bp3RiQUPs4:-N82rn3w68m-1A9RPLjI:b-odvqpp) ([raw file](https://github.com/gramineproject/gramine/blob/4f8d7446d25711a4983cdf6695f44e4c8a8e01d9/CI-Examples/ra-tls-secret-prov/Makefile#L158)):*

I just don't like that we add some hidden assumption that `argv[0]` will always be absolute path
</blockquote></details>

We don't add such an assumption in the Gramine codebase. Gramine will literally override `argv[0] = <libos.entrypoint value>`, nothing else. If `libos.entrypoint = "some/relative/path"`, then this is what the app will see as `argv[0]`.

Or do you mean an assumption *in this particular example* (ra-tls-secret-prov)? If you mean this, then yeah, we have an assumption because our source code for `secret_prov_pf_client.c` uses `execvp(argv[0])` -- this syscall requires an absolute path. And that's why I changed the manifest file for this example (for other examples I changed purely for uniformity, I didn't have to).

So if you mean the latter, sure I can change `execvp()` to `execv()`.
___
*[`libos/test/abi/x86_64/test_entrypoint.py` line 26 at r1](https://reviewable.io/reviews/gramineproject/gramine/801#-N7zuc7jDZS4ChJUim3-:-N82vJCjA0a8UxRY2dKa:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/e4eaa4cca061c1cde5f95e02ad21de9e5f1d6003/libos/test/abi/x86_64/test_entrypoint.py#L26)):*

`subprocess.check_output` does what you want
</blockquote></details>

Done.
___
*[`libos/test/regression/test_libos.py` line 175 at r1](https://reviewable.io/reviews/gramineproject/gramine/801#-N7zvY-E9wHUQbwo0q_i:-N82xnAh2zJMGphE-XYj:b-b25oin) ([raw file](https://github.com/gramineproject/gramine/blob/e4eaa4cca061c1cde5f95e02ad21de9e5f1d6003/libos/test/regression/test_libos.py#L175)):*

There are multiple files witch match point 1 e.g. `libos/test/regression/debug_log_file.manifest.template`. We could reuse one of them (which is not x86 specific and is not c++)
</blockquote></details>

Done.

`debug_log_file` prints all Gramine messages to a `tmp/debug_log_file.log`, so I used instead `debug_log_inline`.
___
*[`pal/regression/test_pal.py` line 59 at r1](https://reviewable.io/reviews/gramineproject/gramine/801#-N7zt3Cw9RhcTL92oGHf:-N82yDBf5Uh90fllcDcu:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/e4eaa4cca061c1cde5f95e02ad21de9e5f1d6003/pal/regression/test_pal.py#L59)):*

Why the regex is so weird, why not `.*`?
Also do you pass `stderr` at the end?
</blockquote></details>

Done.
___
*[`pal/regression/test_pal.py` line 79 at r1](https://reviewable.io/reviews/gramineproject/gramine/801#-N7zwaINELHSgjYkUxuq:-N82yDhi4S4v10JBvV3j:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/e4eaa4cca061c1cde5f95e02ad21de9e5f1d6003/pal/regression/test_pal.py#L79)):*

ditto
</blockquote></details>

Done.
___
*[`pal/src/pal_main.c` line 464 at r1](https://reviewable.io/reviews/gramineproject/gramine/801#-N8-r_uQ7UGNV3VKgQJV:-N82yvKz81rZHqNVelI2:b-5ox222) ([raw file](https://github.com/gramineproject/gramine/blob/e4eaa4cca061c1cde5f95e02ad21de9e5f1d6003/pal/src/pal_main.c#L464)):*

Ok, I've checked `arguments` point to stack memory, so I definitely dislike overwriting it with malloced memory.
</blockquote></details>

Actually:
- In Linux PAL, `arguments` are taken from the stack (as host Linux kernel put them there).
- In Linux-SGX PAL, `arguments` are dynamically allocated.

Well, this is getting annoying. I just realized that `load_cstring_array(argv_src_file, &arguments)` (see comment above) also allocates dynamically. So what do we do now?
___
*[`pal/src/pal_main.c` line 464 at r1](https://reviewable.io/reviews/gramineproject/gramine/801#-N8-r9d22LKfA95QT0BZ:-N82zqRD4sEkJ5zUCRRN:b-egt50t) ([raw file](https://github.com/gramineproject/gramine/blob/e4eaa4cca061c1cde5f95e02ad21de9e5f1d6003/pal/src/pal_main.c#L464)):*

For this case see the discussion below.

> then in general allow such overwriting of function args?

I just consider function argument a normal variable (well, with some exceptions like out-arguments), so reusing it is fine, as long as it makes sense in the specific case.
</blockquote></details>

Ok, I consider this a non-issue then. The current code overwrites `arguments` in many places, so let's just keep it. And I will also use this logic in the future.
___
*[`pal/src/pal_main.c` line 454 at r2](https://reviewable.io/reviews/gramineproject/gramine/801#-N82zaMJ9uJFt-pxFxcb:-N82zaMK9_tVpE4GQbYh:b-ypek1h) ([raw file](https://github.com/gramineproject/gramine/blob/acb2d2d47a5b8b7549f45ba224ba5796adece182/pal/src/pal_main.c#L454)):*
> ```C
>                           "was specified in the manifest.", arguments[0], arguments[1]);
> 
>             ret = load_cstring_array(argv_src_file, &arguments);
> ```

@boryspoplawski Here we also dynamically allocate `arguments`.

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/801)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`CI-Examples/ra-tls-secret-prov/Makefile` line 158 at r2](https://reviewable.io/reviews/gramineproject/gramine/801#-N8-5PHr04bp3RiQUPs4:-N83SYIrCPH5OBQM2CC9:b-htfp8z) ([raw file](https://github.com/gramineproject/gramine/blob/4f8d7446d25711a4983cdf6695f44e4c8a8e01d9/CI-Examples/ra-tls-secret-prov/Makefile#L158)):*

We don't add such an assumption in the Gramine codebase. Gramine will literally override `argv[0] = <libos.entrypoint value>`, nothing else. If `libos.entrypoint = "some/relative/path"`, then this is what the app will see as `argv[0]`.

Or do you mean an assumption *in this particular example* (ra-tls-secret-prov)? If you mean this, then yeah, we have an assumption because our source code for `secret_prov_pf_client.c` uses `execvp(argv[0])` -- this syscall requires an absolute path. And that's why I changed the manifest file for this example (for other examples I changed purely for uniformity, I didn't have to).

So if you mean the latter, sure I can change `execvp()` to `execv()`.
</blockquote></details>

Ok, since it's only the latter, it's ok as it is now
___
*[`pal/src/pal_main.c` line 464 at r1](https://reviewable.io/reviews/gramineproject/gramine/801#-N8-r_uQ7UGNV3VKgQJV:-N83TnWj8os8_-l89azO:b5m8ag5) ([raw file](https://github.com/gramineproject/gramine/blob/e4eaa4cca061c1cde5f95e02ad21de9e5f1d6003/pal/src/pal_main.c#L464)):*

Actually:
- In Linux PAL, `arguments` are taken from the stack (as host Linux kernel put them there).
- In Linux-SGX PAL, `arguments` are dynamically allocated.

Well, this is getting annoying. I just realized that `load_cstring_array(argv_src_file, &arguments)` (see comment above) also allocates dynamically. So what do we do now?
</blockquote></details>

Well, fine, this would require bigger refactoring and adding this line won't increase the difficulty of that refactor in the future

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/801)* status: all files reviewed, 1 unresolved discussion (waiting on @dimakuv)

PR #805, opened 2022-07-30
[tools] Fix OID for RA-TLS

jiazhang0:
The oid parameter of mbedtls_x509write_crt_set_extension() accepts an
encoded OID octet string, rather than an ASN.1 encoded OID object (type 6).

Without the fix, the SGX quote extension in the dump result shows a
weird OID:
            0.6.9.42.840.113741.1337.6:
                ..............r3..L..

The corrected OID should be:
            1.2.840.113741.1337.6:
                ..............r3..L..

Signed-off-by: Jia Zhang <zhang.jia@linux.alibaba.com>

## Description of the changes 

The OID should be canonical and defined correctly. Currently the Evidence Extension format has a wrong OID due to misusing the OID value.

## How to test this PR? 

```shell
cd CI-Examples/ra-tls-mbedtls
vim src/client.c
```

manually apply this patch:
```C
diff --git a/CI-Examples/ra-tls-mbedtls/src/client.c b/CI-Examples/ra-tls-mbedtls/src/client.c
index 4d2f71ba..1655dad2 100644
--- a/CI-Examples/ra-tls-mbedtls/src/client.c
+++ b/CI-Examples/ra-tls-mbedtls/src/client.c
@@ -108,6 +108,10 @@ static int my_verify_measurements(const char* mrenclave, const char* mrsigner,
 static int my_verify_callback(void* data, mbedtls_x509_crt* crt, int depth, uint32_t* flags) {
     (void)data;
 
+    FILE *fp = fopen("/tmp/dump.der", "wb");
+    fwrite(crt->raw.p, crt->raw.len, 1, fp);
+    fclose(fp);
+
     if (depth != 0) {
         /* the cert chain in RA-TLS consists of single self-signed cert, so we expect depth 0 */
         return MBEDTLS_ERR_X509_INVALID_FORMAT;
```

Then dump the certificate and show its content in text:
```shell
make app dcap RA_TYPE=dcap
gramine-sgx ./server &
RA_TLS_ALLOW_DEBUG_ENCLAVE_INSECURE=1 \
RA_TLS_ALLOW_OUTDATED_TCB_INSECURE=1 \
./client dcap
openssl x509 -in /tmp/dump.der -inform der -text | grep 'X509v3 extensions' -A9
```

Results is:
```
        X509v3 extensions:
            X509v3 Basic Constraints: 
                CA:FALSE
            X509v3 Subject Key Identifier: 
                65:23:59:F3:28:2A:5A:DD:79:4E:18:85:48:E6:97:38:73:E6:F7:D6
            X509v3 Authority Key Identifier: 
                keyid:65:23:59:F3:28:2A:5A:DD:79:4E:18:85:48:E6:97:38:73:E6:F7:D6

            0.6.9.42.840.113741.1337.6: 
                ..............r3..L..
```

After applying this patch, the result would be:
```
        X509v3 extensions:
            X509v3 Basic Constraints: 
                CA:FALSE
            X509v3 Subject Key Identifier: 
                3A:09:B0:A2:FA:6C:A8:55:BF:AA:0F:67:74:E9:18:08:E8:B5:2A:E5
            X509v3 Authority Key Identifier: 
                keyid:3A:09:B0:A2:FA:6C:A8:55:BF:AA:0F:67:74:E9:18:08:E8:B5:2A:E5

            1.2.840.113741.1337.6: 
                ..............r3..L..
```

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/805)


kailun-qin (commented):
Thanks for the PR!

Looks reasonable to me since mbedTLS will handle and operate on the ASN.1 format internally.

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/805#-:-N8N-fV6EMpM6vjqWfAy:bpkvccs)* status: 0 of 2 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @jiazhang0)

___
*[`-- commits` line 8 at r1](https://reviewable.io/reviews/gramineproject/gramine/805#-N8MwZRoCjF9KGXZnSGx):*
Typo: weird



_Code quote:_
```
wired
```
___
*[`-- commits` line 8 at r1](https://reviewable.io/reviews/gramineproject/gramine/805#-N8MwlIy-InbLIzrYRCG):*
This is not accurate. `OID tag`is exactly `0x06` only. Suggest to change to `OID`.



_Code quote:_
```
OID tag
```

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/805)* status: 0 of 2 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @jiazhang0)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/805#-N8NKlRv--hDhjy6pe1Z:-N8NKlRw2fMClOiJkVVU:b-1v6up2) (no related file):*
@jiazhang0 @kailun-qin Could you give some links/references about this? How do you know what format mbedTLS expects the OID to be in? What's wrong with the current OID?

I am a dummy in this field, and reading quickly through some OID documentation didn't help.

Also, doesn't this PR break backward-compatibility? How can Gramine with this PR interoperate with e.g. Gramine v1.2?

___

jiazhang0:
@kailun-qin Nice follow up. 

To check out the certificate, modify https://github.com/gramineproject/gramine/blob/master/CI-Examples/ra-tls-mbedtls/src/client.c#L121 and save the content of `crt->raw.p` to a file (e.g, dump.der) and then run `openssl x509 -in dump.der -inform der -text` to show the text.

We can check the validation of a OID value in www.oid-info.com. The expected OID prefix of sgx quote embedded in x509 should be 1.2.840.113741.

jiazhang0:
> > _[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/805)_ status: 0 of 2 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @jiazhang0)
> 
> _[a discussion](https://reviewable.io/reviews/gramineproject/gramine/805#-N8NKlRv--hDhjy6pe1Z:-N8NOFJgAe_SzI2h_8vv:bolwm1u) (no related file):_
> 
> > Could you give some links/references about this? How do you know what format mbedTLS expects the OID to be in? What's wrong with the current OID?
> 
> Currently, `mbedtls_x509write_crt_set_extension()` in our case sets the `quote` octet string to the specified OID in a ASN.1 `Tag(0x06)Length(0x09)Value(0x2A, ...)` aka `TLV` format. Then underneath, mbedTLS simply sets this entire TLV into the `oid.p`: https://github.com/Mbed-TLS/mbedtls/blob/919ff15ecf7b68e7d8ab4d470f3139b9040498e4/library/asn1write.c#L442.
> 
> However, looking at the `oid` field itself: https://github.com/Mbed-TLS/mbedtls/blob/919ff15ecf7b68e7d8ab4d470f3139b9040498e4/include/mbedtls/asn1.h#L152-L158) - it's actually a `mbedtls_asn1_buf` which's already in a ASN.1 TLV format. Hence, setting the entire TLV to `oid.p` does not look reasonable. Instead, we should only set the V part to it (which is proposed by this PR).
> 
> Also, we can see mbedTLS internally manages the TL part of `OID`: https://github.com/Mbed-TLS/mbedtls/blob/919ff15ecf7b68e7d8ab4d470f3139b9040498e4/library/x509_create.c#L336-L339. As the API `int mbedtls_x509write_crt_set_extension( mbedtls_x509write_cert *ctx, const char *oid, size_t oid_len, int critical, const unsigned char *val, size_t val_len )` implies, mbedTLS already knows that it operates on an `oid` of the extension, it's natural that it can manage it itself. Though from document perspective, I didn't find a good reference... (as always)
> 
> > Also, doesn't this PR break backward-compatibility? How can Gramine with this PR interoperate with e.g. Gramine v1.2?
> 
> Good question. Looks like this breaks the backward-compatiblity of Gramine.

I'm not sure what the backward-compatiblity of Gramine exactly means. If we just talk about the backward-compatiblity of  Gramine ra-tls, I think it is not a problem.

The OID detection occurs in ra-tls client side, which searches the OID to detect whether the server side sends a certificate with Evidence Extension.

In old implementation, the ASN.1 encoding for the Evidence Extension OID is 13-byte value `0x06, 0x0B, 0x06, 0x09, 0x2A, 0x86, 0x48, 0x86, 0xF8, 0x4D, 0x8A, 0x39, 0x06`, and client side uses the 11-byte pattern `0x06, 0x09, 0x2A, 0x86, 0x48, 0x86, 0xF8, 0x4D, 0x8A, 0x39, 0x06` in memmem() (see https://github.com/gramineproject/gramine/blob/master/tools/sgx/ra-tls/ra_tls_verify_common.c#L114) to find out it.

In new implementation,  the ASN.1 encoding for the OID becomes 11-byte value `0x06, 0x09, 0x2A, 0x86, 0x48, 0x86, 0xF8, 0x4D, 0x8A, 0x39, 0x06`. The old client side using 11-byte pattern can still find out it. If the server sides uses the 13-byte long value for OID encoding, the new client can still find out it with the 9-byte pattern "0x2A, 0x86, 0x48, 0x86, 0xF8, 0x4D, 0x8A, 0x39, 0x06".

So I think the backward-compatiblity is still kept.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/805)* status: 0 of 2 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @jiazhang0)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/805#-N8NKlRv--hDhjy6pe1Z:-N8RTTY-1DXcOzkBNaxH:b-yd0xaj) (no related file):*
<details><summary><i>Previously, jiazhang0 (Jia Zhang) wrote…</i></summary><blockquote>

I'm not sure what the backward-compatiblity of Gramine exactly means. If we just talk about the backward-compatiblity of  Gramine ra-tls, I think it is not a problem.

The OID detection occurs in ra-tls client side, which searches the OID to detect whether the server side sends a certificate with Evidence Extension.

In old implementation, the ASN.1 encoding for the Evidence Extension OID is 13-byte value `0x06, 0x0B, 0x06, 0x09, 0x2A, 0x86, 0x48, 0x86, 0xF8, 0x4D, 0x8A, 0x39, 0x06`, and client side uses the 11-byte pattern `0x06, 0x09, 0x2A, 0x86, 0x48, 0x86, 0xF8, 0x4D, 0x8A, 0x39, 0x06` in memmem() (see https://github.com/gramineproject/gramine/blob/master/tools/sgx/ra-tls/ra_tls_verify_common.c#L114) to find out it.

In new implementation,  the ASN.1 encoding for the Evidence Extension OID becomes 11-byte value `0x06, 0x09, 0x2A, 0x86, 0x48, 0x86, 0xF8, 0x4D, 0x8A, 0x39, 0x06`. The old client side using 11-byte pattern can still find out it. In another case, the old server sides uses the 13-byte long value for OID encoding, and the new client can still find out it with the 9-byte pattern "0x2A, 0x86, 0x48, 0x86, 0xF8, 0x4D, 0x8A, 0x39, 0x06".

So I think the backward-compatiblity is still kept.

</blockquote></details>

I didn't get a chance to check the client's logic yesterday. I just did it - I agree w/ @jiazhang0 that the backward-compatiblity can be kept as this PR does not alter or extend the content of the OID string (just a substring of the old one) so that `memmem()` should still work.

___

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/805)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @jiazhang0 and @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/805#-N8NKlRv--hDhjy6pe1Z:-N8SM-LY8mPSavJGJkBO:b-2or6wm) (no related file):*

I didn't get a chance to check the client's logic yesterday. I just did it - I agree w/ @jiazhang0 that the backward-compatiblity can be kept as this PR does not alter or extend the content of the OID string (just a substring of the old one) so that `memmem()` should still work.
</blockquote></details>

Thank you @kailun-qin and @jiazhang0, I get it now.

Indeed, the original authors of RA-TLS (it wasn't me) did a mistake of adding the Type and Length ("TL" in the TLV) to the OID constant in the code. Whereas mbedTLS doesn't expect the Type and Length to be submitted, but only the Value: the Type for X.509 extension OIDs is always `0x06`, and the Length is submitted as a function argument:
- https://github.com/gramineproject/gramine/blob/48f4660a4b5e3e469922bb8cef7e99a336c06b59/tools/sgx/ra-tls/ra_tls.h#L39
- https://github.com/gramineproject/gramine/blob/dd07c4a335f1d9c3fa465fe03a63dc79060faaa6/tools/sgx/ra-tls/ra_tls_attest.c#L137

Some other references that helped me:
- https://en.wikipedia.org/wiki/Object_identifier
- https://en.wikipedia.org/wiki/X.690#Identifier_octets
- https://directory.apache.org/api/internal-design-guide/4.1-asn1-tlv.html

The PR looks good to me.

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/805#-N8SN7Tu2O83XM7707dc:-N8SN7Tu2O83XM7707dd:b-36101s) (no related file):*
@mkow @boryspoplawski @woju We'll need a review from one of you on this. Hopefully some of you know OIDs better than me :)

___
___
*[`-- commits` line 8 at r1](https://reviewable.io/reviews/gramineproject/gramine/805#-N8MwZRoCjF9KGXZnSGx):*

Typo: weird
</blockquote></details>

We can fix it during final rebase.
___
*[`-- commits` line 8 at r1](https://reviewable.io/reviews/gramineproject/gramine/805#-N8MwlIy-InbLIzrYRCG):*

This is not accurate. `OID tag`is exactly `0x06` only. Suggest to change to `OID`.
</blockquote></details>

We can fix it during final rebase.
___
*[`-- commits` line 13 at r2](https://reviewable.io/reviews/gramineproject/gramine/805#-N8SLNrm0Yrtb8hmOyIe):*
FYI: In case people wonder what this OID means:

`iso(1) member-body(2) us(840) 113741 1337? 6?`

1. ISO
2. Member-body (country)
3. USA
4. `113741` -- Intel Corporation
5. The last two numbers are not standardized, just some random numbers to form a unique OID

dimakuv:
I created #1039 for new standard Interoperable RA-TLS. I am closing this PR now, since the current RA-TLS implementation is legacy anyway.

PR #813, opened 2022-08-02
[tools/sgx] Improve public API of Secret Prov lib

dimakuv:
## Description of the changes 

This is a breaking change that modifies the API of the Secret Prov lib as follows:

- `secret_provision_start()` allocates a context that should be supplied to all other Secret Prov APIs and must be closed afterwards.

- `secret_provision_get()` requires a context as the first argument, and extracts the secret that is associated with this context. Thus, it allows for several secrets being delivered in parallel from different remote parties (previously, the secret was a global variable).

- `secret_provision_destroy()` is removed. Instead, users should call `secret_provision_close()` that will both close the session and destroy the secret.

With these changes, the context ceases to be global, and the library becomes thread-safe. Also, leaks of mbedTLS resources are prevented.

As a small side fix, the library always returns normal POSIX error codes (previously, the library could return mbedTLS error codes).

Fixes #646. Supersedes #647.

## How to test this PR? 

CI should be enough. But also see manual way in #647 to test resource/memory leaks.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/813)

### DIFF
--- a/tools/sgx/ra-tls/secret_prov_attest.c
+++ b/tools/sgx/ra-tls/secret_prov_attest.c
@@ -31,45 +31,71 @@
 
 #include "ra_tls.h"
 #include "secret_prov.h"
+#include "secret_prov_common.h"
 #include "util.h"
 
-/* these are globals because the user may continue using the SSL session even after invoking
- * secret_provision_start() (in the user-supplied callback) */
-static mbedtls_ctr_drbg_context g_ctr_drbg;
-static mbedtls_entropy_context g_entropy;
-static mbedtls_ssl_config g_conf;
-static mbedtls_x509_crt g_verifier_ca_chain;
-static mbedtls_net_context g_verifier_fd;
-static mbedtls_ssl_context g_ssl;
-static mbedtls_pk_context g_my_ratls_key;
-static mbedtls_x509_crt g_my_ratls_cert;
+struct ra_tls_ctx {
+    mbedtls_ssl_context* ssl;
+    mbedtls_net_context* net;
+    mbedtls_ssl_config* conf;
+    uint8_t* secret;
+    size_t   secret_size;
+};
 
-static uint8_t* provisioned_secret = NULL;
-static size_t provisioned_secret_size = 0;
-
-int secret_provision_get(uint8_t** out_secret, size_t* out_secret_size) {
-    if (!out_secret || !out_secret_size)
+int secret_provision_get(struct ra_tls_ctx* ctx, uint8_t** out_secret, size_t* out_secret_size) {
+    if (!ctx || !out_secret || !out_secret_size)
         return -EINVAL;
 
-    *out_secret      = provisioned_secret;
-    *out_secret_size = provisioned_secret_size;
+    uint8_t* secret_copy = malloc(ctx->secret_size);
+    if (!secret_copy)
+        return -ENOMEM;
+
+    memcpy(secret_copy, ctx->secret, ctx->secret_size);
+    *out_secret      = secret_copy;
+    *out_secret_size = ctx->secret_size;
     return 0;
 }
 
-void secret_provision_destroy(void) {
-    if (provisioned_secret && provisioned_secret_size)
+int secret_provision_write(struct ra_tls_ctx* ctx, const uint8_t* buf, size_t size) {
+    if (!ctx || (size && !buf))
+        return -EINVAL;
+    return secret_provision_common_write(ctx->ssl, buf, size);
+}
+
+int secret_provision_read(struct ra_tls_ctx* ctx, uint8_t* buf, size_t size) {
+    if (!ctx || (size && !buf))
+        return -EINVAL;
+    return secret_provision_common_read(ctx->ssl, buf, size);
+}
+
+int secret_provision_close(struct ra_tls_ctx* ctx) {
+    if (!ctx)
+        return -EINVAL;
+
+    if (ctx->secret && ctx->secret_size) {
 #ifdef __STDC_LIB_EXT1__
-        memset_s(provisioned_secret, 0, provisioned_secret_size);
+        memset_s(ctx->secret, 0, ctx->secret_size);
 #else
-        memset(provisioned_secret, 0, provisioned_secret_size);
+        memset(ctx->secret, 0, ctx->secret_size);
 #endif
-    free(provisioned_secret);
-    provisioned_secret      = NULL;
-    provisioned_secret_size = 0;
+    }
+
+    int ret = secret_provision_common_close(ctx->ssl);
+
+    mbedtls_ssl_free(ctx->ssl);
+    mbedtls_ssl_config_free(ctx->conf);
+    mbedtls_net_free(ctx->net);
+    free(ctx->ssl);
+    free(ctx->conf);
+    free(ctx->net);
+
+    free(ctx->secret);
+    free(ctx);
+    return ret;
 }
 
 int secret_provision_start(const char* in_servers, const char* in_ca_chain_path,
-                           struct ra_tls_ctx* out_ctx) {
+                           struct ra_tls_ctx** out_ctx) {
     int ret;
 
     char* servers       = NULL;
@@ -78,28 +104,62 @@ int secret_provision_start(const char* in_servers, const char* in_ca_chain_path,
     char* connected_addr = NULL;
     char* connected_port = NULL;
 
-    mbedtls_ctr_drbg_init(&g_ctr_drbg);
-    mbedtls_entropy_init(&g_entropy);
-    mbedtls_x509_crt_init(&g_verifier_ca_chain);
+    uint8_t* secret = NULL;
 
-    mbedtls_pk_init(&g_my_ratls_key);
-    mbedtls_x509_crt_init(&g_my_ratls_cert);
+    mbedtls_ctr_drbg_context ctr_drbg;
+    mbedtls_entropy_context entropy;
+    mbedtls_x509_crt verifier_ca_chain;
+    mbedtls_pk_context my_ratls_key;
+    mbedtls_x509_crt my_ratls_cert;
 
-    mbedtls_net_init(&g_verifier_fd);
-    mbedtls_ssl_config_init(&g_conf);
-    mbedtls_ssl_init(&g_ssl);
+    if (!out_ctx)
+        return -EINVAL;
+
+    mbedtls_net_context* net  = malloc(sizeof(*net));
+    mbedtls_ssl_config*  conf = malloc(sizeof(*conf));
+    mbedtls_ssl_context* ssl  = malloc(sizeof(*ssl));
+    struct ra_tls_ctx*   ctx  = malloc(sizeof(*ctx));
+
+    if (!net || !conf || !ssl || !ctx) {
+        free(net);
+        free(conf);
+        free(ssl);
+        free(ctx);
+        return -ENOMEM;
+    }
+
+    ctx->ssl         = ssl;
+    ctx->conf        = conf;
+    ctx->net         = net;
+    ctx->secret      = NULL;
+    ctx->secret_size = 0;
+
+    mbedtls_ctr_drbg_init(&ctr_drbg);
+    mbedtls_entropy_init(&entropy);
+    mbedtls_x509_crt_init(&verifier_ca_chain);
+    mbedtls_pk_init(&my_ratls_key);
+    mbedtls_x509_crt_init(&my_ratls_cert);
+
+    mbedtls_net_init(net);
+    mbedtls_ssl_config_init(conf);
+    mbedtls_ssl_init(ssl);
 
     const char* pers = "secret-provisioning";
-    ret = mbedtls_ctr_drbg_seed(&g_ctr_drbg, mbedtls_entropy_func, &g_entropy,
+    ret = mbedtls_ctr_drbg_seed(&ctr_drbg, mbedtls_entropy_func, &entropy,
                                 (const uint8_t*)pers, strlen(pers));
     if (ret < 0) {
+        ERROR("Secret Provisioning failed during mbedtls_ctr_drbg_seed with error %d\n", ret);
+        ret = -EPERM;
         goto out;
     }
 
     if (!in_ca_chain_path) {
         in_ca_chain_path = getenv(SECRET_PROVISION_CA_CHAIN_PATH);
-        if (!in_ca_chain_path)
-            return -EINVAL;
+        if (!in_ca_chain_path) {
+            ERROR("Secret Provisioning could not find envvar " SECRET_PROVISION_CA_CHAIN_PATH "\n");
+            ret = -EINVAL;
+            goto out;
+        }
     }
 
     ca_chain_path = strdup(in_ca_chain_path);
@@ -124,7 +184,7 @@ int secret_provision_start(const char* in_servers, const char* in_ca_chain_path,
     char* saveptr2 = NULL;
     char* str1;
     for (str1 = servers; /*no condition*/; str1 = NULL) {
-        ret = -ECONNREFUSED;
+        ret = MBEDTLS_ERR_NET_CONNECT_FAILED;
         char* token = strtok_r(str1, ",; ", &saveptr1);
         if (!token)
             break;
@@ -137,78 +197,92 @@ int secret_provision_start(const char* in_servers, const char* in_ca_chain_path,
         if (!connected_port)
             continue;
 
-        ret = mbedtls_net_connect(&g_verifier_fd, connected_addr, connected_port,
-                                  MBEDTLS_NET_PROTO_TCP);
+        ret = mbedtls_net_connect(net, connected_addr, connected_port, MBEDTLS_NET_PROTO_TCP);
         if (!ret)
             break;
     }
 
     if (ret < 0) {
+        ERROR("Secret Provisioning could not connect to any of the servers specified in "
+              SECRET_PROVISION_SERVERS "; last mbedTLS error was %d\n", ret);
         goto out;
     }
 
-    ret = mbedtls_ssl_config_defaults(&g_conf, MBEDTLS_SSL_IS_CLIENT, MBEDTLS_SSL_TRANSPORT_STREAM,
+    ret = mbedtls_ssl_config_defaults(conf, MBEDTLS_SSL_IS_CLIENT, MBEDTLS_SSL_TRANSPORT_STREAM,
                                       MBEDTLS_SSL_PRESET_DEFAULT);
     if (ret < 0) {
+        ERROR("Secret Provisioning failed during mbedtls_ssl_config_defaults with error %d\n", ret);
+        ret = -EPERM;
         goto out;
     }
 
-    ret = mbedtls_x509_crt_parse_file(&g_verifier_ca_chain, ca_chain_path);
+    ret = mbedtls_x509_crt_parse_file(&verifier_ca_chain, ca_chain_path);
     if (ret != 0) {
+        ERROR("Secret Provisioning failed during mbedtls_x509_crt_parse_file with error %d\n", ret);
+        ret = -EPERM;
         goto out;
     }
 
     char crt_issuer[256];
-    ret = mbedtls_x509_dn_gets(crt_issuer, sizeof(crt_issuer), &g_verifier_ca_chain.issuer);
+    ret = mbedtls_x509_dn_gets(crt_issuer, sizeof(crt_issuer), &verifier_ca_chain.issuer);
     if (ret < 0) {
+        ERROR("Secret Provisioning failed during mbedtls_x509_dn_gets with error %d\n", ret);
+        ret = -EPERM;
         goto out;
     }
 
-    mbedtls_ssl_conf_authmode(&g_conf, MBEDTLS_SSL_VERIFY_REQUIRED);
-    mbedtls_ssl_conf_ca_chain(&g_conf, &g_verifier_ca_chain, NULL);
+    mbedtls_ssl_conf_authmode(conf, MBEDTLS_SSL_VERIFY_REQUIRED);
+    mbedtls_ssl_conf_ca_chain(conf, &verifier_ca_chain, NULL);
 
-    ret = ra_tls_create_key_and_crt(&g_my_ratls_key, &g_my_ratls_cert);
+    ret = ra_tls_create_key_and_crt(&my_ratls_key, &my_ratls_cert);
     if (ret < 0) {
+        ERROR("Secret Provisioning failed during ra_tls_create_key_and_crt with error %d\n", ret);
+        ret = -EPERM;
         goto out;
     }
 
-    mbedtls_ssl_conf_rng(&g_conf, mbedtls_ctr_drbg_random, &g_ctr_drbg);
+    mbedtls_ssl_conf_rng(conf, mbedtls_ctr_drbg_random, &ctr_drbg);
 
-    ret = mbedtls_ssl_conf_own_cert(&g_conf, &g_my_ratls_cert, &g_my_ratls_key);
+    ret = mbedtls_ssl_conf_own_cert(conf, &my_ratls_cert, &my_ratls_key);
     if (ret < 0) {
+        ERROR("Secret Provisioning failed during mbedtls_ssl_conf_own_cert with error %d\n", ret);
+        ret = -EPERM;
         goto out;
     }
 
-    ret = mbedtls_ssl_setup(&g_ssl, &g_conf);
+    ret = mbedtls_ssl_setup(ssl, conf);
     if (ret < 0) {
+        ERROR("Secret Provisioning failed during mbedtls_ssl_setup with error %d\n", ret);
+        ret = -EPERM;
         goto out;
     }
 
-    ret = mbedtls_ssl_set_hostname(&g_ssl, connected_addr);
+    ret = mbedtls_ssl_set_hostname(ssl, connected_addr);
     if (ret < 0) {
+        ERROR("Secret Provisioning failed during mbedtls_ssl_set_hostname with error %d\n", ret);
+        ret = -EPERM;
         goto out;
     }
 
-    mbedtls_ssl_set_bio(&g_ssl, &g_verifier_fd, mbedtls_net_send, mbedtls_net_recv, NULL);
+    mbedtls_ssl_set_bio(ssl, net, mbedtls_net_send, mbedtls_net_recv, NULL);
 
-    ret = -1;
-    while (ret < 0) {
-        ret = mbedtls_ssl_handshake(&g_ssl);
-        if (ret == MBEDTLS_ERR_SSL_WANT_READ || ret == MBEDTLS_ERR_SSL_WANT_WRITE) {
-            continue;
-        }
-        if (ret < 0) {
-            goto out;
-        }
+    do {
+        ret = mbedtls_ssl_handshake(ssl);
+    } while (ret == MBEDTLS_ERR_SSL_WANT_READ || ret == MBEDTLS_ERR_SSL_WANT_WRITE);
+    if (ret < 0) {
+        ERROR("Secret Provisioning failed during mbedtls_ssl_handshake with error %d\n", ret);
+        ret = -EPERM;
+        goto out;
     }
 
-    uint32_t flags = mbedtls_ssl_get_verify_result(&g_ssl);
+    uint32_t flags = mbedtls_ssl_get_verify_result(ssl);
     if (flags != 0) {
-        ret = MBEDTLS_ERR_X509_CERT_VERIFY_FAILED;
+        ERROR("Secret Provisioning failed during mbedtls_ssl_get_verify_result (flags = %u)\n",
+              flags);
+        ret = -EPERM;
         goto out;
     }
 
-    struct ra_tls_ctx ctx = {.ssl = &g_ssl};
     uint8_t buf[128] = {0};
     size_t size;
 
@@ -217,75 +291,69 @@ int secret_provision_start(const char* in_servers, const char* in_ca_chain_path,
     size = sprintf((char*)buf, SECRET_PROVISION_REQUEST);
     size += 1; /* include null byte */
 
-    ret = secret_provision_write(&ctx, buf, size);
+    ret = secret_provision_common_write(ssl, buf, size);
     if (ret < 0) {
         goto out;
     }
 
     /* remote verifier sends 32-bit integer over network; we need to ntoh it */
-    uint32_t received_secret_size;
-    static_assert(sizeof(buf) >= sizeof(SECRET_PROVISION_RESPONSE) + sizeof(received_secret_size),
+    uint32_t secret_size;
+    static_assert(sizeof(buf) >= sizeof(SECRET_PROVISION_RESPONSE) + sizeof(secret_size),
                   "buffer must be sufficiently large to hold SECRET_PROVISION_RESPONSE + int32");
 
     memset(buf, 0, sizeof(buf));
-    ret = secret_provision_read(&ctx, buf,
-                                sizeof(SECRET_PROVISION_RESPONSE) + sizeof(received_secret_size));
+    ret = secret_provision_common_read(ssl, buf,
+                                       sizeof(SECRET_PROVISION_RESPONSE) + sizeof(secret_size));
     if (ret < 0) {
         goto out;
     }
 
     if (memcmp(buf, SECRET_PROVISION_RESPONSE, sizeof(SECRET_PROVISION_RESPONSE))) {
+        ERROR("Secret Provisioning read a response that doesn't match the expected "
+              SECRET_PROVISION_RESPONSE "\n");
         ret = -EINVAL;
         goto out;
     }
 
-    memcpy(&received_secret_size, buf + sizeof(SECRET_PROVISION_RESPONSE),
-           sizeof(received_secret_size));
+    memcpy(&secret_size, buf + sizeof(SECRET_PROVISION_RESPONSE), sizeof(secret_size));
 
-    received_secret_size = ntohl(received_secret_size);
-    if (received_secret_size > INT_MAX) {
+    secret_size = ntohl(secret_size);
+    if (secret_size > INT_MAX) {
         ret = -EINVAL;
         goto out;
     }
 
-    provisioned_secret_size = received_secret_size;
-    provisioned_secret = malloc(provisioned_secret_size);
-    if (!provisioned_secret) {
+    secret = malloc(secret_size);
+    if (!secret) {
         ret = -ENOMEM;
         goto out;
     }
 
-    ret = secret_provision_read(&ctx, provisioned_secret, provisioned_secret_size);
+    ret = secret_provision_common_read(ssl, secret, secret_size);
     if (ret < 0) {
         goto out;
     }
 
-    if (out_ctx) {
-        out_ctx->ssl = ctx.ssl;
-    } else {
-        secret_provision_close(&ctx);
-    }
-
+    ctx->secret      = secret;
+    ctx->secret_size = secret_size;
+    *out_ctx = ctx;
     ret = 0;
 out:
     if (ret < 0) {
-        secret_provision_destroy();
+        free(secret);
+        int close_ret = secret_provision_close(ctx);
+        if (close_ret < 0)
+            INFO("WARNING: Closing the secret-prov context failed with error %d.\n", close_ret);
     }
 
-    if (ret < 0 || !out_ctx) {
-        mbedtls_x509_crt_free(&g_my_ratls_cert);
-        mbedtls_pk_free(&g_my_ratls_key);
-        mbedtls_net_free(&g_verifier_fd);
-        mbedtls_ssl_free(&g_ssl);
-        mbedtls_ssl_config_free(&g_conf);
-        mbedtls_x509_crt_free(&g_verifier_ca_chain);
-        mbedtls_ctr_drbg_free(&g_ctr_drbg);
-        mbedtls_entropy_free(&g_entropy);
-    }
+    mbedtls_x509_crt_free(&my_ratls_cert);
+    mbedtls_pk_free(&my_ratls_key);
+    mbedtls_x509_crt_free(&verifier_ca_chain);
+    mbedtls_entropy_free(&entropy);
+    mbedtls_ctr_drbg_free(&ctr_drbg);
 
     free(servers);
     free(ca_chain_path);
-
     return ret;
 }
 
@@ -307,14 +375,15 @@ __attribute__((constructor)) static void secret_provision_constructor(void) {
 
         unsetenv(SECRET_PROVISION_SECRET_STRING);
 
+        struct ra_tls_ctx* ctx;
         int ret = secret_provision_start(/*in_servers=*/NULL, /*in_ca_chain_path=*/NULL,
-                                         /*out_ctx=*/NULL);
+                                         &ctx);
         if (ret < 0) {
             ERROR("Secret provisioning failed, terminating the whole process\n");
             exit(1);
         }
 
-        ret = secret_provision_get(&secret, &secret_size);
+        ret = secret_provision_get(ctx, &secret, &secret_size);
         if (ret < 0 || !secret || !secret_size || secret_size > PATH_MAX ||
                 secret[secret_size - 1] != '\0') {
             ERROR("Secret is not a null-terminated string, cannot do anything about such secret\n");
@@ -354,7 +423,8 @@ __attribute__((constructor)) static void secret_provision_constructor(void) {
 
             size_t total_written = 0;
             while (total_written < sizeof(keydata)) {
-                ssize_t written = write(fd, keydata + total_written, sizeof(keydata) - total_written);
+                ssize_t written = write(fd, keydata + total_written,
+                                        sizeof(keydata) - total_written);
                 if (written > 0) {
                     total_written += written;
                 } else if (written == 0) {
@@ -375,6 +445,6 @@ __attribute__((constructor)) static void secret_provision_constructor(void) {
         /* put the secret into an environment variable */
         setenv(SECRET_PROVISION_SECRET_STRING, (const char*)secret, /*overwrite=*/1);
 
-        secret_provision_destroy();
+        secret_provision_close(ctx);
     }
 }
--- a/CI-Examples/ra-tls-secret-prov/src/secret_prov_client.c
+++ b/CI-Examples/ra-tls-secret-prov/src/secret_prov_client.c
@@ -17,31 +17,20 @@
 
 int main(int argc, char** argv) {
     int ret;
-    int bytes;
 
-    struct ra_tls_ctx ctx = {0};
-
-    uint8_t* secret1   = NULL;
+    uint8_t* secret1 = NULL;
     size_t secret1_size = 0;
-
     uint8_t secret2[3] = {0}; /* we expect second secret to be 2-char string */
 
-    bool is_constructor = false;
-    char* str = getenv(SECRET_PROVISION_CONSTRUCTOR);
-    if (str && (!strcmp(str, "1") || !strcmp(str, "true") || !strcmp(str, "TRUE")))
-        is_constructor = true;
-
-    if (!is_constructor) {
-        /* secret provisioning was not run as part of initialization, run it now */
-        ret = secret_provision_start("dummyserver:80;localhost:4433;anotherdummy:4433",
-                                     CA_CRT_PATH, &ctx);
-        if (ret < 0) {
-            fprintf(stderr, "[error] secret_provision_start() returned %d\n", ret);
-            goto out;
-        }
+    struct ra_tls_ctx* ctx = NULL;
+    ret = secret_provision_start("dummyserver:80;localhost:4433;anotherdummy:4433",
+                                 CA_CRT_PATH, &ctx);
+    if (ret < 0) {
+        fprintf(stderr, "[error] secret_provision_start() returned %d\n", ret);
+        goto out;
     }
 
-    ret = secret_provision_get(&secret1, &secret1_size);
+    ret = secret_provision_get(ctx, &secret1, &secret1_size);
     if (ret < 0) {
         fprintf(stderr, "[error] secret_provision_get() returned %d\n", ret);
         goto out;
@@ -50,36 +39,27 @@ int main(int argc, char** argv) {
         fprintf(stderr, "[error] secret_provision_get() returned secret with size 0\n");
         goto out;
     }
-
     secret1[secret1_size - 1] = '\0';
 
-    if (!is_constructor) {
-        /* let's ask for another secret (just to show communication with secret-prov server) */
-        bytes = secret_provision_write(&ctx, (uint8_t*)SEND_STRING, sizeof(SEND_STRING));
-        if (bytes < 0) {
-            fprintf(stderr, "[error] secret_provision_write() returned %d\n", bytes);
-            goto out;
-        }
-
-        /* the secret we expect in return is a 2-char string */
-        bytes = secret_provision_read(&ctx, secret2, sizeof(secret2));
-        if (bytes < 0) {
-            fprintf(stderr, "[error] secret_provision_read() returned %d\n", bytes);
-            goto out;
-        }
-        if (bytes != sizeof(secret2)) {
-            fprintf(stderr, "[error] secret_provision_read() returned secret with size %d"
-                    " (expected %lu)\n", bytes, sizeof(secret2));
-            goto out;
-        }
-
-        secret2[bytes - 1] = '\0';
+    /* let's ask for another secret (just to show communication with secret-prov server) */
+    ret = secret_provision_write(ctx, (uint8_t*)SEND_STRING, sizeof(SEND_STRING));
+    if (ret < 0) {
+        fprintf(stderr, "[error] secret_provision_write() returned %d\n", ret);
+        goto out;
     }
+
+    /* the secret we expect in return is a 2-char string */
+    ret = secret_provision_read(ctx, secret2, sizeof(secret2));
+    if (ret < 0) {
+        fprintf(stderr, "[error] secret_provision_read() returned %d\n", ret);
+        goto out;
+    }
+    secret2[sizeof(secret2) - 1] = '\0';
 
     printf("--- Received secret1 = '%s', secret2 = '%s' ---\n", secret1, secret2);
     ret = 0;
 out:
-    secret_provision_destroy();
-    secret_provision_close(&ctx);
-    return ret;
+    free(secret1);
+    secret_provision_close(ctx);
+    return ret == 0 ? 0 : 1;
 }
--- a/CI-Examples/ra-tls-secret-prov/src/secret_prov_server.c
+++ b/CI-Examples/ra-tls-secret-prov/src/secret_prov_server.c
@@ -60,41 +60,32 @@ static int communicate_with_client_callback(struct ra_tls_ctx* ctx) {
     printf("--- Sent secret1 = '%s' ---\n", g_secret_string);
 
     /* let's send another secret (just to show communication with secret-awaiting client) */
-    int bytes;
-    uint8_t buf[128] = {0};
+    uint8_t buf[sizeof(EXPECTED_STRING)] = {0};
 
-    bytes = secret_provision_read(ctx, buf, sizeof(EXPECTED_STRING));
-    if (bytes < 0) {
-        if (bytes == -ECONNRESET) {
+    ret = secret_provision_read(ctx, buf, sizeof(buf));
+    if (ret < 0) {
+        if (ret == -ECONNRESET) {
             /* client doesn't want another secret, shutdown communication gracefully */
-            ret = 0;
-            goto out;
+            return 0;
         }
 
-        fprintf(stderr, "[error] secret_provision_read() returned %d\n", bytes);
-        ret = -EINVAL;
-        goto out;
+        fprintf(stderr, "[error] secret_provision_read() returned %d\n", ret);
+        return -EINVAL;
     }
 
-    assert(bytes == sizeof(EXPECTED_STRING));
-    if (memcmp(buf, EXPECTED_STRING, bytes)) {
+    if (memcmp(buf, EXPECTED_STRING, sizeof(EXPECTED_STRING))) {
         fprintf(stderr, "[error] client sent '%s' but expected '%s'\n", buf, EXPECTED_STRING);
-        ret = -EINVAL;
-        goto out;
+        return -EINVAL;
     }
 
-    bytes = secret_provision_write(ctx, (uint8_t*)SECRET_STRING, sizeof(SECRET_STRING));
-    if (bytes < 0) {
-        fprintf(stderr, "[error] secret_provision_write() returned %d\n", bytes);
-        ret = -EINVAL;
-        goto out;
+    ret = secret_provision_write(ctx, (uint8_t*)SECRET_STRING, sizeof(SECRET_STRING));
+    if (ret < 0) {
+        fprintf(stderr, "[error] secret_provision_write() returned %d\n", ret);
+        return -EINVAL;
     }
 
     printf("--- Sent secret2 = '%s' ---\n", SECRET_STRING);
-    ret = 0;
-out:
-    secret_provision_close(ctx);
-    return ret;
+    return 0;
 }
 
 int main(int argc, char** argv) {
--- a/tools/sgx/ra-tls/secret_prov.h
+++ b/tools/sgx/ra-tls/secret_prov.h
@@ -22,9 +22,7 @@
 
 #define DEFAULT_SERVERS "localhost:4433"
 
-struct ra_tls_ctx {
-    void* ssl;
-};
+struct ra_tls_ctx;
 
 typedef int (*verify_measurements_cb_t)(const char* mrenclave, const char* mrsigner,
                                         const char* isv_prod_id, const char* isv_svn);
@@ -43,6 +41,8 @@ typedef int (*secret_provision_cb_t)(struct ra_tls_ctx* ctx);
  *
  * This function can be called after an RA-TLS session is established via client-side call to
  * secret_provision_start() or in the server-side callback secret_provision_cb_t().
+ *
+ * This function always writes all \p size bytes on success (partial writes are not possible).
  */
 __attribute__ ((visibility("default")))
 int secret_provision_write(struct ra_tls_ctx* ctx, const uint8_t* buf, size_t size);
@@ -59,29 +59,36 @@ int secret_provision_write(struct ra_tls_ctx* ctx, const uint8_t* buf, size_t si
  *
  * This function can be called after an RA-TLS session is established via client-side call to
  * secret_provision_start() or in the server-side callback secret_provision_cb_t().
+ *
+ * This function always reads all \p size bytes on success (partial reads are not possible).
  */
 __attribute__ ((visibility("default")))
 int secret_provision_read(struct ra_tls_ctx* ctx, uint8_t* buf, size_t size);
 
 /*!
- * \brief Close an established RA-TLS session.
+ * \brief Close an established RA-TLS session and its associated secret.
  *
  * \param ctx  Established RA-TLS session.
  *
  * \returns 0 on success, specific error code (negative int) otherwise.
  *
  * This function can be called after an RA-TLS session is established via client-side call to
- * secret_provision_start() or in the server-side callback secret_provision_cb_t(). Typically,
- * application-specific protocol to provision secrets is implemented via secret_provision_read()
- * and secret_provision_write(), and this function is called to finish secret provisioning.
+ * secret_provision_start(). Typically, application-specific protocol to provision secrets is
+ * implemented via secret_provision_read() and secret_provision_write(), and this function is called
+ * to finish secret provisioning.
+ *
+ * This function zeroes out the memory where provisioned secret is stored and frees it.
+ *
+ * This function must not be called again even if it returned an error. \p ctx is always freed.
  */
 __attribute__ ((visibility("default")))
 int secret_provision_close(struct ra_tls_ctx* ctx);
 
 /*!
- * \brief Get a provisioned secret.
+ * \brief Get a copy of the provisioned secret.
  *
- * \param[out] out_secret       Pointer to buffer with secret (allocated by the library).
+ * \param      ctx              Established RA-TLS session.
+ * \param[out] out_secret       Pointer to newly allocated buffer with secret.
  * \param[out] out_secret_size  Size of allocated buffer.
  *
  * \returns 0 on success, specific error code (negative int) otherwise.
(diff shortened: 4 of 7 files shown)


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/813)* status: 0 of 5 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/813#-N8UVwWFBZGBJxIRv4ez:-N8UVwWFBZGBJxIRv4f-:b-xli39m) (no related file):*
@mkow @boryspoplawski You were the original reviewers of #647. Hopefully this new PR will be easier to review.

Sorry for creating a new PR, but the scope changed drastically in comparison to #647. Now we're changing the APIs.

___
___
*[`tools/sgx/ra-tls/secret_prov_verify.c` line 186 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8UWH20E0UQ8L0YJeoj:-N8UWH20E0UQ8L0YJeok:be1cajr) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_verify.c#L186)):*
> ```C
>                                 (const uint8_t*)pers, strlen(pers));
>     if (ret < 0) {
>         ret = -EPERM;
> ```

FYI: This is not optimal, but mbedTLS functions return mbedTLS-specific error codes, and there is no way to translate them to UNIX error codes (unless I write my own function that does it). Do we care?

liangintel:
after changing API, cross-language-framework solution can work well on this PR.
https://github.com/intel/confidential-computing-zoo/tree/main/cczoo/cross_lang_framework

dimakuv (commented):
@liangintel Thanks for checking!

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/813#-:-N8be2rK39chRZOaSSns:b-k8xliz)* status: 0 of 5 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/813)* status: all files reviewed, 21 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`CI-Examples/ra-tls-secret-prov/src/secret_prov_client.c` line 35 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8hqGvb3lSdZcMi_9h4:-N8hqGvb3lSdZcMi_9h5:brgxtab) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/CI-Examples/ra-tls-secret-prov/src/secret_prov_client.c#L35)):*
> ```C
> 
>     if (!is_constructor) {
>         /* secret provisioning was not run as part of initialization, run it now */
> ```

How this is supposed to work? `ctx` is a local variable, global constructor couldn't have possibly initialized it.
___
*[`CI-Examples/ra-tls-secret-prov/src/secret_prov_client.c` line 44 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8hyw_H6pCZPHFDR95j:-N8hyw_H6pCZPHFDR95k:boj406j) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/CI-Examples/ra-tls-secret-prov/src/secret_prov_client.c#L44)):*
> ```C
>     }
> 
>     ret = secret_provision_get(ctx, &secret1, &secret1_size);
> ```

This is now wrong in case of `is_constructor == true`
___
*[`tools/sgx/ra-tls/secret_prov.h` line 79 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8hsx1UE_Bb9dU64RxJ:-N8hsx1UE_Bb9dU64RxK:bnrjnzp) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov.h#L79)):*
> ```objc
>  */
> __attribute__ ((visibility("default")))
> int secret_provision_close(struct ra_tls_ctx* ctx);
> ```

Please also add a comment that this function must not be called again even if it returns an error and `ctx` is always freed (semantics similar to `close` syscall on Linux).
___
*[`tools/sgx/ra-tls/secret_prov.h` line 96 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8htOSA4AopqUSxRS4r:-N8htOSA4AopqUSxRS4s:b1iszsu) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov.h#L96)):*
> ```objc
>  */
> __attribute__ ((visibility("default")))
> int secret_provision_get(struct ra_tls_ctx* ctx, uint8_t** out_secret, size_t* out_secret_size);
> ```

This API is weird. This gets a reference to the underlying buffer, not a copy. Doing
```c
secret_provision_get(ctx, &buf, &size);
secret_provision_close(ctx);
use(buf);
```
is highly illegal.
___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 55 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8hrmWC2HyytPtyB7Nu:-N8hrmWC2HyytPtyB7Nv:b-5ohzel) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_attest.c#L55)):*
> ```C
> extern int secret_provision_common_write(mbedtls_ssl_context* ssl, const uint8_t* buf, size_t size);
> extern int secret_provision_common_read(mbedtls_ssl_context* ssl, uint8_t* buf, size_t size);
> extern int secret_provision_common_close(mbedtls_ssl_context* ssl);
> ```

Or maybe a header?



_Suggestion:_
```C
int secret_provision_common_write(mbedtls_ssl_context* ssl, const uint8_t* buf, size_t size);
int secret_provision_common_read(mbedtls_ssl_context* ssl, uint8_t* buf, size_t size);
int secret_provision_common_close(mbedtls_ssl_context* ssl);
```
___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 58 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8hs2xIDJQVJ_MNoE6a:-N8hs2xIDJQVJ_MNoE6b:b2oj4yp) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_attest.c#L58)):*
> ```C
> 
> int secret_provision_write(struct ra_tls_ctx* ctx, const uint8_t* buf, size_t size) {
>     mbedtls_ssl_context* ssl = (mbedtls_ssl_context*)ctx->ssl;
> ```

Why the cast? Why the immediate variable?
___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 63 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8hs4LU-R9PuqYVngiX:-N8hs4LU-R9PuqYVngiY:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_attest.c#L63)):*
> ```C
> 
> int secret_provision_read(struct ra_tls_ctx* ctx, uint8_t* buf, size_t size) {
>     mbedtls_ssl_context* ssl = (mbedtls_ssl_context*)ctx->ssl;
> ```

ditto
___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 136 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8hvhvv4iAm38fFBLWj:-N8hvhvv4iAm38fFBLWk:b-9lfj8b) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_attest.c#L136)):*
> ```C
>                                 (const uint8_t*)pers, strlen(pers));
>     if (ret < 0) {
>         ret = -EPERM;
> ```

Maybe print an error message with original mbedtls error code here?
ditto all other places
___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 252 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8hwa-bAK8rbjpA-0yH:-N8hwa-bAK8rbjpA-0yI:bdzqeqf) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_attest.c#L252)):*
> ```C
>             goto out;
>         }
>     }
> ```

It was already this way, but I couldn't look at it :(



_Suggestion:_
```C
    do {
        ret = mbedtls_ssl_handshake(ssl);
    } while (ret == MBEDTLS_ERR_SSL_WANT_READ || ret == MBEDTLS_ERR_SSL_WANT_WRITE);
    if (ret < 0) {
        ret = -EPERM;
        goto out;
    }
```
___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 309 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8huerECfggy0OWc8_5:-N8huerECfggy0OWc8_6:b-gzdsza) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_attest.c#L309)):*
> ```C
>     }
> 
>     ctx = malloc(sizeof(*ctx));
> ```

Why not create this at the top together with the rest of objects?
___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 324 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8hunBWAdE8yvEAgHSy:-N8hunBWAdE8yvEAgHSz:b-fk0t1l) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_attest.c#L324)):*
> ```C
>     ret = 0;
> out:
>     if (ret < 0) {
> ```

If you'd created `ctx` at the top you could just `secret_priviosion_close(ctx)` here, instead of duplicating this cleanup.
___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 325 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8huNXt-dvs39oA5-4a:-N8huNXt-dvs39oA5-4b:b-oa1ffk) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_attest.c#L325)):*
> ```C
> out:
>     if (ret < 0) {
>         secret_provision_common_close(ssl);
> ```

Lack of error checks.
___
*[`tools/sgx/ra-tls/secret_prov_common.c` line 19 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8hzIaF9jZOYuWaTVmW:-N8hzIaF9jZOYuWaTVmX:bia26w7) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_common.c#L19)):*
> ```C
> #include "secret_prov.h"
> 
> int secret_provision_common_write(mbedtls_ssl_context* ssl, const uint8_t* buf, size_t size);
> ```

Well, just create a header please
___
*[`tools/sgx/ra-tls/secret_prov_common.c` line 41 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8hzfio6KbhM9MSppp4:-N8hzfio6KbhM9MSppp5:b-vvlvg8) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_common.c#L41)):*
> ```C
>     }
>     assert(written == size);
>     return (int)written;
> ```

What's the point of this return value? It's always known to the caller. Also please use correct types.
___
*[`tools/sgx/ra-tls/secret_prov_common.c` line 65 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8hztfGBtnRLA4XT-sp:-N8hztfHB9q-uu8EEZA0:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_common.c#L65)):*
> ```C
> 
>     assert(read == size);
>     return (int)read;
> ```

ditto
___
*[`tools/sgx/ra-tls/secret_prov_common.c` line 83 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8i-Df-BhGZ3D5DqJ8f:-N8i-Df-BhGZ3D5DqJ8g:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_common.c#L83)):*
> ```C
>         }
>     }
>     return 0;
> ```





_Suggestion:_
```C
    int ret;
    do {
        ret = mbedtls_ssl_close_notify(ssl);
    } while (ret == MBEDTLS_ERR_SSL_WANT_READ || ret == MBEDTLS_ERR_SSL_WANT_WRITE);
    if (ret < 0) {
        /* use well-known error code for a typical case when remote party closes connection */
        return ret == MBEDTLS_ERR_SSL_PEER_CLOSE_NOTIFY ? -ECONNRESET : -EPERM;
    }
    return 0;
```
___
*[`tools/sgx/ra-tls/secret_prov_verify.c` line 50 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8hzBjn0T7wQlzAHQ5n:-N8hzBjn0T7wQlzAHQ5o:b-njw8vv) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_verify.c#L50)):*
> ```C
> static pthread_mutex_t g_handshake_lock;
> 
> extern int secret_provision_common_write(mbedtls_ssl_context* ssl, const uint8_t* buf, size_t size);
> ```

ditto (remove `extern` or create a header)
___
*[`tools/sgx/ra-tls/secret_prov_verify.c` line 98 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8i-hcFD7_Umnuinyf0:-N8i-hcFD7_Umnuinyf1:b-f7sgae) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_verify.c#L98)):*
> ```C
>     uint32_t flags = mbedtls_ssl_get_verify_result(&ssl);
>     if (flags != 0) {
>         goto out;
> ```

`ret` uninitialized?
___
*[`tools/sgx/ra-tls/secret_prov_verify.c` line 145 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8i1--372JtNBf2s6ev:-N8i1--372JtNBf2s6ew:b-dhjt5t) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_verify.c#L145)):*
> ```C
>     } else {
>         secret_provision_common_close(&ssl);
>     }
> ```

It's not really `caller`, we pass it to the callback. Caller cannot take the ownership of this objects, because it's local to this scope.
Why not do this and relieve the callback from the need of calling close?



_Suggestion:_
```C
    if (ti->f_cb) {
        struct ra_tls_ctx ctx = { .ssl = &ssl };
        ti->f_cb(&ctx);
    }
    secret_provision_common_close(&ssl);
```
___
*[`tools/sgx/ra-tls/secret_prov_verify.c` line 186 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8UWH20E0UQ8L0YJeoj:-N8i0R-FAVZJhhLUXCuy:bx48h4t) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_verify.c#L186)):*

FYI: This is not optimal, but mbedTLS functions return mbedTLS-specific error codes, and there is no way to translate them to UNIX error codes (unless I write my own function that does it). Do we care?
</blockquote></details>

Myabe just log the original error code?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/813)* status: all files reviewed, 21 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski and @dimakuv)

___
*[`CI-Examples/ra-tls-secret-prov/src/secret_prov_client.c` line 35 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8hqGvb3lSdZcMi_9h4:-N9f8tWmB4ZS2MKLL_tU:btp6de8) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/CI-Examples/ra-tls-secret-prov/src/secret_prov_client.c#L35)):*

How this is supposed to work? `ctx` is a local variable, global constructor couldn't have possibly initialized it.
</blockquote></details>

Very good question. Let me take a look at this example code. It probably only runs one of these flows (the no-constructor flow) ever, so I'll remove the other flow.
___
*[`tools/sgx/ra-tls/secret_prov.h` line 96 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8htOSA4AopqUSxRS4r:-N9f7BET-ZoR8SCwNc_a:b3ao9ko) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov.h#L96)):*

This API is weird. This gets a reference to the underlying buffer, not a copy. Doing
```c
secret_provision_get(ctx, &buf, &size);
secret_provision_close(ctx);
use(buf);
```
is highly illegal.
</blockquote></details>

Just to clarify before I change this API: you suggest to always return a *copy* of the secret here, right?
___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 252 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8hwa-bAK8rbjpA-0yH:-N9f7ZXC-QZAA9kZKP1Z:b69xxpk) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_attest.c#L252)):*

It was already this way, but I couldn't look at it :(
</blockquote></details>

There is a weird `null` line in your code change, I'll just ignore it. Just a typo?
___
*[`tools/sgx/ra-tls/secret_prov_common.c` line 19 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8hzIaF9jZOYuWaTVmW:-N9f7nr6-i4882SoUZCw:b-3cuc7z) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_common.c#L19)):*

Well, just create a header please
</blockquote></details>

A new internal header? Like `secret_prov_internal.h`?
___
*[`tools/sgx/ra-tls/secret_prov_verify.c` line 145 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8i1--372JtNBf2s6ev:-N9f8I8k6JU_7GvgpKS-:bji42y) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_verify.c#L145)):*

It's not really `caller`, we pass it to the callback. Caller cannot take the ownership of this objects, because it's local to this scope.
Why not do this and relieve the callback from the need of calling close?
</blockquote></details>

Yeah, agreed. I wanted to have a symmetry in the code of the secret-prov-verify user: first `secret_provision_start_server(&client_connection_callback)` and then `secret_provision_close()`.

But I agree that it's only confusing and cumbersome for users. Will change.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/813)* status: all files reviewed, 21 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski and @dimakuv)

___
*[`tools/sgx/ra-tls/secret_prov.h` line 96 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8htOSA4AopqUSxRS4r:-N9fz1Y07HmHP8JMqtR-:bbkrp2d) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov.h#L96)):*

Just to clarify before I change this API: you suggest to always return a *copy* of the secret here, right?
</blockquote></details>

I guess? I don't know what's the best solution, but even the comment suggests that it's a copy `Pointer to buffer with secret (allocated by the library).`
___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 252 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8hwa-bAK8rbjpA-0yH:-N9fzOVS1sWC7op_eCrh:b-2s85m9) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_attest.c#L252)):*

There is a weird `null` line in your code change, I'll just ignore it. Just a typo?
</blockquote></details>

Turn off the diff (i.e. look at the resulting code), looks like a reviewable bug
___
*[`tools/sgx/ra-tls/secret_prov_common.c` line 19 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8hzIaF9jZOYuWaTVmW:-N9fzbR9ANxz6FOkJSJ-:b-qazz8z) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_common.c#L19)):*

A new internal header? Like `secret_prov_internal.h`?
</blockquote></details>

Or `secret_prov_common.h`? I don't mind the name, but redefining this everywhere is annoying

tools/sgx/ra-tls/secret_prov_attest.c
```diff
@@ -33,43 +33,62 @@
...
+    mbedtls_ssl_context* ssl = (mbedtls_ssl_context*)ctx->ssl;
+    return secret_provision_common_write(ssl, buf, size);
+}
+
+int secret_provision_read(struct ra_tls_ctx* ctx, uint8_t* buf, size_t size) {
+    mbedtls_ssl_context* ssl = (mbedtls_ssl_context*)ctx->ssl;
+    return secret_provision_common_read(ssl, buf, size);
+}
+
+int secret_provision_close(struct ra_tls_ctx* ctx) {
```
liangintel, line 67: crash happened. please add if(!ctx) return ...
similar in secret_provision_get

dimakuv (commented):
> Reviewed 1 of 5 files at r1, 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/813)* status: all files reviewed, 24 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @liangintel, and @mkow)

___
*[`-- commits` line 11 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N9jt1CMBdpv0Q9xLMUy):*
I need to add here that `secret_provision_get()` returns a copy now.
___
*[`-- commits` line 20 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N9jt5J_1BcfmkAaN4pM):*
...But also prints the original mbedTLS error.
___
*[`CI-Examples/ra-tls-secret-prov/src/secret_prov_client.c` line 35 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8hqGvb3lSdZcMi_9h4:-N9jhahdEp7gUXcP5skl:bv6zqno) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/CI-Examples/ra-tls-secret-prov/src/secret_prov_client.c#L35)):*

Very good question. Let me take a look at this example code. It probably only runs one of these flows (the no-constructor flow) ever, so I'll remove the other flow.
</blockquote></details>

Done. Turns out the `is_constructor == true` flow never executed (because there is no `SECRET_PROVISION_CONSTRUCTOR` in the manifest). So I just removed it.
___
*[`CI-Examples/ra-tls-secret-prov/src/secret_prov_client.c` line 44 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8hyw_H6pCZPHFDR95j:-N9jhjePC8zF1hIX9CUi:b-kynmpj) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/CI-Examples/ra-tls-secret-prov/src/secret_prov_client.c#L44)):*

This is now wrong in case of `is_constructor == true`
</blockquote></details>

Done. Not applicable.
___
*[`tools/sgx/ra-tls/secret_prov.h` line 79 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8hsx1UE_Bb9dU64RxJ:-N9jiJn01mTBz7cgvdah:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov.h#L79)):*

Please also add a comment that this function must not be called again even if it returns an error and `ctx` is always freed (semantics similar to `close` syscall on Linux).
</blockquote></details>

Done.
___
*[`tools/sgx/ra-tls/secret_prov.h` line 96 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8htOSA4AopqUSxRS4r:-N9jkUcBCbGaE5oaPmWQ:b6ac6ee) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov.h#L96)):*

I guess? I don't know what's the best solution, but even the comment suggests that it's a copy `Pointer to buffer with secret (allocated by the library).`
</blockquote></details>

Done. Now we allocate a copy.
___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 55 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8hrmWC2HyytPtyB7Nu:-N9jlcrS9yxyji7ChcWU:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_attest.c#L55)):*

Or maybe a header?
</blockquote></details>

Done.
___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 58 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8hs2xIDJQVJ_MNoE6a:-N9jlqz33vMDekEl2azp:b-o3jhuf) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_attest.c#L58)):*

Why the cast? Why the immediate variable?
</blockquote></details>

Done. Remnant from the old version of the code.
___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 63 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8hs4LU-R9PuqYVngiX:-N9jlszi1MQzcpHh5q8i:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_attest.c#L63)):*

ditto
</blockquote></details>

Done.
___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 67 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8UVSjO5eysjSOv8Oky-r1-67:-N9juUAK3LAcq3C54Ykm:bf69lv) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_attest.c#L67)):*
<details><summary><i>Previously, liangintel (Liang Fang) wrote…</i></summary><blockquote>

crash happened. please add if(!ctx) return ...
similar in secret_provision_get
</blockquote></details>

Done. Added this sanity check to all user-facing (API) functions.
___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 136 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8hvhvv4iAm38fFBLWj:-N9jyVaH9Ot6lEXIWsBN:b-2o8a80) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_attest.c#L136)):*

Maybe print an error message with original mbedtls error code here?
ditto all other places
</blockquote></details>

Done. @mkow will hopefully be happy, he complained about the lack of any printouts recently. Now on errors smth like this is printed:
```
secret_provision_start: Secret Provisioning could not connect to any of the servers specified in SECRET_PROVISION_SERVERS
secret_provision_common_close: Secret Provisioning failed during connection close with mbedTLS error -28928
WARNING: Closing the secret-prov context failed with error -1.
secret_provision_constructor: Secret provisioning failed, terminating the whole process
```
___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 252 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8hwa-bAK8rbjpA-0yH:-N9jmGvu38hcpMx_piFj:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_attest.c#L252)):*

Turn off the diff (i.e. look at the resulting code), looks like a reviewable bug
</blockquote></details>

Done.
___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 309 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8huerECfggy0OWc8_5:-N9jsuWO1yorOUIbs5Sc:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_attest.c#L309)):*

Why not create this at the top together with the rest of objects?
</blockquote></details>

Done.
___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 324 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8hunBWAdE8yvEAgHSy:-N9josoM5d7boZM7LBMM:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_attest.c#L324)):*

If you'd created `ctx` at the top you could just `secret_priviosion_close(ctx)` here, instead of duplicating this cleanup.
</blockquote></details>

Done.
___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 325 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8huNXt-dvs39oA5-4a:-N9josSy179ZRnRRKMxC:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_attest.c#L325)):*

Lack of error checks.
</blockquote></details>

Done.
___
*[`tools/sgx/ra-tls/secret_prov_common.c` line 19 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8hzIaF9jZOYuWaTVmW:-N9jlRkh0or_Orsp_nG-:bb6un1j) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_common.c#L19)):*

Or `secret_prov_common.h`? I don't mind the name, but redefining this everywhere is annoying
</blockquote></details>

Done
___
*[`tools/sgx/ra-tls/secret_prov_common.c` line 41 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8hzfio6KbhM9MSppp4:-N9jqIs02-72nxNcGUfT:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_common.c#L41)):*

What's the point of this return value? It's always known to the caller. Also please use correct types.
</blockquote></details>

Done.
___
*[`tools/sgx/ra-tls/secret_prov_common.c` line 65 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8hztfGBtnRLA4XT-sp:-N9jqJj274Ai3bxJBwiq:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_common.c#L65)):*

ditto
</blockquote></details>

Done.
___
*[`tools/sgx/ra-tls/secret_prov_common.c` line 83 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8i-Df-BhGZ3D5DqJ8f:-N9jqbWNAg2fCeIsoj0l:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_common.c#L83)):*
> ```C
>         }
>     }
>     return 0;
> ```

Done.
___
*[`tools/sgx/ra-tls/secret_prov_common.c` line 10 at r2](https://reviewable.io/reviews/gramineproject/gramine/813#-N9k0I_d-ZgqNkqgiYEw:-N9k0I_d-ZgqNkqgiYEx:bszf1ae) ([raw file](https://github.com/gramineproject/gramine/blob/01542442d573bee0764f1f21c6b3a2a29fb05602/tools/sgx/ra-tls/secret_prov_common.c#L10)):*
> ```C
>  */
> 
> #define _XOPEN_SOURCE 700
> ```

FYI: This is added so that the compiler/linker find `dprintf` which is used in our `INFO`, `ERROR` macros.
___
*[`tools/sgx/ra-tls/secret_prov_verify.c` line 50 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8hzBjn0T7wQlzAHQ5n:-N9jlUmJBLLWwUC0CJSu:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_verify.c#L50)):*

ditto (remove `extern` or create a header)
</blockquote></details>

Done.
___
*[`tools/sgx/ra-tls/secret_prov_verify.c` line 98 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8i-hcFD7_Umnuinyf0:-N9jrEhuEDmML8HcqU8b:bqzriv6) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_verify.c#L98)):*

`ret` uninitialized?
</blockquote></details>

Not relevant. This function returns `NULL` anyway, and the label `out` doesn't care about the value of `ret`.
___
*[`tools/sgx/ra-tls/secret_prov_verify.c` line 145 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8i1--372JtNBf2s6ev:-N9jsq4H0yb0H5FJvT9e:bb6un1j) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_verify.c#L145)):*

Yeah, agreed. I wanted to have a symmetry in the code of the secret-prov-verify user: first `secret_provision_start_server(&client_connection_callback)` and then `secret_provision_close()`.

But I agree that it's only confusing and cumbersome for users. Will change.
</blockquote></details>

Done
___
*[`tools/sgx/ra-tls/secret_prov_verify.c` line 186 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8UWH20E0UQ8L0YJeoj:-N9jybu0A4T7Tbch8IwH:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_verify.c#L186)):*

Myabe just log the original error code?
</blockquote></details>

Done.
___
*[`tools/sgx/ra-tls/secret_prov_verify.c` line 16 at r2](https://reviewable.io/reviews/gramineproject/gramine/813#-N9k09va2dvz9K0qlklV:-N9k09va2dvz9K0qlklW:bszf1ae) ([raw file](https://github.com/gramineproject/gramine/blob/01542442d573bee0764f1f21c6b3a2a29fb05602/tools/sgx/ra-tls/secret_prov_verify.c#L16)):*
> ```C
>  */
> 
> #define _XOPEN_SOURCE 700
> ```

FYI: This is added so that the compiler/linker find `dprintf` which is used in our `INFO`, `ERROR` macros.

mkow (changes requested):
> Reviewed 1 of 5 files at r1, 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/813)* status: all files reviewed, 29 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @liangintel)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/813#-N9hNMiRC3MZ5j9EqPkV:-N9hNMiRC3MZ5j9EqPkW:be6rtsq) (no related file):*
@liangintel: Please use Reviewable, that's the tool we use as a project for reviews.

___
___
*[`CI-Examples/ra-tls-secret-prov/src/secret_prov_client.c` line 22 at r2](https://reviewable.io/reviews/gramineproject/gramine/813#-N9k99yD5GoJuMhWweVB:-N9k99yE92F8XRiRne6J:b-6epr5h) ([raw file](https://github.com/gramineproject/gramine/blob/01542442d573bee0764f1f21c6b3a2a29fb05602/CI-Examples/ra-tls-secret-prov/src/secret_prov_client.c#L22)):*
> ```C
> 
>     uint8_t* secret1    = NULL;
>     size_t secret1_size = 0;
> ```

Please pick one of these: (same with `secret2`)



_Suggestion (i):_
```C
    uint8_t* secret1 = NULL;
    size_t secret1_size = 0;
```



_Suggestion (ii):_
```C
    uint8_t* secret1      = NULL;
    size_t   secret1_size = 0;
```
___
*[`CI-Examples/ra-tls-secret-prov/src/secret_prov_client.c` line 63 at r2](https://reviewable.io/reviews/gramineproject/gramine/813#-N9kEDXECfUc_NEb_h_r:-N9kEDXECfUc_NEb_h_s:bz0pquv) ([raw file](https://github.com/gramineproject/gramine/blob/01542442d573bee0764f1f21c6b3a2a29fb05602/CI-Examples/ra-tls-secret-prov/src/secret_prov_client.c#L63)):*
> ```C
> out:
>     if (secret1) {
>         memset(secret1, 0, secret1_size);
> ```

This is a bad thing to showcase to users, compiler will delete this `memset()` anyways. Better not to showcase fake security, just delete it :)
___
*[`tools/sgx/ra-tls/secret_prov.h` line 82 at r2](https://reviewable.io/reviews/gramineproject/gramine/813#-N9kipVN5S6Iah7Zosf1:-N9kipVOCFr0r7y0Miv2:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/01542442d573bee0764f1f21c6b3a2a29fb05602/tools/sgx/ra-tls/secret_prov.h#L82)):*
> ```objc
>  * This function zeroes out the memory where provisioned secret is stored and frees it.
>  *
>  * This function must not be called again even if it returns an error. \p ctx is always freed.
> ```





_Suggestion:_
```objc
if it returned an error
```
___
*[`tools/sgx/ra-tls/secret_prov.h` line 99 at r2](https://reviewable.io/reviews/gramineproject/gramine/813#-N9kjdWX8FG1bLopgFqo:-N9kjdWYC-D6AUp8SH_v:b86a79t) ([raw file](https://github.com/gramineproject/gramine/blob/01542442d573bee0764f1f21c6b3a2a29fb05602/tools/sgx/ra-tls/secret_prov.h#L99)):*
> ```objc
>  * provisioning via secret_provision_start() which will obtain the secret from the server and
>  * save it in enclave memory. After that, the client can call this function to retrieve the
>  * copy of the secret from memory. The client must erase the memory of the retrieved copy of the
> ```

I wouldn't say "must", it's not really required nor making much sense.



_Code quote:_
```objc
The client must erase the memory of the retrieved copy
```
___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 136 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8hvhvv4iAm38fFBLWj:-N9kmrB44S2e0FhPiNoU:b48fpd6) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_attest.c#L136)):*

Done. @mkow will hopefully be happy, he complained about the lack of any printouts recently. Now on errors smth like this is printed:
```
secret_provision_start: Secret Provisioning could not connect to any of the servers specified in SECRET_PROVISION_SERVERS
secret_provision_common_close: Secret Provisioning failed during connection close with mbedTLS error -28928
WARNING: Closing the secret-prov context failed with error -1.
secret_provision_constructor: Secret provisioning failed, terminating the whole process
```
</blockquote></details>

Yeah, better now :) Although in the future it would be better to print mbedtls errors decoded to strings, not just numbers (same for Linux and PAL errnos).
___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 308 at r2](https://reviewable.io/reviews/gramineproject/gramine/813#-N9kpwwICRPOc-HXvM1S:-N9kpwwICRPOc-HXvM1T:b-sll0v) ([raw file](https://github.com/gramineproject/gramine/blob/01542442d573bee0764f1f21c6b3a2a29fb05602/tools/sgx/ra-tls/secret_prov_attest.c#L308)):*
> ```C
>     memset(buf, 0, sizeof(buf));
>     ret = secret_provision_common_read(ssl, buf, sizeof(SECRET_PROVISION_RESPONSE) +
>                                                   sizeof(secret_size));
> ```

(or 4 spaces, just not _one_ :P)



_Suggestion:_
```C
    ret = secret_provision_common_read(ssl, buf, sizeof(SECRET_PROVISION_RESPONSE) +
                                                 sizeof(secret_size));
```

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/813)* status: 4 of 7 files reviewed, 28 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, @liangintel, and @mkow)

___
*[`CI-Examples/ra-tls-secret-prov/src/secret_prov_client.c` line 22 at r2](https://reviewable.io/reviews/gramineproject/gramine/813#-N9k99yD5GoJuMhWweVB:-N9kxRy-3t0ykeBf-7ny:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/01542442d573bee0764f1f21c6b3a2a29fb05602/CI-Examples/ra-tls-secret-prov/src/secret_prov_client.c#L22)):*

Please pick one of these: (same with `secret2`)
</blockquote></details>

Done.
___
*[`CI-Examples/ra-tls-secret-prov/src/secret_prov_client.c` line 63 at r2](https://reviewable.io/reviews/gramineproject/gramine/813#-N9kEDXECfUc_NEb_h_r:-N9kxWuf4XKbubHu71xi:b7a30tx) ([raw file](https://github.com/gramineproject/gramine/blob/01542442d573bee0764f1f21c6b3a2a29fb05602/CI-Examples/ra-tls-secret-prov/src/secret_prov_client.c#L63)):*

This is a bad thing to showcase to users, compiler will delete this `memset()` anyways. Better not to showcase fake security, just delete it :)
</blockquote></details>

Done. Yeah, I was wondering if I should copy-paste this: https://github.com/gramineproject/gramine/blob/a014bc685bbb3cb8b37f90b482e6ab5de620be56/tools/sgx/ra-tls/secret_prov_attest.c#L61-L65

But I guess let's just leave the "proper memory erasing" to users.
___
*[`tools/sgx/ra-tls/secret_prov.h` line 82 at r2](https://reviewable.io/reviews/gramineproject/gramine/813#-N9kipVN5S6Iah7Zosf1:-N9kxnoVDJUE4JfOxzZC:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/01542442d573bee0764f1f21c6b3a2a29fb05602/tools/sgx/ra-tls/secret_prov.h#L82)):*
> ```objc
>  * This function zeroes out the memory where provisioned secret is stored and frees it.
>  *
>  * This function must not be called again even if it returns an error. \p ctx is always freed.
> ```

Done.
___
*[`tools/sgx/ra-tls/secret_prov.h` line 99 at r2](https://reviewable.io/reviews/gramineproject/gramine/813#-N9kjdWX8FG1bLopgFqo:-N9kxxd05GRR0xsvnLMG:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/01542442d573bee0764f1f21c6b3a2a29fb05602/tools/sgx/ra-tls/secret_prov.h#L99)):*

I wouldn't say "must", it's not really required nor making much sense.
</blockquote></details>

Done.
___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 136 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N8hvhvv4iAm38fFBLWj:-N9kylE11NeW5DmE_ljb:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/71fa466c2ac054387aa55c762643379bb6312544/tools/sgx/ra-tls/secret_prov_attest.c#L136)):*

Yeah, better now :) Although in the future it would be better to print mbedtls errors decoded to strings, not just numbers (same for Linux and PAL errnos).
</blockquote></details>

Done.
___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 308 at r2](https://reviewable.io/reviews/gramineproject/gramine/813#-N9kpwwICRPOc-HXvM1S:-N9kytPtCzm_M16DKM9K:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/01542442d573bee0764f1f21c6b3a2a29fb05602/tools/sgx/ra-tls/secret_prov_attest.c#L308)):*

(or 4 spaces, just not _one_ :P)
</blockquote></details>

Done.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/813)* status: all files reviewed, 14 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @liangintel)

___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 308 at r2](https://reviewable.io/reviews/gramineproject/gramine/813#-N9kpwwICRPOc-HXvM1S:-N9lEemmEFJjYi1xtWIA:b-80bnvp) ([raw file](https://github.com/gramineproject/gramine/blob/01542442d573bee0764f1f21c6b3a2a29fb05602/tools/sgx/ra-tls/secret_prov_attest.c#L308)):*

Also we place operators in the new line in such situations, afaik
</blockquote></details>

True.
___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 60 at r3](https://reviewable.io/reviews/gramineproject/gramine/813#-N9l9VOX2D2wWo0AwOFx:-N9lEKre0fOUkQHqiBNQ:b-6l2n63) ([raw file](https://github.com/gramineproject/gramine/blob/40f7ed0a03f4571895c84423f7770a2a190fec90/tools/sgx/ra-tls/secret_prov_attest.c#L60)):*

This doesn't make much sense to me...
Why you only check `ctx == NULL` and not the rest of the arguments?
</blockquote></details>

I don't like these checks, why would `ctx == NULL` be allowed if it's not explicitly documented as such?
But it's a public API, so I won't block on this.

dimakuv (commented):
> Reviewed 2 of 5 files at r4.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/813)* status: 4 of 7 files reviewed, 14 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, @liangintel, and @mkow)

___
*[`CI-Examples/ra-tls-secret-prov/src/secret_prov_client.c` line 64 at r3](https://reviewable.io/reviews/gramineproject/gramine/813#-N9l7DXq3g-wgr3XlPxW:-N9lHMv64xTwgLKMlQ4y:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/40f7ed0a03f4571895c84423f7770a2a190fec90/CI-Examples/ra-tls-secret-prov/src/secret_prov_client.c#L64)):*

Please do not return negative errno codes from main.
</blockquote></details>

Done.
___
*[`CI-Examples/ra-tls-secret-prov/src/secret_prov_server.c` line 55 at r3](https://reviewable.io/reviews/gramineproject/gramine/813#-N9l7zuPC8dSz_O3qzeN:-N9lHeyS65i_7krzhR8q:b-rmkr0c) ([raw file](https://github.com/gramineproject/gramine/blob/40f7ed0a03f4571895c84423f7770a2a190fec90/CI-Examples/ra-tls-secret-prov/src/secret_prov_server.c#L55)):*

Why would there be a reason to call it? I don't understand this comment.
</blockquote></details>

Done. Felt that users will see `_read` and `_write` function calls, and will automatically assume there needs to be a finalizing `_close` call. But maybe I'm just thinking nonsense.
___
*[`CI-Examples/ra-tls-secret-prov/src/secret_prov_server.c` line 63 at r3](https://reviewable.io/reviews/gramineproject/gramine/813#-N9l8KTw56Ee4hZRyCtW:-N9lI-3IBH40r0-lPAMd:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/40f7ed0a03f4571895c84423f7770a2a190fec90/CI-Examples/ra-tls-secret-prov/src/secret_prov_server.c#L63)):*
> ```C
> 
>     /* let's send another secret (just to show communication with secret-awaiting client) */
>     uint8_t buf[128] = {0};
> ```

Done.
___
*[`CI-Examples/ra-tls-secret-prov/src/secret_prov_server.c` line 65 at r3](https://reviewable.io/reviews/gramineproject/gramine/813#-N9l8Pg440lBwtCLTcUk:-N9lI-PY4Gae61ZdeOE6:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/40f7ed0a03f4571895c84423f7770a2a190fec90/CI-Examples/ra-tls-secret-prov/src/secret_prov_server.c#L65)):*
> ```C
>     uint8_t buf[128] = {0};
> 
>     ret = secret_provision_read(ctx, buf, sizeof(EXPECTED_STRING));
> ```

Done.
___
*[`tools/sgx/ra-tls/secret_prov.h` line 99 at r3](https://reviewable.io/reviews/gramineproject/gramine/813#-N9l8vXP7lrnYYqu3Dtb:-N9lIAeq-48-4qHlTQHH:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/40f7ed0a03f4571895c84423f7770a2a190fec90/tools/sgx/ra-tls/secret_prov.h#L99)):*
> ```objc
>  * provisioning via secret_provision_start() which will obtain the secret from the server and
>  * save it in enclave memory. After that, the client can call this function to retrieve the
>  * copy of the secret from memory. The client must free the copied-secret buffer afterwards.
> ```

Done.
___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 308 at r2](https://reviewable.io/reviews/gramineproject/gramine/813#-N9kpwwICRPOc-HXvM1S:-N9lKDY137mk5u4S00It:b-7qxs5k) ([raw file](https://github.com/gramineproject/gramine/blob/01542442d573bee0764f1f21c6b3a2a29fb05602/tools/sgx/ra-tls/secret_prov_attest.c#L308)):*

True.
</blockquote></details>

Done. I guess this version is the most correct now :)
___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 60 at r3](https://reviewable.io/reviews/gramineproject/gramine/813#-N9l9VOX2D2wWo0AwOFx:-N9lIMtj754Sh4puC9ZF:bc92pj3) ([raw file](https://github.com/gramineproject/gramine/blob/40f7ed0a03f4571895c84423f7770a2a190fec90/tools/sgx/ra-tls/secret_prov_attest.c#L60)):*

I don't like these checks, why would `ctx == NULL` be allowed if it's not explicitly documented as such?
But it's a public API, so I won't block on this.
</blockquote></details>

Done. I'm adding these checks because this is a public API, so we should have some protection against dummy usage.
___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 134 at r3](https://reviewable.io/reviews/gramineproject/gramine/813#-N9lBGdWBNqtatOn5d_5:-N9lIfu8-K2rTD1c9STR:bsw6ze8) ([raw file](https://github.com/gramineproject/gramine/blob/40f7ed0a03f4571895c84423f7770a2a190fec90/tools/sgx/ra-tls/secret_prov_attest.c#L134)):*

`secret_size` is not set and could be garbage.
</blockquote></details>

Done. This is not strictly nesserary -- I'm only setting `secret` to NULL so that `free(ctx->secret)` doesn't break. But added just for sanity.
___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 159 at r3](https://reviewable.io/reviews/gramineproject/gramine/813#-N9l9wxk0x0uHVGdKl7t:-N9lIuDA3AyPbDeZAbdx:bxj84z0) ([raw file](https://github.com/gramineproject/gramine/blob/40f7ed0a03f4571895c84423f7770a2a190fec90/tools/sgx/ra-tls/secret_prov_attest.c#L159)):*

Why this isn't `goto out`?
</blockquote></details>

Done. It was a bug.
___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 198 at r3](https://reviewable.io/reviews/gramineproject/gramine/813#-N9lANDk-79DPC6E89jN:-N9lK3BS3Z7rAmjvtSMj:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/40f7ed0a03f4571895c84423f7770a2a190fec90/tools/sgx/ra-tls/secret_prov_attest.c#L198)):*

This is bad, we mix mbedtls error codes with POSIX errno (`ECONNREFUXED` above)
</blockquote></details>

Done.
___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 204 at r3](https://reviewable.io/reviews/gramineproject/gramine/813#-N9lAByU0WUMPLH2IjGo:-N9lK47W3rArZ6ywzV2X:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/40f7ed0a03f4571895c84423f7770a2a190fec90/tools/sgx/ra-tls/secret_prov_attest.c#L204)):*

Why this special casing? It's also wrong, it could have same value as some mbedtls error
</blockquote></details>

Done.
___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 187 at r4](https://reviewable.io/reviews/gramineproject/gramine/813#-N9lLI1g8IAGxxgTEyeE:-N9lLI1g8IAGxxgTEyeF:b-q3o4ss) ([raw file](https://github.com/gramineproject/gramine/blob/5b6710e3808501d57cd79c57a0c43bfc3cac1593/tools/sgx/ra-tls/secret_prov_attest.c#L187)):*
> ```C
>     char* str1;
>     for (str1 = servers; /*no condition*/; str1 = NULL) {
>         ret = MBEDTLS_ERR_NET_CONNECT_FAILED;
> ```

FYI: https://os.mbed.com/teams/sandbox/code/mbedtls/docs/tip/net_8h_source.html (line 38)

boryspoplawski (changes requested):
> Reviewed 5 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/813)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @liangintel, and @mkow)

___
*[`CI-Examples/ra-tls-secret-prov/src/secret_prov_server.c` line 55 at r3](https://reviewable.io/reviews/gramineproject/gramine/813#-N9l7zuPC8dSz_O3qzeN:-N9lL_w90ZmbKpzH4dY1:b-9uvkr6) ([raw file](https://github.com/gramineproject/gramine/blob/40f7ed0a03f4571895c84423f7770a2a190fec90/CI-Examples/ra-tls-secret-prov/src/secret_prov_server.c#L55)):*

Done. Felt that users will see `_read` and `_write` function calls, and will automatically assume there needs to be a finalizing `_close` call. But maybe I'm just thinking nonsense.
</blockquote></details>

I would say that if there is no open/create/init why would there be close?
___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 60 at r4](https://reviewable.io/reviews/gramineproject/gramine/813#-N9lLmaa5LSCv87YeO1C:-N9lLmaa5LSCv87YeO1D:b-buh0fz) ([raw file](https://github.com/gramineproject/gramine/blob/5b6710e3808501d57cd79c57a0c43bfc3cac1593/tools/sgx/ra-tls/secret_prov_attest.c#L60)):*
> ```C
> 
> int secret_provision_write(struct ra_tls_ctx* ctx, const uint8_t* buf, size_t size) {
>     if (!ctx || !buf)
> ```

What if `size == 0`? Isn't `but == NULL` valid then?
___
*[`tools/sgx/ra-tls/secret_prov_verify.c` line 135 at r4](https://reviewable.io/reviews/gramineproject/gramine/813#-N9lMBFiAUSWtczSHNpA:-N9lMBFiAUSWtczSHNpB:b-pb4i4k) ([raw file](https://github.com/gramineproject/gramine/blob/5b6710e3808501d57cd79c57a0c43bfc3cac1593/tools/sgx/ra-tls/secret_prov_verify.c#L135)):*
> ```C
> 
>     ret = secret_provision_common_write(&ssl, buf, sizeof(SECRET_PROVISION_RESPONSE)
>                                                        + sizeof(send_secret_size));
> ```

Why the 4 spaces? Not really blocking, but we never used such convention before

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/813)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, @liangintel, and @mkow)

___
*[`CI-Examples/ra-tls-secret-prov/src/secret_prov_server.c` line 55 at r3](https://reviewable.io/reviews/gramineproject/gramine/813#-N9l7zuPC8dSz_O3qzeN:-N9lMcI00GMbDFlwnsyI:bgeo3n0) ([raw file](https://github.com/gramineproject/gramine/blob/40f7ed0a03f4571895c84423f7770a2a190fec90/CI-Examples/ra-tls-secret-prov/src/secret_prov_server.c#L55)):*

I would say that if there is no open/create/init why would there be close?
</blockquote></details>

Yeah, I have weird thoughts.
___
*[`tools/sgx/ra-tls/secret_prov_attest.c` line 60 at r4](https://reviewable.io/reviews/gramineproject/gramine/813#-N9lLmaa5LSCv87YeO1C:-N9lPzLJC3PR_C92XYP9:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/5b6710e3808501d57cd79c57a0c43bfc3cac1593/tools/sgx/ra-tls/secret_prov_attest.c#L60)):*

What if `size == 0`? Isn't `but == NULL` valid then?
</blockquote></details>

Done.
___
*[`tools/sgx/ra-tls/secret_prov_verify.c` line 135 at r4](https://reviewable.io/reviews/gramineproject/gramine/813#-N9lMBFiAUSWtczSHNpA:-N9lPyFN4HsYjQxU9nNW:bb6un1j) ([raw file](https://github.com/gramineproject/gramine/blob/5b6710e3808501d57cd79c57a0c43bfc3cac1593/tools/sgx/ra-tls/secret_prov_verify.c#L135)):*

Why the 4 spaces? Not really blocking, but we never used such convention before
</blockquote></details>

Done

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/813)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, @liangintel, and @mkow)

___
*[`CI-Examples/ra-tls-secret-prov/src/secret_prov_client.c` line 64 at r6](https://reviewable.io/reviews/gramineproject/gramine/813#-N9lTNiT1gs1FYedi6cp:-N9lTNiT1gs1FYedi6cq:b9y8rwt) ([raw file](https://github.com/gramineproject/gramine/blob/fe3bec45c344fbd236c5f764ecda204014101999/CI-Examples/ra-tls-secret-prov/src/secret_prov_client.c#L64)):*
> ```C
>     free(secret1);
>     secret_provision_close(ctx);
>     return ret == 0 ? 0 : 1;
> ```

FYI: This was my blooper.

mkow (dismissed):
> Reviewed 3 of 5 files at r4, 2 of 2 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/813)* status: all files reviewed, 4 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @liangintel)

___
*[`CI-Examples/ra-tls-secret-prov/src/secret_prov_client.c` line 64 at r3](https://reviewable.io/reviews/gramineproject/gramine/813#-N9l7DXq3g-wgr3XlPxW:-N9lSh3r1Q-Hp6jLURmt:b-17k1nv) ([raw file](https://github.com/gramineproject/gramine/blob/40f7ed0a03f4571895c84423f7770a2a190fec90/CI-Examples/ra-tls-secret-prov/src/secret_prov_client.c#L64)):*

Done.
</blockquote></details>

Aren't you always returning 1 now?

Update: seems you fixed this before I finished the review :)



_Code quote:_
```C
return ret == 0 ?: 1;
```

boryspoplawski (dismissed):
> Reviewed 2 of 2 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/813)* status: all files reviewed, 3 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @liangintel)

mkow (approved):
> Dismissed @liangintel from a discussion.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/813)* status: all files reviewed, 2 unresolved discussions (waiting on @dimakuv)

___
*[`-- commits` line 11 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N9jt1CMBdpv0Q9xLMUy):*

I need to add here that `secret_provision_get()` returns a copy now.
</blockquote></details>

Done.
___
*[`-- commits` line 20 at r1](https://reviewable.io/reviews/gramineproject/gramine/813#-N9jt5J_1BcfmkAaN4pM):*

...But also prints the original mbedTLS error.
</blockquote></details>

Done.

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/813)* status: all files reviewed, 2 unresolved discussions (waiting on @dimakuv)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/813)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #824, opened 2022-08-05
Encrypted files do not get closed


klassiker:
## Description of the problem

When a file is opened inside an encrypted mount and the file descriptor is not closed by the application no data is ever written to the host. Since the encrypted file header is also missing (file size is zero), the application is broken on restart.

## Steps to reproduce

Application:

```c
#include <unistd.h>
#include <fcntl.h>

int main() {
    int fd = openat(AT_FDCWD, "/secrets/myfile", O_WRONLY|O_CREAT, 0666);
    write(fd, "test", 4);
}
```

Manifest:

```toml
libos.entrypoint = "/app/main"

loader.entrypoint = "file:{{ gramine.libos }}"

loader.log_level = "all"
loader.log_file = "log.txt"

loader.env.LD_LIBRARY_PATH = "/lib:{{ arch_libdir }}"

fs.mounts = [
    { path = "/lib",              uri = "file:{{ gramine.runtimedir() }}" },
    { path = "{{ arch_libdir }}", uri = "file:{{ arch_libdir }}" },
    { path = "/app",              uri = "file:/app" },
    { path = "/secrets",          uri = "file:/data", type = "encrypted" },
]
fs.insecure__keys.default = "ffeeddccbbaa99887766554433221100"

sgx.thread_num = 8
sgx.enclave_size = "512M"

sgx.allowed_files = [
    "file:/",
]
```

## Expected results

File is closed by gramine when the process exits and can be opened again after a restart.

## Actual results

Trace, first start:

```
[P1:T1:main] trace: ---- openat(AT_FDCWD, "/secrets/myfile", O_WRONLY|O_CREAT, 0666) = 0x3
[P1:T1:main] trace: ---- write(3, 0x313ae017, 0x4) ...
[P1:T1:main] trace: ---- return from write(...) = 0x4
[P1:T1:main] debug: ---- exit_group (returning 0)
[P1:T1:main] debug: clearing POSIX locks for pid 1
[P1:T1:main] debug: sync client shutdown: closing handles
[P1:T1:main] debug: sync client shutdown: waiting for confirmation
[P1:T1:main] debug: sync client shutdown: finished
[P1:libos] debug: IPC worker: exiting worker thread
[P1:T1:main] debug: process 1 exited with status 0
```

Trace, second start:

```
[P1:T1:main] warning: cb_read: EOF
[P1:T1:main] warning: encrypted_file_internal_open: pf_open failed: Callback failed
[...]

PR #826, opened 2022-08-08
[common] Init new protected file when file size is zero

kailun-qin:
Currently, Gramine does not close a file opened inside an encrypted mount but not explicitly closed by the application. No data is ever written to the host in this case. A subsequent opening of this file (e.g., on the application restart) will be broken since it's considered as operating on an existing file however the file size is actually zero and the encrypted file header is also missing.

This patch initializes a new protected file instead of an existing one when the file size is found out to be zero.
Fixes https://github.com/gramineproject/gramine/issues/824.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/826)

### DIFF
--- a/common/src/protected_files/protected_files.c
+++ b/common/src/protected_files/protected_files.c
@@ -951,11 +951,10 @@ static pf_context_t* ipf_open(const char* path, pf_file_mode_t mode, bool create
     pf->real_file_size = real_size;
     pf->mode = mode;
 
-    if (!create) {
+    if (!create && real_size > 0) {
         // existing file
         if (!ipf_init_existing_file(pf, path))
             goto out;
-
     } else {
         // new file
         if (!ipf_init_new_file(pf, path))


oshogbo (commented):
LGTM.

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/826#-:-N8wpG7X0S3sxS4OwR-O:b-85i63v)* status: 0 of 1 files reviewed, all discussions resolved, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: )

klassiker:
Thanks for looking at this issue, but this only works in the case where the loss of data written to the file is acceptable.

My reduced application in the issue would work fine with this fix, but I actually encountered this problem with applications that write logfiles without closing them on shutdown, leading to empty logs on application crashes.

In my opinion this does not fix #824 and just hides the actual bug.

kailun-qin:
Thanks for the feedbacks!

> but this only works in the case where the loss of data written to the file is acceptable.
> In my opinion this does not fix https://github.com/gramineproject/gramine/issues/824 and just hides the actual bug.

This depends on what behavior you'd you expect for https://github.com/gramineproject/gramine/issues/824. As cited in the description, this PR only tries to fix the broken app. on a subsequent file open w/o a previous explicit closing - which is clearly not right (i.e., a bug).

However, this PR does not intend to support "no loss of data" - which IMO is more complicated, more close to (several) other enhancements and (in some cases) not even feasible:
1. Gramine protected files (based on a reference counting scheme) do not currently support proper handling of file closing, also do not support auto file closing on exit.
2. Gramine protected files do not handle file corruption when fatal errors occur during the changes are written to the disk (i.e., no corruption recovery mechanism is provided). This also leads to "loss of data". Then even supported, for small sets of data less than PF cache size, there is still no guarantee that there'll be "no loss of data".
3. Also, implementations to support "no loss of data" needs to be careful as this can sometimes be exploited by attackers thru advanced ways, i.e., we'd rather prefer the file to be in a initial status (and to only rely on explicit flush or close) in such cases.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/826)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/826#-N93mvOVCcK-RXOGHeNa:-N93mvOW6k3amtrt_clS:brkmpp0) (no related file):*
I agree with @klassiker, this PR seems to only mask the bug.

> this PR only tries to fix the broken app

The app in #824 is perfectly correct (ok, maybe except missing error checks, but that was just a repro code so whatever).

> However, this PR does not intend to support "no loss of data" - which IMO is more complicated, more close to (several) other enhancements and (in some cases) not even feasible: [...]

We can't easily support the case where it's Gramine code which crashed, but for the user code it's perfectly doable (assuming it didn't corrupt Gramine memory).
And in the case which @klassiker mentioned in #824 the app doesn't crash, but correctly exits, so I don't see how the points you listed here relate to it.

This seems to be a bug in PF, we should just close all the descriptors on exit.

___

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/826)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @klassiker)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/826#-N93mvOVCcK-RXOGHeNa:-N93v7GY4Ly_9Iv1WzuS:bvchx5u) (no related file):*

I agree with @klassiker, this PR seems to only mask the bug.

> this PR only tries to fix the broken app

The app in #824 is perfectly correct (ok, maybe except missing error checks, but that was just a repro code so whatever).

> However, this PR does not intend to support "no loss of data" - which IMO is more complicated, more close to (several) other enhancements and (in some cases) not even feasible: [...]

We can't easily support the case where it's Gramine code which crashed, but for the user code it's perfectly doable (assuming it didn't corrupt Gramine memory).
And in the case which @klassiker mentioned in #824 the app doesn't crash, but correctly exits, so I don't see how the points you listed here relate to it.

This seems to be a bug in PF, we should just close all the descriptors on exit.
</blockquote></details>

We just do not close nor flush any open fds at exit, which most likely we should. It matters for PF where we need an explicit flush at `close()` (which here never happens).

___

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/826)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @klassiker)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/826#-N93mvOVCcK-RXOGHeNa:-N94DhS5-A7TkJKaLJvD:bdmyk19) (no related file):*
> I agree with @klassiker, this PR seems to only mask the bug.

OK, I'll close this one if you think it's not right.

> The app in #824 is perfectly correct

Excuse me for the confusions, I was not intending to say that the app is incorrect, but it's blocked (where I should just use this word instead).

> but correctly exits, so I don't see how the points you listed here relate to it.

"also do not support auto file closing on exit" - isn't this related?

I was just commenting generally on "this only works in the case where the loss of data written to the file is acceptable", i.e., "the loss of data" could happen elsewhere (incl. not close fds at exit, 1st item). I consider the listed as enhancements/TODOs/follow-ups, while this PR is only a quick fix to unblock the app where I don't see adding this check will bring drawbacks even if we implemented "close all fds on exit".

___

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/826)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @klassiker)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/826#-N93mvOVCcK-RXOGHeNa:-N95mUIy63UlqaUZNBgr:bdxih9w) (no related file):*
> we should just close all the descriptors on exit.

Drafted a quick one at https://github.com/gramineproject/gramine/pull/830, pls kindly see if this is sth you'd expect. Thanks!

___

kailun-qin:
Switched to https://github.com/gramineproject/gramine/pull/830. Closing this one.

Issue #836, opened 2022-08-11
[LibOS] Posix file locks are not released on fds with `CLOEXEC` flag on `execve`


boryspoplawski:
## Description of the problem
Posix file locks are not released on fds with `CLOEXEC` flag on `execve`.
Regular file closing: https://github.com/gramineproject/gramine/blob/90f8b9d7eacfa1c9bc07089d9d00ea774888435f/libos/src/bookkeep/libos_handle.c#L309
Closing on execve: https://github.com/gramineproject/gramine/blob/90f8b9d7eacfa1c9bc07089d9d00ea774888435f/libos/src/sys/libos_exec.c#L18

PR #837, opened 2022-08-11
[LibOS] Release POSIX file locks on fds with `CLOEXEC` flag on `execve`

kailun-qin:
Quickly adapted from a previous PR.

Resolves https://github.com/gramineproject/gramine/issues/836.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/837)

### DIFF
--- a/libos/include/libos_handle.h
+++ b/libos/include/libos_handle.h
@@ -269,6 +269,7 @@ struct libos_handle* __detach_fd_handle(struct libos_fd_handle* fd, int* flags,
                                         struct libos_handle_map* map);
 struct libos_handle* detach_fd_handle(uint32_t fd, int* flags, struct libos_handle_map* map);
 void detach_all_fds(void);
+void close_cloexec_handles(struct libos_handle_map* map);
 
 /* manage handle mapping */
 int dup_handle_map(struct libos_handle_map** new_map, struct libos_handle_map* old_map);
--- a/libos/src/bookkeep/libos_handle.c
+++ b/libos/src/bookkeep/libos_handle.c
@@ -291,6 +291,27 @@ struct libos_handle* __detach_fd_handle(struct libos_fd_handle* fd, int* flags,
     return handle;
 }
 
+static int clear_posix_locks(struct libos_handle* handle) {
+    if (handle && handle->dentry) {
+        /* Clear POSIX locks for a file. We are required to do that every time a FD is closed, even
+         * if the process holds other handles for that file, or duplicated FDs for the same
+         * handle. */
+        struct posix_lock pl = {
+            .type = F_UNLCK,
+            .start = 0,
+            .end = FS_LOCK_EOF,
+            .pid = g_process.pid,
+        };
+        int ret = posix_lock_set(handle->dentry, &pl, /*block=*/false);
+        if (ret < 0) {
+            log_warning("error releasing locks: %d", ret);
+            return ret;
+        }
+    }
+
+    return 0;
+}
+
 struct libos_handle* detach_fd_handle(uint32_t fd, int* flags,
                                       struct libos_handle_map* handle_map) {
     struct libos_handle* handle = NULL;
@@ -305,20 +326,7 @@ struct libos_handle* detach_fd_handle(uint32_t fd, int* flags,
 
     unlock(&handle_map->lock);
 
-    if (handle && handle->dentry) {
-        /* Clear POSIX locks for a file. We are required to do that every time a FD is closed, even
-         * if the process holds other handles for that file, or duplicated FDs for the same
-         * handle. */
-        struct posix_lock pl = {
-            .type = F_UNLCK,
-            .start = 0,
-            .end = FS_LOCK_EOF,
-            .pid = g_process.pid,
-        };
-        int ret = posix_lock_set(handle->dentry, &pl, /*block=*/false);
-        if (ret < 0)
-            log_warning("error releasing locks: %d", ret);
-    }
+    (void)clear_posix_locks(handle);
 
     return handle;
 }
@@ -674,6 +682,29 @@ int walk_handle_map(int (*callback)(struct libos_fd_handle*, struct libos_handle
 
     unlock(&map->lock);
     return ret;
+}
+
+void close_cloexec_handles(struct libos_handle_map* map) {
+    lock(&map->lock);
+
+    for (uint32_t i = 0; map->fd_top != FD_NULL && i <= map->fd_top; i++) {
+        struct libos_fd_handle* fd_hdl = map->map[i];
+
+        if (!HANDLE_ALLOCATED(fd_hdl))
+            continue;
+
+        if (fd_hdl->flags & FD_CLOEXEC) {
+            struct libos_handle* hdl = __detach_fd_handle(fd_hdl, NULL, map);
+
+            unlock(&map->lock);
+            (void)clear_posix_locks(hdl);
+
+            put_handle(hdl);
+            lock(&map->lock);
+        }
+    }
+
+    unlock(&map->lock);
 }
 
 BEGIN_CP_FUNC(handle) {
--- a/libos/src/sys/libos_exec.c
+++ b/libos/src/sys/libos_exec.c
@@ -15,18 +15,6 @@
 #include "libos_vma.h"
 #include "pal.h"
 
-static int close_on_exec(struct libos_fd_handle* fd_hdl, struct libos_handle_map* map) {
-    if (fd_hdl->flags & FD_CLOEXEC) {
-        struct libos_handle* hdl = __detach_fd_handle(fd_hdl, NULL, map);
-        put_handle(hdl);
-    }
-    return 0;
-}
-
-static int close_cloexec_handle(struct libos_handle_map* map) {
-    return walk_handle_map(&close_on_exec, map);
-}
-
 /* new_argp: pointer to beginning of first stack frame (argc, argv[0], ...)
  * new_auxv: pointer inside first stack frame (auxv[0], auxv[1], ...) */
 noreturn static void __libos_syscall_execve_rtld(void* new_argp, elf_auxv_t* new_auxv) {
@@ -102,8 +90,7 @@ static int libos_syscall_execve_rtld(struct libos_handle* hdl, char** argv,
     struct libos_thread* cur_thread = get_cur_thread();
     int ret;
 
-    if ((ret = close_cloexec_handle(cur_thread->handle_map)) < 0)
-        return ret;
+    close_cloexec_handles(cur_thread->handle_map);
 
     lock(&g_process.fs_lock);
     put_handle(g_process.exec);
--- a/libos/test/regression/fcntl_lock.c
+++ b/libos/test/regression/fcntl_lock.c
@@ -10,6 +10,7 @@
  * The tests usually start another process, and coordinate with it using pipes.
  */
 
+#define _GNU_SOURCE
 #include <assert.h>
 #include <err.h>
 #include <errno.h>
@@ -400,11 +401,53 @@ static void test_range_with_eof(void) {
     close_pipes(pipes);
 }
 
+/* Test: parent waits for child to close a FD (thus release a lock) on exec. */
+static void test_parent_wait_child_cloexec(void) {
+    printf("testing parent wait for child to release a lock upon exec calls (close-on-exec)...\n");
+    unlock(0, 0);
+
+    int pipes[2][2];
+    open_pipes(pipes);
+
+    pid_t pid = fork();
+    if (pid < 0)
+        err(1, "fork");
+
+    if (pid == 0) {
+        lock(F_RDLCK, 0, 100);
+        write_pipe(pipes[0]);
+        read_pipe(pipes[1]);
+
+        /* A dummy `execve` to close the FD and thus to release the taken lock. We expect child to
+         * not exit to make sure that parent can acquire the same lock indeed released by
+         * close-on-exec. */
+        char* argv[] = {(char*)"sleep", (char*)"10000", NULL};
+        execve("/bin/sleep", argv, NULL);
+        exit(1);
+    }
+
+    /* parent process: */
+
+    read_pipe(pipes[0]);
+
+    /* read lock should succeed */
+    lock(F_RDLCK, 0, 100);
+    lock_fail(F_WRLCK, 0, 100);
+    write_pipe(pipes[1]);
+    lock_wait_ok(F_WRLCK, 0, 100);
+
+    if (kill(pid, SIGKILL) < 0) {
+        err(1, "kill");
+    }
+    wait_for_child();
+    close_pipes(pipes);
+}
+
 
 int main(void) {
     setbuf(stdout, NULL);
 
-    g_fd = open(TEST_FILE, O_RDWR | O_CREAT | O_TRUNC, 0600);
+    g_fd = open(TEST_FILE, O_RDWR | O_CREAT | O_TRUNC | O_CLOEXEC, 0600);
     if (g_fd < 0)
         err(1, "open");
 
@@ -413,6 +456,7 @@ int main(void) {
     test_file_close();
     test_child_wait();
     test_parent_wait();
+    test_parent_wait_child_cloexec();
     test_range_with_eof();
 
     if (close(g_fd) < 0)
--- a/libos/test/regression/manifest.template
+++ b/libos/test/regression/manifest.template
@@ -40,6 +40,7 @@ sgx.trusted_files = [
   "file:{{ binary_dir }}/{{ entrypoint }}",
   "file:{{ binary_dir }}/exec_victim",
   "file:/bin/sh",
+  "file:/bin/sleep",
 ]
 
 sgx.insecure__protected_files_key = "ffeeddccbbaa99887766554433221100"


boryspoplawski (changes requested):
> Reviewed 1 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/837)* status: 1 of 3 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin)

___
*[`libos/src/sys/libos_exec.c` line 21 at r1](https://reviewable.io/reviews/gramineproject/gramine/837#-N9CUskKCPp2p-XbjGwV:-N9CUskKCPp2p-XbjGwW:bo4o75f) ([raw file](https://github.com/gramineproject/gramine/blob/b5d69d69d4dddca22255201c37bd978cba270c5d/libos/src/sys/libos_exec.c#L21)):*
> ```C
>     if (fd_hdl->flags & FD_CLOEXEC) {
>         struct libos_handle* hdl = __detach_fd_handle(fd_hdl, NULL, map);
>         clear_posix_locks(hdl);
> ```

Is this safe to do? We are holding `map->lock` here and `posix_lock_set` takes some locks internally and can also sleep on IPC requests.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/837)* status: 1 of 3 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski)

___
*[`libos/src/sys/libos_exec.c` line 21 at r1](https://reviewable.io/reviews/gramineproject/gramine/837#-N9CUskKCPp2p-XbjGwV:-N9CWh5L590yt0fOT9s0:bowsvig) ([raw file](https://github.com/gramineproject/gramine/blob/b5d69d69d4dddca22255201c37bd978cba270c5d/libos/src/sys/libos_exec.c#L21)):*

Is this safe to do? We are holding `map->lock` here and `posix_lock_set` takes some locks internally and can also sleep on IPC requests.
</blockquote></details>

Just remembered that you seemed to have asked this before - it acutally depends on the order when locks are aquired, but this is usually fragile (i.e., better to avoid). Q: isn't `put_handle()` a similar case here?

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/837)* status: 1 of 3 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin)

___
*[`libos/src/sys/libos_exec.c` line 21 at r1](https://reviewable.io/reviews/gramineproject/gramine/837#-N9CUskKCPp2p-XbjGwV:-N9DbwV_7f-LYrNmeZ28:bbnncb3) ([raw file](https://github.com/gramineproject/gramine/blob/b5d69d69d4dddca22255201c37bd978cba270c5d/libos/src/sys/libos_exec.c#L21)):*
> it acutally depends on the order when locks are aquired, but this is usually fragile (i.e., better to avoid). 

Yes. These are a) multiple locks b) they are all unrelated so it hard to define a order on them and maintain it through the whole codebase. We should rather avoid the problem (i.e. design the solution so we never have to take all of them).

> Q: isn't put_handle() a similar case here?

No, `put_handle` only decreases refcount and possibly releases an object. There is no (need for) locking there.

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/837)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski)

___
*[`libos/src/sys/libos_exec.c` line 21 at r1](https://reviewable.io/reviews/gramineproject/gramine/837#-N9CUskKCPp2p-XbjGwV:-N9fHl_cE90EBayHc-B8:b-iryrzj) ([raw file](https://github.com/gramineproject/gramine/blob/b5d69d69d4dddca22255201c37bd978cba270c5d/libos/src/sys/libos_exec.c#L21)):*
> can also sleep on IPC requests.

I don't think it is true: https://github.com/gramineproject/gramine/blob/2e700c1b795dd2430f0cec68ec9d2a1bd1099c18/libos/include/libos_fs_lock.h#L67-L68

> We should rather avoid the problem (i.e. design the solution so we never have to take all of them).

At this point (when we are in `close_on_exec()`), there is a single thread left that called into `exec()` syscall. I am fairly certain that the implementation of FS locks will never try to acquire `map->lock`. I'm also fairly certain that the `exec()` code here will never touch FS-lock-internal locks. So I feel that we are safe with the current PR. So maybe we could put a comment here explaining our rationale that no deadlocks can happen?

But I agree that this is brittle. But also, doesn't `detach_fd_handle()` and its callers have the exact same problem (of unrelated locks being taken in arbitrary order)?

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/837)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv and @kailun-qin)

___
*[`libos/src/sys/libos_exec.c` line 21 at r1](https://reviewable.io/reviews/gramineproject/gramine/837#-N9CUskKCPp2p-XbjGwV:-N9fwO0J8f-6epZyN8Cp:b-3fuhvw) ([raw file](https://github.com/gramineproject/gramine/blob/b5d69d69d4dddca22255201c37bd978cba270c5d/libos/src/sys/libos_exec.c#L21)):*
>> can also sleep on IPC requests.
>> 
> I don't think it is true: 

Ok, the IPC operation will be finalized eventually, but sill make take a lot of time (this is not a blocker though).

> But also, doesn't detach_fd_handle() and its callers have the exact same problem (of unrelated locks being taken in arbitrary order)?

We don't call `detach_fd_handle` with any locks taken (except in `libos/src/fs/pipe/fs.c`, but i's also a hack).

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/837)* status: 2 of 3 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[`libos/src/sys/libos_exec.c` line 21 at r1](https://reviewable.io/reviews/gramineproject/gramine/837#-N9CUskKCPp2p-XbjGwV:-N9gIF-fBtS88pg7diME:b-9rw3hu) ([raw file](https://github.com/gramineproject/gramine/blob/b5d69d69d4dddca22255201c37bd978cba270c5d/libos/src/sys/libos_exec.c#L21)):*

>> can also sleep on IPC requests.
>> 
> I don't think it is true: 

Ok, the IPC operation will be finalized eventually, but sill make take a lot of time (this is not a blocker though).

> But also, doesn't detach_fd_handle() and its callers have the exact same problem (of unrelated locks being taken in arbitrary order)?

We don't call `detach_fd_handle` with any locks taken (except in `libos/src/fs/pipe/fs.c`, but i's also a hack).
</blockquote></details>

Thanks for checking! Added some comments for explanation.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/837)* status: 2 of 3 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`libos/src/sys/libos_exec.c` line 21 at r1](https://reviewable.io/reviews/gramineproject/gramine/837#-N9CUskKCPp2p-XbjGwV:-N9l-nbk8wb-cVGdufBQ:b-jnku0t) ([raw file](https://github.com/gramineproject/gramine/blob/b5d69d69d4dddca22255201c37bd978cba270c5d/libos/src/sys/libos_exec.c#L21)):*

I'm still against this solution. While this might work in the current state of affairs, we might accidentally break it (and debugging such issues is no fun). I would say we should stick to the old rules of either defining an ordering on locks or never taking them together and do not create any precedents.

cc @mkow we might need a tie-breaker

> At this point (when we are in close_on_exec()), there is a single thread left that called into exec() syscall. 

Single user app thread. There are also IPC worker and async helper thread (or whatever it's called).
</blockquote></details>

So what if we introduce a `walk_handle_map_no_locking()` version of the function? And we can call it in `close_cloexec_handle()`. In this case, there will be no `map->lock` held, and we won't need a lengthy comment that @kailun-qin added now.

This introduced version of the function may even be embedded here, because this is the only place where it's currently needed. @boryspoplawski ?

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/837)* status: 2 of 3 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`libos/src/sys/libos_exec.c` line 21 at r1](https://reviewable.io/reviews/gramineproject/gramine/837#-N9CUskKCPp2p-XbjGwV:-N9lJ8yh0XUp0g58e31p:borjar8) ([raw file](https://github.com/gramineproject/gramine/blob/b5d69d69d4dddca22255201c37bd978cba270c5d/libos/src/sys/libos_exec.c#L21)):*

So what if we introduce a `walk_handle_map_no_locking()` version of the function? And we can call it in `close_cloexec_handle()`. In this case, there will be no `map->lock` held, and we won't need a lengthy comment that @kailun-qin added now.

This introduced version of the function may even be embedded here, because this is the only place where it's currently needed. @boryspoplawski ?
</blockquote></details>

Yeah, that would probably work. I would just move `close_cloexec_handles` into `libos/src/bookkeep/libos_handle.c` and inline the loop in there.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/837)* status: 0 of 3 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

___
*[`libos/src/sys/libos_exec.c` line 21 at r1](https://reviewable.io/reviews/gramineproject/gramine/837#-N9CUskKCPp2p-XbjGwV:-N9lVaxG6cyb4ITvpe5o:bsmrizv) ([raw file](https://github.com/gramineproject/gramine/blob/b5d69d69d4dddca22255201c37bd978cba270c5d/libos/src/sys/libos_exec.c#L21)):*

Yeah, that would probably work. I would just move `close_cloexec_handles` into `libos/src/bookkeep/libos_handle.c` and inline the loop in there.
</blockquote></details>

Updated, PTAL if it's sth you'd expect.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/837)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

___
*[`libos/include/libos_handle.h` line 271 at r3](https://reviewable.io/reviews/gramineproject/gramine/837#-N9lX9QG3tE9fC3q6zIH:-N9lX9QG3tE9fC3q6zII:b1wk7yr) ([raw file](https://github.com/gramineproject/gramine/blob/43e365b80cc0885a1d9e0fb2af3a8badec3666fb/libos/include/libos_handle.h#L271)):*
> ```objc
>                                         struct libos_handle_map* map);
> struct libos_handle* detach_fd_handle(uint32_t fd, int* flags, struct libos_handle_map* map);
> void clear_posix_locks(struct libos_handle* handle);
> ```

This is not needed.
___
*[`libos/include/libos_handle.h` line 272 at r3](https://reviewable.io/reviews/gramineproject/gramine/837#-N9lX5KQ1g7b6TXXDKVX:-N9lX5KQ1g7b6TXXDKVY:b-t1vsyb) ([raw file](https://github.com/gramineproject/gramine/blob/43e365b80cc0885a1d9e0fb2af3a8badec3666fb/libos/include/libos_handle.h#L272)):*
> ```objc
> struct libos_handle* detach_fd_handle(uint32_t fd, int* flags, struct libos_handle_map* map);
> void clear_posix_locks(struct libos_handle* handle);
> void close_cloexec_handle(struct libos_handle_map* map);
> ```

`handles` (plural)
___
*[`libos/src/bookkeep/libos_handle.c` line 263 at r3](https://reviewable.io/reviews/gramineproject/gramine/837#-N9lXE8943XUKZQ67MY9:-N9lXE8943XUKZQ67MYA:brynr1l) ([raw file](https://github.com/gramineproject/gramine/blob/90f8b9d7eacfa1c9bc07089d9d00ea774888435f/libos/src/bookkeep/libos_handle.c#L263)):*
> ```C
> struct libos_handle* __detach_fd_handle(struct libos_fd_handle* fd, int* flags,
>                                         struct libos_handle_map* map) {
>     assert(locked(&map->lock));
> ```

Why did you remove this assert? It should be here
___
*[`libos/src/bookkeep/libos_handle.c` line 292 at r3](https://reviewable.io/reviews/gramineproject/gramine/837#-N9lXpQ798O7nRCq4vYq:-N9lXpQ8EcsNFrUcXZ74:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/43e365b80cc0885a1d9e0fb2af3a8badec3666fb/libos/src/bookkeep/libos_handle.c#L292)):*
> ```C
> }
> 
> void clear_posix_locks(struct libos_handle* handle) {
> ```





_Suggestion:_
```C
static void clear_posix_locks(struct libos_handle* handle) {
```
___
*[`libos/src/bookkeep/libos_handle.c` line 673 at r3](https://reviewable.io/reviews/gramineproject/gramine/837#-N9lY798AAZjTuCCOk9x:-N9lY798AAZjTuCCOk9y:b-qlupit) ([raw file](https://github.com/gramineproject/gramine/blob/43e365b80cc0885a1d9e0fb2af3a8badec3666fb/libos/src/bookkeep/libos_handle.c#L673)):*
> ```C
> 
> void close_cloexec_handle(struct libos_handle_map* map) {
>     for (uint32_t i = 0; map->fd_top != FD_NULL && i <= map->fd_top; i++) {
> ```

Where is locking? You need to take `map->lock`
___
*[`libos/src/bookkeep/libos_handle.c` line 681 at r3](https://reviewable.io/reviews/gramineproject/gramine/837#-N9lYHmK5plgicpk4NVW:-N9lYHmL578EMh-muvHa:b-tgi05a) ([raw file](https://github.com/gramineproject/gramine/blob/43e365b80cc0885a1d9e0fb2af3a8badec3666fb/libos/src/bookkeep/libos_handle.c#L681)):*
> ```C
>         if (fd_hdl->flags & FD_CLOEXEC) {
>             struct libos_handle* hdl = __detach_fd_handle(fd_hdl, NULL, map);
>             clear_posix_locks(hdl);
> ```

The idea is that you release lock here, but make sure that when you reaquire it, it's still safe to continue iteration (i.e. that nothing gets stale)

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/837)* status: 0 of 3 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`libos/include/libos_handle.h` line 271 at r3](https://reviewable.io/reviews/gramineproject/gramine/837#-N9lX9QG3tE9fC3q6zIH:-N9l_LCQBi_5okclE2ji:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/43e365b80cc0885a1d9e0fb2af3a8badec3666fb/libos/include/libos_handle.h#L271)):*

This is not needed.
</blockquote></details>

Done.
___
*[`libos/include/libos_handle.h` line 272 at r3](https://reviewable.io/reviews/gramineproject/gramine/837#-N9lX5KQ1g7b6TXXDKVX:-N9l_LZ0CsQHtPo5pxRA:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/43e365b80cc0885a1d9e0fb2af3a8badec3666fb/libos/include/libos_handle.h#L272)):*

`handles` (plural)
</blockquote></details>

Done.
___
*[`libos/src/bookkeep/libos_handle.c` line 263 at r3](https://reviewable.io/reviews/gramineproject/gramine/837#-N9lXE8943XUKZQ67MY9:-N9laDXDDEq6YHpuX1hj:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/90f8b9d7eacfa1c9bc07089d9d00ea774888435f/libos/src/bookkeep/libos_handle.c#L263)):*

Why did you remove this assert? It should be here
</blockquote></details>

Done.
___
*[`libos/src/bookkeep/libos_handle.c` line 292 at r3](https://reviewable.io/reviews/gramineproject/gramine/837#-N9lXpQ798O7nRCq4vYq:-N9l_NsBBLaNPhE69rx4:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/43e365b80cc0885a1d9e0fb2af3a8badec3666fb/libos/src/bookkeep/libos_handle.c#L292)):*
> ```C
> }
> 
> void clear_posix_locks(struct libos_handle* handle) {
> ```

Done.
___
*[`libos/src/bookkeep/libos_handle.c` line 673 at r3](https://reviewable.io/reviews/gramineproject/gramine/837#-N9lY798AAZjTuCCOk9x:-N9l_jMRA2hh31kc7RsU:bipbblu) ([raw file](https://github.com/gramineproject/gramine/blob/43e365b80cc0885a1d9e0fb2af3a8badec3666fb/libos/src/bookkeep/libos_handle.c#L673)):*

Where is locking? You need to take `map->lock`
</blockquote></details>

Isn't `walk_handle_map_no_locking()` suggested (where `no map->lock` is held)?
___
*[`libos/src/bookkeep/libos_handle.c` line 681 at r3](https://reviewable.io/reviews/gramineproject/gramine/837#-N9lYHmK5plgicpk4NVW:-N9lcB7l69AWUwIbIWRV:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/43e365b80cc0885a1d9e0fb2af3a8badec3666fb/libos/src/bookkeep/libos_handle.c#L681)):*

The idea is that you release lock here, but make sure that when you reaquire it, it's still safe to continue iteration (i.e. that nothing gets stale)
</blockquote></details>

Done.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/837)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

___
*[`libos/src/bookkeep/libos_handle.c` line 673 at r3](https://reviewable.io/reviews/gramineproject/gramine/837#-N9lY798AAZjTuCCOk9x:-N9lh2hf-AGNpSua_Zg0:b3oovnw) ([raw file](https://github.com/gramineproject/gramine/blob/43e365b80cc0885a1d9e0fb2af3a8badec3666fb/libos/src/bookkeep/libos_handle.c#L673)):*

Isn't `walk_handle_map_no_locking()` suggested (where `no map->lock` is held)?
</blockquote></details>

I think the suggestion was that the callback is called without any locks.
___
*[`libos/src/bookkeep/libos_handle.c` line 294 at r4](https://reviewable.io/reviews/gramineproject/gramine/837#-N9lh_MJ4zicplbKMf7X:-N9lh_MJ4zicplbKMf7Y:b2pm316) ([raw file](https://github.com/gramineproject/gramine/blob/15c2af9ef4c0dc7cccdf192b4b439f0cd402dc16/libos/src/bookkeep/libos_handle.c#L294)):*
> ```C
> }
> 
> static void clear_posix_locks(struct libos_handle* handle) {
> ```

Actually, this function should return errors. 
___
*[`libos/src/bookkeep/libos_handle.c` line 690 at r4](https://reviewable.io/reviews/gramineproject/gramine/837#-N9lhG7oB77CP8lO4DDY:-N9lhG7p0sodTIFYWYxH:b9i09e) ([raw file](https://github.com/gramineproject/gramine/blob/15c2af9ef4c0dc7cccdf192b4b439f0cd402dc16/libos/src/bookkeep/libos_handle.c#L690)):*
> ```C
>             lock(&map->lock);
> 
>             put_handle(hdl);
> ```

Please move this before `lock`

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/837)* status: 2 of 3 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`libos/src/bookkeep/libos_handle.c` line 673 at r3](https://reviewable.io/reviews/gramineproject/gramine/837#-N9lY798AAZjTuCCOk9x:-N9ljtty8GEBH7JXokeV:b-3b3mqc) ([raw file](https://github.com/gramineproject/gramine/blob/43e365b80cc0885a1d9e0fb2af3a8badec3666fb/libos/src/bookkeep/libos_handle.c#L673)):*

I think the suggestion was that the callback is called without any locks.
</blockquote></details>

OK, got it. I misunderstood that...
___
*[`libos/src/bookkeep/libos_handle.c` line 294 at r4](https://reviewable.io/reviews/gramineproject/gramine/837#-N9lh_MJ4zicplbKMf7X:-N9ljr-CAGzQQM4NTgoC:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/15c2af9ef4c0dc7cccdf192b4b439f0cd402dc16/libos/src/bookkeep/libos_handle.c#L294)):*

Actually, this function should return errors. 
</blockquote></details>

Done.
___
*[`libos/src/bookkeep/libos_handle.c` line 690 at r4](https://reviewable.io/reviews/gramineproject/gramine/837#-N9lhG7oB77CP8lO4DDY:-N9lk4QX9z861rIxmE3E:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/15c2af9ef4c0dc7cccdf192b4b439f0cd402dc16/libos/src/bookkeep/libos_handle.c#L690)):*

Please move this before `lock`
</blockquote></details>

Done.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/837)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

___
*[`libos/src/bookkeep/libos_handle.c` line 673 at r3](https://reviewable.io/reviews/gramineproject/gramine/837#-N9lY798AAZjTuCCOk9x:-N9lw7yXAHbkFpyFvHC0:b-rhev3p) ([raw file](https://github.com/gramineproject/gramine/blob/43e365b80cc0885a1d9e0fb2af3a8badec3666fb/libos/src/bookkeep/libos_handle.c#L673)):*

OK, got it. I misunderstood that...
</blockquote></details>

Yeah, the function name was mislading
___
*[`libos/src/bookkeep/libos_handle.c` line 329 at r5](https://reviewable.io/reviews/gramineproject/gramine/837#-N9lwd2096aox4TxloPs:-N9lwKMU4pB-fgyyhSdf:b-egnfbp) ([raw file](https://github.com/gramineproject/gramine/blob/f3eee8652a75b7aa0a8465763fb758d5dad5f050/libos/src/bookkeep/libos_handle.c#L329)):*
> ```C
>     unlock(&handle_map->lock);
> 
>     clear_posix_locks(handle);
> ```

I'm not sure what to do here. The previous code did not check for errors here and there's no easy way to do that, but I feel like we should handle it somehow?
___
*[`libos/src/bookkeep/libos_handle.c` line 691 at r5](https://reviewable.io/reviews/gramineproject/gramine/837#-N9lwDiQDw7-lKQayd-S:-N9lwDiQDw7-lKQayd-T:b-dw3cqm) ([raw file](https://github.com/gramineproject/gramine/blob/f3eee8652a75b7aa0a8465763fb758d5dad5f050/libos/src/bookkeep/libos_handle.c#L691)):*
> ```C
> 
>             unlock(&map->lock);
>             clear_posix_locks(hdl);
> ```

Please add error checking

mkow (changes requested):
> Reviewed 2 of 3 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/837)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/837#-N9nMPsnBmWQfRtyMUqY:-N9nMPsnBmWQfRtyMUqZ:bn67dh2) (no related file):*
Could you add a test for this? (or change some existing, related one, to test this flow)

___
___
*[`libos/src/bookkeep/libos_handle.c` line 329 at r5](https://reviewable.io/reviews/gramineproject/gramine/837#-N9lwd2096aox4TxloPs:-N9nMzaW7rvjwdAHF2Vk:b42welt) ([raw file](https://github.com/gramineproject/gramine/blob/f3eee8652a75b7aa0a8465763fb758d5dad5f050/libos/src/bookkeep/libos_handle.c#L329)):*

I'm not sure what to do here. The previous code did not check for errors here and there's no easy way to do that, but I feel like we should handle it somehow?
</blockquote></details>

Please at least ignore the result more explicitly, i.e. with `(void)` cast.
___
*[`libos/src/bookkeep/libos_handle.c` line 681 at r5](https://reviewable.io/reviews/gramineproject/gramine/837#-N9nOV606gt4VZcsOANc:-N9nOV606gt4VZcsOANd:b71vvhr) ([raw file](https://github.com/gramineproject/gramine/blob/f3eee8652a75b7aa0a8465763fb758d5dad5f050/libos/src/bookkeep/libos_handle.c#L681)):*
> ```C
>     lock(&map->lock);
> 
>     for (uint32_t i = 0; map->fd_top != FD_NULL && i <= map->fd_top; i++) {
> ```

FYI/warning for reviewers: This way of iteration is dangerous, although currently I don't any direct bug resulting from it.
I think it's possible to reach `map->fd_top == UINT32_MAX`, in which case this loop will deallocate everything, and then start again from 0. This isn't a problem, because either `map->fd_top != FD_NULL` or `HANDLE_ALLOCATED(fd_hdl)` checks will stop it from doing anything on freed FDs.
Anyways, this quirk is probably hard to notice, and IMO it's risky to leave it like this, without a warning comment?

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/837)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

___
*[`libos/src/bookkeep/libos_handle.c` line 681 at r5](https://reviewable.io/reviews/gramineproject/gramine/837#-N9nOV606gt4VZcsOANc:-N9nWEp0-OO1meTPnhDY:bwqgld1) ([raw file](https://github.com/gramineproject/gramine/blob/f3eee8652a75b7aa0a8465763fb758d5dad5f050/libos/src/bookkeep/libos_handle.c#L681)):*

FYI/warning for reviewers: This way of iteration is dangerous, although currently I don't any direct bug resulting from it.
I think it's possible to reach `map->fd_top == UINT32_MAX`, in which case this loop will deallocate everything, and then start again from 0. This isn't a problem, because either `map->fd_top != FD_NULL` or `HANDLE_ALLOCATED(fd_hdl)` checks will stop it from doing anything on freed FDs.
Anyways, this quirk is probably hard to notice, and IMO it's risky to leave it like this, without a warning comment?
</blockquote></details>

I think `FD_NULL == UINT32_MAX` and it's done this way exactly to support cases like htis

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/837)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/837#-N9nMPsnBmWQfRtyMUqY:-N9n_oSD1P2MMvpOnkbv:b-6c5jmj) (no related file):*

Could you add a test for this? (or change some existing, related one, to test this flow)
</blockquote></details>

I think the flow can be covered by the exisiting one: [pipe_ocloexec](https://github.com/gramineproject/gramine/blob/a014bc685bbb3cb8b37f90b482e6ab5de620be56/libos/test/regression/pipe_ocloexec.c). WDYT?

___
___
*[`libos/src/bookkeep/libos_handle.c` line 329 at r5](https://reviewable.io/reviews/gramineproject/gramine/837#-N9lwd2096aox4TxloPs:-N9nZCLPDzfD-YhpQlfG:bd8qhui) ([raw file](https://github.com/gramineproject/gramine/blob/f3eee8652a75b7aa0a8465763fb758d5dad5f050/libos/src/bookkeep/libos_handle.c#L329)):*

Please at least ignore the result more explicitly, i.e. with `(void)` cast.
</blockquote></details>

What would you expect as the right behavior on errors? Currently, we simply warn w/ log on clear lock failures (in this introduced function, also here: https://github.com/gramineproject/gramine/blob/2e700c1b795dd2430f0cec68ec9d2a1bd1099c18/libos/src/sys/libos_exit.c#L99-L101).

I assume explicitly ignore the result should be OK?

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/837)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[`libos/src/bookkeep/libos_handle.c` line 329 at r5](https://reviewable.io/reviews/gramineproject/gramine/837#-N9lwd2096aox4TxloPs:-N9nb0qS9laBxkbRH7d6:b-bpkrpn) ([raw file](https://github.com/gramineproject/gramine/blob/f3eee8652a75b7aa0a8465763fb758d5dad5f050/libos/src/bookkeep/libos_handle.c#L329)):*

What would you expect as the right behavior on errors? Currently, we simply warn w/ log on clear lock failures (in this introduced function, also here: https://github.com/gramineproject/gramine/blob/2e700c1b795dd2430f0cec68ec9d2a1bd1099c18/libos/src/sys/libos_exit.c#L99-L101).

I assume explicitly ignore the result should be OK?
</blockquote></details>

Done by explicitly ignoring. Let me know if you really want to handle it in some way.
___
*[`libos/src/bookkeep/libos_handle.c` line 691 at r5](https://reviewable.io/reviews/gramineproject/gramine/837#-N9lwDiQDw7-lKQayd-S:-N9nb-4m6BiOhXP8uFar:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/f3eee8652a75b7aa0a8465763fb758d5dad5f050/libos/src/bookkeep/libos_handle.c#L691)):*

Please add error checking
</blockquote></details>

Done.

dimakuv (changes requested):
> Reviewed 1 of 3 files at r4, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/837)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/837#-N9nMPsnBmWQfRtyMUqY:-N9p-HO194tTeu8pFsCB:b-fxq4xv) (no related file):*

I think the flow can be covered by the exisiting one: [pipe_ocloexec](https://github.com/gramineproject/gramine/blob/a014bc685bbb3cb8b37f90b482e6ab5de620be56/libos/test/regression/pipe_ocloexec.c). WDYT?
</blockquote></details>

@kailun-qin I guess you can take `pipe_ocloexec.c` flow as an example, but you should create a new test. The test will open some tmp file (with CLOEXEC), then it will perform a fork, then a child process will acquire a lock and then perform a dummy `execve` (to close the tmp file and thus to release the taken lock), and the parent process will wait until the child is done and check whether it can acquire the same lock (the parent should succeed, otherwise the POSIX lock was not released on CLOEXEC).

Some primitives and operations can be peeked from here: https://github.com/gramineproject/gramine/blob/master/libos/test/regression/fcntl_lock.c

I would also verify first that the above test indeed fails on current master, and then verify that your PR fixes this test.

___

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/837)* status: 3 of 5 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/837#-N9nMPsnBmWQfRtyMUqY:-N9pcsiS3_jRzRsaSXZB:b8kcmhk) (no related file):*

@kailun-qin I guess you can take `pipe_ocloexec.c` flow as an example, but you should create a new test. The test will open some tmp file (with CLOEXEC), then it will perform a fork, then a child process will acquire a lock and then perform a dummy `execve` (to close the tmp file and thus to release the taken lock), and the parent process will wait until the child is done and check whether it can acquire the same lock (the parent should succeed, otherwise the POSIX lock was not released on CLOEXEC).

Some primitives and operations can be peeked from here: https://github.com/gramineproject/gramine/blob/master/libos/test/regression/fcntl_lock.c

I would also verify first that the above test indeed fails on current master, and then verify that your PR fixes this test.
</blockquote></details>

Added a test, PTAL, thanks.

___

boryspoplawski (changes requested):
> Reviewed 1 of 1 files at r6, 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/837)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @kailun-qin, and @mkow)

___
*[`libos/test/regression/fcntl_lock.c` line 441 at r7](https://reviewable.io/reviews/gramineproject/gramine/837#-N9pihmxB_539wUzEyY7:-N9pihmxB_539wUzEyY8:b-weawbo) ([raw file](https://github.com/gramineproject/gramine/blob/9a6ef2902fb6459ea221c79a45b9c078c3992d6a/libos/test/regression/fcntl_lock.c#L441)):*
> ```C
>     if (kill(pid, SIGKILL) < 0) {
>         err(1, "kill");
>     }
> ```

wait missing (I think this test has a wrapper for it)

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/837)* status: 4 of 5 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

___
*[`libos/test/regression/fcntl_lock.c` line 441 at r7](https://reviewable.io/reviews/gramineproject/gramine/837#-N9pihmxB_539wUzEyY7:-N9pmc-H7tV8UGzI1rRq:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9a6ef2902fb6459ea221c79a45b9c078c3992d6a/libos/test/regression/fcntl_lock.c#L441)):*

wait missing (I think this test has a wrapper for it)
</blockquote></details>

Done.

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/837)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

dimakuv (changes requested):
> Reviewed 1 of 3 files at r4, 1 of 2 files at r7, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/837)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/837#-N9nMPsnBmWQfRtyMUqY:-N9qIPSb6AIwB8nXS81m:brobcbv) (no related file):*

Added a test, PTAL, thanks.
</blockquote></details>

Awesome test!

Did you verify that the newly added test indeed fails on the master branch of Gramine? Just for sanity.

___

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/837)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/837#-N9nMPsnBmWQfRtyMUqY:-N9qPkJq5M5vWiikONIW:b-b0bgva) (no related file):*

Awesome test!

Did you verify that the newly added test indeed fails on the master branch of Gramine? Just for sanity.
</blockquote></details>

Yes, I've verified it against our lastest master (that it indeed fails) .

___

mkow (changes requested):
Just one wording of a log message, otherwise looks good.

> Reviewed 1 of 1 files at r6, 1 of 2 files at r7, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/837#-:-N9s9gA7BOh4EAHulVDu:ba6typ)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @kailun-qin)

___
*[`libos/src/bookkeep/libos_handle.c` line 681 at r5](https://reviewable.io/reviews/gramineproject/gramine/837#-N9nOV606gt4VZcsOANc:-N9s6mzOBClJgiJxWLvG:b568pco) ([raw file](https://github.com/gramineproject/gramine/blob/f3eee8652a75b7aa0a8465763fb758d5dad5f050/libos/src/bookkeep/libos_handle.c#L681)):*

I think `FD_NULL == UINT32_MAX` and it's done this way exactly to support cases like htis
</blockquote></details>

Ah, right, so that's how it works...
___
*[`libos/test/regression/fcntl_lock.c` line 406 at r8](https://reviewable.io/reviews/gramineproject/gramine/837#-N9s8gzb6v67sV-2Qlpt:-N9s8gzb6v67sV-2Qlpu:b-hoyihv) ([raw file](https://github.com/gramineproject/gramine/blob/c73ba5d25560039e7ccd5f44d0d4ac100a76622b/libos/test/regression/fcntl_lock.c#L406)):*
> ```C
> /* Test: parent waits for child to close a FD (thus release a lock) on exec. */
> static void test_parent_wait_child_cloexec(void) {
>     printf("testing parent wait child cloexec...\n");
> ```

This is so ungrammatical that I really wouldn't understand what's happening here without reading the source. It's just 5 nouns now glued together... ;)



_Code quote:_
```C
testing parent wait child cloexec
```

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/837)* status: 4 of 5 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

___
*[`libos/test/regression/fcntl_lock.c` line 406 at r8](https://reviewable.io/reviews/gramineproject/gramine/837#-N9s8gzb6v67sV-2Qlpt:-N9s_zvY7MMkCqJwq4Ua:b-pmso9w) ([raw file](https://github.com/gramineproject/gramine/blob/c73ba5d25560039e7ccd5f44d0d4ac100a76622b/libos/test/regression/fcntl_lock.c#L406)):*

This is so ungrammatical that I really wouldn't understand what's happening here without reading the source. It's just 5 nouns now glued together... ;)
</blockquote></details>

;) Updated the wording, hope it's a bit better.

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/837)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`libos/test/regression/fcntl_lock.c` line 406 at r8](https://reviewable.io/reviews/gramineproject/gramine/837#-N9s8gzb6v67sV-2Qlpt:-NA3WPhh1bDEK7Jg8QRS:b-y2o7ab) ([raw file](https://github.com/gramineproject/gramine/blob/c73ba5d25560039e7ccd5f44d0d4ac100a76622b/libos/test/regression/fcntl_lock.c#L406)):*

;) Updated the wording, hope it's a bit better.
</blockquote></details>

But that string with 5 keywords explained the test perfectly with no redundancy :)

dimakuv:
Jenkins, test this please

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/837)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

___
*[`libos/test/regression/fcntl_lock.c` line 406 at r8](https://reviewable.io/reviews/gramineproject/gramine/837#-N9s8gzb6v67sV-2Qlpt:-NA9UdgoA664GqhMe5mK:b-ucct20) ([raw file](https://github.com/gramineproject/gramine/blob/c73ba5d25560039e7ccd5f44d0d4ac100a76622b/libos/test/regression/fcntl_lock.c#L406)):*
> But that string with 5 keywords explained the test perfectly with no redundancy :)

I couldn't understand it, it was just 5 English nouns glued together without even any prepositions, which didn't parse to any meaningful sentence ¯\\\_(ツ)\_/¯ Maybe if you knew what the test does before reading this text you could understand it, but that's not how describing something works ;)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/837)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

dimakuv:
Jenkins, test this please

boryspoplawski (approved):
> Reviewed 1 of 1 files at r9, 2 of 2 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/837)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #840, opened 2022-08-12
Unexpected memory fault occurred inside PAL (strnlen at strlen.c)


Villain88:
## Description of the problem
I found a very strange bug that is reproduced only by a set of certain steps and using the following gramineproject components:
gsc, encrypted_fs, tracing

## Steps to reproduce

1. Build a container with the following content
```
FROM ubuntu:20.04

RUN mkdir -p /solution/inputs /solution/output /solution/run
RUN apt update \
    && apt install -y libnss-mdns libnss-myhostname lsb-release \
    && apt install -y python3.8 python3-pip libcurl4

ENTRYPOINT ["/usr/bin/python3.8", "/solution/run/entrypoint.py"]
```
2. Graminize image using GSC with option **-d** (debug) and a manifest with the following content:
```
...
fs.mounts = [
  { type = "encrypted", path = "/solution/run", uri = "file:/solution/run", key_name = "default" },
]

fs.insecure__keys.default = "ffeeddccbbaa99887766554433221100"
loader.env.PYTHONDONTWRITEBYTECODE = "1"
loader.env.PYTHONUSERBASE = "/solution/run/pypi"
...
```

3. Take an example of pytorch from [here](https://github.com/gramineproject/examples/tree/master/pytorch)

4. Go to the directory with pytorch and do something like the following:

```
mv pytorchexample.py entrypoint.py
export PYTHONUSERBASE=$PWD/pypi
pip3 install --user --ignore-installed torchvision pillow
python3 download-pretrained-model.py
#encrypt pypi, script and resources
gramine-sgx-pf-crypt encrypt -w <KEY_ffeeddccbbaa99887766554433221100_FILE> -i <PYTORCH_FOLDER> -o /solution/run
```

5.  Run a graminized container:
```
docker run --rm --device=/dev/sgx_enclave -v /dev/sgx:/dev/sgx -v  /solution/run:/solution/run -v /var/run/aesmd/aesm.socket:/var/run/aesmd/aesm.socket -w /solution/run <GRAMINIZED_IMAGE>
```

## Expected results
The calculation results will appear in the encrypted file results.txt
## Actual results
Exception
```
[P1:T1:python3.8] debug: glibc register library /solution/run/pypi/lib/python3.8/site-packages/torch/lib/libtorch_global_deps.so loaded at 0x1c686d000
[P1:T1:python3.8] debug: append_r_debug: adding (null) at 0x1c686d000
[...]

PR #841, opened 2022-08-15
[LibOS] Set `handle->uri` on opening a protected file

kailun-qin:
Fixes https://github.com/gramineproject/gramine/issues/840.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/841)

### DIFF
--- a/libos/src/fs/chroot/encrypted.c
+++ b/libos/src/fs/chroot/encrypted.c
@@ -176,6 +176,11 @@ static int chroot_encrypted_open(struct libos_handle* hdl, struct libos_dentry* 
     __UNUSED(flags);
 
     int ret;
+    char* uri = NULL;
+
+    ret = chroot_dentry_uri(dent, dent->inode->type, &uri);
+    if (ret < 0)
+        return ret;
 
     if (dent->inode->type == S_IFREG) {
         struct libos_encrypted_file* enc = dent->inode->data;
@@ -183,15 +188,22 @@ static int chroot_encrypted_open(struct libos_handle* hdl, struct libos_dentry* 
         lock(&dent->inode->lock);
         ret = encrypted_file_get(enc);
         unlock(&dent->inode->lock);
-        if (ret < 0)
-            return ret;
+        if (ret < 0) {
+            goto out;
+        }
     }
 
+    hdl->uri = uri;
+    uri = NULL;
     hdl->inode = dent->inode;
     get_inode(dent->inode);
     hdl->type = TYPE_CHROOT_ENCRYPTED;
     hdl->pos = 0;
-    return 0;
+    ret = 0;
+
+out:
+    free(uri);
+    return ret;
 }
 
 static int chroot_encrypted_creat(struct libos_handle* hdl, struct libos_dentry* dent, int flags,


mkow:
Jenkins, test this please

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/841)* status: 0 of 1 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/841#-N9Wuazn252_ENIQVHHC:-N9Wuazo9hD6C6YpnEND:b-caafxq) (no related file):*
Could you elaborate on why this is the correct fix? I'm not familiar with most of FS code and it seems that it allows for having `handle->uri == NULL` and this code is invalid: https://github.com/gramineproject/gramine/blob/d8217c8ad08fc2be289384b90100e29c0506c097/libos/src/libos_rtld.c#L1037

___
___
*[`libos/src/fs/chroot/encrypted.c` line 184 at r1](https://reviewable.io/reviews/gramineproject/gramine/841#-N9WqBN_-r_GHwJiIBV-:-N9WqBN_-r_GHwJiIBV0:b-f5rufv) ([raw file](https://github.com/gramineproject/gramine/blob/89521c7dcfdb9760ecc624e19d620d02c093c41e/libos/src/fs/chroot/encrypted.c#L184)):*
> ```C
>     if (ret < 0)
>         return ret;
> ```

Couldn't we just always set `hdl->uri` here (i.e. regardless of whether code below will succeed)?
___
*[`libos/src/fs/chroot/encrypted.c` line 209 at r1](https://reviewable.io/reviews/gramineproject/gramine/841#-N9WuZ9aCjWF7QxbC-8X:-N9WuZ9aCjWF7QxbC-8Y:baki5ok) ([raw file](https://github.com/gramineproject/gramine/blob/89521c7dcfdb9760ecc624e19d620d02c093c41e/libos/src/fs/chroot/encrypted.c#L209)):*
> ```C
> }
> 
> static int chroot_encrypted_creat(struct libos_handle* hdl, struct libos_dentry* dent, int flags,
> ```

This doesn't set `uri` either.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/841)* status: 0 of 1 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/841#-N9Wuazn252_ENIQVHHC:-N9WutNb1-noaTthR_7J:b-gwmmd0) (no related file):*

Could you elaborate on why this is the correct fix? I'm not familiar with most of FS code and it seems that it allows for having `handle->uri == NULL` and this code is invalid: https://github.com/gramineproject/gramine/blob/d8217c8ad08fc2be289384b90100e29c0506c097/libos/src/libos_rtld.c#L1037
</blockquote></details>

cc @dimakuv @mkow maybe you know more about this

___

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/841)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/841#-N9Wuazn252_ENIQVHHC:-N9aSwiw00qkoZLHc1TL:b-e1u8eg) (no related file):*

cc @dimakuv @mkow maybe you know more about this
</blockquote></details>

The LibOS codebase doesn't require `handle::uri` to be set: https://github.com/gramineproject/gramine/blob/aaf34c87bfebd071cdb18a187a7b7cbe510da88a/libos/include/libos_handle.h#L180

Ideally, this `handle::uri` should be removed and instead moved to handle-specific struct. Like we do for Encrypted Files here: https://github.com/gramineproject/gramine/blob/aaf34c87bfebd071cdb18a187a7b7cbe510da88a/libos/include/libos_fs_encrypted.h#L46

Many parts of the FS code do not set the URI. For example, `tmpfs` handles do not set URI.

I agree with @boryspoplawski that this line is wrong: https://github.com/gramineproject/gramine/blob/d8217c8ad08fc2be289384b90100e29c0506c097/libos/src/libos_rtld.c#L1037

It should have an IF clause similar to this case: https://github.com/gramineproject/gramine/blob/d8217c8ad08fc2be289384b90100e29c0506c097/libos/src/libos_rtld.c#L839-L841

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/841#-N9aV1a5AKnECh52y7Ar:-N9aV1a5AKnECh52y7As:bvbd1gm) (no related file):*
@kailun-qin I think this solution is wrong. We need to make Gramine LibOS codebase resistant to cases of `handle::uri == NULL`. Maybe you could look into this? See also my other comment.

But to fix #840 quickly, we can just add an `if (hdl->uri) ...only then call append_r_debug()...`

___
___
*[`libos/src/fs/chroot/encrypted.c` line 179 at r1](https://reviewable.io/reviews/gramineproject/gramine/841#-N9aUvTN1p1xUeUdO1-u:-N9aUvTN1p1xUeUdO1-v:b-s52rym) ([raw file](https://github.com/gramineproject/gramine/blob/89521c7dcfdb9760ecc624e19d620d02c093c41e/libos/src/fs/chroot/encrypted.c#L179)):*
> ```C
> 
>     int ret;
>     char* uri = NULL;
> ```

I have a strong feeling that this code is unnecessary and wrong.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/841)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @kailun-qin, and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/841#-N9Wuazn252_ENIQVHHC:-N9auxhuC-97Unb40Yew:bmduywv) (no related file):*

The LibOS codebase doesn't require `handle::uri` to be set: https://github.com/gramineproject/gramine/blob/aaf34c87bfebd071cdb18a187a7b7cbe510da88a/libos/include/libos_handle.h#L180

Ideally, this `handle::uri` should be removed and instead moved to handle-specific struct. Like we do for Encrypted Files here: https://github.com/gramineproject/gramine/blob/aaf34c87bfebd071cdb18a187a7b7cbe510da88a/libos/include/libos_fs_encrypted.h#L46

Many parts of the FS code do not set the URI. For example, `tmpfs` handles do not set URI.

I agree with @boryspoplawski that this line is wrong: https://github.com/gramineproject/gramine/blob/d8217c8ad08fc2be289384b90100e29c0506c097/libos/src/libos_rtld.c#L1037

It should have an IF clause similar to this case: https://github.com/gramineproject/gramine/blob/d8217c8ad08fc2be289384b90100e29c0506c097/libos/src/libos_rtld.c#L839-L841
</blockquote></details>

I see, make sense. Thanks!

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/841#-N9aV1a5AKnECh52y7Ar:-N9av3mQ7ytyqJmMXWhD:b-612znw) (no related file):*

@kailun-qin I think this solution is wrong. We need to make Gramine LibOS codebase resistant to cases of `handle::uri == NULL`. Maybe you could look into this? See also my other comment.

But to fix #840 quickly, we can just add an `if (hdl->uri) ...only then call append_r_debug()...`
</blockquote></details>

Yes, I can look more thoroughly into this. Did a quick fix at https://github.com/gramineproject/gramine/pull/845 for https://github.com/gramineproject/gramine/issues/840.

___

boryspoplawski:
Closing this PR as we decided not to proceed with it

PR #843, opened 2022-08-16
[CI-Examples]Modify ra_tls_mbedtls' Makefile to enable it in Anolis OS

billionairiam:
Signed-off-by: Liang, Ma <liang3.ma@intel.com>

## Description of the changes 

1. When we configure gramine in [Anolis OS](https://openanolis.cn/), The[ ra_tls_mbedtls](https://github.com/gramineproject/gramine/tree/master/CI-Examples/ra-tls-mbedtls) under CI-Example can't work correctly. This is caused by the [Makefile](https://github.com/gramineproject/gramine/blob/master/CI-Examples/ra-tls-mbedtls/Makefile) is debian/ubuntu specific, so I think we can modify it to fit different operating systems.
2. In ubuntu the `ARCH_LIBDIR ?= /lib/$(shell $(CC) -dumpmachine)`, while in Anolis OS it should be `/usr/lib64`

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/843)

### DIFF
--- a/CI-Examples/ra-tls-mbedtls/Makefile
+++ b/CI-Examples/ra-tls-mbedtls/Makefile
@@ -1,7 +1,16 @@
 GRAMINEDIR ?= ../..
-GRAMINE_PKGLIBDIR ?= /usr/lib/x86_64-linux-gnu/gramine # this is debian/ubuntu specific
-
+OS_TYPE ?= $(shell cat /etc/os-release | grep -w NAME)
+ifeq ($(findstring Ubuntu, $(OS_TYPE)),Ubuntu)
+GRAMINE_PKGLIBDIR ?= /usr/lib/x86_64-linux-gnu/gramine
 ARCH_LIBDIR ?= /lib/$(shell $(CC) -dumpmachine)
+else
+ifeq ($(findstring Anolis OS,$(OS_TYPE)),Anolis OS)
+GRAMINE_PKGLIBDIR ?= /usr/lib64/gramine
+ARCH_LIBDIR ?= /lib64
+else
+echo "Unsupported operation system!"
+endif
+endif
 
 ifeq ($(DEBUG),1)
 GRAMINE_LOG_LEVEL = debug


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/843)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @billionairiam)

___
*[`CI-Examples/ra-tls-mbedtls/Makefile` line 13 at r1](https://reviewable.io/reviews/gramineproject/gramine/843#-N9aXrfk2T5Y0xcSjEkM:-N9aXrfl6vRVXPdnAcz0:bbc9dzh) ([raw file](https://github.com/gramineproject/gramine/blob/ce974a4b0b11249493bf0b4f2ffaf6d9348b87ef/CI-Examples/ra-tls-mbedtls/Makefile#L13)):*
> ```Makefile
> echo "Unsupported operation system!"
> endif
> endif
> ```

I veto this change. We will actually need to change **all** our Makefiles, not just this one.

We need to come up with a better way of specifying the paths based on the OS distro. Unfortunately, I do not know of better ways...

I think we had this conversation before with @mkow and @woju, and didn't come up with a better way.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/843)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @mkow and @woju)

___
*[`CI-Examples/ra-tls-mbedtls/Makefile` line 13 at r1](https://reviewable.io/reviews/gramineproject/gramine/843#-N9aXrfk2T5Y0xcSjEkM:-N9bV9OR7omhsJ9gqngk:b-2kimvl) ([raw file](https://github.com/gramineproject/gramine/blob/ce974a4b0b11249493bf0b4f2ffaf6d9348b87ef/CI-Examples/ra-tls-mbedtls/Makefile#L13)):*

I veto this change. We will actually need to change **all** our Makefiles, not just this one.

We need to come up with a better way of specifying the paths based on the OS distro. Unfortunately, I do not know of better ways...

I think we had this conversation before with @mkow and @woju, and didn't come up with a better way.
</blockquote></details>

These are only examples, what's the problem with them not working on some OSes? For a production env you need to pass all these files (that are in ARCH_DIR) together with the app anyway (easiest way being docker with Gramine or GSC)

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/843)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @mkow and @woju)

___
*[`CI-Examples/ra-tls-mbedtls/Makefile` line 13 at r1](https://reviewable.io/reviews/gramineproject/gramine/843#-N9aXrfk2T5Y0xcSjEkM:-N9fv97T6ZsjvjnW5u5g:br8jrfs) ([raw file](https://github.com/gramineproject/gramine/blob/ce974a4b0b11249493bf0b4f2ffaf6d9348b87ef/CI-Examples/ra-tls-mbedtls/Makefile#L13)):*

These are only examples, what's the problem with them not working on some OSes? For a production env you need to pass all these files (that are in ARCH_DIR) together with the app anyway (easiest way being docker with Gramine or GSC)
</blockquote></details>

Sure, but then why do we modify only this one example?

And if we modify *all* examples like this, then we introduced 10 identical lines repeated X times... I dislike it very much.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/843)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @mkow and @woju)

___
*[`CI-Examples/ra-tls-mbedtls/Makefile` line 13 at r1](https://reviewable.io/reviews/gramineproject/gramine/843#-N9aXrfk2T5Y0xcSjEkM:-N9g0IEs5isslevEuA96:bdm0ja3) ([raw file](https://github.com/gramineproject/gramine/blob/ce974a4b0b11249493bf0b4f2ffaf6d9348b87ef/CI-Examples/ra-tls-mbedtls/Makefile#L13)):*

Sure, but then why do we modify only this one example?

And if we modify *all* examples like this, then we introduced 10 identical lines repeated X times... I dislike it very much.
</blockquote></details>

No, my point was to modify nothing - these examples are only usable on ubuntu/debian. Other users are on their own.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/843)* status: 0 of 1 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @billionairiam, @mkow, and @woju)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/843#-N9kzZWY2kiCSLG9e5a_:-N9kzZWY2kiCSLG9e5aa:bxeb4hs) (no related file):*
@billionairiam What was the reason for this PR? Do you think it would be possible for Anolis OS users to run the build like this:
```
SGX=1 GRAMINE_PKGLIBDIR=/usr/lib64/gramine ARCH_LIBDIR=/lib64 make
```

(This is what users of CentOS/RHEL/Fedora do currently anyway. Anolis users could do the same.)

As mentioned in the other comment, we don't think we won't such changes in our examples. They are purely for Debian/Ubuntu, and we don't want to complicate the examples.

___
___
*[`CI-Examples/ra-tls-mbedtls/Makefile` line 13 at r1](https://reviewable.io/reviews/gramineproject/gramine/843#-N9aXrfk2T5Y0xcSjEkM:-N9kzXia0qFUIF9eQvA7:b-pbdqz3) ([raw file](https://github.com/gramineproject/gramine/blob/ce974a4b0b11249493bf0b4f2ffaf6d9348b87ef/CI-Examples/ra-tls-mbedtls/Makefile#L13)):*

No, my point was to modify nothing - these examples are only usable on ubuntu/debian. Other users are on their own.
</blockquote></details>

Agreed.

PR #845, opened 2022-08-16
[LibOS] Append `handle::uri` to GDB link map only when it's set

### LINKED ISSUE #840: Unexpected memory fault occurred inside PAL (strnlen at strlen.c)

kailun-qin:
LibOS codebase doesn't require `handle::uri` to be set. Appending NULL URI directly to GDB link map will lead to an unexpected memory fault occurred inside PAL.

This patch appends `handle::uri` to GDB link map only when it's set.

Fixes #840.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/845)

### DIFF
--- a/libos/src/libos_rtld.c
+++ b/libos/src/libos_rtld.c
@@ -1034,7 +1034,8 @@ int register_library(const char* name, unsigned long load_address) {
         return err;
     }
 
-    append_r_debug(hdl->uri, (void*)load_address);
+    if (hdl->uri)
+        append_r_debug(hdl->uri, (void*)load_address);
     put_handle(hdl);
     return 0;
 }


boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/845)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin)

___
*[`libos/src/libos_rtld.c` line 1037 at r1](https://reviewable.io/reviews/gramineproject/gramine/845#-N9baDL27JZGOTEnGXsy:-N9baDL27JZGOTEnGXsz:b-t1uk0n) ([raw file](https://github.com/gramineproject/gramine/blob/17db68263b5fb1f9a36ed3afe8627f1e42a48060/libos/src/libos_rtld.c#L1037)):*
> ```C
>     }
> 
>     if (hdl->uri)
> ```

But with this PF won't be reflected in debug maps and we won't get symbols e.g. when using GDB...

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/845)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin)

___
*[`libos/src/libos_rtld.c` line 1037 at r1](https://reviewable.io/reviews/gramineproject/gramine/845#-N9baDL27JZGOTEnGXsy:-N9bbPDY6xkpJ4HJX5cX:bkjbu43) ([raw file](https://github.com/gramineproject/gramine/blob/17db68263b5fb1f9a36ed3afe8627f1e42a48060/libos/src/libos_rtld.c#L1037)):*

But with this PF won't be reflected in debug maps and we won't get symbols e.g. when using GDB...
</blockquote></details>

But PFs are encrypted anyway, so GDB won't be able to read it?

I don't know how much info GDB takes from the loaded image (`load_address`) and how much info GDB takes from the file on disk (`hdl->uri`). But I would assume that giving a filepath with encrypted content breaks GDB's ability to reflect this file correctly.

So I don't think we lose anything here.

Actually, would be interesting to check this, by marking some lib as a Protected File and running under GDB. @kailun-qin Could you do this experiment?

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/845)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin)

___
*[`libos/src/libos_rtld.c` line 1037 at r1](https://reviewable.io/reviews/gramineproject/gramine/845#-N9baDL27JZGOTEnGXsy:-N9be3j349P6Ha9XO6nf:b-jxdpik) ([raw file](https://github.com/gramineproject/gramine/blob/17db68263b5fb1f9a36ed3afe8627f1e42a48060/libos/src/libos_rtld.c#L1037)):*

But PFs are encrypted anyway, so GDB won't be able to read it?

I don't know how much info GDB takes from the loaded image (`load_address`) and how much info GDB takes from the file on disk (`hdl->uri`). But I would assume that giving a filepath with encrypted content breaks GDB's ability to reflect this file correctly.

So I don't think we lose anything here.

Actually, would be interesting to check this, by marking some lib as a Protected File and running under GDB. @kailun-qin Could you do this experiment?
</blockquote></details>

+1 please check this.

But even if this won't work, @dimakuv I think you said we want to migrate from `handle::uri` to other handle specific fields. This change would only hide problems.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/845)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin)

___
*[`libos/src/libos_rtld.c` line 1037 at r1](https://reviewable.io/reviews/gramineproject/gramine/845#-N9baDL27JZGOTEnGXsy:-N9bfI3G87ulXzb-1r_H:b-k3ob6t) ([raw file](https://github.com/gramineproject/gramine/blob/17db68263b5fb1f9a36ed3afe8627f1e42a48060/libos/src/libos_rtld.c#L1037)):*

+1 please check this.

But even if this won't work, @dimakuv I think you said we want to migrate from `handle::uri` to other handle specific fields. This change would only hide problems.
</blockquote></details>

@boryspoplawski Yes, this PR is not the final solution, more like a bandaid. But we don't have any timeline when (and if) we'll remove `handle::uri`, so having this PR that "hides the problem" now is better than segfaulting.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/845)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin)

___
*[`libos/src/libos_rtld.c` line 1037 at r1](https://reviewable.io/reviews/gramineproject/gramine/845#-N9baDL27JZGOTEnGXsy:-N9bmGjw5cRG0GqMvve2:b-x8dhoo) ([raw file](https://github.com/gramineproject/gramine/blob/17db68263b5fb1f9a36ed3afe8627f1e42a48060/libos/src/libos_rtld.c#L1037)):*

But can't we have a proper solution now while we still have `handle::uri`? By proper I mean working with what we have right now (including every kind of file)
</blockquote></details>

What do you mean by a proper solution? Do you mean to check for the type of the handle (is it a file? is it an encrypted file?) and then choose the URI based on the handle type? And then do `append_r_debug(hdl_type == encrypted ? hdl->encrypted->uri : hdl->uri, ...)`? Yes, this is also a good idea. I'd prefer such a solution.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/845)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski and @dimakuv)

___
*[`libos/src/libos_rtld.c` line 1037 at r1](https://reviewable.io/reviews/gramineproject/gramine/845#-N9baDL27JZGOTEnGXsy:-N9dbdTa-fT5c0Vm81k2:b-fv9zzs) ([raw file](https://github.com/gramineproject/gramine/blob/17db68263b5fb1f9a36ed3afe8627f1e42a48060/libos/src/libos_rtld.c#L1037)):*
> Actually, would be interesting to check this, by marking some lib as a Protected File and running under GDB.

Checked marking some lib as a PF under GDB (w/ the filepath given), the behavior is basically expected - `giving a filepath with encrypted content breaks GDB's ability to reflect this file correctly`. Pls kindly see below.
```
[P1:T1:helloworld] debug: append_r_debug: adding file:./xxx/libxxx.so at 0x3338000
Python Exception <class 'elftools.common.exceptions.ELFError'> Magic number does not match:
Python Exception <class 'gdb.error'> Error occurred in Python: Magic number does not match:

Thread 1 "loader" hit Breakpoint -1, debug_map_update_debugger () at ../pal/src/host/linux-common/debug_map.c:37
37      __attribute__((__noinline__)) void debug_map_update_debugger(void) {
(gdb) c
Continuing.
...
(gdb) n
Single stepping until exit from function add@plt,
which has no line number information.
0x00000000033390f9 in ?? ()
```

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/845)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @pwmarcz)

___
*[`libos/src/libos_rtld.c` line 1037 at r1](https://reviewable.io/reviews/gramineproject/gramine/845#-N9baDL27JZGOTEnGXsy:-N9flGYsEs54YcihIwTP:b4iz7pa) ([raw file](https://github.com/gramineproject/gramine/blob/17db68263b5fb1f9a36ed3afe8627f1e42a48060/libos/src/libos_rtld.c#L1037)):*

> Actually, would be interesting to check this, by marking some lib as a Protected File and running under GDB.

Checked marking some lib as a PF under GDB (w/ the filepath given), the behavior is basically expected - `giving a filepath with encrypted content breaks GDB's ability to reflect this file correctly`. Pls kindly see below.
```
[P1:T1:helloworld] debug: append_r_debug: adding file:./xxx/libxxx.so at 0x3338000
Python Exception <class 'elftools.common.exceptions.ELFError'> Magic number does not match:
Python Exception <class 'gdb.error'> Error occurred in Python: Magic number does not match:

Thread 1 "loader" hit Breakpoint -1, debug_map_update_debugger () at ../pal/src/host/linux-common/debug_map.c:37
37      __attribute__((__noinline__)) void debug_map_update_debugger(void) {
(gdb) c
Continuing.
...
(gdb) n
Single stepping until exit from function add@plt,
which has no line number information.
0x00000000033390f9 in ?? ()
```
</blockquote></details>

@pwmarcz Could we summon you to shed light on this issue?

We have `handle::uri`: https://github.com/gramineproject/gramine/blob/9af4ad88b41e281b603c5dc7600a5a20bfdfffad/libos/include/libos_handle.h#L180

But we also have `libos_encrypted_file::uri` (which is stashed into the handle via `handle::dentry::inode::data`, see e.g. https://github.com/gramineproject/gramine/blob/74e74a87a1127124cf89bd487a021dc1ceb8fa75/libos/src/fs/chroot/encrypted.c#L159?plain=1):
https://github.com/gramineproject/gramine/blob/9af4ad88b41e281b603c5dc7600a5a20bfdfffad/libos/include/libos_fs_encrypted.h#L46

One thing to note is that `handle::uri` is not supposed to change and thus cannot be used for file handles (as they can be renamed). But `libos_encrypted_file::uri` can be changed, e.g. https://github.com/gramineproject/gramine/blob/74e74a87a1127124cf89bd487a021dc1ceb8fa75/libos/src/fs/libos_fs_encrypted.c#L661

So what is the relationship between the two `uri` fields?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/845)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski and @dimakuv)

___
*[`libos/src/libos_rtld.c` line 1037 at r1](https://reviewable.io/reviews/gramineproject/gramine/845#-N9baDL27JZGOTEnGXsy:-N9gRQNz6AW3ljJytfiS:b-yfeamz) ([raw file](https://github.com/gramineproject/gramine/blob/17db68263b5fb1f9a36ed3afe8627f1e42a48060/libos/src/libos_rtld.c#L1037)):*

@pwmarcz Could we summon you to shed light on this issue?

We have `handle::uri`: https://github.com/gramineproject/gramine/blob/9af4ad88b41e281b603c5dc7600a5a20bfdfffad/libos/include/libos_handle.h#L180

But we also have `libos_encrypted_file::uri` (which is stashed into the handle via `handle::dentry::inode::data`, see e.g. https://github.com/gramineproject/gramine/blob/74e74a87a1127124cf89bd487a021dc1ceb8fa75/libos/src/fs/chroot/encrypted.c#L159?plain=1):
https://github.com/gramineproject/gramine/blob/9af4ad88b41e281b603c5dc7600a5a20bfdfffad/libos/include/libos_fs_encrypted.h#L46

One thing to note is that `handle::uri` is not supposed to change and thus cannot be used for file handles (as they can be renamed). But `libos_encrypted_file::uri` can be changed, e.g. https://github.com/gramineproject/gramine/blob/74e74a87a1127124cf89bd487a021dc1ceb8fa75/libos/src/fs/libos_fs_encrypted.c#L661

So what is the relationship between the two `uri` fields?
</blockquote></details>

Yeah, our current setup for GDB assumes that we can read the file from host. Perhaps we could make GDB look at memory instead, but that sounds difficult: you will need to ensure that all relevant parts of the binary are mapped in memory.

@dimakuv Two things:
* `enc->uri` is for the encrypted files library, which I wanted to keep separate from the rest of LibOS. It has no notion of LibOS handle, but needs to have this information in order to call PAL and open a file.
*  If I recall collectly, `enc` doesn't actually correspond to a handle, but to an inode. Multiple handles use the same `enc` so that an encrypted file is opened only once (and you can read/write it through multiple handles without problems with consistency).

As for `handle::uri` being immutable, yeah... this is convenient because you don't need a lock, but when you rename a file, you're not able to update that URI. That's actually not a problem for non-encrypted files, because you don't actually need to open them again, so it's OK to keep the old URI (for debugging purposes I guess).

Actually I think `handle::uri` is actually not very useful except for things like debug maps or reporting errors? Usually you need the URI when opening a file, not afterwards. This is a bit similar to `handle::dentry` - in principle, it's not necessary, in practice it might be hard to get rid of from Gramine.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/845)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski)

___
*[`libos/src/libos_rtld.c` line 1037 at r1](https://reviewable.io/reviews/gramineproject/gramine/845#-N9baDL27JZGOTEnGXsy:-N9l6F2C4Yu2W3gVTpMW:b9mib1q) ([raw file](https://github.com/gramineproject/gramine/blob/17db68263b5fb1f9a36ed3afe8627f1e42a48060/libos/src/libos_rtld.c#L1037)):*

Yeah, our current setup for GDB assumes that we can read the file from host. Perhaps we could make GDB look at memory instead, but that sounds difficult: you will need to ensure that all relevant parts of the binary are mapped in memory.

@dimakuv Two things:
* `enc->uri` is for the encrypted files library, which I wanted to keep separate from the rest of LibOS. It has no notion of LibOS handle, but needs to have this information in order to call PAL and open a file.
*  If I recall collectly, `enc` doesn't actually correspond to a handle, but to an inode. Multiple handles use the same `enc` so that an encrypted file is opened only once (and you can read/write it through multiple handles without problems with consistency).

As for `handle::uri` being immutable, yeah... this is convenient because you don't need a lock, but when you rename a file, you're not able to update that URI. That's actually not a problem for non-encrypted files, because you don't actually need to open them again, so it's OK to keep the old URI (for debugging purposes I guess).

Actually I think `handle::uri` is actually not very useful except for things like debug maps or reporting errors? Usually you need the URI when opening a file, not afterwards. This is a bit similar to `handle::dentry` - in principle, it's not necessary, in practice it might be hard to get rid of from Gramine.
</blockquote></details>

Thanks!

So we have the following set:
- `enc->uri` cannot be removed from the codebase
- `handle->uri` could be removed from the codebase
- `handle->uri` seems generally useful as @boryspoplawski mentioned (it's a URI of an underlying PAL object, if any) -- used by normal files, named pipes, and maybe some other types of handles

So to me, sounds like it doesn't make much sense to remove the redundancy between `enc->uri` and `handle->uri` (this would require removing `handle->uri` but it seems sufficiently developer-friendly and useful).

Then maybe the current PR is fine? @boryspoplawski @kailun-qin

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/845)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv)

___
*[`libos/src/libos_rtld.c` line 1037 at r1](https://reviewable.io/reviews/gramineproject/gramine/845#-N9baDL27JZGOTEnGXsy:-N9lHu93ElMeWB1TjhOD:bitdazo) ([raw file](https://github.com/gramineproject/gramine/blob/17db68263b5fb1f9a36ed3afe8627f1e42a48060/libos/src/libos_rtld.c#L1037)):*

Thanks!

So we have the following set:
- `enc->uri` cannot be removed from the codebase
- `handle->uri` could be removed from the codebase
- `handle->uri` seems generally useful as @boryspoplawski mentioned (it's a URI of an underlying PAL object, if any) -- used by normal files, named pipes, and maybe some other types of handles

So to me, sounds like it doesn't make much sense to remove the redundancy between `enc->uri` and `handle->uri` (this would require removing `handle->uri` but it seems sufficiently developer-friendly and useful).

Then maybe the current PR is fine? @boryspoplawski @kailun-qin 
</blockquote></details>

I'm not sure if `handle->uri` is that useful, as @pwmarcz pointed out, outside of debugging. But again, I have no good overview of FS subsystem, so it's hard for me to judge this.
As for PF inside GDB - it would be nice to have such support, someday it will come useful.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/845)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski and @dimakuv)

___
*[`libos/src/libos_rtld.c` line 1037 at r1](https://reviewable.io/reviews/gramineproject/gramine/845#-N9baDL27JZGOTEnGXsy:-N9lL4mV-i-5rGkQ0LXN:b-ab7dxq) ([raw file](https://github.com/gramineproject/gramine/blob/17db68263b5fb1f9a36ed3afe8627f1e42a48060/libos/src/libos_rtld.c#L1037)):*

I'm not sure if `handle->uri` is that useful, as @pwmarcz pointed out, outside of debugging. But again, I have no good overview of FS subsystem, so it's hard for me to judge this.
As for PF inside GDB - it would be nice to have such support, someday it will come useful.
</blockquote></details>

It seems reasonable to me to keep the `handle->uri` at least for now (as it's still helping w/ debug stuff to some extent). I assume it should be thru another PR if we do want to remove the redundancy?

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/845)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski)

___
*[`libos/src/libos_rtld.c` line 1037 at r1](https://reviewable.io/reviews/gramineproject/gramine/845#-N9baDL27JZGOTEnGXsy:-N9ovd57ALMaZxcR4xfr:b-t50xm5) ([raw file](https://github.com/gramineproject/gramine/blob/17db68263b5fb1f9a36ed3afe8627f1e42a48060/libos/src/libos_rtld.c#L1037)):*
I'm fine with this PR and accepting it from my side.

> As for PF inside GDB - it would be nice to have such support, someday it will come useful.

Yes, but that's a whole different story...

> I assume it should be thru another PR if we do want to remove the redundancy?

Yes, someone will need to carefully analyze the usages of `handle->uri` and figure out how to remove it (or make a definitive statement why it shouldn't be removed).

boryspoplawski (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/845)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required) (waiting on @kailun-qin)

___
*[`-- commits` line 6 at r1](https://reviewable.io/reviews/gramineproject/gramine/845#-N9pbHCB26Hx9mZqzHmd):*




_Suggestion:_
```
occurring
```
___
*[`-- commits` line 8 at r1](https://reviewable.io/reviews/gramineproject/gramine/845#-N9pbNtj9tO6MMbKw4_O):*
This is already stated in one liner, I don't think we need it

boryspoplawski:
Jenkins, test this please

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/845)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required) (waiting on @boryspoplawski)

___
*[`-- commits` line 6 at r1](https://reviewable.io/reviews/gramineproject/gramine/845#-N9pbHCB26Hx9mZqzHmd):*
Pls kindly help fix during rebase, thanks!
___
*[`-- commits` line 8 at r1](https://reviewable.io/reviews/gramineproject/gramine/845#-N9pbNtj9tO6MMbKw4_O):*

This is already stated in one liner, I don't think we need it
</blockquote></details>

Pls kindly help fix during rebase, thanks!

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/845)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski)

___
*[`-- commits` line 6 at r1](https://reviewable.io/reviews/gramineproject/gramine/845#-N9pbHCB26Hx9mZqzHmd):*

Pls kindly help fix during rebase, thanks!
</blockquote></details>

Done
___
*[`-- commits` line 8 at r1](https://reviewable.io/reviews/gramineproject/gramine/845#-N9pbNtj9tO6MMbKw4_O):*

Pls kindly help fix during rebase, thanks!
</blockquote></details>

Done

dimakuv:
Jenkins, test this please

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/845)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #862, opened 2022-08-25
[LibOS] Add support for `MSG_DONTWAIT` in UNIX domain sockets

boryspoplawski:
## Description of the changes 
Emulated purely in LibOS. 

## How to test this PR? 
You can use following program (in libos regression tests, as it uses some common code from there) to play with different races (highlighted a few places you can edit for different results). I did not add it to CI, since it just races some operations and it's hard to define what counts as the test passed.

```c
#include <err.h>
#include <errno.h>
#include <fcntl.h>
#include <stddef.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/ioctl.h>
#include <sys/socket.h>
#include <sys/stat.h>
#include <sys/types.h>
#include <sys/un.h>
#include <sys/wait.h>
#include <unistd.h>
#include <pthread.h>
#include <sys/time.h>

#include "common.h"

static void* func(void* arg) {
    int s = (long)arg;
    /* You can play with this loop size. */
    for (size_t i = 0; i < 100000; ++i) {
        int val = 1;
        CHECK(ioctl(s, FIONBIO, &val));
        val = 0;
        CHECK(ioctl(s, FIONBIO, &val));
    }
    return NULL;
}

int main(void) {
    setbuf(stdout, NULL);
    setbuf(stderr, NULL);

    int p[2];
    CHECK(pipe(p));

    struct sockaddr_un sa = {
        .sun_family = AF_UNIX,
        .sun_path = "u",
    };

    int pid = CHECK(fork());
    if (pid == 0) {
        int x;
        int s = CHECK(socket(AF_UNIX, SOCK_STREAM, 0));

        /* You can comment out the thread spawning. */
        pthread_t th;
        x = pthread_create(&th, NULL, func, (void*)(long)s);
        errno = -x;
        CHECK(-(x != 0));
        (void)func;

        char buf[0x10];
        CHECK(read(p[0], buf, sizeof(buf)));

        CHECK(connect(s, (void*)&sa, sizeof(sa)));

        /* You can play with the loop size. */
        for (size_t i = 0; i < 100; ++i) {
            errno = 0;
            x = recv(s, buf, sizeof(buf), MSG_DONTWAIT);
            if (x != -1) {
                printf("got: i=%zu, ret=%d\n", i, x);
            } else if (errno != EAGAIN) {
                printf("WTF: %d: %d\n", x, errno);
                exit(1);
            }
        }

        errno = 0;
        struct timeval t1 = { 0 };
        struct timeval t2 = { 0 };
        CHECK(gettimeofday(&t1, NULL));
        x = recv(s, buf, sizeof(buf), 0);
        CHECK(gettimeofday(&t2, NULL));
        long ms = (t2.tv_sec - t1.tv_sec) * 1000 + (t2.tv_usec - t1.tv_usec) / 1000;
        printf("recv returned: %d: %d - %m in %ld (ms)\n", x, errno, ms);

        return 0;
    }

    int s = CHECK(socket(AF_UNIX, SOCK_STREAM, 0));

    CHECK(bind(s, (void*)&sa, sizeof(sa)));
    CHECK(listen(s, 5));

    char buf[0x10] = { 0 };
    CHECK(write(p[1], buf, 1));

    int c = CHECK(accept(s, NULL, NULL));
    printf("client: %d\n", c);

    sleep(1);
    CHECK(write(c, buf, 1));

    CHECK(wait(NULL));
    return 0;
}
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/862)

### DIFF
--- a/libos/include/libos_handle.h
+++ b/libos/include/libos_handle.h
@@ -81,6 +81,8 @@ enum libos_sock_state {
  * `ops`, `domain`, `type` and `protocol` are read-only and do not need any locking.
  * Access to `peek` struct is protected by `recv_lock`. This lock also ensures proper ordering of
  * stream reads (see the comment in `do_recvmsg` in "libos/src/sys/libos_socket.c").
+ * Access to `force_nonblocking_users_count` is protected by the lock of the handle wrapping this
+ * struct.
  * `pal_handle` should be accessed using atomic operations.
  * If you need to take both `recv_lock` and `lock`, take the former first.
  */
@@ -105,9 +107,11 @@ struct libos_sock_handle {
         size_t data_size;
     } peek;
     struct libos_lock recv_lock;
-    unsigned int last_error;
+    /* This field is only used by UNIX sockets. */
+    size_t force_nonblocking_users_count;
     uint64_t sendtimeout_us;
     uint64_t receivetimeout_us;
+    unsigned int last_error;
     /* This field denotes whether the socket was ever bound. */
     bool was_bound;
     /* This field indicates if the socket is ready for read-like operations (`recv`/`read` or
--- a/libos/src/fs/socket/fs.c
+++ b/libos/src/fs/socket/fs.c
@@ -81,6 +81,8 @@ static int setflags(struct libos_handle* handle, unsigned int flags, unsigned in
     assert(mask != 0);
     assert((flags & ~mask) == 0);
 
+    /* If you add support for other flags, check "libos/src/net/unix.c", which also (temporarily)
+     * changes these flags. */
     if (!WITHIN_MASK(flags, O_NONBLOCK)) {
         return -EINVAL;
     }
@@ -95,6 +97,12 @@ static int setflags(struct libos_handle* handle, unsigned int flags, unsigned in
     PAL_HANDLE pal_handle = __atomic_load_n(&sock->pal_handle, __ATOMIC_ACQUIRE);
     if (!pal_handle) {
         /* Just save the flags for later. */
+        goto out_set_flags;
+    }
+
+    if (handle->info.sock.force_nonblocking_users_count) {
+        /* Some thread is forcing a nonblocking operation, it will set the correct flags in PAL, we
+         * just need to set flags in LibOS. */
         goto out_set_flags;
     }
 
--- a/libos/src/net/unix.c
+++ b/libos/src/net/unix.c
@@ -309,6 +309,100 @@ static int getsockopt(struct libos_handle* handle, int level, int optname, void*
     return -ENOPROTOOPT;
 }
 
+static int maybe_force_nonblocking_wrapper(bool force_nonblocking, struct libos_handle* handle,
+                                           PAL_HANDLE pal_handle,
+                                           int (*func)(PAL_HANDLE, uint64_t, size_t*, void*),
+                                           void* buf, size_t* size) {
+    /*
+     * There are 3 kinds of operations that can race here:
+     * 1) operation with force_nonblocking set,
+     * 2) operation with force_nonblocking not set,
+     * 3) changing blockingness of the handle,
+     * and any combination of any of them should yield correct results.
+     *
+     * 1-1, 2-2, 3-3 races are "normal" and proper locking orders them correctly.
+     * 1-2 races - operation 2 does not affect 1; operation 2 will be repeated if handle is
+     *             temporarily nonblocking (concurrent operation 1 happening).
+     * 1-3 races - we count threads using the handle in temporary nonblocking mode. The first thread
+     *             to use it does the change on PAL level and the last thread restores the correct
+     *             (currently set in LibOS) mode. Operation 3 takes that into account - sets flags
+     *             only in LibOS if `handle->info.sock.force_nonblocking_users_count` is nonzero
+     *             (see "libos/src/fs/socket/fs.c"). `handle->lock` provides proper ordering.
+     * 2-3 races - this is inherently racy - operation 2 will be either in the old or new mode,
+     *             depending on how it ends up ordered vs operation 3. This works exactly the same
+     *             on normal Linux - doing operations 2 and 3 concurrently, user app cannot know
+     *             whether operation 2 ends up being blocking or not.
+     */
+    int ret;
+    if (force_nonblocking) {
+        /* We already have `pal_handle` set, so there is no need for taking `sock->lock`. */
+        lock(&handle->lock);
+        handle->info.sock.force_nonblocking_users_count += 1;
+        if (!(handle->flags & O_NONBLOCK) && handle->info.sock.force_nonblocking_users_count == 1) {
+            /* Temporarily set `pal_handle` in nonblocking mode. */
+            PAL_STREAM_ATTR attr;
+            ret = PalStreamAttributesQueryByHandle(pal_handle, &attr);
+            if (ret < 0) {
+                unlock(&handle->lock);
+                return pal_to_unix_errno(ret);
+            }
+            assert(!attr.nonblocking);
+            attr.nonblocking = true;
+            ret = PalStreamAttributesSetByHandle(pal_handle, &attr);
+            if (ret < 0) {
+                unlock(&handle->lock);
+                return pal_to_unix_errno(ret);
+            }
+        }
+        unlock(&handle->lock);
+    }
+
+again:
+    ret = func(pal_handle, /*offset=*/0, size, buf);
+    if (ret < 0) {
+        ret = (ret == -PAL_ERROR_TOOLONG) ? -EMSGSIZE : pal_to_unix_errno(ret);
+        if (ret == -EAGAIN && !force_nonblocking) {
+            lock(&handle->lock);
+            bool handle_is_blocking = !(handle->flags & O_NONBLOCK);
+            unlock(&handle->lock);
+            if (handle_is_blocking) {
+                /* Spurious `EAGAIN`, retry. */
+                goto again;
+            }
+        }
+    }
+
+    if (force_nonblocking) {
+        lock(&handle->lock);
+        handle->info.sock.force_nonblocking_users_count -= 1;
+        if (!(handle->flags & O_NONBLOCK) && handle->info.sock.force_nonblocking_users_count == 0) {
+            /* `pal_handle` was temporarily in nonblocking mode, fix it. */
+            PAL_STREAM_ATTR attr;
+            int tmp_ret = PalStreamAttributesQueryByHandle(pal_handle, &attr);
+            if (tmp_ret < 0) {
+                unlock(&handle->lock);
+                tmp_ret = pal_to_unix_errno(tmp_ret);
+                log_error("%s: nonblocking restore: failed to get handle attrs: %d", __func__,
+                          tmp_ret);
+                PalProcessExit(1);
+            }
+            assert(attr.nonblocking);
+            attr.nonblocking = false;
+            tmp_ret = PalStreamAttributesSetByHandle(pal_handle, &attr);
+            if (tmp_ret < 0) {
+                unlock(&handle->lock);
+                tmp_ret = pal_to_unix_errno(tmp_ret);
+                log_error("%s: nonblocking restore: failed to set handle attrs: %d", __func__,
+                          tmp_ret);
+                PalProcessExit(1);
+            }
+        }
+        unlock(&handle->lock);
+    }
+
+    return ret;
+}
+
 static int send(struct libos_handle* handle, struct iovec* iov, size_t iov_len, size_t* out_size,
                 void* addr, size_t addrlen, bool force_nonblocking) {
     __UNUSED(addr);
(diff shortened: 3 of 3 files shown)


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/862)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski)

___
*[`libos/include/libos_handle.h` line 90 at r1](https://reviewable.io/reviews/gramineproject/gramine/862#-NAOPpzP2m_GJ8dG2Vuy:-NAOPpzP2m_GJ8dG2Vuz:b-pqr5af) ([raw file](https://github.com/gramineproject/gramine/blob/2924c419620e637d5eb2b01a43a109cdf27c6b69/libos/include/libos_handle.h#L90)):*
> ```objc
>  *
>  * TODO: add a `void* private` and move type specific fields to there (e.g. `broadcast` or
>  * `force_nonblocking_users`).
> ```

Well, if it's only 2-3 fields, I wouldn't bother.
___
*[`libos/include/libos_handle.h` line 113 at r1](https://reviewable.io/reviews/gramineproject/gramine/862#-NAOPzQW3EnXQOwZm3IP:-NAOPzQW3EnXQOwZm3IQ:b-2ukvj6) ([raw file](https://github.com/gramineproject/gramine/blob/2924c419620e637d5eb2b01a43a109cdf27c6b69/libos/include/libos_handle.h#L113)):*
> ```objc
>     } peek;
>     struct libos_lock recv_lock;
>     /* Only used by UNIX sockets. */
> ```

Maybe remove this comment? You have a top-level comment on this, and this current comment is ambigious -- does it refer only to the next line, or to all next lines until the next comment?
___
*[`libos/include/libos_handle.h` line 114 at r1](https://reviewable.io/reviews/gramineproject/gramine/862#-NAOSVf76p0kA9wGjCfd:-NAOSVf76p0kA9wGjCfe:b-djtidc) ([raw file](https://github.com/gramineproject/gramine/blob/2924c419620e637d5eb2b01a43a109cdf27c6b69/libos/include/libos_handle.h#L114)):*
> ```objc
>     struct libos_lock recv_lock;
>     /* Only used by UNIX sockets. */
>     size_t force_nonblocking_users;
> ```

Maybe append `_cnt`?
___
*[`libos/src/net/unix.c` line 324 at r1](https://reviewable.io/reviews/gramineproject/gramine/862#-NAORTQv9KIlQAfyAlWb:-NAORTQw41-0OuGQPqPS:b8rr5aj) ([raw file](https://github.com/gramineproject/gramine/blob/2924c419620e637d5eb2b01a43a109cdf27c6b69/libos/src/net/unix.c#L324)):*
> ```C
>      *
>      * 1-1, 2-2, 3-3 races are "normal" and proper locking orders them correctly.
>      * 1-2 races - operation 2 does not affect 1; operation 2 will be repated if handle is blocking.
> ```

`repeated`
___
*[`libos/src/net/unix.c` line 324 at r1](https://reviewable.io/reviews/gramineproject/gramine/862#-NAORUJe9yzivcldAQJ0:-NAORUJe9yzivcldAQJ1:b-sjrhu1) ([raw file](https://github.com/gramineproject/gramine/blob/2924c419620e637d5eb2b01a43a109cdf27c6b69/libos/src/net/unix.c#L324)):*
> ```C
>      *
>      * 1-1, 2-2, 3-3 races are "normal" and proper locking orders them correctly.
>      * 1-2 races - operation 2 does not affect 1; operation 2 will be repated if handle is blocking.
> ```

I don't understand the phrase `operation 2 will be repeated if handle is blocking`. Could you expand?

I understand that we have two scenarios:
1. Underlying non-blocking handle. Then `force_nonblocking` is irrelevant, all operations are non-blocking. So doesn't matter in which order we execute operation 1 and operation 2.
2. Underlying blocking handle. Operation 1 temporarily makes the handle non-blocking and performs the read/write, no questions here. But operation 2 will somehow notice that the handle became non-blocking, and defer its execution until it finds out that the handle is again blocking? (I didn't look at the code yet.) But how does this "noticing and defering" works exactly?

UPDATE: Actually, the code simply checks for `-EAGAIN` error which signals that the handle is currently in non-blocking mode, and retries until the `-EAGAIN` goes away. The code is easier to understand than this comment :) Maybe just add smth like `operation 2 will be repeated (until it stops returning -EAGAIN) if handle is blocking`.
___
*[`libos/src/net/unix.c` line 328 at r1](https://reviewable.io/reviews/gramineproject/gramine/862#-NAOSyzjAiGNmrLSmqS0:-NAOSyzjAiGNmrLSmqS1:bsavnlg) ([raw file](https://github.com/gramineproject/gramine/blob/2924c419620e637d5eb2b01a43a109cdf27c6b69/libos/src/net/unix.c#L328)):*
> ```C
>      *             to use it does the change on PAL level and the last thread restores the correct
>      *             (currently set in LibOS) mode. Operation 3 takes that into account - sets flags
>      *             only in LibOS if `handle->info.sock.force_nonblocking_users` is nonzero.
> ```

Please add a reference to `libos/src/fs/socket/fs.c`
___
*[`libos/src/net/unix.c` line 332 at r1](https://reviewable.io/reviews/gramineproject/gramine/862#-NAOTF0ZDGNcvpLS90Be:-NAOTF0ZDGNcvpLS90Bf:bt5pr6q) ([raw file](https://github.com/gramineproject/gramine/blob/2924c419620e637d5eb2b01a43a109cdf27c6b69/libos/src/net/unix.c#L332)):*
> ```C
>      * 2-3 races - this is inherently racy - operation 2 will be either in the old or new mode,
>      *             depending on how it ends up ordered vs operation 3. This works exactly the same
>      *             on normal Linux.
> ```

Again too short comment, and thus ambigious. Do you mean that we can basically ignore any special handling of this case because a read/write may be blocking or non-blocking anyhow? Like, the user can never make an assumption that in such case, operation 2 will actually block.
___
*[`libos/src/net/unix.c` line 385 at r1](https://reviewable.io/reviews/gramineproject/gramine/862#-NAOV6g78wm8vsKynlCA:-NAOV6g78wm8vsKynlCB:b-owmxen) ([raw file](https://github.com/gramineproject/gramine/blob/2924c419620e637d5eb2b01a43a109cdf27c6b69/libos/src/net/unix.c#L385)):*
> ```C
>                 log_error("%s: nonblocking restore: failed to get handle attrs: %d", __func__,
>                           tmp_ret);
>                 PalProcessExit(1);
> ```

Why do you do die here (when we restore "blockiness"), but not in the above code (when we "force non-blockiness")?

I would expect to die in both cases...

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/862)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

___
*[`libos/include/libos_handle.h` line 90 at r1](https://reviewable.io/reviews/gramineproject/gramine/862#-NAOPpzP2m_GJ8dG2Vuy:-NAOxZHgEUjbfSkrp4xQ:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/2924c419620e637d5eb2b01a43a109cdf27c6b69/libos/include/libos_handle.h#L90)):*

Well, if it's only 2-3 fields, I wouldn't bother.
</blockquote></details>

Done.
___
*[`libos/include/libos_handle.h` line 113 at r1](https://reviewable.io/reviews/gramineproject/gramine/862#-NAOPzQW3EnXQOwZm3IP:-NAOwkm-Dd-95UGbBzq7:b-9g3ruo) ([raw file](https://github.com/gramineproject/gramine/blob/2924c419620e637d5eb2b01a43a109cdf27c6b69/libos/include/libos_handle.h#L113)):*

Maybe remove this comment? You have a top-level comment on this, and this current comment is ambigious -- does it refer only to the next line, or to all next lines until the next comment?
</blockquote></details>

I think such comments always refer to the next line, unless stated otherwise
___
*[`libos/include/libos_handle.h` line 114 at r1](https://reviewable.io/reviews/gramineproject/gramine/862#-NAOSVf76p0kA9wGjCfd:-NAOzYB26h6tLjaQhGSc:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/2924c419620e637d5eb2b01a43a109cdf27c6b69/libos/include/libos_handle.h#L114)):*

Maybe append `_cnt`?
</blockquote></details>

Done.
___
*[`libos/src/net/unix.c` line 324 at r1](https://reviewable.io/reviews/gramineproject/gramine/862#-NAORTQv9KIlQAfyAlWb:-NAOzeM92DtFL6RQ7rRp:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/2924c419620e637d5eb2b01a43a109cdf27c6b69/libos/src/net/unix.c#L324)):*

`repeated`
</blockquote></details>

Done.
___
*[`libos/src/net/unix.c` line 324 at r1](https://reviewable.io/reviews/gramineproject/gramine/862#-NAORUJe9yzivcldAQJ0:-NAP-3363AXAQEZkbRCH:b81hmnb) ([raw file](https://github.com/gramineproject/gramine/blob/2924c419620e637d5eb2b01a43a109cdf27c6b69/libos/src/net/unix.c#L324)):*

I don't understand the phrase `operation 2 will be repeated if handle is blocking`. Could you expand?

I understand that we have two scenarios:
1. Underlying non-blocking handle. Then `force_nonblocking` is irrelevant, all operations are non-blocking. So doesn't matter in which order we execute operation 1 and operation 2.
2. Underlying blocking handle. Operation 1 temporarily makes the handle non-blocking and performs the read/write, no questions here. But operation 2 will somehow notice that the handle became non-blocking, and defer its execution until it finds out that the handle is again blocking? (I didn't look at the code yet.) But how does this "noticing and defering" works exactly?

UPDATE: Actually, the code simply checks for `-EAGAIN` error which signals that the handle is currently in non-blocking mode, and retries until the `-EAGAIN` goes away. The code is easier to understand than this comment :) Maybe just add smth like `operation 2 will be repeated (until it stops returning -EAGAIN) if handle is blocking`.
</blockquote></details>

This comment was intended to explain why the code below works w.r.t. different races. Logically it should be put after the code - you need to read the implementation first - but that would look weird. I don't want to explain what the code below does - it's rather simple, there's no need for that.

I've expanded this comment a bit, let me know what you think.
___
*[`libos/src/net/unix.c` line 328 at r1](https://reviewable.io/reviews/gramineproject/gramine/862#-NAOSyzjAiGNmrLSmqS0:-NAP0FfhCLh8EtoWtV6M:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/2924c419620e637d5eb2b01a43a109cdf27c6b69/libos/src/net/unix.c#L328)):*

Please add a reference to `libos/src/fs/socket/fs.c`
</blockquote></details>

Done.
___
*[`libos/src/net/unix.c` line 332 at r1](https://reviewable.io/reviews/gramineproject/gramine/862#-NAOTF0ZDGNcvpLS90Be:-NAP0KCY0PFtWIZSDP0A:b-umjc0v) ([raw file](https://github.com/gramineproject/gramine/blob/2924c419620e637d5eb2b01a43a109cdf27c6b69/libos/src/net/unix.c#L332)):*

Again too short comment, and thus ambigious. Do you mean that we can basically ignore any special handling of this case because a read/write may be blocking or non-blocking anyhow? Like, the user can never make an assumption that in such case, operation 2 will actually block.
</blockquote></details>

Yes, if user concurrently does 2 and 3, it's racy by definition.
___
*[`libos/src/net/unix.c` line 385 at r1](https://reviewable.io/reviews/gramineproject/gramine/862#-NAOV6g78wm8vsKynlCA:-NAP13vwCma_IBteAK6B:b2x9jdi) ([raw file](https://github.com/gramineproject/gramine/blob/2924c419620e637d5eb2b01a43a109cdf27c6b69/libos/src/net/unix.c#L385)):*

Why do you do die here (when we restore "blockiness"), but not in the above code (when we "force non-blockiness")?

I would expect to die in both cases...
</blockquote></details>

Because above can be cleanly handled - we can just return the error, nothing was changed yet. Here we have no way of restoring the balance - the error happened while restoring, thus we have no other choice, but to kill everything.

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/862)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

___
*[`libos/src/net/unix.c` line 385 at r1](https://reviewable.io/reviews/gramineproject/gramine/862#-NAOV6g78wm8vsKynlCA:-NAP4x2v4VZS_yNHpfCw:b-qezkjt) ([raw file](https://github.com/gramineproject/gramine/blob/2924c419620e637d5eb2b01a43a109cdf27c6b69/libos/src/net/unix.c#L385)):*

Because above can be cleanly handled - we can just return the error, nothing was changed yet. Here we have no way of restoring the balance - the error happened while restoring, thus we have no other choice, but to kill everything.
</blockquote></details>

Ok, fair enough.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/862)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #866, opened 2022-08-29
[PAL] Remove `PAL_OPTION_CLOEXEC` flag

### LINKED ISSUE #864: [PAL] Host fds from checkpointing are leaked on `fork()`
## Description of the problem
When Gramine emulates `fork()` syscall, it spawns another Gramine instance by doing `fork() + execve()` on the host OS. All open file descriptors are supposed to have `O_CLOEXEC` flag set, so that they are all closed - Gramine then sends (via ancillary data on unix domain socket) only those that are required in the child process.
The problem is that these fds don't have `O_CLOEXEC` set as they are received in the child process. Sending fds via ancillary data is lie doing `dup()` - `O_CLOEXEC` is not set on the new fd.

## Steps to reproduce
```c
#define _GNU_SORUCE
#include <stdlib.h>
#include <unistd.h>
#include <sys/wait.h>
#include <stdio.h>
#include <fcntl.h>

#include "common.h"

static void open_fds(int* fds, size_t len) {
    for (size_t i = 0; i < len; i++) {
        fds[i] = CHECK(open("/bin/sh", O_RDONLY));
    }
}

int main(void) {
    setbuf(stdout, NULL);
    setbuf(stderr, NULL);

    /* Get some never-closed fds. */
    int fds[100];
    open_fds(fds, ARRAY_LEN(fds));

    pid_t p = CHECK(fork());
    if (p == 0) {
        while (1) {
            printf("process %d looping\n", getpid());
            p = CHECK(fork());
            if (p != 0) {
                /* parent */
                exit(0);
            }
        }
        return 0;
    }

    while (1) {
        __asm__ volatile ("pause");
    }
    return 0;
}
```

## Expected results
Inf loop without any errors (except for IPC notification, which are unrelated and due to parent exiting).

## Actual results
Processes run out of fds and error out.

## Gramine commit hash
33576d39a844c3ab554d76e150baa3058a97f638

boryspoplawski:
## Description of the changes 
This flag was used to control `CLOEXEC` flag of host OS fd, and as such, was Linux specific. It was also not really needed - Linux and Linux SGX PALs always set it on host fd anyway.
This commit also adds all missing `CLOEXEC` flag usages in Linux and Linux SGX PALs.

Fixes #864

## How to test this PR? 
`fdleak` test augmented.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/866)

### DIFF
--- a/common/include/linux_socket.h
+++ b/common/include/linux_socket.h
@@ -61,6 +61,7 @@ struct mmsghdr {
 #define MSG_TRUNC 0x20
 #define MSG_DONTWAIT 0x40
 #define MSG_NOSIGNAL 0x4000
+#define MSG_CMSG_CLOEXEC 0x40000000
 
 /* Option levels. */
 #define SOL_SOCKET 1
--- a/libos/include/libos_flags_conv.h
+++ b/libos/include/libos_flags_conv.h
@@ -53,6 +53,5 @@ static inline enum pal_create_mode LINUX_OPEN_FLAGS_TO_PAL_CREATE(int flags) {
 }
 
 static inline pal_stream_options_t LINUX_OPEN_FLAGS_TO_PAL_OPTIONS(int flags) {
-    return (flags & O_CLOEXEC  ? PAL_OPTION_CLOEXEC  : 0) |
-           (flags & O_NONBLOCK ? PAL_OPTION_NONBLOCK : 0);
+    return flags & O_NONBLOCK ? PAL_OPTION_NONBLOCK : 0;
 }
--- a/libos/src/libos_pollable_event.c
+++ b/libos/src/libos_pollable_event.c
@@ -23,7 +23,7 @@ int create_pollable_event(struct libos_pollable_event* event) {
     PAL_HANDLE write_handle;
     do {
         ret = PalStreamOpen(uri, PAL_ACCESS_RDWR, /*share_flags=*/0, PAL_CREATE_IGNORED,
-                            PAL_OPTION_NONBLOCK | PAL_OPTION_CLOEXEC, &write_handle);
+                            PAL_OPTION_NONBLOCK, &write_handle);
     } while (ret == -PAL_ERROR_INTERRUPTED);
     if (ret < 0) {
         log_error("%s: PalStreamOpen failed: %d", __func__, ret);
--- a/libos/src/sys/libos_eventfd.c
+++ b/libos/src/sys/libos_eventfd.c
@@ -45,7 +45,6 @@ static int create_eventfd(PAL_HANDLE* efd, uint64_t initial_count, int flags) {
     int pal_flags = 0;
 
     pal_flags |= flags & EFD_NONBLOCK ? PAL_OPTION_NONBLOCK : 0;
-    pal_flags |= flags & EFD_CLOEXEC ? PAL_OPTION_CLOEXEC : 0;
     pal_flags |= flags & EFD_SEMAPHORE ? PAL_OPTION_EFD_SEMAPHORE : 0;
 
     ret = PalStreamOpen(URI_PREFIX_EVENTFD, PAL_ACCESS_RDWR, /*share_flags=*/0,
--- a/libos/test/regression/fdleak.c
+++ b/libos/test/regression/fdleak.c
@@ -1,27 +1,81 @@
+#define _GNU_SOURCE
+#include <err.h>
 #include <fcntl.h>
 #include <stdio.h>
 #include <stdlib.h>
-#include <sys/stat.h>
 #include <sys/types.h>
+#include <sys/wait.h>
 #include <unistd.h>
 
-/* This is supposed to expose resource leaks where close()d files are not properly cleaned up. */
+#include "common.h"
 
-int main(int argc, char** argv) {
-    for (int i = 0; i < 10000; i++) {
-        int fd = open(argv[0], O_RDONLY);
-        if (fd == -1)
-            abort();
-        char buf[1024];
-        ssize_t read_ret = read(fd, buf, sizeof(buf));
-        if (read_ret == -1)
-            abort();
-        int ret = close(fd);
-        if (ret == -1)
-            abort();
+#define OPEN_CLOSE_FILES_COUNT 1000
+#define INITIAL_OPEN_FDS 10
+#define SPAWN_CHILDREN_COUNT 10
+
+static void test_open_close(const char* fname) {
+    for (size_t i = 0; i < OPEN_CLOSE_FILES_COUNT; i++) {
+        int fd = CHECK(open(fname, O_RDONLY));
+        char buf[0x10];
+        ssize_t ret = CHECK(read(fd, buf, sizeof(buf)));
+        if (ret != sizeof(buf)) {
+            errx(1, "short read on a file: %zd", ret);
+        }
+        CHECK(close(fd));
+    }
+}
+
+static void test_open_fork(const char* fname) {
+    int fds[INITIAL_OPEN_FDS];
+    for (size_t i = 0; i < ARRAY_LEN(fds); i++) {
+        fds[i] = CHECK(open(fname, O_RDONLY));
     }
 
-    puts("Test succeeded.");
+    int pipefds[2];
+    CHECK(pipe(pipefds));
 
+    pid_t p = CHECK(fork());
+    if (p == 0) {
+        CHECK(close(pipefds[0]));
+
+        for (size_t i = 0; i < SPAWN_CHILDREN_COUNT; i++) {
+            p = CHECK(fork());
+            if (p != 0) {
+                /* Parent just exits. */
+                exit(0);
+            }
+            /* Child continues and spawns the next child (nested forks). */
+        }
+
+        /* Last child done, inform the first process. */
+        char c = 0;
+        CHECK(write(pipefds[1], &c, sizeof(c)));
+        exit(0);
+    }
+
+    CHECK(close(pipefds[1]));
+
+    char c = 0;
+    ssize_t ret = CHECK(read(pipefds[0], &c, sizeof(c)));
+    if (ret != sizeof(c)) {
+        errx(1, "confirmation message too short: %zd", ret);
+    }
+    CHECK(wait(NULL));
+
+    for (size_t i = 0; i < ARRAY_LEN(fds); i++) {
+        CHECK(close(fds[i]));
+    }
+}
+
+int main(int argc, char** argv) {
+    if (argc != 1) {
+        errx(1, "unexpected argc: %d (expected 1)", argc);
+    }
+
+    test_open_close(argv[0]);
+
+    test_open_fork(argv[0]);
+
+    puts("TEST OK");
     return 0;
 }
--- a/libos/test/regression/test_libos.py
+++ b/libos/test/regression/test_libos.py
@@ -1019,8 +1019,11 @@ class TC_40_FileSystem(RegressionTestCase):
         self.assertIn('/proc/stat test passed', stdout)
 
     def test_030_fdleak(self):
-        stdout, _ = self.run_binary(['fdleak'], timeout=10)
-        self.assertIn("Test succeeded.", stdout)
+        # The fd limit is rather arbitrary, but must be in sync with numbers from the test.
+        # Currently test opens 10 fds simultaneously, so 50 is a safe margin for any fds that
+        # Gramine might open internally.
+        stdout, _ = self.run_binary(['fdleak'], timeout=40, open_fds_limit=50)
+        self.assertIn("TEST OK", stdout)
 
     def get_cache_levels_cnt(self):
         cpu0 = '/sys/devices/system/cpu/cpu0/'
--- a/pal/include/host/linux-common/pal_flags_conv.h
+++ b/pal/include/host/linux-common/pal_flags_conv.h
@@ -63,7 +63,6 @@ static inline int PAL_CREATE_TO_LINUX_OPEN(enum pal_create_mode create) {
 }
 
 static inline int PAL_OPTION_TO_LINUX_OPEN(pal_stream_options_t options) {
-    assert(WITHIN_MASK(options, PAL_OPTION_CLOEXEC | PAL_OPTION_NONBLOCK | PAL_OPTION_PASSTHROUGH));
-    return (options & PAL_OPTION_CLOEXEC  ? O_CLOEXEC  : 0) |
-           (options & PAL_OPTION_NONBLOCK ? O_NONBLOCK : 0);
+    assert(WITHIN_MASK(options, PAL_OPTION_NONBLOCK | PAL_OPTION_PASSTHROUGH));
+    return options & PAL_OPTION_NONBLOCK ? O_NONBLOCK : 0;
 }
--- a/pal/include/pal/pal.h
+++ b/pal/include/pal/pal.h
@@ -50,6 +50,12 @@ typedef struct {
     struct handle_ops* ops;
 } PAL_HDR;
 
+/*
+ * This header provides `PAL_HANDLE` type definition.
+ * All host resources being part of `PAL_HANDLE` must be released when spawning another process
+ * (`PalProcessCreate`), e.g. Linux PAL must set `OCLOEXEC` flag on all open file descriptors.
+ * LibOS layer takes care of migrating all necessary handles using `PalSendHandle`.
+ */
 #include "pal_host.h"
 
 static inline void init_handle_hdr(PAL_HANDLE handle, int pal_type) {
@@ -259,11 +265,10 @@ enum pal_create_mode {
 
 /*! stream misc flags */
 typedef uint32_t pal_stream_options_t; /* bitfield */
-#define PAL_OPTION_CLOEXEC         0x1
-#define PAL_OPTION_EFD_SEMAPHORE   0x2 /*!< specific to `eventfd` syscall */
-#define PAL_OPTION_NONBLOCK        0x4
-#define PAL_OPTION_PASSTHROUGH     0x8 /*!< Disregard `sgx.{allowed,trusted}_files` */
-#define PAL_OPTION_MASK            0xF
+#define PAL_OPTION_EFD_SEMAPHORE   0x1 /*!< specific to `eventfd` syscall */
+#define PAL_OPTION_NONBLOCK        0x2
+#define PAL_OPTION_PASSTHROUGH     0x4 /*!< Disregard `sgx.{allowed,trusted}_files` */
+#define PAL_OPTION_MASK            0x7
 
 /*!
  * \brief Open/create a stream resource specified by `uri`.
--- a/pal/src/host/linux-sgx/gdb_integration/sgx_gdb.c
+++ b/pal/src/host/linux-sgx/gdb_integration/sgx_gdb.c
@@ -429,7 +429,7 @@ static int open_memdevice(pid_t tid, int* out_memdev, struct enclave_dbginfo** o
     }
 
     snprintf(memdev_path, sizeof(memdev_path), "/proc/%d/mem", tid);
-    fd = open(memdev_path, O_RDWR);
+    fd = open(memdev_path, O_RDWR | O_CLOEXEC);
     if (fd < 0) {
         DEBUG_LOG("Cannot open %s\n", memdev_path);
         ret = -2;
--- a/pal/src/host/linux-sgx/host_log.c
+++ b/pal/src/host/linux-sgx/host_log.c
@@ -24,7 +24,7 @@ int host_log_init(const char* path) {
             return ret;
     }
 
-    ret = DO_SYSCALL(open, path, O_WRONLY | O_APPEND | O_CREAT, PERM_rw_______);
+    ret = DO_SYSCALL(open, path, O_WRONLY | O_APPEND | O_CREAT | O_CLOEXEC, PERM_rw_______);
     if (ret < 0)
         return ret;
     g_host_log_fd = ret;
--- a/pal/src/host/linux-sgx/host_main.c
+++ b/pal/src/host/linux-sgx/host_main.c
@@ -214,7 +214,8 @@ static int initialize_enclave(struct pal_enclave* enclave, const char* manifest_
     /* this array may overflow the stack, so we allocate it in BSS */
     static void* tcs_addrs[MAX_DBG_THREADS];
 
-    enclave_image = DO_SYSCALL(open, enclave->libpal_uri + URI_PREFIX_FILE_LEN, O_RDONLY, 0);
+    enclave_image = DO_SYSCALL(open, enclave->libpal_uri + URI_PREFIX_FILE_LEN,
+                               O_RDONLY | O_CLOEXEC, 0);
     if (enclave_image < 0) {
         log_error("Cannot find enclave image: %s", enclave->libpal_uri);
         ret = enclave_image;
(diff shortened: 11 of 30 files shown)


boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/866)* status: 0 of 27 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/866#-NAeHuupAqdbu4l9wsi8:-NAeHuupAqdbu4l9wsi9:bhvzqf3) (no related file):*
I think it would be worth adding a test to CI, but I have no good idea on how to do that so it doesn't take too much time.
Maybe `ulimit` on host + some well calculated loop forking processes? Still, that would take some time on SGX PAL.

Well, I'll do that.

___

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/866)* status: 0 of 30 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/866#-NAeHuupAqdbu4l9wsi8:-NAeeJ8TBM4lyOKKki2v:b7mgdod) (no related file):*

I think it would be worth adding a test to CI, but I have no good idea on how to do that so it doesn't take too much time.
Maybe `ulimit` on host + some well calculated loop forking processes? Still, that would take some time on SGX PAL.

Well, I'll do that.
</blockquote></details>

Ok, added a test to CI.

___

dimakuv (changes requested):
> Reviewed 27 of 27 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/866)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`-- commits` line 8 at r2](https://reviewable.io/reviews/gramineproject/gramine/866#-NAhTEaF55Aa7LXIJx7Y):*
Don't you want to add that this fixes FD leaks in nested-fork scenarios? This sounds like an important phrase to add here, for historial reasons.

Also, could we make it properly separate paragraphs? It deserves its own para (with an empty line in between).
___
*[`libos/test/regression/fdleak.c` line 46 at r2](https://reviewable.io/reviews/gramineproject/gramine/866#-NAhUrEo3UxRhd_XpZnj:-NAhUrEo3UxRhd_XpZnk:b-85p8kc) ([raw file](https://github.com/gramineproject/gramine/blob/ff2eabaf92f7c4e3de31f071d628f7e44b6edaaf/libos/test/regression/fdleak.c#L46)):*
> ```C
>                 /* Parent just exits. */
>                 exit(0);
>             }
> ```

Can you maybe add a comment after this `if (p != 0) { }`:
```
/* child continues the loop and will spawn another child (deeply nested fork) */
```

It took me a couple minutes to understand what happens here
___
*[`libos/test/regression/fdleak.c` line 62 at r2](https://reviewable.io/reviews/gramineproject/gramine/866#-NAhVZ9lBx3KjAzI6-06:-NAhVZ9lBx3KjAzI6-07:borer8y) ([raw file](https://github.com/gramineproject/gramine/blob/ff2eabaf92f7c4e3de31f071d628f7e44b6edaaf/libos/test/regression/fdleak.c#L62)):*
> ```C
>         errx(1, "confirmation message too short: %zd", ret);
>     }
>     CHECK(wait(NULL));
> ```

How are we guaranteed that at this point other (nested) children are finished? Or we just don't care whether they performed `exit(0)` already?
___
*[`libos/test/regression/fdleak.c` line 71 at r2](https://reviewable.io/reviews/gramineproject/gramine/866#-NAhTvUD5BfJl-GfjYqt:-NAhTvUD5BfJl-GfjYqu:b-ccczzj) ([raw file](https://github.com/gramineproject/gramine/blob/ff2eabaf92f7c4e3de31f071d628f7e44b6edaaf/libos/test/regression/fdleak.c#L71)):*
> ```C
> int main(int argc, char** argv) {
>     if (argc != 1) {
>         errx(1, "invalid argc: %d", argc);
> ```

Maybe print a message that says `test expects argc == 1 (received: %d)`?
___
*[`libos/test/regression/test_libos.py` line 1015 at r2](https://reviewable.io/reviews/gramineproject/gramine/866#-NAhWKiZ6KcsH6wcSEDc:-NAhWKiZ6KcsH6wcSEDd:b-hu0jqj) ([raw file](https://github.com/gramineproject/gramine/blob/ff2eabaf92f7c4e3de31f071d628f7e44b6edaaf/libos/test/regression/test_libos.py#L1015)):*
> ```Python
> 
>     def test_030_fdleak(self):
>         stdout, _ = self.run_binary(['fdleak'], timeout=10, open_fds_limit=50)
> ```

Oh, this `open_fds_limit` trick is cool!

But maybe we should add a comment explaining why we have such a high limit? (That Gramine internally creates several more FDs, so we can't use a tight bound like ~15).
___
*[`libos/test/regression/test_libos.py` line 1015 at r2](https://reviewable.io/reviews/gramineproject/gramine/866#-NAhWbrX8s-LTDJvOpay:-NAhWbrX8s-LTDJvOpaz:b78mvwk) ([raw file](https://github.com/gramineproject/gramine/blob/ff2eabaf92f7c4e3de31f071d628f7e44b6edaaf/libos/test/regression/test_libos.py#L1015)):*
> ```Python
> 
>     def test_030_fdleak(self):
>         stdout, _ = self.run_binary(['fdleak'], timeout=10, open_fds_limit=50)
> ```

How can the timeout be so small in the SGX case? You're creating 11 enclaves! Something feels wrong here.
___
*[`pal/src/host/linux/pal_memory.c` line 93 at r2](https://reviewable.io/reviews/gramineproject/gramine/866#-NAhXH81DL-tswd2FlEC:-NAhXH81DL-tswd2FlED:bn9tp7h) ([raw file](https://github.com/gramineproject/gramine/blob/ff2eabaf92f7c4e3de31f071d628f7e44b6edaaf/pal/src/host/linux/pal_memory.c#L93)):*
> ```C
> 
> static int read_proc_meminfo(const char* key, unsigned long* val) {
>     int fd = DO_SYSCALL(open, "/proc/meminfo", O_RDONLY | O_CLOEXEC, 0);
> ```

Why would it matter here?
___
*[`pal/src/host/linux/pal_streams.c` line 253 at r2](https://reviewable.io/reviews/gramineproject/gramine/866#-NAhXTsw9ng55mHMZ8tb:-NAhXTsw9ng55mHMZ8tc:b9rk2mw) ([raw file](https://github.com/gramineproject/gramine/blob/ae6cad8a5758acaf5f4fe8073234718c57cfd08f/pal/src/host/linux/pal_streams.c#L253)):*
> ```C
>         /* This check is to shield from a Iago attack. We know that sendmsg() in _PalSendHandle()
>          * transfers the message atomically, and that our recvmsg() receives it atomically. So
>          * the only valid values for ret must be zero or the size of the header. */
> ```

That's just not true because `ret == 0` should never happen?

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/866)* status: 28 of 30 files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`-- commits` line 8 at r2](https://reviewable.io/reviews/gramineproject/gramine/866#-NAhTEaF55Aa7LXIJx7Y):*

Don't you want to add that this fixes FD leaks in nested-fork scenarios? This sounds like an important phrase to add here, for historial reasons.

Also, could we make it properly separate paragraphs? It deserves its own para (with an empty line in between).
</blockquote></details>

Something like:
```
[PAL] Remove `PAL_OPTION_CLOEXEC` flag                                   
                                                                         
This flag was used to control `CLOEXEC` flag of host OS fd and, as such, 
was Linux specific. It was also not really needed - Linux and Linux SGX  
PALs always set it on host fd anyway.                                    
                                                                         
This commit also adds all missing `CLOEXEC` flag usages in Linux and     
Linux SGX PALs, which fixes multiple fd leaks (all Linux PALs do `fork`  
+ `execve` when spawning a new process, which leaks fds without          
`CLOEXEC` flag).
```
?
___
*[`libos/test/regression/fdleak.c` line 46 at r2](https://reviewable.io/reviews/gramineproject/gramine/866#-NAhUrEo3UxRhd_XpZnj:-NAixS_XDnmSi_5Vn4qa:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/ff2eabaf92f7c4e3de31f071d628f7e44b6edaaf/libos/test/regression/fdleak.c#L46)):*

Can you maybe add a comment after this `if (p != 0) { }`:
```
/* child continues the loop and will spawn another child (deeply nested fork) */
```

It took me a couple minutes to understand what happens here
</blockquote></details>

Done.
___
*[`libos/test/regression/fdleak.c` line 62 at r2](https://reviewable.io/reviews/gramineproject/gramine/866#-NAhVZ9lBx3KjAzI6-06:-NAixWEaA-xarq6TPpiv:bs5b8so) ([raw file](https://github.com/gramineproject/gramine/blob/ff2eabaf92f7c4e3de31f071d628f7e44b6edaaf/libos/test/regression/fdleak.c#L62)):*

How are we guaranteed that at this point other (nested) children are finished? Or we just don't care whether they performed `exit(0)` already?
</blockquote></details>

We are not, but why would we care? There is no way of doing this without `PR_SET_CHILD_SUBREAPER`, which is not supported in Gramine. All we are is that we do enough nested forks, which we know happened, because we successfully read data from the pipe.
___
*[`libos/test/regression/fdleak.c` line 71 at r2](https://reviewable.io/reviews/gramineproject/gramine/866#-NAhTvUD5BfJl-GfjYqt:-NAiy6Zs4IqpklfD7rGF:b-iufoh4) ([raw file](https://github.com/gramineproject/gramine/blob/ff2eabaf92f7c4e3de31f071d628f7e44b6edaaf/libos/test/regression/fdleak.c#L71)):*

Maybe print a message that says `test expects argc == 1 (received: %d)`?
</blockquote></details>

If you want I can do this, but this is really just a sanity check...
___
*[`libos/test/regression/test_libos.py` line 1015 at r2](https://reviewable.io/reviews/gramineproject/gramine/866#-NAhWKiZ6KcsH6wcSEDc:-NAiylwe1JXUiRKegSf0:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/ff2eabaf92f7c4e3de31f071d628f7e44b6edaaf/libos/test/regression/test_libos.py#L1015)):*

Oh, this `open_fds_limit` trick is cool!

But maybe we should add a comment explaining why we have such a high limit? (That Gramine internally creates several more FDs, so we can't use a tight bound like ~15).
</blockquote></details>

Done.
___
*[`libos/test/regression/test_libos.py` line 1015 at r2](https://reviewable.io/reviews/gramineproject/gramine/866#-NAhWbrX8s-LTDJvOpay:-NAizXeJDN7FP9XT96jM:baid7bq) ([raw file](https://github.com/gramineproject/gramine/blob/ff2eabaf92f7c4e3de31f071d628f7e44b6edaaf/libos/test/regression/test_libos.py#L1015)):*

How can the timeout be so small in the SGX case? You're creating 11 enclaves! Something feels wrong here.
</blockquote></details>

This is a very good question - this test takes 12s in debug mode on a decent (but not beefy) machine...
Update: ok, we take `max(timeout, DEFAULT_TIMEOUT)`, which is 20s for SGX.

I've increased timeout to 40, should be enough even for NUCs
___
*[`pal/src/host/linux/pal_memory.c` line 93 at r2](https://reviewable.io/reviews/gramineproject/gramine/866#-NAhXH81DL-tswd2FlEC:-NAjH4fqDDxjbxSe2vRT:bti14xy) ([raw file](https://github.com/gramineproject/gramine/blob/ff2eabaf92f7c4e3de31f071d628f7e44b6edaaf/pal/src/host/linux/pal_memory.c#L93)):*

Why would it matter here?
</blockquote></details>

Why not? We don't want this to stay open after execve, so why would we state otherwise?
Also keep in mind that there always could be a concurrent thread doing fork at any moment, so the fact that a function always closes the fd is not enough.
___
*[`pal/src/host/linux/pal_streams.c` line 253 at r2](https://reviewable.io/reviews/gramineproject/gramine/866#-NAhXTsw9ng55mHMZ8tb:-NAjHPyT-4Nmti4QEeHe:b4mm98h) ([raw file](https://github.com/gramineproject/gramine/blob/ae6cad8a5758acaf5f4fe8073234718c57cfd08f/pal/src/host/linux/pal_streams.c#L253)):*

That's just not true because `ret == 0` should never happen?
</blockquote></details>

This comment made no sense.
1) `ret == 0` could happen and we just return an error of our choice. I don't see a point in treating `0` in any special way.
2) This is Linux PAL, we trust the host OS.
3) Even if we didn't, we just check return value - this is normal, expected code, no need for fancy names.
4) I believe Iago attacks refer to malicious host returning wrong values, especially different/unexpected error codes and we already checked for `< 0` above, not here. Also I don't understand why this has a special name (which frankly speaking I've never heard before), this is nothing fancy/novel.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/866)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`-- commits` line 8 at r2](https://reviewable.io/reviews/gramineproject/gramine/866#-NAhTEaF55Aa7LXIJx7Y):*

Something like:
```
[PAL] Remove `PAL_OPTION_CLOEXEC` flag                                   
                                                                         
This flag was used to control `CLOEXEC` flag of host OS fd and, as such, 
was Linux specific. It was also not really needed - Linux and Linux SGX  
PALs always set it on host fd anyway.                                    
                                                                         
This commit also adds all missing `CLOEXEC` flag usages in Linux and     
Linux SGX PALs, which fixes multiple fd leaks (all Linux PALs do `fork`  
+ `execve` when spawning a new process, which leaks fds without          
`CLOEXEC` flag).
```
?
</blockquote></details>

Yes, this text is good.
___
*[`libos/test/regression/fdleak.c` line 62 at r2](https://reviewable.io/reviews/gramineproject/gramine/866#-NAhVZ9lBx3KjAzI6-06:-NAmtf2EEOJYe3gv2XyT:b-gxztsp) ([raw file](https://github.com/gramineproject/gramine/blob/ff2eabaf92f7c4e3de31f071d628f7e44b6edaaf/libos/test/regression/fdleak.c#L62)):*

We are not, but why would we care? There is no way of doing this without `PR_SET_CHILD_SUBREAPER`, which is not supported in Gramine. All we are is that we do enough nested forks, which we know happened, because we successfully read data from the pipe.
</blockquote></details>

Fair enough.
___
*[`libos/test/regression/fdleak.c` line 71 at r2](https://reviewable.io/reviews/gramineproject/gramine/866#-NAhTvUD5BfJl-GfjYqt:-NAmtk34EXHroI09RqXK:b-17rccj) ([raw file](https://github.com/gramineproject/gramine/blob/ff2eabaf92f7c4e3de31f071d628f7e44b6edaaf/libos/test/regression/fdleak.c#L71)):*

If you want I can do this, but this is really just a sanity check...
</blockquote></details>

Well, I frequently run LibOS tests manually, so this error could throw me off... So I'd want this.
___
*[`pal/src/host/linux/pal_memory.c` line 93 at r2](https://reviewable.io/reviews/gramineproject/gramine/866#-NAhXH81DL-tswd2FlEC:-NAmu7cbAXqaehfG7b_Q:b-qezkjt) ([raw file](https://github.com/gramineproject/gramine/blob/ff2eabaf92f7c4e3de31f071d628f7e44b6edaaf/pal/src/host/linux/pal_memory.c#L93)):*

Why not? We don't want this to stay open after execve, so why would we state otherwise?
Also keep in mind that there always could be a concurrent thread doing fork at any moment, so the fact that a function always closes the fd is not enough.
</blockquote></details>

Ok, fair enough.
___
*[`pal/src/host/linux/pal_streams.c` line 253 at r2](https://reviewable.io/reviews/gramineproject/gramine/866#-NAhXTsw9ng55mHMZ8tb:-NAmuWc5E8P-b5l2JsEg:bxjhpoq) ([raw file](https://github.com/gramineproject/gramine/blob/ae6cad8a5758acaf5f4fe8073234718c57cfd08f/pal/src/host/linux/pal_streams.c#L253)):*
Ok, even item 2 is enough to explain this deletion.

> Also I don't understand why this has a special name

Do you mean the term "Iago attack"? I don't know the reason for such a name, but this is from a conference paper. And all conference papers introduce some new names, just because that's how conferences work...

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/866)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[`libos/test/regression/fdleak.c` line 71 at r2](https://reviewable.io/reviews/gramineproject/gramine/866#-NAhTvUD5BfJl-GfjYqt:-NAntG7P06H5sU9Naxvq:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/ff2eabaf92f7c4e3de31f071d628f7e44b6edaaf/libos/test/regression/fdleak.c#L71)):*

Well, I frequently run LibOS tests manually, so this error could throw me off... So I'd want this.
</blockquote></details>

Done.
___
*[`pal/src/host/linux/pal_streams.c` line 253 at r2](https://reviewable.io/reviews/gramineproject/gramine/866#-NAhXTsw9ng55mHMZ8tb:-NAntVB_9SHK4bqP7N1c:b-juo85r) ([raw file](https://github.com/gramineproject/gramine/blob/ae6cad8a5758acaf5f4fe8073234718c57cfd08f/pal/src/host/linux/pal_streams.c#L253)):*

Ok, even item 2 is enough to explain this deletion.

> Also I don't understand why this has a special name

Do you mean the term "Iago attack"? I don't know the reason for such a name, but this is from a conference paper. And all conference papers introduce some new names, just because that's how conferences work...
</blockquote></details>

Well, I'll just stick to the term "checking return values".

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/866)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/866)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/866#-NAo22371GcshTU27Dr3:-NAo22371GcshTU27Dr4:b-pf1fdw) (no related file):*
Hmm, it would be good to document somewhere this requirement of having `CLOEXEC` on all host fds - the problem is that it's Linux and Linux SGX PALs specific (it's not PAL generic)

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/866)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/866#-NAo22371GcshTU27Dr3:-NArqeoi7qssptNDTp3Y:b-eo8ac3) (no related file):*

Hmm, it would be good to document somewhere this requirement of having `CLOEXEC` on all host fds - the problem is that it's Linux and Linux SGX PALs specific (it's not PAL generic)
</blockquote></details>

We have our documentation on PAL: https://gramine.readthedocs.io/en/latest/pal/host-abi.html. But it's old and not good and nobody ever looks at it, so I wouldn't put there.

Maybe just a high-level comment here: https://github.com/gramineproject/gramine/blob/7b369de0787eab57b80a569604155fb6d0e598d3/pal/include/pal/pal.h#L6

Something like "All PAL implementations must create handles on the host that will be automatically closed when switching to a new process image. In case of Linux-backed PALs, this is achieved by adding `O_CLOEXEC` to all created host fds."

___

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/866)* status: 29 of 30 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/866#-NAo22371GcshTU27Dr3:-NAszGZZDtkDVFPuaoU6:b-896fix) (no related file):*

We have our documentation on PAL: https://gramine.readthedocs.io/en/latest/pal/host-abi.html. But it's old and not good and nobody ever looks at it, so I wouldn't put there.

Maybe just a high-level comment here: https://github.com/gramineproject/gramine/blob/7b369de0787eab57b80a569604155fb6d0e598d3/pal/include/pal/pal.h#L6

Something like "All PAL implementations must create handles on the host that will be automatically closed when switching to a new process image. In case of Linux-backed PALs, this is achieved by adding `O_CLOEXEC` to all created host fds."
</blockquote></details>

Done.

___

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/866)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/866#-NAo22371GcshTU27Dr3:-NAt2C5DCftFWTYNVoh4:bi8rkil) (no related file):*

Done.
</blockquote></details>

Yes, my proposed text was incorrect. Yours is much better. Thanks.

___

mkow (approved):
> Reviewed 26 of 27 files at r1, 1 of 3 files at r2, 1 of 2 files at r3, 1 of 1 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/866)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/866)* status: all files reviewed, 1 unresolved discussion, not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski)

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/866)* status: all files reviewed, 1 unresolved discussion, not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski)

___
*[`-- commits` line 8 at r2](https://reviewable.io/reviews/gramineproject/gramine/866#-NAhTEaF55Aa7LXIJx7Y):*

Yes, this text is good.
</blockquote></details>

@boryspoplawski The text was not updated?

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/866)* status: all files reviewed, 1 unresolved discussion, not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

___
*[`-- commits` line 8 at r2](https://reviewable.io/reviews/gramineproject/gramine/866#-NAhTEaF55Aa7LXIJx7Y):*

@boryspoplawski The text was not updated?
</blockquote></details>

Yes, that's why I prefer squash commits...

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/866)* status: :shipit: complete! all files reviewed, all discussions resolved

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/866)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #875, opened 2022-09-02
[CI-Examples] Use `-Wl,--enable-new-dtags` in RA-TLS examples

### LINKED ISSUE #752: Error running the ra-tls-mbedtls example
I was trying to run the ra-tls-mbedtls example, intially it did not allow me to use the current update where the the value of sx.remote_attestation was changed to attestation type, it gave the following error -

**error: Cannot parse 'sgx.remote_attestation' (the value must be `true` or `false`)**

So I changed the value to true on the manifest template. 
After which it is currently giving me this error -

**Inconsistency detected by ld.so: dl-call-libc-early-init.c: 37: _dl_call_libc_early_init: Assertion `sym != NULL' failed!**

dimakuv:
## Description of the changes 

`ra-tls-mbedtls` and `ra-tls-secret-prov` examples link against MbedTLS version provided by Gramine, via `pkg-config --libs mbedtls_gramine` in Makefile. The pkg-config command generates a line that contains `-Wl,-rpath,<path to Gramine libs installation>`.

This line should instruct the linker to add the `DT_RUNPATH` search path, but some linkers add `DT_RPATH` instead, which results in `LD_LIBRARY_PATH` being ignored.

The final outcome is that Gramine loads e.g. host-level `/usr/lib/x86_64-linux-gnu/libc.so.6` instead of Gramine-specific `/lib/libc.so.6`, which conflicts with `/lib/ld-linux-x86-64.so.2`:

    Inconsistency detected by ld.so: dl-call-libc-early-init.c: ...

To fix this problem, this commit adds `-Wl,--enable-new-dtags` to all applications that link against `mbedtls_gramine` libs. This forces all linkers to use `DR_RUNPATH` instead of `DT_RPATH`.

Fixes #752.

## How to test this PR? 

N/A. Hopefully people affected by this bug can test it (I haven't encountered it myself). On my local machines, stuff like `ldd server` and `readelf -a server | grep PATH` shows correct things:
```
$ readelf -a secret_prov_server_epid  | grep PATH
 0x000000000000001d (RUNPATH)            Library runpath: [/home/dimakuv/gramineproject/gramine/built-debug/lib/x86_64-linux-gnu]
```

### DIFF
--- a/CI-Examples/ra-tls-mbedtls/Makefile
+++ b/CI-Examples/ra-tls-mbedtls/Makefile
@@ -1,6 +1,3 @@
-GRAMINEDIR ?= ../..
-GRAMINE_PKGLIBDIR ?= /usr/lib/x86_64-linux-gnu/gramine # this is debian/ubuntu specific
-
 ARCH_LIBDIR ?= /lib/$(shell $(CC) -dumpmachine)
 
 ifeq ($(DEBUG),1)
@@ -42,11 +39,8 @@ ssl/server.crt: ssl/ca_config.conf
 
 ############################# MBEDTLS DEPENDENCY ##############################
 
-#
-# This download is done to get headers in include/, because we currently fail
-# to provide the headers.
-# TODO: install headers, and `make install` before copying them
-#
+# This download is done to get headers in include/, because we currently fail to provide the headers
+# (`pkg-config --cflags mbedtls_gramine` in the below CFLAGS line returns a non-existing directory).
 
 MBEDTLS_VERSION ?= 3.2.1
 MBEDTLS_SRC ?= mbedtls-$(MBEDTLS_VERSION).tar.gz
@@ -78,7 +72,7 @@ mbedtls/.mbedtls_configured: mbedtls/.mbedtls_downloaded
 ######################### CLIENT/SERVER EXECUTABLES ###########################
 
 CFLAGS += -I./mbedtls/include $(shell pkg-config --cflags mbedtls_gramine)
-LDFLAGS += -ldl $(shell pkg-config --libs mbedtls_gramine)
+LDFLAGS += -ldl -Wl,--enable-new-dtags $(shell pkg-config --libs mbedtls_gramine)
 
 server: src/server.c mbedtls/.mbedtls_configured
 	$(CC) $< $(CFLAGS) $(LDFLAGS) -o $@
--- a/CI-Examples/ra-tls-secret-prov/Makefile
+++ b/CI-Examples/ra-tls-secret-prov/Makefile
@@ -1,6 +1,3 @@
-GRAMINEDIR ?= ../..
-GRAMINE_PKGLIBDIR ?= /usr/lib/x86_64-linux-gnu/gramine # this is debian/ubuntu specific
-
 ARCH_LIBDIR ?= /lib/$(shell $(CC) -dumpmachine)
 
 ifeq ($(DEBUG),1)
@@ -49,9 +46,13 @@ ssl/server.crt: ssl/ca_config.conf
 
 ######################### CLIENT/SERVER EXECUTABLES ###########################
 
+# Use hard-coded GRAMINEDIR because we currently fail to provide the headers in Gramine installation
+# (`pkg-config --cflags mbedtls_gramine` returns a non-existing directory).
+
+GRAMINEDIR ?= ../..
 CFLAGS += -Wall -std=c11 -I$(GRAMINEDIR)/tools/sgx/ra-tls \
           $(shell pkg-config --cflags mbedtls_gramine)
-LDFLAGS += $(shell pkg-config --libs mbedtls_gramine)
+LDFLAGS += -Wl,--enable-new-dtags $(shell pkg-config --libs mbedtls_gramine)
 
 secret_prov_server_epid: src/secret_prov_server.c
 	$(CC) $< $(CFLAGS) $(LDFLAGS) -lsecret_prov_verify_epid -pthread -o $@


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/875)* status: 0 of 2 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

___
*[`CI-Examples/ra-tls-mbedtls/Makefile` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/875#-NAxPnHoAk9Kixzd8z_L:-NAxPnHpDLHVMmd3_76U:b8e4fgq) ([raw file](https://github.com/gramineproject/gramine/blob/f1e1d22d29a08c05a6cf19188aa0241787608371/CI-Examples/ra-tls-mbedtls/Makefile#L2)):*
> ```Makefile
> GRAMINEDIR ?= ../..
> GRAMINE_PKGLIBDIR ?= /usr/lib/x86_64-linux-gnu/gramine # this is debian/ubuntu specific
> ```

FYI: These two lines were useless
___
*[`CI-Examples/ra-tls-mbedtls/Makefile` line 74 at r1](https://reviewable.io/reviews/gramineproject/gramine/875#-NAxPwVVC7WkUbmjRZmy:-NAxPwVVC7WkUbmjRZmz:bv5l1im) ([raw file](https://github.com/gramineproject/gramine/blob/35b19b29c88760e08023be4af1956adcc64502be/CI-Examples/ra-tls-mbedtls/Makefile#L74)):*
> ```Makefile
> ######################### CLIENT/SERVER EXECUTABLES ###########################
> 
> CFLAGS += -I./mbedtls/include $(shell pkg-config --cflags mbedtls_gramine)
> ```

FYI: I kept this `shell pkg-config --cflags mbedtls_gramine` because one day it will work hopefully...
___
*[`CI-Examples/ra-tls-secret-prov/Makefile` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/875#-NAxPrNg1pHdBrdXaCK-:-NAxPrNg1pHdBrdXaCK0:b96fjk0) ([raw file](https://github.com/gramineproject/gramine/blob/f1e1d22d29a08c05a6cf19188aa0241787608371/CI-Examples/ra-tls-secret-prov/Makefile#L2)):*
> ```Makefile
> GRAMINEDIR ?= ../..
> GRAMINE_PKGLIBDIR ?= /usr/lib/x86_64-linux-gnu/gramine # this is debian/ubuntu specific
> ```

FYI: The second line was useless. The first line I moved to its only usage.
___
*[`CI-Examples/ra-tls-secret-prov/Makefile` line 54 at r1](https://reviewable.io/reviews/gramineproject/gramine/875#-NAxQ4Mp7Kg7EGTtEFAY:-NAxQ4Mp7Kg7EGTtEFAZ:bv5l1im) ([raw file](https://github.com/gramineproject/gramine/blob/35b19b29c88760e08023be4af1956adcc64502be/CI-Examples/ra-tls-secret-prov/Makefile#L54)):*
> ```Makefile
> GRAMINEDIR ?= ../..
> CFLAGS += -Wall -std=c11 -I$(GRAMINEDIR)/tools/sgx/ra-tls \
>           $(shell pkg-config --cflags mbedtls_gramine)
> ```

FYI: I kept this `shell pkg-config --cflags mbedtls_gramine` because one day it will work hopefully...

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/875)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`-- commits` line 5 at r1](https://reviewable.io/reviews/gramineproject/gramine/875#-NAyM2E28cVFflf6PhvT):*
This command doesn't link, it provides arguments to the linker.
___
*[`-- commits` line 9 at r1](https://reviewable.io/reviews/gramineproject/gramine/875#-NAyMFjx-BRUlrb5aMYs):*
I think the fact that this is done only by some older linker is important in the context of how we fix it



_Suggestion:_
```
older linkers
```
___
*[`-- commits` line 10 at r1](https://reviewable.io/reviews/gramineproject/gramine/875#-NAyMRfo73r21QNIITXo):*
`LD_LIBRARY_PATH` is not ignored, it just has lower priority and it so happens that `libc.so` is found at directory pointed at `DR_RPATH`, which takes precedence.
___
*[`-- commits` line 11 at r1](https://reviewable.io/reviews/gramineproject/gramine/875#-NAyL-650HmLHYm3m2uB):*
These paths are:
1) host specific (e.g. different distros will have different paths), whereas the problem is not,
2) can be different, depending on where you have stuff installed,
3) `/lib/libc.so.6` is a in-Gramine path, which we've chosen to use in our examples. It's not the host path (which is what you are talking about here).

Maybe just remove the paths and stick to `host libc` vs `Gramine provided libc`? Or at least use host paths and take the default installation paths from our apt package. 
___
*[`-- commits` line 14 at r1](https://reviewable.io/reviews/gramineproject/gramine/875#-NAyK6KoDOzGqtGdRDIv):*
Is it worth having this error message here?
The problem I see is that this is only one of the things that can  go wrong / one of the errors. Simple crash (segmentation fault) is a likely outcome as well (this is what I always encountered when mixed ld vs libc). This one just happens to be printed on the specific combination of ld.so we currently use and current version of glibc from Ubuntu 20.04.

Not blocking though.
___
*[`-- commits` line 18 at r1](https://reviewable.io/reviews/gramineproject/gramine/875#-NAyN2Ft2p9SPIDu_c0E):*
> all linkers

Isn't this flag GNU ld specific?
___
*[`CI-Examples/ra-tls-mbedtls/Makefile` line 48 at r1](https://reviewable.io/reviews/gramineproject/gramine/875#-NAyNHwq3-h451KaixJy:-NAyNHwq3-h451KaixJz:b-7qkta6) ([raw file](https://github.com/gramineproject/gramine/blob/f1e1d22d29a08c05a6cf19188aa0241787608371/CI-Examples/ra-tls-mbedtls/Makefile#L48)):*
> ```Makefile
> # This download is done to get headers in include/, because we currently fail
> # to provide the headers.
> # TODO: install headers, and `make install` before copying them
> ```

Could you add some TODO here? I don't understand the old one (it was outdated?), but it would be good to have something here.
___
*[`CI-Examples/ra-tls-mbedtls/Makefile` line 74 at r1](https://reviewable.io/reviews/gramineproject/gramine/875#-NAxPwVVC7WkUbmjRZmy:-NAyNdih7_pEidKPCnbu:b-7wvtjh) ([raw file](https://github.com/gramineproject/gramine/blob/35b19b29c88760e08023be4af1956adcc64502be/CI-Examples/ra-tls-mbedtls/Makefile#L74)):*

FYI: I kept this `shell pkg-config --cflags mbedtls_gramine` because one day it will work hopefully...
</blockquote></details>

Maybe change this into a TODO? I think these two includes are mutually exclusive
___
*[`CI-Examples/ra-tls-secret-prov/Makefile` line 54 at r1](https://reviewable.io/reviews/gramineproject/gramine/875#-NAxQ4Mp7Kg7EGTtEFAY:-NAyO-qz9WokMN46ZSea:b-fpy7sm) ([raw file](https://github.com/gramineproject/gramine/blob/35b19b29c88760e08023be4af1956adcc64502be/CI-Examples/ra-tls-secret-prov/Makefile#L54)):*

FYI: I kept this `shell pkg-config --cflags mbedtls_gramine` because one day it will work hopefully...
</blockquote></details>

But what for? This example doesn't use mbedtls

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/875)* status: 0 of 2 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "squash! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`-- commits` line 5 at r1](https://reviewable.io/reviews/gramineproject/gramine/875#-NAyM2E28cVFflf6PhvT):*

This command doesn't link, it provides arguments to the linker.
</blockquote></details>

Done.
___
*[`-- commits` line 9 at r1](https://reviewable.io/reviews/gramineproject/gramine/875#-NAyMFjx-BRUlrb5aMYs):*

I think the fact that this is done only by some older linker is important in the context of how we fix it
</blockquote></details>

Done.
___
*[`-- commits` line 10 at r1](https://reviewable.io/reviews/gramineproject/gramine/875#-NAyMRfo73r21QNIITXo):*

`LD_LIBRARY_PATH` is not ignored, it just has lower priority and it so happens that `libc.so` is found at directory pointed at `DR_RPATH`, which takes precedence.
</blockquote></details>

Done.
___
*[`-- commits` line 11 at r1](https://reviewable.io/reviews/gramineproject/gramine/875#-NAyL-650HmLHYm3m2uB):*

These paths are:
1) host specific (e.g. different distros will have different paths), whereas the problem is not,
2) can be different, depending on where you have stuff installed,
3) `/lib/libc.so.6` is a in-Gramine path, which we've chosen to use in our examples. It's not the host path (which is what you are talking about here).

Maybe just remove the paths and stick to `host libc` vs `Gramine provided libc`? Or at least use host paths and take the default installation paths from our apt package. 
</blockquote></details>

Done.
___
*[`-- commits` line 14 at r1](https://reviewable.io/reviews/gramineproject/gramine/875#-NAyK6KoDOzGqtGdRDIv):*

Is it worth having this error message here?
The problem I see is that this is only one of the things that can  go wrong / one of the errors. Simple crash (segmentation fault) is a likely outcome as well (this is what I always encountered when mixed ld vs libc). This one just happens to be printed on the specific combination of ld.so we currently use and current version of glibc from Ubuntu 20.04.

Not blocking though.
</blockquote></details>

Done. I still think it's useful to have this `Inconsistency detected ...` message here, so that we can grep for this in git commit history. But I added "segfaults" as well.
___
*[`-- commits` line 18 at r1](https://reviewable.io/reviews/gramineproject/gramine/875#-NAyN2Ft2p9SPIDu_c0E):*

> all linkers

Isn't this flag GNU ld specific?
</blockquote></details>

Well, all linkers I know of (old ld, new Gold, LLVM ld `lld`) have this flag. What other linkers you're thinking of?
___
*[`CI-Examples/ra-tls-mbedtls/Makefile` line 48 at r1](https://reviewable.io/reviews/gramineproject/gramine/875#-NAyNHwq3-h451KaixJy:-NBBkoVbCCJ0nsaf5REk:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/f1e1d22d29a08c05a6cf19188aa0241787608371/CI-Examples/ra-tls-mbedtls/Makefile#L48)):*

Could you add some TODO here? I don't understand the old one (it was outdated?), but it would be good to have something here.
</blockquote></details>

Done.
___
*[`CI-Examples/ra-tls-mbedtls/Makefile` line 74 at r1](https://reviewable.io/reviews/gramineproject/gramine/875#-NAxPwVVC7WkUbmjRZmy:-NBBkk8IC5FqnloZTLHE:b-839ez8) ([raw file](https://github.com/gramineproject/gramine/blob/35b19b29c88760e08023be4af1956adcc64502be/CI-Examples/ra-tls-mbedtls/Makefile#L74)):*

Maybe change this into a TODO? I think these two includes are mutually exclusive
</blockquote></details>

Done. Yes, you are right, they are mutually exclusive.
___
*[`CI-Examples/ra-tls-secret-prov/Makefile` line 54 at r1](https://reviewable.io/reviews/gramineproject/gramine/875#-NAxQ4Mp7Kg7EGTtEFAY:-NBBmbTzDuaXHFj4MJwN:bvv9l27) ([raw file](https://github.com/gramineproject/gramine/blob/35b19b29c88760e08023be4af1956adcc64502be/CI-Examples/ra-tls-secret-prov/Makefile#L54)):*

But what for? This example doesn't use mbedtls
</blockquote></details>

Done. You are right. This is from those times when mbedTLS was linked dynamically in SecretProv lib. Now mbedTLS is linked statically, so we don't need to care. Removed.

But notice that we still need to get the Secret Prov header (`#include "secret_prov.h"`) from somewhere. We should actually introduce `ratls_gramine` and `secretprov_gramine` packages, I guess.
___
*[`CI-Examples/ra-tls-secret-prov/Makefile` line 55 at r1](https://reviewable.io/reviews/gramineproject/gramine/875#-NBBoHq2CIaVkztvSiTO:-NBBoHq2CIaVkztvSiTP:btxp6na) ([raw file](https://github.com/gramineproject/gramine/blob/35b19b29c88760e08023be4af1956adcc64502be/CI-Examples/ra-tls-secret-prov/Makefile#L55)):*
> ```Makefile
> CFLAGS += -Wall -std=c11 -I$(GRAMINEDIR)/tools/sgx/ra-tls \
>           $(shell pkg-config --cflags mbedtls_gramine)
> LDFLAGS += -Wl,--enable-new-dtags $(shell pkg-config --libs mbedtls_gramine)
> ```

FYI: This `mbedtls_gramine` package is misleading. It should have been `secretprov_gramine`. But we get away with this because all Gramine-specific shared libraries are put in the same directory (e.g. `-L/home/dimakuv/gramineproject/gramine/built-debug/lib/x86_64-linux-gnu`), so this "hacky" invocation works.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/875)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "squash! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`-- commits` line 14 at r1](https://reviewable.io/reviews/gramineproject/gramine/875#-NAyK6KoDOzGqtGdRDIv):*

Done. I still think it's useful to have this `Inconsistency detected ...` message here, so that we can grep for this in git commit history. But I added "segfaults" as well.
</blockquote></details>

Doubt you will ever need to grep for it. Next time we hit a similar issue, it will be different message...
___
*[`-- commits` line 18 at r1](https://reviewable.io/reviews/gramineproject/gramine/875#-NAyN2Ft2p9SPIDu_c0E):*

Well, all linkers I know of (old ld, new Gold, LLVM ld `lld`) have this flag. What other linkers you're thinking of?
</blockquote></details>

IDK, just asking, since you wrote "all linkers". Seems this flag is GNU specific, but as with a lof of others, is also supported by `lld`. Which is all we care about, I guess.
___
*[`CI-Examples/ra-tls-secret-prov/Makefile` line 54 at r1](https://reviewable.io/reviews/gramineproject/gramine/875#-NAxQ4Mp7Kg7EGTtEFAY:-NBCYiOnA36NvYtinFPA:b-2rimeh) ([raw file](https://github.com/gramineproject/gramine/blob/35b19b29c88760e08023be4af1956adcc64502be/CI-Examples/ra-tls-secret-prov/Makefile#L54)):*

Done. You are right. This is from those times when mbedTLS was linked dynamically in SecretProv lib. Now mbedTLS is linked statically, so we don't need to care. Removed.

But notice that we still need to get the Secret Prov header (`#include "secret_prov.h"`) from somewhere. We should actually introduce `ratls_gramine` and `secretprov_gramine` packages, I guess.
</blockquote></details>

`secretprov_gramine` is just an example of RALTS usage, isn't it? So I guess no, we don't need such package. But `ratls_gramine` is a set of libraries we provide for convenience, so maybe it would be good to have it as a package (even though it's just also an example).
___
*[`CI-Examples/ra-tls-secret-prov/Makefile` line 55 at r1](https://reviewable.io/reviews/gramineproject/gramine/875#-NBBoHq2CIaVkztvSiTO:-NBCZqJ11WFcLO9fc-M7:b-bcfesh) ([raw file](https://github.com/gramineproject/gramine/blob/35b19b29c88760e08023be4af1956adcc64502be/CI-Examples/ra-tls-secret-prov/Makefile#L55)):*

FYI: This `mbedtls_gramine` package is misleading. It should have been `secretprov_gramine`. But we get away with this because all Gramine-specific shared libraries are put in the same directory (e.g. `-L/home/dimakuv/gramineproject/gramine/built-debug/lib/x86_64-linux-gnu`), so this "hacky" invocation works.
</blockquote></details>

Wouldn't `sgx_util` be more suitable then?
___
*[`CI-Examples/ra-tls-secret-prov/Makefile` line 52 at r2](https://reviewable.io/reviews/gramineproject/gramine/875#-NBCaVzo-99wbFCPMzIG:-NBCaVzp1FCc3N1V0ous:bt4tvpq) ([raw file](https://github.com/gramineproject/gramine/blob/979cb2153a0cc11bdf0f9ae8e753bba7edb92680/CI-Examples/ra-tls-secret-prov/Makefile#L52)):*
> ```Makefile
> # installation. We also use `mbedtls_gramine` package because we don't have a secret prov one.
> # TODO: Install secretprov_gramine package during Gramine install, and use it in below
> #       CFLAGS/LDFLAGS lines (via `pkg-config {--cflags|--libs} secretprov_gramine`).
> ```

1. We install such lib. The problem is it doesn't have pkgconfig file
2. Actually, why this is even a lib we install??? Isn't it an example usage of RA-TLS?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/875)* status: 1 of 2 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[`-- commits` line 14 at r1](https://reviewable.io/reviews/gramineproject/gramine/875#-NAyK6KoDOzGqtGdRDIv):*

Doubt you will ever need to grep for it. Next time we hit a similar issue, it will be different message...
</blockquote></details>

I guess I agree. Let me block myself on this, I'll fix the commit message during final rebase.

TODO(myself): remove the `Inconsistency detected` example message.
___
*[`CI-Examples/ra-tls-secret-prov/Makefile` line 54 at r1](https://reviewable.io/reviews/gramineproject/gramine/875#-NAxQ4Mp7Kg7EGTtEFAY:-NBCeoRE0Ahx3BKilmOh:boe8f9l) ([raw file](https://github.com/gramineproject/gramine/blob/35b19b29c88760e08023be4af1956adcc64502be/CI-Examples/ra-tls-secret-prov/Makefile#L54)):*

`secretprov_gramine` is just an example of RALTS usage, isn't it? So I guess no, we don't need such package. But `ratls_gramine` is a set of libraries we provide for convenience, so maybe it would be good to have it as a package (even though it's just also an example).
</blockquote></details>

No, Secret Prov is a library in itself. It builds on top of RA-TLS, but it is a normal library to use by people.

I would argue that we need three packages (or more specifically, pkgconfig files):
- `mbedtls_gramine`
- `ratls_gramine`
- `secretprov_gramine`
___
*[`CI-Examples/ra-tls-secret-prov/Makefile` line 55 at r1](https://reviewable.io/reviews/gramineproject/gramine/875#-NBBoHq2CIaVkztvSiTO:-NBCfdi1EydM2fR49Qss:bwbe17h) ([raw file](https://github.com/gramineproject/gramine/blob/35b19b29c88760e08023be4af1956adcc64502be/CI-Examples/ra-tls-secret-prov/Makefile#L55)):*

Wouldn't `sgx_util` be more suitable then?
</blockquote></details>

Done. Still a hack.
___
*[`CI-Examples/ra-tls-secret-prov/Makefile` line 52 at r2](https://reviewable.io/reviews/gramineproject/gramine/875#-NBCaVzo-99wbFCPMzIG:-NBCdmGEEfZuwirisfN2:b-bupbz2) ([raw file](https://github.com/gramineproject/gramine/blob/979cb2153a0cc11bdf0f9ae8e753bba7edb92680/CI-Examples/ra-tls-secret-prov/Makefile#L52)):*

1. We install such lib. The problem is it doesn't have pkgconfig file
2. Actually, why this is even a lib we install??? Isn't it an example usage of RA-TLS?
</blockquote></details>

1. Done, rephrased.
2. No, Secret Prov is a perfectly normal library. It is built on top of RA-TLS, but it can definitely be linked against by applications. It has sufficient functionality, and we describe and endorse it here: https://gramine.readthedocs.io/en/latest/attestation.html

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/875)* status: 1 of 2 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[`CI-Examples/ra-tls-secret-prov/Makefile` line 54 at r1](https://reviewable.io/reviews/gramineproject/gramine/875#-NAxQ4Mp7Kg7EGTtEFAY:-NBCmQPb1VSj4SndC8pL:bi7cfn0) ([raw file](https://github.com/gramineproject/gramine/blob/35b19b29c88760e08023be4af1956adcc64502be/CI-Examples/ra-tls-secret-prov/Makefile#L54)):*

I don't think the latter two should be parts of the core Gramine
</blockquote></details>

I disagree. Yes, they can be developed independently of Gramine, but they are sufficiently widely used to be considered a part of the core Gramine deployment/package.
___
*[`CI-Examples/ra-tls-secret-prov/Makefile` line 55 at r1](https://reviewable.io/reviews/gramineproject/gramine/875#-NBBoHq2CIaVkztvSiTO:-NBCmcU-ATHq7YJn-OsY:bbdbc1k) ([raw file](https://github.com/gramineproject/gramine/blob/35b19b29c88760e08023be4af1956adcc64502be/CI-Examples/ra-tls-secret-prov/Makefile#L55)):*

Another problem here is that we use these `LDFLAGS` for 5 different programs, each with different requirements. Some don't need any libs. Some do...
</blockquote></details>

Yep. @mkow is working on splitting this example in several smaller ones. Maybe he will also split this Makefile, or at least the `CFLAGS`/`LDFLAGS` requirements?
___
*[`CI-Examples/ra-tls-secret-prov/Makefile` line 50 at r3](https://reviewable.io/reviews/gramineproject/gramine/875#-NBChljG5RxRVJgxtz7l:-NBCmN8V4RTd1bvLBR5K:b3ylir8) ([raw file](https://github.com/gramineproject/gramine/blob/b0f6dc147366f53ea4d44a900855eed9a134f09f/CI-Examples/ra-tls-secret-prov/Makefile#L50)):*

It's not a package (in apt and such sense), it's a library.
</blockquote></details>

Changed to pkgconfig.

boryspoplawski (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/875)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/875)* status: all files reviewed, 1 unresolved discussion, "squash! " and "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

mkow (changes requested):
> Reviewed 1 of 2 files at r2, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/875)* status: all files reviewed, 2 unresolved discussions, "squash! " and "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`CI-Examples/ra-tls-secret-prov/Makefile` line 52 at r5](https://reviewable.io/reviews/gramineproject/gramine/875#-NBt8oLLBycoSEW4ofiM:-NBt8oLLBycoSEW4ofiN:b-8d7z) ([raw file](https://github.com/gramineproject/gramine/blob/8ccfac67b5fa4157853f4ac31dc2b58138479cf2/CI-Examples/ra-tls-secret-prov/Makefile#L52)):*
> ```
> pckconfig
> ```

pkg-config, I guess?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/875)* status: 1 of 2 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

___
*[`CI-Examples/ra-tls-secret-prov/Makefile` line 52 at r5](https://reviewable.io/reviews/gramineproject/gramine/875#-NBt8oLLBycoSEW4ofiM:-NBuo25P84RJ_lINuDmY:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/8ccfac67b5fa4157853f4ac31dc2b58138479cf2/CI-Examples/ra-tls-secret-prov/Makefile#L52)):*

> ```
> pckconfig
> ```

pkg-config, I guess?
</blockquote></details>

Done.

boryspoplawski (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/875)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/875)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/875)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

___
*[`-- commits` line 14 at r1](https://reviewable.io/reviews/gramineproject/gramine/875#-NAyK6KoDOzGqtGdRDIv):*

Well, this one is rather short, so I don't block if you leave this. Up to you
</blockquote></details>

Done

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/875)* status: :shipit: complete! all files reviewed, all discussions resolved

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/875)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #877, opened 2022-09-02
[PAL/Linux-SGX] Introduce `sgx.cpu_features.[...]` manifest options

dimakuv:
## Description of the changes 

Introduce `sgx.cpu_features.[...] = "[autodetect|required|disabled]"` instead of `sgx.require_[...] = true|false`.

Mapping of deprecated and new manifest options:
- `sgx.cpu_features.[...] = "autodetect"` -> `sgx.require_[...] = false`
- `sgx.cpu_features.[...] = "required"` -> `sgx.require_[...] = true`

`sgx.cpu_features.[...] = "disabled"` is new and it disables the corresponding CPU feature inside the SGX enclave even if this CPU
feature is available on the system: this may improve enclave performance because this CPU feature will *not* be saved and restored during enclave entry/exit.

This PR is a re-creation of https://github.com/gramineproject/gramine/pull/321 and https://github.com/gramineproject/gramine/pull/461, since it turned out to be useful for performance. Without disclosing any perf numbers, I performed a set of micro-benchmarks and they prove that disabling Intel AMX inside of the enclave (when it is available on the machine but is not needed by the app inside the enclave) gives a measurable performance boost (for the worst-case scenarios). So this change actually is beneficial.

## How to test this PR? 

I tested this PR manually on the AMX-enabled machine:
```bash
# with nothing, should default to sgx.cpu_features.amx = "autodetect"
$ gramine-sgx helloworld
debug: LibOS xsave_enabled 1, xsave_size 0x2b00(11008), xsave_features 0x600e7

# with sgx.cpu_features.amx = "autodetect"
$ gramine-sgx helloworld
debug: LibOS xsave_enabled 1, xsave_size 0x2b00(11008), xsave_features 0x600e7 

# with sgx.cpu_features.amx = "required"
$ gramine-sgx helloworld
debug: LibOS xsave_enabled 1, xsave_size 0x2b00(11008), xsave_features 0x600e7

# with sgx.cpu_features.amx = "disabled"
$ gramine-sgx helloworld
debug: LibOS xsave_enabled 1, xsave_size 0xa80(2688), xsave_features 0xe7
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/877)

PR #461, opened 2022-03-16
[Pal/Linux-SGX] Add `sgx.disable_[cpu-feature]` manifest options

dimakuv:
## Description of the changes 

This PR adds three new manifest options: `sgx.disable_avx`, `sgx.disable_avx512`, `sgx.disable_amx`. Setting each of these options to `true` disables the corresponding CPU feature inside the SGX enclave even if this CPU feature is available on the system: this may improve enclave performance because this CPU feature will *not* be saved and restored during enclave entry/exit.

This PR is a re-creation of https://github.com/gramineproject/gramine/pull/321, since it turned out to be useful for performance. Without disclosing any perf numbers, I performed a set of micro-benchmarks and they prove that disabling Intel AMX inside of the enclave (when it is available on the machine but is not needed by the app inside the enclave) gives a measurable performance boost (for the worst-case scenarios). So this change actually is beneficial.

## How to test this PR? 

I tested this PR manually on the AMX-enabled machine:
```bash
# with   sgx.require_amx = false,   sgx.disable_amx = false 
$ gramine-sgx helloworld
[::] debug: LibOS xsave_enabled 1, xsave_size 0x2b00(11008), xsave_features 0x600e7 

# with   sgx.require_amx = true,   sgx.disable_amx = false 
$ gramine-sgx helloworld
[::] debug: LibOS xsave_enabled 1, xsave_size 0x2b00(11008), xsave_features 0x600e7

# with   sgx.require_amx = false,   sgx.disable_amx = true
$ gramine-sgx helloworld
[::] debug: LibOS xsave_enabled 1, xsave_size 0xa80(2688), xsave_features 0xe7

# with   sgx.require_amx = true,   sgx.disable_amx = true   -> this one is not recommended
$ gramine-sgx helloworld
[::] debug: LibOS xsave_enabled 1, xsave_size 0xa80(2688), xsave_features 0xe7
```

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/461)

PR #321, opened 2022-01-06
[Pal/Linux-SGX] Add `sgx.disable_[cpu-feature]` manifest options

dimakuv:
## Description of the changes 

This commit adds three new manifest options: `sgx.disable_avx`, `sgx.disable_avx512`, `sgx.disable_amx`. Setting each of these options to `true` disables the corresponding CPU feature inside the SGX enclave even if this CPU feature is available on the system: this may improve enclave performance because this CPU feature will *not* be saved and restored during enclave entry/exit.

~~For example, disabling Intel AMX may improve performance of some workloads by around 5%.~~ UPDATE: The previous statement turned out to be false. It seems that this PR doesn't provide any perf improvement. I am closing it and may resurrect in the future if this PR turns out to be useful for anything.

## How to test this PR? 

I'll test it manually on some machines (to make sure it actually improves performance).

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/321)

PR #881, opened 2022-09-05
[PAL/Linux-SGX] Introduce `sgx.cpu_features.[...]` manifest options

dimakuv:
## Description of the changes 

This PR introduces `sgx.cpu_features.[...] = "[unspecified|required|disabled]"` instead of `sgx.require_[...] = true|false`.

Mapping of deprecated and new manifest options:
- `sgx.cpu_features.[...] = "unspecified"` -> `sgx.require_[...] = false`
- `sgx.cpu_features.[...] = "required"` -> `sgx.require_[...] = true`

`sgx.cpu_features.[...] = "disabled"` is new and it disables the corresponding CPU feature inside the SGX enclave even if this CPU feature is available on the system: this may improve enclave performance because this CPU feature will *not* be saved and restored during enclave entry/exit.

As a side effect, this PR removes unneeded attrs in `get_mrenclave_and_manifest()`. The only enclave attributes required for the SGX enclave measurement are the enclave size and the number of enclave threads. Other attributes such as `ISV_PROD_ID`, `ISV_SVN`, `XFRM` are not included in the measurement.

This PR is a re-creation of https://github.com/gramineproject/gramine/pull/321 and https://github.com/gramineproject/gramine/pull/461 and https://github.com/gramineproject/gramine/pull/877, since it turned out to be useful for performance. Without disclosing any perf numbers, I performed a set of micro-benchmarks and they prove that disabling Intel AMX inside of the enclave (when it is available on the machine but is not needed by the app inside the enclave) gives a measurable performance boost (for the worst-case scenarios). So this change actually is beneficial.

## How to test this PR? 

I tested this PR manually on the AMX-enabled machine:
```bash
# with nothing, should default to sgx.cpu_features.amx = "unspecified"
$ gramine-sgx helloworld
debug: LibOS xsave_enabled 1, xsave_size 0x2b00(11008), xsave_features 0x600e7

# with sgx.cpu_features.amx = "unspecified"
$ gramine-sgx helloworld
debug: LibOS xsave_enabled 1, xsave_size 0x2b00(11008), xsave_features 0x600e7 

# with sgx.cpu_features.amx = "required"
$ gramine-sgx helloworld
debug: LibOS xsave_enabled 1, xsave_size 0x2b00(11008), xsave_features 0x600e7

# with sgx.cpu_features.amx = "disabled"
$ gramine-sgx helloworld
debug: LibOS xsave_enabled 1, xsave_size 0xa80(2688), xsave_features 0xe7
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/881)

### DIFF
--- a/Documentation/manifest-syntax.rst
+++ b/Documentation/manifest-syntax.rst
@@ -775,28 +775,75 @@ untrusted memory in potentially insecure manner - susceptible to
 CVE-2022-21233 (INTEL-SA-00657) and CVE-2022-21166 (INTEL-SA-00615)
 respectively.
 
-Optional CPU features (AVX, AVX512, MPX, PKRU, AMX, EXINFO)
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+SGX EXINFO
+^^^^^^^^^^
 
 ::
 
-    sgx.require_avx    = [true|false]
-    sgx.require_avx512 = [true|false]
-    sgx.require_mpx    = [true|false]
-    sgx.require_pkru   = [true|false]
-    sgx.require_amx    = [true|false]
-    sgx.require_exinfo = [true|false]
+    sgx.use_exinfo = [true|false]
     (Default: false)
 
-This syntax ensures that the CPU features are available and enabled for the
-enclave. If the options are set in the manifest but the features are unavailable
-on the platform, enclave initialization will fail. If the options are unset,
-enclave initialization will succeed even if these features are unavailable on
-the platform.
+If ``sgx.use_exinfo`` is set, user application can retrieve faulting address in
+signal handler in case of a page fault. Otherwise (set to ``false``), the
+faulting address will always be provided as ``0``. The default is ``false``
+because some frameworks/runtimes could otherwise print the callstack and
+variables/registers on exceptions, potentially leaking data.
 
-If ``require_exinfo`` is set, user application can retrieve faulting address in
-signal handler in case of a page fault. Otherwise (set to ``false``),
-the faulting address will always be provided as ``0``.
+Optional CPU features (AVX, AVX512, AMX, MPX, PKRU)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+::
+
+    sgx.cpu_features.avx    = "[unspecified|disabled|required]"
+    sgx.cpu_features.avx512 = "[unspecified|disabled|required]"
+    sgx.cpu_features.amx    = "[unspecified|disabled|required]"
+    (Default: "unspecified")
+
+    sgx.cpu_features.mpx    = "[disabled|required]"
+    sgx.cpu_features.pkru   = "[disabled|required]"
+    (Default: "disabled")
+
+For the SGX threat model, Gramine divides CPU features in two subsets:
+not-security-hardening features (currently AVX, AVX512 and AMX) and
+security-hardening features (MPX and PKRU).
+
+The ``"unspecified"`` syntax applies only to not-security-hardening features. It
+means that the enclave initialization will succeed regardless of whether the CPU
+feature is available on the platform or not. The CPU features will be enabled in
+the enclave if they are available on the platform. Security-hardening features
+cannot be marked as ``"unspecified"`` because this would lead to different
+security properties of the application on different platforms.
+
+The ``"disabled"`` syntax disables the CPU feature inside the enclave even if
+this CPU feature is available on the platform. This may improve enclave
+performance because this CPU feature will *not* be saved and restored during
+enclave entry/exit. This syntax is provided to improve performance of
+applications that are known to *not* rely on certain CPU features. Be aware that
+if the application relies on some disabled CPU features, the application will
+fail with SIGILL ("illegal instruction"). For example, if the application is
+built with AVX support, and AVX is disabled in the manifest, the application
+will crash.
+
+The ``"required"`` syntax ensures that the CPU feature is available and enabled
+for the enclave. If such option is set in the manifest but the CPU feature is
+unavailable on the platform, enclave initialization will fail.
+
+When in doubt, it is recommended to keep the default values for these features.
+In this case, Gramine auto-detects the corresponding not-security-hardening CPU
+features on the platform and enables them if available, and disables
+security-hardening CPU features. This allows the SGX enclave to be executed on
+the widest range of platforms.
+
+At the technical level, the mapping from CPU features to SGX fields is as
+follows:
+
+- ``"unspecified"``: ``SIGSTRUCT.ATTRIBUTEMASK[feature] = 0``. The untrusted
+  loader of Gramine sets ``SECS.ATTRIBUTES[feature] = 0`` if it can't detect the
+  feature on the platform, and ``SECS.ATTRIBUTES[feature] = 1`` otherwise.
+- ``"required"``: ``SIGSTRUCT.ATTRIBUTEMASK[feature] = 1`` and
+  ``SIGSTRUCT.ATTRIBUTES[feature] = 1``.
+- ``"disabled"``: ``SIGSTRUCT.ATTRIBUTEMASK[feature] = 1`` and
+  ``SIGSTRUCT.ATTRIBUTES[feature] = 0``.
 
 ISV Product ID and SVN
 ^^^^^^^^^^^^^^^^^^^^^^
@@ -1134,3 +1181,47 @@ In addition, the application manifest must also contain ``sgx.debug = true``.
    independently.
 
 See :ref:`vtune-sgx-profiling` for more information.
+
+Deprecated options
+------------------
+
+Optional CPU features (deprecated syntax)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+::
+
+    sgx.require_avx    = [true|false]
+    sgx.require_avx512 = [true|false]
+    sgx.require_mpx    = [true|false]
+    sgx.require_pkru   = [true|false]
+    sgx.require_amx    = [true|false]
+
+This syntax specified whether to require certain CPU features to be available on
+the platform where the enclave executes. This syntax has been replaced with
+``sgx.cpu_features.[avx|avx512|mpx|pkru|amx]``.
+
+Previously, the default value for these features was ``false``. This was
+ambiguous because e.g. ``sgx.require_avx = false`` meant that the AVX feature
+was auto-detected on the platform, whereas ``sgx.require_mpx = false`` meant
+that the MPX feature was always disabled inside the enclave.
+
+.. note ::
+   Previously, security-hardening features (MPX and PKRU) were underspecified.
+   For example, ``sgx.require_mpx = false`` meant that the MPX feature was
+   always disabled inside the enclave (regardless of whether the CPU supported
+   it or not), and ``sgx.require_mpx = true`` meant that the MPX feature was
+   always enabled inside the enclave (even if the CPU did not support it, which
+   would result in enclave failures). This is in contrast to
+   non-security-hardening features (AVX, AVX512, AMX) which could be
+   "unspecified". The new manifest syntax fixes this ambiguity.
+
+SGX EXINFO (deprecated syntax)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+::
+
+    sgx.require_exinfo = [true|false]
+
+This syntax specified whether a user application can retrieve faulting address
+in signal handler in case of a page fault. This syntax was renamed to
+``sgx.use_exinfo``. The default value was ``false``.
--- a/Documentation/performance.rst
+++ b/Documentation/performance.rst
@@ -216,26 +216,33 @@ untrusted RPC threads on another set of cores (e.g., on second hyper-threads).
 In general, the classical performance-tuning strategies are applicable for
 Gramine and Exitless multi-threaded workloads.
 
-Optional CPU features (AVX, AVX512, MPX, PKRU, AMX)
+Optional CPU features (AVX, AVX512, AMX, MPX, PKRU)
 ---------------------------------------------------
 
 SGX technology allows to specify which CPU features are required to run the SGX
-enclave. Gramine "inherits" this and has the following manifest options:
-``sgx.require_avx``, ``sgx.require_avx512``, ``sgx.require_mpx``,
-``sgx.require_pkru``, ``sgx.require_amx``. By default, all of them are set to
-``false`` – this means that SGX hardware will allow running the SGX enclave on
-any system, whether the system has AVX/AVX512/MPX/PKRU/AMX features or not.
+enclave. Gramine "inherits" this and has the manifest options for
+AVX/AVX512/AMX/MPX/PKRU CPU features under ``sgx.cpu_features``. By default, all
+not-security-hardening CPU features (AVX, AVX512, AMX) are set to
+``"unspecified"`` -- this means that Gramine will allow running the SGX enclave
+on any platform, whether the platform has the CPU features or not.
 
 Gramine typically correctly identifies the features of the underlying platform
-and propagates the information on AVX/AVX512/MPX/PKRU/AMX inside the enclave and
-to the application. It is recommended to leave these manifest options as-is (set
-to ``false``). However, we observed on some platforms that the graminized
+and propagates the information on AVX/AVX512/AMX inside the enclave and to the
+application. It is recommended to leave these manifest options as-is (set to
+``"unspecified"``). However, we observed on some platforms that the graminized
 application cannot detect these features and falls back to a slow
 implementation. For example, some crypto libraries do not recognize AVX on the
 platform and use very slow functions, leading to 10-100x overhead over native
 (we still don't know the reason for this behavior). If you suspect this can be
-your case, enable the features in the manifest, e.g., set ``sgx.require_avx =
-true``.
+your case, enable the features in the manifest, e.g., set ``sgx.cpu_features.avx
+= "required"``.
+
+Gramine also allows to explicitly disable CPU features using the ``"disabled"``
+keyword -- this disables the corresponding CPU feature inside the SGX enclave
+even if this CPU feature is available on the system. This may improve enclave
+performance because this CPU feature will *not* be saved and restored during
+enclave entry/exit. But be aware that if the graminized application relies on
+this CPU feature, the application will crash with "illegal instruction".
 
 For more information on SGX logic regarding optional CPU features, see the Intel
 Software Developer Manual, Table 38-3 ("Layout of ATTRIBUTES Structure") under
--- a/pal/src/host/linux-sgx/sgx_arch.h
+++ b/pal/src/host/linux-sgx/sgx_arch.h
@@ -98,6 +98,11 @@ typedef uint8_t sgx_isvfamily_id_t[SGX_ISV_FAMILY_ID_SIZE];
 /* EINIT must verify most of the SECS.ATTRIBUTES[127..64] bits (XFRM/XCR0 bits) against
  * SIGSTRUCT.ATTRIBUTES[127..64].
  *
+ * This default XFRM mask may be modified via the manifest options `sgx.cpu_features.[feature]`. If
+ * the manifest option for some feature is set to "required" or "disabled", then the corresponding
+ * bits in the XFRM mask are set. If the manifest option is set to "unspecified", then the
+ * corresponding bits are unset.
+ *
  * Notes:
  *   - Verified bits include: bit 0 + bit 1 (X87 + SSE, always enabled in SGX), bit 3 + bit 4
  *     (BNDREG + BNDCSR, enables Intel MPX), bit 9 (PKRU, enables Intel MPK), and all reserved bits.
--- a/python/graminelibos/manifest.py
+++ b/python/graminelibos/manifest.py
@@ -93,12 +93,6 @@ class Manifest:
         sgx.setdefault('isvsvn', 0)
         sgx.setdefault('remote_attestation', "none")
         sgx.setdefault('debug', False)
-        sgx.setdefault('require_avx', False)
-        sgx.setdefault('require_avx512', False)
-        sgx.setdefault('require_mpx', False)
-        sgx.setdefault('require_pkru', False)
-        sgx.setdefault('require_amx', False)
-        sgx.setdefault('require_exinfo', False)
         sgx.setdefault('enable_stats', False)
         sgx.setdefault('edmm_enable', False)
 
@@ -106,6 +100,22 @@ class Manifest:
             sgx.setdefault('enclave_size', DEFAULT_ENCLAVE_SIZE_WITH_EDMM)
         else:
             sgx.setdefault('enclave_size', DEFAULT_ENCLAVE_SIZE_NO_EDMM)
+
+        # TODO: below was deprecated in release v1.6, remove this check in v1.7
+        #       (but keep the `if` body)
+        if not 'require_exinfo' in sgx:
+            sgx.setdefault('use_exinfo', False)
+
+        # TODO: below were deprecated in release v1.6, remove this check in v1.7
+        #       (but keep the `if` body)
+        deprecated = ['require_avx', 'require_avx512', 'require_amx', 'require_mpx', 'require_pkru']
+        if not any(key in sgx for key in deprecated):
+            sgx_cpu_features = sgx.setdefault('cpu_features', {})
+            sgx_cpu_features.setdefault('avx', "unspecified")
+            sgx_cpu_features.setdefault('avx512', "unspecified")
+            sgx_cpu_features.setdefault('amx', "unspecified")
+            sgx_cpu_features.setdefault('mpx', "disabled")
+            sgx_cpu_features.setdefault('pkru', "disabled")
 
         if not isinstance(sgx['trusted_files'], list):
             raise ValueError("Unsupported trusted files syntax, more info: " +
--- a/python/graminelibos/sgx_sign.py
+++ b/python/graminelibos/sgx_sign.py
@@ -77,36 +77,78 @@ def parse_size(value):
 def collect_bits(manifest_sgx, options_dict):
     val = 0
     for opt, bits in options_dict.items():
-        if manifest_sgx[opt] == 1:
+        if manifest_sgx.get(opt) is True:
             val |= bits
     return val
+
+
+def collect_cpu_feature_bits(manifest_cpu_features, options_dict, val, mask, security_hardening):
+    for opt, bits in options_dict.items():
+        if opt not in manifest_cpu_features:
+            continue
+        if manifest_cpu_features[opt] == "required":
+            val |= bits
+            mask |= bits
+        elif manifest_cpu_features[opt] == "disabled":
+            val &= ~bits
+            mask |= bits
+        elif security_hardening or manifest_cpu_features[opt] != "unspecified":
+            raise KeyError(f'Manifest option `sgx.cpu_features.{opt}` has a disallowed value')
+    return val, mask
 
 
 def get_enclave_attributes(manifest_sgx):
     flags_dict = {
         'debug': offs.SGX_FLAGS_DEBUG,
     }
+    flags = collect_bits(manifest_sgx, flags_dict)
+    if ARCHITECTURE == 'amd64':
+        flags |= offs.SGX_FLAGS_MODE64BIT
 
-    xfrms_dict = {
+    # TODO: 'require_exinfo' was deprecated in release v1.6, should be removed in v1.7
+    if 'require_exinfo' in manifest_sgx:
+        if 'use_exinfo' in manifest_sgx:
+            raise KeyError(f'`sgx.require_exinfo` cannot coexist with `sgx.use_exinfo`')
+        manifest_sgx['use_exinfo'] = manifest_sgx.pop('require_exinfo')
+
+    miscs_dict = {
+        'use_exinfo': offs.SGX_MISCSELECT_EXINFO,
+    }
+    miscs = collect_bits(manifest_sgx, miscs_dict)
+
+    # TODO: these were deprecated in release v1.6, so they should be removed in v1.7
+    deprecated_xfrms_dict = {
         'require_avx': offs.SGX_XFRM_AVX,
         'require_avx512': offs.SGX_XFRM_AVX512,
         'require_mpx': offs.SGX_XFRM_MPX,
         'require_pkru': offs.SGX_XFRM_PKRU,
         'require_amx': offs.SGX_XFRM_AMX,
     }
-
-    miscs_dict = {
-        'require_exinfo': offs.SGX_MISCSELECT_EXINFO,
+    xfrms_dict = {
+        'avx': offs.SGX_XFRM_AVX,
+        'avx512': offs.SGX_XFRM_AVX512,
+        'amx': offs.SGX_XFRM_AMX,
+    }
+    secure_xfrms_dict = {
+        'mpx': offs.SGX_XFRM_MPX,
+        'pkru': offs.SGX_XFRM_PKRU,
     }
 
-    flags = collect_bits(manifest_sgx, flags_dict)
-    if ARCHITECTURE == 'amd64':
-        flags |= offs.SGX_FLAGS_MODE64BIT
+    xfrms, xfrms_mask = offs.SGX_XFRM_LEGACY, offs.SGX_XFRM_MASK_CONST
+    if manifest_sgx.get('cpu_features') is None:
+        # collect deprecated `sgx.require_xxx` options; remove this in v1.7
+        xfrms |= collect_bits(manifest_sgx, deprecated_xfrms_dict)
+    else:
+        for deprecated_key in deprecated_xfrms_dict:
+            if deprecated_key in manifest_sgx:
+                raise KeyError(f'`sgx.cpu_features` cannot coexist with `sgx.{deprecated_key}`')
+        xfrms, xfrms_mask = collect_cpu_feature_bits(manifest_sgx['cpu_features'], xfrms_dict,
+                                                     xfrms, xfrms_mask, security_hardening=False)
+        xfrms, xfrms_mask = collect_cpu_feature_bits(manifest_sgx['cpu_features'],
+                                                     secure_xfrms_dict, xfrms, xfrms_mask,
+                                                     security_hardening=True)
 
-    xfrms = offs.SGX_XFRM_LEGACY | collect_bits(manifest_sgx, xfrms_dict)
-    miscs = collect_bits(manifest_sgx, miscs_dict)
-
-    return flags, xfrms, miscs
+    return flags, miscs, xfrms, xfrms_mask
 
 
 # Populate Enclave Memory
@@ -527,10 +569,11 @@ def get_tbssigstruct(manifest_path, date, libpal=SGX_LIBPAL, verbose=False):
     sig['isv_prod_id'] = manifest_sgx['isvprodid']
     sig['isv_svn'] = manifest_sgx['isvsvn']
 
-    attribute_flags, attribute_xfrms, misc_select = get_enclave_attributes(manifest_sgx)
+    attribute_flags, misc_select, attribute_xfrms, xfrms_mask = get_enclave_attributes(manifest_sgx)
     sig['attribute_flags'] = attribute_flags
-    sig['attribute_xfrms'] = attribute_xfrms
     sig['misc_select'] = misc_select
+    sig['attribute_xfrms'] = attribute_xfrms
+    sig['attribute_xfrm_mask'] = xfrms_mask
 
     return sig
 


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/881)* status: 0 of 6 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

___
*[`pal/src/host/linux-sgx/sgx_arch.h` line 102 at r1](https://reviewable.io/reviews/gramineproject/gramine/881#-NBCaaZF4Ix9OcYM0xYq:-NBCaaZGBmC_mRrF9_Ka:buwsxhy) ([raw file](https://github.com/gramineproject/gramine/blob/0c10c1436aa347a74d8308a29bc843f15f61b949/pal/src/host/linux-sgx/sgx_arch.h#L102)):*
> ```objc
>  * the manifest option for some feature is set to "required", then the corresponding bits in the
>  * XFRM mask are set. If the manifest option is set to "disabled", then the corresponding bits are
>  * unset.
> ```

FYI: another way of thinking about `sgx.cpu_features.[feature] = "autodetect"` is like this:
- the default XFRM mask stays;
- for not-security-critical CPU features, the default XFRM mask bits are `0`, so the untrusted Gramine loader can choose to enable the CPU feature;
- for security-critical CPU features, the default XFRM mask bits are `1`, so the untrusted Gramine loader can *not* enable the CPU feature, and it just stays disabled.

So there is an interesting logic for security-critical CPU features: they are initially definitely-disabled. The only way to set them is to do `sgx.cpu_features.[feature] = "required"`, so in this case the CPU feature is definitely-enabled. Why so strict? Because we have two contradicting requirements: (a) we want to run our enclave on as many platforms as possible, (b) we must be sure that the platforms do not just enable/disable such CPU features as they wish. So security-critical CPU features are definitely-disabled by default (to allow for this enclave to run on more platforms), and they can only be definitely-enabled.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/881)* status: 0 of 6 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/881#-NBClLee4Q_7mpeK40DZ:-NBClLee4Q_7mpeK40D_:b-f44vbb) (no related file):*
FYI: Based on an offline chat with @boryspoplawski, I did some changes to the second (main) commit and force-pushed, since nobody yet reviewed this. The changes were:
- Replace `security-critical` with `security-hardening` (otherwise "security-critical CPU features can be disabled" sounds bizzare)
- Replace `autodetect` with `unspecified`
- Allow `autodetect` only for not-security-critical CPU features

___

donporter (approved):
> Reviewed 3 of 6 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/881)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`Documentation/manifest-syntax.rst` line 575 at r2](https://reviewable.io/reviews/gramineproject/gramine/881#-NL6KmxPAF66ZOHg5Vpx:-NL6KmxPAF66ZOHg5Vpy:bhskkhk) ([raw file](https://github.com/gramineproject/gramine/blob/d9feaae81024618610572d15577c67a4734387d3/Documentation/manifest-syntax.rst#L575)):*
> ```reStructuredText
> unavailable on the platform, enclave initialization will fail.
> 
> In case of doubt, it is recommended to keep the default values for these
> ```

A more common English phrasing might be: "When in doubt", but I know what you mean.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/881)* status: all files reviewed, 8 unresolved discussions (waiting on @dimakuv)

___
*[`-- commits` line 22 at r2](https://reviewable.io/reviews/gramineproject/gramine/881#-NP16YW19KRR6SmdSzja):*
We default not-security-harening options to (from `unspecified`) `disabled` now. Does this mean we changed the default behavior? If so, should note it down somewhere.



_Code quote:_
```
  - `sgx.cpu_features.[...] = "unspecified"` -> `sgx.require_[...] = false`
  - `sgx.cpu_features.[...] = "required"` -> `sgx.require_[...] = true`

  `sgx.cpu_features.[...] = "disabled"` is new and it disables the
```
___
*[`Documentation/manifest-syntax.rst` line 556 at r2](https://reviewable.io/reviews/gramineproject/gramine/881#-NP1HyIeD-43AwIDll9Z:-NP1HyIeD-43AwIDll9_:b-y9crew) ([raw file](https://github.com/gramineproject/gramine/blob/d9feaae81024618610572d15577c67a4734387d3/Documentation/manifest-syntax.rst#L556)):*
> ```reStructuredText
> security-hardening features (MPX and PKRU).
> 
> The ``"unspecified"`` syntax applies only to not-security-hardening features. It
> ```

Where do we validate this in this code?



_Code quote:_
```reStructuredText
The ``"unspecified"`` syntax applies only to not-security-hardening features
```
___
*[`Documentation/manifest-syntax.rst` line 999 at r2](https://reviewable.io/reviews/gramineproject/gramine/881#-NP100eK8ldJcb-5rLNQ:-NP100eK8ldJcb-5rLNR:b-wvwtvu) ([raw file](https://github.com/gramineproject/gramine/blob/d9feaae81024618610572d15577c67a4734387d3/Documentation/manifest-syntax.rst#L999)):*
> ```reStructuredText
>     sgx.require_mpx    = [true|false]
>     sgx.require_pkru   = [true|false]
>     sgx.require_amx    = [true|false]
> ```

Default is not specified any more. So what the users should expect exactly?
___
*[`pal/src/host/linux-sgx/sgx_arch.h` line 102 at r2](https://reviewable.io/reviews/gramineproject/gramine/881#-NP18cX5B0utAPOkpT9K:-NP18cX5B0utAPOkpT9L:b3dhtfi) ([raw file](https://github.com/gramineproject/gramine/blob/d9feaae81024618610572d15577c67a4734387d3/pal/src/host/linux-sgx/sgx_arch.h#L102)):*
> - ``"disabled"``: ``SIGSTRUCT.ATTRIBUTEMASK[feature] = 1`` and
  ``SIGSTRUCT.ATTRIBUTES[feature] = 0``.

Is the corresponding bit unset?



_Code quote:_
```objc
 * XFRM mask are set. If the manifest option is set to "disabled", then the corresponding bits are
 * unset.
```
___
*[`python/graminelibos/manifest.py` line 100 at r2](https://reviewable.io/reviews/gramineproject/gramine/881#-NP0x4KF0xSVEQAEioSw:-NP0x4KF0xSVEQAEioSx:bimng2h) ([raw file](https://github.com/gramineproject/gramine/blob/d9feaae81024618610572d15577c67a4734387d3/python/graminelibos/manifest.py#L100)):*
> ```Python
>         sgx_cpu_features = sgx.setdefault('cpu_features', {})
>         sgx_cpu_features.setdefault('mpx', "disabled")
>         sgx_cpu_features.setdefault('pkru', "disabled")
> ```

As we removed the defaults of the legacy options, this changes the default behavior, right?



_Code quote:_
```Python
        sgx_cpu_features.setdefault('mpx', "disabled")
        sgx_cpu_features.setdefault('pkru', "disabled")
```
___
*[`python/graminelibos/sgx_sign.py` line 76 at r2](https://reviewable.io/reviews/gramineproject/gramine/881#-NP0wHbM3b5cS5Bp-O1A:-NP0wHbM3b5cS5Bp-O1B:bhpms1f) ([raw file](https://github.com/gramineproject/gramine/blob/d9feaae81024618610572d15577c67a4734387d3/python/graminelibos/sgx_sign.py#L76)):*
> ```Python
>     val = 0
>     for opt, bits in options_dict.items():
>         if manifest_sgx.get(opt, 0) == 1:
> ```

This won't generate a KeyError anymore. I know this works for below but is this sth. we intended for?



_Code quote:_
```Python
get(opt, 0)
```
___
*[`python/graminelibos/sgx_sign.py` line 92 at r2](https://reviewable.io/reviews/gramineproject/gramine/881#-NP1EBJXEtieRyuuvmL9:-NP1EBJXEtieRyuuvmLA:bw7xkto) ([raw file](https://github.com/gramineproject/gramine/blob/d9feaae81024618610572d15577c67a4734387d3/python/graminelibos/sgx_sign.py#L92)):*
> ```Python
>         elif manifest_sgx['cpu_features'].get(opt, "") == "disabled":
>             val &= ~bits
>             mask |= bits
> ```

Users can input things like `sgx.cpu_features.amx = "abcd"` (same semantics as `unspecified`)?
___
*[`python/graminelibos/sgx_sign.py` line 126 at r2](https://reviewable.io/reviews/gramineproject/gramine/881#-NP0xISlBGez94l-KXII:-NP0xISlBGez94l-KXIJ:b-p7sfwo) ([raw file](https://github.com/gramineproject/gramine/blob/d9feaae81024618610572d15577c67a4734387d3/python/graminelibos/sgx_sign.py#L126)):*
> ```Python
>         'require_amx': offs.SGX_XFRM_AMX,
>     }
>     xfrms |= collect_bits(manifest_sgx, deprecated_xfrms_dict)
> ```

So we allow the co-existance and discrepancies between the legacy and new options?



_Code quote:_
```Python
xfrms |= collect_bits(manifest_sgx, deprecated_xfrms_dict)
```

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/881)* status: 2 of 6 files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @donporter and @kailun-qin)

___
*[`-- commits` line 22 at r2](https://reviewable.io/reviews/gramineproject/gramine/881#-NP16YW19KRR6SmdSzja):*
> We default not-security-harening options to (from `unspecified`) `disabled` now. 

Is this a typo in your sentence? Did you mean **We default security-hardening options...**?

But you are right, there was a complex ambiguity in the previous syntax and the way it was described in the documentation. I added a note in the docs. ( The state of this feature was really poor :( )
___
*[`Documentation/manifest-syntax.rst` line 556 at r2](https://reviewable.io/reviews/gramineproject/gramine/881#-NP1HyIeD-43AwIDll9Z:-NP20m9n1hsFPTJKQK3o:b-376vyb) ([raw file](https://github.com/gramineproject/gramine/blob/d9feaae81024618610572d15577c67a4734387d3/Documentation/manifest-syntax.rst#L556)):*

Where do we validate this in this code?
</blockquote></details>

Done. Good catch.
___
*[`Documentation/manifest-syntax.rst` line 575 at r2](https://reviewable.io/reviews/gramineproject/gramine/881#-NL6KmxPAF66ZOHg5Vpx:-NP1rKSI4rdxXtL8ZXHc:bb6un1j) ([raw file](https://github.com/gramineproject/gramine/blob/d9feaae81024618610572d15577c67a4734387d3/Documentation/manifest-syntax.rst#L575)):*
<details><summary><i>Previously, donporter (Don Porter) wrote…</i></summary><blockquote>

A more common English phrasing might be: "When in doubt", but I know what you mean.
</blockquote></details>

Done
___
*[`Documentation/manifest-syntax.rst` line 999 at r2](https://reviewable.io/reviews/gramineproject/gramine/881#-NP100eK8ldJcb-5rLNQ:-NP1rQks8Mw-mxRKEELM:b2on52e) ([raw file](https://github.com/gramineproject/gramine/blob/d9feaae81024618610572d15577c67a4734387d3/Documentation/manifest-syntax.rst#L999)):*

Default is not specified any more. So what the users should expect exactly?
</blockquote></details>

Not sure I understand your question.

If users don't specify these deprecated CPU features, then the default from the **corresponding new CPU feature** is used by Gramine. E.g. if user wonders what is the default for MPX now, then the answer is `disabled` because this is the default for `sgx.cpu_features.mpx `.
___
*[`pal/src/host/linux-sgx/sgx_arch.h` line 102 at r2](https://reviewable.io/reviews/gramineproject/gramine/881#-NP18cX5B0utAPOkpT9K:-NP1uyO-2UTrm11uO2Ny:bm79zx5) ([raw file](https://github.com/gramineproject/gramine/blob/d9feaae81024618610572d15577c67a4734387d3/pal/src/host/linux-sgx/sgx_arch.h#L102)):*

> - ``"disabled"``: ``SIGSTRUCT.ATTRIBUTEMASK[feature] = 1`` and
  ``SIGSTRUCT.ATTRIBUTES[feature] = 0``.

Is the corresponding bit unset?
</blockquote></details>

Oops, you are right. Fixed the comment now.
___
*[`python/graminelibos/manifest.py` line 100 at r2](https://reviewable.io/reviews/gramineproject/gramine/881#-NP0x4KF0xSVEQAEioSw:-NP1v48C2nUJrWEevs1-:b-n7k8tf) ([raw file](https://github.com/gramineproject/gramine/blob/d9feaae81024618610572d15577c67a4734387d3/python/graminelibos/manifest.py#L100)):*

As we removed the defaults of the legacy options, this changes the default behavior, right?
</blockquote></details>

No, the default behavior is the same for MPX and PKRU.

It's important to realize that security-hardening features (MPX and PKRU) and non-security-hardening features (AVX, AMX, ...) always had different behavior in Gramine. It is clear from this note: https://github.com/gramineproject/gramine/blob/9eff514112d9ca5398e0575c80e524e55a6a9a88/pal/src/host/linux-sgx/sgx_arch.h#L101-L114

So the previous manifest syntax with `sgx.require_xxx` was totally bad... This PR attempts at fixing the syntax and to describe it more clearly.

UPDATE: I removed this explicit setting of defaults in `manifest.py`, because they are already covered via `offs.SGX_XFRM_LEGACY` and `offs.SGX_XFRM_MASK_CONST`.
___
*[`python/graminelibos/sgx_sign.py` line 76 at r2](https://reviewable.io/reviews/gramineproject/gramine/881#-NP0wHbM3b5cS5Bp-O1A:-NP1ik-ZAdCe0-yGPIwK:b8l5ra1) ([raw file](https://github.com/gramineproject/gramine/blob/d9feaae81024618610572d15577c67a4734387d3/python/graminelibos/sgx_sign.py#L76)):*

This won't generate a KeyError anymore. I know this works for below but is this sth. we intended for?
</blockquote></details>

Yes, this is intended. Or to be more specific, we never intended to throw a KeyError here. 

This error never happened because we explicitly used `sgx.setdefault()` on those now-deprecated CPU feature options `require_avx`, etc. But now this error could happen because we don't force this explicit setting of now-deprecated options, so these keys may not exist now. That's why the change here.

UPDATE: I replaced `0` with `False` and `1` with `True`, because these are booleans. We are just lucky that Python maps 0 to False and 1 to True.
___
*[`python/graminelibos/sgx_sign.py` line 92 at r2](https://reviewable.io/reviews/gramineproject/gramine/881#-NP1EBJXEtieRyuuvmL9:-NP1wx2y7_aRW2WHfz-g:b35wj2g) ([raw file](https://github.com/gramineproject/gramine/blob/d9feaae81024618610572d15577c67a4734387d3/python/graminelibos/sgx_sign.py#L92)):*

Users can input things like `sgx.cpu_features.amx = "abcd"` (same semantics as `unspecified`)?
</blockquote></details>

Done. Good point.
___
*[`python/graminelibos/sgx_sign.py` line 126 at r2](https://reviewable.io/reviews/gramineproject/gramine/881#-NP0xISlBGez94l-KXII:-NP1zU9aCnED8G9V-6DY:b-z9zeqe) ([raw file](https://github.com/gramineproject/gramine/blob/d9feaae81024618610572d15577c67a4734387d3/python/graminelibos/sgx_sign.py#L126)):*

So we allow the co-existance and discrepancies between the legacy and new options?
</blockquote></details>

Done. Good catch. Significantly rewrote this part.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/881)* status: 0 of 11 files reviewed, 9 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "DONTMERGE" found in commit messages' one-liners (waiting on @donporter and @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/881#-NTScaQl1cpxTtW3N4ZN:-NTScaQm59zb52pwGS0T:bpujkq2) (no related file):*
I rebased this PR to the latest master (it got really stale, 200 commits behind master and 6 months old).

As a next step, I will test this PR again on AMX-enabled machines, using the `amxtest` application that I add in the third commit (this commit will be dropped at final rebase, since this is just a perf test).

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/881)* status: 0 of 11 files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "DONTMERGE" found in commit messages' one-liners (waiting on @donporter and @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/881#-NTScaQl1cpxTtW3N4ZN:-NTx945V4sSRFFgUyaVl:bypz9nx) (no related file):*
> As a next step, I will test this PR again on AMX-enabled machines ...

Done. Everything looks good.

___

donporter (changes requested):
> Reviewed 6 of 11 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/881)* status: 6 of 11 files reviewed, 9 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "DONTMERGE" found in commit messages' one-liners (waiting on @dimakuv and @kailun-qin)

___
*[`-- commits` line 33 at r4](https://reviewable.io/reviews/gramineproject/gramine/881#-NbKOGm9A41cFRtScrLL):*
I am not going to review these files, then, and leave a blocking comment here.  but will review the rest.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/881)* status: 2 of 11 files reviewed, 10 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "DONTMERGE" found in commit messages' one-liners (waiting on @donporter and @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/881#-NeCe4wr2qpSu8IwcPn3:-NeCe4wr2qpSu8IwcPn4:b-llbmeq) (no related file):*
I will need to make sure that other Gramine repos (gsc, examples, contrib) will also be updated to reflect this change in the manifest syntax.

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/881#-NeCeuwK0v4OALPkGyHa:-NeCeuwK0v4OALPkGyHb:bq40lov) (no related file):*
FYI: rebased to latest master.

The first commit in this rebased PR is also extracted as a separate PR: https://github.com/gramineproject/gramine/pull/1550

I also kept this first commit here, for the ease of testing this PR (it won't work without the first commit).

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/881)* status: 2 of 11 files reviewed, 10 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "DONTMERGE" found in commit messages' one-liners (waiting on @donporter and @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/881#-NeCeuwK0v4OALPkGyHa:-NeE_PxR98n4hBhG9U2x:b-k5z4bo) (no related file):*

FYI: rebased to latest master.

The first commit in this rebased PR is also extracted as a separate PR: https://github.com/gramineproject/gramine/pull/1550

I also kept this first commit here, for the ease of testing this PR (it won't work without the first commit).
</blockquote></details>

The first commit was extracted into #1550, which was just merged. So I rebased on top of the just-merged commit, and now there are only two commits in this PR (one of the commits is an AMX test that will be removed before final merge).

___

mkow (changes requested):
> Reviewed 1 of 11 files at r3.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/881)* status: 2 of 11 files reviewed, 11 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "DONTMERGE" found in commit messages' one-liners (waiting on @dimakuv, @donporter, and @kailun-qin)

___
*[`Documentation/manifest-syntax.rst` line 999 at r2](https://reviewable.io/reviews/gramineproject/gramine/881#-NP100eK8ldJcb-5rLNQ:-NeG8vRwBH5DLJOQCoCr:b-rehpzt) ([raw file](https://github.com/gramineproject/gramine/blob/d9feaae81024618610572d15577c67a4734387d3/Documentation/manifest-syntax.rst#L999)):*

Not sure I understand your question.

If users don't specify these deprecated CPU features, then the default from the **corresponding new CPU feature** is used by Gramine. E.g. if user wonders what is the default for MPX now, then the answer is `disabled` because this is the default for `sgx.cpu_features.mpx `.
</blockquote></details>

What Kailun means is that previously the docs said that these options are `false` by default, but you dropped it when moving the docs here.
___
*[`Documentation/manifest-syntax.rst` line 783 at r6](https://reviewable.io/reviews/gramineproject/gramine/881#-NeG5yGt2jCBmS1Xi3as:-NeG5yGt2jCBmS1Xi3at:b-hdmidg) ([raw file](https://github.com/gramineproject/gramine/blob/fc0ff085892f8a31959b79fa3f1b3b5757f807a2/Documentation/manifest-syntax.rst#L783)):*
> ```reStructuredText
> ::
> 
>     sgx.require_exinfo = [true|false]
> ```

Why is this left unchanged? `require_exinfo == false` is still misleading - the meaning in English is "don't require exinfo", but in practice it disables exinfo even on CPUs which support it.

kailun-qin (changes requested):
> Reviewed 6 of 11 files at r3, 1 of 1 files at r4, 4 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/881)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "DONTMERGE" found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`-- commits` line 22 at r2](https://reviewable.io/reviews/gramineproject/gramine/881#-NP16YW19KRR6SmdSzja):*
> Is this a typo in your sentence? Did you mean **We default security-hardening options...**?

Yeah, typo, excuse me.
___
*[`Documentation/manifest-syntax.rst` line 999 at r2](https://reviewable.io/reviews/gramineproject/gramine/881#-NP100eK8ldJcb-5rLNQ:-NeGdPs81OHywRNy5guv:b81dg54) ([raw file](https://github.com/gramineproject/gramine/blob/d9feaae81024618610572d15577c67a4734387d3/Documentation/manifest-syntax.rst#L999)):*
> What Kailun means is that previously the docs said that these options are `false` by default, but you dropped it when moving the docs here.

Yes, exactly.
___
*[`Documentation/manifest-syntax.rst` line 808 at r6](https://reviewable.io/reviews/gramineproject/gramine/881#-NeGoMtZ0rFq1dfqvM9b:-NeGoMtZ0rFq1dfqvM9c:bq3fugc) ([raw file](https://github.com/gramineproject/gramine/blob/fc0ff085892f8a31959b79fa3f1b3b5757f807a2/Documentation/manifest-syntax.rst#L808)):*
> ```reStructuredText
> security-hardening features (MPX and PKRU).
> 
> The ``"unspecified"`` syntax applies only to not-security-hardening features. It
> ```

Should we explain a bit why `"unspecified"` is not made for security-hardening features?



_Code quote:_
```reStructuredText
The ``"unspecified"`` syntax applies only to not-security-hardening features.
```
___
*[`python/graminelibos/sgx_sign.py` line 139 at r6](https://reviewable.io/reviews/gramineproject/gramine/881#-NeGklCb056UKkAn9wqx:-NeGklCb056UKkAn9wqy:b-a9bqb7) ([raw file](https://github.com/gramineproject/gramine/blob/fc0ff085892f8a31959b79fa3f1b3b5757f807a2/python/graminelibos/sgx_sign.py#L139)):*
> ```Python
>             if deprecated_key in manifest_sgx:
>                 raise KeyError(f'`sgx.cpu_features` cannot coexist with `sgx.{deprecated_key}`')
>         xfrms, xfrms_mask = collect_cpu_feature_bits(manifest_sgx, xfrms_dict, xfrms, xfrms_mask,
> ```

It's guaranteed that we have `manifest_sgx['cpu_features']` when we're here, can we input it directly (instead of inputing `manifest_sgx`)?



_Code quote:_
```Python
xfrms, xfrms_mask = collect_cpu_feature_bits(manifest_sgx, xfrms_dict, xfrms, xfrms_mask
```
___
*[`python/graminelibos/sgx_sign.py` line 140 at r6](https://reviewable.io/reviews/gramineproject/gramine/881#-NeGlKbq0BaZQCk3Glny:-NeGlKbq0BaZQCk3Glnz:bu2ngx9) ([raw file](https://github.com/gramineproject/gramine/blob/fc0ff085892f8a31959b79fa3f1b3b5757f807a2/python/graminelibos/sgx_sign.py#L140)):*
> ```Python
>                 raise KeyError(f'`sgx.cpu_features` cannot coexist with `sgx.{deprecated_key}`')
>         xfrms, xfrms_mask = collect_cpu_feature_bits(manifest_sgx, xfrms_dict, xfrms, xfrms_mask,
>                                                      security_hardening=False)
> ```

I don't insist, but is it better to have a nested dict for `xfrms_dict` and `secure_xfrms_dict` so that we can get rid of this `security_hardening=xxx`?



_Code quote:_
```Python
security_hardening=False
```

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/881)* status: 8 of 11 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/881#-NeCe4wr2qpSu8IwcPn3:-NeHsjxw0YwWTfmYoWma:b42d1c9) (no related file):*

I will need to make sure that other Gramine repos (gsc, examples, contrib) will also be updated to reflect this change in the manifest syntax.
</blockquote></details>

Done:
- GSC -- requires no changes (doesn't use `sgx.require_*` at all)
- Examples -- one PR submitted: https://github.com/gramineproject/examples/pull/77
- Contrib-- requires no changes (doesn't use `sgx.require_*` at all)

___
___
*[`Documentation/manifest-syntax.rst` line 999 at r2](https://reviewable.io/reviews/gramineproject/gramine/881#-NP100eK8ldJcb-5rLNQ:-NeHmqgj4qVDkMOcLwRb:buxcdnw) ([raw file](https://github.com/gramineproject/gramine/blob/d9feaae81024618610572d15577c67a4734387d3/Documentation/manifest-syntax.rst#L999)):*

> What Kailun means is that previously the docs said that these options are `false` by default, but you dropped it when moving the docs here.

Yes, exactly.
</blockquote></details>

Done. Added a new paragraph to explain this.
___
*[`Documentation/manifest-syntax.rst` line 783 at r6](https://reviewable.io/reviews/gramineproject/gramine/881#-NeG5yGt2jCBmS1Xi3as:-NeHreo08xwgqPN0c2En:b-ruhvpn) ([raw file](https://github.com/gramineproject/gramine/blob/fc0ff085892f8a31959b79fa3f1b3b5757f807a2/Documentation/manifest-syntax.rst#L783)):*

Why is this left unchanged? `require_exinfo == false` is still misleading - the meaning in English is "don't require exinfo", but in practice it disables exinfo even on CPUs which support it.
</blockquote></details>

Done, in a separate commit
___
*[`Documentation/manifest-syntax.rst` line 808 at r6](https://reviewable.io/reviews/gramineproject/gramine/881#-NeGoMtZ0rFq1dfqvM9b:-NeHljs9D3Mg5VPaAZvo:b-sh10pq) ([raw file](https://github.com/gramineproject/gramine/blob/fc0ff085892f8a31959b79fa3f1b3b5757f807a2/Documentation/manifest-syntax.rst#L808)):*

Should we explain a bit why `"unspecified"` is not made for security-hardening features?
</blockquote></details>

Done. Not sure if the new sentence brings anything new, but check if you like it.
___
*[`python/graminelibos/sgx_sign.py` line 139 at r6](https://reviewable.io/reviews/gramineproject/gramine/881#-NeGklCb056UKkAn9wqx:-NeHnoPzBKcWab_2-i2B:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/fc0ff085892f8a31959b79fa3f1b3b5757f807a2/python/graminelibos/sgx_sign.py#L139)):*

It's guaranteed that we have `manifest_sgx['cpu_features']` when we're here, can we input it directly (instead of inputing `manifest_sgx`)?
</blockquote></details>

Done.
___
*[`python/graminelibos/sgx_sign.py` line 140 at r6](https://reviewable.io/reviews/gramineproject/gramine/881#-NeGlKbq0BaZQCk3Glny:-NeHnpZI1j5d6QqLxE8v:b8bb60p) ([raw file](https://github.com/gramineproject/gramine/blob/fc0ff085892f8a31959b79fa3f1b3b5757f807a2/python/graminelibos/sgx_sign.py#L140)):*

I don't insist, but is it better to have a nested dict for `xfrms_dict` and `secure_xfrms_dict` so that we can get rid of this `security_hardening=xxx`?
</blockquote></details>

I like the current code with two invocations -- one for non-security-critical features (`xfrms_dict`) and one for security-critical features (`secure_xfrms_dict`). I would like to keep it that way, I think it's more readable.

mkow (changes requested):
> Reviewed 1 of 11 files at r3, 1 of 4 files at r5, 2 of 3 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/881)* status: 10 of 11 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @kailun-qin)

___
*[`Documentation/manifest-syntax.rst` line 789 at r8](https://reviewable.io/reviews/gramineproject/gramine/881#-NeJ4CYk2MrcAptnOQyh:-NeJ4CYk2MrcAptnOQyi:bq7v6wl) ([raw file](https://github.com/gramineproject/gramine/blob/9cc5c63857478c7602e1ebdc3941aaf093e5cf3a/Documentation/manifest-syntax.rst#L789)):*
> ```reStructuredText
> signal handler in case of a page fault. Otherwise (set to ``false``), the
> faulting address will always be provided as ``0``.
> ```

I'd also mention here _why_ the default is `false` - some frameworks/runtimes would dump the callstack + variables/registers on exceptions, potentially leaking data.
___
*[`python/graminelibos/manifest.py` line 99 at r8](https://reviewable.io/reviews/gramineproject/gramine/881#-NeJ9idk3ILzaDjVoLAW:-NeJ9idk3ILzaDjVoLAX:beq0g9m) ([raw file](https://github.com/gramineproject/gramine/blob/9cc5c63857478c7602e1ebdc3941aaf093e5cf3a/python/graminelibos/manifest.py#L99)):*
> ```Python
>         sgx.setdefault('enable_stats', False)
>         sgx.setdefault('edmm_enable', False)
> ```

No defaults for `sgx.features`? All other options have the defaults set up here, why making an exception for `sgx.features`?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/881)* status: 4 of 11 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "DONTMERGE" and "fixup! " found in commit messages' one-liners (waiting on @kailun-qin and @mkow)

___
*[`Documentation/manifest-syntax.rst` line 789 at r8](https://reviewable.io/reviews/gramineproject/gramine/881#-NeJ4CYk2MrcAptnOQyh:-NeJpTn70h7vbO-0fTJb:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9cc5c63857478c7602e1ebdc3941aaf093e5cf3a/Documentation/manifest-syntax.rst#L789)):*

I'd also mention here _why_ the default is `false` - some frameworks/runtimes would dump the callstack + variables/registers on exceptions, potentially leaking data.
</blockquote></details>

Done.
___
*[`python/graminelibos/manifest.py` line 99 at r8](https://reviewable.io/reviews/gramineproject/gramine/881#-NeJ9idk3ILzaDjVoLAW:-NeJpYyJBizo8daiqi2l:b-c5i7fy) ([raw file](https://github.com/gramineproject/gramine/blob/9cc5c63857478c7602e1ebdc3941aaf093e5cf3a/python/graminelibos/manifest.py#L99)):*

No defaults for `sgx.features`? All other options have the defaults set up here, why making an exception for `sgx.features`?
</blockquote></details>

Done.

I actually don't like setting these defaults (I don't know/remember who started this), because it is redundant (Gramine's parsing code already has the defaults) and every time we modify some manifest options, we also need to modify here...

But I understand that we need to be consistent, so let me add the defaults here, and maybe later we'll decide to remove this defaults list.

mkow (changes requested):
> Reviewed 1 of 1 files at r8, 2 of 6 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/881)* status: 7 of 11 files reviewed, 12 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "DONTMERGE" and "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @kailun-qin)

___
*[`python/graminelibos/manifest.py` line 99 at r8](https://reviewable.io/reviews/gramineproject/gramine/881#-NeJ9idk3ILzaDjVoLAW:-NeKNz8O0XqBsQRPsWCl:boh08mm) ([raw file](https://github.com/gramineproject/gramine/blob/9cc5c63857478c7602e1ebdc3941aaf093e5cf3a/python/graminelibos/manifest.py#L99)):*

Done.

I actually don't like setting these defaults (I don't know/remember who started this), because it is redundant (Gramine's parsing code already has the defaults) and every time we modify some manifest options, we also need to modify here...

But I understand that we need to be consistent, so let me add the defaults here, and maybe later we'll decide to remove this defaults list.
</blockquote></details>

Hmm, maybe it was for the Python code which uses some of these, so that it doesn't have to use `.get()` everywhere? I'm not sure...
___
*[`python/graminelibos/manifest.py` line 105 at r9](https://reviewable.io/reviews/gramineproject/gramine/881#-NeKP-7Q6KpY4Rx7CQTT:-NeKP-7Q6KpY4Rx7CQTU:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/b4dc821c28630311bece86b8b8822a97d499cf65/python/graminelibos/manifest.py#L105)):*
> ```Python
>             sgx.setdefault('enclave_size', DEFAULT_ENCLAVE_SIZE_NO_EDMM)
> 
>         # TODO: below were deprecated in release v1.6, remove this check in v1.7;
> ```





_Suggestion:_
```Python
remove this check in v1.7: (but keep the `if` body)
```
___
*[`python/graminelibos/sgx_sign.py` line 139 at r6](https://reviewable.io/reviews/gramineproject/gramine/881#-NeGklCb056UKkAn9wqx:-NeK_LHsCsZgTi29mcmR:b-ejhn6a) ([raw file](https://github.com/gramineproject/gramine/blob/fc0ff085892f8a31959b79fa3f1b3b5757f807a2/python/graminelibos/sgx_sign.py#L139)):*

Done.
</blockquote></details>

It actually is, because of the defaults in `manifest.py`.
___
*[`python/graminelibos/sgx_sign.py` line 140 at r6](https://reviewable.io/reviews/gramineproject/gramine/881#-NeGlKbq0BaZQCk3Glny:-NeK_XT-ExpYc7r5ONn1:b-gefjnq) ([raw file](https://github.com/gramineproject/gramine/blob/fc0ff085892f8a31959b79fa3f1b3b5757f807a2/python/graminelibos/sgx_sign.py#L140)):*

I like the current code with two invocations -- one for non-security-critical features (`xfrms_dict`) and one for security-critical features (`secure_xfrms_dict`). I would like to keep it that way, I think it's more readable.
</blockquote></details>

I think I prefer the current version also.
___
*[`python/graminelibos/sgx_sign.py` line 80 at r9](https://reviewable.io/reviews/gramineproject/gramine/881#-NeKQawP4YEO9lDXy-bB:-NeKQawP4YEO9lDXy-bC:b6wcgfc) ([raw file](https://github.com/gramineproject/gramine/blob/b4dc821c28630311bece86b8b8822a97d499cf65/python/graminelibos/sgx_sign.py#L80)):*
> ```Python
>     val = 0
>     for opt, bits in options_dict.items():
>         if manifest_sgx.get(opt) is True:
> ```

This change is not needed anymore.
___
*[`python/graminelibos/sgx_sign.py` line 87 at r9](https://reviewable.io/reviews/gramineproject/gramine/881#-NeKRaPx-Y9_k2Hl5joj:-NeKRaPx-Y9_k2Hl5jok:bs3lftb) ([raw file](https://github.com/gramineproject/gramine/blob/b4dc821c28630311bece86b8b8822a97d499cf65/python/graminelibos/sgx_sign.py#L87)):*
> ```Python
> def collect_cpu_feature_bits(manifest_cpu_features, options_dict, val, mask, security_hardening):
>     for opt, bits in options_dict.items():
>         if manifest_cpu_features.get(opt) is None:
> ```

More Pythonic way:



_Suggestion:_
```Python
if opt not in manifest_cpu_features:
```
___
*[`python/graminelibos/sgx_sign.py` line 88 at r9](https://reviewable.io/reviews/gramineproject/gramine/881#-NeKPTpmD8wYCzh_4Eyk:-NeKPTpmD8wYCzh_4Eyl:b2oqogt) ([raw file](https://github.com/gramineproject/gramine/blob/b4dc821c28630311bece86b8b8822a97d499cf65/python/graminelibos/sgx_sign.py#L88)):*
> ```Python
>     for opt, bits in options_dict.items():
>         if manifest_cpu_features.get(opt) is None:
>             continue
> ```

This is not needed, `manifest.py` guarantees that these options are always set. Though, we aren't consistent in this I think, sometimes we rely on this here and sometimes not?
___
*[`python/graminelibos/sgx_sign.py` line 96 at r9](https://reviewable.io/reviews/gramineproject/gramine/881#-NeKVRDJ1cPJKacz77HV:-NeKVRDJ1cPJKacz77HW:bwyrxip) ([raw file](https://github.com/gramineproject/gramine/blob/b4dc821c28630311bece86b8b8822a97d499cf65/python/graminelibos/sgx_sign.py#L96)):*
> ```Python
>             mask |= bits
>         elif security_hardening or manifest_cpu_features[opt] != "unspecified":
>             raise KeyError(f'Manifest option `sgx.cpu_features.{opt}` has disallowed value')
> ```

Otherwise "has disallowed" would form a past tense.
OTOH, in plural this would be ambiguous, I guess? :)



_Suggestion:_
```Python
has a disallowed value
```
___
*[`python/graminelibos/sgx_sign.py` line 108 at r9](https://reviewable.io/reviews/gramineproject/gramine/881#-NeKZXXeANeKVkR70yvi:-NeKZXXeANeKVkR70yvj:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/b4dc821c28630311bece86b8b8822a97d499cf65/python/graminelibos/sgx_sign.py#L108)):*
> ```Python
>         flags |= offs.SGX_FLAGS_MODE64BIT
> 
>     if manifest_sgx.get('require_exinfo') is not None:
> ```

ditto
___
*[`python/graminelibos/sgx_sign.py` line 110 at r9](https://reviewable.io/reviews/gramineproject/gramine/881#-NeKZbwh5Y00-ZuRiiCf:-NeKZbwh5Y00-ZuRiiCg:b1c9lrl) ([raw file](https://github.com/gramineproject/gramine/blob/b4dc821c28630311bece86b8b8822a97d499cf65/python/graminelibos/sgx_sign.py#L110)):*
> ```Python
>     if manifest_sgx.get('require_exinfo') is not None:
>         raise KeyError(f'Manifest option `sgx.require_exinfo` was renamed to `sgx.use_exinfo`, '
>                         'please modify the manifest file accordingly')
> ```

Hmm, wasn't the old name only deprecated, not removed?

kailun-qin (commented):
> Reviewed 1 of 1 files at r8, 6 of 6 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/881)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "DONTMERGE" and "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`python/graminelibos/sgx_sign.py` line 140 at r6](https://reviewable.io/reviews/gramineproject/gramine/881#-NeGlKbq0BaZQCk3Glny:-NeLb33a1vtazc2oEDCN:bbix6dv) ([raw file](https://github.com/gramineproject/gramine/blob/fc0ff085892f8a31959b79fa3f1b3b5757f807a2/python/graminelibos/sgx_sign.py#L140)):*

I think I prefer the current version also.
</blockquote></details>

OK then :)
___
*[`python/graminelibos/sgx_sign.py` line 96 at r9](https://reviewable.io/reviews/gramineproject/gramine/881#-NeKVRDJ1cPJKacz77HV:-NeLaLeF5ZHUksGAkp_f:b-bn19v3) ([raw file](https://github.com/gramineproject/gramine/blob/b4dc821c28630311bece86b8b8822a97d499cf65/python/graminelibos/sgx_sign.py#L96)):*

Otherwise "has disallowed" would form a past tense.
OTOH, in plural this would be ambiguous, I guess? :)
</blockquote></details>

What about "contains disallowed value"?
___
*[`python/graminelibos/sgx_sign.py` line 110 at r9](https://reviewable.io/reviews/gramineproject/gramine/881#-NeKZbwh5Y00-ZuRiiCf:-NeLb-5xErIy7Z-fo7Me:b32v57) ([raw file](https://github.com/gramineproject/gramine/blob/b4dc821c28630311bece86b8b8822a97d499cf65/python/graminelibos/sgx_sign.py#L110)):*

Hmm, wasn't the old name only deprecated, not removed?
</blockquote></details>

+1

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/881)* status: 8 of 11 files reviewed, 9 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin and @mkow)

___
*[`python/graminelibos/manifest.py` line 105 at r9](https://reviewable.io/reviews/gramineproject/gramine/881#-NeKP-7Q6KpY4Rx7CQTT:-NeMbmpk5QbnJV-FAfpb:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/b4dc821c28630311bece86b8b8822a97d499cf65/python/graminelibos/manifest.py#L105)):*
> ```Python
>             sgx.setdefault('enclave_size', DEFAULT_ENCLAVE_SIZE_NO_EDMM)
> 
>         # TODO: below were deprecated in release v1.6, remove this check in v1.7;
> ```

Done.
___
*[`python/graminelibos/sgx_sign.py` line 139 at r6](https://reviewable.io/reviews/gramineproject/gramine/881#-NeGklCb056UKkAn9wqx:-NeMeY80DJ1d83x9t9Lf:b-iansbk) ([raw file](https://github.com/gramineproject/gramine/blob/fc0ff085892f8a31959b79fa3f1b3b5757f807a2/python/graminelibos/sgx_sign.py#L139)):*

It actually is, because of the defaults in `manifest.py`.
</blockquote></details>

@mkow I don't understand your comment. What are you blocking on? What action should I do?
___
*[`python/graminelibos/sgx_sign.py` line 80 at r9](https://reviewable.io/reviews/gramineproject/gramine/881#-NeKQawP4YEO9lDXy-bB:-NeMc8Wd5g-KpOII1lAR:b8yqdqx) ([raw file](https://github.com/gramineproject/gramine/blob/b4dc821c28630311bece86b8b8822a97d499cf65/python/graminelibos/sgx_sign.py#L80)):*

This change is not needed anymore.
</blockquote></details>

This change is still needed, because we can have deprecated options in the manifest-template file. E.g.:
```
# app.manifest.template

sgx.require_avx = true
sgx.require_mpx = false
```

In this case, our `gramine-manifest` tool detects the deprecated options and **does not** add the `sgx.cpu_features.xxx` options at all (to not have a conflict between the deprecated ones and the new ones).

Also note that the other deprecated options will **not be added** to the manifest file. E.g. `sgx.require_avx512` will not be added (because it's one of the deprecated options, and we simply don't have the logic to add it explicitly).

So this `.get(opt)` is still required, because there are legitimate cases when the (deprecated) option was not set to a default and thus is missing completely from the manifest dict.

P.S. This whole thing can indeed go away when we don't support the deprecated options anymore, and in that case we'll always have the defaults in the manifest dict, and the `.get(opt)` trick could be replaced with a simple `[opt]`. But not now.
___
*[`python/graminelibos/sgx_sign.py` line 87 at r9](https://reviewable.io/reviews/gramineproject/gramine/881#-NeKRaPx-Y9_k2Hl5joj:-NeMdQR0AjI8r5JQGmkN:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/b4dc821c28630311bece86b8b8822a97d499cf65/python/graminelibos/sgx_sign.py#L87)):*

More Pythonic way:
</blockquote></details>

Done.
___
*[`python/graminelibos/sgx_sign.py` line 88 at r9](https://reviewable.io/reviews/gramineproject/gramine/881#-NeKPTpmD8wYCzh_4Eyk:-NeMd8Oc9nmZyVPNm0lD:b-n9ettp) ([raw file](https://github.com/gramineproject/gramine/blob/b4dc821c28630311bece86b8b8822a97d499cf65/python/graminelibos/sgx_sign.py#L88)):*

This is not needed, `manifest.py` guarantees that these options are always set. Though, we aren't consistent in this I think, sometimes we rely on this here and sometimes not?
</blockquote></details>

No, we are not guaranteed that these options are set. See my other comment.
___
*[`python/graminelibos/sgx_sign.py` line 96 at r9](https://reviewable.io/reviews/gramineproject/gramine/881#-NeKVRDJ1cPJKacz77HV:-NeMdYXIB_xbkSxAhK7B:b-wtx14u) ([raw file](https://github.com/gramineproject/gramine/blob/b4dc821c28630311bece86b8b8822a97d499cf65/python/graminelibos/sgx_sign.py#L96)):*

What about "contains disallowed value"?
</blockquote></details>

Done. I like Michal's suggestion better, so went with it.
___
*[`python/graminelibos/sgx_sign.py` line 108 at r9](https://reviewable.io/reviews/gramineproject/gramine/881#-NeKZXXeANeKVkR70yvi:-NeMeQ2l1pp7xM5zLJtq:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/b4dc821c28630311bece86b8b8822a97d499cf65/python/graminelibos/sgx_sign.py#L108)):*

ditto
</blockquote></details>

Done.
___
*[`python/graminelibos/sgx_sign.py` line 110 at r9](https://reviewable.io/reviews/gramineproject/gramine/881#-NeKZbwh5Y00-ZuRiiCf:-NeMeQX75x05jHs199cf:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/b4dc821c28630311bece86b8b8822a97d499cf65/python/graminelibos/sgx_sign.py#L110)):*

+1
</blockquote></details>

Done.

mkow (approved):
> Reviewed 3 of 3 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/881)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`python/graminelibos/sgx_sign.py` line 139 at r6](https://reviewable.io/reviews/gramineproject/gramine/881#-NeGklCb056UKkAn9wqx:-NeOk-shDV558T_kwOrh:btzrl6e) ([raw file](https://github.com/gramineproject/gramine/blob/fc0ff085892f8a31959b79fa3f1b3b5757f807a2/python/graminelibos/sgx_sign.py#L139)):*

@mkow I don't understand your comment. What are you blocking on? What action should I do?
</blockquote></details>

Hmm, I think I accidentally replied under a wrong thread, sorry.
___
*[`python/graminelibos/sgx_sign.py` line 80 at r9](https://reviewable.io/reviews/gramineproject/gramine/881#-NeKQawP4YEO9lDXy-bB:-NeOj03h16vT2j65rYxA:b-vy5pcl) ([raw file](https://github.com/gramineproject/gramine/blob/b4dc821c28630311bece86b8b8822a97d499cf65/python/graminelibos/sgx_sign.py#L80)):*

This change is still needed, because we can have deprecated options in the manifest-template file. E.g.:
```
# app.manifest.template

sgx.require_avx = true
sgx.require_mpx = false
```

In this case, our `gramine-manifest` tool detects the deprecated options and **does not** add the `sgx.cpu_features.xxx` options at all (to not have a conflict between the deprecated ones and the new ones).

Also note that the other deprecated options will **not be added** to the manifest file. E.g. `sgx.require_avx512` will not be added (because it's one of the deprecated options, and we simply don't have the logic to add it explicitly).

So this `.get(opt)` is still required, because there are legitimate cases when the (deprecated) option was not set to a default and thus is missing completely from the manifest dict.

P.S. This whole thing can indeed go away when we don't support the deprecated options anymore, and in that case we'll always have the defaults in the manifest dict, and the `.get(opt)` trick could be replaced with a simple `[opt]`. But not now.
</blockquote></details>

Ah, I see, makes sense.

kailun-qin (approved):
> Reviewed 3 of 3 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/881)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

donporter (changes requested):
> Reviewed 1 of 11 files at r3, 1 of 4 files at r5, 4 of 6 files at r9, 3 of 3 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/881)* status: all files reviewed, 3 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`CI-Examples/amxtest/README.md` line 127 at r10](https://reviewable.io/reviews/gramineproject/gramine/881#-NecTXIcBeO2Z427nEl4:-NecTXIcBeO2Z427nEl5:b-pyyn2h) ([raw file](https://github.com/gramineproject/gramine/blob/cbc4684a644f21ec0b825a10a2a0633423afa1dc/CI-Examples/amxtest/README.md#L127)):*
> ```Markdown
>  static long sgx_ocall_sched_yield(void* args) {
>      __UNUSED(args);
> +#if 1
> ```

Is this #define needed in the documentation?
___
*[`Documentation/manifest-syntax.rst` line 787 at r10](https://reviewable.io/reviews/gramineproject/gramine/881#-NecTsPj54bt5yhywTia:-NecTsPj54bt5yhywTib:b6bknn0) ([raw file](https://github.com/gramineproject/gramine/blob/cbc4684a644f21ec0b825a10a2a0633423afa1dc/Documentation/manifest-syntax.rst#L787)):*
> ```reStructuredText
> 
> If ``sgx.use_exinfo`` is set, user application can retrieve faulting address in
> signal handler in case of a page fault. Otherwise (set to ``false``), the
> ```

Worth explicating that this is untrusted/non-enclave user code?

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/881)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @donporter)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/881#-NTScaQl1cpxTtW3N4ZN:-NecW-NU4Gh93sFiQMtU:be5d6b2) (no related file):*

> As a next step, I will test this PR again on AMX-enabled machines ...

Done. Everything looks good.
</blockquote></details>

Blocking until the AMX test gets removed.

___
___
*[`CI-Examples/amxtest/README.md` line 127 at r10](https://reviewable.io/reviews/gramineproject/gramine/881#-NecTXIcBeO2Z427nEl4:-NecVhGO7Wa3mRXf9XHl:bb82bso) ([raw file](https://github.com/gramineproject/gramine/blob/cbc4684a644f21ec0b825a10a2a0633423afa1dc/CI-Examples/amxtest/README.md#L127)):*
<details><summary><i>Previously, donporter (Don Porter) wrote…</i></summary><blockquote>

Is this #define needed in the documentation?
</blockquote></details>

This test will be dropped before merging, there was a blocking comment somewhere, but seems that someone unlocked it accidentally, let me fix that.
___
*[`Documentation/manifest-syntax.rst` line 787 at r10](https://reviewable.io/reviews/gramineproject/gramine/881#-NecTsPj54bt5yhywTia:-NecVXO33tQm67BvD7nw:b-i49qt5) ([raw file](https://github.com/gramineproject/gramine/blob/cbc4684a644f21ec0b825a10a2a0633423afa1dc/Documentation/manifest-syntax.rst#L787)):*
<details><summary><i>Previously, donporter (Don Porter) wrote…</i></summary><blockquote>

Worth explicating that this is untrusted/non-enclave user code?
</blockquote></details>

What do you mean? The fault address is passed to the app inside Gramine? There's no "non-enclave user code" in Gramine?

donporter (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/881)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`CI-Examples/amxtest/README.md` line 127 at r10](https://reviewable.io/reviews/gramineproject/gramine/881#-NecTXIcBeO2Z427nEl4:-NecWb9Q1pHOhEdo3cBY:b-h3l4o4) ([raw file](https://github.com/gramineproject/gramine/blob/cbc4684a644f21ec0b825a10a2a0633423afa1dc/CI-Examples/amxtest/README.md#L127)):*

This test will be dropped before merging, there was a blocking comment somewhere, but seems that someone unlocked it accidentally, let me fix that.
</blockquote></details>

Ah, right.  My mistake
___
*[`Documentation/manifest-syntax.rst` line 787 at r10](https://reviewable.io/reviews/gramineproject/gramine/881#-NecTsPj54bt5yhywTia:-NecWESq--uVYh7xK1Zv:bpa8hvy) ([raw file](https://github.com/gramineproject/gramine/blob/cbc4684a644f21ec0b825a10a2a0633423afa1dc/Documentation/manifest-syntax.rst#L787)):*

What do you mean? The fault address is passed to the app inside Gramine? There's no "non-enclave user code" in Gramine?
</blockquote></details>

Ah, My mistake.  I was thinking this was referring the signal handler outside the enclave.

dimakuv:
Jenkins, retest Jenkins-SGX-EDMM please (there seemed to be some problem with Intel PCCS service, but now it's gone?
)

donporter (approved):
> Reviewed 5 of 5 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/881)* status: all files reviewed, 1 unresolved discussion

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/881)* status: all files reviewed, 1 unresolved discussion (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/881#-NTScaQl1cpxTtW3N4ZN:-Necfire1TEYaX7-Mhfg:b6wksql) (no related file):*

Blocking until the AMX test gets removed.
</blockquote></details>

I extracted the AMX test in my private repo (just not to lose that test) and removed from this PR.

___
___
*[`-- commits` line 19 at r11](https://reviewable.io/reviews/gramineproject/gramine/881#-NeciwWt6zfMzBqgzouG):*
FYI: previously I had a separate commit "[PAL/Linux-SGX] Rename `sgx.require_exinfo` to `sgx.use_exinfo`". But I messed up the fixup commits, so instead of untangling the commits, I melded the second commit into the first one. Sorry :(

mkow (approved):
> Reviewed 5 of 5 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/881)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #894, opened 2022-09-07
[PAL/Linux-SGX] Mitigate CVE-2022-21233 aka INTEL-SA-00657

### LINKED ISSUE #846: Evaluate and mitigate stale data reads in Gramine SGX
Intel disclosed the Stale Data Read vulnerability recently. In order to mitigate this issue completely, reading data from outside an SGX enclave's own linear memory range (ELRANGE) must be aligned and in 8-bytes chunks. We'd better evaluate if there is any impact to Gramine SGX and mitigate accordingly.

References:

https://www.intel.com/content/www/us/en/security-center/advisory/intel-sa-00657.html
https://www.intel.com/content/www/us/en/developer/articles/technical/software-security-guidance/advisory-guidance/stale-data-read-from-xapic.html

boryspoplawski:
## Description of the changes 
If malicious host maps APIC registers in some untrusted memory (MMIO) and this memory is subsequently read by the enclave (e.g. when reading some results of an ocall) such read can potentially return some stale, enclave-originated data. Naturally aligned 8-byte reads are not affected by this vulnerability, so as software mitigation, this commit makes all untrusted memory accesses 8-byte wide and naturally aligned.

As a side effect the exitless ocalls feature becomes potentially insecure. It uses untrusted memory in some complex manner, which we did not bother changing for now, as the feature is not widely used.

Fixes #846

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/894)

### DIFF
--- a/CI-Examples/nginx/nginx.manifest.template
+++ b/CI-Examples/nginx/nginx.manifest.template
@@ -24,9 +24,6 @@ sgx.nonpie_binary = true
 sgx.enclave_size = "512M"
 sgx.thread_num = 4
 
-# Nginx benefits from Exitless. Uncomment the below line to use it.
-#sgx.rpc_thread_num = 4
-
 sgx.trusted_files = [
   "file:{{ gramine.libos }}",
   "file:{{ install_dir }}/sbin/nginx",
--- a/Documentation/manifest-syntax.rst
+++ b/Documentation/manifest-syntax.rst
@@ -511,7 +511,7 @@ Number of RPC threads (Exitless feature)
 
 ::
 
-    sgx.rpc_thread_num = [NUM]
+    sgx.insecure__rpc_thread_num = [NUM]
     (Default: 0)
 
 This syntax specifies the number of RPC threads that are created outside of
@@ -524,7 +524,7 @@ the enclave (except for a few syscalls where there is no benefit, e.g.,
 If user specifies ``0`` or omits this directive, then no RPC threads are
 created and all system calls perform an enclave exit ("normal" execution).
 
-Note that the number of created RPC threads must match the maximum number of
+Note that the number of created RPC threads should match the maximum number of
 simultaneous enclave threads. If there are more RPC threads, then CPU time is
 wasted. If there are less RPC threads, some enclave threads may starve,
 especially if there are many blocking system calls by other enclave threads.
@@ -534,6 +534,9 @@ OCALLs/ECALLs for fast shared-memory communication at the cost of occupying
 more CPU cores and burning more CPU cycles. For example, a single-threaded
 Redis instance on Linux becomes 5-threaded on Gramine with Exitless. Thus,
 Exitless may negatively impact throughput but may improve latency.
+
+This feature is currently marked as insecure, because it reads untrusted memory
+in potentially insecure manner - susceptible to CVE-2022-21233 (INTEL-SA-00657).
 
 Optional CPU features (AVX, AVX512, MPX, PKRU, AMX)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/Documentation/performance.rst
+++ b/Documentation/performance.rst
@@ -163,6 +163,9 @@ to OCALLs:
 Exitless feature
 ----------------
 
+Note this feature is currently insecure and not recommended for production
+usage (potentially susceptible to CVE-2022-21233 aka INTEL-SA-00657).
+
 Gramine supports the Exitless (or Switchless) feature – it trades off CPU cores
 for faster OCALL execution. More specifically, with Exitless, enclave threads do
 not exit the enclave on OCALLs but instead busy wait for untrusted helper
@@ -172,28 +175,28 @@ requests for OCALLs from enclave threads (untrusted helper threads periodically
 sleep if there have been no OCALL requests for a long time to save some CPU
 cycles).
 
-Exitless is configured by ``sgx.rpc_thread_num = xyz``. By default, the Exitless
-feature is disabled – all enclave threads perform an actual OCALL for each
-system call and exit the enclave. The feature can be disabled by specifying
-``sgx.rpc_thread_num = 0``.
+Exitless is configured by ``sgx.insecure__rpc_thread_num = xyz``. By default,
+the Exitless feature is disabled – all enclave threads perform an actual OCALL
+for each system call and exit the enclave. The feature can be disabled by
+specifying ``sgx.insecure__rpc_thread_num = 0``.
 
 You must decide how many untrusted helper RPC threads your application needs. A
-rule of thumb: specify ``sgx.rpc_thread_num == sgx.thread_num``, i.e., the
-number of untrusted RPC threads should be the same as the number of enclave
+rule of thumb: specify ``sgx.insecure__rpc_thread_num == sgx.thread_num``, i.e.,
+the number of untrusted RPC threads should be the same as the number of enclave
 threads. For example, native Redis 6.0 uses 3-4 enclave threads during its
 execution, plus Gramine uses another 1-2 helper enclave threads. So Redis
 manifest has an over-approximation of this number: ``sgx.thread_num = 8``. Thus,
-to correctly enable the Exitless feature, specify ``sgx.rpc_thread_num = 8``.
-Here is an example:
+to correctly enable the Exitless feature, specify
+``sgx.insecure__rpc_thread_num = 8``. Here is an example:
 
 ::
 
-   # exitless disabled: `sgx.thread_num = 8` and `sgx.rpc_thread_num = 0`
+   # exitless disabled: `sgx.thread_num = 8` and `sgx.insecure__rpc_thread_num = 0`
    CI-Examples/redis$ gramine-sgx redis-server --save '' --protected-mode no &
    CI-Examples/redis$ src/src/redis-benchmark -t set
    43010.75 requests per second
 
-   # exitless enabled: `sgx.thread_num = 8` and `sgx.rpc_thread_num = 8`
+   # exitless enabled: `sgx.thread_num = 8` and `sgx.insecure__rpc_thread_num = 8`
    CI-Examples/redis$ gramine-sgx redis-server --save '' --protected-mode no &
    CI-Examples/redis$ src/src/redis-benchmark -t set
    68119.89 requests per second
--- a/libos/test/regression/multi_pthread_exitless.manifest.template
+++ b/libos/test/regression/multi_pthread_exitless.manifest.template
@@ -10,7 +10,7 @@ fs.mounts = [
 
 # app runs with 4 parallel threads + Gramine has couple internal threads
 sgx.thread_num = 8
-sgx.rpc_thread_num = 8
+sgx.insecure__rpc_thread_num = 8
 
 sgx.nonpie_binary = true
 sgx.debug = true
--- a/pal/regression/Thread2_exitless.manifest.template
+++ b/pal/regression/Thread2_exitless.manifest.template
@@ -3,7 +3,7 @@
 loader.entrypoint = "file:{{ binary_dir }}/{{ entrypoint }}"
 
 sgx.thread_num = 2
-sgx.rpc_thread_num = 2
+sgx.insecure__rpc_thread_num = 2
 sgx.enable_stats = true
 sgx.nonpie_binary = true
 sgx.debug = true
--- a/pal/src/host/linux-sgx/enclave_api.h
+++ b/pal/src/host/linux-sgx/enclave_api.h
@@ -16,10 +16,17 @@ void* sgx_alloc_on_ustack(uint64_t size);
 void* sgx_copy_to_ustack(const void* ptr, uint64_t size);
 void sgx_reset_ustack(const void* old_ustack);
 
+void sgx_copy_to_enclave_verified(void* ptr, const void* uptr, size_t size);
 bool sgx_copy_to_enclave(void* ptr, size_t maxsize, const void* uptr, size_t usize);
 void* sgx_import_array_to_enclave(const void* uptr, size_t elem_size, size_t elem_cnt);
 void* sgx_import_array2d_to_enclave(const void* uptr, size_t elem_size, size_t elem_cnt1,
                                     size_t elem_cnt2);
+
+#define COPY_UNTRUSTED_VALUE(untrusted_ptr) ({                          \
+    __typeof__(*(untrusted_ptr)) val;                                   \
+    sgx_copy_to_enclave_verified(&val, (untrusted_ptr), sizeof(val));   \
+    val;                                                                \
+})
 
 /*!
  * \brief Low-level wrapper around EREPORT instruction leaf.
--- a/pal/src/host/linux-sgx/enclave_ecalls.c
+++ b/pal/src/host/linux-sgx/enclave_ecalls.c
@@ -4,30 +4,12 @@
 #include "enclave_ecalls.h"
 #include "pal_ecall_types.h"
 #include "pal_linux.h"
-#include "pal_rpc_queue.h"
 #include "sgx_arch.h"
 
 extern uintptr_t g_enclave_base;
 extern uintptr_t g_enclave_top;
 
 static int64_t g_enclave_start_called = 0;
-
-/* returns 0 if rpc_queue is valid/not requested, otherwise -1 */
-static int verify_and_init_rpc_queue(void* untrusted_rpc_queue) {
-    if (!untrusted_rpc_queue) {
-        /* user app didn't request RPC queue (i.e., the app didn't request exitless syscalls) */
-        return 0;
-    }
-
-    if (!sgx_is_valid_untrusted_ptr(untrusted_rpc_queue, sizeof(*g_rpc_queue),
-                                    alignof(__typeof__(*g_rpc_queue)))) {
-        /* malicious RPC queue object, return error */
-        return -1;
-    }
-
-    g_rpc_queue = untrusted_rpc_queue;
-    return 0;
-}
 
 /*
  * Called from enclave_entry.S to execute ecalls.
@@ -93,9 +75,6 @@ void handle_ecall(long ecall_index, void* ecall_args, void* exit_target, void* e
         }
         ms = ecall_args;
 
-        if (verify_and_init_rpc_queue(READ_ONCE(ms->rpc_queue)))
-            return;
-
         /* xsave size must be initialized early, from a trusted source (EREPORT result) */
         // TODO: This eats 1KB of a stack frame which lives for the whole lifespan of this enclave.
         //       We should move it somewhere else and deallocate right after use.
@@ -108,10 +87,14 @@ void handle_ecall(long ecall_index, void* ecall_args, void* exit_target, void* e
         init_xsave_size(report.body.attributes.xfrm);
 
         /* pal_linux_main is responsible for checking the passed arguments */
-        pal_linux_main(READ_ONCE(ms->ms_libpal_uri), READ_ONCE(ms->ms_libpal_uri_len),
-                       READ_ONCE(ms->ms_args), READ_ONCE(ms->ms_args_size), READ_ONCE(ms->ms_env),
-                       READ_ONCE(ms->ms_env_size), READ_ONCE(ms->ms_parent_stream_fd),
-                       READ_ONCE(ms->ms_qe_targetinfo), READ_ONCE(ms->ms_topo_info));
+        pal_linux_main(COPY_UNTRUSTED_VALUE(&ms->ms_libpal_uri),
+                       COPY_UNTRUSTED_VALUE(&ms->ms_libpal_uri_len),
+                       COPY_UNTRUSTED_VALUE(&ms->ms_args), COPY_UNTRUSTED_VALUE(&ms->ms_args_size),
+                       COPY_UNTRUSTED_VALUE(&ms->ms_env), COPY_UNTRUSTED_VALUE(&ms->ms_env_size),
+                       COPY_UNTRUSTED_VALUE(&ms->ms_parent_stream_fd),
+                       COPY_UNTRUSTED_VALUE(&ms->ms_qe_targetinfo),
+                       COPY_UNTRUSTED_VALUE(&ms->ms_topo_info),
+                       COPY_UNTRUSTED_VALUE(&ms->rpc_queue));
     } else {
         // ENCLAVE_START already called (maybe successfully, maybe not), so
         // only valid ecall is THREAD_START.
--- a/pal/src/host/linux-sgx/enclave_framework.c
+++ b/pal/src/host/linux-sgx/enclave_framework.c
@@ -129,13 +129,76 @@ void sgx_reset_ustack(const void* old_ustack) {
     UPDATE_USTACK(old_ustack);
 }
 
+static void copy_u64s(void* dst, const void* untrusted_src, size_t count) {
+    assert((uintptr_t)untrusted_src % 8 == 0);
+    /* The output operands are needed because we cannot specify the same register as both input and
+     * clobbered in GCC inline assembly. */
+    __asm__ volatile (
+        "rep movsq\n"
+        : "+D"(dst), "+S"(untrusted_src), "+c"(count)
+        :  "D"(dst),  "S"(untrusted_src),  "c"(count)
+        : "memory", "cc"
+    );
+}
+
+void sgx_copy_to_enclave_verified(void* ptr, const void* uptr, size_t size) {
+    assert(sgx_is_valid_untrusted_ptr(uptr, size, /*alignment=*/1));
+    assert(sgx_is_completely_within_enclave(ptr, size));
+
+    if (size == 0) {
+        return;
+    }
+
+    /*
+     * This should be simple `memcpy(ptr, uptr, size)`, but CVE-2022-21233 (INTEL-SA-00657).
+     * To mitigate this issue, all reads of untrusted memory from within the enclave must be done
+     * in 8-byte chunks aligned to 8-bytes boundary. Since x64 allocates memory in pages of
+     * (at least) 0x1000 in size, we can safely 8-align the pointer down and the size up.
+     */
+    size_t copy_len;
+    size_t prefix_misalignment = (uintptr_t)uptr & 7;
+    if (prefix_misalignment) {
+        /* Beginning of the copied range is misaligned. */
+        char prefix_val[8] = { 0 };
+        copy_u64s(prefix_val, (char*)uptr - prefix_misalignment, /*count=*/1);
+
+        copy_len = MIN(sizeof(prefix_val) - prefix_misalignment, size);
+        memcpy(ptr, prefix_val + prefix_misalignment, copy_len);
+        ptr = (char*)ptr + copy_len;
+        uptr = (const char*)uptr + copy_len;
+        size -= copy_len;
+
+        if (size == 0) {
+            return;
+        }
+    }
+    assert((uintptr_t)uptr % 8 == 0);
+
+    size_t suffix_misalignment = size & 7;
+    copy_len = size - suffix_misalignment;
+    assert(copy_len % 8 == 0);
+    copy_u64s(ptr, uptr, copy_len / 8);
+    ptr = (char*)ptr + copy_len;
+    uptr = (const char*)uptr + copy_len;
+    size -= copy_len;
+
+    assert(size == suffix_misalignment);
+    if (suffix_misalignment) {
+        /* End of the copied range is misaligned. */
+        char suffix_val[8] = { 0 };
+        copy_u64s(suffix_val, uptr, /*count=*/1);
+        memcpy(ptr, suffix_val, suffix_misalignment);
+    }
+}
+
 bool sgx_copy_to_enclave(void* ptr, size_t maxsize, const void* uptr, size_t usize) {
     if (usize > maxsize ||
         !sgx_is_valid_untrusted_ptr(uptr, usize, /*alignment=*/1) ||
         !sgx_is_completely_within_enclave(ptr, maxsize)) {
         return false;
     }
-    memcpy(ptr, uptr, usize);
+
+    sgx_copy_to_enclave_verified(ptr, uptr, usize);
     return true;
 }
 
--- a/pal/src/host/linux-sgx/enclave_ocalls.c
+++ b/pal/src/host/linux-sgx/enclave_ocalls.c
@@ -134,7 +134,7 @@ static long sgx_exitless_ocall(uint64_t code, void* ms) {
     }
 
     sgx_reset_ustack(old_ustack);
-    return READ_ONCE(req->result);
+    return COPY_UNTRUSTED_VALUE(&req->result);
 }
 
 __attribute_no_sanitize_address
@@ -218,7 +218,7 @@ int ocall_mmap_untrusted(void** addrptr, size_t size, int prot, int flags, int f
         return retval;
     }
 
-    void* returned_addr = READ_ONCE(ms->ms_addr);
+    void* returned_addr = COPY_UNTRUSTED_VALUE(&ms->ms_addr);
     if (flags & MAP_FIXED) {
         /* addrptr already contains the mmap'ed address, no need to update it */
         if (returned_addr != requested_addr) {
@@ -365,10 +365,10 @@ int ocall_cpuid(unsigned int leaf, unsigned int subleaf, unsigned int values[4])
     }
 
     if (!retval) {
-        values[0] = READ_ONCE(ms->ms_values[0]);
-        values[1] = READ_ONCE(ms->ms_values[1]);
-        values[2] = READ_ONCE(ms->ms_values[2]);
-        values[3] = READ_ONCE(ms->ms_values[3]);
+        values[0] = COPY_UNTRUSTED_VALUE(&ms->ms_values[0]);
+        values[1] = COPY_UNTRUSTED_VALUE(&ms->ms_values[1]);
+        values[2] = COPY_UNTRUSTED_VALUE(&ms->ms_values[2]);
+        values[3] = COPY_UNTRUSTED_VALUE(&ms->ms_values[3]);
     }
 
     sgx_reset_ustack(old_ustack);
@@ -1024,7 +1024,7 @@ int ocall_create_process(size_t nargs, const char** args, int* stream_fd) {
 
     if (!retval) {
         if (stream_fd)
-            *stream_fd = READ_ONCE(ms->ms_stream_fd);
+            *stream_fd = COPY_UNTRUSTED_VALUE(&ms->ms_stream_fd);
     }
 
     sgx_reset_ustack(old_ustack);
@@ -1072,7 +1072,7 @@ int ocall_futex(uint32_t* futex, int op, int val, uint64_t* timeout_us) {
     }
 
     if (timeout_us) {
-        uint64_t remaining_time_us = READ_ONCE(ms->ms_timeout_us);
+        uint64_t remaining_time_us = COPY_UNTRUSTED_VALUE(&ms->ms_timeout_us);
         if (retval == -ETIMEDOUT) {
             remaining_time_us = 0;
         }
@@ -1145,7 +1145,7 @@ int ocall_bind(int fd, struct sockaddr_storage* addr, size_t addrlen, uint16_t* 
         goto out;
     }
 
-    uint16_t new_port = READ_ONCE(ms->ms_new_port);
+    uint16_t new_port = COPY_UNTRUSTED_VALUE(&ms->ms_new_port);
     if (new_port == 0) {
         ret = -EPERM;
         goto out;
@@ -1223,7 +1223,7 @@ int ocall_listen(int domain, int type, int protocol, int ipv6_v6only, struct soc
 
     if (retval >= 0) {
         if (addr && len) {
-            size_t untrusted_addrlen = READ_ONCE(ms->ms_addrlen);
+            size_t untrusted_addrlen = COPY_UNTRUSTED_VALUE(&ms->ms_addrlen);
             if (!sgx_copy_to_enclave(addr, len, untrusted_addr, untrusted_addrlen)) {
                 sgx_reset_ustack(old_ustack);
                 return -EPERM;
@@ -1279,7 +1279,7 @@ int ocall_accept(int sockfd, struct sockaddr* addr, size_t* addrlen, struct sock
 
     if (retval >= 0) {
         if (addr && len) {
-            size_t untrusted_addrlen = READ_ONCE(ms->ms_addrlen);
+            size_t untrusted_addrlen = COPY_UNTRUSTED_VALUE(&ms->ms_addrlen);
             if (!sgx_copy_to_enclave(addr, len, untrusted_addr, untrusted_addrlen)) {
                 sgx_reset_ustack(old_ustack);
                 return -EPERM;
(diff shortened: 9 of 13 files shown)


boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/894)* status: 0 of 13 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/894#-NBOz5JO5vhTIxi9VB3E:-NBOz5JO5vhTIxi9VB3F:b-z6wrsq) (no related file):*
I've debugged the failures, but still working on a fix

___

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/894)* status: 0 of 13 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski)

___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 153 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQKOtM85IFSiF-wzVC:-NBQKOtM85IFSiF-wzVD:bw45wu9) ([raw file](https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/pal/src/host/linux-sgx/enclave_framework.c#L153)):*
> ```C
> 
>     /*
>      * This should be simple `memcpy(ptr, uptr, size)`, but CVE-2022-21233 (INTEL-SA-00657).
> ```

Sentence not complete?



_Code quote:_
```C
but CVE-2022-21233 (INTEL-SA-00657)
```
___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 188 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQKVqjDSlcEI7WDdmP:-NBQKVqjDSlcEI7WDdmQ:b-8lxu07) ([raw file](https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/pal/src/host/linux-sgx/enclave_framework.c#L188)):*
> ```C
>     if (suffix_misalignment) {
>         /* End of the copied range is misaligned. */
>         char suffix_val[8] = { 0 };
> ```

Q: is it worth having `prefix_val` and `suffix_val` in one buffer and saving one `memcpy()`?



_Code quote:_
```C
char suffix_val[8] = { 0 }
```
___
*[`pal/src/host/linux-sgx/pal_main.c` line 62 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQH6R45J7_tCImLzNK:-NBQH6R45J7_tCImLzNL:bjz81tt) ([raw file](https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/pal/src/host/linux-sgx/pal_main.c#L62)):*
> ```C
>     }
> 
>     if (!sgx_is_valid_untrusted_ptr(untrusted_rpc_queue, sizeof(*g_rpc_queue),
> ```

Q: should we make the check part earlier, e.g., before copying to the trusted rather than in `pal_linux_main()`, 
___
*[`pal/src/host/linux-sgx/pal_rpc_queue.h` line 5 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQHzlQ1gk_reXKFfz_:-NBQHzlQ1gk_reXKFfza:bnpt6z3) ([raw file](https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/pal/src/host/linux-sgx/pal_rpc_queue.h#L5)):*
> ```objc
>  * functions, like `spinlock_lock`, on that memory. Since we have to use 8-byte naturally aligned
>  * memory accesses to mitigate CVE-2022-21233, we must mark this feature as insecure until we
>  * rewrite it or that CVE is fully fixed in hardware or microcode.
> ```

The full fix in HW or ucode would be not exposing any stale data in XAPIC or disabling xAPIC completely (which maybe quite long-term). Then this "until" does not sound that useful in such case. Also, if fully fixed in HW or ucode, we'd also revert all our SW mitigations introduced in this PR.



_Code quote:_
```objc
or that CVE is fully fixed in hardware or microcode
```

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/894)* status: 0 of 13 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski)

___
*[`-- commits` line 7 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQMGuKAjJxF41issb1):*
Maybe it's better to also mention a subsequent unintentional write may leak that data, cited from the https://www.intel.com/content/www/us/en/developer/articles/technical/software-security-guidance/advisory-guidance/stale-data-read-from-xapic.html:
`The enclave may then unintentionally perform an operation that could allow an attacker to infer this data.`



_Code quote:_
```
  some results of an ocall) such read can potentially return some stale,
  enclave-originated data. Naturally aligned 8-byte reads are not affected
```
___
*[`-- commits` line 14 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQN3XQ7n0v1wJw-o2O):*
Should we put the CVE and SA reference links here?

dimakuv (changes requested):
> Reviewed 12 of 13 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/894)* status: all files reviewed, 15 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski and @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/894#-NBRAHULB_iptDerVpPw:-NBRAHULB_iptDerVpPx:b-q9pjjv) (no related file):*
I guess you made some analysis on the Exitless feature, right? Would be good to post the analysis (even if it was partial) here, in this PR's comments, just for the future reference. Maybe we'll *have to* fix Exitless in the future, and it would be good to have some starting point.

___
___
*[`-- commits` line 7 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQMGuKAjJxF41issb1):*

Maybe it's better to also mention a subsequent unintentional write may leak that data, cited from the https://www.intel.com/content/www/us/en/developer/articles/technical/software-security-guidance/advisory-guidance/stale-data-read-from-xapic.html:
`The enclave may then unintentionally perform an operation that could allow an attacker to infer this data.`
</blockquote></details>

+1, for completeness of description
___
*[`-- commits` line 14 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQN3XQ7n0v1wJw-o2O):*

Should we put the CVE and SA reference links here?
</blockquote></details>

I don't think we need this. We have the CVE and SA numbers in the title of this commit, so people can easily google them.
___
*[`-- commits` line 14 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQyxVZ6rsuV01PeR1P):*
Maybe we should also have one paragraph that these changes can (will be?) reverted in some distant future, when all CPUs in the wild are patched against this vulnerability? Then e.g. we can move Exitless back to "secure" options.
___
*[`Documentation/manifest-syntax.rst` line 539 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQzik4ETyJ7gwzf8ZU:-NBQzik4ETyJ7gwzf8ZV:b-8gr430) ([raw file](https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/Documentation/manifest-syntax.rst#L539)):*
> ```reStructuredText
> 
> This feature is currently marked as insecure, because it reads untrusted memory
> in potentially insecure manner.
> ```

Maybe add `In particular, this feature is succeptible to the CVE-2022-21233 (aka INTEL-SA-00657) vulnerability.`
___
*[`Documentation/performance.rst` line 167 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQzzEkBmi5Cizhhb6p:-NBQzzEkBmi5Cizhhb6q:b-8gr430) ([raw file](https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/Documentation/performance.rst#L167)):*
> ```reStructuredText
> 
> Note this feature is currently insecure and not recommended for production
> usage.
> ```

Maybe add `In particular, this feature is succeptible to the CVE-2022-21233 (aka INTEL-SA-00657) vulnerability.`
___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 153 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQKOtM85IFSiF-wzVC:-NBR35PC6E07PGgNRbf1:bik4if2) ([raw file](https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/pal/src/host/linux-sgx/enclave_framework.c#L153)):*

Sentence not complete?
</blockquote></details>

I think this was intended :) I don't mind this shortened sentence, it's clear that it means "but see CVE... for details"
___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 180 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBR5D-sC_a5XMCvIiAK:-NBR5D-sC_a5XMCvIiAL:b2qztu0) ([raw file](https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/pal/src/host/linux-sgx/enclave_framework.c#L180)):*
> ```C
>     copy_len = size - suffix_misalignment;
>     assert(copy_len % 8 == 0);
>     copy_u64s(ptr, uptr, copy_len / 8);
> ```

At this point, `copy_len` may be `0`. Do we want to move these 4 lines under `if (copy_len) { ... }`?

I know that this doesn't change anything semantically, but it will make more clear that there is such a special case possible ("the size to copy is less than 8 bytes, so we skip the middle logic of this function")
___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 188 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQKVqjDSlcEI7WDdmP:-NBR6BTxAmsqwQn09EzX:b3tqz0d) ([raw file](https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/pal/src/host/linux-sgx/enclave_framework.c#L188)):*

Q: is it worth having `prefix_val` and `suffix_val` in one buffer and saving one `memcpy()`?
</blockquote></details>

I don't understand what @kailun-qin means. I don't see how we can safe one `memcpy()`. All the logic here seems correct and already minimal.
___
*[`pal/src/host/linux-sgx/pal_main.c` line 62 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQH6R45J7_tCImLzNK:-NBR7Cru5xTQi3EPJr59:bemk6t) ([raw file](https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/pal/src/host/linux-sgx/pal_main.c#L62)):*

Q: should we make the check part earlier, e.g., before copying to the trusted rather than in `pal_linux_main()`, 
</blockquote></details>

@kailun-qin I don't see a need for this. Who cares if we copied the "bad" pointer into the enclave? We'll check this "bad" pointer now and we will fail the whole Gramine process if it's really bad.

Also, I'm unclear what you mean by "before copying to the trusted". The only thing that we copy to the trusted enclave memory is the pointer value itself. Gramine never copies the RPC queue itself -- it is always outside of the enclave memory.
___
*[`pal/src/host/linux-sgx/pal_main.c` line 588 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBR7hXCBmVJZ66-hzaW:-NBR7hXD5xPfMt1Z8u7w:b-5zke8k) ([raw file](https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/pal/src/host/linux-sgx/pal_main.c#L588)):*
> ```C
>     g_pal_common_state.raw_manifest_data = manifest_addr;
>     g_pal_public_state.manifest_root = manifest_root;
> ```

Please remove one empty line
___
*[`pal/src/host/linux-sgx/pal_main.c` line 592 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBR97pF2aO0VLbe8RRh:-NBR97pF2aO0VLbe8RRi:bd245wh) ([raw file](https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/pal/src/host/linux-sgx/pal_main.c#L592)):*
> ```C
>     int64_t rpc_thread_num;
>     ret = toml_int_in(g_pal_public_state.manifest_root, "sgx.insecure__rpc_thread_num",
>                       /*defaultval=*/0, &rpc_thread_num);
> ```

Looks like you added this read-from-manifest to prevent the malicious host from lying about the RPC queue? This is good, we should have had it before, but we missed this...
___
*[`pal/src/host/linux-sgx/pal_rpc_queue.h` line 2 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBR9kWN8bmOzJ88gU6Q:-NBR9kWO0f2gEzKNdnT1:btmi0uf) ([raw file](https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/pal/src/host/linux-sgx/pal_rpc_queue.h#L2)):*
> ```objc
> /*
>  * FIXME: This feature is currently insecure. It keeps most object in untrusted memory and uses some
> ```

`objects`
___
*[`pal/src/host/linux-sgx/pal_rpc_queue.h` line 4 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBR9nUIDkzUDrZGLsIf:-NBR9nUIDkzUDrZGLsIg:b-eu3mbw) ([raw file](https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/pal/src/host/linux-sgx/pal_rpc_queue.h#L4)):*
> ```objc
>  * FIXME: This feature is currently insecure. It keeps most object in untrusted memory and uses some
>  * functions, like `spinlock_lock`, on that memory. Since we have to use 8-byte naturally aligned
>  * memory accesses to mitigate CVE-2022-21233, we must mark this feature as insecure until we
> ```

`...memory accesses to untrusted memory ...`. This makes it more concrete.
___
*[`pal/src/host/linux-sgx/pal_rpc_queue.h` line 5 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQHzlQ1gk_reXKFfz_:-NBRA4595fKFVqKXfOC4:b78m3ui) ([raw file](https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/pal/src/host/linux-sgx/pal_rpc_queue.h#L5)):*

The full fix in HW or ucode would be not exposing any stale data in XAPIC or disabling xAPIC completely (which maybe quite long-term). Then this "until" does not sound that useful in such case. Also, if fully fixed in HW or ucode, we'd also revert all our SW mitigations introduced in this PR.
</blockquote></details>

I agree with Kailun's text, but I don't see what we can change in this comment. I think the comment is fine as-is.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/894)* status: all files reviewed, 13 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski, @dimakuv, and @kailun-qin)

___
*[`-- commits` line 14 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQN3XQ7n0v1wJw-o2O):*

I don't think we need this. We have the CVE and SA numbers in the title of this commit, so people can easily google them.
</blockquote></details>

Yeah, make sense, thanks!
___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 153 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQKOtM85IFSiF-wzVC:-NBRI77f1bVJGV0Tiad4:b-owq44j) ([raw file](https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/pal/src/host/linux-sgx/enclave_framework.c#L153)):*

I think this was intended :) I don't mind this shortened sentence, it's clear that it means "but see CVE... for details"
</blockquote></details>

OK ;) I'm fine w/ it.
___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 188 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQKVqjDSlcEI7WDdmP:-NBRIAeC-JuiuUJ6ZFei:bmjjddj) ([raw file](https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/pal/src/host/linux-sgx/enclave_framework.c#L188)):*
> I don't understand what @kailun-qin means. I don't see how we can safe one memcpy(). All the logic here seems correct and already minimal.

I was just thinking about having a `buf[16] = { 0 }` and having both `prefix` and `suffix` (or only one of them) copy into it, and then having one single `memcpy()` at the end.
___
*[`pal/src/host/linux-sgx/pal_main.c` line 62 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQH6R45J7_tCImLzNK:-NBRJB2_BRIW3hmhbtZb:b8de42r) ([raw file](https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/pal/src/host/linux-sgx/pal_main.c#L62)):*
> @kailun-qin I don't see a need for this. Who cares if we copied the "bad" pointer into the enclave? We'll check this "bad" pointer now and we will fail the whole Gramine process if it's really bad.

OK, make sense to me.

> Also, I'm unclear what you mean by "before copying to the trusted".

I was just thinking about moving this check before: https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/pal/src/host/linux-sgx/enclave_ecalls.c#L97

as it seems that we had some assumptions here (though not exactly the same purpose):
https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/pal/src/host/linux-sgx/enclave_framework.c#L145.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/894)* status: all files reviewed, 13 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski, @dimakuv, and @kailun-qin)

___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 188 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQKVqjDSlcEI7WDdmP:-NBRLyaR2f8K3ZzHFJAz:bfys4x3) ([raw file](https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/pal/src/host/linux-sgx/enclave_framework.c#L188)):*

> I don't understand what @kailun-qin means. I don't see how we can safe one memcpy(). All the logic here seems correct and already minimal.

I was just thinking about having a `buf[16] = { 0 }` and having both `prefix` and `suffix` (or only one of them) copy into it, and then having one single `memcpy()` at the end.
</blockquote></details>

Update: just rethink more thoroughly, yep, the logic here seems already minimal... Excuse me for the distraction here.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/894)* status: all files reviewed, 11 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/894#-NBRAHULB_iptDerVpPw:-NBRVqNHEIeGoMCd0P-D:b-r9cyj0) (no related file):*

I guess you made some analysis on the Exitless feature, right? Would be good to post the analysis (even if it was partial) here, in this PR's comments, just for the future reference. Maybe we'll *have to* fix Exitless in the future, and it would be good to have some starting point.
</blockquote></details>

I just looked at the code, if that's what you mean.
It touches untrusted memory in many places, most of them are not marked as such and hidden behind functions. So I guess there are two cases:
1) Normal untrusted memory access - would have to be changed to copying the value to local variable and only then using it, I guess. Doable, but very annoying.
2) Spinlocks on untrusted memory - I think easiest way to solve it would be to inline all of their code, change the underlying word to 64bits (it's `uint32_t` now) and make sure it's always 8bytes aligned.

Also there are other possible issues, e.g. some values are read multiple times and they are expected to stay the same. In the one place I looked at it was most likely fine (i.e. didn't result in any vulnerabilities), but I think we should just look closely at all of it, once we decide to bring it back. To speed things up I did not bother for now, because I got a green light on marking this insecure (at least temporarily). I think it's all fine, but wouldn't hurt to verify.

___
___
*[`-- commits` line 7 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQMGuKAjJxF41issb1):*

+1, for completeness of description
</blockquote></details>

How about I add a sentence between these two (i.e. `...enclave-originated data. (here goes new sentence). Naturally aligned...`):
`This data may then be unintentionally leaked by the enclave, e.g. written back to untrusted part or handled in non-side-channel-secure manner.`
___
*[`-- commits` line 14 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQN3XQ7n0v1wJw-o2O):*

Yeah, make sense, thanks!
</blockquote></details>

Yes, I intentionally didn't want to put links, because they often get stale.
___
*[`-- commits` line 14 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQyxVZ6rsuV01PeR1P):*

Maybe we should also have one paragraph that these changes can (will be?) reverted in some distant future, when all CPUs in the wild are patched against this vulnerability? Then e.g. we can move Exitless back to "secure" options.
</blockquote></details>

Well, if somebody wants this feature, then we should fix it much sooner. If nobody uses it, then there is no point and we could completely deprecate it?
___
*[`Documentation/manifest-syntax.rst` line 539 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQzik4ETyJ7gwzf8ZU:-NBRaxGzE1b8GB2kZzEi:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/Documentation/manifest-syntax.rst#L539)):*

Maybe add `In particular, this feature is succeptible to the CVE-2022-21233 (aka INTEL-SA-00657) vulnerability.`
</blockquote></details>

Done.
___
*[`Documentation/performance.rst` line 167 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQzzEkBmi5Cizhhb6p:-NBRbDJo5Dvp0H-KNJTL:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/Documentation/performance.rst#L167)):*

Maybe add `In particular, this feature is succeptible to the CVE-2022-21233 (aka INTEL-SA-00657) vulnerability.`
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 153 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQKOtM85IFSiF-wzVC:-NBRbH-aEkgRPFF4wk-H:b-3pto2a) ([raw file](https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/pal/src/host/linux-sgx/enclave_framework.c#L153)):*

OK ;) I'm fine w/ it.
</blockquote></details>

Yes, this was intended
___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 180 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBR5D-sC_a5XMCvIiAK:-NBRblG-0UJl-FkC0Hqd:b-k6yafu) ([raw file](https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/pal/src/host/linux-sgx/enclave_framework.c#L180)):*

At this point, `copy_len` may be `0`. Do we want to move these 4 lines under `if (copy_len) { ... }`?

I know that this doesn't change anything semantically, but it will make more clear that there is such a special case possible ("the size to copy is less than 8 bytes, so we skip the middle logic of this function")
</blockquote></details>

I don't mind either. This code works correctly for `copy_len == 0` and it's unclear to me which would be faster, an if or couple of adds/subs. So this would be purely visual / for code clarity change.
Let's hear from others which version they would prefer.
___
*[`pal/src/host/linux-sgx/pal_main.c` line 62 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQH6R45J7_tCImLzNK:-NBRckhs-700IgCJ_lYo:b76rahe) ([raw file](https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/pal/src/host/linux-sgx/pal_main.c#L62)):*
> I was just thinking about moving this check before: 

How do you check a pointer without actually having it (i.e. copying it into enclave - the pointer, not the underlying object)?

> as it seems that we had some assumptions here (though not exactly the same purpose):

I don't understand your point, this is exactly the same check.
___
*[`pal/src/host/linux-sgx/pal_main.c` line 588 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBR7hXCBmVJZ66-hzaW:-NBRdcwPDWyG4MUrUsIL:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/pal/src/host/linux-sgx/pal_main.c#L588)):*

Please remove one empty line
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/pal_main.c` line 592 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBR97pF2aO0VLbe8RRh:-NBRdeoq-pSuywZIazis:b-hx0jjg) ([raw file](https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/pal/src/host/linux-sgx/pal_main.c#L592)):*

Looks like you added this read-from-manifest to prevent the malicious host from lying about the RPC queue? This is good, we should have had it before, but we missed this...
</blockquote></details>

Correct. Before it wasn't required per se (assuming the implementation is correct, this could only impact performance), but now this is a must have.
___
*[`pal/src/host/linux-sgx/pal_rpc_queue.h` line 2 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBR9kWN8bmOzJ88gU6Q:-NBRe05cEyG2E4uu0fMg:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/pal/src/host/linux-sgx/pal_rpc_queue.h#L2)):*

`objects`
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/pal_rpc_queue.h` line 4 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBR9nUIDkzUDrZGLsIf:-NBReCCz23usyTdEVAGB:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/pal/src/host/linux-sgx/pal_rpc_queue.h#L4)):*

`...memory accesses to untrusted memory ...`. This makes it more concrete.
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/pal_rpc_queue.h` line 5 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQHzlQ1gk_reXKFfz_:-NBReNku6w5FggBFVvyV:b74hu6k) ([raw file](https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/pal/src/host/linux-sgx/pal_rpc_queue.h#L5)):*

I agree with Kailun's text, but I don't see what we can change in this comment. I think the comment is fine as-is.
</blockquote></details>

We have no fixes in exitless code, that's the whole point (and that's why we mark it insecure).

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/894)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @kailun-qin, and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/894#-NBRAHULB_iptDerVpPw:-NBRmVrz0lqpo5-YhYdu:b-1xms1w) (no related file):*

I just looked at the code, if that's what you mean.
It touches untrusted memory in many places, most of them are not marked as such and hidden behind functions. So I guess there are two cases:
1) Normal untrusted memory access - would have to be changed to copying the value to local variable and only then using it, I guess. Doable, but very annoying.
2) Spinlocks on untrusted memory - I think easiest way to solve it would be to inline all of their code, change the underlying word to 64bits (it's `uint32_t` now) and make sure it's always 8bytes aligned.

Also there are other possible issues, e.g. some values are read multiple times and they are expected to stay the same. In the one place I looked at it was most likely fine (i.e. didn't result in any vulnerabilities), but I think we should just look closely at all of it, once we decide to bring it back. To speed things up I did not bother for now, because I got a green light on marking this insecure (at least temporarily). I think it's all fine, but wouldn't hurt to verify.
</blockquote></details>

Thank you, this is exactly the analysis I wanted us to capture here.

___
___
*[`-- commits` line 7 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQMGuKAjJxF41issb1):*

How about I add a sentence between these two (i.e. `...enclave-originated data. (here goes new sentence). Naturally aligned...`):
`This data may then be unintentionally leaked by the enclave, e.g. written back to untrusted part or handled in non-side-channel-secure manner.`
</blockquote></details>

Looks good to me.
___
*[`-- commits` line 14 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQyxVZ6rsuV01PeR1P):*

Well, if somebody wants this feature, then we should fix it much sooner. If nobody uses it, then there is no point and we could completely deprecate it?
</blockquote></details>

But my comment was generally about this commit, not about Exitless. Like, in 2 years from now, we won't need that contrived logic in `sgx_copy_to_enclave_verified()`, and it would make sense to remove it again. That's what I wanted to highlight in the commit text.
___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 180 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBR5D-sC_a5XMCvIiAK:-NBRnnP28GhCOwnR9sDd:bcryzdf) ([raw file](https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/pal/src/host/linux-sgx/enclave_framework.c#L180)):*

I don't mind either. This code works correctly for `copy_len == 0` and it's unclear to me which would be faster, an if or couple of adds/subs. So this would be purely visual / for code clarity change.
Let's hear from others which version they would prefer.
</blockquote></details>

The others are @mkow, let's ping him :)
___
*[`pal/src/host/linux-sgx/pal_main.c` line 62 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQH6R45J7_tCImLzNK:-NBRnSgTCCNkh1sgbJ13:b-o47ab5) ([raw file](https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/pal/src/host/linux-sgx/pal_main.c#L62)):*

> I was just thinking about moving this check before: 

How do you check a pointer without actually having it (i.e. copying it into enclave - the pointer, not the underlying object)?

> as it seems that we had some assumptions here (though not exactly the same purpose):

I don't understand your point, this is exactly the same check.
</blockquote></details>

Yes, I agree with Borys. I don't see any problems here. And again -- we only copy and check the **pointer value**, not the **underlying object this pointer points to**.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/894)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, @kailun-qin, and @mkow)

___
*[`-- commits` line 14 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQyxVZ6rsuV01PeR1P):*

But my comment was generally about this commit, not about Exitless. Like, in 2 years from now, we won't need that contrived logic in `sgx_copy_to_enclave_verified()`, and it would make sense to remove it again. That's what I wanted to highlight in the commit text.
</blockquote></details>

So you want me to add a sentence saying something like "when this issues is fixed in HW we will remove our mitigations"?

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/894)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @kailun-qin, and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/894#-NBZ3Ua_-osReAv3b7dA:-NBZ3Ua_-osReAv3b7dB:b9rupzq) (no related file):*
TODO: add this to list of the breaking changes, if we merge this `insecure__` prefix addition

___

mkow (changes requested):
> Reviewed 8 of 13 files at r1, 1 of 1 files at r2, 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/894)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @kailun-qin)

___
*[`-- commits` line 7 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQMGuKAjJxF41issb1):*

Looks good to me.
</blockquote></details>

+1 to the addition proposed by Borys.
___
*[`-- commits` line 13 at r3](https://reviewable.io/reviews/gramineproject/gramine/894#-NBqKI4b2GgzEHXweFyk):*
Could you also mention that there was a bug that untrusted host could force Gramine to use exitless, despite the app didn't want it? It's ofc not a security issue after we marked exitless as insecure, but IMO worth to mention this change here.
___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 180 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBR5D-sC_a5XMCvIiAK:-NBo_Ei87gJfkMwC3jZi:bp7xq02) ([raw file](https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/pal/src/host/linux-sgx/enclave_framework.c#L180)):*

The others are @mkow, let's ping him :)
</blockquote></details>

I don't see why zero would be a special case here, it's a normal size and everything works correctly on it.
___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 134 at r3](https://reviewable.io/reviews/gramineproject/gramine/894#-NBqBIn41aSfQmutGhQF:-NBqBIn599LBGvotjh3g:b-m722xp) ([raw file](https://github.com/gramineproject/gramine/blob/c05ea6c64f2b7c555f3ddc027b62b3a13fccf6d0/pal/src/host/linux-sgx/enclave_framework.c#L134)):*
> ```C
> static void copy_u64s(void* dst, const void* untrusted_src, size_t count) {
>     assert((uintptr_t)untrusted_src % 8 == 0);
>     /* The output operands are needed because we cannot specify the same register as both input and
> ```

I'd reword this to say it's just the GAS inline assembly syntax which doesn't support this, in this version it's unclear why we can't do this.



_Code quote:_
```C
we cannot specify
```
___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 187 at r3](https://reviewable.io/reviews/gramineproject/gramine/894#-NBqIbmUETCcmvVska_A:-NBqIbmV20ONEMDeR2Gh:bp7br0o) ([raw file](https://github.com/gramineproject/gramine/blob/c05ea6c64f2b7c555f3ddc027b62b3a13fccf6d0/pal/src/host/linux-sgx/enclave_framework.c#L187)):*
> ```C
>     assert(size == suffix_misalignment);
>     if (suffix_misalignment) {
>         /* End of the copied range is misaligned. */
> ```

This case is actually different than `prefix_misalignment` - here the _length_ of the suffix is misaligned, not the end of the range. And currently you have this comment exactly the same as under `prefix_misalignment` case.



_Code quote:_
```C
/* End of the copied range is misaligned. */
```

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/894)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/894#-NBZ3Ua_-osReAv3b7dA:-NBqSds6C-6vsLzeLNh_:bm26bt7) (no related file):*

TODO: add this to list of the breaking changes, if we merge this `insecure__` prefix addition
</blockquote></details>

Done (in the draft version)

___
___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 180 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBR5D-sC_a5XMCvIiAK:-NBqSWsq7G3Y9ZbyZLnN:b-wv7diu) ([raw file](https://github.com/gramineproject/gramine/blob/9fc3749bc6762a54be49df76883bd5d37ee8d956/pal/src/host/linux-sgx/enclave_framework.c#L180)):*

I don't see why zero would be a special case here, it's a normal size and everything works correctly on it.
</blockquote></details>

OK, then let's leave the code as is.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/894)* status: all files reviewed, 3 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: ITL), "squash! " found in commit messages' one-liners (waiting on @dimakuv and @kailun-qin)

___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 187 at r3](https://reviewable.io/reviews/gramineproject/gramine/894#-NBqIbmUETCcmvVska_A:-NBr6SNJ-9yVMNRa6FzV:b-oag90k) ([raw file](https://github.com/gramineproject/gramine/blob/c05ea6c64f2b7c555f3ddc027b62b3a13fccf6d0/pal/src/host/linux-sgx/enclave_framework.c#L187)):*

I'm confused. What's the difference? Here the start is aligned and the end is misaligned. Yes, this means that the length is misaligned (length of the suffix, not necessarily the total length).

> And currently you have this comment exactly the same as under prefix_misalignment case.

Yes, and I believe these cases are mostly the same (length < 8, being or end misaligned)
</blockquote></details>

Ahhh, by "end" here you mean the real end of the string. Somehow I read this as that "suffix chunk", my bad.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/894)* status: all files reviewed, 2 unresolved discussions, "squash! " found in commit messages' one-liners (waiting on @kailun-qin)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/894)* status: all files reviewed, 2 unresolved discussions (waiting on @kailun-qin)

mkow (approved):
> Dismissed @kailun-qin from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/894)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`-- commits` line 7 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQMGuKAjJxF41issb1):*

Done.
</blockquote></details>

Dismissing @kailun-qin from this discussion because it was already fixed and he didn't reply since then (and he's not a maintainer to have veto rights ;) ).
___
*[`-- commits` line 14 at r2](https://reviewable.io/reviews/gramineproject/gramine/894#-NBQN3XQ7n0v1wJw-o2O):*

Yes, I intentionally didn't want to put links, because they often get stale.
</blockquote></details>

ditto

Issue #800, opened 2022-07-27
Add support for KSS (key separation and sharing)


DL8:
KSS (key separation and sharing) provides additional fields to SECS and EGETKEY, which allows more fine-grained control over remote attestation and key derivation. Some examples where it might be useful:
1. The same enclave binary may be executed in the context of two different processes (they might even belong to different users). They will have the same MRSIGNER and MRENCLAVE, so without they will derive the same keys, which might lead to unintended data disclosure
2. A remote party may want to attest not only which binary is run, but also which specific configuration is used (CONFIGID and CONFIGSVN)
3. An application may consist of several enclaves ("microservices"), which may want to share keys (e.g. for shared protected storage)

Several places where changes may be required (not a complete list and open for discussion):
1. The loader should support enumeration (CPUID) and setting new SECS attributes (CONFIGID, CONFIGSVN, ISVEXTPRODID, ISVFAMILYID)
2. A mechanism to define KSS requirement should be added to the manifest (one option: must enable, may enable, must disable)
3. EGETKEY wrapper should support new key policy attributes and new fields in key request structure (e.g. CONFIGSVN)

mkow:
> The same enclave binary may be executed in the context of two different processes (they might even belong to different users). They will have the same MRSIGNER and MRENCLAVE, so without they will derive the same keys, which might lead to unintended data disclosure

I don't think this is the case in Gramine model:
- The entrypoint path and contents are measured in almost all cases (the only exception is when it's provided via encrypted files, but then see below).
- The true "identity" of the enclave will always be its MRENCLAVE in Gramine model. If we change the identity on `execve()` from "Gramine ver XYZ+binary QWER" to "Gramine ver XYZ+binary ASDF" then it may seem that the current identity is really "Gramine ver XYZ+binary ASDF", while in fact QWER could make persistent changes inside the enclave and still be in control. This is because Gramine doesn't provide in-enclave sandboxing (this is technically infeasible on SGX).

> A remote party may want to attest not only which binary is run, but also which specific configuration is used (CONFIGID and CONFIGSVN)

This sounds more relevant to Gramine, but why wouldn't the config be placed in the manifest? Could you describe some real-world use case for this?

> An application may consist of several enclaves ("microservices"), which may want to share keys (e.g. for shared protected storage)

But they already can, via our remote attestation APIs? This seems unrelated to KSS, it's just about attestation?

kailun-qin:
[...]

PR #903, opened 2022-09-12
[python,ci-examples] Support SGX extended product ID

DL8:
KSS is a new feature that introduces additional fields to identify enclave in build time: ISVEXTPRODID and ISVFAMILYID. These fields are defined the the developer and signed in SIGSTRUCT. Add new fields to the manifest to add support, and support in SGX signing tool.

Add a hello world example with KSS enabled: Product ID fields are configured in the manifest and the enclave prints them. This example must be built with remote attestation.

At this point SGX loader is not updated and there are no sanity checks prior to enclave load. Therefore, if an enclave with KSS is loaded on an unsupported platform, it might crash.

Signed-off-by: Lavi, Nir <nir.lavi@intel.com>

## Description of the changes 

The following fields were added to the manifest:
```
sgx.kss = [true|false] (default: false)
sgx.isvextprodid = "[16-byte hex value]"  (default: "0x00000000000000000000000000000000")
sgx.isvfamily    = "[16-byte hex value]"  (default: "0x00000000000000000000000000000000")
```

if `sgx.kss = true`, SGX signing tool will do the following in SIGSTRUCT:
1. Set `attributes.kss = 1`
2. Set `isvextprodid` and isvfamilyid` with values from the manifest

This is a partial fix for #800 . It does not address the following:
1. CONFIGID handling
2. RA-TLS verification callback changes

## How to test this PR? 

Build and run `kss-helloworld` example on a platform with SGX and KSS support. The following should be in the output:
```
isvextprodid = feca0df0feca0df00df0feca0df0feca
isvfamilyid = ffeeddccbbaa99887766554433221100
```

If the platform has SGX enabled but does not support KSS, enclave should not load


donporter (changes requested):
> Reviewed 1 of 12 files at r1.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/903)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @DL8)

___
*[`CI-Examples/kss-helloworld/kss-helloworld.c` line 33 at r1](https://reviewable.io/reviews/gramineproject/gramine/903#-NLLwR3P2LQLnYz46X7s:-NLQkZJ_BxQFSkku8pKC:bhdpv7w) ([raw file](https://github.com/gramineproject/gramine/blob/fa954c93f4aef629624a1647368b7ca7db98b3a4/CI-Examples/kss-helloworld/kss-helloworld.c#L33)):*
<details><summary><i>Previously, DL8 (Nir) wrote…</i></summary><blockquote>

this example uses /dev/attestation/report, so we want ot make sure it's loaded. it will be loaded if remote attestation is enabled, but we don't really care which mechanism is used
</blockquote></details>

I see.  Perhaps add a comment to this effect for future maintenance.

DL8 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/903)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @donporter)

___
*[`CI-Examples/kss-helloworld/kss-helloworld.c` line 33 at r1](https://reviewable.io/reviews/gramineproject/gramine/903#-NLLwR3P2LQLnYz46X7s:-NLQoxWg5vxOkG3VLjCI:b-d7bhn0) ([raw file](https://github.com/gramineproject/gramine/blob/fa954c93f4aef629624a1647368b7ca7db98b3a4/CI-Examples/kss-helloworld/kss-helloworld.c#L33)):*
<details><summary><i>Previously, donporter (Don Porter) wrote…</i></summary><blockquote>

I see.  Perhaps add a comment to this effect for future maintenance.
</blockquote></details>

will do

DL8 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/903)* status: 4 of 13 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @donporter)

___
*[`CI-Examples/kss-helloworld/kss-helloworld.c` line 33 at r1](https://reviewable.io/reviews/gramineproject/gramine/903#-NLLwR3P2LQLnYz46X7s:-NLW2C696vYNfvQjXEFE:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/fa954c93f4aef629624a1647368b7ca7db98b3a4/CI-Examples/kss-helloworld/kss-helloworld.c#L33)):*
<details><summary><i>Previously, DL8 (Nir) wrote…</i></summary><blockquote>

will do
</blockquote></details>

Done.
___
*[`CI-Examples/kss-helloworld/kss-helloworld.c` line 59 at r1](https://reviewable.io/reviews/gramineproject/gramine/903#-NLM0HiO3KloMvON8KwK:-NLW2Dnw-8Dh0i9A51La:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/fa954c93f4aef629624a1647368b7ca7db98b3a4/CI-Examples/kss-helloworld/kss-helloworld.c#L59)):*
<details><summary><i>Previously, DL8 (Nir) wrote…</i></summary><blockquote>

will do
</blockquote></details>

Done.

DL8 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/903)* status: 4 of 13 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @donporter)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/903#-NLLvwAn1bnD4Fyq57Qq:-NLW2JYy4Zn04-fyciCL:bfg93tt) (no related file):*
<details><summary><i>Previously, DL8 (Nir) wrote…</i></summary><blockquote>

loading on a platform without support will probably lead to a crash due to a reserved bit being set in SECS.ATTRIBUTES. this scenario is not covered in this PR
</blockquote></details>

i tried loading an enclave with KSS on a platform without this feature. the result is `ECREATE` and the loader fails gracefully. is this behavior acceptable?

___

donporter (changes requested):
> Reviewed 9 of 9 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/903)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @DL8)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/903#-NLLvwAn1bnD4Fyq57Qq:-NLv5ZvO74h_EMayBVAR:b-h54nlb) (no related file):*
<details><summary><i>Previously, DL8 (Nir) wrote…</i></summary><blockquote>

i tried loading an enclave with KSS on a platform without this feature. the result is `ECREATE` and the loader fails gracefully. is this behavior acceptable?
</blockquote></details>

Can you show an example of what you mean, by "the result is ECREATE" or sample output of the loader failing?  I hope that the output gives the user a clear indication of what the problem is and how to learn more.

___
___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 1194 at r1](https://reviewable.io/reviews/gramineproject/gramine/903#-NLv7hY_6nA6pM1SHQ_S:-NLv7hY_6nA6pM1SHQ_T:b-mneuu) ([raw file](https://github.com/gramineproject/gramine/blob/fa954c93f4aef629624a1647368b7ca7db98b3a4/pal/src/host/linux-sgx/enclave_framework.c#L1194)):*
> ```C
>     assert(stream->hdr.type == PAL_TYPE_PROCESS);
> 
>     memset(&target_info, 0, sizeof(target_info));
> ```

I worry that relying on get_current_enclave_target_info to always zero out all fields is prone to error, since it point-initializes fields and there could be an ambiguous expectation between caller and callee.

Assuming no other maintainers object, I would be ok with adding comments documenting an explicit pre/post condition that the caller need not initialize the tartet_info and that the post-condition is that all fields will be initialized, potentially to zero (and ensuring that this is what the current code does).

DL8 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/903)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @donporter)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/903#-NLLvwAn1bnD4Fyq57Qq:-NLv9Mf_8-3No_Q-ahWK:b-ip5sc1) (no related file):*
<details><summary><i>Previously, donporter (Don Porter) wrote…</i></summary><blockquote>

Can you show an example of what you mean, by "the result is ECREATE" or sample output of the loader failing?  I hope that the output gives the user a clear indication of what the problem is and how to learn more.
</blockquote></details>

at the beginning i didn't have a platform without KSS, so i couldn't test this scenario. according to the SDM, `ECREATE` throws an exception in this case, so i assumed the loader would crash. however, now i have a machine without kss supported so i tried it and saw that it results in graceful failure. the output itself is not very informative:
```
$ gramine-sgx ./kss-helloworld
error: ECREATE failed in enclave creation ioctl (errno = -5)
error: Creating enclave failed: -5
error: load_enclave() failed with error -5
```

___
___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 1194 at r1](https://reviewable.io/reviews/gramineproject/gramine/903#-NLv7hY_6nA6pM1SHQ_S:-NLvCXRRAfFoYh1x-tqa:bh0spq7) ([raw file](https://github.com/gramineproject/gramine/blob/fa954c93f4aef629624a1647368b7ca7db98b3a4/pal/src/host/linux-sgx/enclave_framework.c#L1194)):*
<details><summary><i>Previously, donporter (Don Porter) wrote…</i></summary><blockquote>

I worry that relying on get_current_enclave_target_info to always zero out all fields is prone to error, since it point-initializes fields and there could be an ambiguous expectation between caller and callee.

Assuming no other maintainers object, I would be ok with adding comments documenting an explicit pre/post condition that the caller need not initialize the tartet_info and that the post-condition is that all fields will be initialized, potentially to zero (and ensuring that this is what the current code does).
</blockquote></details>

i think this discussion should be in a separate PR/issue

donporter (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/903)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @DL8)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/903#-NLLvwAn1bnD4Fyq57Qq:-NLzwuGq-sTrKHJQODML:b-8mcakj) (no related file):*
<details><summary><i>Previously, DL8 (Nir) wrote…</i></summary><blockquote>

at the beginning i didn't have a platform without KSS, so i couldn't test this scenario. according to the SDM, `ECREATE` throws an exception in this case, so i assumed the loader would crash. however, now i have a machine without kss supported so i tried it and saw that it results in graceful failure. the output itself is not very informative:
```
$ gramine-sgx ./kss-helloworld
error: ECREATE failed in enclave creation ioctl (errno = -5)
error: Creating enclave failed: -5
error: load_enclave() failed with error -5
```
</blockquote></details>

Is there a way to make this message more informative?

___
___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 1194 at r1](https://reviewable.io/reviews/gramineproject/gramine/903#-NLv7hY_6nA6pM1SHQ_S:-NLzwyf85fwA4SiOPNyl:b-dwah7o) ([raw file](https://github.com/gramineproject/gramine/blob/fa954c93f4aef629624a1647368b7ca7db98b3a4/pal/src/host/linux-sgx/enclave_framework.c#L1194)):*
<details><summary><i>Previously, DL8 (Nir) wrote…</i></summary><blockquote>

i think this discussion should be in a separate PR/issue
</blockquote></details>

Perhaps in the interim we should not remove the memset, then?

donporter (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/903)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @DL8)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/903#-NLLvwAn1bnD4Fyq57Qq:-NMTz29y5fKdnMitCPCz:bd3x35i) (no related file):*
<details><summary><i>Previously, DL8 (Nir) wrote…</i></summary><blockquote>

the error code can be translated with `strerror()`. a more informative solution will require adding an explicit check to the loader, but then why do we explicitly check this specific case but not others (e.g. an unsupported feature is requested, invalid launch token, invalid enclave state and so on)?
</blockquote></details>

Ideally, we would check them all.  I hate to burden you if this is a big ask; even a placeholder to slot in more informative messages for other cases would be an improvement.

___
___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 1194 at r1](https://reviewable.io/reviews/gramineproject/gramine/903#-NLv7hY_6nA6pM1SHQ_S:-NMTzdDY8FxpuWQV-cLn:b-766d6a) ([raw file](https://github.com/gramineproject/gramine/blob/fa954c93f4aef629624a1647368b7ca7db98b3a4/pal/src/host/linux-sgx/enclave_framework.c#L1194)):*
<details><summary><i>Previously, DL8 (Nir) wrote…</i></summary><blockquote>

looks like this memset was removed in commit `652d08af5d56a5396d707700a9c8ca43ba6c3dfb`
</blockquote></details>

I see.  I misunderstood the diff.

DL8 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/903)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @donporter)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/903#-NLLvwAn1bnD4Fyq57Qq:-NN1qhm4EuItDbcC7Ndl:bgra6av) (no related file):*
<details><summary><i>Previously, donporter (Don Porter) wrote…</i></summary><blockquote>

Ideally, we would check them all.  I hate to burden you if this is a big ask; even a placeholder to slot in more informative messages for other cases would be an improvement.
</blockquote></details>

do you think this PR is a good reference?
https://github.com/gramineproject/gramine/pull/1149

___

PR #929, opened 2022-09-23
[LibOS] Return -EINVAL for unsupport timer in setitimer syscall

stefanberger:
## Description of the changes 

Return -EINVAL for an unsupported timer in setitimer syscall and enable the LTP test case setitimer03.

## How to test this PR? 

Run LTP.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/929)

### DIFF
--- a/libos/src/sys/libos_alarm.c
+++ b/libos/src/sys/libos_alarm.c
@@ -78,7 +78,7 @@ static void signal_itimer(IDTYPE caller, void* arg) {
 long libos_syscall_setitimer(int which, struct __kernel_itimerval* value,
                              struct __kernel_itimerval* ovalue) {
     if (which != ITIMER_REAL)
-        return -ENOSYS;
+        return -EINVAL;
 
     if (!value)
         return -EFAULT;
--- a/libos/test/ltp/ltp.cfg
+++ b/libos/test/ltp/ltp.cfg
@@ -1999,9 +1999,6 @@ skip = yes
 [sethostname03]
 skip = yes
 
-[setitimer03]
-skip = yes
-
 [setns01]
 skip = yes
 


boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/929)* status: 0 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @stefanberger)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/929#-NCke6AqAFs7JsxtW8RC:-NCke6AqAFs7JsxtW8RD:buaamhq) (no related file):*
What's the point of this change? Generally I don't feel like returning `EINVAL` for syscalls we do not implement is a good idea.
Also `setitimer03` from LTP is pretty useless test, so we gain nothing from enabling it (it just checks that invalid timer type returns EINVAL).

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/929)* status: 0 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/929#-NCke6AqAFs7JsxtW8RC:-NCsZuR7DEpYICz8B_qc:b-kyk0cj) (no related file):*

What's the point of this change? Generally I don't feel like returning `EINVAL` for syscalls we do not implement is a good idea.
Also `setitimer03` from LTP is pretty useless test, so we gain nothing from enabling it (it just checks that invalid timer type returns EINVAL).
</blockquote></details>

I agree with @boryspoplawski. It makes more sense to return `ENOSYS` (as we do now) for things that Gramine does not implement. Rather than `EINVAL` (proposed in this PR).

The correct fix would be to implement the other two types of timers (`ITIMER_VIRTUAL` and `ITIMER_PROF`), and return `EINVAL` when `which` is not one of the supported three.

I vote for closing this PR.

___

PR #937, opened 2022-09-28
[tools/pf_crypt] Normalize paths during PF encrypt/decrypt

### LINKED ISSUE #923: warning: encrypted_file_internal_open: pf_open failed: Invalid path
## Description of the problem
I am trying to run a ML model where the inputs to the model are encrypted (kind of like the pytorch example in the docs, but more involved). I encrypted all of the input image files using the wrap key, and mount the input directory as an encrypted directory.

The DCAP remote attestation works ok and my enclave is able to receive the wrap key, however, when the model tries to open the actual files, I get the following error:
`warning: encrypted_file_internal_open: pf_open failed: Invalid path`

Here are the some more log messages:
```
[P1:T1:python3] trace: ---- shim_stat("./datasets/cd-sem-enc/AB/train", 0x1c6980bd
0) = 0x0
[P1:T1:python3] trace: ---- shim_openat(AT_FDCWD, "./datasets/cd-sem-enc/AB/train"
, O_RDONLY|0x90800, 0000) = 0x3
[P1:T1:python3] trace: ---- shim_newfstatat(3, "", 0x1c6980af0, 4096) = 0x0
[P1:T1:python3] warning: encrypted_file_internal_open: pf_open failed: Invalid path
```
the `./datasets/cd-sem-enc/AB/train` directory is the directory holding the encrypted images. I've mounted it like so:

```
# need this line for compatibility reasons
# loader.preload = "file:{{ gramine.libos }}"

loader.entrypoint = "file:{{ gramine.libos }}"
libos.entrypoint = "{{ entrypoint }}"
loader.insecure__use_cmdline_argv = true

loader.env.LD_LIBRARY_PATH = "/lib:{{ arch_libdir }}:/usr/lib:/usr/bin/:/usr/{{ arch_libdir }}"
loader.env.LD_PRELOAD = "libsecret_prov_attest.so"
loader.env.SECRET_PROVISION_CONSTRUCTOR = "1"
loader.env.SECRET_PROVISION_SET_KEY = "default"
loader.env.SECRET_PROVISION_CA_CHAIN_PATH = "ssl/ca.crt"
loader.env.SECRET_PROVISION_SERVERS = "localhost:4433"

loader.argv0_override = "{{ entrypoint }}"

fs.mounts = [
	{ path = "{{ arch_libdir }}", uri = "file:{{ arch_libdir }}" },
	{ path = "/datasets/cd-sem-enc/", uri = "file:datasets/cd-sem-enc/", type = "encrypted"},
	{ path = "/etc", uri = "file:/etc" },
	{ path = "/usr", uri = "file:/usr" },
	{ path = "/tmp", uri = "file:/tmp" },
	{ path = "/usr/lib", uri = "file:/usr/lib" },
]

loader.pal_internal_mem_size = "256M"
sgx.debug = true
sgx.thread_num = 10
sgx.nonpie_binary = true
sgx.enclave_size = "8192M"
sgx.remote_attestation = "dcap"

# make sure you install numpy, torch and torchvision as sudo so they are installed to the distlib not the local user
sgx.trusted_files = [
	"file:{{ gramine.libos }}",
	"file:{{ entrypoint }}",
	"file:{{ arch_libdir }}/",
	"file:/usr{{ arch_libdir }}/",
	"file:{{ python.stdlib }}/",
	"file:{{ python.distlib }}/",
	"file:/usr/local/lib/python3.8/dist-packages/",
	"file:/usr/lib/python3/dist-packages/",
	"file:{{python.get_path('stdlib', vars={'installed_base': '/usr/local'}) }}/",
	"file:ssl/ca.crt",
	"file:/usr/lib/python3.8/",
	"file:{{ env.HOME }}/gramine/CI-Examples/pix2pix/",
	"file:./"
]

sgx.allowed_files = [
	"file:/etc/nsswitch.conf",
	"file:/etc/ethers",
	"file:/etc/host.conf",
	"file:/etc/hosts",
	"file:/etc/group",
	"file:/etc/passwd",
	"file:/etc/gai.conf",
	"file:/etc/resolv.conf",
	"file:/usr/share/dpkg/cputable",
	"file:/usr/share/dpkg/tupletable",
	"file:/etc/apt/apt.conf.d/",
	"file:/usr/lib/ssl/openssl.cnf",
	"file:/etc/default/apport",
	"file:/tmp/", #to cache stuff
	"file:./checkpoints/", #cd_sem_pix2pix_enc/"
	"file:./results/",
]
```

## Steps to reproduce
run `make train-cd` with the following Makefile
```
#!/bin/bash
GRAMINE_LOG_LEVEL ?= all
ARCH_LIBDIR ?= /lib/$(shell $(CC) -dumpmachine)
ENTRYPOINT ?= python3
SGX_SIGNER_KEY ?= /home/trishita/.config/gramine/enclave-key.pem
CFLAGS = -Wall -Wextra -O3
CC = gcc
SGX = 1

# note that entry point needs to be full path
python3.manifest: python3.manifest.template train.py #~/gramine/CI-Examples/io-secure/input.txt
	gramine-manifest \
	-Dentrypoint=/usr/bin/$(ENTRYPOINT) \
	-Darch_libdir=$(ARCH_LIBDIR) \
	-Dlog_level=$(GRAMINE_LOG_LEVEL) \
	$< $@

python3.manifest.sgx: python3.manifest
	@test -s $(SGX_SIGNER_KEY) || \
		{ echo "SGX signer private key was not found, please specify SGX_SIGNER_KEY!"; exit 1; }
	sudo gramine-sgx-sign \
		--key $(SGX_SIGNER_KEY) \
		--manifest $< \
		--output $@
	sudo chown trishita:trishita python3.manifest.sgx python3.sig

python3.sig: python3.manifest.sgx

python3.token: python3.sig
	gramine-sgx-get-token \
		--output $@ --sig $<

ifeq ($(SGX),)
GRAMINE = gramine-direct
else
GRAMINE = gramine-sgx
endif

encrypt:
	gramine-sgx-pf-crypt encrypt -w files/wrap-key -i datasets/cd-sem -o datasets/cd-sem-enc

train-cd: python3.token
	sudo $(GRAMINE) $(ENTRYPOINT) train.py \
	--dataroot ./datasets/cd-sem-enc/AB \ # this is the path to the encrypted dataset
	--name cd_sem_pix2pix_enc \
	--model pix2pix \
	--direction AtoB \
	--gpu_ids -1 \
	--display_id 0 \
	--num_threads 0 \
	--checkpoints_dir ./checkpoints \
	--n_epochs 100 \
	--gan_mode vanilla \
	--beta1 0.5     

test-cd: python3.token
	sudo $(GRAMINE) $(ENTRYPOINT) test.py \
	--dataroot ./datasets/cd-sem/AB \
	--checkpoints_dir ./checkpoints \
	--name cd_sem_pix2pix \
	--model pix2pix \
	--direction AtoB \
	--gpu_ids -1
```

## Expected results
The model should be able to open the encrypted files transparently

## Actual results
`warning: encrypted_file_internal_open: pf_open failed: Invalid path`

## Gramine commit hash
`003efc4c8819de47ff11b5a0af7ba09aee7f5fc1`

Thank you so much for your help!

dimakuv:
## Description of the changes 

Protected Files common logic requires file paths to be normalized, otherwise filename comparison (done to prevent file-swap attacks) fails. E.g., `gramine-sgx-pf-crypt encrypt ... -o ./encrypted-file` would previously save the path verbatim (`./encrypted-file`) which would confuse Gramine that expects a normalized path.

Fixes #923.

## How to test this PR? 

Play with `gramine-sgx-pf-crypt` tool like this:
```
gramine-sgx-pf-crypt encrypt -w key.key -i input.file -o ./somedir/.././encrypted.file

gramine-sgx-pf-crypt decrypt -w key.key -i ./gramine/./../encrypted.file -o ./somedir/../decrypted.file -V
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/937)

### DIFF
--- a/tools/sgx/common/meson.build
+++ b/tools/sgx/common/meson.build
@@ -18,6 +18,7 @@ sgx_util = shared_library('sgx_util',
     ],
     dependencies: [
         cjson_dep,
+        common_utils_dep,
         libcurl_dep,
         protected_files_dep,
         mbedtls_static_dep,
--- a/tools/sgx/common/pf_util.c
+++ b/tools/sgx/common/pf_util.c
@@ -23,6 +23,7 @@
 
 #define USE_STDLIB
 #include "api.h"
+#include "path_utils.h"
 #include "perm.h"
 #include "util.h"
 
@@ -269,9 +270,24 @@ int pf_encrypt_file(const char* input_path, const char* output_path, const pf_ke
     int input = -1;
     int output = -1;
     pf_context_t* pf = NULL;
+    char* norm_output_path = NULL;
+
     void* chunk = malloc(PF_NODE_SIZE);
     if (!chunk) {
         ERROR("Out of memory\n");
+        goto out;
+    }
+
+    size_t output_path_size = strlen(output_path) + 1;
+    norm_output_path = malloc(output_path_size);
+    if (!norm_output_path) {
+        ERROR("Out of memory\n");
+        goto out;
+    }
+
+    bool norm_ret = get_norm_path(output_path, norm_output_path, &output_path_size);
+    if (!norm_ret) {
+        ERROR("Failed to normalize path '%s'\n", output_path);
         goto out;
     }
 
@@ -281,19 +297,19 @@ int pf_encrypt_file(const char* input_path, const char* output_path, const pf_ke
         goto out;
     }
 
-    output = open(output_path, O_RDWR | O_CREAT, PERM_rw_rw_r__);
+    output = open(norm_output_path, O_RDWR | O_CREAT, PERM_rw_rw_r__);
     if (output < 0) {
-        ERROR("Failed to create output file '%s': %s\n", output_path, strerror(errno));
+        ERROR("Failed to create output file '%s': %s\n", norm_output_path, strerror(errno));
         goto out;
     }
 
-    INFO("Encrypting: %s -> %s\n", input_path, output_path);
+    INFO("Encrypting: %s -> %s\n", input_path, norm_output_path);
     INFO("            (Gramine's sgx.protected_files must contain this exact path: \"%s\")\n",
-                      output_path);
+                      norm_output_path);
 
     pf_handle_t handle = (pf_handle_t)&output;
-    pf_status_t pfs = pf_open(handle, output_path, /*size=*/0, PF_FILE_MODE_WRITE, /*create=*/true,
-                              wrap_key, &pf);
+    pf_status_t pfs = pf_open(handle, norm_output_path, /*size=*/0, PF_FILE_MODE_WRITE,
+                              /*create=*/true, wrap_key, &pf);
     if (PF_FAILURE(pfs)) {
         ERROR("Failed to open output PF: %s\n", pf_strerror(pfs));
         goto out;
@@ -341,6 +357,7 @@ out:
     }
 
     free(chunk);
+    free(norm_output_path);
     if (input >= 0)
         close(input);
     if (output >= 0)
@@ -355,6 +372,8 @@ int pf_decrypt_file(const char* input_path, const char* output_path, bool verify
     int input = -1;
     int output = -1;
     pf_context_t* pf = NULL;
+    char* norm_input_path = NULL;
+
     void* chunk = malloc(PF_NODE_SIZE);
     if (!chunk) {
         ERROR("Out of memory\n");
@@ -382,8 +401,22 @@ int pf_decrypt_file(const char* input_path, const char* output_path, bool verify
         goto out;
     }
 
-    const char* path = verify_path ? input_path : NULL;
-    pf_status_t pfs = pf_open((pf_handle_t)&input, path, input_size, PF_FILE_MODE_READ,
+    if (verify_path) {
+        size_t input_path_size = strlen(input_path) + 1;
+        norm_input_path = malloc(input_path_size);
+        if (!norm_input_path) {
+            ERROR("Out of memory\n");
+            goto out;
+        }
+
+        bool norm_ret = get_norm_path(input_path, norm_input_path, &input_path_size);
+        if (!norm_ret) {
+            ERROR("Failed to normalize path '%s'\n", input_path);
+            goto out;
+        }
+    }
+
+    pf_status_t pfs = pf_open((pf_handle_t)&input, norm_input_path, input_size, PF_FILE_MODE_READ,
                               /*create=*/false, wrap_key, &pf);
     if (PF_FAILURE(pfs)) {
         ERROR("Opening protected input file failed: %s\n", pf_strerror(pfs));
@@ -438,6 +471,7 @@ int pf_decrypt_file(const char* input_path, const char* output_path, bool verify
     ret = 0;
 
 out:
+    free(norm_input_path);
     free(chunk);
     if (pf)
         pf_close(pf);
--- a/tools/sgx/pf_crypt/pf_crypt.c
+++ b/tools/sgx/pf_crypt/pf_crypt.c
@@ -23,8 +23,8 @@ struct option g_options[] = {
     { 0, 0, 0, 0 }
 };
 
-static void usage(void) {
-    INFO("\nUsage: pf_crypt mode [options]\n");
+static void usage(const char* argv0) {
+    INFO("\nUsage: %s mode [options]\n", argv0);
     INFO("Available modes:\n");
     INFO("  gen-key                 Generate and save wrap key to file\n");
     INFO("  encrypt                 Encrypt plaintext files\n");
@@ -80,17 +80,17 @@ int main(int argc, char* argv[]) {
                 verify = true;
                 break;
             case 'h':
-                usage();
+                usage(argv[0]);
                 exit(0);
             default:
                 ERROR("Unknown option: %c\n", this_option);
-                usage();
+                usage(argv[0]);
         }
     }
 
     if (optind >= argc) {
         ERROR("Mode not specified\n");
-        usage();
+        usage(argv[0]);
         goto out;
     }
 
@@ -114,7 +114,7 @@ int main(int argc, char* argv[]) {
         case 'e': /* encrypt */
             if (!input_path || !output_path) {
                 ERROR("Input or output path not specified\n");
-                usage();
+                usage(argv[0]);
                 goto out;
             }
             ret = pf_encrypt_files(input_path, output_path, wrap_key_path);
@@ -123,14 +123,14 @@ int main(int argc, char* argv[]) {
         case 'd': /* decrypt */
             if (!input_path || !output_path) {
                 ERROR("Input or output path not specified\n");
-                usage();
+                usage(argv[0]);
                 goto out;
             }
             ret = pf_decrypt_files(input_path, output_path, verify, wrap_key_path);
             break;
 
         default:
-            usage();
+            usage(argv[0]);
             goto out;
     }
 
--- a/tools/sgx/pf_tamper/pf_tamper.c
+++ b/tools/sgx/pf_tamper/pf_tamper.c
@@ -32,8 +32,8 @@ struct option g_options[] = {
     { 0, 0, 0, 0 }
 };
 
-static void usage(void) {
-    INFO("\nUsage: pf_tamper [options]\n");
+static void usage(const char* argv0) {
+    INFO("\nUsage: %s [options]\n", argv0);
     INFO("\nAvailable options:\n");
     INFO("  --help, -h           Display this help\n");
     INFO("  --verbose, -v        Enable verbose output\n");
@@ -415,29 +415,29 @@ int main(int argc, char* argv[]) {
                 set_verbose(true);
                 break;
             case 'h':
-                usage();
+                usage(argv[0]);
                 return 0;
             default:
                 ERROR("Unknown option: %c\n", option);
-                usage();
+                usage(argv[0]);
         }
     }
 
     if (!input_path) {
         ERROR("Input path not specified\n");
-        usage();
+        usage(argv[0]);
         goto out;
     }
 
     if (!g_output_dir) {
         ERROR("Output path not specified\n");
-        usage();
+        usage(argv[0]);
         goto out;
     }
 
     if (!wrap_key_path) {
         ERROR("Wrap key path not specified\n");
-        usage();
+        usage(argv[0]);
         goto out;
     }
 


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/937)* status: 0 of 5 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

___
*[`tools/sgx/common/util.c` line 255 at r1](https://reviewable.io/reviews/gramineproject/gramine/937#-ND2Vmis9mwd1m-PwQqG:-ND2VmitEOWNtJz71vCl:bbmogn4) ([raw file](https://github.com/gramineproject/gramine/blob/77fb399200331498a8a6906db277b000cc0669e8/tools/sgx/common/util.c#L255)):*
> ```C
> }
> 
> /* below functions are copied from common/src/path.c */
> ```

FYI: This copy-paste is ugly, but I couldn't come up with a better alternative.

There is no function like `get_norm_path()` in the standard C library. The closest I found is `man 3 realpath`, but it produces an **absolute path**, which is not what we want.

I didn't want to mark `common_dep` (the `common/` source files) as the dependency for `libsgx_util.so`, because this leads to tremendous name clashes. Recall that these tools do *not* use Gramine's `common/` APIs, instead these tools use normal Glibc functions.

Also, I didn't want to move this `get_norm_path()` code to a separate dependency/library. This would be plain ugly, and just for this corner-case. So I decided to duplicate the logic.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/937)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/937#-ND36sLoDS0uzG9-iaOM:-ND36sLoDS0uzG9-iaON:baxl07) (no related file):*
I'm confused. Is this `-o` parameter a path on the host where the encrypted file should have saved, or in-Gramine path where it must be mounted? From the code looks like the former, but to prevent swap attacks it should be letter. Which makes me believe it's both, that would be a bad design...
Also, shouldn't we use absolute in-Gramine paths to be really swap resistant?

___
___
*[`tools/sgx/common/pf_util.c` line 280 at r1](https://reviewable.io/reviews/gramineproject/gramine/937#-ND36Tes23jO8AnPpQvL:-ND36Tes23jO8AnPpQvM:bhozu2u) ([raw file](https://github.com/gramineproject/gramine/blob/77fb399200331498a8a6906db277b000cc0669e8/tools/sgx/common/pf_util.c#L280)):*
> ```C
>     }
> 
>     norm_output_path = strdup(output_path);
> ```

Since you get the length of the string below anyway, why not use `malloc`? This looks weird.
___
*[`tools/sgx/common/pf_util.c` line 404 at r1](https://reviewable.io/reviews/gramineproject/gramine/937#-ND37VzF0_iW52L4NQ1b:-ND37VzF0_iW52L4NQ1c:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/77fb399200331498a8a6906db277b000cc0669e8/tools/sgx/common/pf_util.c#L404)):*
> ```C
> 
>     if (verify_path) {
>         norm_input_path = strdup(input_path);
> ```

ditto
___
*[`tools/sgx/common/util.c` line 255 at r1](https://reviewable.io/reviews/gramineproject/gramine/937#-ND2Vmis9mwd1m-PwQqG:-ND37wEE-h0jlVkiuayZ:b-9ulk5) ([raw file](https://github.com/gramineproject/gramine/blob/77fb399200331498a8a6906db277b000cc0669e8/tools/sgx/common/util.c#L255)):*

FYI: This copy-paste is ugly, but I couldn't come up with a better alternative.

There is no function like `get_norm_path()` in the standard C library. The closest I found is `man 3 realpath`, but it produces an **absolute path**, which is not what we want.

I didn't want to mark `common_dep` (the `common/` source files) as the dependency for `libsgx_util.so`, because this leads to tremendous name clashes. Recall that these tools do *not* use Gramine's `common/` APIs, instead these tools use normal Glibc functions.

Also, I didn't want to move this `get_norm_path()` code to a separate dependency/library. This would be plain ugly, and just for this corner-case. So I decided to duplicate the logic.
</blockquote></details>

In theory you could `realpath` and `lstrip` `getcwd`, but that would resolve all symlinks. Also see the top level comments, this 
___
*[`tools/sgx/common/util.c` line 255 at r1](https://reviewable.io/reviews/gramineproject/gramine/937#-ND3A3Vu3T3JA3nkyUe4:-ND3A3Vv2cVkHfkHxVSU:b14x4ca) ([raw file](https://github.com/gramineproject/gramine/blob/77fb399200331498a8a6906db277b000cc0669e8/tools/sgx/common/util.c#L255)):*
> ```C
> }
> 
> /* below functions are copied from common/src/path.c */
> ```

Why did you remove most of the comments from these functions? Not saying we need them, just saw that in the diff
___
*[`tools/sgx/pf_crypt/pf_crypt.c` line 27 at r1](https://reviewable.io/reviews/gramineproject/gramine/937#-ND39OGU0dkhPlnPepF0:-ND39OGU0dkhPlnPepF1:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/77fb399200331498a8a6906db277b000cc0669e8/tools/sgx/pf_crypt/pf_crypt.c#L27)):*
> ```C
> 
> static void usage(void) {
>     INFO("\nUsage: gramine-sgx-pf-crypt mode [options]\n");
> ```





_Suggestion:_
```C
static void usage(const char* argv0) {
    INFO("\nUsage: %s mode [options]\n", argv0);
```
___
*[`tools/sgx/pf_tamper/pf_tamper.c` line 36 at r1](https://reviewable.io/reviews/gramineproject/gramine/937#-ND39V4T-p_KLSkRCjBX:-ND39V4T-p_KLSkRCjBY:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/77fb399200331498a8a6906db277b000cc0669e8/tools/sgx/pf_tamper/pf_tamper.c#L36)):*
> ```C
> 
> static void usage(void) {
>     INFO("\nUsage: gramine-sgx-pf-tamper [options]\n");
> ```

ditto

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/937)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/937#-ND36sLoDS0uzG9-iaOM:-ND3S7OX3RL44HqJ8cal:b-lcgzoa) (no related file):*
> Why do you say that it should be latter? I mean, yes, the design sounds complicated (with the path on the host, i.e. URI). But the user controls the manifest file which contains the mapping between in-Gramine paths and on-the-host paths. So how can a swap attack happen?

The host is free to change the host path as it wishes. So in reality we just threat that path as an unique identifier - the path requested by Gramine enclave trusted code must match the string that's inside encrypted file. Relative paths are not unique, assuming the starting point might differ. I'm fuzzy on Gramine PF implementation details, so not sure what exactly could go wrong (if anything).
I guess all paths are considered relative to the manifest location, i.e. always have the same starting point, so they are unique after all.

> I don't know how this can be even achieved. The user must know in advance where in Gramine the path will be put, and this in-Gramine path may have absolutely no correlation to the host path. And when user does gramine-sgx-pf-crypt, the user only knows about and operates on host paths.

This would probably be an additional argument to the tool.

___
___
*[`tools/sgx/common/util.c` line 255 at r1](https://reviewable.io/reviews/gramineproject/gramine/937#-ND2Vmis9mwd1m-PwQqG:-ND3U8Xv2dds11Lj79HT:bft9njn) ([raw file](https://github.com/gramineproject/gramine/blob/77fb399200331498a8a6906db277b000cc0669e8/tools/sgx/common/util.c#L255)):*

Did you lose the last part of the sentence?

I really don't like the idea with resolving symlinks.
</blockquote></details>

Yeah, something got cut in the middle.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/937)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/937#-ND36sLoDS0uzG9-iaOM:-ND5oV9s1EiAdK4veUpl:b-k09bde) (no related file):*

> Why do you say that it should be latter? I mean, yes, the design sounds complicated (with the path on the host, i.e. URI). But the user controls the manifest file which contains the mapping between in-Gramine paths and on-the-host paths. So how can a swap attack happen?

The host is free to change the host path as it wishes. So in reality we just threat that path as an unique identifier - the path requested by Gramine enclave trusted code must match the string that's inside encrypted file. Relative paths are not unique, assuming the starting point might differ. I'm fuzzy on Gramine PF implementation details, so not sure what exactly could go wrong (if anything).
I guess all paths are considered relative to the manifest location, i.e. always have the same starting point, so they are unique after all.

> I don't know how this can be even achieved. The user must know in advance where in Gramine the path will be put, and this in-Gramine path may have absolutely no correlation to the host path. And when user does gramine-sgx-pf-crypt, the user only knows about and operates on host paths.

This would probably be an additional argument to the tool.
</blockquote></details>

Hmm, this seems to be a bit weird legacy design and it's causing troubles now. I think it was written this way because in the old times the mounts in Gramine were mess and that was the only viable way. But ideally, this should be decoupled with the path on the host and be tied to the manifest mount via different means (in-gramine path? some id? dunno). Any thoughts?

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/937)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/937#-ND36sLoDS0uzG9-iaOM:-ND715n67FxNAmlsWDfG:b-f8hzt6) (no related file):*
> via different means (in-gramine path? some id? dunno)

I don't see how we can use `some id`. E.g., when Gramine opens an Encrypted File for read, Gramine doesn't have any additional metadata on the file other than two paths: the in-Gramine one and the on-host one (what we also call URI). So Gramine must use only one of these pieces of metadata. I don't see how we can rely on some other `id`.

The `in-gramine path` is a more reasonable candidate. Maybe even **an absolute in-gramine path**. As Borys suggests, we will need an additional argument to our `gramine-sgx-pf-crypt` tool.

I think the `absolute in-gramine path` is the cleanest approach, and it will make sense for seasoned Gramine users. (On the other hand, this is again a breaking change, and it will be *very unintuitive* for many novice users.)

___

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/937)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/937#-ND36sLoDS0uzG9-iaOM:-NDKvef_2L2lHEZalGiQ:b-odeuof) (no related file):*
> I don't see how we can use some `id`.

We could create it and put into manifest. But I don't think there's a way of doing so without making it complex for the users.

> The `in-gramine path` is a more reasonable candidate. Maybe even an **absolute in-gramine path**. As Borys suggests, we will need an additional argument to our `gramine-sgx-pf-crypt` tool.

Yeah, I think we should go this way. @boryspoplawski: What do you think?

> it will be *very unintuitive* for many novice users

Why? Assuming we describe this argument clearly. We can also default it to the host path if not given.

___

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/937)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/937#-ND36sLoDS0uzG9-iaOM:-NDL6sFdDGo31n3o7JLP:b-ra2vwk) (no related file):*
> Yeah, I think we should go this way. @boryspoplawski: What do you think?

Well, I suggested this in the first place, so +1 from my side :)

___

mkow (changes requested):
> Reviewed 3 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/937)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/937#-ND36sLoDS0uzG9-iaOM:-NDNnyriBtk98YAuxvWL:b-hwihw7) (no related file):*

> Yeah, I think we should go this way. @boryspoplawski: What do you think?

Well, I suggested this in the first place, so +1 from my side :)
</blockquote></details>

https://github.com/gramineproject/gramine/issues/949

___
___
*[`-- commits` line 8 at r2](https://reviewable.io/reviews/gramineproject/gramine/937#-NDO0jSbCmZp7GTfrV2F):*
I'm not sure if I understand this description and PR correctly. What you seem to suggest in this sentence is that paths like `./encrypted-file` are not supported correctly without this PR, and this PR adds normalization to paths before saving them in PF headers. But `./encrypted-file` is already a normalized path? It's not absolute, but it _is_ normalized. Or maybe normalized version is just `encrypted-file`? Not sure what's the exact definition of normalization.
___
*[`tools/sgx/common/util.c` line 255 at r1](https://reviewable.io/reviews/gramineproject/gramine/937#-ND2Vmis9mwd1m-PwQqG:-NDO4SCW2l6ULo7VY8nU:b3mr0k3) ([raw file](https://github.com/gramineproject/gramine/blob/77fb399200331498a8a6906db277b000cc0669e8/tools/sgx/common/util.c#L255)):*

Yeah, something got cut in the middle.
</blockquote></details>

Ouch, I really don't like this, it's a non-trivial and quite a long function. Maybe we could split `common_dep` into non-libc-colliding part and a part which supplies our own mini-libc?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/937)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/937#-ND36sLoDS0uzG9-iaOM:-NDX2R3MCZsUAMTZm6em:bwnlkov) (no related file):*

https://github.com/gramineproject/gramine/issues/949
</blockquote></details>

Thanks for creating the issue. I assume we can still continue with this PR, right?

___
___
*[`-- commits` line 8 at r2](https://reviewable.io/reviews/gramineproject/gramine/937#-NDO0jSbCmZp7GTfrV2F):*

I'm not sure if I understand this description and PR correctly. What you seem to suggest in this sentence is that paths like `./encrypted-file` are not supported correctly without this PR, and this PR adds normalization to paths before saving them in PF headers. But `./encrypted-file` is already a normalized path? It's not absolute, but it _is_ normalized. Or maybe normalized version is just `encrypted-file`? Not sure what's the exact definition of normalization.
</blockquote></details>

In this PR, I don't introduce any new definition of "normalization". "Normalization" is whatever the arleady-existing `get_norm_path()` function is doing. In particular, this function removes `./` symbols, yes.

Am I supposed to do anything, like write comments about what we mean with "path normalization"? If yes, then at which place in the code?
___
*[`tools/sgx/common/util.c` line 255 at r1](https://reviewable.io/reviews/gramineproject/gramine/937#-ND2Vmis9mwd1m-PwQqG:-NDX32O85sRi5n6fSgNK:betyihf) ([raw file](https://github.com/gramineproject/gramine/blob/77fb399200331498a8a6906db277b000cc0669e8/tools/sgx/common/util.c#L255)):*

Ouch, I really don't like this, it's a non-trivial and quite a long function. Maybe we could split `common_dep` into non-libc-colliding part and a part which supplies our own mini-libc?
</blockquote></details>

Yes, we could do this. Do you want it to be a pre-requisite to this PR, or do you want me to create a GitHub issue about this and put a TODO comment here?

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/937)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/937#-ND36sLoDS0uzG9-iaOM:-NDYPKN9BfX5BusNzaT-:bb77u1k) (no related file):*

Thanks for creating the issue. I assume we can still continue with this PR, right?
</blockquote></details>

Yeah

___
___
*[`-- commits` line 8 at r2](https://reviewable.io/reviews/gramineproject/gramine/937#-NDO0jSbCmZp7GTfrV2F):*

In this PR, I don't introduce any new definition of "normalization". "Normalization" is whatever the arleady-existing `get_norm_path()` function is doing. In particular, this function removes `./` symbols, yes.

Am I supposed to do anything, like write comments about what we mean with "path normalization"? If yes, then at which place in the code?
</blockquote></details>

I guess path normalization is just making the path as short as possible, while it still points to the same file (and is anchored at the same starting point). Not sure if there is any official version of the notion

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/937)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/937#-ND36sLoDS0uzG9-iaOM:-ND_AUasCosCRfT2uFTq:bb7867z) (no related file):*

Yeah
</blockquote></details>

Yup.

___
___
*[`-- commits` line 8 at r2](https://reviewable.io/reviews/gramineproject/gramine/937#-NDO0jSbCmZp7GTfrV2F):*

I guess path normalization is just making the path as short as possible, while it still points to the same file (and is anchored at the same starting point). Not sure if there is any official version of the notion
</blockquote></details>

Maybe just change the example in the commit message to `././//encrypted-file` or something similar, to make it more obvious what you mean? I was confused by the current description.
___
*[`tools/sgx/common/util.c` line 255 at r1](https://reviewable.io/reviews/gramineproject/gramine/937#-ND2Vmis9mwd1m-PwQqG:-ND_Aogv0EaMb9Bhw3Ua:b-xkhhze) ([raw file](https://github.com/gramineproject/gramine/blob/77fb399200331498a8a6906db277b000cc0669e8/tools/sgx/common/util.c#L255)):*

Yes, we could do this. Do you want it to be a pre-requisite to this PR, or do you want me to create a GitHub issue about this and put a TODO comment here?
</blockquote></details>

Please do it here (or another PR merged before this one), so that you don't need to duplicate code in this PR. That is, assuming we're not in hurry to fix this problem.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/937)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

___
*[`tools/sgx/common/util.c` line 255 at r1](https://reviewable.io/reviews/gramineproject/gramine/937#-ND2Vmis9mwd1m-PwQqG:-NDb4K383iqh9WIYGZKO:bn2p6k1) ([raw file](https://github.com/gramineproject/gramine/blob/77fb399200331498a8a6906db277b000cc0669e8/tools/sgx/common/util.c#L255)):*

Please do it here (or another PR merged before this one), so that you don't need to duplicate code in this PR. That is, assuming we're not in hurry to fix this problem.
</blockquote></details>

We can do it as another PR before merging this one. Let me block this PR then.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/937)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @kailun-qin)

___
*[`tools/sgx/common/util.c` line 255 at r1](https://reviewable.io/reviews/gramineproject/gramine/937#-ND2Vmis9mwd1m-PwQqG:-NDlG7ZY1dQFmLC86zLA:b-kt1ht9) ([raw file](https://github.com/gramineproject/gramine/blob/77fb399200331498a8a6906db277b000cc0669e8/tools/sgx/common/util.c#L255)):*

We can do it as another PR before merging this one. Let me block this PR then.
</blockquote></details>

I had a brief discussion with @kailun-qin. He or one of his teammates could work on this (separation of `common/` code into two dependencies: one with glibc functions, another with non-glibc functions).

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/937)* status: 2 of 6 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski, @kailun-qin, and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/937#-NHPHpC5EcEfJ6r9k9uB:-NHPHpC5EcEfJ6r9k9uC:b-tmz9fq) (no related file):*
I rebased to the latest master. Ready for review (and much simpler now).

___
___
*[`-- commits` line 8 at r2](https://reviewable.io/reviews/gramineproject/gramine/937#-NDO0jSbCmZp7GTfrV2F):*

Maybe just change the example in the commit message to `././//encrypted-file` or something similar, to make it more obvious what you mean? I was confused by the current description.
</blockquote></details>

Done.
___
*[`tools/sgx/common/util.c` line 255 at r1](https://reviewable.io/reviews/gramineproject/gramine/937#-ND2Vmis9mwd1m-PwQqG:-NHPF1opCOunWRhBgv2W:buybcrz) ([raw file](https://github.com/gramineproject/gramine/blob/77fb399200331498a8a6906db277b000cc0669e8/tools/sgx/common/util.c#L255)):*

I had a brief discussion with @kailun-qin. He or one of his teammates could work on this (separation of `common/` code into two dependencies: one with glibc functions, another with non-glibc functions).
</blockquote></details>

Done. The dependent PR was merged (https://github.com/gramineproject/gramine/pull/1002/files). I rebased on top of current master, to incorporate those changes.

boryspoplawski (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/937)* status: all files reviewed, 2 unresolved discussions (waiting on @kailun-qin and @mkow)

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/937)* status: :shipit: complete! all files reviewed, all discussions resolved

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/937)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #946, opened 2022-09-29
[LibOS,PAL] Remove `struct pal_iovec`

boryspoplawski:
## Description of the changes 
This struct was essentially `struct iovec` from Linux and it's usage required some otherwise unneeded `malloc()` + `free()`, which became a performance bottleneck in some cases.
This commit also removes erroneous include of `sys/socket.h`, which also got in the way of these changes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/946)

### DIFF
--- a/Documentation/Doxyfile-pal
+++ b/Documentation/Doxyfile-pal
@@ -1,7 +1,7 @@
 @INCLUDE = Doxyfile
 PROJECT_NAME = "PAL"
 OUTPUT_DIRECTORY = _build/doxygen-pal
-INPUT = ../pal/include/arch/x86_64 ../pal/include/pal ../pal/include
+INPUT = ../pal/include/arch/x86_64 ../pal/include/pal ../pal/include ../common/include/iovec.h
 MACRO_EXPANSION = YES
 EXPAND_ONLY_PREDEF = YES
 PREDEFINED = noreturn= __x86_64__=1
--- a/Documentation/pal/host-abi.rst
+++ b/Documentation/pal/host-abi.rst
@@ -241,7 +241,7 @@ Socket handling
 .. doxygenstruct:: pal_socket_addr
    :project: pal
 
-.. doxygenstruct:: pal_iovec
+.. doxygenstruct:: iovec
    :project: pal
 
 .. doxygenfunction:: PalSocketCreate
--- a/common/include/iovec.h
+++ b/common/include/iovec.h
@@ -1,0 +1,11 @@
+/* SPDX-License-Identifier: LGPL-3.0-or-later */
+/* Copyright (C) 2022 Intel Corporation
+ *                    Borys Popławski <borysp@invisiblethingslab.com>
+ */
+
+#pragma once
+
+struct iovec {
+    void* iov_base;
+    size_t iov_len;
+};
--- a/common/include/linux_socket.h
+++ b/common/include/linux_socket.h
@@ -11,6 +11,8 @@
 #include <linux/un.h>
 #include <stddef.h>
 
+#include "iovec.h"
+
 #define SOCKADDR_MAX_SIZE 128
 
 struct sockaddr_storage {
@@ -19,11 +21,6 @@ struct sockaddr_storage {
         char _size[SOCKADDR_MAX_SIZE];
         void* _align;
     };
-};
-
-struct iovec {
-    void* iov_base;
-    size_t iov_len;
 };
 
 struct msghdr {
@@ -40,6 +37,24 @@ struct mmsghdr {
     struct msghdr msg_hdr;
     unsigned int msg_len;
 };
+
+struct cmsghdr {
+    size_t cmsg_len;
+    int cmsg_level;
+    int cmsg_type;
+    unsigned char __cmsg_data[];
+};
+
+#define CMSG_DATA(cmsg) ((cmsg)->__cmsg_data)
+#define CMSG_FIRSTHDR(mhdr)                                   \
+    ((size_t)(mhdr)->msg_controllen >= sizeof(struct cmsghdr) \
+         ? (struct cmsghdr*)(mhdr)->msg_control               \
+         : (struct cmsghdr*)0)
+#define CMSG_ALIGN(len) ALIGN_UP(len, sizeof(size_t))
+#define CMSG_SPACE(len) (CMSG_ALIGN(sizeof(struct cmsghdr)) + CMSG_ALIGN(len))
+#define CMSG_LEN(len)   (CMSG_ALIGN(sizeof(struct cmsghdr)) + (len))
+
+#define SCM_RIGHTS 1
 
 #define AF_UNSPEC 0
 #define AF_UNIX 1
--- a/libos/src/net/ip.c
+++ b/libos/src/net/ip.c
@@ -652,19 +652,9 @@ static int send(struct libos_handle* handle, struct iovec* iov, size_t iov_len, 
         linux_to_pal_sockaddr(addr, &pal_ip_addr);
     }
 
-    struct pal_iovec* pal_iov = malloc(iov_len * sizeof(*pal_iov));
-    if (!pal_iov) {
-        return -ENOMEM;
-    }
-    for (size_t i = 0; i < iov_len; i++) {
-        pal_iov[i].iov_base = iov[i].iov_base;
-        pal_iov[i].iov_len = iov[i].iov_len;
-    }
-
-    int ret = PalSocketSend(sock->pal_handle, pal_iov, iov_len, out_size,
-                            addr ? &pal_ip_addr : NULL, force_nonblocking);
+    int ret = PalSocketSend(sock->pal_handle, iov, iov_len, out_size, addr ? &pal_ip_addr : NULL,
+                            force_nonblocking);
     ret = (ret == -PAL_ERROR_TOOLONG) ? -EMSGSIZE : pal_to_unix_errno(ret);
-    free(pal_iov);
     return ret;
 }
 
@@ -684,19 +674,9 @@ static int recv(struct libos_handle* handle, struct iovec* iov, size_t iov_len,
             __builtin_unreachable();
     }
 
-    struct pal_iovec* pal_iov = malloc(iov_len * sizeof(*pal_iov));
-    if (!pal_iov) {
-        return -ENOMEM;
-    }
-    for (size_t i = 0; i < iov_len; i++) {
-        pal_iov[i].iov_base = iov[i].iov_base;
-        pal_iov[i].iov_len = iov[i].iov_len;
-    }
-
     struct pal_socket_addr pal_ip_addr;
-    int ret = PalSocketRecv(handle->info.sock.pal_handle, pal_iov, iov_len, out_total_size,
+    int ret = PalSocketRecv(handle->info.sock.pal_handle, iov, iov_len, out_total_size,
                             addr ? &pal_ip_addr : NULL, force_nonblocking);
-    free(pal_iov);
     if (ret < 0) {
         return pal_to_unix_errno(ret);
     }
--- a/pal/include/pal/pal.h
+++ b/pal/include/pal/pal.h
@@ -14,6 +14,8 @@
 #include <stdint.h>
 #include <stdnoreturn.h>
 
+#include "iovec.h"
+
 // TODO: fix this (but see pal/include/arch/x86_64/pal_arch.h)
 #define INSIDE_PAL_H
 
@@ -509,13 +511,6 @@ struct pal_socket_addr {
     };
 };
 
-/* This could be just `struct iovec` from Linux, but we don't want to set a precedent of using Linux
- * types in PAL API. */
-struct pal_iovec {
-    void* iov_base;
-    size_t iov_len;
-};
-
 /*!
  * \brief Create a socket handle.
  *
@@ -602,8 +597,12 @@ int PalSocketConnect(PAL_HANDLE handle, struct pal_socket_addr* addr,
  * \returns 0 on success, negative error code on failure.
  *
  * Data is sent atomically, i.e. data from two `PalSocketSend` calls will not be interleaved.
+ *
+ * We use Linux `struct iovec` as argument here, because the alternative is to use a custom
+ * structure, but it would contain exactly the same fields, which would achieve nothing, but could
+ * worsen performance in certain cases.
  */
-int PalSocketSend(PAL_HANDLE handle, struct pal_iovec* iov, size_t iov_len, size_t* out_size,
+int PalSocketSend(PAL_HANDLE handle, struct iovec* iov, size_t iov_len, size_t* out_size,
                   struct pal_socket_addr* addr, bool force_nonblocking);
 
 /*!
@@ -622,8 +621,12 @@ int PalSocketSend(PAL_HANDLE handle, struct pal_iovec* iov, size_t iov_len, size
  * \returns 0 on success, negative error code on failure.
  *
  * Data is received atomically, i.e. data from two `PalSocketRecv` calls will not be interleaved.
+ *
+ * We use Linux `struct iovec` as argument here, because the alternative is to use a custom
+ * structure, but it would contain exactly the same fields, which would achieve nothing, but could
+ * worsen performance in certain cases.
  */
-int PalSocketRecv(PAL_HANDLE handle, struct pal_iovec* iov, size_t iov_len, size_t* out_total_size,
+int PalSocketRecv(PAL_HANDLE handle, struct iovec* iov, size_t iov_len, size_t* out_total_size,
                   struct pal_socket_addr* addr, bool force_nonblocking);
 
 /*
--- a/pal/include/pal_internal.h
+++ b/pal/include/pal_internal.h
@@ -118,9 +118,9 @@ struct socket_ops {
                   struct pal_socket_addr* out_client_addr, struct pal_socket_addr* out_local_addr);
     int (*connect)(PAL_HANDLE handle, struct pal_socket_addr* addr,
                    struct pal_socket_addr* out_local_addr);
-    int (*send)(PAL_HANDLE handle, struct pal_iovec* iov, size_t iov_len, size_t* out_size,
+    int (*send)(PAL_HANDLE handle, struct iovec* iov, size_t iov_len, size_t* out_size,
                 struct pal_socket_addr* addr, bool force_nonblocking);
-    int (*recv)(PAL_HANDLE handle, struct pal_iovec* iov, size_t iov_len, size_t* out_size,
+    int (*recv)(PAL_HANDLE handle, struct iovec* iov, size_t iov_len, size_t* out_size,
                 struct pal_socket_addr* addr, bool force_nonblocking);
 };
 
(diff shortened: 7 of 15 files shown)


dimakuv (changes requested):
> Reviewed 16 of 16 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/946)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/946#-NDCQ_N1Arb0JaQ4uHOX:-NDCQ_N1Arb0JaQ4uHOY:b-d1t48f) (no related file):*
Docs failed.

___
___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/946#-NDCPU-V4oRNmL-6joqw):*
Maybe "Replace `struct pal_iovec` with `struct iovec`" ?

Otherwise it's hard to notice the difference -- that you introduce `struct iovec` (same as UNIX) instead of `struct pal_iovec` (Gramine specific).
___
*[`common/include/iovec.h` line 8 at r1](https://reviewable.io/reviews/gramineproject/gramine/946#-NDCPptQ4yxK-AU1Y33B:-NDCPptQ4yxK-AU1Y33C:b7a623c) ([raw file](https://github.com/gramineproject/gramine/blob/96cf960dc2e63488037f2f9de724ed78b6a0dae4/common/include/iovec.h#L8)):*
> ```objc
> #pragma once
> 
> struct iovec {
> ```

Do we want to add somewhere that:
1. UNIX host doesn't need to re-shape objects of this type and can just directly forward them to host syscalls, but
2. non-UNIX hosts (e.g. Windows) must re-shape objects of this type, which may hamper performance

Mainly, I want to have a comment somewhere that we do this ABI-compatible declaration of `struct iovec` for performance reasons on UNIX systems. We have this in the commit message, but I also want some clear messaging in the code itself.
___
*[`pal/include/pal/pal.h` line 511 at r1](https://reviewable.io/reviews/gramineproject/gramine/946#-NDCQiIc2XTrg4sv8EYp:-NDCQiIc2XTrg4sv8EYq:b-66k93n) ([raw file](https://github.com/gramineproject/gramine/blob/bf9d2758b6dbe3b973d707df157ec54caec1bc05/pal/include/pal/pal.h#L511)):*
> ```objc
> 
> /* This could be just `struct iovec` from Linux, but we don't want to set a precedent of using Linux
>  * types in PAL API. */
> ```

Performance vs code purity, 1:0
___
*[`pal/src/host/linux-sgx/pal_streams.c` line 19 at r1](https://reviewable.io/reviews/gramineproject/gramine/946#-NDCRSHX1sN4arU4g5rD:-NDCRSHX1sN4arU4g5rE:b-c9mymt) ([raw file](https://github.com/gramineproject/gramine/blob/96cf960dc2e63488037f2f9de724ed78b6a0dae4/pal/src/host/linux-sgx/pal_streams.c#L19)):*
> ```C
> #include "crypto.h"
> #include "enclave_pages.h"
> #include "linux_socket.h"
> ```

Why did you have to add it here? If it's needed, how did it work before?

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/946)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/946#-NDCQ_N1Arb0JaQ4uHOX:-NDEnaKh-RnVzfZWNN6Z:b-896fix) (no related file):*

Docs failed.
</blockquote></details>

Done.

___
___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/946#-NDCPU-V4oRNmL-6joqw):*

Maybe "Replace `struct pal_iovec` with `struct iovec`" ?

Otherwise it's hard to notice the difference -- that you introduce `struct iovec` (same as UNIX) instead of `struct pal_iovec` (Gramine specific).
</blockquote></details>

Well, `sutrct iovec` was already there, I did not introduced it, just used in PAL API. Maybe:
```[PAL] Replace `struct pal_iovec` with `struct iovec"` in PAL API```?
___
*[`common/include/iovec.h` line 8 at r1](https://reviewable.io/reviews/gramineproject/gramine/946#-NDCPptQ4yxK-AU1Y33B:-NDEeUCy6apRzW4o_R00:b-s12tjr) ([raw file](https://github.com/gramineproject/gramine/blob/96cf960dc2e63488037f2f9de724ed78b6a0dae4/common/include/iovec.h#L8)):*

Do we want to add somewhere that:
1. UNIX host doesn't need to re-shape objects of this type and can just directly forward them to host syscalls, but
2. non-UNIX hosts (e.g. Windows) must re-shape objects of this type, which may hamper performance

Mainly, I want to have a comment somewhere that we do this ABI-compatible declaration of `struct iovec` for performance reasons on UNIX systems. We have this in the commit message, but I also want some clear messaging in the code itself.
</blockquote></details>

Not sure I understand exactly what you want here. There are few (kind of unrelated) things:
- Here we have this struct because it's a global definition - it must be exactly the same as on Linux.
- We use it in LibOS, because that's the type used by Linux syscalls.
- We also use it in PAL API - yes, performance is one (and for sure key) reason, but note that we had the same struct, albeit named differently, before, when performance was not optimal (because of unnecessary copying). So it's not like we have this purely for UNIX host performance -  such struct would be used in PAL API anyway.
- Originaly `struct iovec` was not used, so that when somebody implements a new PAL API, they could check this (i.e. an existing one), see a Linux type and think that it's fine to just use Linux types everywhere (which is rarely the case).
- Also note that `ocall_send` does the copying anyway (what you call "re-shaping"), so it's not even that all Linux hosts can use it directly.

So the only place I can think of where we could put something is `pal.h` near all functions which use this struct, but not sure what exactly to put there.
___
*[`pal/include/pal/pal.h` line 511 at r1](https://reviewable.io/reviews/gramineproject/gramine/946#-NDCQiIc2XTrg4sv8EYp:-NDEiHzN-gU80qeQTerm:b7uh4ct) ([raw file](https://github.com/gramineproject/gramine/blob/bf9d2758b6dbe3b973d707df157ec54caec1bc05/pal/include/pal/pal.h#L511)):*

Performance vs code purity, 1:0
</blockquote></details>

It's not code purity, I just didn't want to set a precedent for future cases.
___
*[`pal/src/host/linux-sgx/pal_streams.c` line 19 at r1](https://reviewable.io/reviews/gramineproject/gramine/946#-NDCRSHX1sN4arU4g5rD:-NDEitcjEe3E6rHJetAY:bk0d1su) ([raw file](https://github.com/gramineproject/gramine/blob/96cf960dc2e63488037f2f9de724ed78b6a0dae4/pal/src/host/linux-sgx/pal_streams.c#L19)):*

Why did you have to add it here? If it's needed, how did it work before?
</blockquote></details>

It uses `CMSG_*` stuff, which we had in a PAL-SGX specific header `pal_linux_types.h`.

**Review by dimakuv — DISMISSED** [review]
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/946)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/946#-NDCPU-V4oRNmL-6joqw):*

Well, `sutrct iovec` was already there, I did not introduced it, just used in PAL API. Maybe:
```[PAL] Replace `struct pal_iovec` with `struct iovec"` in PAL API```?
</blockquote></details>

Yes, the proposed title is also good. (Bar the accidental `"` symbol.)
___
*[`common/include/iovec.h` line 8 at r1](https://reviewable.io/reviews/gramineproject/gramine/946#-NDCPptQ4yxK-AU1Y33B:-NDWuy2W79R-7dcbbo0b:b-rcd2ih) ([raw file](https://github.com/gramineproject/gramine/blob/96cf960dc2e63488037f2f9de724ed78b6a0dae4/common/include/iovec.h#L8)):*

Not sure I understand exactly what you want here. There are few (kind of unrelated) things:
- Here we have this struct because it's a global definition - it must be exactly the same as on Linux.
- We use it in LibOS, because that's the type used by Linux syscalls.
- We also use it in PAL API - yes, performance is one (and for sure key) reason, but note that we had the same struct, albeit named differently, before, when performance was not optimal (because of unnecessary copying). So it's not like we have this purely for UNIX host performance -  such struct would be used in PAL API anyway.
- Originaly `struct iovec` was not used, so that when somebody implements a new PAL API, they could check this (i.e. an existing one), see a Linux type and think that it's fine to just use Linux types everywhere (which is rarely the case).
- Also note that `ocall_send` does the copying anyway (what you call "re-shaping"), so it's not even that all Linux hosts can use it directly.

So the only place I can think of where we could put something is `pal.h` near all functions which use this struct, but not sure what exactly to put there.
</blockquote></details>

Yes, I see your point. I don't think putting comments about it in `pal.h` is worth it.

Ok, I guess I'm resolving because indeed there is no good place to put the comment. Anyway, it will be obvious that for non-UNIX hosts, this struct will need to be re-shaped and thus will loose some performance (on allocation/deallocation of the object + copy). And for historical context, we have the commit message.

**Review by mkow — CHANGES_REQUESTED** [review]
> Reviewed 15 of 16 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/946)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`common/include/linux_socket.h` line 53 at r2](https://reviewable.io/reviews/gramineproject/gramine/946#-NEM8x4q0FFbt40vThXn:-NEM8x4rA-RS0cI7TMg3:b66tyj) ([raw file](https://github.com/gramineproject/gramine/blob/da13eee3b6dc57ab6a777a67972b9463b36ef3cc/common/include/linux_socket.h#L53)):*
> ```objc
>          ? (struct cmsghdr*)(mhdr)->msg_control               \
>          : (struct cmsghdr*)0)
> #define CMSG_ALIGN(len) ALIGN_UP((len), sizeof(size_t))
> ```

Why did you add additional parens here and below?



_Code quote:_
```objc
(len)
```

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/946)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`common/include/linux_socket.h` line 53 at r2](https://reviewable.io/reviews/gramineproject/gramine/946#-NEM8x4q0FFbt40vThXn:-NEMozTDDD90ClxAi5Hl:bi8p4o0) ([raw file](https://github.com/gramineproject/gramine/blob/da13eee3b6dc57ab6a777a67972b9463b36ef3cc/common/include/linux_socket.h#L53)):*

Why did you add additional parens here and below?
</blockquote></details>

I thought we had discussion about this once and the conclusion was that we should use parens around macro args always, even if the chances  of miscompilation are minimal (because highest op priority)

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/946)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`common/include/linux_socket.h` line 53 at r2](https://reviewable.io/reviews/gramineproject/gramine/946#-NEM8x4q0FFbt40vThXn:-NEWEmAB95kdx0irmaaS:bcyvitl) ([raw file](https://github.com/gramineproject/gramine/blob/da13eee3b6dc57ab6a777a67972b9463b36ef3cc/common/include/linux_socket.h#L53)):*

I thought we had discussion about this once and the conclusion was that we should use parens around macro args always, even if the chances  of miscompilation are minimal (because highest op priority)
</blockquote></details>

Hmm, I thought the conclusion then was that in this case we shouldn't. By quickly grepping and looking at random results I couldn't find any place where we parenthesize such a construction (and it's a very popular case, I found maaany examples, all without parens).

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/946)* status: 16 of 17 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`common/include/linux_socket.h` line 53 at r2](https://reviewable.io/reviews/gramineproject/gramine/946#-NEM8x4q0FFbt40vThXn:-NEWPTTiABr6cPbRY68V:b4y5gs9) ([raw file](https://github.com/gramineproject/gramine/blob/da13eee3b6dc57ab6a777a67972b9463b36ef3cc/common/include/linux_socket.h#L53)):*
Idk either, I just thought you wanted them, so I'm fine with removing.

> and below?

Not sure what you meant by that

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/946)* status: all files reviewed, 1 unresolved discussion, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`common/include/linux_socket.h` line 53 at r2](https://reviewable.io/reviews/gramineproject/gramine/946#-NEM8x4q0FFbt40vThXn:-NEYExWGE4Q7hr4J2gKU:b3t4mcd) ([raw file](https://github.com/gramineproject/gramine/blob/da13eee3b6dc57ab6a777a67972b9463b36ef3cc/common/include/linux_socket.h#L53)):*
> Idk either, I just thought you wanted them, so I'm fine with removing.

Hmm, I think it was either a different case or maybe Dmitrii?

> > and below?
> 
> Not sure what you meant by that

Ah, sorry, it's already good.

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/946)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/946)* status: all files reviewed, 2 unresolved discussions (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/946#-NE_UvTdA9ytIKYSZveH:-NE_UvTdA9ytIKYSZveI:b-g8mpq5) (no related file):*
Last minute thought: I just noticed that we don't say anywhere that this is an exception and we don't state the reason for it. Maybe we should do this in a comment in `pal.h`? Just to prevent others from following this in other cases.

___

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/946)* status: all files reviewed, 2 unresolved discussions (waiting on @boryspoplawski and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/946#-NE_UvTdA9ytIKYSZveH:-NE_fqG2DfMvlBqLYjkM:b-pmjo9u) (no related file):*

Last minute thought: I just noticed that we don't say anywhere that this is an exception and we don't state the reason for it. Maybe we should do this in a comment in `pal.h`? Just to prevent others from following this in other cases.
</blockquote></details>

This was already discussed in this PR (please check one of the resolved issues, Borys wrote a big reply). We decided against this since it wasn't obvious where to put the comment and what exactly to say in the comment. @mkow Do you still want such a comment?

___

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/946)* status: all files reviewed, 1 unresolved discussion (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/946#-NE_UvTdA9ytIKYSZveH:-NEb35oqAbHhDwHQkJJt:b-75vyrb) (no related file):*
I don't have more to add than in the other thread. I think that having a small comment would be nice, but adding it to every function is a bit overkill (and we cannot add it near type, because it's in header unrelated to PAL).

> This API directly uses struct iovec from Linux for performance reasons.

The problem is that the old struct had the exactly same layout, so the change traded basically nothing for performance. If the old struct was different, it's not so obvious we would still want this change to `struct iovec`, even for performance.

___

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/946)* status: all files reviewed, 1 unresolved discussion (waiting on @boryspoplawski and @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/946#-NE_UvTdA9ytIKYSZveH:-NEfDtzY27o6KB9TNGSB:b-wbc63c) (no related file):*
> but adding it to every function is a bit overkill

But there are only 2 such functions right now, and we hope we'll never add more. So, IMO it's ok to add it to both of them.

> The problem is that the old struct had the exactly same layout, so the change traded basically nothing for performance. If the old struct was different, it's not so obvious we would still want this change to `struct iovec`, even for performance.

Yeah, this case is very mild, but I'm really afraid that people will start using this exception to justify more aggressive stuff in PAL API in their PRs quoting this place with "I saw you already do it this way here, so I did the same".

___

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/946)* status: all files reviewed, 1 unresolved discussion (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/946#-NE_UvTdA9ytIKYSZveH:-NEfWrQm-xoRr7dJS8qH:b5dt8fg) (no related file):*

> but adding it to every function is a bit overkill

But there are only 2 such functions right now, and we hope we'll never add more. So, IMO it's ok to add it to both of them.

> The problem is that the old struct had the exactly same layout, so the change traded basically nothing for performance. If the old struct was different, it's not so obvious we would still want this change to `struct iovec`, even for performance.

Yeah, this case is very mild, but I'm really afraid that people will start using this exception to justify more aggressive stuff in PAL API in their PRs quoting this place with "I saw you already do it this way here, so I did the same".
</blockquote></details>

So should I add to both functions:
```
We use Linux `struct iovec` as argument here, because the alternative is to use a custom structure,  
but it would contain exactly the same fields, which would achieve nothing, but could worsen          
performance in certain cases.
```

___

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/946)* status: all files reviewed, 1 unresolved discussion (waiting on @boryspoplawski and @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/946#-NE_UvTdA9ytIKYSZveH:-NEffXUZ7W9Kblq7pv7o:brxp9sq) (no related file):*

So should I add to both functions:
```
We use Linux `struct iovec` as argument here, because the alternative is to use a custom structure,  
but it would contain exactly the same fields, which would achieve nothing, but could worsen          
performance in certain cases.
```
</blockquote></details>

Sounds good.

___

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/946)* status: 16 of 17 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/946#-NE_UvTdA9ytIKYSZveH:-NEfmjGa0QdoON4gI663:b-896fix) (no related file):*

Sounds good.
</blockquote></details>

Done.

___

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/946)* status: all files reviewed, 1 unresolved discussion (waiting on @mkow)

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/946)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #972, opened 2022-10-12
[common] Add truncation to arbitrary size to Protected Files

### LINKED ISSUE #950: [LibOS] Add support for shrinking encrypted files via truncate(2) to arbitrary size
Graminizing the ethereum [geth client](https://github.com/ethereum/go-ethereum), we are running into another issue regarding encrypted files. To protect the state of geth's persistent data from host access and manipulation, files are stored in a encrypted mount point. 

Syncing `goerli` testnet with geth into encrypted mount point on Gramine 1.3.1, we run into this error:
```
[P1:T12:geth] warning: encrypted_file_set_size: pf_set_size failed: Functionality not implemented
CRIT [10-03|17:28:11.357] Failed to truncate ancient data          number=1,325,056 err="truncate /root/.ethereum/goerli/geth/chaindata/ancient/chain/bodies.cidx: permission denied"
```
At times, geth needs to rewind the state of the blockchain to a previous block. It does so by shrinking the size of the database file to the specific block. Since shrinking encrypted files is not supported within gramine, this operation fails.

dimakuv:
## Description of the changes 

Truncation of PFs to arbitrary size is implemented in a slow and trivial way: read the file's contents into a temporary buffer, shrink the file to zero, write back the contents into the file.

Fixes #950.

## How to test this PR? 

CI. I added tests to `libos/test/fs`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/972)

### DIFF
--- a/common/src/protected_files/README.rst
+++ b/common/src/protected_files/README.rst
@@ -66,7 +66,9 @@ Some tests in ``libos/test/regression`` also work with encrypted files.
 TODO
 ====
 
-- Shrinking protected files via truncate(2) to arbitrary size is not yet implemented.
+- Shrinking protected files via truncate(2) to arbitrary size is implemented in
+  a slow and trivial way: read the file's contents into a temporary buffer,
+  shrink the file to zero, write back the contents into the file.
 - The recovery file feature is disabled, this needs to be discussed if it's
   needed in Gramine.
 - Tests for invalid/malformed/corrupted files need to be ported to the new
--- a/common/src/protected_files/protected_files.c
+++ b/common/src/protected_files/protected_files.c
@@ -1236,10 +1236,13 @@ pf_status_t pf_set_size(pf_context_t* pf, uint64_t size) {
         return PF_STATUS_SUCCESS;
     }
 
-    /* shrink the file
-     *    - to zero: emulated by recreating the file metadata as if it was just created
+    /* Shrink the file:
+     *    - to zero: emulated by recreating the file metadata as if it was just created,
      *    - to arbitrary size: emulated by reading file contents in a buffer, shrinking the file to
-     *                         zero as above, and writing back contents into the file */
+     *                         zero as above, and writing back contents into the file.
+     *
+     * TODO: implement "shrink to arbitrary size" properly.
+     */
     int ret;
     char* buf = NULL;
 
--- a/common/src/protected_files/protected_files.c
+++ b/common/src/protected_files/protected_files.c
@@ -1,7 +1,7 @@
 /* SPDX-License-Identifier: LGPL-3.0-or-later */
-/* Copyright (C) 2020 Invisible Things Lab
+/* Copyright (C) 2022 Intel Corporation
+ * Copyright (C) 2020 Invisible Things Lab
  *                    Rafal Wojdyla <omeg@invisiblethingslab.com>
- * Copyright (C) 2019 Intel Corporation
  */
 
 #include "protected_files.h"
@@ -1216,7 +1216,6 @@ pf_status_t pf_get_size(pf_context_t* pf, uint64_t* size) {
     return PF_STATUS_SUCCESS;
 }
 
-// TODO: File truncation to arbitrary size.
 pf_status_t pf_set_size(pf_context_t* pf, uint64_t size) {
     if (!g_initialized)
         return PF_STATUS_UNINITIALIZED;
@@ -1228,7 +1227,7 @@ pf_status_t pf_set_size(pf_context_t* pf, uint64_t size) {
         return PF_STATUS_SUCCESS;
 
     if (size > pf->encrypted_part_plain.size) {
-        // Extend the file.
+        /* extend the file */
         uint64_t offset = pf->encrypted_part_plain.size;
         DEBUG_PF("extending the file from %lu to %lu", offset, size);
         if (ipf_write(pf, NULL, offset, size - offset) != size - offset)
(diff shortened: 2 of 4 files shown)


boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/972)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`common/src/protected_files/protected_files.c` line 1239 at r1](https://reviewable.io/reviews/gramineproject/gramine/972#-NEDz1rsDOiJkKB-49M3:-NEDz1rsDOiJkKB-49M4:bq0vgym) ([raw file](https://github.com/gramineproject/gramine/blob/db5e74d8a2281621de8ebab313d22e8ea2a33a16/common/src/protected_files/protected_files.c#L1239)):*
> ```C
>     }
> 
>     /* shrink the file
> ```

Maybe we should have some `TODO` somewhere here to implement it properly someday?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/972)* status: 3 of 4 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`common/src/protected_files/protected_files.c` line 1239 at r1](https://reviewable.io/reviews/gramineproject/gramine/972#-NEDz1rsDOiJkKB-49M3:-NEFFc2g4iFtfqb0dknV:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/db5e74d8a2281621de8ebab313d22e8ea2a33a16/common/src/protected_files/protected_files.c#L1239)):*

Maybe we should have some `TODO` somewhere here to implement it properly someday?
</blockquote></details>

Done.
___
*[`common/src/protected_files/protected_files.c` line 1248 at r1](https://reviewable.io/reviews/gramineproject/gramine/972#-NEFFce15j3pfj3YD8-t:-NEFFce15j3pfj3YD8-u:brk6ac6) ([raw file](https://github.com/gramineproject/gramine/blob/db5e74d8a2281621de8ebab313d22e8ea2a33a16/common/src/protected_files/protected_files.c#L1248)):*
> ```C
>     if (size > 0) {
>         /* read the file's contents into a temporary buffer */
>         buf = malloc(size);
> ```

Do we want to introduce a stopper like
```
if (size > 10 * 1024 * 1024)
    log_error("Shrinking a file requires to copy in a temporary buffer, but the size of the buffer will be too big");
    return PF_STATUS_DENIED;
```

If yes, what would be the size at which we cut off?

I just want to highlight the problem of allocating an arbitrarily-large buffer here. I actually don't think we need a stopper IF clause, because it will be a random heuristic...

oshogbo (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/972)* status: 3 of 4 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[`common/src/protected_files/protected_files.c` line 1248 at r1](https://reviewable.io/reviews/gramineproject/gramine/972#-NEFFce15j3pfj3YD8-t:-NEFwGc5-A9BGKm2Jwnk:b-ezuuh4) ([raw file](https://github.com/gramineproject/gramine/blob/db5e74d8a2281621de8ebab313d22e8ea2a33a16/common/src/protected_files/protected_files.c#L1248)):*

Do we want to introduce a stopper like
```
if (size > 10 * 1024 * 1024)
    log_error("Shrinking a file requires to copy in a temporary buffer, but the size of the buffer will be too big");
    return PF_STATUS_DENIED;
```

If yes, what would be the size at which we cut off?

I just want to highlight the problem of allocating an arbitrarily-large buffer here. I actually don't think we need a stopper IF clause, because it will be a random heuristic...
</blockquote></details>

I think that allocating arbitrary size might be a problem in case of files.
People don't use protected files with databases which can grow kinda large?
Won't naive solution be just to change the size of data and erase the encrypted part? Even if the hosts file won't shrink it might be a safer solution.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/972)* status: 3 of 4 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @oshogbo)

___
*[`common/src/protected_files/protected_files.c` line 1248 at r1](https://reviewable.io/reviews/gramineproject/gramine/972#-NEFFce15j3pfj3YD8-t:-NEG6x1c1Urwp98QS1ZF:b-79olfk) ([raw file](https://github.com/gramineproject/gramine/blob/db5e74d8a2281621de8ebab313d22e8ea2a33a16/common/src/protected_files/protected_files.c#L1248)):*
> I think that allocating arbitrary size might be a problem in case of files.
> People don't use protected files with databases which can grow kinda large?

Yes, technically people can use PFs for databases, and then this `malloc(size)` may potentially allocate GB-sized buffers. That's why I'm asking about adding a stopper here.

> Won't naive solution be just to change the size of data and erase the encrypted part?

But erase how? Like, what is the pseudo-code for erasing a part of the file contents? For example, how do I truncate from 2 characters to 1 character -- what do I erase? (Recall that encrypted file contents are stored in 4KB blocks, and each block must be re-encrypted with a new derived key after each update.)

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/972)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @oshogbo)

___
*[`common/src/protected_files/protected_files.c` line 1248 at r1](https://reviewable.io/reviews/gramineproject/gramine/972#-NEFFce15j3pfj3YD8-t:-NEGJ66NCuWO2Q2oDMDe:b-b04er) ([raw file](https://github.com/gramineproject/gramine/blob/db5e74d8a2281621de8ebab313d22e8ea2a33a16/common/src/protected_files/protected_files.c#L1248)):*

> I think that allocating arbitrary size might be a problem in case of files.
> People don't use protected files with databases which can grow kinda large?

Yes, technically people can use PFs for databases, and then this `malloc(size)` may potentially allocate GB-sized buffers. That's why I'm asking about adding a stopper here.

> Won't naive solution be just to change the size of data and erase the encrypted part?

But erase how? Like, what is the pseudo-code for erasing a part of the file contents? For example, how do I truncate from 2 characters to 1 character -- what do I erase? (Recall that encrypted file contents are stored in 4KB blocks, and each block must be re-encrypted with a new derived key after each update.)
</blockquote></details>

Regarding cut off: I don't think it's needed - this memory will be freed soon after so having the natural limit of enclave size is IMHO enough.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/972)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @oshogbo)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/972#-NEkEl-Y4Fl34APMDzAf:-NEkEl-Y4Fl34APMDzAg:bkma4m9) (no related file):*
Requesting review of @mkow and blocking until he checks.

___

oshogbo (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/972)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

___
*[`common/src/protected_files/protected_files.c` line 1248 at r1](https://reviewable.io/reviews/gramineproject/gramine/972#-NEFFce15j3pfj3YD8-t:-NEvBxu8E25ez7cO8j7c:bsu3kaa) ([raw file](https://github.com/gramineproject/gramine/blob/db5e74d8a2281621de8ebab313d22e8ea2a33a16/common/src/protected_files/protected_files.c#L1248)):*

Regarding cut off: I don't think it's needed - this memory will be freed soon after so having the natural limit of enclave size is IMHO enough. 
</blockquote></details>

I was thinking something like:
```
set_header(newsize);

memset(zero_buf, 0 sizeof(*buf));
for (size_t i = newsize; i < actuall_size; i +=sizeof(*buf)) {
      pf_write(pf, size + i, MIN(sizeof(*buf), actuall_size - i), zero_buf);
}
``` 

Or something like that. My goal was just to implement truncate without, really changing the size on the disk.
But I believe mkow prefer other solution.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/972)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`common/src/protected_files/protected_files.c` line 1248 at r1](https://reviewable.io/reviews/gramineproject/gramine/972#-NEFFce15j3pfj3YD8-t:-NF8Dc9e4WOx6umI1PYJ:b-8by9ra) ([raw file](https://github.com/gramineproject/gramine/blob/db5e74d8a2281621de8ebab313d22e8ea2a33a16/common/src/protected_files/protected_files.c#L1248)):*

I was thinking something like:
```
set_header(newsize);

memset(zero_buf, 0 sizeof(*buf));
for (size_t i = newsize; i < actuall_size; i +=sizeof(*buf)) {
      pf_write(pf, size + i, MIN(sizeof(*buf), actuall_size - i), zero_buf);
}
``` 

Or something like that. My goal was just to implement truncate without, really changing the size on the disk.
But I believe mkow prefer other solution.
</blockquote></details>

I see, yes, we could use @oshogbo solution as well.

@mkow What solution do you prefer?

dimakuv:
We hit this again on a MariaDB workload -- without this PR it fails with `ftruncate: Permission denied` on a Protected File. With this PR, it succeeds initialization and continues working.

We should prioritize finishing this PR.

PR #976, opened 2022-10-13
[LibOS] Add uid and gid to inodes

dimakuv:
## Description of the changes 

Previously, inodes didn't have user/group IDs, thus `stat("file")` returned `statbuf.st_uid == statbuf.st_uid == 0`. This is wrong because Gramine allows to specify initial user/group IDs via `loader.uid` and `loader.gid` manifest options, and file ownership must default to these.

This PR adds `inode->uid` and `inode->gid`. Also, `chown()`, `fchown()` and `fchownat()` syscalls now change file uid/gid. Several LTP tests were enabled to test this feature.

## How to test this PR? 

I enabled some LTP tests.

### DIFF
--- a/libos/include/libos_fs.h
+++ b/libos/include/libos_fs.h
@@ -257,6 +257,9 @@ struct libos_inode {
     /* File size */
     file_off_t size;
 
+    /* credentials */
+    IDTYPE uid, gid;
+
     /* Create/modify/access time */
     time_t ctime;
     time_t mtime;
--- a/libos/src/fs/libos_dcache.c
+++ b/libos/src/fs/libos_dcache.c
@@ -361,6 +361,12 @@ struct libos_inode* get_new_inode(struct libos_mount* mount, mode_t type, mode_t
     inode->mtime = 0;
     inode->atime = 0;
 
+    struct libos_thread* current = get_cur_thread();
+    lock(&current->lock);
+    inode->uid = current->uid;
+    inode->gid = current->gid;
+    unlock(&current->lock);
+
     inode->mount = mount;
     get_mount(mount);
     inode->fs = mount->fs;
@@ -610,6 +616,9 @@ BEGIN_CP_FUNC(inode) {
         new_inode->type = inode->type;
         new_inode->perm = inode->perm;
         new_inode->size = inode->size;
+
+        new_inode->uid = inode->uid;
+        new_inode->gid = inode->gid;
 
         new_inode->ctime = inode->ctime;
         new_inode->mtime = inode->mtime;
--- a/libos/src/fs/libos_fs_pseudo.c
+++ b/libos/src/fs/libos_fs_pseudo.c
@@ -210,6 +210,9 @@ static int pseudo_istat(struct libos_dentry* dent, struct libos_inode* inode, st
     memset(buf, 0, sizeof(*buf));
     buf->st_dev = 1;
     buf->st_mode = inode->type | inode->perm;
+    buf->st_uid  = inode->uid;
+    buf->st_gid  = inode->gid;
+
     struct pseudo_node* node = inode->data;
     switch (node->type) {
         case PSEUDO_DIR: {
--- a/libos/src/fs/libos_fs_util.c
+++ b/libos/src/fs/libos_fs_util.c
@@ -62,6 +62,9 @@ static int generic_istat(struct libos_inode* inode, struct stat* buf) {
     lock(&inode->lock);
     buf->st_mode = inode->type | inode->perm;
     buf->st_size = inode->size;
+    buf->st_uid  = inode->uid;
+    buf->st_gid  = inode->gid;
+
     /* Some programs (e.g. some tests from LTP) require this value. We've picked some random,
      * pretty looking constant - exact value should not affect anything (perhaps except
      * performance). */
--- a/libos/src/libos_init.c
+++ b/libos/src/libos_init.c
@@ -420,8 +420,8 @@ noreturn void libos_init(const char* const* argv, const char* const* envp) {
 
     RUN_INIT(init_ipc);
     RUN_INIT(init_process);
-    RUN_INIT(init_mount_root);
     RUN_INIT(init_threading);
+    RUN_INIT(init_mount_root);
     RUN_INIT(init_mount);
     RUN_INIT(init_std_handles);
 
--- a/libos/src/sys/libos_file.c
+++ b/libos/src/sys/libos_file.c
@@ -267,13 +267,17 @@ long libos_syscall_fchownat(int dfd, const char* filename, uid_t uid, gid_t gid,
 
     lock(&g_dcache_lock);
     ret = path_lookupat(dir, filename, LOOKUP_FOLLOW, &dent);
-    unlock(&g_dcache_lock);
     if (ret < 0)
         goto out;
 
-    /* XXX: do nothing now */
+    lock(&dent->inode->lock);
+    dent->inode->uid = (uid == (uid_t)-1) ? dent->inode->uid : uid;
+    dent->inode->gid = (gid == (gid_t)-1) ? dent->inode->gid : gid;
+    unlock(&dent->inode->lock);
+
     put_dentry(dent);
 out:
+    unlock(&g_dcache_lock);
     if (dir)
         put_dentry(dir);
     return ret;
@@ -287,8 +291,25 @@ long libos_syscall_fchown(int fd, uid_t uid, gid_t gid) {
     if (!hdl)
         return -EBADF;
 
-    /* XXX: do nothing now */
-    return 0;
+    int ret;
+    struct libos_dentry* dent = hdl->dentry;
+
+    lock(&g_dcache_lock);
+    if (!dent || !dent->inode) {
+        ret = -ENOENT;
+        goto out;
+    }
+
+    lock(&dent->inode->lock);
+    dent->inode->uid = (uid == (uid_t)-1) ? dent->inode->uid : uid;
+    dent->inode->gid = (gid == (gid_t)-1) ? dent->inode->gid : gid;
+    unlock(&dent->inode->lock);
+
+    ret = 0;
+out:
+    unlock(&g_dcache_lock);
+    put_handle(hdl);
+    return ret;
 }
 
 static int do_rename(struct libos_dentry* old_dent, struct libos_dentry* new_dent) {
--- a/libos/test/ltp/ltp.cfg
+++ b/libos/test/ltp/ltp.cfg
@@ -121,8 +121,20 @@ skip = yes
 [chmod07]
 skip = yes
 
-# chown() does nothing
-[chown*]
+# 16-bit unsupported
+[chown*_16]
+skip = yes
+
+# chown() does not clear setuid/setgid bits set on executable files
+[chown02]
+skip = yes
+
+# no seteuid()
+[chown03]
+skip = yes
+
+# requires using a device
+[chown04]
 skip = yes
 
 # no seteuid()
@@ -355,8 +367,24 @@ skip = yes
 [fchmodat01]
 timeout = 40
 
-# fchown* does nothing
-[fchown*]
+# 16-bit unsupported
+[fchown*_16]
+skip = yes
+
+# fchown() does not clear setuid/setgid bits set on executable files
+[fchown02]
+skip = yes
+
+# no seteuid()
+[fchown03]
+skip = yes
+
+# requires using a device
+[fchown04]
+skip = yes
+
+# no utimensat()
+[fchownat*]
 skip = yes
 
 [fcntl02_64]


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/976)* status: 0 of 7 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

___
*[`libos/src/libos_init.c` line 424 at r1](https://reviewable.io/reviews/gramineproject/gramine/976#-NEGqIwR9ph8JPM-Ck4L:-NEGqIwR9ph8JPM-Ck4M:b-5efgkz) ([raw file](https://github.com/gramineproject/gramine/blob/55af33d7e475326a29f7ef96f84ea6df1ec525c7/libos/src/libos_init.c#L424)):*
> ```C
>     RUN_INIT(init_process);
>     RUN_INIT(init_threading);
>     RUN_INIT(init_mount_root);
> ```

FYI: This was needed because mounting creates files (and thus sets up `inode->uid/gid = current_thread->uid/gid`). And setting up uid/gid happens in `init_threading()`.
___
*[`libos/src/fs/libos_dcache.c` line 367 at r1](https://reviewable.io/reviews/gramineproject/gramine/976#-NEGqZPCCj1tmqj_GeLg:-NEGqZPCCj1tmqj_GeLh:b-y0j4ia) ([raw file](https://github.com/gramineproject/gramine/blob/55af33d7e475326a29f7ef96f84ea6df1ec525c7/libos/src/fs/libos_dcache.c#L367)):*
> ```C
>     lock(&current->lock);
>     inode->uid = current->uid;
>     inode->gid = current->gid;
> ```

FYI: I was surprised to learn that files are created with a real UID, not effective UID. I first wrote `current->euid`, but LTP tests failed, and then I learned that it should be `current->uid`.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/976)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`libos/src/libos_init.c` line 424 at r1](https://reviewable.io/reviews/gramineproject/gramine/976#-NEGqIwR9ph8JPM-Ck4L:-NEGv03o0A9zSe9OdBsD:bxpoepn) ([raw file](https://github.com/gramineproject/gramine/blob/55af33d7e475326a29f7ef96f84ea6df1ec525c7/libos/src/libos_init.c#L424)):*

FYI: This was needed because mounting creates files (and thus sets up `inode->uid/gid = current_thread->uid/gid`). And setting up uid/gid happens in `init_threading()`.
</blockquote></details>

Seems that nothing in `init_threading` touches filesystem, so it should be fine
___
*[`libos/src/fs/libos_dcache.c` line 367 at r1](https://reviewable.io/reviews/gramineproject/gramine/976#-NEGqZPCCj1tmqj_GeLg:-NEGv736CVnXEcelzRcp:bsd23r2) ([raw file](https://github.com/gramineproject/gramine/blob/55af33d7e475326a29f7ef96f84ea6df1ec525c7/libos/src/fs/libos_dcache.c#L367)):*

FYI: I was surprised to learn that files are created with a real UID, not effective UID. I first wrote `current->euid`, but LTP tests failed, and then I learned that it should be `current->uid`.
</blockquote></details>

Are you sure? I though `fsuid` is used, which is almost always equal to `euid`. Also man states `The owner (user ID) of the new file is set to the effective user ID of the process.` (tho I would assume `fsuid` is used).

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/976)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski)

___
*[`libos/src/fs/libos_dcache.c` line 367 at r1](https://reviewable.io/reviews/gramineproject/gramine/976#-NEGqZPCCj1tmqj_GeLg:-NEH4w2581laOi3BIRx0:b-fub8t2) ([raw file](https://github.com/gramineproject/gramine/blob/55af33d7e475326a29f7ef96f84ea6df1ec525c7/libos/src/fs/libos_dcache.c#L367)):*

Are you sure? I though `fsuid` is used, which is almost always equal to `euid`. Also man states `The owner (user ID) of the new file is set to the effective user ID of the process.` (tho I would assume `fsuid` is used).
</blockquote></details>

I encountered this in the `stat01` test case (which failed with `current->euid`): https://github.com/linux-test-project/ltp/blob/master/testcases/kernel/syscalls/stat/stat01.c

If you go through the source code of this LTP test, you see that it:
1. Sets `euid` to the `nobody` user: https://github.com/linux-test-project/ltp/blob/ff9465c5b2a3284790c9955c3ea980fa39736b46/testcases/kernel/syscalls/stat/stat01.c#L57
2. Creates a temporary file: https://github.com/linux-test-project/ltp/blob/ff9465c5b2a3284790c9955c3ea980fa39736b46/testcases/kernel/syscalls/stat/stat01.c#L60
3. For verification of this temporary file, it uses the `uid`: https://github.com/linux-test-project/ltp/blob/ff9465c5b2a3284790c9955c3ea980fa39736b46/testcases/kernel/syscalls/stat/stat01.c#L66
4. Then verification of `stat_buf.st_uid` happens: https://github.com/linux-test-project/ltp/blob/ff9465c5b2a3284790c9955c3ea980fa39736b46/testcases/kernel/syscalls/stat/stat01.c#L45

I quickly browsed through Linux source code, but couldn't find the corresponding place and gave up.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/976)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`libos/src/fs/libos_dcache.c` line 367 at r1](https://reviewable.io/reviews/gramineproject/gramine/976#-NEGqZPCCj1tmqj_GeLg:-NEHD-AOA--OrwBTJHLz:b-igq8hq) ([raw file](https://github.com/gramineproject/gramine/blob/55af33d7e475326a29f7ef96f84ea6df1ec525c7/libos/src/fs/libos_dcache.c#L367)):*

I encountered this in the `stat01` test case (which failed with `current->euid`): https://github.com/linux-test-project/ltp/blob/master/testcases/kernel/syscalls/stat/stat01.c

If you go through the source code of this LTP test, you see that it:
1. Sets `euid` to the `nobody` user: https://github.com/linux-test-project/ltp/blob/ff9465c5b2a3284790c9955c3ea980fa39736b46/testcases/kernel/syscalls/stat/stat01.c#L57
2. Creates a temporary file: https://github.com/linux-test-project/ltp/blob/ff9465c5b2a3284790c9955c3ea980fa39736b46/testcases/kernel/syscalls/stat/stat01.c#L60
3. For verification of this temporary file, it uses the `uid`: https://github.com/linux-test-project/ltp/blob/ff9465c5b2a3284790c9955c3ea980fa39736b46/testcases/kernel/syscalls/stat/stat01.c#L66
4. Then verification of `stat_buf.st_uid` happens: https://github.com/linux-test-project/ltp/blob/ff9465c5b2a3284790c9955c3ea980fa39736b46/testcases/kernel/syscalls/stat/stat01.c#L45

I quickly browsed through Linux source code, but couldn't find the corresponding place and gave up.
</blockquote></details>

This test must be run as root: https://github.com/linux-test-project/ltp/blob/ff9465c5b2a3284790c9955c3ea980fa39736b46/testcases/kernel/syscalls/stat/stat01.c#L74 and `setuid` behaves differently depending if previous `euid` was `0`.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/976)* status: 6 of 8 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski)

___
*[`-- commits` line 11 at r2](https://reviewable.io/reviews/gramineproject/gramine/976#-NEKb3dv7X2REsF79GY6):*
@boryspoplawski Is this a reasonable simplification?
___
*[`libos/src/fs/libos_dcache.c` line 367 at r1](https://reviewable.io/reviews/gramineproject/gramine/976#-NEGqZPCCj1tmqj_GeLg:-NEKaa7aBpxKELoHWO1y:b-azzumo) ([raw file](https://github.com/gramineproject/gramine/blob/55af33d7e475326a29f7ef96f84ea6df1ec525c7/libos/src/fs/libos_dcache.c#L367)):*

This test must be run as root: https://github.com/linux-test-project/ltp/blob/ff9465c5b2a3284790c9955c3ea980fa39736b46/testcases/kernel/syscalls/stat/stat01.c#L74 and `setuid` behaves differently depending if previous `euid` was `0`.
</blockquote></details>

Done. Thanks, I completely missed this corner case of `setuid`! I added a new commit (the first one, to preserve correctness of the second commit) and had to rebase this PR.

Now it all works and makes sense. (For context, this is how setuid logic looks in Linux: https://elixir.bootlin.com/linux/v6.0/source/kernel/sys.c#L602)
___
*[`libos/src/sys/libos_getuid.c` line 61 at r2](https://reviewable.io/reviews/gramineproject/gramine/976#-NEKb9E30HhSpmkSUnEb:-NEKb9E30HhSpmkSUnEc:b-cx7fja) ([raw file](https://github.com/gramineproject/gramine/blob/0411978a21519bf03551600419e491c61a93106f/libos/src/sys/libos_getuid.c#L61)):*
> ```C
>     struct libos_thread* current = get_cur_thread();
>     lock(&current->lock);
>     if (current->uid == 0) {
> ```

FYI: Note that I use `uid` (user ID) even for gid checking.

boryspoplawski (changes requested):
> Reviewed 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/976)* status: 7 of 8 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`-- commits` line 3 at r2](https://reviewable.io/reviews/gramineproject/gramine/976#-NELYQGTA9aDRUAYEuZf):*
This oneliner definitely lacks context.
___
*[`-- commits` line 11 at r2](https://reviewable.io/reviews/gramineproject/gramine/976#-NEKb3dv7X2REsF79GY6):*

@boryspoplawski Is this a reasonable simplification?
</blockquote></details>

Let's first resolve other comments, the commit msg depends on them
___
*[`libos/src/sys/libos_getuid.c` line 49 at r2](https://reviewable.io/reviews/gramineproject/gramine/976#-NELZI39DGF4T5YiY8sB:-NELZI3A0jIykl8gA-mZ:b59f8ez) ([raw file](https://github.com/gramineproject/gramine/blob/0411978a21519bf03551600419e491c61a93106f/libos/src/sys/libos_getuid.c#L49)):*
> ```C
>     struct libos_thread* current = get_cur_thread();
>     lock(&current->lock);
>     if (current->uid == 0) {
> ```

So we never really talked about it, mainly because it makes no sense in Gramine - what do we consider a privileged process? For now we considered everything, but here you added some local assumption that `uid == 0` is privileged.
___
*[`libos/src/sys/libos_getuid.c` line 50 at r2](https://reviewable.io/reviews/gramineproject/gramine/976#-NELZF9M1E7RvC1CudQN:-NELZF9M1E7RvC1CudQO:b-40fafr) ([raw file](https://github.com/gramineproject/gramine/blob/0411978a21519bf03551600419e491c61a93106f/libos/src/sys/libos_getuid.c#L50)):*
> ```C
>     lock(&current->lock);
>     if (current->uid == 0) {
>         /* if the process is privileged, the real UID is also set */
> ```

It's not really how this works in Linux. If process is privileged then yes, all uids are set, but if it's not privileged, then:
- `user_requested_uid == saved_uid == euid` then rest is also set to `user_requested_uid` (so effectively all uids are now equal, just like in privileged case)
- not equal - syscall fails.

So either syscall fails, or all uids are the same.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/976)* status: 7 of 8 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski and @mkow)

___
*[`libos/src/sys/libos_getuid.c` line 49 at r2](https://reviewable.io/reviews/gramineproject/gramine/976#-NELZI39DGF4T5YiY8sB:-NELhv_94agKI2j6WgDl:b93g908) ([raw file](https://github.com/gramineproject/gramine/blob/0411978a21519bf03551600419e491c61a93106f/libos/src/sys/libos_getuid.c#L49)):*

So we never really talked about it, mainly because it makes no sense in Gramine - what do we consider a privileged process? For now we considered everything, but here you added some local assumption that `uid == 0` is privileged.
</blockquote></details>

Ok, I have another proposal: let's remove `euid` and keep only `uid`. Gramine doesn't distinguish between these two anyway, so having just one UID should be enough for our simple logic. This will simplify the logic and Gramine will stop pretending as if `euid` means anything to it.

What do you think? Also summon @mkow 

By the way, we *do* have this assumption of `uid == 0` being privileged in this file (in two places): https://github.com/gramineproject/gramine/blob/546c598b694d091121babc215bf6b9f9f52bbb4d/libos/src/sys/libos_getrlimit.c#L104

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/976)* status: 7 of 8 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv and @mkow)

___
*[`libos/src/sys/libos_getuid.c` line 49 at r2](https://reviewable.io/reviews/gramineproject/gramine/976#-NELZI39DGF4T5YiY8sB:-NELokYfC1a5hDgtivjE:b-cqxtns) ([raw file](https://github.com/gramineproject/gramine/blob/0411978a21519bf03551600419e491c61a93106f/libos/src/sys/libos_getuid.c#L49)):*
> Gramine doesn't distinguish between these two anyway,

Why do you say so?

> let's remove euid and keep only uid

I don't like it, what if some app tries to change euid temporarily and then restore it to uid? It might brake if it has some sanity checks...

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/976)* status: 7 of 10 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[`libos/src/sys/libos_getuid.c` line 49 at r2](https://reviewable.io/reviews/gramineproject/gramine/976#-NELZI39DGF4T5YiY8sB:-NE_Dyxo6Isnk6YjV4bm:bjwluad) ([raw file](https://github.com/gramineproject/gramine/blob/0411978a21519bf03551600419e491c61a93106f/libos/src/sys/libos_getuid.c#L49)):*
> > Gramine doesn't distinguish between these two anyway,
> 
> Why do you say so?

Fair enough, my comment was too vague. I meant that Gramine doesn't enforce anything based on `uid` and `euid`. But I agree that even the fact that Gramine allows to set `uid` and `euid` separately and allows to return these values separately may be important for execution of some programs (which e.g. do some sanity checks on `uid == euid`).

>> let's remove euid and keep only uid
>
> I don't like it, what if some app tries to change euid temporarily and then restore it to uid? It might brake if it has some sanity checks...

Yes, agreed. Then I'll need to add the logic from your other comment in this PR, and keep `uid` and `euid` separately, as is currently done.
___
*[`libos/src/sys/libos_getuid.c` line 50 at r2](https://reviewable.io/reviews/gramineproject/gramine/976#-NELZF9M1E7RvC1CudQN:-NE_dDW67P8VSr8Sj1Tu:biv5dix) ([raw file](https://github.com/gramineproject/gramine/blob/0411978a21519bf03551600419e491c61a93106f/libos/src/sys/libos_getuid.c#L50)):*

It's not really how this works in Linux. If process is privileged then yes, all uids are set, but if it's not privileged, then:
- `user_requested_uid == saved_uid == euid` then rest is also set to `user_requested_uid` (so effectively all uids are now equal, just like in privileged case)
- not equal - syscall fails.

So either syscall fails, or all uids are the same.
</blockquote></details>

To be honest, I didn't understand your comment. I think your comment is wrong. I read the following references:
- https://man7.org/linux/man-pages/man2/setuid.2.html
- https://en.wikipedia.org/wiki/User_identifier#Process_attributes
- https://people.eecs.berkeley.edu/~daw/papers/setuid-usenix02.pdf
- https://elixir.bootlin.com/linux/v6.0/source/kernel/sys.c#L602

And implemented the logic based on what I've understood from my readings. Please check.

`suid` and `sgid` are currently useless, but I wanted to introduce them now and then have one more PR to provide more syscalls on UIDs.
___
*[`libos/src/sys/libos_getuid.c` line 61 at r2](https://reviewable.io/reviews/gramineproject/gramine/976#-NEKb9E30HhSpmkSUnEb:-NE_d5JQ6IoC8zPZngey:bowo8hu) ([raw file](https://github.com/gramineproject/gramine/blob/0411978a21519bf03551600419e491c61a93106f/libos/src/sys/libos_getuid.c#L61)):*

FYI: Note that I use `uid` (user ID) even for gid checking.
</blockquote></details>

FYI: Now I used `gid` in this check, which looks more in line with what Linux does.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/976)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`libos/src/sys/libos_getuid.c` line 50 at r2](https://reviewable.io/reviews/gramineproject/gramine/976#-NELZF9M1E7RvC1CudQN:-NEajgig-i0O-ApY9IBm:b10os7w) ([raw file](https://github.com/gramineproject/gramine/blob/0411978a21519bf03551600419e491c61a93106f/libos/src/sys/libos_getuid.c#L50)):*

To be honest, I didn't understand your comment. I think your comment is wrong. I read the following references:
- https://man7.org/linux/man-pages/man2/setuid.2.html
- https://en.wikipedia.org/wiki/User_identifier#Process_attributes
- https://people.eecs.berkeley.edu/~daw/papers/setuid-usenix02.pdf
- https://elixir.bootlin.com/linux/v6.0/source/kernel/sys.c#L602

And implemented the logic based on what I've understood from my readings. Please check.

`suid` and `sgid` are currently useless, but I wanted to introduce them now and then have one more PR to provide more syscalls on UIDs.
</blockquote></details>

Yes, sorry, it was supposed to be that only one of `suid`, `ruid` must be equal to the request.

Compile and chwon the executable file to some other user (but not root) and `chmod u+s`, then run.
```c
#define CHECK(x) ({ \
    __typeof__(x) _x = (x); \
    if (_x == -1) { \
        err(1, "error at %s (line %d)", #x, __LINE__); \
    } \
    _x; \
})

int main(void) {
    setbuf(stdout, NULL);
    setbuf(stderr, NULL);

    uid_t ruid, euid, suid;
    CHECK(syscall(SYS_getresuid, &ruid, &euid, &suid));
    printf("ruid: %d, euid: %d, suid: %d\n", ruid, euid, suid);

    CHECK(syscall(SYS_setresuid, -1, -1, ruid));

    CHECK(syscall(SYS_getresuid, &ruid, &euid, &suid));
    printf("ruid: %d, euid: %d, suid: %d\n", ruid, euid, suid);

    CHECK(syscall(SYS_setuid, euid));

    CHECK(syscall(SYS_getresuid, &ruid, &euid, &suid));
    printf("ruid: %d, euid: %d, suid: %d\n", ruid, euid, suid);

    return 0;
}
```
___
*[`libos/src/sys/libos_getuid.c` line 61 at r2](https://reviewable.io/reviews/gramineproject/gramine/976#-NEKb9E30HhSpmkSUnEb:-NEaoJKWChIqCaeg7PlH:b-b2mzxb) ([raw file](https://github.com/gramineproject/gramine/blob/0411978a21519bf03551600419e491c61a93106f/libos/src/sys/libos_getuid.c#L61)):*

FYI: Now I used `gid` in this check, which looks more in line with what Linux does.
</blockquote></details>

The kernel checks if process has `CAP_SETGID` which is what `uid == 0` was before there were capabilities.
___
*[`libos/src/sys/libos_getuid.c` line 61 at r3](https://reviewable.io/reviews/gramineproject/gramine/976#-NEajpg1CgGKjr4-jzOf:-NEajpg1CgGKjr4-jzOg:b9thr77) ([raw file](https://github.com/gramineproject/gramine/blob/bfefa3d484f5af557b58caf8e400d479d1abc217/libos/src/sys/libos_getuid.c#L61)):*
> ```C
>     }
> 
>     if (uid != current->uid && uid != current->euid && uid != current->suid) {
> ```

See the comment above.



_Suggestion:_
```C
if (uid != current->uid && uid != current->suid) {
```
___
*[`libos/src/sys/libos_getuid.c` line 61 at r3](https://reviewable.io/reviews/gramineproject/gramine/976#-NEakgvTD7m7Fzl4X4IS:-NEakgvTD7m7Fzl4X4IT:bx1nmg2) ([raw file](https://github.com/gramineproject/gramine/blob/bfefa3d484f5af557b58caf8e400d479d1abc217/libos/src/sys/libos_getuid.c#L61)):*
> ```C
>     }
> 
>     if (uid != current->uid && uid != current->euid && uid != current->suid) {
> ```

Why not do it exactly as in kernel, i.e. use `else if` here? The current version has more gotos than necessary, including one for a non-error path, which I dislike. You can:
```
if (current->uid == 0) {
    current->uid = uid;
    current->suid  = uid;
} else if (uid != current->uid && uid != current->suid) {
    ret = -EPERM;
    goto error;
}

current->euid = uid;
ret = 0;
```

Not blocking though
___
*[`libos/src/sys/libos_getuid.c` line 88 at r3](https://reviewable.io/reviews/gramineproject/gramine/976#-NEaoZ9GCMCOYtXWLZoB:-NEaoZ9GCMCOYtXWLZoC:b-2jurdy) ([raw file](https://github.com/gramineproject/gramine/blob/bfefa3d484f5af557b58caf8e400d479d1abc217/libos/src/sys/libos_getuid.c#L88)):*
> ```C
>     }
> 
>     if (gid != current->gid && gid != current->egid && gid != current->sgid) {
> ```

ditto as in `setuid`

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/976)* status: 9 of 10 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`libos/src/sys/libos_getuid.c` line 50 at r2](https://reviewable.io/reviews/gramineproject/gramine/976#-NELZF9M1E7RvC1CudQN:-NEbcnSU2q8WTe8T1SOa:b-lfodp1) ([raw file](https://github.com/gramineproject/gramine/blob/0411978a21519bf03551600419e491c61a93106f/libos/src/sys/libos_getuid.c#L50)):*

Yes, sorry, it was supposed to be that only one of `suid`, `ruid` must be equal to the request.

Compile and chwon the executable file to some other user (but not root) and `chmod u+s`, then run.
```c
#define CHECK(x) ({ \
    __typeof__(x) _x = (x); \
    if (_x == -1) { \
        err(1, "error at %s (line %d)", #x, __LINE__); \
    } \
    _x; \
})

int main(void) {
    setbuf(stdout, NULL);
    setbuf(stderr, NULL);

    uid_t ruid, euid, suid;
    CHECK(syscall(SYS_getresuid, &ruid, &euid, &suid));
    printf("ruid: %d, euid: %d, suid: %d\n", ruid, euid, suid);

    CHECK(syscall(SYS_setresuid, -1, -1, ruid));

    CHECK(syscall(SYS_getresuid, &ruid, &euid, &suid));
    printf("ruid: %d, euid: %d, suid: %d\n", ruid, euid, suid);

    CHECK(syscall(SYS_setuid, euid));

    CHECK(syscall(SYS_getresuid, &ruid, &euid, &suid));
    printf("ruid: %d, euid: %d, suid: %d\n", ruid, euid, suid);

    return 0;
}
```
</blockquote></details>

You are right, surprisingly `uid == current->euid` is not allowed:
```
$ ./a.out
ruid: 1002, euid: 65534, suid: 65534
ruid: 1002, euid: 65534, suid: 1002
a.out: error at syscall(SYS_setuid, euid) (line 30): Operation not permitted
```

My user is 1002, I chowned the executable to `nobody` (65534). So we see that we can't `setuid()` to the already-set `euid`, which I find interesting. Well, the man page of `setuid` is quite clear, so it's my blooper:
```
EPERM  The user is not privileged (Linux: does not have the CAP_SETUID capability in its user  namespace)  and  uid
       does not match the real UID or saved set-user-ID of the calling process.
```

---

When I run the same program from set-user-ID-root, then I get the following:
```
$ ll -h a.out
-rwsrwxr-x 1 root root 17K Oct 17 13:20 a.out*

$ ./a.out
ruid: 1002, euid: 0, suid: 0
ruid: 1002, euid: 0, suid: 1002
ruid: 0, euid: 0, suid: 0
```

Which corresponds to this snippet from `man setuid`:
```
       If  the  user is root or the program is set-user-ID-root, special care must be taken: setuid() checks the effective
       user ID of the caller and if it is the superuser, all process-related user ID's are set to uid.  After this has oc‐
       curred, it is impossible for the program to regain root privileges.
```

Just for info, maybe others will find this useful.
___
*[`libos/src/sys/libos_getuid.c` line 61 at r2](https://reviewable.io/reviews/gramineproject/gramine/976#-NEKb9E30HhSpmkSUnEb:-NEbgUXy08TneFt8kNrM:ba97nta) ([raw file](https://github.com/gramineproject/gramine/blob/0411978a21519bf03551600419e491c61a93106f/libos/src/sys/libos_getuid.c#L61)):*

The kernel checks if process has `CAP_SETGID` which is what `uid == 0` was before there were capabilities.
</blockquote></details>

Done. Thanks for the historical info.
___
*[`libos/src/sys/libos_getuid.c` line 61 at r3](https://reviewable.io/reviews/gramineproject/gramine/976#-NEajpg1CgGKjr4-jzOf:-NEbgNpg6CvTfbmM-NTk:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/bfefa3d484f5af557b58caf8e400d479d1abc217/libos/src/sys/libos_getuid.c#L61)):*

See the comment above.
</blockquote></details>

Done.
___
*[`libos/src/sys/libos_getuid.c` line 61 at r3](https://reviewable.io/reviews/gramineproject/gramine/976#-NEakgvTD7m7Fzl4X4IS:-NEbgOIS3-Nx_HDA77Uh:b-u3shu1) ([raw file](https://github.com/gramineproject/gramine/blob/bfefa3d484f5af557b58caf8e400d479d1abc217/libos/src/sys/libos_getuid.c#L61)):*

Why not do it exactly as in kernel, i.e. use `else if` here? The current version has more gotos than necessary, including one for a non-error path, which I dislike. You can:
```
if (current->uid == 0) {
    current->uid = uid;
    current->suid  = uid;
} else if (uid != current->uid && uid != current->suid) {
    ret = -EPERM;
    goto error;
}

current->euid = uid;
ret = 0;
```

Not blocking though
</blockquote></details>

Done. Doesn't matter for me.
___
*[`libos/src/sys/libos_getuid.c` line 88 at r3](https://reviewable.io/reviews/gramineproject/gramine/976#-NEaoZ9GCMCOYtXWLZoB:-NEbgo1T7SxS2TR-zvDp:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/bfefa3d484f5af557b58caf8e400d479d1abc217/libos/src/sys/libos_getuid.c#L88)):*

ditto as in `setuid`
</blockquote></details>

Done.

boryspoplawski (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/976)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`-- commits` line 3 at r2](https://reviewable.io/reviews/gramineproject/gramine/976#-NELYQGTA9aDRUAYEuZf):*

This oneliner definitely lacks context.
</blockquote></details>

Ok, so we need something about this change being inside `setuid/setgid` syscalls. I guess I would be fine with just ```[LibOS] Rework `setuid` and `setgid` syscalls```

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/976)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[`-- commits` line 3 at r2](https://reviewable.io/reviews/gramineproject/gramine/976#-NELYQGTA9aDRUAYEuZf):*

Ok, so we need something about this change being inside `setuid/setgid` syscalls. I guess I would be fine with just ```[LibOS] Rework `setuid` and `setgid` syscalls```
</blockquote></details>

Yes, I agree. I don't see a need to have a long commit message since we simply made `setuid` and `setgid` syscalls more functional.

So I'll change to this during final rebase:
```
[LibOS] Rework `setuid` and `setgid` syscalls

Signed-off-by: Dmitrii Kuvaiskii <dmitrii.kuvaiskii@intel.com>
```

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/976)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`-- commits` line 3 at r2](https://reviewable.io/reviews/gramineproject/gramine/976#-NELYQGTA9aDRUAYEuZf):*

Yes, I agree. I don't see a need to have a long commit message since we simply made `setuid` and `setgid` syscalls more functional.

So I'll change to this during final rebase:
```
[LibOS] Rework `setuid` and `setgid` syscalls

Signed-off-by: Dmitrii Kuvaiskii <dmitrii.kuvaiskii@intel.com>
```
</blockquote></details>

I meant just as one liner, IMO you can leave the commit message body as is

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/976)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #977, opened 2022-10-13
Generate unique sgx trusted files for manifest.sgx

anjalirai-intel:
Gramine currently does not throw error when there are duplicate
entries in manifest.sgx
It will generate unique entries for sgx trusted files in manifest.sgx

Signed-off-by: Anjali Rai <anjali.rai@intel.com>

## How to test this PR? 
Python example reference has been taken to test this PR
1. Create duplicate entries for sgx.trusted section in python.manifest.template.
2. make SGX=1
3. Notice in manifest.sgx there will be unique set of entries for trusted file section

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/977)

### DIFF
--- a/python/graminelibos/gen_jinja_env.py
+++ b/python/graminelibos/gen_jinja_env.py
@@ -29,6 +29,12 @@ def ldd(*args):
             ret.add(line[0])
     return sorted(ret)
 
+def get_python_sys_paths():
+    for path in sys.path:
+        if "gramine" not in path:
+            if os.path.exists(path) and os.access(path, os.R_OK):
+                yield path
+
 def add_globals_from_python(env):
     paths = sysconfig.get_paths()
     env.globals['python'] = {
@@ -46,6 +52,7 @@ def add_globals_from_python(env):
 
         'get_path': sysconfig.get_path,
         'get_paths': sysconfig.get_paths,
+        'get_sys_path':get_python_sys_paths,
 
         'implementation': sys.implementation,
     }
--- a/python/graminelibos/manifest.py
+++ b/python/graminelibos/manifest.py
@@ -70,6 +70,32 @@ def append_trusted_dir_or_file(trusted_files, val, expanded):
         append_tf(trusted_files, uri, hash_file_contents(path))
         expanded.append(path)
 
+def get_unique_trusted_files(trusted_files):
+    """ Remove duplicates from ``sgx.trusted_files`` manifest entry and returns this new list.
+
+    Returns:
+        ``sgx.trusted_files`` list with duplicate entries removed.
+    """
+    iter_list = [trusted_files[0]]
+    unique_trusted_files = [trusted_files[0]]
+    for x in trusted_files:
+        add = True
+        temp_1 = x["uri"].replace("//", '/').replace("file:", '')
+        for tf in iter_list:
+            temp_2 = tf["uri"].replace("//", '/').replace("file:", '')
+            if temp_2.startswith(temp_1) and temp_2 != temp_1:
+                unique_trusted_files.remove(tf)
+            elif temp_1.startswith(temp_2) and temp_1 != temp_2:
+                add = False
+            elif temp_2 == temp_1:
+                add = False
+        if x not in iter_list and add:
+            unique_trusted_files.append(x)
+            iter_list = unique_trusted_files.copy()
+
+    return unique_trusted_files
+
+
 class Manifest:
     """Just a representation of a manifest.
 
@@ -169,12 +195,13 @@ class Manifest:
                 or some of them could not be loaded from the filesystem.
 
         """
-        trusted_files = []
+        final_trusted_files = []
         expanded = []
-        for tf in self['sgx']['trusted_files']:
-            append_trusted_dir_or_file(trusted_files, tf, expanded)
+        unique_trusted_files = get_unique_trusted_files(self['sgx']['trusted_files'])
+        for tf in unique_trusted_files:
+            append_trusted_dir_or_file(final_trusted_files, tf, expanded)
 
-        self['sgx']['trusted_files'] = trusted_files
+        self['sgx']['trusted_files'] = final_trusted_files
         return expanded
 
     def get_dependencies(self):


mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/977)* status: 0 of 1 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @anjalirai-intel)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/977#-NEIrV1v8vLXEzhWZ3uh):*
https://commit.style/
___
*[`-- commits` line 3 at r1](https://reviewable.io/reviews/gramineproject/gramine/977#-NEIrZqq83Z8y0TvWQ1V):*
> Gramine currently does not prevent when there are duplicate entries in manifest.sgx

What's wrong with the current behavior?
___
*[`python/graminelibos/manifest.py` line 210 at r1](https://reviewable.io/reviews/gramineproject/gramine/977#-NEIrMrSAGdC2Brvbd21:-NEIrMrSAGdC2Brvbd22:b-xlae4f) ([raw file](https://github.com/gramineproject/gramine/blob/8cec4bf91d477f7073e01f1ad3c77d786708d7f7/python/graminelibos/manifest.py#L210)):*
> ```Python
>         uri_list = list(set([t_f["uri"] for t_f in self['sgx']['trusted_files']]))
>         for uri in uri_list:
>             for t_f in self['sgx']['trusted_files']:
> ```

Isn't this O(n^2) on the number of trusted files?

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/977)* status: all files reviewed, 10 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @anjalirai-intel and @mkow)

___
*[`-- commits` line 3 at r1](https://reviewable.io/reviews/gramineproject/gramine/977#-NEIrZqq83Z8y0TvWQ1V):*

> Gramine currently does not prevent when there are duplicate entries in manifest.sgx

What's wrong with the current behavior?
</blockquote></details>

This was detected when doing something like this, see https://github.com/gramineproject/gramine/pull/978:
```
# makefile
NUMPY_PATH ?= $(shell pip3 show numpy | grep Location: | cut -d" " -f2)/
SCIPY_PATH ?= $(shell pip3 show scipy | grep Location: | cut -d" " -f2)/
SIX_PATH ?= $(shell pip3 show six | grep Location: | cut -d" " -f2)/

gramine-manifest
		-Dnumpy_path=$(NUMPY_PATH) \
		-Dscipy_path=$(SCIPY_PATH) \
		-Dsix_path=$(SIX_PATH) \

# manifest.template
sgx.trusted_files = [
  "file:{{ numpy_path }}/",
  "file:{{ scipy_path }}/",
  "file:{{ six_path }}/",
]
```

I argue that this is a reasonable way for users to write their manifests in a generic and OS-agnostic and Python-quirks-agnostic ways.

But the downside of this is that Numpy and Scipy may be installed in the same directory (you think this is what typically happens, but this is *so very much untrue* on different OS distros, and I don't even care why). And you end up having the same `sgx.trusted_files` directory with all its files listed twice -- which increases your manifest size from e.g. 2MB to 4MB. And as we know now, the TOML C parser eats up a lot of enclave memory.
___
*[`-- commits` line 4 at r1](https://reviewable.io/reviews/gramineproject/gramine/977#-NEKQrKi7SRmsHarUO25):*
The good message would be this:
```
[python] Remove duplicates in `sgx.trusted_files`

Sometimes the `.manifest` file may contain duplicate entries in the
`sgx.trusted_files` list, especially if these are directories obtained
in a programmatic way (e.g. via `pip3 show packagename | grep Location`).
Since the final `.manifest.sgx` file contains the unfolded list of files under
the directories, duplicated dir entries may significantly increase the size of such
final manifest file. To avoid this increase, we remove duplicate entries.
```
___
*[`python/graminelibos/manifest.py` line 172 at r1](https://reviewable.io/reviews/gramineproject/gramine/977#-NEKT8joCeEsDAZtNBh6:-NEKT8joCeEsDAZtNBh7:b3deqme) ([raw file](https://github.com/gramineproject/gramine/blob/8cec4bf91d477f7073e01f1ad3c77d786708d7f7/python/graminelibos/manifest.py#L172)):*
> ```Python
> 
>         """
>         trusted_files = []
> ```

Please rename to `final_trusted_files`
___
*[`python/graminelibos/manifest.py` line 174 at r1](https://reviewable.io/reviews/gramineproject/gramine/977#-NEKS8ma8zBpxk_5JN77:-NEKS8ma8zBpxk_5JN78:byj1huc) ([raw file](https://github.com/gramineproject/gramine/blob/8cec4bf91d477f7073e01f1ad3c77d786708d7f7/python/graminelibos/manifest.py#L174)):*
> ```Python
>         trusted_files = []
>         expanded = []
>         tunique_trusted_files = self.get_unique_trusted_files()
> ```

`tunique_trusted_files` -> `unique_trusted_files`
___
*[`python/graminelibos/manifest.py` line 201 at r1](https://reviewable.io/reviews/gramineproject/gramine/977#-NEKSWO8DFPfAgBPYPpS:-NEKSWO8DFPfAgBPYPpT:b-5bqi4i) ([raw file](https://github.com/gramineproject/gramine/blob/8cec4bf91d477f7073e01f1ad3c77d786708d7f7/python/graminelibos/manifest.py#L201)):*
> ```Python
>         return deps
> 
>     def get_unique_trusted_files(self):
> ```

Please move this function out of the class. And put it right after `append_trusted_dir_or_file`. (This is because methods of this class are "public API", and we don't want `Manifest::get_unique_trusted_files()` to be called by users; it is a purely private helper function.)

In this case, instead of `self` arg, you'll have `trusted_files` arg.
___
*[`python/graminelibos/manifest.py` line 202 at r1](https://reviewable.io/reviews/gramineproject/gramine/977#-NEKTT3761q_7YGcycEc:-NEKTT3761q_7YGcycEd:b-qkxns1) ([raw file](https://github.com/gramineproject/gramine/blob/8cec4bf91d477f7073e01f1ad3c77d786708d7f7/python/graminelibos/manifest.py#L202)):*
> ```Python
> 
>     def get_unique_trusted_files(self):
>         """ Generates a list of unique trusted files from sgx.trusted_files section in the manifest
> ```

Actually, I would say simply:
```
Remove duplicates from ``sgx.trusted_files`` manifest entry and returns this new list.
```
___
*[`python/graminelibos/manifest.py` line 205 at r1](https://reviewable.io/reviews/gramineproject/gramine/977#-NEKTOKP9sMqE9Z4-JN1:-NEKTOKP9sMqE9Z4-JN2:b8092hk) ([raw file](https://github.com/gramineproject/gramine/blob/8cec4bf91d477f7073e01f1ad3c77d786708d7f7/python/graminelibos/manifest.py#L205)):*
> ```Python
> 
>         Returns:
>             List of unique trusted files
> ```

I would say simply:
```
    ``sgx.trusted_files`` list with duplicate entries removed.
```
___
*[`python/graminelibos/manifest.py` line 207 at r1](https://reviewable.io/reviews/gramineproject/gramine/977#-NEKTzSU7xHn-xoxSi0T:-NEKTzSU7xHn-xoxSi0U:b-1wtwd5) ([raw file](https://github.com/gramineproject/gramine/blob/8cec4bf91d477f7073e01f1ad3c77d786708d7f7/python/graminelibos/manifest.py#L207)):*
> ```Python
>             List of unique trusted files
>         """
>         trusted_list = []
> ```

Let's rename to `unique_trusted_files`, for uniformity with the rest of the code
___
*[`python/graminelibos/manifest.py` line 210 at r1](https://reviewable.io/reviews/gramineproject/gramine/977#-NEIrMrSAGdC2Brvbd21:-NEKUS3-Dc_IwtR9oheY:bi8liep) ([raw file](https://github.com/gramineproject/gramine/blob/8cec4bf91d477f7073e01f1ad3c77d786708d7f7/python/graminelibos/manifest.py#L210)):*

Isn't this O(n^2) on the number of trusted files?
</blockquote></details>

Yes, but who cares. Recall that at this point, we're expanding from `.manifest` to `.manifest.sgx`. So the input `self['sgx']['trusted_files']` is +- written by the user, and this list contains like 10 items (typically dirs) at maximum. So the `n` is small, and we can ignore the quadratic complexity.

Also, I agree that the code snippet looks simplistic, but I'm no Python expert and cannot come up with a cleaner code to perform the same task ("get unique URIs, then for each unique URI find the first corresponding item in sgx.trusted_files (and ignore the rest items if any)").

Also, don't forget that `sgx.trusted_files` entries may have different formats in this intermediate `.manifest` file, see: https://github.com/gramineproject/gramine/blob/master/libos/test/regression/file_check_policy_strict.manifest.template#L26-L34

So what we must do is to carefully copy items *as-is* from `self['sgx']['trusted_files']`, and this code does exactly this.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/977)* status: all files reviewed, 11 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @anjalirai-intel, @boryspoplawski, and @mkow)

___
*[`-- commits` line 3 at r1](https://reviewable.io/reviews/gramineproject/gramine/977#-NEIrZqq83Z8y0TvWQ1V):*
> How's that generic and OS-agnostic? You need the very same layout on both the machine you are building the manifest and running the enclave, no way around that.

You misunderstood me. I mean, for manifest writers. The manifest writer can use the same manifest template for both Ubuntu and CentOS. Of course the development machine and the deployment machine must have the same FS layout (and thus the same OS distro).
___
*[`python/graminelibos/manifest.py` line 178 at r1](https://reviewable.io/reviews/gramineproject/gramine/977#-NELQk23CPoNVmRJ7S98:-NELXq-N9MRPjWuX6nQr:bf2nrtd) ([raw file](https://github.com/gramineproject/gramine/blob/8cec4bf91d477f7073e01f1ad3c77d786708d7f7/python/graminelibos/manifest.py#L178)):*

These `trusted_files` can still have duplicates, so what's the point of this PR?
</blockquote></details>

Why? `trusted_files` is a new list, that is initially empty and then `tf` items from `unique_trusted_files` are added to it one by one (with expansion of directories).
___
*[`python/graminelibos/manifest.py` line 210 at r1](https://reviewable.io/reviews/gramineproject/gramine/977#-NEIrMrSAGdC2Brvbd21:-NELYGnx0XFaREIzit_s:b9jxwzz) ([raw file](https://github.com/gramineproject/gramine/blob/8cec4bf91d477f7073e01f1ad3c77d786708d7f7/python/graminelibos/manifest.py#L210)):*
> So getting your python example from above, if user has /usr in manifest and NUMPY_PATH becomes /usr/lib/xxxxxx, manifest will still have duplicates.

Oh, that's a good point...

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/977)* status: all files reviewed, 11 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @anjalirai-intel, @dimakuv, and @mkow)

___
*[`python/graminelibos/manifest.py` line 178 at r1](https://reviewable.io/reviews/gramineproject/gramine/977#-NELQk23CPoNVmRJ7S98:-NELazJc3k6Nj51fYFgX:b-xw5e5s) ([raw file](https://github.com/gramineproject/gramine/blob/8cec4bf91d477f7073e01f1ad3c77d786708d7f7/python/graminelibos/manifest.py#L178)):*

Why? `trusted_files` is a new list, that is initially empty and then `tf` items from `unique_trusted_files` are added to it one by one (with expansion of directories).
</blockquote></details>

Multiple examples, one of them I shown below. e.g.: `/a/` and `/a/b/` will produce duplicates

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/977)* status: all files reviewed, 11 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @anjalirai-intel, @boryspoplawski, and @mkow)

___
*[`python/graminelibos/manifest.py` line 178 at r1](https://reviewable.io/reviews/gramineproject/gramine/977#-NELQk23CPoNVmRJ7S98:-NELempADUhX0V8Xwhs6:b-xghmb2) ([raw file](https://github.com/gramineproject/gramine/blob/8cec4bf91d477f7073e01f1ad3c77d786708d7f7/python/graminelibos/manifest.py#L178)):*

Multiple examples, one of them I shown below. e.g.: `/a/` and `/a/b/` will produce duplicates
</blockquote></details>

Yep, this PR covers only literal duplicates in `sgx.trusted_files` (like `/a/` and `/a/`, but not your example with nested dirs). We need to fix it...

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/977)* status: all files reviewed, 11 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @anjalirai-intel, @dimakuv, and @mkow)

___
*[`python/graminelibos/manifest.py` line 178 at r1](https://reviewable.io/reviews/gramineproject/gramine/977#-NELQk23CPoNVmRJ7S98:-NELgdB1AAzfNzq-lgb4:bgfdm7f) ([raw file](https://github.com/gramineproject/gramine/blob/8cec4bf91d477f7073e01f1ad3c77d786708d7f7/python/graminelibos/manifest.py#L178)):*

Yep, this PR covers only literal duplicates in `sgx.trusted_files` (like `/a/` and `/a/`, but not your example with nested dirs). We need to fix it...
</blockquote></details>

First we need to decide whether we want to do it at all, and if we do, then what to do on duplicate entries (which gets prio, etc)

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/977)* status: all files reviewed, 11 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @anjalirai-intel, @dimakuv, and @mkow)

___
*[`python/graminelibos/manifest.py` line 178 at r1](https://reviewable.io/reviews/gramineproject/gramine/977#-NELQk23CPoNVmRJ7S98:-NELo0XR2R_Gtmf93Pa0:b-c68qkz) ([raw file](https://github.com/gramineproject/gramine/blob/8cec4bf91d477f7073e01f1ad3c77d786708d7f7/python/graminelibos/manifest.py#L178)):*

> First we need to decide whether we want to do it at all

I think yes. It's definitely possible that users will end up with `sgx.trusted_files = ["/usr", "/usr"]`, and then the manifest file is 2x bigger. It's already a problem for us, these 20MB-sized manifest files...

>  what to do on duplicate entries (which gets prio, etc)

What is the difference? It's the same file, so it will have the same filename and sha256 hash. So, just any of them?
</blockquote></details>

Maybe I want to overwrite a file with some special one which I don't have on building machine? E.g. `sgx.trusted_files = [ "/data/", { "uri": "/data/special_file", "sha256": "xxxx" }]`. Then I would expect the explicitly listed file to take precedence.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/977)* status: all files reviewed, 10 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @anjalirai-intel and @mkow)

___
*[`python/graminelibos/gen_jinja_env.py` line 32 at r2](https://reviewable.io/reviews/gramineproject/gramine/977#-NEpBM8450o-fbRaDwe9:-NEpBM8450o-fbRaDweA:b-xysh8w) ([raw file](https://github.com/gramineproject/gramine/blob/beb90704b713417975ff92d95480065b385a31a8/python/graminelibos/gen_jinja_env.py#L32)):*
> ```Python
>     return sorted(ret)
> 
> def get_python_sys_paths():
> ```

Why is this change in this PR? It has nothing to do with "remove duplicates". Please move this change to another PR of yours: https://github.com/gramineproject/gramine/pull/978
___
*[`python/graminelibos/gen_jinja_env.py` line 34 at r2](https://reviewable.io/reviews/gramineproject/gramine/977#-NEpBb8F7Pstgbhsg7yF:-NEpBb8F7Pstgbhsg7yG:b-z63vmn) ([raw file](https://github.com/gramineproject/gramine/blob/beb90704b713417975ff92d95480065b385a31a8/python/graminelibos/gen_jinja_env.py#L34)):*
> ```Python
> def get_python_sys_paths():
>     for path in sys.path:
>         if "gramine" not in path:
> ```

Actually, now that I think about it, let's remove this `gramine` check. First of all, some Python system path may legitimately have `gramine` subdir. Second of all, there is no harm in listing Gramine python scripts -- who knows, maybe that's exactly what users want.
___
*[`python/graminelibos/gen_jinja_env.py` line 55 at r2](https://reviewable.io/reviews/gramineproject/gramine/977#-NEpC4F6B7dQdy_zQOwJ:-NEpC4F735vXMORyFDrb:b3ccwl5) ([raw file](https://github.com/gramineproject/gramine/blob/beb90704b713417975ff92d95480065b385a31a8/python/graminelibos/gen_jinja_env.py#L55)):*
> ```Python
>         'get_path': sysconfig.get_path,
>         'get_paths': sysconfig.get_paths,
>         'get_sys_path':get_python_sys_paths,
> ```

Should be `get_sys_paths`.

Also, add a space after `:`
___
*[`python/graminelibos/manifest.py` line 74 at r2](https://reviewable.io/reviews/gramineproject/gramine/977#-NEpC9MI-Jfobr8YsxyE:-NEpC9MI-Jfobr8YsxyF:b-jmnwcz) ([raw file](https://github.com/gramineproject/gramine/blob/beb90704b713417975ff92d95480065b385a31a8/python/graminelibos/manifest.py#L74)):*
> ```Python
> 
> def get_unique_trusted_files(trusted_files):
>     """ Remove duplicates from ``sgx.trusted_files`` manifest entry and returns this new list.
> ```

`Remove` -> `Removes`

`this` -> `a`
___
*[`python/graminelibos/manifest.py` line 94 at r2](https://reviewable.io/reviews/gramineproject/gramine/977#-NEpCRl0DesB77KarUez:-NEpCRl0DesB77KarUf-:b7nriso) ([raw file](https://github.com/gramineproject/gramine/blob/beb90704b713417975ff92d95480065b385a31a8/python/graminelibos/manifest.py#L94)):*
> ```Python
>         if x not in iter_list and add:
>             unique_trusted_files.append(x)
>             iter_list = unique_trusted_files.copy()
> ```

Oof, this loop is O(2) and extremely memory consuming. I'll (or someone else) will think of a better implementation...

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/977)* status: all files reviewed, 10 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @anjalirai-intel, @dimakuv, and @mkow)

___
*[`python/graminelibos/manifest.py` line 94 at r2](https://reviewable.io/reviews/gramineproject/gramine/977#-NEpCRl0DesB77KarUez:-NEqDlPME1bSZM-vUUts:b-unnm87) ([raw file](https://github.com/gramineproject/gramine/blob/beb90704b713417975ff92d95480065b385a31a8/python/graminelibos/manifest.py#L94)):*

Oof, this loop is O(2) and extremely memory consuming. I'll (or someone else) will think of a better implementation...
</blockquote></details>

Haven't I posted already a short and working version below? Should I just create a PR?

woju (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/977)* status: all files reviewed, 10 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @anjalirai-intel, @dimakuv, and @mkow)

___
*[`-- commits` line 3 at r1](https://reviewable.io/reviews/gramineproject/gramine/977#-NEIrZqq83Z8y0TvWQ1V):*
> I argue that this is a reasonable way for users to write their manifests

I argue it's not. It's responsibility of manifest author to ensure the paths are unique. I'm worrying it will be a footgun.

What is also wrong is, this is at best a half-solution to the problem of python paths (#978), and I argue this is a wrong solution. The correct solution is to just list all the paths from `sys.path`.

I've submitted the original branch, which I've earlier shown to @anjalirai-intel, as a separate PR: #999.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/977)* status: all files reviewed, 10 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @anjalirai-intel, @dimakuv, and @mkow)

___
*[`-- commits` line 3 at r1](https://reviewable.io/reviews/gramineproject/gramine/977#-NEIrZqq83Z8y0TvWQ1V):*

> I argue that this is a reasonable way for users to write their manifests

I argue it's not. It's responsibility of manifest author to ensure the paths are unique. I'm worrying it will be a footgun.

What is also wrong is, this is at best a half-solution to the problem of python paths (#978), and I argue this is a wrong solution. The correct solution is to just list all the paths from `sys.path`.

I've submitted the original branch, which I've earlier shown to @anjalirai-intel, as a separate PR: #999.
</blockquote></details>

But python paths can collide with non-python paths...

anjalirai-intel:
*[`-- commits` line 3 at r1](https://reviewable.io/reviews/gramineproject/gramine/977#-NEIrZqq83Z8y0TvWQ1V):*

But python paths can collide with non-python paths...
</blockquote></details>

@woju I think you are getting confused b/w the manifest changes and python centos support PR. I have already put the changes you suggested in this PR.

anjalirai-intel:
*[`-- commits` line 3 at r1](https://reviewable.io/reviews/gramineproject/gramine/977#-NEIrZqq83Z8y0TvWQ1V):*
<details><summary><i>Previously, anjalirai-intel wrote…</i></summary><blockquote>

@woju I think you are getting confused b/w the manifest changes and python centos support PR. I have already put the changes you suggested in this PR.
</blockquote></details>

I have not pushed any recent changes in this PR because @dimakuv told that he will look into the optimization to have unique trusted files.

anjalirai-intel:
*[`python/graminelibos/manifest.py` line 94 at r2](https://reviewable.io/reviews/gramineproject/gramine/977#-NEpCRl0DesB77KarUez:-NEsg30b4lyfNhhRvqNp:b-hndxpi) ([raw file](https://github.com/gramineproject/gramine/blob/beb90704b713417975ff92d95480065b385a31a8/python/graminelibos/manifest.py#L94)):*

Haven't I posted already a short and working version below? Should I just create a PR?
</blockquote></details>

With the changes you proposed, it does not handle subdirectory logic.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/977)* status: all files reviewed, 10 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @anjalirai-intel, @dimakuv, @mkow, and @woju)

___
*[`python/graminelibos/manifest.py` line 94 at r2](https://reviewable.io/reviews/gramineproject/gramine/977#-NEpCRl0DesB77KarUez:-NEv5--261EmBKwEmA5D:b-7xq55g) ([raw file](https://github.com/gramineproject/gramine/blob/beb90704b713417975ff92d95480065b385a31a8/python/graminelibos/manifest.py#L94)):*
<details><summary><i>Previously, anjalirai-intel wrote…</i></summary><blockquote>

With the changes you proposed, it does not handle subdirectory logic. 
</blockquote></details>

We are going in circles. Not to waste any more time I've prepared a simple PR implementing this: https://github.com/gramineproject/gramine/pull/1003

PR #982, opened 2022-10-17
[PAL/Linux-SGX] Replace `sgx.thread_num` with `sgx.max_threads`

dimakuv:
## Description of the changes 

Similarly, `sgx.insecure__rpc_thread_num` is replaced with `sgx.insecure__rpc_max_threads`. Note that only the manifest-option names are replaced, but the internal C code names are unmodified.

See #981 for the context and discussions.

## How to test this PR? 

CI. I modified tests/examples + I left one test using the deprecated old names.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/982)

### DIFF
--- a/CI-Examples/bash/manifest.template
+++ b/CI-Examples/bash/manifest.template
@@ -21,7 +21,7 @@ fs.mounts = [
 sgx.debug = true
 sgx.nonpie_binary = true
 sgx.enclave_size = "512M"
-sgx.thread_num = 4
+sgx.max_threads = 4
 
 sgx.trusted_files = [
   "file:{{ gramine.libos }}",
--- a/CI-Examples/blender/blender.manifest.template
+++ b/CI-Examples/blender/blender.manifest.template
@@ -24,7 +24,7 @@ sgx.debug = true
 sgx.nonpie_binary = true
 sys.stack.size = "8M"
 sgx.enclave_size = "2048M"
-sgx.thread_num = 64
+sgx.max_threads = 64
 
 sgx.trusted_files = [
   "file:{{ gramine.libos }}",
--- a/CI-Examples/lighttpd/lighttpd.manifest.template
+++ b/CI-Examples/lighttpd/lighttpd.manifest.template
@@ -22,7 +22,7 @@ fs.mounts = [
 sgx.debug = true
 sgx.nonpie_binary = true
 sgx.enclave_size = "256M"
-sgx.thread_num = 3
+sgx.max_threads = 3
 
 sgx.trusted_files = [
   "file:{{ gramine.libos }}",
--- a/CI-Examples/memcached/memcached.manifest.template
+++ b/CI-Examples/memcached/memcached.manifest.template
@@ -25,7 +25,7 @@ fs.mounts = [
 
 sgx.debug = true
 sgx.nonpie_binary = true
-sgx.thread_num = 16
+sgx.max_threads = 16
 
 # Memcached does not fail explicitly when enclave memory is exhausted. Instead, Memcached goes into
 # infinite loop without a listening socket. You can trigger this incorrect behavior by increasing
--- a/CI-Examples/nginx/nginx.manifest.template
+++ b/CI-Examples/nginx/nginx.manifest.template
@@ -28,7 +28,7 @@ fs.mounts = [
 sgx.debug = true
 sgx.nonpie_binary = true
 sgx.enclave_size = "512M"
-sgx.thread_num = 4
+sgx.max_threads = 4
 
 sgx.trusted_files = [
   "file:{{ gramine.libos }}",
--- a/CI-Examples/python/python.manifest.template
+++ b/CI-Examples/python/python.manifest.template
@@ -30,7 +30,7 @@ sys.enable_extra_runtime_domain_names_conf = true
 sgx.debug = true
 sgx.nonpie_binary = true
 sgx.enclave_size = "512M"
-sgx.thread_num = 32
+sgx.max_threads = 32
 
 sgx.remote_attestation = "{{ ra_type }}"
 sgx.ra_client_spid = "{{ ra_client_spid }}"
--- a/CI-Examples/ra-tls-mbedtls/client.manifest.template
+++ b/CI-Examples/ra-tls-mbedtls/client.manifest.template
@@ -26,7 +26,7 @@ sys.enable_extra_runtime_domain_names_conf = true
 
 sgx.debug = true
 sgx.enclave_size = "512M"
-sgx.thread_num = 4
+sgx.max_threads = 4
 
 sgx.trusted_files = [
   "file:{{ gramine.libos }}",
--- a/CI-Examples/redis/redis-server.manifest.template
+++ b/CI-Examples/redis/redis-server.manifest.template
@@ -78,7 +78,7 @@ sgx.enclave_size = "1024M"
 # and one for asynchronous events/alarms. Redis is technically single-threaded
 # but spawns couple additional threads to do background bookkeeping. Therefore,
 # specifying '8' allows to run a maximum of 6 Redis threads which is enough.
-sgx.thread_num = 8
+sgx.max_threads = 8
 
 # Redis executable is typically a PIE (Position Independent Executable) on most
 # modern OS distros (e.g., Ubuntu 18.04). However, on some OS distros (notably,
--- a/CI-Examples/rust/rust-hyper-http-server.manifest.template
+++ b/CI-Examples/rust/rust-hyper-http-server.manifest.template
@@ -40,4 +40,4 @@ sys.insecure__allow_eventfd = true
 # - any threads and threadpools you might be starting
 # - helper threads internal to Gramine — see:
 #   https://gramine.readthedocs.io/en/latest/manifest-syntax.html#number-of-threads
-sgx.thread_num = 8
+sgx.max_threads = 8
--- a/CI-Examples/rust/src/main.rs
+++ b/CI-Examples/rust/src/main.rs
@@ -11,7 +11,7 @@ async fn hello_world(_req: Request<Body>) -> Result<Response<Body>, Infallible> 
 // because you need to specify in the Gramine manifest the maximal number of threads per
 // process, and ideally this wouldn't depend on your hardware.
 //
-// See sgx.thread_num in the manifest.
+// See sgx.max_threads in the manifest.
 #[tokio::main(worker_threads = 4)]
 async fn main() {
     let addr = SocketAddr::from(([127, 0, 0, 1], 3000));
--- a/CI-Examples/sqlite/manifest.template
+++ b/CI-Examples/sqlite/manifest.template
@@ -28,7 +28,7 @@ fs.insecure__keys.default = "ffeeddccbbaa99887766554433221100"
 
 sgx.debug = true
 sgx.enclave_size = "256M"
-sgx.thread_num = 4
+sgx.max_threads = 4
 
 sgx.trusted_files = [
   "file:{{ gramine.libos }}",
--- a/Documentation/manifest-syntax.rst
+++ b/Documentation/manifest-syntax.rst
@@ -496,7 +496,7 @@ Number of threads
 
 ::
 
-    sgx.thread_num = [NUM]
+    sgx.max_threads = [NUM]
     (Default: 4)
 
 This syntax specifies the maximum number of threads that can be created inside
@@ -518,9 +518,9 @@ Note that Gramine uses several helper threads internally:
   each time a new pipe is created. It terminates itself immediately after the
   TLS handshake is performed.
 
-Given these internal threads, ``sgx.thread_num`` should be set to at least ``4``
-even for single-threaded applications (to accommodate for the main thread, the
-IPC thread, the Async thread and one TLS-handshake thread).
+Given these internal threads, ``sgx.max_threads`` should be set to at least
+``4`` even for single-threaded applications (to accommodate for the main thread,
+the IPC thread, the Async thread and one TLS-handshake thread).
 
 
 Number of RPC threads (Exitless feature)
@@ -983,3 +983,12 @@ This syntax specified how much additional memory Gramine used to reserve for its
 internal use (e.g., metadata for trusted files, internal handles,
 etc.). Currently Gramine correctly tracks all internal memory allocations and
 does not require this workaround.
+
+Number of threads (deprecated syntax)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+::
+
+    sgx.thread_num = [NUM]
+
+This name was ambiguous and was replaced with ``sgx.max_threads``.
--- a/Documentation/performance.rst
+++ b/Documentation/performance.rst
@@ -181,22 +181,22 @@ for each system call and exit the enclave. The feature can be disabled by
 specifying ``sgx.insecure__rpc_thread_num = 0``.
 
 You must decide how many untrusted helper RPC threads your application needs. A
-rule of thumb: specify ``sgx.insecure__rpc_thread_num == sgx.thread_num``, i.e.,
-the number of untrusted RPC threads should be the same as the number of enclave
-threads. For example, native Redis 6.0 uses 3-4 enclave threads during its
-execution, plus Gramine uses another 1-2 helper enclave threads. So Redis
-manifest has an over-approximation of this number: ``sgx.thread_num = 8``. Thus,
+rule of thumb: specify ``sgx.insecure__rpc_thread_num == sgx.max_threads``,
+i.e., the number of untrusted RPC threads should be the same as the number of
+enclave threads. For example, native Redis 6.0 uses 3-4 enclave threads during
+its execution, plus Gramine uses another 1-2 helper enclave threads. So Redis
+manifest has an over-approximation of this number: ``sgx.max_threads = 8``. Thus,
 to correctly enable the Exitless feature, specify
 ``sgx.insecure__rpc_thread_num = 8``. Here is an example:
 
 ::
 
-   # exitless disabled: `sgx.thread_num = 8` and `sgx.insecure__rpc_thread_num = 0`
+   # exitless disabled: `sgx.max_threads = 8` and `sgx.insecure__rpc_thread_num = 0`
    CI-Examples/redis$ gramine-sgx redis-server --save '' --protected-mode no &
    CI-Examples/redis$ src/src/redis-benchmark -t set
    43010.75 requests per second
 
-   # exitless enabled: `sgx.thread_num = 8` and `sgx.insecure__rpc_thread_num = 8`
+   # exitless enabled: `sgx.max_threads = 8` and `sgx.insecure__rpc_thread_num = 8`
    CI-Examples/redis$ gramine-sgx redis-server --save '' --protected-mode no &
    CI-Examples/redis$ src/src/redis-benchmark -t set
    68119.89 requests per second
@@ -419,8 +419,8 @@ enclave size by tweaking ``sgx.enclave_size = "512M"``,
 doesn't help, it could be due to insufficient stack size: in this case try to
 increase ``sys.stack.size = "256K"``, ``sys.stack.size = "2M"``,
 ``sys.stack.size = "4M"`` and so on. Finally, if Gramine complains about
-insufficient number of TCSs or threads, increase ``sgx.thread_num = 4``,
-``sgx.thread_num = 8``, ``sgx.thread_num = 16``, and so on.
+insufficient number of TCSs or threads, increase ``sgx.max_threads = 4``,
+``sgx.max_threads = 8``, ``sgx.max_threads = 16``, and so on.
 
 Do not forget about the cost of software encryption! Gramine transparently
 encrypts many means of communication:
--- a/common/include/toml_utils.h
+++ b/common/include/toml_utils.h
@@ -32,7 +32,7 @@ int toml_bool_in(const toml_table_t* root, const char* key, bool defaultval, boo
  * \brief Find an integer key-value in TOML manifest.
  *
  * \param root        Root table of the TOML manifest.
- * \param key         Dotted key (e.g. "sgx.thread_num").
+ * \param key         Dotted key (e.g. "sgx.max_threads").
  * \param defaultval  `retval` is set to this value if not found in the manifest.
  * \param retval      Pointer to output integer.
  *
--- a/libos/test/abi/x86_64/manifest.template
+++ b/libos/test/abi/x86_64/manifest.template
@@ -8,7 +8,7 @@ fs.mounts = [
 
 sgx.nonpie_binary = true
 sgx.debug = true
-sgx.thread_num = 4
+sgx.max_threads = 4
 
 sgx.trusted_files = [
   "file:{{ gramine.libos }}",
--- a/libos/test/abi/x86_64/stack_arg.manifest.template
+++ b/libos/test/abi/x86_64/stack_arg.manifest.template
@@ -13,7 +13,7 @@ fs.mounts = [
 
 sgx.nonpie_binary = true
 sgx.debug = true
-sgx.thread_num = 4
+sgx.max_threads = 4
 
 sgx.trusted_files = [
   "file:{{ gramine.libos }}",
--- a/libos/test/abi/x86_64/stack_env.manifest.template
+++ b/libos/test/abi/x86_64/stack_env.manifest.template
@@ -13,7 +13,7 @@ fs.mounts = [
 
 sgx.nonpie_binary = true
 sgx.debug = true
-sgx.thread_num = 4
+sgx.max_threads = 4
 
 sgx.trusted_files = [
   "file:{{ gramine.libos }}",
--- a/libos/test/fs/manifest.template
+++ b/libos/test/fs/manifest.template
@@ -22,7 +22,7 @@ fs.insecure__keys.default = "ffeeddccbbaa99887766554433221100"
 
 sgx.nonpie_binary = true
 sgx.debug = true
-sgx.thread_num = 16
+sgx.max_threads = 16
 
 sgx.allowed_files = [
   "file:tmp/",
--- a/libos/test/regression/bootstrap_cpp.manifest.template
+++ b/libos/test/regression/bootstrap_cpp.manifest.template
@@ -13,7 +13,7 @@ fs.mounts = [
   { path = "/usr/{{ arch_libdir }}", uri = "file:/usr/{{ arch_libdir }}" },
 ]
 
-sgx.thread_num = 8
+sgx.max_threads = 8
 sgx.nonpie_binary = true
 sgx.debug = true
 
--- a/libos/test/regression/manifest.template
+++ b/libos/test/regression/manifest.template
@@ -21,7 +21,7 @@ fs.mounts = [
   { type = "encrypted", path = "/encrypted_file_mrsigner.dat", uri = "file:encrypted_file_mrsigner.dat", key_name = "_sgx_mrsigner" },
 ]
 
-sgx.thread_num = 16
+sgx.max_threads = 16
 sgx.nonpie_binary = true
 sgx.debug = true
 
--- a/libos/test/regression/multi_pthread.manifest.template
+++ b/libos/test/regression/multi_pthread.manifest.template
@@ -9,7 +9,7 @@ fs.mounts = [
 ]
 
 # app runs with 4 parallel threads + Gramine has couple internal threads
-sgx.thread_num = 8
+sgx.max_threads = 8
 
 sgx.nonpie_binary = true
 sgx.debug = true
--- a/libos/test/regression/multi_pthread_exitless.manifest.template
+++ b/libos/test/regression/multi_pthread_exitless.manifest.template
@@ -9,6 +9,7 @@ fs.mounts = [
 ]
 
 # app runs with 4 parallel threads + Gramine has couple internal threads
+# TODO: legacy `sgx.thread_num` name just for testing, deprecated in v1.4, remove in v1.5
 sgx.thread_num = 8
 sgx.insecure__rpc_thread_num = 8
 
--- a/libos/test/regression/openmp.manifest.template
+++ b/libos/test/regression/openmp.manifest.template
@@ -21,7 +21,7 @@ fs.mounts = [
   { path = "/usr/{{ arch_libdir }}", uri = "file:/usr/{{ arch_libdir }}" },
 ]
 
-sgx.thread_num = 32
+sgx.max_threads = 32
 sgx.nonpie_binary = true
 sgx.debug = true
 
--- a/libos/test/regression/pthread_set_get_affinity.c
+++ b/libos/test/regression/pthread_set_get_affinity.c
@@ -19,7 +19,7 @@
 
 #define MAIN_THREAD_CNT         1
 #define INTERNAL_THREAD_CNT     2
-#define MANIFEST_SGX_THREAD_CNT 8 /* corresponds to sgx.thread_num in the manifest template */
+#define MANIFEST_SGX_THREAD_CNT 8 /* corresponds to sgx.max_threads in the manifest template */
 
 /* barrier to synchronize between parent and children */
 pthread_barrier_t barrier;
@@ -88,7 +88,7 @@ int main(int argc, const char** argv) {
         errx(EXIT_FAILURE, "Parent should have affinity set to all online cores!");
     }
 
-    /* If you want to run on all cores then increase sgx.thread_num in the manifest.template and
+    /* If you want to run on all cores then increase sgx.max_threads in the manifest.template and
      * also set MANIFEST_SGX_THREAD_CNT to the same value.
      */
     size_t numthreads = MIN(online_cores, (MANIFEST_SGX_THREAD_CNT
--- a/libos/test/regression/shebang_test_script.manifest.template
+++ b/libos/test/regression/shebang_test_script.manifest.template
@@ -10,7 +10,7 @@ fs.mounts = [
   { path = "/bin", uri = "file:/bin" },
 ]
 
-sgx.thread_num = 16
+sgx.max_threads = 16
 sgx.nonpie_binary = true
 sgx.debug = true
 
--- a/pal/regression/Thread2.manifest.template
+++ b/pal/regression/Thread2.manifest.template
@@ -1,6 +1,6 @@
 loader.entrypoint = "file:{{ binary_dir }}/{{ entrypoint }}"
 
-sgx.thread_num = 2
+sgx.max_threads = 2
 sgx.enable_stats = true
 sgx.nonpie_binary = true
 sgx.debug = true
--- a/pal/regression/Thread2_exitless.manifest.template
+++ b/pal/regression/Thread2_exitless.manifest.template
@@ -2,7 +2,7 @@
 
 loader.entrypoint = "file:{{ binary_dir }}/{{ entrypoint }}"
 
-sgx.thread_num = 2
+sgx.max_threads = 2
 sgx.insecure__rpc_thread_num = 2
 sgx.enable_stats = true
 sgx.nonpie_binary = true
--- a/pal/src/host/linux-sgx/host_main.c
+++ b/pal/src/host/linux-sgx/host_main.c
@@ -663,26 +663,41 @@ static int parse_loader_config(char* manifest, struct pal_enclave* enclave_info,
     }
 
     int64_t thread_num_int64;
-    ret = toml_int_in(manifest_root, "sgx.thread_num", /*defaultval=*/0, &thread_num_int64);
+    ret = toml_int_in(manifest_root, "sgx.max_threads", /*defaultval=*/-1, &thread_num_int64);
     if (ret < 0) {
-        log_error("Cannot parse 'sgx.thread_num'");
+        log_error("Cannot parse 'sgx.max_threads'");
         ret = -EINVAL;
         goto out;
     }
 
     if (thread_num_int64 < 0) {
-        log_error("Negative 'sgx.thread_num' is impossible");
+        /* TODO: sgx.thread_num is deprecated in v1.4, remove in v1.5 */
+        ret = toml_int_in(manifest_root, "sgx.thread_num", /*defaultval=*/-1, &thread_num_int64);
+        if (ret < 0) {
+            log_error("Cannot parse 'sgx.thread_num'");
+            ret = -EINVAL;
+            goto out;
+        }
+        if (thread_num_int64 < 0) {
+            log_error("'sgx.max_threads' not found in the manifest");
+            ret = -EINVAL;
+            goto out;
+        }
+        log_error("Detected deprecated syntax: 'sgx.thread_num'. Consider switching to "
+                  "'sgx.max_threads'.");
+    }
+
+    if (!thread_num_int64) {
+        log_error("'sgx.max_threads' must be a positive number");
         ret = -EINVAL;
         goto out;
     }
-
-    enclave_info->thread_num = thread_num_int64 ?: 1;
-
-    if (enclave_info->thread_num > MAX_DBG_THREADS) {
-        log_error("Too large 'sgx.thread_num', maximum allowed is %d", MAX_DBG_THREADS);
+    if (thread_num_int64 > MAX_DBG_THREADS) {
+        log_error("Too large 'sgx.max_threads', maximum allowed is %d", MAX_DBG_THREADS);
         ret = -EINVAL;
         goto out;
     }
+    enclave_info->thread_num = thread_num_int64;
 
     int64_t rpc_thread_num_int64;
     ret = toml_int_in(manifest_root, "sgx.insecure__rpc_thread_num", /*defaultval=*/0,
@@ -699,14 +714,13 @@ static int parse_loader_config(char* manifest, struct pal_enclave* enclave_info,
         goto out;
     }
 
-    enclave_info->rpc_thread_num = rpc_thread_num_int64;
-
-    if (enclave_info->rpc_thread_num > MAX_RPC_THREADS) {
+    if (rpc_thread_num_int64 > MAX_RPC_THREADS) {
         log_error("Too large 'sgx.insecure__rpc_thread_num', maximum allowed is %d",
                   MAX_RPC_THREADS);
         ret = -EINVAL;
         goto out;
     }
+    enclave_info->rpc_thread_num = rpc_thread_num_int64;
 
     if (enclave_info->rpc_thread_num && enclave_info->thread_num > RPC_QUEUE_SIZE) {
         log_error("Too many threads for exitless feature (more than capacity of RPC queue)");
--- a/pal/src/host/linux-sgx/host_thread.c
+++ b/pal/src/host/linux-sgx/host_thread.c
@@ -176,7 +176,7 @@ int pal_thread_init(void* tcbptr) {
     if (!tcb->tcs) {
         log_error(
             "There are no available TCS pages left for a new thread!\n"
-            "Please try to increase sgx.thread_num in the manifest.\n"
+            "Please try to increase sgx.max_threads in the manifest.\n"
             "The current value is %d",
             g_enclave_thread_num);
         ret = -ENOMEM;
--- a/python/graminelibos/manifest.py
+++ b/python/graminelibos/manifest.py
@@ -87,7 +87,11 @@ class Manifest:
         sgx = manifest.setdefault('sgx', {})
         sgx.setdefault('trusted_files', [])
         sgx.setdefault('enclave_size', DEFAULT_ENCLAVE_SIZE)
-        sgx.setdefault('thread_num', DEFAULT_THREAD_NUM)
+
+        # TODO: sgx.thread_num is deprecated in v1.4, simplify below logic in v1.5
+        if 'thread_num' not in sgx:
+            sgx.setdefault('max_threads', DEFAULT_THREAD_NUM)
+
         sgx.setdefault('isvprodid', 0)
         sgx.setdefault('isvsvn', 0)
         sgx.setdefault('remote_attestation', "none")
--- a/python/graminelibos/sgx_sign.py
+++ b/python/graminelibos/sgx_sign.py
@@ -163,17 +163,17 @@ def get_memory_areas(attr, libpal):
     areas = []
     areas.append(
         MemoryArea('ssa',
-                   size=attr['thread_num'] * offs.SSA_FRAME_SIZE * offs.SSA_FRAME_NUM,
+                   size=attr['max_threads'] * offs.SSA_FRAME_SIZE * offs.SSA_FRAME_NUM,
                    flags=PAGEINFO_R | PAGEINFO_W | PAGEINFO_REG))
-    areas.append(MemoryArea('tcs', size=attr['thread_num'] * offs.TCS_SIZE,
+    areas.append(MemoryArea('tcs', size=attr['max_threads'] * offs.TCS_SIZE,
                             flags=PAGEINFO_TCS))
-    areas.append(MemoryArea('tls', size=attr['thread_num'] * offs.PAGESIZE,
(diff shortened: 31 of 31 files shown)


mkow (changes requested):
> Reviewed 33 of 33 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/982)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`pal/src/host/linux-sgx/host_main.c` line 667 at r1](https://reviewable.io/reviews/gramineproject/gramine/982#-NEaRBaB-xBJJk25Jkq9:-NEaRBaB-xBJJk25JkqA:bm7kg6r) ([raw file](https://github.com/gramineproject/gramine/blob/f2728828f8ef641fd13b47ad6908b4ed0ba74dea/pal/src/host/linux-sgx/host_main.c#L667)):*
> ```C
>     int64_t thread_num_int64;
>     ret = toml_int_in(manifest_root, "sgx.max_threads", /*defaultval=*/0, &thread_num_int64);
>     if (ret < 0) {
> ```

I don't think it works, which means the tests you added don't actually test it :P
___
*[`pal/src/host/linux-sgx/host_main.c` line 674 at r1](https://reviewable.io/reviews/gramineproject/gramine/982#-NEaQlbvEbfY8kdumi_R:-NEaQlbvEbfY8kdumi_S:b39nudr) ([raw file](https://github.com/gramineproject/gramine/blob/f2728828f8ef641fd13b47ad6908b4ed0ba74dea/pal/src/host/linux-sgx/host_main.c#L674)):*
> ```C
>             ret = -EINVAL;
>             goto out;
>         }
> ```

there should also be a deprecation warning here
___
*[`pal/src/host/linux-sgx/host_main.c` line 695 at r1](https://reviewable.io/reviews/gramineproject/gramine/982#-NEaRQMxBMDpF7vOnEfZ:-NEaRQMxBMDpF7vOnEf_:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/f2728828f8ef641fd13b47ad6908b4ed0ba74dea/pal/src/host/linux-sgx/host_main.c#L695)):*
> ```C
>     ret = toml_int_in(manifest_root, "sgx.insecure__rpc_max_threads", /*defaultval=*/0,
>                       &rpc_thread_num_int64);
>     if (ret < 0) {
> ```

ditto
___
*[`pal/src/host/linux-sgx/host_main.c` line 704 at r1](https://reviewable.io/reviews/gramineproject/gramine/982#-NEaR-lz9L9FPgHSt0yC:-NEaR-lz9L9FPgHSt0yD:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/f2728828f8ef641fd13b47ad6908b4ed0ba74dea/pal/src/host/linux-sgx/host_main.c#L704)):*
> ```C
>             ret = -EINVAL;
>             goto out;
>         }
> ```

ditto
___
*[`pal/src/host/linux-sgx/pal_main.c` line 711 at r1](https://reviewable.io/reviews/gramineproject/gramine/982#-NEaRap_AgtDT_d6t5T2:-NEaRap_AgtDT_d6t5T3:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/f2728828f8ef641fd13b47ad6908b4ed0ba74dea/pal/src/host/linux-sgx/pal_main.c#L711)):*
> ```C
>     ret = toml_int_in(g_pal_public_state.manifest_root, "sgx.insecure__rpc_max_threads",
>                       /*defaultval=*/0, &rpc_thread_num);
>     if (ret < 0) {
> ```

ditto

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/982)* status: 31 of 33 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`pal/src/host/linux-sgx/host_main.c` line 667 at r1](https://reviewable.io/reviews/gramineproject/gramine/982#-NEaRBaB-xBJJk25Jkq9:-NEaVW9s4_ScbPo3YQFA:bslxcdv) ([raw file](https://github.com/gramineproject/gramine/blob/f2728828f8ef641fd13b47ad6908b4ed0ba74dea/pal/src/host/linux-sgx/host_main.c#L667)):*

I don't think it works, which means the tests you added don't actually test it :P
</blockquote></details>

Done.

This one is interesting. I modified the `manifest.py` script to forcefuly switch from `thread_num` -> `max_threads` (removing the former). That's why I couldn't detect this bug in the C code, even though I have one LibOS regression test with a deprecated syntax (i.e., the Python helper script removes `sgx.thread_num` so the C code doesn't even get into the `if` clause).

Now I wonder -- maybe I should just remove the check from the C code, and unconditionally use `sgx.max_threads`? Because to correctly sign the enclave, the user must run it through the `manifest.py` (which is used by `gramine-sgx-sign`) anyway.

Thus, using our Python tools, it's simply impossible to have the final `manifest.sgx` file with legacy `sgx.thread_num`...

So, should I just delete this second `toml_int_in("sgx.thread_num")`? Ditto for the RPC threads (exitless).
___
*[`pal/src/host/linux-sgx/host_main.c` line 674 at r1](https://reviewable.io/reviews/gramineproject/gramine/982#-NEaQlbvEbfY8kdumi_R:-NEaVVm_E4USk5eQGeRq:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/f2728828f8ef641fd13b47ad6908b4ed0ba74dea/pal/src/host/linux-sgx/host_main.c#L674)):*

there should also be a deprecation warning here
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/host_main.c` line 695 at r1](https://reviewable.io/reviews/gramineproject/gramine/982#-NEaRQMxBMDpF7vOnEfZ:-NEaURbp1hXYeAq4u4JB:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/f2728828f8ef641fd13b47ad6908b4ed0ba74dea/pal/src/host/linux-sgx/host_main.c#L695)):*

ditto
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/host_main.c` line 704 at r1](https://reviewable.io/reviews/gramineproject/gramine/982#-NEaR-lz9L9FPgHSt0yC:-NEaVVBX9i9thA_dJgua:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/f2728828f8ef641fd13b47ad6908b4ed0ba74dea/pal/src/host/linux-sgx/host_main.c#L704)):*

ditto
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/pal_main.c` line 711 at r1](https://reviewable.io/reviews/gramineproject/gramine/982#-NEaRap_AgtDT_d6t5T2:-NEaUOtNCwtZnuszTdNG:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/f2728828f8ef641fd13b47ad6908b4ed0ba74dea/pal/src/host/linux-sgx/pal_main.c#L711)):*

ditto
</blockquote></details>

Done.

boryspoplawski (changes requested):
> Reviewed 31 of 33 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/982)* status: all files reviewed, 11 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`libos/test/regression/multi_pthread_exitless.manifest.template` line 12 at r2](https://reviewable.io/reviews/gramineproject/gramine/982#-NEbH5dN8L_xKmgoS1h3:-NEbH5dN8L_xKmgoS1h4:bikp5z4) ([raw file](https://github.com/gramineproject/gramine/blob/b78e350059d49d806c7c73ccfd3ff755b549e340/libos/test/regression/multi_pthread_exitless.manifest.template#L12)):*
> ```
> 
> # app runs with 4 parallel threads + Gramine has couple internal threads
> # TODO: legacy names just for testing, deprecated in v1.4, remove two versions after
> ```

I probably ask it for one too many time, but didn't we decide to make this cycle 2 releases, i.e. deprecate it in the next version (here v1.4) and the remove in the following (here v1.5) ?
___
*[`pal/src/host/linux-sgx/host_main.c` line 667 at r1](https://reviewable.io/reviews/gramineproject/gramine/982#-NEaRBaB-xBJJk25Jkq9:-NEcB2ub35NDrXSHkR3g:b-r88p4) ([raw file](https://github.com/gramineproject/gramine/blob/f2728828f8ef641fd13b47ad6908b4ed0ba74dea/pal/src/host/linux-sgx/host_main.c#L667)):*

Done.

This one is interesting. I modified the `manifest.py` script to forcefuly switch from `thread_num` -> `max_threads` (removing the former). That's why I couldn't detect this bug in the C code, even though I have one LibOS regression test with a deprecated syntax (i.e., the Python helper script removes `sgx.thread_num` so the C code doesn't even get into the `if` clause).

Now I wonder -- maybe I should just remove the check from the C code, and unconditionally use `sgx.max_threads`? Because to correctly sign the enclave, the user must run it through the `manifest.py` (which is used by `gramine-sgx-sign`) anyway.

Thus, using our Python tools, it's simply impossible to have the final `manifest.sgx` file with legacy `sgx.thread_num`...

So, should I just delete this second `toml_int_in("sgx.thread_num")`? Ditto for the RPC threads (exitless).
</blockquote></details>

IMO no, we are not forcing users to use our scripts. IMO remove the `pop()` from python (see a comment there)
___
*[`pal/src/host/linux-sgx/host_main.c` line 667 at r2](https://reviewable.io/reviews/gramineproject/gramine/982#-NEcASC4ALLw49CFqO82:-NEcASC4ALLw49CFqO83:b9nw06b) ([raw file](https://github.com/gramineproject/gramine/blob/b78e350059d49d806c7c73ccfd3ff755b549e340/pal/src/host/linux-sgx/host_main.c#L667)):*
> ```C
>     int64_t thread_num_int64;
>     ret = toml_int_in(manifest_root, "sgx.max_threads", /*defaultval=*/-1, &thread_num_int64);
>     if (ret < 0 || thread_num_int64 < 0) {
> ```

Shouldn't this be 2 checks? `ret < 0` -> always fail, `thread_num_int64 < 0` -> check deprecated `sgx.thread_num`
___
*[`pal/src/host/linux-sgx/host_main.c` line 671 at r2](https://reviewable.io/reviews/gramineproject/gramine/982#-NEcArwR4iOlUIpC4Bzw:-NEcArwR4iOlUIpC4Bzx:b8xjwl4) ([raw file](https://github.com/gramineproject/gramine/blob/b78e350059d49d806c7c73ccfd3ff755b549e340/pal/src/host/linux-sgx/host_main.c#L671)):*
> ```C
>         ret = toml_int_in(manifest_root, "sgx.thread_num", /*defaultval=*/0, &thread_num_int64);
>         if (ret < 0) {
>             log_error("Cannot parse 'sgx.max_threads' (legacy name 'sgx.thread_num')");
> ```

What's the point of this `(legacy name ...)`? If you use deprecated version, you get a warning with both names...
___
*[`pal/src/host/linux-sgx/host_main.c` line 697 at r2](https://reviewable.io/reviews/gramineproject/gramine/982#-NEcAlXa-Kj4Dq4hLHk-:-NEcAlXa-Kj4Dq4hLHk0:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/b78e350059d49d806c7c73ccfd3ff755b549e340/pal/src/host/linux-sgx/host_main.c#L697)):*
> ```C
>     ret = toml_int_in(manifest_root, "sgx.insecure__rpc_max_threads", /*defaultval=*/-1,
>                       &rpc_thread_num_int64);
>     if (ret < 0 || rpc_thread_num_int64 < 0) {
> ```

ditto
___
*[`pal/src/host/linux-sgx/pal_main.c` line 711 at r2](https://reviewable.io/reviews/gramineproject/gramine/982#-NEcBCpd3ToKmV4UCOIM:-NEcBCpd3ToKmV4UCOIN:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/b78e350059d49d806c7c73ccfd3ff755b549e340/pal/src/host/linux-sgx/pal_main.c#L711)):*
> ```C
>     ret = toml_int_in(g_pal_public_state.manifest_root, "sgx.insecure__rpc_max_threads",
>                       /*defaultval=*/-1, &rpc_thread_num);
>     if (ret < 0 || rpc_thread_num < 0) {
> ```

ditto
___
*[`python/graminelibos/manifest.py` line 92 at r2](https://reviewable.io/reviews/gramineproject/gramine/982#-NEcA1J4Ds2NKWMOR8Ir:-NEcA1J4Ds2NKWMOR8Is:b-sdfovk) ([raw file](https://github.com/gramineproject/gramine/blob/b78e350059d49d806c7c73ccfd3ff755b549e340/python/graminelibos/manifest.py#L92)):*
> ```Python
>         # TODO: sgx.thread_num and sgx.insecure__rpc_thread_num are deprecated in v1.4,
>         #       simplify below logic two versions after
>         sgx.setdefault('max_threads', sgx.setdefault('thread_num', DEFAULT_THREAD_NUM))
> ```

Maybe just add a simple check here:
```python
if 'max_threads' in sgx and 'thread_num' in sgx:
    print_error_message_and_exit
```
and then set both defaults without `pop`?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/982)* status: all files reviewed, 11 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[`libos/test/regression/multi_pthread_exitless.manifest.template` line 12 at r2](https://reviewable.io/reviews/gramineproject/gramine/982#-NEbH5dN8L_xKmgoS1h3:-NEdyqLE6Xm6bupvYEO4:b-qrtmio) ([raw file](https://github.com/gramineproject/gramine/blob/b78e350059d49d806c7c73ccfd3ff755b549e340/libos/test/regression/multi_pthread_exitless.manifest.template#L12)):*

I probably ask it for one too many time, but didn't we decide to make this cycle 2 releases, i.e. deprecate it in the next version (here v1.4) and the remove in the following (here v1.5) ?
</blockquote></details>

No, from what I see, we have the following rule:
1. Deprecate in the upcoming release (v1.4).
2. Do nothing in the next release (v1.5). This gives users time to migrate.
3. Remove in the next-next release (v1.6).
___
*[`python/graminelibos/manifest.py` line 92 at r2](https://reviewable.io/reviews/gramineproject/gramine/982#-NEcA1J4Ds2NKWMOR8Ir:-NEdzqdiD07Ko_7G3ErJ:b5tueha) ([raw file](https://github.com/gramineproject/gramine/blob/b78e350059d49d806c7c73ccfd3ff755b549e340/python/graminelibos/manifest.py#L92)):*
> and then set both defaults without `pop`?

How exactly do you want to do it? If we do this:
```
sgx.setdefault('max_threads', sgx.setdefault('thread_num', DEFAULT_THREAD_NUM))
```
or this:
```
sgx.setdefault('max_threads', DEFAULT_THREAD_NUM)
sgx.setdefault('thread_num', DEFAULT_THREAD_NUM)
```
without popping (removing) `sgx.thread_num`, then we can have this problematic manifest:
```
sgx.max_threads = 8  # because it was set by the user in the template
sgx.thread_num = 4   # taken from default value because it was not set by the user in the template
```

In this case, Gramine will correctly use the "newer" `sgx.max_threads = 8` at startup, but if the user ever looks inside the `manifest.sgx` file, she will be confused by the two aliases of the same option but with different values. I wanted to prevent this behavior, that's why I added `pop` -- then the final `manifest.sgx` always has one correct `sgx.max_threads` option.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/982)* status: all files reviewed, 11 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`libos/test/regression/multi_pthread_exitless.manifest.template` line 12 at r2](https://reviewable.io/reviews/gramineproject/gramine/982#-NEbH5dN8L_xKmgoS1h3:-NEf9UjY1IFNN5QcNfip:b1511wh) ([raw file](https://github.com/gramineproject/gramine/blob/b78e350059d49d806c7c73ccfd3ff755b549e340/libos/test/regression/multi_pthread_exitless.manifest.template#L12)):*

No, from what I see, we have the following rule:
1. Deprecate in the upcoming release (v1.4).
2. Do nothing in the next release (v1.5). This gives users time to migrate.
3. Remove in the next-next release (v1.6).
</blockquote></details>

But users have time to migrate already in point one (i.e. they have one release cycle - 3-4 months - to migrate).
___
*[`python/graminelibos/manifest.py` line 92 at r2](https://reviewable.io/reviews/gramineproject/gramine/982#-NEcA1J4Ds2NKWMOR8Ir:-NEfA0lw9BzjS_2p-Hz0:bfjpwsm) ([raw file](https://github.com/gramineproject/gramine/blob/b78e350059d49d806c7c73ccfd3ff755b549e340/python/graminelibos/manifest.py#L92)):*

> and then set both defaults without `pop`?

How exactly do you want to do it? If we do this:
```
sgx.setdefault('max_threads', sgx.setdefault('thread_num', DEFAULT_THREAD_NUM))
```
or this:
```
sgx.setdefault('max_threads', DEFAULT_THREAD_NUM)
sgx.setdefault('thread_num', DEFAULT_THREAD_NUM)
```
without popping (removing) `sgx.thread_num`, then we can have this problematic manifest:
```
sgx.max_threads = 8  # because it was set by the user in the template
sgx.thread_num = 4   # taken from default value because it was not set by the user in the template
```

In this case, Gramine will correctly use the "newer" `sgx.max_threads = 8` at startup, but if the user ever looks inside the `manifest.sgx` file, she will be confused by the two aliases of the same option but with different values. I wanted to prevent this behavior, that's why I added `pop` -- then the final `manifest.sgx` always has one correct `sgx.max_threads` option.
</blockquote></details>

Look at the snipped I posed above

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/982)* status: all files reviewed, 11 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`python/graminelibos/manifest.py` line 92 at r2](https://reviewable.io/reviews/gramineproject/gramine/982#-NEcA1J4Ds2NKWMOR8Ir:-NEfA9Ms837LldjxqFFZ:bwtaqct) ([raw file](https://github.com/gramineproject/gramine/blob/b78e350059d49d806c7c73ccfd3ff755b549e340/python/graminelibos/manifest.py#L92)):*

Look at the snipped I posed above
</blockquote></details>

Ah, sorry, you meant `manifest.sgx`, not original one.
Then you can do:
```python
if 'thread_num' not in sgx:
    sgx.setdefault('max_threads', DEFAULT_THREAD_NUM)
```
i.e. only set default value for the non-deprecated version

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/982)* status: all files reviewed, 11 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, and @woju)

___
*[`libos/test/regression/multi_pthread_exitless.manifest.template` line 12 at r2](https://reviewable.io/reviews/gramineproject/gramine/982#-NEbH5dN8L_xKmgoS1h3:-NEfHGc212VfqiE2GIPw:bxha1to) ([raw file](https://github.com/gramineproject/gramine/blob/b78e350059d49d806c7c73ccfd3ff755b549e340/libos/test/regression/multi_pthread_exitless.manifest.template#L12)):*

But users have time to migrate already in point one (i.e. they have one release cycle - 3-4 months - to migrate).
</blockquote></details>

@mkow @woju 
___
*[`python/graminelibos/manifest.py` line 92 at r2](https://reviewable.io/reviews/gramineproject/gramine/982#-NEcA1J4Ds2NKWMOR8Ir:-NEfHL7g6m6H3bWzM0uT:b551868) ([raw file](https://github.com/gramineproject/gramine/blob/b78e350059d49d806c7c73ccfd3ff755b549e340/python/graminelibos/manifest.py#L92)):*

Ah, sorry, you meant `manifest.sgx`, not original one.
Then you can do:
```python
if 'thread_num' not in sgx:
    sgx.setdefault('max_threads', DEFAULT_THREAD_NUM)
```
i.e. only set default value for the non-deprecated version
</blockquote></details>

This will pose a problem because in this case the manifest file we'll have `sgx.thread_num` but won't have `sgx.max_threads`, and the `sgx_sign.py` will fail because it only expects `sgx.max_threads` (see this file in this PR).

Do you want me to still do your change + modify `sgx_sign.py` such that this file does smth like:
```
attr.get('max_threads', attr['thread_num'])
```
?

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/982)* status: all files reviewed, 10 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @woju)

___
*[`libos/test/regression/multi_pthread_exitless.manifest.template` line 12 at r2](https://reviewable.io/reviews/gramineproject/gramine/982#-NEbH5dN8L_xKmgoS1h3:-NEfJd3V5fXMNWHAy4k2:bqb09uw) ([raw file](https://github.com/gramineproject/gramine/blob/b78e350059d49d806c7c73ccfd3ff755b549e340/libos/test/regression/multi_pthread_exitless.manifest.template#L12)):*

@mkow @woju 
</blockquote></details>

I think it was rather as Borys said, but I'm not 100% sure.
___
*[`pal/src/host/linux-sgx/host_main.c` line 667 at r1](https://reviewable.io/reviews/gramineproject/gramine/982#-NEaRBaB-xBJJk25Jkq9:-NEfK1uYAmxTivUx1g_r:bz132hv) ([raw file](https://github.com/gramineproject/gramine/blob/f2728828f8ef641fd13b47ad6908b4ed0ba74dea/pal/src/host/linux-sgx/host_main.c#L667)):*

IMO no, we are not forcing users to use our scripts. IMO remove the `pop()` from python (see a comment there)
</blockquote></details>

In general I like the idea of handling deprecated stuff externally and simplify our trusted C code. But here, as Borys said, the problem is that not everyone may be using our templating scripts and only use the signer.
But I'd gladly hear if you have a good idea how to redesign the flow to allow handling deprecation outside of C code.
___
*[`pal/src/host/linux-sgx/host_main.c` line 674 at r1](https://reviewable.io/reviews/gramineproject/gramine/982#-NEaQlbvEbfY8kdumi_R:-NEfL5UxEEedwVxLcMGn:bd2h69h) ([raw file](https://github.com/gramineproject/gramine/blob/f2728828f8ef641fd13b47ad6908b4ed0ba74dea/pal/src/host/linux-sgx/host_main.c#L674)):*

Done.
</blockquote></details>

But now you print it also when none of the two options was given in the manifest.
___
*[`pal/src/host/linux-sgx/host_main.c` line 704 at r1](https://reviewable.io/reviews/gramineproject/gramine/982#-NEaR-lz9L9FPgHSt0yC:-NEfLDTQCkyVMHY-Ab2e:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/f2728828f8ef641fd13b47ad6908b4ed0ba74dea/pal/src/host/linux-sgx/host_main.c#L704)):*

Done.
</blockquote></details>

ditto
___
*[`pal/src/host/linux-sgx/pal_main.c` line 721 at r2](https://reviewable.io/reviews/gramineproject/gramine/982#-NEfLFts7hTl59ZC6CIJ:-NEfLFts7hTl59ZC6CIK:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/b78e350059d49d806c7c73ccfd3ff755b549e340/pal/src/host/linux-sgx/pal_main.c#L721)):*
> ```C
>         }
>         log_warning("Detected deprecated syntax: 'sgx.insecure__rpc_thread_num'. Consider "
>                     "switching to 'sgx.insecure__rpc_max_threads'.");
> ```

ditto

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/982)* status: all files reviewed, 10 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @woju)

___
*[`libos/test/regression/multi_pthread_exitless.manifest.template` line 12 at r2](https://reviewable.io/reviews/gramineproject/gramine/982#-NEbH5dN8L_xKmgoS1h3:-NEfYIDg9cTHNwBKRDXR:bhjhzk9) ([raw file](https://github.com/gramineproject/gramine/blob/b78e350059d49d806c7c73ccfd3ff755b549e340/libos/test/regression/multi_pthread_exitless.manifest.template#L12)):*

I think it was rather as Borys said, but I'm not 100% sure.
</blockquote></details>

We definitely write everywhere (i.e. in code where we have such TODOs) that it's after two versions, but I also think we talked about it at some point and the conclusion was: next version deprecates, next-next removes, but not sure
___
*[`python/graminelibos/manifest.py` line 92 at r2](https://reviewable.io/reviews/gramineproject/gramine/982#-NEcA1J4Ds2NKWMOR8Ir:-NEfYtUkEyrNq-FQEo4Y:b6f1yfa) ([raw file](https://github.com/gramineproject/gramine/blob/b78e350059d49d806c7c73ccfd3ff755b549e340/python/graminelibos/manifest.py#L92)):*

This will pose a problem because in this case the manifest file we'll have `sgx.thread_num` but won't have `sgx.max_threads`, and the `sgx_sign.py` will fail because it only expects `sgx.max_threads` (see this file in this PR).

Do you want me to still do your change + modify `sgx_sign.py` such that this file does smth like:
```
attr.get('max_threads', attr['thread_num'])
```
?
</blockquote></details>

Idk, I trust you can come up with a version that works and makes sense :) But I guess there are 2 options (?) - add support for both in `sgx_sign.py` or always force users to use `max_threads` - both are not ideal.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/982)* status: 28 of 33 files reviewed, 10 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, and @woju)

___
*[`libos/test/regression/multi_pthread_exitless.manifest.template` line 12 at r2](https://reviewable.io/reviews/gramineproject/gramine/982#-NEbH5dN8L_xKmgoS1h3:-NEfqR3S-J7vQkyXPSWn:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/b78e350059d49d806c7c73ccfd3ff755b549e340/libos/test/regression/multi_pthread_exitless.manifest.template#L12)):*

We definitely write everywhere (i.e. in code where we have such TODOs) that it's after two versions, but I also think we talked about it at some point and the conclusion was: next version deprecates, next-next removes, but not sure
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/host_main.c` line 667 at r1](https://reviewable.io/reviews/gramineproject/gramine/982#-NEaRBaB-xBJJk25Jkq9:-NEgGFbBC5NZ0BvyaCpr:bcdfqu1) ([raw file](https://github.com/gramineproject/gramine/blob/f2728828f8ef641fd13b47ad6908b4ed0ba74dea/pal/src/host/linux-sgx/host_main.c#L667)):*

In general I like the idea of handling deprecated stuff externally and simplify our trusted C code. But here, as Borys said, the problem is that not everyone may be using our templating scripts and only use the signer.
But I'd gladly hear if you have a good idea how to redesign the flow to allow handling deprecation outside of C code.
</blockquote></details>

Ok, did as you both wanted.
___
*[`pal/src/host/linux-sgx/host_main.c` line 674 at r1](https://reviewable.io/reviews/gramineproject/gramine/982#-NEaQlbvEbfY8kdumi_R:-NEgCDh877J_ybEWr4lu:b-pse7px) ([raw file](https://github.com/gramineproject/gramine/blob/f2728828f8ef641fd13b47ad6908b4ed0ba74dea/pal/src/host/linux-sgx/host_main.c#L674)):*

But now you print it also when none of the two options was given in the manifest.
</blockquote></details>

We actually didn't have any logic for "the option is not given in the manifest" before... I fixed it.
___
*[`pal/src/host/linux-sgx/host_main.c` line 704 at r1](https://reviewable.io/reviews/gramineproject/gramine/982#-NEaR-lz9L9FPgHSt0yC:-NEgDHszDlQEQJM3h2E5:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/f2728828f8ef641fd13b47ad6908b4ed0ba74dea/pal/src/host/linux-sgx/host_main.c#L704)):*

ditto
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/host_main.c` line 667 at r2](https://reviewable.io/reviews/gramineproject/gramine/982#-NEcASC4ALLw49CFqO82:-NEgAjwMAj52GaXAoN5G:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/b78e350059d49d806c7c73ccfd3ff755b549e340/pal/src/host/linux-sgx/host_main.c#L667)):*

Shouldn't this be 2 checks? `ret < 0` -> always fail, `thread_num_int64 < 0` -> check deprecated `sgx.thread_num`
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/host_main.c` line 671 at r2](https://reviewable.io/reviews/gramineproject/gramine/982#-NEcArwR4iOlUIpC4Bzw:-NEgAhy-53lC57XJDWHi:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/b78e350059d49d806c7c73ccfd3ff755b549e340/pal/src/host/linux-sgx/host_main.c#L671)):*

What's the point of this `(legacy name ...)`? If you use deprecated version, you get a warning with both names...
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/host_main.c` line 697 at r2](https://reviewable.io/reviews/gramineproject/gramine/982#-NEcAlXa-Kj4Dq4hLHk-:-NEgDGxj-o2-INg-2kPM:bb6un1j) ([raw file](https://github.com/gramineproject/gramine/blob/b78e350059d49d806c7c73ccfd3ff755b549e340/pal/src/host/linux-sgx/host_main.c#L697)):*

ditto
</blockquote></details>

Done
___
*[`pal/src/host/linux-sgx/host_main.c` line 689 at r3](https://reviewable.io/reviews/gramineproject/gramine/982#-NEgIi9D77XoVpwhq5AM:-NEgIi9D77XoVpwhq5AN:bai7zq4) ([raw file](https://github.com/gramineproject/gramine/blob/83cf079288ee3743b96cbec1b82c156d62524e3a/pal/src/host/linux-sgx/host_main.c#L689)):*
> ```C
>         thread_num_option = "sgx.thread_num";
>         log_error("Detected deprecated syntax: 'sgx.thread_num'. Consider switching to "
>                   "'sgx.max_threads'.");
> ```

FYI: Note that this is now `log_error()`. Previously it was `log_warning()`, but this is printed even before we parse and set the log level, so it was always skipped (because default log level is `error`). So I changed to error log level to make sure it definitely prints, and verified via the `multithread_exitless` LibOS test.
___
*[`pal/src/host/linux-sgx/pal_main.c` line 711 at r2](https://reviewable.io/reviews/gramineproject/gramine/982#-NEcBCpd3ToKmV4UCOIM:-NEgDeqA8TAc5Zr8M-Kd:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/b78e350059d49d806c7c73ccfd3ff755b549e340/pal/src/host/linux-sgx/pal_main.c#L711)):*

ditto
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/pal_main.c` line 721 at r2](https://reviewable.io/reviews/gramineproject/gramine/982#-NEfLFts7hTl59ZC6CIJ:-NEgDw9wE3wiB48mh__W:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/b78e350059d49d806c7c73ccfd3ff755b549e340/pal/src/host/linux-sgx/pal_main.c#L721)):*

ditto
</blockquote></details>

Done.
___
*[`python/graminelibos/manifest.py` line 92 at r2](https://reviewable.io/reviews/gramineproject/gramine/982#-NEcA1J4Ds2NKWMOR8Ir:-NEfqyG_3CqQ3ZrH_KMr:b-dvall) ([raw file](https://github.com/gramineproject/gramine/blob/b78e350059d49d806c7c73ccfd3ff755b549e340/python/graminelibos/manifest.py#L92)):*

Idk, I trust you can come up with a version that works and makes sense :) But I guess there are 2 options (?) - add support for both in `sgx_sign.py` or always force users to use `max_threads` - both are not ideal. 
</blockquote></details>

Done. Ok, going with the first option, as you and Michal both requested smth like this.

boryspoplawski (changes requested):
> Reviewed 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/982)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`pal/src/host/linux-sgx/host_main.c` line 674 at r1](https://reviewable.io/reviews/gramineproject/gramine/982#-NEaQlbvEbfY8kdumi_R:-NEgOZXA7xamrZZoOTbm:b-nqnpd7) ([raw file](https://github.com/gramineproject/gramine/blob/f2728828f8ef641fd13b47ad6908b4ed0ba74dea/pal/src/host/linux-sgx/host_main.c#L674)):*

We actually didn't have any logic for "the option is not given in the manifest" before... I fixed it.
</blockquote></details>

We had logic of defaulting to `1`. Now you changed it to fail if no option given, but change `0` -> `1`. Is that what we want? I would rather fail on `0`.
___
*[`pal/src/host/linux-sgx/host_main.c` line 671 at r2](https://reviewable.io/reviews/gramineproject/gramine/982#-NEcArwR4iOlUIpC4Bzw:-NEgNB3q1WgzeQ8y7Lqq:b-alfjlu) ([raw file](https://github.com/gramineproject/gramine/blob/b78e350059d49d806c7c73ccfd3ff755b549e340/pal/src/host/linux-sgx/host_main.c#L671)):*

Done.
</blockquote></details>

I meant why not always print the current name? If you use legacy, you get a warning about switching to the new version (so you know both names and seeing the new one won't be a surprise).
___
*[`pal/src/host/linux-sgx/host_main.c` line 719 at r3](https://reviewable.io/reviews/gramineproject/gramine/982#-NEgOq_GBn8RLlJMJ66I:-NEgOq_GBn8RLlJMJ66J:b-viuke8) ([raw file](https://github.com/gramineproject/gramine/blob/83cf079288ee3743b96cbec1b82c156d62524e3a/pal/src/host/linux-sgx/host_main.c#L719)):*
> ```C
>             goto out;
>         }
>         if (rpc_thread_num_int64 < 0) {
> ```

How does it work? Each manifest must have this option now and if it doesn't want this feature it must specify `0`?
___
*[`pal/src/host/linux-sgx/host_main.c` line 732 at r3](https://reviewable.io/reviews/gramineproject/gramine/982#-NEgOxJb-yMlN86nquJR:-NEgOxJb-yMlN86nquJS:b-9n5pp2) ([raw file](https://github.com/gramineproject/gramine/blob/83cf079288ee3743b96cbec1b82c156d62524e3a/pal/src/host/linux-sgx/host_main.c#L732)):*
> ```C
> 
>     if (enclave_info->rpc_thread_num > MAX_RPC_THREADS) {
>         log_error("Too large '%s', maximum allowed is %d", rpc_thread_num_option, MAX_RPC_THREADS);
> ```

ditto (IMO just print the new name)
___
*[`pal/src/host/linux-sgx/pal_main.c` line 724 at r3](https://reviewable.io/reviews/gramineproject/gramine/982#-NEgP9_B5q8B_ivg-9BN:-NEgP9_B5q8B_ivg-9BO:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/83cf079288ee3743b96cbec1b82c156d62524e3a/pal/src/host/linux-sgx/pal_main.c#L724)):*
> ```C
>         }
>         if (rpc_thread_num < 0) {
>             log_error("'sgx.insecure__rpc_max_threads' not found in the manifest");
> ```

ditto

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/982)* status: 31 of 33 files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[`pal/src/host/linux-sgx/host_main.c` line 674 at r1](https://reviewable.io/reviews/gramineproject/gramine/982#-NEaQlbvEbfY8kdumi_R:-NEj-J0J2rM_MqI3g8je:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/f2728828f8ef641fd13b47ad6908b4ed0ba74dea/pal/src/host/linux-sgx/host_main.c#L674)):*

We had logic of defaulting to `1`. Now you changed it to fail if no option given, but change `0` -> `1`. Is that what we want? I would rather fail on `0`.
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/host_main.c` line 671 at r2](https://reviewable.io/reviews/gramineproject/gramine/982#-NEcArwR4iOlUIpC4Bzw:-NEizj1cEUSsn4N1bucz:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/b78e350059d49d806c7c73ccfd3ff755b549e340/pal/src/host/linux-sgx/host_main.c#L671)):*

I meant why not always print the current name? If you use legacy, you get a warning about switching to the new version (so you know both names and seeing the new one won't be a surprise).
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/host_main.c` line 719 at r3](https://reviewable.io/reviews/gramineproject/gramine/982#-NEgOq_GBn8RLlJMJ66I:-NEj-itj7mZIVaxQIIvL:b-6ekkvr) ([raw file](https://github.com/gramineproject/gramine/blob/83cf079288ee3743b96cbec1b82c156d62524e3a/pal/src/host/linux-sgx/host_main.c#L719)):*

How does it work? Each manifest must have this option now and if it doesn't want this feature it must specify `0`?
</blockquote></details>

Done. Now explicitly set to zero if no option was provided.

Previously it worked because our Python scripts always specify this option (if not specified by user, then the Python script sets it to zero). So that's why I didn't notice this bug.
___
*[`pal/src/host/linux-sgx/host_main.c` line 732 at r3](https://reviewable.io/reviews/gramineproject/gramine/982#-NEgOxJb-yMlN86nquJR:-NEj-Lo-BAfObDALYQDK:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/83cf079288ee3743b96cbec1b82c156d62524e3a/pal/src/host/linux-sgx/host_main.c#L732)):*

ditto (IMO just print the new name)
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/pal_main.c` line 724 at r3](https://reviewable.io/reviews/gramineproject/gramine/982#-NEgP9_B5q8B_ivg-9BN:-NEj0HiR52spS_qxr9iI:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/83cf079288ee3743b96cbec1b82c156d62524e3a/pal/src/host/linux-sgx/pal_main.c#L724)):*

ditto
</blockquote></details>

Done.

boryspoplawski (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/982)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow)

mkow (changes requested):
> Reviewed 1 of 5 files at r3, 1 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/982)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`pal/src/host/linux-sgx/host_main.c` line 725 at r4](https://reviewable.io/reviews/gramineproject/gramine/982#-NFE9DVk0IlERTGOMsSd:-NFE9DVk0IlERTGOMsSe:b-ohhox6) ([raw file](https://github.com/gramineproject/gramine/blob/a84f86d03f5b7ade9c411ea60b3756613bc918e1/pal/src/host/linux-sgx/host_main.c#L725)):*
> ```C
>         }
>         log_error("Detected deprecated syntax: 'sgx.insecure__rpc_thread_num'. Consider "
>                   "switching to 'sgx.insecure__rpc_max_threads'.");
> ```

Actually... isn't "num" the correct term in this case? For exitless we just spawn all these threads, it's not their maximal number, but their exact number?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/982)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`pal/src/host/linux-sgx/host_main.c` line 725 at r4](https://reviewable.io/reviews/gramineproject/gramine/982#-NFE9DVk0IlERTGOMsSd:-NFEE2z--V1ZaN-TJMhk:b-nfr37r) ([raw file](https://github.com/gramineproject/gramine/blob/a84f86d03f5b7ade9c411ea60b3756613bc918e1/pal/src/host/linux-sgx/host_main.c#L725)):*

Actually... isn't "num" the correct term in this case? For exitless we just spawn all these threads, it's not their maximal number, but their exact number?
</blockquote></details>

Hm. Well, yes, but these threads do not spin all the time, they sleep periodically (when there's no new requests): https://github.com/gramineproject/gramine/blob/f512b33303548a4cb0c6c8117cfe8ea5cb7999ff/pal/src/host/linux-sgx/host_ocalls.c#L815

I'm not sure we want to keep the old name, because then we'll have a discrepancy in naming (`max_threads` vs `rpc_thread_num`). Also, maybe we'll change Exitless to spawn new RPC threads only when needed -- then `rpc_max_threads` will be a maximal number.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/982)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`pal/src/host/linux-sgx/host_main.c` line 725 at r4](https://reviewable.io/reviews/gramineproject/gramine/982#-NFE9DVk0IlERTGOMsSd:-NFEpCzV7b1-aZvQu4Bd:b2asbbz) ([raw file](https://github.com/gramineproject/gramine/blob/a84f86d03f5b7ade9c411ea60b3756613bc918e1/pal/src/host/linux-sgx/host_main.c#L725)):*
Ah, true, missed that.

> I'm not sure we want to keep the old name, because then we'll have a discrepancy in naming (max_threads vs rpc_thread_num). Also, maybe we'll change Exitless to spawn new RPC threads only when needed -- then rpc_max_threads will be a maximal number.


    then we'll have a discrepancy in naming

But the discrepancy isn't undue, in RPC we spawn each thread on the host upfront, in `max_threads` case we spawn them on demand.

>> then we'll have a discrepancy in naming
>>
> That's intended IMO, because both options have different semantics (one is max, the other is precise number).

Yes, exactly.

So I think the old name can stay (i.e. `max`)

boryspoplawski (dismissed):
> Reviewed 7 of 8 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/982)* status: all files reviewed, 3 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`pal/src/host/linux-sgx/host_main.c` line 723 at r5](https://reviewable.io/reviews/gramineproject/gramine/982#-NFF45gX5gKhiqgmNpLO:-NFG-LXD46J-uX909Yek:b-f01hev) ([raw file](https://github.com/gramineproject/gramine/blob/26c7283ad0ad5576f58e870fb3dc51a72330c90b/pal/src/host/linux-sgx/host_main.c#L723)):*

FYI: Just switched these two statements places. I think it's more readable.
</blockquote></details>

hmm? I don't see any changes here

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/982)* status: all files reviewed, 3 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`pal/src/host/linux-sgx/host_main.c` line 723 at r5](https://reviewable.io/reviews/gramineproject/gramine/982#-NFF45gX5gKhiqgmNpLO:-NFI9Orl-ugzRGEXJedg:b-dukh1e) ([raw file](https://github.com/gramineproject/gramine/blob/26c7283ad0ad5576f58e870fb3dc51a72330c90b/pal/src/host/linux-sgx/host_main.c#L723)):*

hmm? I don't see any changes here
</blockquote></details>

I mean, I reverted all changes to `rpc_thread_num` in this PR, except this particular change (where I swapped the two statements places). You can see this change when you look at the full range of commits in this file.

mkow (approved):
> Reviewed 1 of 5 files at r3, 7 of 8 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/982)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`Documentation/manifest-syntax.rst` line 1000 at r6](https://reviewable.io/reviews/gramineproject/gramine/982#-NFR52UL8-4RBbVv2Q5G:-NFR52UL8-4RBbVv2Q5H:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/2e86548d21818e39ff30391480c33045221143ae/Documentation/manifest-syntax.rst#L1000)):*
> ```reStructuredText
>     sgx.thread_num = [NUM]
> 
> This name was ambigious and was replaced with ``sgx.max_threads``.
> ```





_Suggestion:_
```reStructuredText
ambiguous
```

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/982)* status: 27 of 33 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski and @mkow)

___
*[`-- commits` line 5 at r4](https://reviewable.io/reviews/gramineproject/gramine/982#-NFF-UJu9eqNQZXDisml):*

I will remove the first sentence, because we decided to revert the change with RPC (Exitless) threads.
</blockquote></details>

Done
___
*[`Documentation/manifest-syntax.rst` line 1000 at r6](https://reviewable.io/reviews/gramineproject/gramine/982#-NFR52UL8-4RBbVv2Q5G:-NFSVDH5-RRB6pzQM5ZI:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/2e86548d21818e39ff30391480c33045221143ae/Documentation/manifest-syntax.rst#L1000)):*
> ```reStructuredText
>     sgx.thread_num = [NUM]
> 
> This name was ambigious and was replaced with ``sgx.max_threads``.
> ```

Done.

mkow (approved):
> Reviewed 6 of 6 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/982)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #983, opened 2022-10-17
[LibOS] Add `sys.disallow_subprocesses` manifest option

dimakuv:
## Description of the changes 

Some apps (in particular Python) tend to spawn helper subprocesses. There is no simple way to disable such functionality in the apps themselves, so we introduce a new manifest option to force apps to use only the main process and fall back to no-subprocesses logic.

~~**This is a draft for quick testing! After we'll test that it helps sometimes, we can have discussions on the terribleness of this approach.**~~ Based on customer feedback, this looks like useful feature, so I mark this PR as ready-for-review.

This PR literally disables `fork/vfork/clone` syscalls used to spawn subprocesses.

## How to test this PR? 

Tested some Python workloads; they stopped spawning child enclaves and still behave correctly.

Here is an example of Flask:
- Install Flask: `pip3 install Flask` (installs in HOME dir of user)
- Improve manifest file:
```diff
diff --git a/CI-Examples/python/python.manifest.template b/CI-Examples/python/python.manifest.template
index da1a904e..9ba70b14 100644
@@ -8,2 +8,5 @@ loader.log_level = "{{ log_level }}"
 loader.env.LD_LIBRARY_PATH = "/lib:/lib:{{ arch_libdir }}:/usr/{{ arch_libdir }}"
+loader.env.HOME = { passthrough = true }  # so that Python knows where to search for home-installed modules
+
+sys.disallow_subprocesses = true

@@ -27,2 +30,3 @@ fs.mounts = [
   { path = "/etc/hosts", uri = "file:helper-files/hosts" },
+  { path = "/bin/uname", uri = "file:/bin/uname" },  # this is the binary that is fork+execve'd

@@ -55,2 +59,3 @@ sgx.trusted_files = [
   "file:helper-files/",
+  "file:/bin/uname",
 ]
```
- Improve Helloworld Python script:
```diff
diff --git a/CI-Examples/python/scripts/helloworld.py b/CI-Examples/python/scripts/helloworld.py
index b11fb8ab..24f90b94 100644
@@ -2,2 +2,4 @@

+from flask import Flask
+
 print("Hello World")
```
- Run with `sys.disallow_subprocesses = true` commented out (old behavior):
```sh
~/gramineproject/gramine/CI-Examples/python$ time gramine-sgx python scripts/helloworld.py
Hello World

real    0m34.935s
user    0m9.781s
sys     0m8.668s
```
- Run with `sys.disallow_subprocesses = true`  (new behavior):
```sh
~/gramineproject/gramine/CI-Examples/python$ time gramine-sgx python  scripts/helloworld.py
Hello World

real    0m17.889s
user    0m9.462s
sys     0m8.390s
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/983)

### DIFF
--- a/Documentation/manifest-syntax.rst
+++ b/Documentation/manifest-syntax.rst
@@ -337,6 +337,27 @@ Be careful! In SGX environment, the untrusted host could inject that signal in
 an arbitrary moment. Examine what your application's `SIGTERM` handler does and
 whether it poses any security threat.
 
+Disallowing subprocesses (fork)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+::
+
+    sys.disallow_subprocesses = [true|false]
+    (Default: false)
+
+This specifies whether to block applications from creating child processes (e.g.
+via ``fork()`` or ``clone()`` system calls). The intuition is that many
+applications have fallbacks when they fail to spawn a child process (e.g.
+Python). Could be useful in SGX environments: child processes consume
+:term:`EPC` memory which is a limited resource.
+
+.. note ::
+   This option is *not* a security feature - Gramine by-design is only a one-way
+   sandbox, which doesn't protect the host from the enclave. Don't use this
+   option if you want to somehow mitigate running untrusted enclaves. Instead,
+   to achieve this, you need to run the whole Gramine inside a proper security
+   sandbox.
+
 Root FS mount point
 ^^^^^^^^^^^^^^^^^^^
 
--- a/libos/src/sys/libos_clone.c
+++ b/libos/src/sys/libos_clone.c
@@ -17,6 +17,7 @@
 #include "libos_types.h"
 #include "libos_vma.h"
 #include "pal.h"
+#include "toml_utils.h"
 
 struct libos_clone_args {
     PAL_HANDLE create_event;
@@ -306,6 +307,25 @@ long libos_syscall_clone(unsigned long flags, unsigned long user_stack_addr, int
     }
 
     long ret = 0;
+
+    if (clone_new_process) {
+        assert(g_manifest_root);
+        bool disallow_subprocesses;
+        ret = toml_bool_in(g_manifest_root, "sys.disallow_subprocesses", /*defaultval=*/false,
+                           &disallow_subprocesses);
+        if (ret < 0) {
+            log_error("Cannot parse \'sys.disallow_subprocesses\' (the value must be `true` or "
+                      "`false`)");
+            return -ENOSYS;
+        }
+        if (disallow_subprocesses) {
+            if (FIRST_TIME()) {
+                log_warning("The app tried to create a subprocess, but this is disabled "
+                            "(sys.disallow_subprocesses = true)");
+            }
+            return -EAGAIN;
+        }
+    }
 
     struct libos_thread* thread = get_new_thread();
     if (!thread) {
--- a/libos/test/regression/fork_disallowed.manifest.template
+++ b/libos/test/regression/fork_disallowed.manifest.template
@@ -1,0 +1,25 @@
+{% set entrypoint = "fork_and_exec" -%}
+
+loader.entrypoint = "file:{{ gramine.libos }}"
+libos.entrypoint = "{{ entrypoint }}"
+
+loader.log_level = "warning"
+loader.env.LD_LIBRARY_PATH = "/lib"
+
+fs.mounts = [
+  { path = "/lib", uri = "file:{{ gramine.runtimedir(libc) }}" },
+  { path = "/{{ entrypoint }}", uri = "file:{{ binary_dir }}/{{ entrypoint }}" },
+]
+
+# must print a warning: "The app tried to create a subprocess, but this is disabled"
+sys.disallow_subprocesses = true
+
+sgx.nonpie_binary = true
+sgx.debug = true
+sgx.edmm_enable = {{ 'true' if env.get('EDMM', '0') == '1' else 'false' }}
+
+sgx.trusted_files = [
+  "file:{{ gramine.libos }}",
+  "file:{{ gramine.runtimedir(libc) }}/",
+  "file:{{ binary_dir }}/{{ entrypoint }}",
+]
--- a/libos/test/regression/test_libos.py
+++ b/libos/test/regression/test_libos.py
@@ -206,21 +206,29 @@ class TC_01_Bootstrap(RegressionTestCase):
         self.assertIn('child exited with status: 0', stdout)
         self.assertIn('test completed successfully', stdout)
 
-    def test_203_vfork_and_exec(self):
+    def test_203_fork_disallowed(self):
+        try:
+            self.run_binary(['fork_disallowed'])
+            self.fail('expected to return nonzero')
+        except subprocess.CalledProcessError as e:
+            stderr = e.stderr.decode()
+            self.assertIn('The app tried to create a subprocess, but this is disabled', stderr)
+
+    def test_204_vfork_and_exec(self):
         stdout, _ = self.run_binary(['vfork_and_exec'], timeout=60)
 
         # vfork and exec 2 page child binary
         self.assertIn('child exited with status: 0', stdout)
         self.assertIn('test completed successfully', stdout)
 
-    def test_204_exec_fork(self):
+    def test_205_exec_fork(self):
         stdout, _ = self.run_binary(['exec_fork'], timeout=60)
         self.assertNotIn('Handled SIGCHLD', stdout)
         self.assertIn('Set up handler for SIGCHLD', stdout)
         self.assertIn('child exited with status: 0', stdout)
         self.assertIn('test completed successfully', stdout)
 
-    def test_205_double_fork(self):
+    def test_206_double_fork(self):
         stdout, stderr = self.run_binary(['double_fork'])
         self.assertIn('TEST OK', stdout)
         self.assertNotIn('grandchild', stderr)
--- a/libos/test/regression/tests.toml
+++ b/libos/test/regression/tests.toml
@@ -37,6 +37,7 @@ manifests = [
   "file_size",
   "fopen_cornercases",
   "fork_and_exec",
+  "fork_disallowed",
   "fp_multithread",
   "fstat_cwd",
   "futex_bitset",
--- a/libos/test/regression/tests_musl.toml
+++ b/libos/test/regression/tests_musl.toml
@@ -39,6 +39,7 @@ manifests = [
   "file_size",
   "fopen_cornercases",
   "fork_and_exec",
+  "fork_disallowed",
   "fp_multithread",
   "fstat_cwd",
   "futex_bitset",


heavenboy8:
Hi, 

This feature is very interesting and important for our own usage of Gramine. :+1: Indeed, we are spawning enclaves on demand containing applications we do not write by ourself and can't monitor. In our scenario, on an SGX machine, several enclaves should run independently and do not exceed the resources they own. Especially, each application can run only one enclave and no more (otherwise an app will take resources (memory) from another app). Therefore having the ability to disallow the fork is a major concern for us. This feature is exactly what we are expecting for. On a security point of view, it enables us to prevent an application to act as a fork bomb. 

Great job and thanks for this feature. :clap: Could we expect it to be integrated in the incoming release 1.4 ?

dimakuv:
FYI: Jenkins-Direct-Sanitizers failed due to timeout on Blender download (known problem), but I can't restart the pipeline because our Gramine web-site certificate expired.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/983)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/983#-NNYrT4H2Rsefh5-B7B8:-NNYrT4I9nddqe1puxSj:b-3afv3q) (no related file):*
> On a security point of view, it enables us to prevent an application to act as a fork bomb.

@heavenboy8: This PR doesn't provide such guarantees. Gramine is a one-way sandbox - it protects the app from the host, but not the other way. If you want to run untrusted apps inside Gramine then you should wrap everything inside a security sandbox, e.g. https://github.com/google/nsjail.

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/983#-NNYrw8S57LU9vej-bUl:-NNYrw8S57LU9vej-bUm:b8kud8d) (no related file):*
> Great job and thanks for this feature. 👏 Could we expect it to be integrated in the incoming release 1.4 ?

@heavenboy8: Nope, sorry. v1.4 is going out in a moment and this PR isn't ready yet.

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/983#-NNYsBSuBkVsD3Boiz78:-NNYsBSuBkVsD3Boiz79:b88lg5c) (no related file):*
> There is no simple way to disable such functionality in the apps themselves, so we introduce a new manifest option to force apps to use only the main process

But there's a mechanism made exactly for this in Linux - resource limits. Why not just add emulation for the Linux limit of processes, instead of this custom and less flexible solution?

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/983)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/983#-NNYrT4H2Rsefh5-B7B8:-NN_qOVR-RHyKKe4CE6a:b-pn0rq0) (no related file):*

> On a security point of view, it enables us to prevent an application to act as a fork bomb.

@heavenboy8: This PR doesn't provide such guarantees. Gramine is a one-way sandbox - it protects the app from the host, but not the other way. If you want to run untrusted apps inside Gramine then you should wrap everything inside a security sandbox, e.g. https://github.com/google/nsjail.
</blockquote></details>

I think there is a point to what @heavenboy8 says. Imagine we have the following scenario:
- Benign host
- Random, possibly malicious, apps on top of Gramine

The host will never do a fork bomb, as it is benign. However, some malicious app may be a fork bomb. There would be two ways to prevent this:
- Limit the number of (child) processes on the host,
- Have something like this PR, and rely on Gramine doing the right thing.

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/983#-NNYsBSuBkVsD3Boiz78:-NN_rQWf3G7X_Vny04QZ:bsh159q) (no related file):*
@mkow I like the idea, but could you elaborate?

The main advantage of this PR is that it doesn't require any interaction/request from the application. IIUC, you suggest to implement proper emulation of `nproc` limit inside Gramine. But this would require the application to issue `setrlimit(RLIMIT_NPROC)` on startup/at some point.

Or maybe you mean we should add something like this in the manifest?
```
sys.rlimits.nproc.soft = 1
sys.rlimits.nproc.hard = -1 (aka infinity)
```

If that's your proposal, I really like this idea. We could even shove our already-existing manifest options (like `stack.size` and `brk.size`) under this manifest-options group.

> ... instead of this custom and less flexible solution?

Well, I can only agree. My solution was quick-and-dirty and is not flexible at all. The idea with Linux's resource limits I like much more.

___

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/983)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/983#-NNYrT4H2Rsefh5-B7B8:-NNaayi_DMCf70OlvQ5H:bxtljpw) (no related file):*
> some malicious app may be a fork bomb

> Have something like this PR, and rely on Gramine doing the right thing.

But this PR is _not_ protecting against this scenario. As I said, Gramine is only a one way sandbox and we have no way to restrict the app anyhow if it's malicious. This PR is restricting only non-malicious apps which are being nice to Gramine.

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/983#-NNYsBSuBkVsD3Boiz78:-NNabV-V7RPzbS_LSOU9:b-2iz4r7) (no related file):*
> Or maybe you mean we should add something like this in the manifest?
> 
> ```
> sys.rlimits.nproc.soft = 1
> sys.rlimits.nproc.hard = -1 (aka infinity)
> ```

Yes, exactly. Emulate the Linux mechanism and add a way to specify its initial state in the manifest.

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/983)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/983#-NNYrT4H2Rsefh5-B7B8:-NNaf4H30x4_qkQA9HdT:b-kcgo0e) (no related file):*
> This PR is restricting only non-malicious apps which are being nice to Gramine.

Hm, ok, you're right. My scenario with a malicious app is bogus. The only real scenario that this PR would cover is when both the host and the app are benign, but the app may do something security- or privacy-critical when it notices that it's run under root.

Anyway, in the other comment we agreed on having proper Linux resource limits emulation, which would make this PR unnecessary.

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/983#-NNYsBSuBkVsD3Boiz78:-NNaf_uL31glcg_ruVlp:bb4g17k) (no related file):*

> Or maybe you mean we should add something like this in the manifest?
> 
> ```
> sys.rlimits.nproc.soft = 1
> sys.rlimits.nproc.hard = -1 (aka infinity)
> ```

Yes, exactly. Emulate the Linux mechanism and add a way to specify its initial state in the manifest.
</blockquote></details>

Yes, good idea. Let me try to cook a PR this week, let's see how hard it is. If it will turn out to be too time-consuming, I'll create a GitHub issue instead.

I'll keep this PR open for now, until I either create another PR or a new GitHub issue.

___

heavenboy8:
Hi, 

Actually it's more a matter of "quality of service" than security in our case scenario. 
If my application is supposed to take 8GB of the EPC but it actually takes 16GB because of a fork (which generates a second enclave), it will eat the EPC memory space dedicated to another application and make the other enclaves running in the same SGX machine swap. The performance will then be bad. 

I like the solution implemented in this PR because it controls a behaviour implemented by Gramine: spawn another enclave while a fork occurs. Another choice could have been to refuse the fork because it generates more memory than we have set in the manifest. With that option, we have a better control of the fork behaviour. Let's remind that the problem issues when we import Flask or tensorflow: very common python library.

I've tried to play with `ulimit` and `nproc` instead. It looks like a good idea. However: 

> RLIMIT_NPROC (https://man7.org/linux/man-pages/man2/getrlimit.2.html): This is a limit on the number of extant process (or, more precisely on Linux, **threads**) for the **real user ID** of the calling process.  So long as the current number of processes belonging to this process's real user ID is  greater than or equal to this limit, **fork(2) fails with the error EAGAIN.**

Several things are annoying here:  

1. `ulimit` works on a user and not a single process. So it's not convenient (even impossible in some cases) to put a limit  for the whole processes of a given user whereas we just want to control the fork of a single application under Gramine. 
2. Put `nproc=1` also prevents us from spawning threads (which was completely okay here because spawning a thread does not spawn another enclave). `OpenBLAS blas_thread_init: pthread_create failed for thread 47 of 48: Resource temporarily unavailable.`
2. `EAGAIN` makes the enclave stop under Gramine. Whereas this implementation, does not.

dimakuv (commented):
@heavenboy8 Thanks for looking at this again, very interesting observations. Some comments.

> Another choice could have been to refuse the fork because it generates more memory than we have set in the manifest.

Hm, I don't like this. This is significantly more complicated to track than number of processes. Especially when we take into account EDMM.

> ulimit works on a user and not a single process. So it's not convenient (even impossible in some cases) to put a limit for the whole processes of a given user whereas we just want to control the fork of a single application under Gramine.

This is not a problem for Gramine. Gramine kinda re-creates a Linux environment from the application perspective, so the limit will be only applicable **inside Gramine** and not on your host machine. And since Gramine runs as a single user, the limit will always work for the application that is started under Gramine.

> Put `nproc=1` also prevents us from spawning threads (which was completely okay here because spawning a thread does not spawn another enclave).

Ok, this is something I didn't realize, and this is a real complication for us. I didn't know that `RLIMIT_NPROC` works on Linux tasks (which are both processes and threads), and so indeed -- we have a problem. @mkow @oshogbo  any thoughts? One idea is to introduce a new, Gramine-specific limit like `RLIMIT_NFORK`...

> EAGAIN makes the enclave stop under Gramine. Whereas this implementation, does not.

I'm not sure what you mean here. I guess you mean that you tested some application under Gramine, and if it got EAGAIN, it stopped/terminated? I don't understand why some app would do that on EAGAIN (and won't do it on ENOSYS, as in this PR).

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/983#-:-NNvgzX237fLql9ZxA7t:b-bi9xu9)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

mkow (commented):
@heavenboy8: If you could, please use Reviewable when replying, this way the discussions will be properly threaded.

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/983#-:-NO1tVfi7lYa9DO0VM2x:b-ancs0i)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/983#-NNYsBSuBkVsD3Boiz78:-NO1u5hE3ycZ605PH4EO:b-i6drps) (no related file):*
> Several things are annoying here:
> 1. `ulimit` works on a user and not a single process. So it's not convenient (even impossible in some cases) to put a limit  for the whole processes of a given user whereas we just want to control the fork of a single application under Gramine.
> 2. Put `nproc=1` also prevents us from spawning threads (which was completely okay here because spawning a thread does not spawn another enclave). `OpenBLAS blas_thread_init: pthread_create failed for thread 47 of 48: Resource temporarily unavailable.`

Ah, right. This makes `RLIMIT_NPROC` quite useless for your case.

> One idea is to introduce a new, Gramine-specific limit like `RLIMIT_NFORK`...

I don't like the idea of modifying the Linux ABI we emulate.

@dimakuv: I don't have other ideas, so I guess we have to proceed with the approach from this PR?

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/983)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/983#-NNYsBSuBkVsD3Boiz78:-NO99TYW7Jh80Te9owev:b-q3q3j4) (no related file):*

> Several things are annoying here:
> 1. `ulimit` works on a user and not a single process. So it's not convenient (even impossible in some cases) to put a limit  for the whole processes of a given user whereas we just want to control the fork of a single application under Gramine.
> 2. Put `nproc=1` also prevents us from spawning threads (which was completely okay here because spawning a thread does not spawn another enclave). `OpenBLAS blas_thread_init: pthread_create failed for thread 47 of 48: Resource temporarily unavailable.`

Ah, right. This makes `RLIMIT_NPROC` quite useless for your case.

> One idea is to introduce a new, Gramine-specific limit like `RLIMIT_NFORK`...

I don't like the idea of modifying the Linux ABI we emulate.

@dimakuv: I don't have other ideas, so I guess we have to proceed with the approach from this PR?
</blockquote></details>

Well, ok. If we don't want to introduce any new interfaces, then yes, my approach seems the most straight-forward.

I will add a test case now, and this PR is ready to go then.

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/983)* status: 0 of 6 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/983#-NNYsBSuBkVsD3Boiz78:-NO9I-Ie0Jc0evNYNhqn:b4rc7df) (no related file):*

Well, ok. If we don't want to introduce any new interfaces, then yes, my approach seems the most straight-forward.

I will add a test case now, and this PR is ready to go then.
</blockquote></details>

Done, added documentation and a test.

___

heavenboy8:
Hi @dimakuv , 
Thanks for the last changes. Can we expect a merge on `main` branch soon? We will be glad to use this new feature.

dimakuv (commented):
@heavenboy8 We don't have a `main` branch, we have a `master` branch :)

I think yes, we'll merge this soon-ish. @mkow needs to re-review, but he was generally Ok with this implementation.

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/983#-:-NOV3qwF6bZVrdnvxyMz:b3y304x)* status: 0 of 6 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

mkow (changes requested):
> Reviewed 1 of 1 files at r2, 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/983)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`Documentation/manifest-syntax.rst` line 353 at r3](https://reviewable.io/reviews/gramineproject/gramine/983#-NOgBZ4A2gOaivhI3juH:-NOgBZ4B3C7hO1ht4UHa:b-smugc6) ([raw file](https://github.com/gramineproject/gramine/blob/38fda632e5329c843d908e9802a0e08e3cb3d8e5/Documentation/manifest-syntax.rst#L353)):*
To avoid misuse, please add:

> This option is _not_ a security feature - Gramine by-design is only a one-way sandbox, which doesn't protect the host from the enclave. Don't use this option if you want to somehow mitigate running untrusted enclaves. Instead, to achieve this, you need to run the whole Gramine inside a proper security sandbox.
___
*[`libos/src/sys/libos_clone.c` line 326 at r3](https://reviewable.io/reviews/gramineproject/gramine/983#-NOgDkOw2lKondm2oWHX:-NOgDkOw2lKondm2oWHY:b-2e58hb) ([raw file](https://github.com/gramineproject/gramine/blob/38fda632e5329c843d908e9802a0e08e3cb3d8e5/libos/src/sys/libos_clone.c#L326)):*
> ```C
>                             "(sys.disallow_subprocesses = true)");
>             }
>             return -ENOSYS;
> ```

Not something else? E.g. the same error as returned when you exceed rlimits.
___
*[`libos/test/regression/fork_disallowed.manifest.template` line 14 at r3](https://reviewable.io/reviews/gramineproject/gramine/983#-NOgED771zcdIl1lDd73:-NOgED771zcdIl1lDd74:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/38fda632e5329c843d908e9802a0e08e3cb3d8e5/libos/test/regression/fork_disallowed.manifest.template#L14)):*
> ```
> ]
> 
> # must print a warning `The app tried to create a subprocess, but this is disabled`
> ```





_Suggestion:_
```
warning: "The app tried to create a subprocess, but this is disabled"
```

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/983)* status: 3 of 6 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`Documentation/manifest-syntax.rst` line 353 at r3](https://reviewable.io/reviews/gramineproject/gramine/983#-NOgBZ4A2gOaivhI3juH:-NOiHfue4hIofR5ZtYhg:bb6un1j) ([raw file](https://github.com/gramineproject/gramine/blob/38fda632e5329c843d908e9802a0e08e3cb3d8e5/Documentation/manifest-syntax.rst#L353)):*

To avoid misuse, please add:

> This option is _not_ a security feature - Gramine by-design is only a one-way sandbox, which doesn't protect the host from the enclave. Don't use this option if you want to somehow mitigate running untrusted enclaves. Instead, to achieve this, you need to run the whole Gramine inside a proper security sandbox.
</blockquote></details>

Done
___
*[`libos/src/sys/libos_clone.c` line 326 at r3](https://reviewable.io/reviews/gramineproject/gramine/983#-NOgDkOw2lKondm2oWHX:-NOiIUTj1B7Sa9mNZ_0a:brcbkow) ([raw file](https://github.com/gramineproject/gramine/blob/38fda632e5329c843d908e9802a0e08e3cb3d8e5/libos/src/sys/libos_clone.c#L326)):*

Not something else? E.g. the same error as returned when you exceed rlimits.
</blockquote></details>

Done. Surprisingly, the error is `EAGAIN`, see `man 2 fork` for details (`man 2 clone` redirects to `man 2 fork` for this error code).
___
*[`libos/test/regression/fork_disallowed.manifest.template` line 14 at r3](https://reviewable.io/reviews/gramineproject/gramine/983#-NOgED771zcdIl1lDd73:-NOiHpvc8BvMfE2c06SN:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/38fda632e5329c843d908e9802a0e08e3cb3d8e5/libos/test/regression/fork_disallowed.manifest.template#L14)):*
> ```
> ]
> 
> # must print a warning `The app tried to create a subprocess, but this is disabled`
> ```

Done.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/983)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`libos/src/sys/libos_clone.c` line 326 at r3](https://reviewable.io/reviews/gramineproject/gramine/983#-NOgDkOw2lKondm2oWHX:-NOiSKOJCoaX_QWNe2Ro:bm9om98) ([raw file](https://github.com/gramineproject/gramine/blob/38fda632e5329c843d908e9802a0e08e3cb3d8e5/libos/src/sys/libos_clone.c#L326)):*
> Tested some Python workloads; they stopped spawning child enclaves and still behave correctly.

Just to be sure, could you check these workloads again after this change?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/983)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`libos/src/sys/libos_clone.c` line 326 at r3](https://reviewable.io/reviews/gramineproject/gramine/983#-NOgDkOw2lKondm2oWHX:-NOil62k6b_-83OmNT-H:b55hsjd) ([raw file](https://github.com/gramineproject/gramine/blob/38fda632e5329c843d908e9802a0e08e3cb3d8e5/libos/src/sys/libos_clone.c#L326)):*

> Tested some Python workloads; they stopped spawning child enclaves and still behave correctly.

Just to be sure, could you check these workloads again after this change?
</blockquote></details>

Done.

I also updated the PR description (see "How to test this PR" section), so that everyone can quickly reproduce my tests.

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/983)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/983)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`libos/src/sys/libos_clone.c` line 315 at r4](https://reviewable.io/reviews/gramineproject/gramine/983#-NOlaBOq4uDbFISmrxtH:-NOlaBOq4uDbFISmrxtI:b-80oidc) ([raw file](https://github.com/gramineproject/gramine/blob/721d2b48f509ce0bbca93b124833d82ae3b06b7f/libos/src/sys/libos_clone.c#L315)):*
> ```C
>         bool disallow_subprocesses;
>         ret = toml_bool_in(g_manifest_root, "sys.disallow_subprocesses",
>                            /*defaultval=*/false, &disallow_subprocesses);
> ```

Please rewrap.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/983)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

___
*[`libos/src/sys/libos_clone.c` line 315 at r4](https://reviewable.io/reviews/gramineproject/gramine/983#-NOlaBOq4uDbFISmrxtH:-NOn2W_rEhenQUMxjck_:b89z1lz) ([raw file](https://github.com/gramineproject/gramine/blob/721d2b48f509ce0bbca93b124833d82ae3b06b7f/libos/src/sys/libos_clone.c#L315)):*

Please rewrap.
</blockquote></details>

There is nothing to rewrap? Everything seems to be already conforming to 100-char limit.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/983)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`libos/src/sys/libos_clone.c` line 315 at r4](https://reviewable.io/reviews/gramineproject/gramine/983#-NOlaBOq4uDbFISmrxtH:-NOn5ULv0621q_vFniE1:b-6irua0) ([raw file](https://github.com/gramineproject/gramine/blob/721d2b48f509ce0bbca93b124833d82ae3b06b7f/libos/src/sys/libos_clone.c#L315)):*

There is nothing to rewrap? Everything seems to be already conforming to 100-char limit.
</blockquote></details>

Shouldn't we rewrap to below (or I misunderstood the guidelines)?



_Suggestion:_
```C
        ret = toml_bool_in(g_manifest_root, "sys.disallow_subprocesses", /*defaultval=*/false,
                           &disallow_subprocesses);
```

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/983)* status: 5 of 6 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin and @mkow)

___
*[`libos/src/sys/libos_clone.c` line 315 at r4](https://reviewable.io/reviews/gramineproject/gramine/983#-NOlaBOq4uDbFISmrxtH:-NOn9NjA9g8zcUrpeH5v:begbomf) ([raw file](https://github.com/gramineproject/gramine/blob/721d2b48f509ce0bbca93b124833d82ae3b06b7f/libos/src/sys/libos_clone.c#L315)):*

Shouldn't we rewrap to below (or I misunderstood the guidelines)?
</blockquote></details>

Done. You are right. I am used to automatically rewrapping in my Vim editor, using `Gq` key combination. Interesting, it didn't work correctly this time, no idea why.

mkow (approved):
> Reviewed 1 of 1 files at r5, 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/983)* status: :shipit: complete! all files reviewed, all discussions resolved

dimakuv:
Jenkins, retest pkg-rpm-almalinux9 please (`Curl error (6): Couldn't resolve host name for https://mirrors.almalinux.org/mirrorlist/9/appstream [getaddrinfo() thread failed to start]`)

dimakuv:
Jenkins, retest pkg-rpm-almalinux9 please

Issue #984, opened 2022-10-17
Pathnames for named UNIX Domain Sockets


dimakuv:
### Description of the feature

We currently do not implement pathnames for named UDSes (see `man 7 unix`): https://github.com/gramineproject/gramine/blob/546c598b694d091121babc215bf6b9f9f52bbb4d/libos/src/net/unix.c#L11-L15

Actually, we need two features:
1. Pathnames for named UDSes.
2. Ability to find the UDS created in another, sibling Gramine process.

### Pathnames for named UDSes

This shouldn't be too hard. We can re-use the logic already implemented for Named Pipes (FIFOs, see `man 7 fifo`):
- https://github.com/gramineproject/gramine/blob/546c598b694d091121babc215bf6b9f9f52bbb4d/libos/src/fs/pipe/fs.c#L304-L308
- https://github.com/gramineproject/gramine/blob/546c598b694d091121babc215bf6b9f9f52bbb4d/libos/src/fs/pipe/fs.c#L201
- https://github.com/gramineproject/gramine/blob/546c598b694d091121babc215bf6b9f9f52bbb4d/libos/src/fs/pipe/fs.c#L63
- https://github.com/gramineproject/gramine/blob/546c598b694d091121babc215bf6b9f9f52bbb4d/libos/src/sys/libos_pipe.c#L184

Basically, we create a `uds_built_fs` that allows to "see" the file in Gramine FS, and then create dentry and inode objects and associate them with the UDS handle.

### Finding the UDS created in another, sibling Gramine process

This is harder. Because we have a scenario where one subprocess (`wineserver`, spawned from the main process `bash`) creates a UDS, and another subprocess (`wine64`, spawned from the main process `bash`) connects to it.

This means that `wine64` will not see the created UDS in its Gramine FS view, because the UDS inode is not on the host (recall that what's created on the host is the UDS called smth like `/tmp/gramine/0001/12345edcba`). And this UDS inode is not in `wine64` because the parent process (`bash`) doesn't have it too; this UDS inode is solely in `wineserver`.

So we need a mechanism to make the UDS synced to other processes (or some simpler approximation of this). Several suggestions:
1. Use the existing [sync engine](https://github.com/gramineproject/gramine/blob/master/libos/include/libos_sync.h).
    - Pros: may be used to properly notify all sibling processes about UDS creation/rename/deletion/etc.
    - Cons: very complicated to use; we don't have a proper "checkpoint and restore a dentry+inode" mechanism to reuse.
[...]

PR #986, opened 2022-10-18
[LibOS] Create file on host corresponding to named UNIX socket

dimakuv:
## Description of the changes 

Named UNIX domain sockets are expected to be backed by files. Previously, Gramine didn't have such backing files. However, some workloads require this, e.g., two subprocesses spawned in the same Gramine instance and communicating over a named UNIX socket.

This is a poor-man's implementation: LibOS creates a dummy regular file on the host, so that both subprocesses can see it. This poor implementatoin is chosen because there is no file synchronization mechanism available in Gramine.

See also discussions in #984.

## How to test this PR? 

A new LibOS regression test is added that tests this.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/986)

### DIFF
--- a/libos/include/libos_fs.h
+++ b/libos/include/libos_fs.h
@@ -890,6 +890,9 @@ struct libos_inode* get_new_inode(struct libos_mount* mount, mode_t type, mode_t
 void get_inode(struct libos_inode* inode);
 void put_inode(struct libos_inode* inode);
 
+/* same as libos_syscall_openat() but filename can be in Gramine-private address space */
+long do_openat(int dfd, const char* filename, int flags, int mode);
+
 /*
  * Hashing utilities for paths.
  *
--- a/libos/src/net/unix.c
+++ b/libos/src/net/unix.c
@@ -9,9 +9,18 @@
  */
 
 /*
- * TODO: Currently pathname UNIX domain sockets are not visible on the Gramine filesystem (they
- * do not have a corresponding dentry). This shouldn't be hard to implement, but leaving this as
- * a todo for now - nothing seemed to require it, at least so far.
+ * FIXME: Currently named UNIX domain sockets are visible on the Gramine filesystem via a dummy file
+ *        (created on the host). Ideally, the file should be in-Gramine only and synchronized among
+ *        all processes of the Gramine instance, but we do not have such synchronization mechanism.
+ *
+ *        There are several issues with current approach of a dummy on-the-host file:
+ *          - If the file is created on a tmpfs mount, a sibling Gramine process won't be able to
+ *            connect to it anyway.
+ *          - Two independent Gramine instances may race on creating this dummy file on the host,
+ *            and one of them will fail (this wouldn't be the case with in-Gramine-only files).
+ *          - The dummy file is a regular file, it can be read/written but this won't affect the
+ *            behavior of the actual UNIX socket. On the other hand, real-world apps won't access
+ *            this file knowing that it is a UNIX socket.
  */
 
 #include "crypto.h"
@@ -19,7 +28,9 @@
 #include "libos_fs.h"
 #include "libos_internal.h"
 #include "libos_socket.h"
+#include "libos_table.h"
 #include "pal.h"
+#include "perm.h"
 
 /*!
  * \brief Verify UNIX socket address and convert it to a unique socket name.
@@ -72,6 +83,45 @@ static int unaddr_to_sockname(void* addr, size_t* addrlen, char* sock_name, size
     }
     assert(sock_name_size >= 2 * sizeof(hash) + 1);
     bytes2hex(hash, sizeof(hash), sock_name, sock_name_size);
+    return 0;
+}
+
+/*!
+ * \brief Extract UNIX socket address if it is a named UNIX socket.
+ *
+ * \param         addr           The socket address to convert.
+ * \param[in,out] addrlen        Pointer to the size of \p addr.
+ * \param[out]    named          Is it a named UNIX socket. If yes then \p pathname is updated.
+ * \param[out]    pathname       Buffer for the output pathname of the socket. On success
+ *                               contains a null terminated string.
+ * \param         pathname_size  Size of \p pathname.
+ *
+ * This function nust be called after unaddr_to_sockname() because it assumes \p addr to point to a
+ * correct sockaddr_un object.
+ */
+static int unaddr_to_pathname(void* addr, size_t addrlen, bool* named, char* pathname,
+                              size_t pathname_size) {
+    if (addrlen > sizeof(struct sockaddr_un)) {
+        addrlen = sizeof(struct sockaddr_un);
+    }
+
+    const char* path = (char*)addr + offsetof(struct sockaddr_un, sun_path);
+    size_t pathlen = addrlen - offsetof(struct sockaddr_un, sun_path);
+    assert(pathlen >= 1);
+    if (!path[0]) {
+        /* Unnamed UNIX socket. */
+        *named = false;
+        return 0;
+    }
+
+    /* Named UNIX socket. */
+    pathlen = strnlen(path, pathlen);
+    if (pathname_size < pathlen + 1) {
+        return -EINVAL;
+    }
+    memset(pathname, 0, pathname_size);
+    memcpy(pathname, path, pathlen);
+    *named = true;
     return 0;
 }
 
@@ -129,6 +179,27 @@ static int bind(struct libos_handle* handle, void* addr, size_t addrlen) {
                                  sizeof(pipe_name) - static_strlen(URI_PREFIX_PIPE_SRV));
     if (ret < 0) {
         return ret;
+    }
+
+    /* Named UNIX domain sockets are currently backed by a dummy file. This allows dir-specific
+     * operations like lookup, stat, and unlink. */
+    bool named;
+    size_t pathname_size = sizeof(struct sockaddr_un) - offsetof(struct sockaddr_un, sun_path) + 1;
+    char pathname[pathname_size];
+    ret = unaddr_to_pathname(addr, addrlen, &named, pathname, pathname_size);
+    if (ret < 0) {
+        return ret;
+    }
+
+    if (named) {
+        int dummy_file_fd = do_openat(AT_FDCWD, pathname, O_CREAT | O_EXCL, PERM_rw_______);
+        if (dummy_file_fd < 0) {
+            return dummy_file_fd == -EEXIST ? -EADDRINUSE: dummy_file_fd;
+        }
+        ret = libos_syscall_close(dummy_file_fd);
+        if (ret < 0) {
+            return ret;
+        }
     }
 
     lock(&handle->lock);
@@ -226,6 +297,26 @@ static int connect(struct libos_handle* handle, void* addr, size_t addrlen) {
                                  sizeof(pipe_name) - static_strlen(URI_PREFIX_PIPE));
     if (ret < 0) {
         return ret;
+    }
+
+    /* Named UNIX domain sockets are currently backed by a dummy file. Check that file exists. */
+    bool named;
+    size_t pathname_size = sizeof(struct sockaddr_un) - offsetof(struct sockaddr_un, sun_path) + 1;
+    char pathname[pathname_size];
+    ret = unaddr_to_pathname(addr, addrlen, &named, pathname, pathname_size);
+    if (ret < 0) {
+        return ret;
+    }
+
+    if (named) {
+        int dummy_file_fd = do_openat(AT_FDCWD, pathname, O_RDWR, /*mode=*/0);
+        if (dummy_file_fd < 0) {
+            return dummy_file_fd;
+        }
+        ret = libos_syscall_close(dummy_file_fd);
+        if (ret < 0) {
+            return ret;
+        }
     }
 
     lock(&handle->lock);
--- a/libos/src/sys/libos_open.c
+++ b/libos/src/sys/libos_open.c
@@ -100,7 +100,7 @@ long libos_syscall_creat(const char* path, mode_t mode) {
     return libos_syscall_open(path, O_CREAT | O_TRUNC | O_WRONLY, mode);
 }
 
-long libos_syscall_openat(int dfd, const char* filename, int flags, int mode) {
+long do_openat(int dfd, const char* filename, int flags, int mode) {
     /* Clear invalid flags. */
     flags &= O_ACCMODE | O_APPEND |  O_CLOEXEC | O_CREAT | O_DIRECT | O_DIRECTORY | O_DSYNC | O_EXCL
              | O_LARGEFILE | O_NOATIME | O_NOCTTY | O_NOFOLLOW | O_NONBLOCK | O_PATH | O_SYNC
@@ -111,9 +111,6 @@ long libos_syscall_openat(int dfd, const char* filename, int flags, int mode) {
     if (flags & O_PATH) {
         flags &= O_PATH | O_CLOEXEC | O_DIRECTORY | O_NOFOLLOW;
     }
-
-    if (!is_user_string_readable(filename))
-        return -EFAULT;
 
     if (!(flags & O_CREAT)) {
         /* `mode` should be ignored if O_CREAT is not specified, according to man */
@@ -158,6 +155,13 @@ out:
     if (dir)
         put_dentry(dir);
     return ret;
+}
+
+long libos_syscall_openat(int dfd, const char* filename, int flags, int mode) {
+    if (!is_user_string_readable(filename))
+        return -EFAULT;
+
+    return do_openat(dfd, filename, flags, mode);
 }
 
 long libos_syscall_close(int fd) {
--- a/libos/test/regression/meson.build
+++ b/libos/test/regression/meson.build
@@ -127,6 +127,7 @@ tests = {
     'udp': {},
     'uid_gid': {},
     'unix': {},
+    'unix_subprocesses': {},
     'vfork_and_exec': {},
 }
 
--- a/libos/test/regression/test_libos.py
+++ b/libos/test/regression/test_libos.py
@@ -1284,10 +1284,10 @@ class TC_80_Socket(RegressionTestCase):
         self.assertIn('[parent] TEST OK', stdout)
 
     def test_100_socket_unix(self):
-        if os.path.exists("dummy"):
-            os.remove("dummy")
-        if os.path.exists("u"):
-            os.remove("u")
+        if os.path.exists("tmp/dummy"):
+            os.remove("tmp/dummy")
+        if os.path.exists("tmp/u"):
+            os.remove("tmp/u")
 
         stdout, _ = self.run_binary(['unix'])
         self.assertIn('Data: This is packet 0', stdout)
@@ -1300,6 +1300,13 @@ class TC_80_Socket(RegressionTestCase):
         self.assertIn('Data: This is packet 7', stdout)
         self.assertIn('Data: This is packet 8', stdout)
         self.assertIn('Data: This is packet 9', stdout)
+
+    def test_101_socket_unix_subprocesses(self):
+        if os.path.exists("tmp/uds_subprocesses"):
+            os.remove("tmp/uds_subprocesses")
+
+        stdout, _ = self.run_binary(['unix_subprocesses'])
+        self.assertIn('TEST OK', stdout)
 
     def test_200_socket_udp(self):
         stdout, _ = self.run_binary(['udp'])
--- a/libos/test/regression/tests.toml
+++ b/libos/test/regression/tests.toml
@@ -111,6 +111,7 @@ manifests = [
   "udp",
   "uid_gid",
   "unix",
+  "unix_subprocesses",
   "vfork_and_exec",
 ]
 
--- a/libos/test/regression/tests_musl.toml
+++ b/libos/test/regression/tests_musl.toml
@@ -112,6 +112,7 @@ manifests = [
   "udp",
   "uid_gid",
   "unix",
+  "unix_subprocesses",
   "vfork_and_exec",
 ]
 
--- a/libos/test/regression/unix.c
+++ b/libos/test/regression/unix.c
@@ -47,7 +47,7 @@ static int server_dummy_socket(void) {
         printf("Dummy socket was created\n");
 
     address.sun_family = AF_UNIX;
-    strncpy(address.sun_path, "dummy", sizeof(address.sun_path));
+    strncpy(address.sun_path, "tmp/dummy", sizeof(address.sun_path));
     socklen_t minimal_address_size = offsetof(struct sockaddr_un, sun_path) +
                                      strlen(address.sun_path) + 1;
 
@@ -78,7 +78,7 @@ static int server(void) {
         printf("The socket was created\n");
 
     address.sun_family = AF_UNIX;
-    strncpy(address.sun_path, "u", sizeof(address.sun_path));
+    strncpy(address.sun_path, "tmp/u", sizeof(address.sun_path));
 
     if (bind(create_socket, (struct sockaddr*)&address, sizeof(address)) < 0) {
         perror("bind");
@@ -155,7 +155,7 @@ static int client(void) {
         printf("The socket was created\n");
 
     address.sun_family = AF_UNIX;
-    strncpy(address.sun_path, "u", sizeof(address.sun_path));
+    strncpy(address.sun_path, "tmp/u", sizeof(address.sun_path));
 
     if (connect(create_socket, (struct sockaddr*)&address, sizeof(address)) == 0)
         printf("The connection was accepted with the server\n");
--- a/libos/test/regression/unix_subprocesses.c
+++ b/libos/test/regression/unix_subprocesses.c
@@ -1,0 +1,125 @@
+#define _GNU_SOURCE
+#include <err.h>
+#include <errno.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/socket.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <sys/un.h>
+#include <sys/wait.h>
+#include <unistd.h>
+
+#include "common.h"
+
+#define SRV_ADDR "tmp/uds_subprocesses"
+
+static const char g_buffer[] = "Hello from UDS server!";
+
+static void server(int pipefd) {
+    int s = CHECK(socket(AF_UNIX, SOCK_STREAM, 0));
+
+    struct sockaddr_un sa;
+    sa.sun_family = AF_UNIX;
+    strncpy(sa.sun_path, SRV_ADDR, sizeof(sa.sun_path));
+
+    CHECK(bind(s, (void*)&sa, sizeof(sa)));
+    CHECK(listen(s, 5));
+
+    char c = 0;
+    ssize_t x = CHECK(write(pipefd, &c, sizeof(c)));
+    if (x != 1) {
+        errx(1, "client terminated unexpectedly");
+    }
+    CHECK(close(pipefd));
+
+    int client = CHECK(accept(s, NULL, NULL));
+
+    CHECK(close(s));
+
+    size_t written = 0;
+    while (written < sizeof(g_buffer)) {
+        x = CHECK(write(client, g_buffer + written, sizeof(g_buffer) - written));
+        if (!x) {
+            /* technically impossible, but let's fail loudly if we ever hit this */
+            errx(1, "write to client returned zero");
+        }
+        written += x;
+    }
+
+    CHECK(close(client));
+}
+
+static void client(int pipefd) {
+    char c = 0;
+    ssize_t x = CHECK(read(pipefd, &c, sizeof(c)));
+    if (x != 1) {
+        errx(1, "server terminated unexpectedly");
+    }
+    CHECK(close(pipefd));
+
+    /* named UNIX domain sockets must create FS files, verify it; recall that by default Gramine
+     * creates files with root UID/GID permissions if not specified otherwise in manifest */
+    struct stat statbuf;
+    CHECK(stat(SRV_ADDR, &statbuf));
+    if (statbuf.st_uid != 0 || statbuf.st_gid != 0) {
+        errx(1, "unexpected UID/GID of file `%s`", SRV_ADDR);
+    }
+
+    int s = CHECK(socket(AF_UNIX, SOCK_STREAM, 0));
+
+    struct sockaddr_un sa;
+    sa.sun_family = AF_UNIX;
+    strncpy(sa.sun_path, SRV_ADDR, sizeof(sa.sun_path));
+
+    CHECK(connect(s, (void*)&sa, sizeof(sa)));
+
+    char buf[sizeof(g_buffer) + 1] = { 0 };
+    size_t got = 0;
+    while (got < sizeof(g_buffer)) {
+        x = CHECK(read(s, buf + got, sizeof(g_buffer) - got));
+        if (!x) {
+            /* let's fail loudly if there is no data from server */
+            errx(1, "read from server returned zero");
+        }
+        got += x;
+    }
+
+    if (strcmp(buf, g_buffer)) {
+        errx(1, "unexpected message from server: expected %s, got %s", g_buffer, buf);
+    }
+
+    CHECK(close(s));
+}
+
+int main(void) {
+    int pipefds[2];
+    CHECK(pipe(pipefds));
+
+    pid_t p_s = CHECK(fork());
+    if (p_s == 0) {
+        CHECK(close(pipefds[0]));
+        server(pipefds[1]);
+        return 0;
+    }
+
+    pid_t p_c = CHECK(fork());
+    if (p_c == 0) {
+        CHECK(close(pipefds[1]));
+        client(pipefds[0]);
+        return 0;
+    }
+
+    for (int i = 0; i < 2; i++) {
+        int status = 0;
+        CHECK(wait(&status));
+        if (!WIFEXITED(status) || WEXITSTATUS(status)) {
+            errx(1, "child wait status: %#x", status);
+        }
+    }
+
+    puts("TEST OK");
+    return 0;
+}


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/986)* status: 0 of 9 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfLvnQ2fAMtnQInU0S:-NEfLvnR6d8MclZZ789m:b-frmwoi) (no related file):*
I admit, this is hacky. Consider this PR more of an RFC.

I couldn't come up with a better approach to make UDS files visible, without going into the "let's create a file synchronization mechanism" (which would be cool, but looks very complicated).

___
___
*[`libos/src/net/unix.c` line 197 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfMLba7IVZdprNmvc0:-NEfMLba7IVZdprNmvc1:bwflrdl) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/src/net/unix.c#L197)):*
> ```C
>         int dummy_file_fd = do_openat(AT_FDCWD, pathname, O_CREAT | O_EXCL, PERM_rw_______);
>         if (dummy_file_fd < 0) {
>             return dummy_file_fd == -EEXIST ? -EADDRINUSE: dummy_file_fd;
> ```

FYI: This `EEXIST` -> `EADDRINUSE` is required to return correct error in `bind()`.
___
*[`libos/test/regression/test_libos.py` line 1286 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfMV8245jbTeEf0f-f:-NEfMV8245jbTeEf0f-g:b68u8ba) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/test/regression/test_libos.py#L1286)):*
> ```Python
>         self.assertIn('[parent] TEST OK', stdout)
> 
>     def test_100_socket_unix(self):
> ```

FYI: Actually, I would prefer to remove the old `unix.c` test. My newly added test is a superset of that old one.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/986)* status: all files reviewed, 15 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`-- commits` line 7 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfbIAQAFHrK6OuAKG7):*
> e.g., two subprocesses spawned in the same Gramine instance and communicating over a named UNIX socket.

That's not correct - currently they can connect correctly. What they cannot do is `stat*()` or find the entry in `getdents` returned list.
___
*[`libos/src/net/unix.c` line 16 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfs1EOBSQ2r0xi41YT:-NEfs1EOBSQ2r0xi41YU:b6imd2a) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/src/net/unix.c#L16)):*
> ```C
>  *        all processes of the Gramine instance, but we do not have such synchronization mechanism.
>  *
>  *        There are several issues with current approach of a dummy on-the-host file:
> ```

Aren't these (especially 2nd) blockers for this PR?
___
*[`libos/src/net/unix.c` line 24 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfvctW0DurnWUFbngW:-NEfvctW0DurnWUFbngX:bkwdk4k) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/src/net/unix.c#L24)):*
> ```C
>  *            behavior of the actual UNIX socket. On the other hand, real-world apps won't access
>  *            this file knowing that it is a UNIX socket.
>  */
> ```

Also `stat` will show the file as regular not socket with this PR, I believe? If so then I don't think it's what we want...
___
*[`libos/src/net/unix.c` line 93 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfsQhrDYshvC-We90M:-NEfsQhrDYshvC-We90N:b9vuo2a) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/src/net/unix.c#L93)):*
> ```C
>  *
>  * \param         addr           The socket address to convert.
>  * \param[in,out] addrlen        Pointer to the size of \p addr.
> ```

It's not a pointer?
___
*[`libos/src/net/unix.c` line 104 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfu1x0EI0ducSYo02X:-NEfu1x0EI0ducSYo02Y:b-xdv5ku) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/src/net/unix.c#L104)):*
> ```C
> static int unaddr_to_pathname(void* addr, size_t addrlen, bool* named, char* pathname,
>                               size_t pathname_size) {
>     if (addrlen > sizeof(struct sockaddr_un)) {
> ```

Why do you need this? `addrlen` should be already correct, also you assume that (since there is no check for `< offsetof`)
___
*[`libos/src/net/unix.c` line 122 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfuVcKEwvHN12r-lY2:-NEfuVcLDVgREvzx-i98:bbi3s3v) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/src/net/unix.c#L122)):*
> ```C
>         return -EINVAL;
>     }
>     memset(pathname, 0, pathname_size);
> ```

Why not `pathname[pathlen] = 0;`?
___
*[`libos/src/net/unix.c` line 302 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfvRij7zsw3NgxqxWz:-NEfvRij7zsw3NgxqxX-:b-3v1cex) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/src/net/unix.c#L302)):*
> ```C
>     }
> 
>     /* Named UNIX domain sockets are currently backed by a dummy file. Check that file exists. */
> ```

What's the point of this check? It's racy and not needed - connect will succeed iff socket exist.
___
*[`libos/test/regression/unix_subprocesses.c` line 26 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfdzEpBOidFNFsDenv:-NEfdzEpBOidFNFsDenw:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/test/regression/unix_subprocesses.c#L26)):*
> ```C
>     struct sockaddr_un sa;
>     sa.sun_family = AF_UNIX;
>     strncpy(sa.sun_path, SRV_ADDR, sizeof(sa.sun_path));
> ```





_Suggestion:_
```C
    struct sockaddr_un sa = {
        .sun_family = AF_UNIX,
        .sun_path = SRV_ADDR,
    };
```
___
*[`libos/test/regression/unix_subprocesses.c` line 33 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEffCxBBhvoqcGlh-zI:-NEffCxBBhvoqcGlh-zJ:bv6dfy0) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/test/regression/unix_subprocesses.c#L33)):*
> ```C
>     char c = 0;
>     ssize_t x = CHECK(write(pipefd, &c, sizeof(c)));
>     if (x != 1) {
> ```

Either `1` or `sizeof(c)` everywhere



_Code quote:_
```C
    ssize_t x = CHECK(write(pipefd, &c, sizeof(c)));
    if (x != 1) {
```
___
*[`libos/test/regression/unix_subprocesses.c` line 50 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfegu_8qlJQtlC1Lbz:-NEfegu_8qlJQtlC1Lc-:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/test/regression/unix_subprocesses.c#L50)):*
> ```C
>         }
>         written += x;
>     }
> ```





_Suggestion:_
```C
    x = CHECK(posix_fs_write(client, g_buffer, sizeof(g_buffer)));
    if (x != sizeof(g_buffer)) {
        errx(1, "partial write to client");
    }
```
___
*[`libos/test/regression/unix_subprocesses.c` line 58 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEffIk80tDHi2jxQqJp:-NEffIk80tDHi2jxQqJq:bsfd532) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/test/regression/unix_subprocesses.c#L58)):*
> ```C
>     char c = 0;
>     ssize_t x = CHECK(read(pipefd, &c, sizeof(c)));
>     if (x != 1) {
> ```

ditto (`1` or `sizeof(c)`)



_Code quote:_
```C
    ssize_t x = CHECK(read(pipefd, &c, sizeof(c)));
    if (x != 1) {
```
___
*[`libos/test/regression/unix_subprocesses.c` line 67 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEffSPKD9asM1B8yn4q:-NEffSPKD9asM1B8yn4r:b72007a) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/test/regression/unix_subprocesses.c#L67)):*
> ```C
>     struct stat statbuf;
>     CHECK(stat(SRV_ADDR, &statbuf));
>     if (statbuf.st_uid != 0 || statbuf.st_gid != 0) {
> ```

Please also check `(statbuf.st_mode & S_IFMT) != S_IFSOCK`
___
*[`libos/test/regression/unix_subprocesses.c` line 75 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEffxjj7lneSfI00Xf_:-NEffxjj7lneSfI00Xfa:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/test/regression/unix_subprocesses.c#L75)):*
> ```C
>     struct sockaddr_un sa;
>     sa.sun_family = AF_UNIX;
>     strncpy(sa.sun_path, SRV_ADDR, sizeof(sa.sun_path));
> ```





_Suggestion:_
```C
    struct sockaddr_un sa = {
        .sun_family = AF_UNIX,
        .sun_path = SRV_ADDR,
    };
```
___
*[`libos/test/regression/unix_subprocesses.c` line 81 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfg3AL86YIbnjle7td:-NEfg3AL86YIbnjle7te:bl3g1n1) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/test/regression/unix_subprocesses.c#L81)):*
> ```C
>     char buf[sizeof(g_buffer) + 1] = { 0 };
>     size_t got = 0;
>     while (got < sizeof(g_buffer)) {
> ```

ditto `posix_fd_read(s, buf, sizeof(buf)) - 1`

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/986)* status: all files reviewed, 16 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski)

___
*[`-- commits` line 7 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfbIAQAFHrK6OuAKG7):*

> e.g., two subprocesses spawned in the same Gramine instance and communicating over a named UNIX socket.

That's not correct - currently they can connect correctly. What they cannot do is `stat*()` or find the entry in `getdents` returned list.

</blockquote></details>

Yes, that's what I meant by "backing files". I can tweak the description later.
___
*[`libos/src/net/unix.c` line 16 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfs1EOBSQ2r0xi41YT:-NEjB5-p18gsOvh8c1Ur:b-jwov3t) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/src/net/unix.c#L16)):*

Aren't these (especially 2nd) blockers for this PR?
</blockquote></details>

"Blockers" in which sense?

On the one hand, I want this functionality because some important workloads do not work without it.

On the other hand, I admit that this functionality is implemented in a suboptimal way in this PR. But the optimal way is probably months of work...

So my assumption was: if this change is rather small and not security critical, the PR could be considered. And in (some distant) future, when we have proper sync of files (at least file metadata and notifications), this PR can be reverted in favor of a normal solution.

Basically, it boils down to correctness of functionality vs speed of execution. This PR chooses the latter.

In practical terms, why the 2nd item is a blocker? Can't we say the same for other (regular) files -- if two independent Gramine files operate on `awesome.data` encrypted file, then one/both of them will fail? How is this situation different?
___
*[`libos/src/net/unix.c` line 24 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfvctW0DurnWUFbngW:-NEjCG9sAa_KHthgIcap:b-s3jeh5) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/src/net/unix.c#L24)):*

Also `stat` will show the file as regular not socket with this PR, I believe? If so then I don't think it's what we want...
</blockquote></details>

Yes. Here I have no idea how to fix it, and this is indeed bad.
___
*[`libos/src/net/unix.c` line 93 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfsQhrDYshvC-We90M:-NEjCQuv0Uklus-jlaZ3:b-5386qa) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/src/net/unix.c#L93)):*

It's not a pointer?
</blockquote></details>

No, we already "fixed" it in the above function (`unaddr_to_sockname()`).
___
*[`libos/src/net/unix.c` line 104 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfu1x0EI0ducSYo02X:-NEjCWgI8crM1s0J_uJY:bcdi5cc) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/src/net/unix.c#L104)):*

Why do you need this? `addrlen` should be already correct, also you assume that (since there is no check for `< offsetof`)
</blockquote></details>

Agreed, I can remove this.
___
*[`libos/src/net/unix.c` line 122 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfuVcKEwvHN12r-lY2:-NEjC_NyDnRFJHkgaFrd:b-2jzzcd) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/src/net/unix.c#L122)):*

Why not `pathname[pathlen] = 0;`?
</blockquote></details>

Agreed, I can do this.
___
*[`libos/src/net/unix.c` line 302 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfvRij7zsw3NgxqxWz:-NEjCdLB-8RGiMyQRN8G:bb3xvot) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/src/net/unix.c#L302)):*

What's the point of this check? It's racy and not needed - connect will succeed iff socket exist.
</blockquote></details>

Well, yes, I could remove it. Just felt right to have this -- in this PR, every created named socket has a corresponding (dummy) regular file, so it feels logical to first check this file and then connect to the socket.

But I agree, since this whole "dummy file" is a hack, we can remove this check and immediately try the actual named socket.
___
*[`libos/test/regression/unix.c` line 50 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEjDIUT812_NfTd0tju:-NEjDIUT812_NfTd0tjv:bshvbnw) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/test/regression/unix.c#L50)):*
> ```C
> 
>     address.sun_family = AF_UNIX;
>     strncpy(address.sun_path, "tmp/dummy", sizeof(address.sun_path));
> ```

@boryspoplawski Do you think I should remove this old test and replace it with the new one?

If yes, then I will split it into a separate PR. Even if we don't merge this "dummy-file" PR, at least some good will result out of my efforts.
___
*[`libos/test/regression/unix_subprocesses.c` line 26 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfdzEpBOidFNFsDenv:-NEjD8zb-bM5sXYOG5ik:b-bhiq1l) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/test/regression/unix_subprocesses.c#L26)):*
> ```C
>     struct sockaddr_un sa;
>     sa.sun_family = AF_UNIX;
>     strncpy(sa.sun_path, SRV_ADDR, sizeof(sa.sun_path));
> ```

But `.sun_path` is not a pointer, it's an array embedded in `struct sockaddr_un`. This won't work.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/986)* status: all files reviewed, 16 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`libos/src/net/unix.c` line 16 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfs1EOBSQ2r0xi41YT:-NEkhZ_L6eocf4b1yifv:b-rgpgai) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/src/net/unix.c#L16)):*

"Blockers" in which sense?

On the one hand, I want this functionality because some important workloads do not work without it.

On the other hand, I admit that this functionality is implemented in a suboptimal way in this PR. But the optimal way is probably months of work...

So my assumption was: if this change is rather small and not security critical, the PR could be considered. And in (some distant) future, when we have proper sync of files (at least file metadata and notifications), this PR can be reverted in favor of a normal solution.

Basically, it boils down to correctness of functionality vs speed of execution. This PR chooses the latter.

In practical terms, why the 2nd item is a blocker? Can't we say the same for other (regular) files -- if two independent Gramine files operate on `awesome.data` encrypted file, then one/both of them will fail? How is this situation different?
</blockquote></details>

Idk, I could imagine mysql using some config files from `/var` read-only (so they can be trusted files shared between multiple instances) and creating `/var/socket` 
___
*[`libos/src/net/unix.c` line 24 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfvctW0DurnWUFbngW:-NEkc5xrDqbDruQ-V5yN:bpivs0j) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/src/net/unix.c#L24)):*

Actually, I think I came up with a better approach. Still hacky, but at least "decent" from LibOS perspective:
1. I will create proper `socket_builtin_fs.d_ops` with `stat`, `lookup` and probably some other callbacks: https://github.com/gramineproject/gramine/blob/268dcf45bf15a8e780be48e0600c8c87911446e8/libos/src/fs/socket/fs.c#L223?plain=1
2. I will also use `PalStreamOpen()` to create an underlying file, that will denote the UNIX socket, similar to how `chroot` files do it: https://github.com/gramineproject/gramine/blob/268dcf45bf15a8e780be48e0600c8c87911446e8/libos/src/fs/chroot/fs.c#L222?plain=1
3. The `stat` and `lookup` callbacks will only be used for UNIX domain sockets.
4. The `lookup` callback will use `PalStreamOpen()` to check that the underlying file exists.
5. I can `PalStreamOpen(concatenated_uri_path)`, so that I still disambiguite among Gramine instances.

This will solve the `stat` problem. It will solve all 3 items in the above list, if I understand correctly.

Yep, looks like this will be much better.
</blockquote></details>

How would you check in point 5 that the file is indeed a socket, not some unrelated normal file?
___
*[`libos/src/net/unix.c` line 93 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfsQhrDYshvC-We90M:-NEkcG34AFerw4jRqGvM:b-454gl1) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/src/net/unix.c#L93)):*

No, we already "fixed" it in the above function (`unaddr_to_sockname()`).
</blockquote></details>

I know, but the comment says it's a pointer.
___
*[`libos/test/regression/unix.c` line 50 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEjDIUT812_NfTd0tju:-NEkgpNR3lcM5KM4rskw:b-lvrx0j) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/test/regression/unix.c#L50)):*

@boryspoplawski Do you think I should remove this old test and replace it with the new one?

If yes, then I will split it into a separate PR. Even if we don't merge this "dummy-file" PR, at least some good will result out of my efforts.
</blockquote></details>

Yes please. But check, it might have some more subtests, which we might want to integrate here (but probably rewrite while moving)
___
*[`libos/test/regression/unix_subprocesses.c` line 26 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfdzEpBOidFNFsDenv:-NEkgzOO0LUrxVx3hUg8:bivkw48) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/test/regression/unix_subprocesses.c#L26)):*

But `.sun_path` is not a pointer, it's an array embedded in `struct sockaddr_un`. This won't work.
</blockquote></details>

It will (https://en.cppreference.com/w/c/language/array_initialization)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/986)* status: all files reviewed, 16 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski)

___
*[`libos/src/net/unix.c` line 24 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfvctW0DurnWUFbngW:-NEktENc8T7kFEpwlYFc:b-ba9ob9) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/src/net/unix.c#L24)):*
> How would you check in point 5 that the file is indeed a socket, not some unrelated normal file?

I'm unsure what you're asking about. I will call `PalStreamOpen()` in two places:
1. During UNIX socket `bind()` -- to create a corresponding file on the host.
2. During `socket_buildin_fs::lookup()` callback -- to check if the corresponding file exists.

The underlying file on the host is a regular file and can have whatever name, actually. So I can do a concatenation `/tmp/gramine/<instance_id>/unix_sockets/` + `<unix_socket_name>`. That's what I wanted to say in point 5.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/986)* status: all files reviewed, 16 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`libos/src/net/unix.c` line 24 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfvctW0DurnWUFbngW:-NEkvk5eEYoXZOmLVPaf:b-do3lsi) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/src/net/unix.c#L24)):*

> How would you check in point 5 that the file is indeed a socket, not some unrelated normal file?

I'm unsure what you're asking about. I will call `PalStreamOpen()` in two places:
1. During UNIX socket `bind()` -- to create a corresponding file on the host.
2. During `socket_buildin_fs::lookup()` callback -- to check if the corresponding file exists.

The underlying file on the host is a regular file and can have whatever name, actually. So I can do a concatenation `/tmp/gramine/<instance_id>/unix_sockets/` + `<unix_socket_name>`. That's what I wanted to say in point 5.
</blockquote></details>

And if some other process creates that file, point 5 will think it's a socket (whereas it's not)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/986)* status: all files reviewed, 16 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski)

___
*[`libos/src/net/unix.c` line 24 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfvctW0DurnWUFbngW:-NEkycaN5LnMdgkwRGv8:bj1oqhy) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/src/net/unix.c#L24)):*
> And if some other process creates that file, point 5 will think it's a socket (whereas it's not)

But that's why I mention the prefix to the full path: `/tmp/gramine/<instance_id>/unix_sockets/`. No process is ever supposed to create "dummy-for-unix-sockets" file under this directory.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/986)* status: all files reviewed, 16 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`libos/src/net/unix.c` line 24 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfvctW0DurnWUFbngW:-NEl1-ZU8EQR9RaRKszJ:bvihw1b) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/src/net/unix.c#L24)):*

Clarification: this full path is a URI on the host (i.e., a PAL path). So to avoid confusion, think of it as `file:/tmp/gramine/<instance_id>/unix_sockets/<unix_socket_name>`.

The LibOS path is whatever the application specified as the UNIX domain socket path, which is typically a relative path `<unix_socket_name>`.
</blockquote></details>

Ah, I didn't get this part about concatenation of some Gramine-specific prefix (btw that prefix is now concatenated in PAL and `<instance_id>` is only in PAL, but that should be easily changeable).

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/986)* status: all files reviewed, 16 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski)

___
*[`libos/src/net/unix.c` line 24 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfvctW0DurnWUFbngW:-NEo92V6ApVqRD7OzZUh:b-e8mur4) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/src/net/unix.c#L24)):*

Ah, I didn't get this part about concatenation of some Gramine-specific prefix (btw that prefix is now concatenated in PAL and `<instance_id>` is only in PAL, but that should be easily changeable). 
</blockquote></details>

Yes, as part of the PR I'll expose `instance_id` to LibOS.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/986)* status: all files reviewed, 15 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski)

___
*[`libos/test/regression/unix.c` line 50 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEjDIUT812_NfTd0tju:-NEpWOP955Ru1yFXlYZ-:bed0lxw) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/test/regression/unix.c#L50)):*

Yes please. But check, it might have some more subtests, which we might want to integrate here (but probably rewrite while moving)
</blockquote></details>

Done: https://github.com/gramineproject/gramine/pull/996
___
*[`libos/test/regression/unix_subprocesses.c` line 26 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfdzEpBOidFNFsDenv:-NEpIaSf1JzYqyn0qO-N:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/test/regression/unix_subprocesses.c#L26)):*

It will (https://en.cppreference.com/w/c/language/array_initialization)
</blockquote></details>

Done.
___
*[`libos/test/regression/unix_subprocesses.c` line 33 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEffCxBBhvoqcGlh-zI:-NEpIiMJ30G4-oelUcvl:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/test/regression/unix_subprocesses.c#L33)):*

Either `1` or `sizeof(c)` everywhere
</blockquote></details>

Done.
___
*[`libos/test/regression/unix_subprocesses.c` line 50 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfegu_8qlJQtlC1Lbz:-NEpItB38CITK5zjBItJ:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/test/regression/unix_subprocesses.c#L50)):*
> ```C
>         }
>         written += x;
>     }
> ```

Done.
___
*[`libos/test/regression/unix_subprocesses.c` line 58 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEffIk80tDHi2jxQqJp:-NEpIvra86mdo1yHdR0Z:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/test/regression/unix_subprocesses.c#L58)):*

ditto (`1` or `sizeof(c)`)
</blockquote></details>

Done.
___
*[`libos/test/regression/unix_subprocesses.c` line 67 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEffSPKD9asM1B8yn4q:-NEpJCtr6cZwImuQ6ct1:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/test/regression/unix_subprocesses.c#L67)):*

Please also check `(statbuf.st_mode & S_IFMT) != S_IFSOCK`
</blockquote></details>

Done.
___
*[`libos/test/regression/unix_subprocesses.c` line 75 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEffxjj7lneSfI00Xf_:-NEpJE--BA0SzRYtTU4l:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/test/regression/unix_subprocesses.c#L75)):*
> ```C
>     struct sockaddr_un sa;
>     sa.sun_family = AF_UNIX;
>     strncpy(sa.sun_path, SRV_ADDR, sizeof(sa.sun_path));
> ```

Done.
___
*[`libos/test/regression/unix_subprocesses.c` line 81 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfg3AL86YIbnjle7td:-NEpJj3M7XUC5nVtHnkF:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/test/regression/unix_subprocesses.c#L81)):*

ditto `posix_fd_read(s, buf, sizeof(buf)) - 1`
</blockquote></details>

Done.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/986)* status: all files reviewed, 15 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/986#-NEqEUVU1Xj59X10dOXq:-NEqEUVU1Xj59X10dOXr:b-3gkmx0) (no related file):*
Did you forget to push your changes?

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/986)* status: all files reviewed, 15 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/986#-NEqEUVU1Xj59X10dOXq:-NEqM8tb0Otpbk4yqghi:b2u735) (no related file):*

Did you forget to push your changes?
</blockquote></details>

I marked as "Done" only the changes that I split into this PR: https://github.com/gramineproject/gramine/pull/996. So the changes are there.

___

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/986)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`libos/src/net/unix.c` line 24 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfvctW0DurnWUFbngW:-NEqU4F0D8hwATFXEwU_:b-3uh0ry) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/src/net/unix.c#L24)):*

Yes, as part of the PR I'll expose `instance_id` to LibOS.
</blockquote></details>

I still don't get it. How do you plan on making this work:
```
bind("/var/socket");
stat("/var/socket");
```
I.e. how `stat` knows that the path is a socket not a normal file? I.e. how you want to decide whether to check host `/var/socket` vs `/tmp/gramiine/(...)/var/socket`? What if both exist?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/986)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski)

___
*[`libos/src/net/unix.c` line 24 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfvctW0DurnWUFbngW:-NEuMMi1BV0Yhvx2o2XF:b-mcy163) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/src/net/unix.c#L24)):*

I still don't get it. How do you plan on making this work:
```
bind("/var/socket");
stat("/var/socket");
```
I.e. how `stat` knows that the path is a socket not a normal file? I.e. how you want to decide whether to check host `/var/socket` vs `/tmp/gramiine/(...)/var/socket`? What if both exist?
</blockquote></details>

It will be a bit similar to how we do it for FIFOs (named pipes). In particular, this is the important step: https://github.com/gramineproject/gramine/blob/db86ad2eab2ab449508b85263818f4d6d6e10a50/libos/src/sys/libos_pipe.c#L223

In other words, when I'll perform `bind("/var/socket")`, I will:
1. First check if such a file already exists (as in the code snippet above); if exists then fail.
2. Then create this dentry + inode, as shown here: https://github.com/gramineproject/gramine/blob/f7995b60c7b49e075adefa18573767c6e24a4cc9/libos/src/fs/pipe/fs.c#L63

The important part is that this new dentry+inode will have a new `socket_builtin_fs.d_ops` with `stat` and `lookup` callbacks. So our FS code, whenever the app does `stat("/var/socket")`, will end up performing `socket_builtin_fs.d_ops.lookup()` -- where I will check for my dummy file.

Btw, I refined my initial idea in the meantime -- I don't need a dummy file on the host. I can just query the abstract UNIX socket that was created -- if it exists, then I will return some hard-coded `statbuf` values; if it doesn't exist then `lookup` and `stat` will fail.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/986)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`libos/src/net/unix.c` line 24 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfvctW0DurnWUFbngW:-NEuTFwD3ZBXJqWHJVE9:bt96df1) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/src/net/unix.c#L24)):*

It will be a bit similar to how we do it for FIFOs (named pipes). In particular, this is the important step: https://github.com/gramineproject/gramine/blob/db86ad2eab2ab449508b85263818f4d6d6e10a50/libos/src/sys/libos_pipe.c#L223

In other words, when I'll perform `bind("/var/socket")`, I will:
1. First check if such a file already exists (as in the code snippet above); if exists then fail.
2. Then create this dentry + inode, as shown here: https://github.com/gramineproject/gramine/blob/f7995b60c7b49e075adefa18573767c6e24a4cc9/libos/src/fs/pipe/fs.c#L63

The important part is that this new dentry+inode will have a new `socket_builtin_fs.d_ops` with `stat` and `lookup` callbacks. So our FS code, whenever the app does `stat("/var/socket")`, will end up performing `socket_builtin_fs.d_ops.lookup()` -- where I will check for my dummy file.

Btw, I refined my initial idea in the meantime -- I don't need a dummy file on the host. I can just query the abstract UNIX socket that was created -- if it exists, then I will return some hard-coded `statbuf` values; if it doesn't exist then `lookup` and `stat` will fail.
</blockquote></details>

But that only works in one Gramine process, other processes would have no idea about this dentry+inode

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/986)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski)

___
*[`libos/src/net/unix.c` line 24 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfvctW0DurnWUFbngW:-NEuYkfcEp7NxtmBiark:b6kdsb2) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/src/net/unix.c#L24)):*

But that only works in one Gramine process, other processes would have no idea about this dentry+inode
</blockquote></details>

Yes, you are correct, I also realized it. (Well, it works also in child processes because they get the FS view -- including dentry+inode -- from the parent. But it doesn't work in siblings which only see the FS structure based on the manifest's `fs.mounts`).

The only way to circumvent this that I see is to introduce a new `fs.mounts = [ {type="unix_socket", path="/var/socket" }]`.

This will of course require to be able to know the name of the UNIX socket filename(s) that app creates beforehand... But I guess this is configurable in real-world apps. For example, WINE does it like this: `$HOME/.wine/socket` (in general `$WINEPREFIX/socket`).

@boryspoplawski Do you think it is Ok-ish? I can't come up with anything else...

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/986)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`libos/src/net/unix.c` line 24 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfvctW0DurnWUFbngW:-NEuinF69wldUFWk2eSW:b-72xu3u) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/src/net/unix.c#L24)):*
> The only way to circumvent this that I see is to introduce a new fs.mounts = [ {type="unix_socket", path="/var/socket" }].

But that would require every app to use this syntax (i.e. there seem to be no way of opt-out).

BTW, why do you even need this whole change? Just for wine? Are you sure wine will work in Gramine? The first problem that comes to my mind: Windows uses `gs` segment register for TLS, but Gramine uses the very same `gs` (on Linux `fs` is used for TLS, so no clash)

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/986)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`libos/src/net/unix.c` line 24 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfvctW0DurnWUFbngW:-NF9eZ1I0iJnN41e_liw:b-xaukvs) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/src/net/unix.c#L24)):*
> I am not sure what you mean. What do you mean by "every app"? Every app with the same manifest (that contains this fs.mounts.type = "unix_socket")? Why would that be a problem?

I mean if you add a check to `bind` and `connect` verifying the existence of the file and the app does not specify this file in the manifest, it wouldn't be able to use the unix socket (even if it does not need this fs file otherwise). Is that a problem? No idea, but it's definitely a breaking change.

> What's the problem with the gs register? We can always simply add a swapgs instruction when going into Gramine 

We can't, `swapgs` is ring0 instruction.

> (well, maybe not that simply, but there are definitely tools available to work around this).

We use `gs` for **going into Gramine** and app not using it is currently a hard requirement of Gramine

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/986)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski)

___
*[`libos/src/net/unix.c` line 24 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfvctW0DurnWUFbngW:-NF9kIhf2IQz2tneP87p:b-a2zybi) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/src/net/unix.c#L24)):*
> I mean if you add a check to bind and connect verifying the existence of the file and the app does not specify this file in the manifest, it wouldn't be able to use the unix socket (even if it does not need this fs file otherwise). Is that a problem? No idea, but it's definitely a breaking change.

I don't think this will be needed. Currently, `connect` can only connect to those UDSes that were created in the same process or in the parent. In both cases, the dentry will already exist, so nothing needs to be read from the manifest (in other words, the `lookup` FS callback will never be involved).

Similarly, `bind` will only create a dentry, not look it up. The only case where the `fs.mounts.type="unix_socket"` will play any role is during the `lookup` FS callback, which will only happen in my case of "siblings want to talk over UDS".

> We can't, swapgs is ring0 instruction.

True, I mentioned a wrong instruction. I meant `rdgsbase` and `wrgsbase`.

> We use gs for going into Gramine and app not using it is currently a hard requirement of Gramine

Yes, this one is interesting, and I'll still need to think how to circumvent this.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/986)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`libos/src/net/unix.c` line 24 at r1](https://reviewable.io/reviews/gramineproject/gramine/986#-NEfvctW0DurnWUFbngW:-NFAxxsHAudCr_wQAvlB:b-aou3g4) ([raw file](https://github.com/gramineproject/gramine/blob/c4252d1278b7a13928d36131c17047ebb17ff4cb/libos/src/net/unix.c#L24)):*
> True, I mentioned a wrong instruction. I meant rdgsbase and wrgsbase.

Any how any of these help? What would you write into `gs`?

> Yes, this one is interesting, and I'll still need to think how to circumvent this.

I guess the whole point is - do we need to bother with this file-for-socket at all? Because if `wine` won't work anyway, then it seems pointless to me...
Otherwise we can continue with your latest idea discussed above (which I guess would involve both the new manifest entry and some `/tmp` files).

dimakuv:
Superseded by https://github.com/gramineproject/gramine/pull/1021. Closing.

PR #978, opened 2022-10-13
Python does not have distlib path in CentOS/RHEL.

anjalirai-intel:
## Description of the changes 

To support Python for CentOS, RHEL & Ubuntu from Gramine, we need to replace distlib with numpy, scipy & six libraries path, because distlib path does not exist in CentOS & RHEL

Fixes #973. See it for context.

## How to test this PR? 
1. Make Python example using make SGX=1 on CentOS/RHEL
2. make SGX=1 check

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/978)

[rhel_python3_sysconfig.txt](https://github.com/gramineproject/gramine/files/9798424/rhel_python3_sysconfig.txt)
[centos_python3_sysconfig.txt](https://github.com/gramineproject/gramine/files/9798426/centos_python3_sysconfig.txt)

PR #999, opened 2022-10-20
gramine-manifest: API to list python paths

woju:
## Description of the changes 

Because of differences between distributions, specifically filesystem layout and content of `sys.path`, it's hard to write a manifest that would run Python apps correctly on different distributions. This is an attempt to fix this.

This PR supersedes ~#977 and~ #978.

Closes #978.

## How to test this PR? 

`CI-Examples/python` is modified — run in CI or manually

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/999)

### DIFF
--- a/CI-Examples/python/python.manifest.template
+++ b/CI-Examples/python/python.manifest.template
@@ -15,8 +15,9 @@ fs.mounts = [
   { path = "/lib", uri = "file:{{ gramine.runtimedir() }}" },
   { path = "{{ arch_libdir }}", uri = "file:{{ arch_libdir }}" },
   { path = "/usr/{{ arch_libdir }}", uri = "file:/usr/{{ arch_libdir }}" },
-  { path = "{{ python.stdlib }}", uri = "file:{{ python.stdlib }}" },
-  { path = "{{ python.distlib }}", uri = "file:{{ python.distlib }}" },
+{% for path in python.get_sys_path(entrypoint) %}
+  { path = "{{ path }}", uri = "file:{{ path }}" },
+{% endfor %}
   { path = "{{ entrypoint }}", uri = "file:{{ entrypoint }}" },
   { path = "/etc/hosts", uri = "file:helper-files/hosts" },
 
@@ -41,8 +42,9 @@ sgx.trusted_files = [
   "file:{{ gramine.runtimedir() }}/",
   "file:{{ arch_libdir }}/",
   "file:/usr/{{ arch_libdir }}/",
-  "file:{{ python.stdlib }}/",
-  "file:{{ python.distlib }}/",
+{% for path in python.get_sys_path(entrypoint) %}
+  "file:{{ path }}{{ '/' if path.is_dir() else '' }}",
+{% endfor %}
   "file:scripts/",
   "file:helper-files/",
 ]
--- a/python/graminelibos/gen_jinja_env.py
+++ b/python/graminelibos/gen_jinja_env.py
@@ -29,6 +29,15 @@ def ldd(*args):
             ret.add(line[0])
     return sorted(ret)
 
+def python_get_sys_path(interpreter, include_nonexisting=False):
+    for path in subprocess.check_output([interpreter, '-c',
+        '''import sys; print('\\0'.join(path for path in sys.path if path), end='')'''
+    ]).split(b'\0'):
+        path = pathlib.Path(os.fsdecode(path))
+        if not include_nonexisting and not path.exists():
+            continue
+        yield path
+
 def add_globals_from_python(env):
     paths = sysconfig.get_paths()
     env.globals['python'] = {
@@ -48,6 +57,8 @@ def add_globals_from_python(env):
         'get_paths': sysconfig.get_paths,
 
         'implementation': sys.implementation,
+
+        'get_sys_path': python_get_sys_path,
     }
 
 class Runtimedir:


boryspoplawski (changes requested):
> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/999)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @woju)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/999#-NEq__LZ36ExD-WJrJe_:-NEq__LZ36ExD-WJrJea:b-atd67t) (no related file):*
Not sure about 978, but it does not supersede 977

___
___
*[`CI-Examples/python/python.manifest.template` line 18 at r2](https://reviewable.io/reviews/gramineproject/gramine/999#-NErXAev4zJjzI4X22r4:-NErXAev4zJjzI4X22r5:bn1pzyq) ([raw file](https://github.com/gramineproject/gramine/blob/acceeec095c420b12c6cd89ad4be73add52a08bf/CI-Examples/python/python.manifest.template#L18)):*
> ```
>   { path = "{{ arch_libdir }}", uri = "file:{{ arch_libdir }}" },
>   { path = "/usr/{{ arch_libdir }}", uri = "file:/usr/{{ arch_libdir }}" },
> {% for path in python.get_sys_path(entrypoint) %}
> ```

Shouldn't this be:



_Suggestion:_
```
{% for path in python.get_sys_path(entrypoint, skip_nonexisting=True) %}
```
___
*[`CI-Examples/python/python.manifest.template` line 45 at r2](https://reviewable.io/reviews/gramineproject/gramine/999#-NErXL5xDDwdgsmAMRmG:-NErXL5xDDwdgsmAMRmH:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/acceeec095c420b12c6cd89ad4be73add52a08bf/CI-Examples/python/python.manifest.template#L45)):*
> ```
>   "file:{{ arch_libdir }}/",
>   "file:/usr/{{ arch_libdir }}/",
> {% for path in python.get_sys_path(entrypoint) %}
> ```

ditto
___
*[`CI-Examples/python/python.manifest.template` line 46 at r4](https://reviewable.io/reviews/gramineproject/gramine/999#-NErXpgD0_V38qNgLWYI:-NErXpgD0_V38qNgLWYJ:bqtfkbm) ([raw file](https://github.com/gramineproject/gramine/blob/403dac791951912f8fe04e52703d208822e1179f/CI-Examples/python/python.manifest.template#L46)):*
> ```
>   "file:/usr/{{ arch_libdir }}/",
> {% for path in python.get_sys_path(entrypoint) %}
>   "file:{{ path }}{{ '/' if path.is_dir() else '' }}",
> ```

If you skipped non-existing paths this `if` wouldn't be needed, right?
___
*[`python/graminelibos/gen_jinja_env.py` line 32 at r1](https://reviewable.io/reviews/gramineproject/gramine/999#-NErWbKaBvx5RMzy2qhl:-NErWbKaBvx5RMzy2qhm:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/457288812a5f8cce0e7670ddd94e82456076b2f5/python/graminelibos/gen_jinja_env.py#L32)):*
> ```Python
>     return sorted(ret)
> 
> def python_get_sys_path(interpreter, nonexisting=False):
> ```





_Suggestion:_
```Python
skip_nonexisting
```
___
*[`python/graminelibos/gen_jinja_env.py` line 37 at r4](https://reviewable.io/reviews/gramineproject/gramine/999#-NErXeHI18X2-5hRFXwW:-NErXeHI18X2-5hRFXwX:b-w4mpbc) ([raw file](https://github.com/gramineproject/gramine/blob/403dac791951912f8fe04e52703d208822e1179f/python/graminelibos/gen_jinja_env.py#L37)):*
> ```Python
>     ]).split(b'\0'):
>         path = pathlib.Path(os.fsdecode(path))
>         if not nonexisting and not path.exists():
> ```

What? Why this `not` here? It should not be here, but you should pass `True` in manifest instead

dimakuv (changes requested):
> Reviewed 1 of 1 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/999)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @woju)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/999#-NEq__LZ36ExD-WJrJe_:-NEthWj50T8zFZtdc4LE:bl5i2ux) (no related file):*

Not sure about 978, but it does not supersede 977
</blockquote></details>

Yes, this PR definitely doesn't remove duplicates from `sgx.trusted_files`. Please strike-through it in the PR description.

___
___
*[`CI-Examples/python/python.manifest.template` line 18 at r2](https://reviewable.io/reviews/gramineproject/gramine/999#-NErXAev4zJjzI4X22r4:-NEtfy04AGmFwFoFNy95:b32v57) ([raw file](https://github.com/gramineproject/gramine/blob/acceeec095c420b12c6cd89ad4be73add52a08bf/CI-Examples/python/python.manifest.template#L18)):*

Shouldn't this be:
</blockquote></details>

+1
___
*[`python/graminelibos/gen_jinja_env.py` line 37 at r4](https://reviewable.io/reviews/gramineproject/gramine/999#-NErXeHI18X2-5hRFXwW:-NEtgj2-2awQ7w6vvhWA:b-i5evnc) ([raw file](https://github.com/gramineproject/gramine/blob/403dac791951912f8fe04e52703d208822e1179f/python/graminelibos/gen_jinja_env.py#L37)):*

What? Why this `not` here? It should not be here, but you should pass `True` in manifest instead
</blockquote></details>

+1

Looks like @woju implemented a reverted logic as @boryspoplawski wants/expects. But I agree with Borys that `skip_nonexisting` will be a more user-friendly option.

Alternatively, we can keep @woju's current logic but rename it to a more understandable `list_nonexisting` (or even `list_even_if_nonexisting`)

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/999)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/999#-NEq__LZ36ExD-WJrJe_:-NEtt9mGCF9XeWVmS_C1:b-4o78cb) (no related file):*

Yes, this PR definitely doesn't remove duplicates from `sgx.trusted_files`. Please strike-through it in the PR description.
</blockquote></details>

Like I wrote in a comment there, #977 it's a wrong solution to what is fixed here. But I guess the title doesn't match, so I can strike it.

___
___
*[`CI-Examples/python/python.manifest.template` line 18 at r2](https://reviewable.io/reviews/gramineproject/gramine/999#-NErXAev4zJjzI4X22r4:-NEtx-_TClCWiCCGzqwd:bdikwrb) ([raw file](https://github.com/gramineproject/gramine/blob/acceeec095c420b12c6cd89ad4be73add52a08bf/CI-Examples/python/python.manifest.template#L18)):*

+1
</blockquote></details>

No, it should not. Omitting nonexisting paths should be the default behaviour, because this will be the most frequent (or perhaps the only) use case. So this line is correct irrespective of the discussion about the exact
___
*[`CI-Examples/python/python.manifest.template` line 45 at r2](https://reviewable.io/reviews/gramineproject/gramine/999#-NErXL5xDDwdgsmAMRmG:-NEtxMgv2r0cdSlwnZo4:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/acceeec095c420b12c6cd89ad4be73add52a08bf/CI-Examples/python/python.manifest.template#L45)):*

ditto
</blockquote></details>

ditto
___
*[`CI-Examples/python/python.manifest.template` line 46 at r4](https://reviewable.io/reviews/gramineproject/gramine/999#-NErXpgD0_V38qNgLWYI:-NEtrqh-0m6dux6h-3_M:b-yf6oyn) ([raw file](https://github.com/gramineproject/gramine/blob/403dac791951912f8fe04e52703d208822e1179f/CI-Examples/python/python.manifest.template#L46)):*

If you skipped non-existing paths this `if` wouldn't be needed, right?
</blockquote></details>

No, this is because directories need to end with `/` and regular files not. Some of paths could be zipfiles. Those zipfiles don't usually exist, but the logic here is correct.
___
*[`python/graminelibos/gen_jinja_env.py` line 37 at r4](https://reviewable.io/reviews/gramineproject/gramine/999#-NErXeHI18X2-5hRFXwW:-NEtsPWb8Bdu00_xmQ9C:b-fqdpwo) ([raw file](https://github.com/gramineproject/gramine/blob/403dac791951912f8fe04e52703d208822e1179f/python/graminelibos/gen_jinja_env.py#L37)):*

+1

Looks like @woju implemented a reverted logic as @boryspoplawski wants/expects. But I agree with Borys that `skip_nonexisting` will be a more user-friendly option.

Alternatively, we can keep @woju's current logic but rename it to a more understandable `list_nonexisting` (or even `list_even_if_nonexisting`)
</blockquote></details>

I did it this way to have the default value of False (so if you override, you write `=True`).

`include_nonexisting=`? (there could also be `nonexisting_ok=`, which would conform to `pathlib` style, but I think it's wrong).

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/999)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @woju)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/999#-NEq__LZ36ExD-WJrJe_:-NEtyxvP5n8zSPHO3B-r:b5zdsgl) (no related file):*

Like I wrote in a comment there, #977 it's a wrong solution to what is fixed here. But I guess the title doesn't match, so I can strike it.
</blockquote></details>

I see your point. What you mean is that #977 will become irrelevant after this PR (because there will be no more duplicated Python paths for sure).

___
___
*[`python/graminelibos/gen_jinja_env.py` line 37 at r4](https://reviewable.io/reviews/gramineproject/gramine/999#-NErXeHI18X2-5hRFXwW:-NEtzBHq6ZwPMSh65Jaf:b1jvy6g) ([raw file](https://github.com/gramineproject/gramine/blob/403dac791951912f8fe04e52703d208822e1179f/python/graminelibos/gen_jinja_env.py#L37)):*

I did it this way to have the default value of False (so if you override, you write `=True`).

`include_nonexisting=`? (there could also be `nonexisting_ok=`, which would conform to `pathlib` style, but I think it's wrong).
</blockquote></details>

Ok, Woju has a point. But then let's rename to `include_nonexisting`, good name.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/999)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @woju)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/999#-NEq__LZ36ExD-WJrJe_:-NEuDiP61w-tx6SZSci6:bru9ars) (no related file):*

I see your point. What you mean is that #977 will become irrelevant after this PR (because there will be no more duplicated Python paths for sure).
</blockquote></details>

That's not true, python paths can collide with other paths already in the manifest.

___
___
*[`python/graminelibos/gen_jinja_env.py` line 37 at r4](https://reviewable.io/reviews/gramineproject/gramine/999#-NErXeHI18X2-5hRFXwW:-NEuE0icAc2ZDL3Sr0Zd:b-yc3yui) ([raw file](https://github.com/gramineproject/gramine/blob/403dac791951912f8fe04e52703d208822e1179f/python/graminelibos/gen_jinja_env.py#L37)):*

Ok, Woju has a point. But then let's rename to `include_nonexisting`, good name.
</blockquote></details>

`include_nonexisting` is fine and `skip_nonexisting=True` would be fine as well

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/999)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @woju)

___
*[`CI-Examples/python/python.manifest.template` line 18 at r2](https://reviewable.io/reviews/gramineproject/gramine/999#-NErXAev4zJjzI4X22r4:-NEuEbzH-8IBufqPq4pV:bc8l2j1) ([raw file](https://github.com/gramineproject/gramine/blob/acceeec095c420b12c6cd89ad4be73add52a08bf/CI-Examples/python/python.manifest.template#L18)):*

No, it should not. Omitting nonexisting paths should be the default behaviour, because this will be the most frequent (or perhaps the only) use case. So this line is correct irrespective of the discussion about the exact
</blockquote></details>

When I've put this comment ommiting wasn't the default (in 1st commit)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/999)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @woju)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/999#-NEq__LZ36ExD-WJrJe_:-NEuHqNN8tCB_udXrzg-:b7629b9) (no related file):*
> That's not true, python paths can collide with other paths already in the manifest.

You mean, because Python paths may be subdirs of other paths? Yeah...

___

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/999)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @woju)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/999#-NEq__LZ36ExD-WJrJe_:-NEuQoi7DV1b5m90tCQD:b-h5yzit) (no related file):*

> That's not true, python paths can collide with other paths already in the manifest.

You mean, because Python paths may be subdirs of other paths? Yeah...
</blockquote></details>

Most likely other way round, but yes

___

boryspoplawski (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/999)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/999)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/999#-NEq__LZ36ExD-WJrJe_:-NEuSZpvCgbODiEL6BkW:be2dvw6) (no related file):*

Most likely other way round, but yes
</blockquote></details>

I argue if those paths collide, then something in manifest and/or template is wrong, and there are simple ways around it.

For example, if you have `/usr/` in trusted files (a stupid move, but people do it), then you can:

```jinja
sgx.trusted_files = [
  "file:/usr/",
{% for path in python.get_sys_path(entrypoint) if not path.is_relative_to('/usr') %}
  "file:{{ path }}{{ '/' if path.is_dir() else '' }}",
{% endfor %}
  # ...
]
```

You just need to think a bit, which is the whole point. Without this, people will start dumping everything in trusted_files without thinking. Explicit is better than implicit.

___
___
*[`python/graminelibos/gen_jinja_env.py` line 37 at r4](https://reviewable.io/reviews/gramineproject/gramine/999#-NErXeHI18X2-5hRFXwW:-NEuSY0M33kWhVK8sKxi:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/403dac791951912f8fe04e52703d208822e1179f/python/graminelibos/gen_jinja_env.py#L37)):*

`include_nonexisting` is fine and `skip_nonexisting=True` would be fine as well
</blockquote></details>

Done.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/999)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @woju)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/999#-NEq__LZ36ExD-WJrJe_:-NEuaMyD8Wv1k2An3pd1:b-u6zltj) (no related file):*

I argue if those paths collide, then something in manifest and/or template is wrong, and there are simple ways around it.

For example, if you have `/usr/` in trusted files (a stupid move, but people do it), then you can:

```jinja
sgx.trusted_files = [
  "file:/usr/",
{% for path in python.get_sys_path(entrypoint) if not path.is_relative_to('/usr') %}
  "file:{{ path }}{{ '/' if path.is_dir() else '' }}",
{% endfor %}
  # ...
]
```

You just need to think a bit, which is the whole point. Without this, people will start dumping everything in trusted_files without thinking. Explicit is better than implicit.
</blockquote></details>

This snippet looks cool, an interesting trick (bar the built-in `is_relative_to()` method which is only available in Python 3.9+).

Is there a way to replace `is_relative_to()` with smth that works on Python 3.6? Then we could include this trick here, in `python.manifest.template`. (To exclude subdirs of `{{arch_libdir}}` and `/usr/{{arch_libdir}}`)

___

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/999)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/999#-NEq__LZ36ExD-WJrJe_:-NEudZ_h67dazC1EtCWn:b-npkpg6) (no related file):*

This snippet looks cool, an interesting trick (bar the built-in `is_relative_to()` method which is only available in Python 3.9+).

Is there a way to replace `is_relative_to()` with smth that works on Python 3.6? Then we could include this trick here, in `python.manifest.template`. (To exclude subdirs of `{{arch_libdir}}` and `/usr/{{arch_libdir}}`)
</blockquote></details>

- right now this works: `(path|string).startswith('/usr')`, but I don't know if it won't break on non-unicode paths (it's possible it would work, but I don't know enough about pathlib to say for sure)
- the correct thing would be `fsdecode(path).startswith('/usr'), but we'd need to export `os.fsdecode` function to jinja env (I have nothing against it)
- there's also https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.relative_to, but if it's not relative, you need to catch an exception, so we'd clearly need to write a wrapper

___

mkow (approved):
> Reviewed 1 of 1 files at r3, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/999)* status: all files reviewed, 2 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @woju)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/999#-NEq__LZ36ExD-WJrJe_:-NEv87WjA1kh6wTJBNnz:bnhr1e6) (no related file):*
> `/usr/` in trusted files (a stupid move, but people do it)

> Without this, people will start dumping everything in trusted_files without thinking.

Why? It's a completely valid approach, it removes the need to estimate dynamic dependencies of your workload, which you may do wrong and notice only after long time running on production.

___
___
*[`-- commits` line 7 at r5](https://reviewable.io/reviews/gramineproject/gramine/999#-NEv9eiE2hPGErsUab3y):*
This should be just one commit.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/999)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @woju)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/999)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`-- commits` line 7 at r5](https://reviewable.io/reviews/gramineproject/gramine/999#-NEv9eiE2hPGErsUab3y):*

This should be just one commit.
</blockquote></details>

@mkow What do you mean exactly? Do you want to merge two commits into one? Why?

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/999)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`-- commits` line 7 at r5](https://reviewable.io/reviews/gramineproject/gramine/999#-NEv9eiE2hPGErsUab3y):*

@mkow What do you mean exactly? Do you want to merge two commits into one? Why?
</blockquote></details>

Yes - it's the same change and the first commit alone doesn't make much sense, because it adds dead code in the current version.

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/999)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`-- commits` line 7 at r5](https://reviewable.io/reviews/gramineproject/gramine/999#-NEv9eiE2hPGErsUab3y):*

Yes - it's the same change and the first commit alone doesn't make much sense, because it adds dead code in the current version.
</blockquote></details>

Ok, we can merge them into one during final rebase.

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/999)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @mkow)

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/999)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/999)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #1006, opened 2022-10-24
[DRAFT][RFC][LibOS,PAL] Add naive enclave dynamic memory management(EDMM) support

vijaydhanraj:
## Description of the changes 
This PR adds EDMM support but is still WIP. Creating this PR to get early feedback from maintainers.

## How to test this PR? 
To test this PR, add `sgx.edmm_enable_heap = true` as part of the manifest. By default, this syntax is turned off. Please note the current status is that all PAL regression tests pass but LibOS regression tests are failing. Debug in progress.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1006)


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1006)* status: 0 of 47 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @vijaydhanraj)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1006#-NFEO2g89xnKECa08ApO:-NFEO2g89xnKECa08ApP:bcy2c0b) (no related file):*
I looked into the PR briefly. Main points:
1. EDMM hardware ABI and subsequently the kernel driver API distinguish between "relaxing" page permissions and "restricting" them:
    - relaxing is done via `mprotect()` on the host;
    - restricting is done via `ioctl(SGX_IOC_ENCLAVE_RESTRICT_PERMISSIONS)` on the host.
2. To understand which API to use, Gramine-SGX needs to keep current page permissions for each memory region (aka VMA).
    - This is already done via `libos_vma_info::prot`.
    - I'm unsure why Vijay adds `cur_prot` and `prev_prot` fields instead of that single `prot`...
3.  `PalVirtualMemoryProtect()` needs to be expanded with a new argument -- current page permissions.
    - Note that in the latest Gramine code, there is no memory management in PAL, so there is no place to "keep" page permissions.
    - So looks like there is no way to hide this pecularity from LibOS code.
4. I don't understand the need for `merge_vmas()` in the LibOS VMA code. Is it just an optimization? If yes, then it is surely premature and should be removed...
5. The need for `is_allocated` boolean and the associated trick is not clear to me.
    - Looks like some file-backed mappings are allocated twice? Then maybe we should fix the root cause, not patch over.
6. Changes in the PAL layer generally make sense to me.

___

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1006)* status: 0 of 47 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1006#-NFEO2g89xnKECa08ApO:-NFGcqiC2uyA8IxyzoXa:b-58t7ei) (no related file):*
Thanks @dimakuv for the feedback.

> I'm unsure why Vijay adds cur_prot and prev_prot fields instead of that single prot...

`libos_syscall_mprotect` does two things, 1) bkeeps the new page permissions by calling `bkeep_mprotect` which overwrites the previous page permission. (After this `bkeep_mprotect`, if we dump VMA, it will only reflect new page permission) 2) Invokes `PalVirtualMemoryProtect` to apply the new page permission. Currently step 2 is a dummy function. 

But with EDMM, we need to pass the previous and current page permission to `PAL` so that it can decide if the permission needs to be relaxed or restricted. So modified `bkeep_mprotect` to store the previous page permission (`prev_prot`) in addition to the new request page permission (`cur_prot`) which can be passed to `PalVirtualMemoryProtect`.

> I don't understand the need for merge_vmas() in the LibOS VMA code. Is it just an optimization? If yes, then it is surely premature and should be removed...

This is used to identify holes, that is, unallocated (not `EAUG`ed and `EACCEPT`ed) memory regions within the requested range. Something like below esp. for `mmap` with `MAP_FIXED` flag case. Only for the unallocated memory `PalVirtualMemoryAlloc` is invoked and for other regions, just the page permission is changed by calling `PalVirtualMemoryProtect`.
```
+-------+ Requested addr end
| vma3  |
|       |
+-------+
| vma2  |
|       |
+-------+
|XXXXXXX|  unallocated
+-------+
| vma1  |
+-------+ Requested addr start
```

And yes, as an optimization, as we parse the VMAs we also merge them if they have the same characteristics like if they are unallocated or have the same page permissions.

> The need for is_allocated boolean and the associated trick is not clear to me.

There are certain VMAs (`VMA_UNMAPPED`) that are used only for bookkeeping purposes, but memory is not allocated. To differentiate between such kind of VMA and regular VMA, I have introduced `is_allocated` flag. `PalVirtualMemoryAlloc` is invoked when `is_allocated` is false and `PalVirtualMemoryProtect` otherwise. One example is `reserve_dyn` function where `VMA_UNMAPPED` is used.

___

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1006)* status: 0 of 47 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1006#-NFEO2g89xnKECa08ApO:-NFtXbZ87aZltFy_k6z2:b-g10kqw) (no related file):*

Thanks @dimakuv for the feedback.

> I'm unsure why Vijay adds cur_prot and prev_prot fields instead of that single prot...

`libos_syscall_mprotect` does two things, 1) bkeeps the new page permissions by calling `bkeep_mprotect` which overwrites the previous page permission. (After this `bkeep_mprotect`, if we dump VMA, it will only reflect new page permission) 2) Invokes `PalVirtualMemoryProtect` to apply the new page permission. Currently step 2 is a dummy function. 

But with EDMM, we need to pass the previous and current page permission to `PAL` so that it can decide if the permission needs to be relaxed or restricted. So modified `bkeep_mprotect` to store the previous page permission (`prev_prot`) in addition to the new request page permission (`cur_prot`) which can be passed to `PalVirtualMemoryProtect`.

> I don't understand the need for merge_vmas() in the LibOS VMA code. Is it just an optimization? If yes, then it is surely premature and should be removed...

This is used to identify holes, that is, unallocated (not `EAUG`ed and `EACCEPT`ed) memory regions within the requested range. Something like below esp. for `mmap` with `MAP_FIXED` flag case. Only for the unallocated memory `PalVirtualMemoryAlloc` is invoked and for other regions, just the page permission is changed by calling `PalVirtualMemoryProtect`.
```
+-------+ Requested addr end
| vma3  |
|       |
+-------+
| vma2  |
|       |
+-------+
|XXXXXXX|  unallocated
+-------+
| vma1  |
+-------+ Requested addr start
```

And yes, as an optimization, as we parse the VMAs we also merge them if they have the same characteristics like if they are unallocated or have the same page permissions.

> The need for is_allocated boolean and the associated trick is not clear to me.

There are certain VMAs (`VMA_UNMAPPED`) that are used only for bookkeeping purposes, but memory is not allocated. To differentiate between such kind of VMA and regular VMA, I have introduced `is_allocated` flag. `PalVirtualMemoryAlloc` is invoked when `is_allocated` is false and `PalVirtualMemoryProtect` otherwise. One example is `reserve_dyn` function where `VMA_UNMAPPED` is used.



</blockquote></details>

1. Adding `edmm` specific arguments to `mmap` callback is definitely wrong.
2. `EDMM` is SGX PAL feature, LibOS must not be aware of. It might require a lot of changes, that by itself is fine, but they should not be EDMM specific. I didn't look closely at current changes, so idk if they are generic just with wrong names, but I suspect they are not
3. I dont see how merging VMAs is needed. It doesn't change anything, only complicates stuff.

___

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1006)* status: 0 of 47 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1006#-NFEO2g89xnKECa08ApO:-NFxjpFK5FSUkHNlICGj:ba75uev) (no related file):*
Thanks @boryspoplawski for the feedback.

> 1. Adding edmm specific arguments to mmap callback is definitely wrong.

I think you are referring to this `int (*libos_fs_ops::mmap)(struct libos_handle *hdl, void *addr, size_t size, int prot, int flags, uint64_t offset, struct edmm_heap_request *vma_range) `. If so, this is not EDMM specific but just VMA ranges and the count of such ranges. 

I named the struct as `struct edmm_heap_request` maybe this is misleading? Please see below these have nothing to do with EDMM.
```
struct edmm_heap_vma {
    void* addr;
    size_t length;
    int cur_prot;  // current memory protection flags: PROT_*
    int prev_prot;
    bool is_allocated;
};

/* To pass along with `bkeep_mmap_fixed` in case of MAP_FIXED to get overlapping VMA ranges.
 * Note: Adjacent VMA ranges with same `PROT_*` are merged. */
struct edmm_heap_request {
    int range_cnt;
    struct edmm_heap_vma vma[DEFAULT_VMA_COUNT];
};
```

> 2. EDMM is SGX PAL feature, LibOS must not be aware of. It might require a lot of changes, that by itself is fine, but they should not be EDMM specific. I didn't look closely at current changes, so idk if they are generic just with wrong names, but I suspect they are not

You are right, as part of this PR I have added some logic that makes the LibOS aware of EDMM. I did this esp. for mmap with `MAP-FIXED` flag case because for SGX1 (no EDMM), we can simply alloc but for EDMM case, we cannot accept an already `EACCEPT`ed page. Any suggestions on how to tackle this?

> 3. I dont see how merging VMAs is needed. It doesn't change anything, only complicates stuff.

Please see my response to @dimakuv he also had the same question. Maybe the name is misleading. The basic idea is to parse the VMA regions that falls within a requested range and to identify if there are any unallocated (not `EAUG`ed and `EACCEPT`ed) memory region.

___

vijaydhanraj (commented):
Current Status: All LibOS/PAL regression tests pass, and the PR is ready for evaluation with workloads.

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1006#-:-NHzgeCt-xbhW35Ld3t1:br5ol9d)* status: 0 of 53 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv)

mkow:
Closing, as #1054 was merged which supersedes it.

PR #1021, opened 2022-10-31
[LibOS] Bind named UNIX domain sockets to FS pathnames

dimakuv:
## Description of the changes 

Named UNIX domain sockets are expected to be backed by files. Previously, Gramine didn't have such backing files. However, some workloads require this, e.g., two subprocesses spawned in the same Gramine instance and communicating over a named UNIX socket.

This commit adds pseudo-files for named UNIX sockets (there are no real files created on the host). These pseudo-files are looked up using `socket_builtin_fs`. The `stat()` operation is also implemented.

This design also fixes a bug with relative paths in UNIX sockets' names: previously Gramine would refuse to connect to a socket which name resolves to the same absolute path, but the relative name is different.

This is a *breaking change*: named UNIX sockets must be explicitly specified in the manifest file as `fs.mounts` entries. This requirement fixes a bug when Gramine allowed to create UNIX sockets with arbitrary names, even those that did not map to any mounted directories.

## How to test this PR? 

I modified the `unix` LibOS regression test.


donporter (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1021)* status: all files reviewed, 1 unresolved discussion (waiting on @dimakuv)

___
*[`libos/src/fs/socket/fs.c` line 23 at r1](https://reviewable.io/reviews/gramineproject/gramine/1021#-NFiWRRu2M0YD1HnbctB:-NL1WDVY4I4SxsZbLkQd:boq1chq) ([raw file](https://github.com/gramineproject/gramine/blob/28a9e4593554d6d116405af374dfd43f156f0364/libos/src/fs/socket/fs.c#L23)):*

This is a hack. I'm not sure if we want the legacy logic or not...

If we remove legacy logic, some scenarios will be impossible to support because the manifest must hard-code the socket name. But some apps create named sockets with random names, e.g. LTP's `bind03` test.

If we keep legacy logic, then it will not be a "clean" implementation. But we'll preserve compatibility with old manifests & apps.

Of course, instead of a "side channel" of peeking into `fs.mounts = { type="socket" }`, I could introduce a new true/false manifest option. But I didn't want to introduce yet another super-specific keyword in the manifest.
</blockquote></details>

I am not sure I understand the argument for legacy logic, long-term.  I tend to think we should mirror linux/posix as much as we can.

Perhaps keep it until the next release and announce a sunset period (i.e., drop it 2 releases from now)?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1021)* status: 2 of 6 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @donporter)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1021#-NavKbqe4XHjhgJPaozW:-NavKbqe4XHjhgJPaozX:b-i23z3q) (no related file):*
I rebased this PR, because it started to get stale.

I still don't like this solution for two reasons:
1. Hack to keep legacy behavior.
2. A need to specify an exact UNIX socket name in the manifest, which disallows a typical case of randomly-named UNIX sockets.

I guess it would make sense to just always create/lookup dentries for named UNIX sockets and drop the legacy behavior... I should probably revive this effort with a series of PRs that have a less hacky approach as in this PR.

___

donporter (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1021)* status: all files reviewed, 2 unresolved discussions (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1021#-NavKbqe4XHjhgJPaozW:-Nb-xWpd3e44639EAiTy:b29p41v) (no related file):*

I rebased this PR, because it started to get stale.

I still don't like this solution for two reasons:
1. Hack to keep legacy behavior.
2. A need to specify an exact UNIX socket name in the manifest, which disallows a typical case of randomly-named UNIX sockets.

I guess it would make sense to just always create/lookup dentries for named UNIX sockets and drop the legacy behavior... I should probably revive this effort with a series of PRs that have a less hacky approach as in this PR.
</blockquote></details>

Yeah, fair points.  

1) In the long run, I would support non-standard "personalities" when there is a good reason not to do things the Unix way
2) This is a good case to support in the long run, but totally agree that keeping random names in sync across liboses is going to be challenging.

___

PR #1065, opened 2022-11-28
[WIP] Project Amber client with KBS support

bigdata-memory:
This patch upgrades the project Amber (alpha) integration code from Gramine v1.2 to v1.3.1, interfaces with project Amber (beta) and adds KBS support.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1065)


yzz127:
With Ubuntu 20.04 and gcc 9.4, there is a compilation error complaining api_fortified.h. The error could be fixed with #define USE_STDLIB in https.h header file. However, please check and make sure this changes make sense and doesn't break expected logics. Thanks!

bigdata-memory:
> With Ubuntu 20.04 and gcc 9.4, there is a compilation error complaining api_fortified.h. The error could be fixed with #define USE_STDLIB in https.h header file. However, please check and make sure this changes make sense and doesn't break expected logics. Thanks!

@yzz127, Thank you so much. I will update this PR according to your patch fix. Thanks.

mssun:
Is KBS-related settings in the manifest required? Maybe not all users needs the secret provisioning provided by Project Amber. I suggest to make the KBS optional.

bigdata-memory:
> Is KBS-related settings in the manifest required? Maybe not all users needs the secret provisioning provided by Project Amber. I suggest to make the KBS optional.

@mssun Thank you for your suggestion. You are right, It is required for manifest settings, but not be used if the workload doesn't open & read "/dev/amber/secret". I will make it optional for manifest loading, thanks.

mssun:
Note that `sgx.remote_attestation = "dcap"` should be explicitly set in the manifest file to use Amber for attestation. The default value is none if not set. It's better to check whether both `sgx.remote_attestation` and `sgx.amber_*` are set when compiling the manifest template file.

mkow:
Seems to be a duplicate of #1083, closing.
@bigdata-memory: please explain if there's a reason why would we keep both PRs open.

PR #1132, opened 2023-01-19
Enable unix domain socket between two different gramine instances

ying2liu:
Currently gramine only supports unix domain socket between child/parent within the same gramine instance. Some cloud use cases need unix domain socket between two different gramine instances. Add this support based on customers' requirement.

Signed-off-by: Ying Liu <ying2.liu@intel.com>

## How to test this PR? 

- - -
- 
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1132)

[uds_test.zip](https://github.com/gramineproject/gramine/files/10453853/uds_test.zip)

### DIFF
--- a/libos/src/ipc/libos_ipc.c
+++ b/libos/src/ipc/libos_ipc.c
@@ -113,12 +113,19 @@ static int ipc_connect(IDTYPE dest, struct libos_ipc_connection** conn_ptr) {
         }
 
         char uri[PIPE_URI_SIZE];
+        char  uri_new[PIPE_URI_SIZE];
         if (vmid_to_uri(dest, uri, sizeof(uri)) < 0) {
             log_error("buffer for IPC pipe URI too small");
             BUG();
         }
+
+        uint64_t len = snprintf(uri_new, sizeof(uri_new), "%s/%lu", uri, g_pal_public_state->instance_id);
+        if (len >= sizeof(uri_new)) {
+                return -ERANGE;
+        }
+
         do {
-            ret = PalStreamOpen(uri, PAL_ACCESS_RDONLY, /*share_flags=*/0, PAL_CREATE_IGNORED,
+            ret = PalStreamOpen(uri_new, PAL_ACCESS_RDONLY, /*share_flags=*/0, PAL_CREATE_IGNORED,
                                 /*options=*/0, &conn->handle);
         } while (ret == -PAL_ERROR_INTERRUPTED);
         if (ret < 0) {
--- a/libos/src/libos_init.c
+++ b/libos/src/libos_init.c
@@ -554,7 +554,7 @@ int create_pipe(char* name, char* uri, size_t size, PAL_HANDLE* hdl, bool use_vm
 
     while (true) {
         if (use_vmid_for_name) {
-            len = snprintf(pipename, sizeof(pipename), "%u", g_process_ipc_ids.self_vmid);
+            len = snprintf(pipename, sizeof(pipename), "%u/%lu",  g_process_ipc_ids.self_vmid, g_pal_public_state->instance_id);
             if (len >= sizeof(pipename))
                 return -ERANGE;
         } else {
--- a/pal/include/host/linux-common/linux_utils.h
+++ b/pal/include/host/linux-common/linux_utils.h
@@ -45,8 +45,7 @@ void time_get_now_plus_ns(struct timespec* ts, uint64_t addend_ns);
  * can be negative! */
 int64_t time_ns_diff_from_now(struct timespec* ts);
 
-int get_gramine_unix_socket_addr(uint64_t instance_id, const char* name,
-                                 struct sockaddr_un* out_addr);
+int get_gramine_unix_socket_addr(const char* name, struct sockaddr_un* out_addr);
 
 int file_stat_type(struct stat* stat);
 
--- a/pal/include/pal/pal.h
+++ b/pal/include/pal/pal.h
@@ -174,6 +174,7 @@ struct pal_public_state {
 
     bool extra_runtime_domain_names_conf;
     struct pal_dns_host_conf dns_host;
+    uint64_t instance_id;
 };
 
 /* We cannot mark this as returning a pointer to `const` object, because LibOS can
--- a/pal/src/host/linux-common/gramine_unix_socket_addr.c
+++ b/pal/src/host/linux-common/gramine_unix_socket_addr.c
@@ -8,7 +8,7 @@
 #include "api.h"
 #include "linux_utils.h"
 
-int get_gramine_unix_socket_addr(uint64_t instance_id, const char* name,
+int get_gramine_unix_socket_addr(const char* name,
                                   struct sockaddr_un* out_addr) {
     /* Apparently there is no way to get this define without including whole "sys/socket.h". */
     out_addr->sun_family = /*AF_UNIX*/1;
@@ -17,8 +17,7 @@ int get_gramine_unix_socket_addr(uint64_t instance_id, const char* name,
      * buffer (i.e. do *not* consider the rest of the buffer as null-terminated C string, but rather
      * a fixed length byte array). */
     memset(out_addr->sun_path, 0, sizeof(out_addr->sun_path));
-    int ret = snprintf(out_addr->sun_path + 1, sizeof(out_addr->sun_path) - 1, "/gramine/%lu/%s",
-                       instance_id, name);
+    int ret = snprintf(out_addr->sun_path + 1, sizeof(out_addr->sun_path) - 1, "/gramine/%s", name);
     if (ret < 0) {
         return ret;
     }
--- a/pal/src/host/linux-sgx/pal_main.c
+++ b/pal/src/host/linux-sgx/pal_main.c
@@ -544,6 +544,7 @@ noreturn void pal_linux_main(void* uptr_libpal_uri, size_t libpal_uri_len, void*
                              void* urts_reserved_mem_ranges, size_t urts_reserved_mem_ranges_size) {
     /* All our arguments are coming directly from the host. We are responsible to check them. */
     int ret;
+    sgx_key_128bit_t seal_key;
 
     /* Relocate PAL */
     ret = setup_pal_binary();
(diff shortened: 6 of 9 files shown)


mkbhanda:
Do we need a two tests here, socket between parent & child and two random enclaves?

ying2liu:
@mkbhanda The test was uploaded. It includes both client and server.

ckaratzas:
As I have seen unix domain sockets currently don't have dentry exposed. Does this PR handle this issue or am I missing something?

ying2liu:
@ckaratzas No, this PR doesn't cover dentry. Currently unix domain socket only works between child/parent within the same gramine instance. This PR allows unix domain socket to work between two different gramine instances.

dimakuv:
@ckaratzas Do you mean what this PR achieves: https://github.com/gramineproject/gramine/pull/1021 ? This PR is currently deprioritized, since we're busy working on/reviewing more urgent issues. But I plan to revive this PR soon (it is in a working condition, you can test it).

mkow (changes requested):
> Reviewed 2 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1132)* status: 2 of 9 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @ying2liu)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1132#-NMnAgk51AJReYk-mJCa:-NMnAgk51AJReYk-mJCb:bp4tjcx) (no related file):*
If I read correctly, this approach allows the untrusted host to juggle UDS connections between Gramine processes. We discussed this already a few times on calls, this is not secure and we won't merge such an approach...

___
___
*[`libos/src/ipc/libos_ipc.c` line 116 at r1](https://reviewable.io/reviews/gramineproject/gramine/1132#-NMn9oNxBb1AzRGWRpI_:-NMn9oNxBb1AzRGWRpIa:b204qjl) ([raw file](https://github.com/gramineproject/gramine/blob/627b82a7b6675949672a68de328fce1185aede80/libos/src/ipc/libos_ipc.c#L116)):*
> ```C
> 
>         char uri[PIPE_URI_SIZE];
>         char  uri_new[PIPE_URI_SIZE];
> ```

please fix formatting
___
*[`libos/src/ipc/libos_ipc.c` line 122 at r1](https://reviewable.io/reviews/gramineproject/gramine/1132#-NMn9xp9Ct1sRY4LXCRe:-NMn9xp9Ct1sRY4LXCRf:brpvd2h) ([raw file](https://github.com/gramineproject/gramine/blob/627b82a7b6675949672a68de328fce1185aede80/libos/src/ipc/libos_ipc.c#L122)):*
> ```C
>         }
> 
>         uint64_t len = snprintf(uri_new, sizeof(uri_new), "%s/%lu", uri, g_pal_public_state->instance_id);
> ```

Shouldn't this be part of `vmid_to_uri()`?
___
*[`libos/src/ipc/libos_ipc.c` line 122 at r1](https://reviewable.io/reviews/gramineproject/gramine/1132#-NMnAOz56msHjtHJkTSB:-NMnAOz56msHjtHJkTSC:b-jm2hbe) ([raw file](https://github.com/gramineproject/gramine/blob/627b82a7b6675949672a68de328fce1185aede80/libos/src/ipc/libos_ipc.c#L122)):*
> ```C
>         }
> 
>         uint64_t len = snprintf(uri_new, sizeof(uri_new), "%s/%lu", uri, g_pal_public_state->instance_id);
> ```

`snprintf` doesn't return `uint64_t`
___
*[`pal/src/host/linux-common/gramine_unix_socket_addr.c` line 11 at r1](https://reviewable.io/reviews/gramineproject/gramine/1132#-NMnB6md1wb6Hel7Xpjt:-NMnB6md1wb6Hel7Xpju:b-mun5l7) ([raw file](https://github.com/gramineproject/gramine/blob/627b82a7b6675949672a68de328fce1185aede80/pal/src/host/linux-common/gramine_unix_socket_addr.c#L11)):*
> ```C
> #include "linux_utils.h"
> 
> int get_gramine_unix_socket_addr(const char* name,
> ```

You dropped instance_id from the namespace id, so how does it work now if you start two instances of the same workload in two different Gramine instances? They should be isolated and behave the same as if you ran two Docker containers.

ying2liu:
@mkow There is use case that requires Unix domain socket connection between two different gramine instances. This is the first patch. There will be another patch to address the security concern.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1132)* status: 2 of 9 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @ying2liu)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1132#-NMnAgk51AJReYk-mJCa:-NN-Q3ZgEWSfpgGNbxSH:b-l5t5xg) (no related file):*
> @mkow There is use case that requires Unix domain socket connection between two different gramine instances. This is the first patch. There will be another patch to address the security concern.

@ying2liu: We won't merge something which is not secure, and an existence of a use case for it doesn't justify making Gramine insecure.
Also, again, please use Reviewable when replying to keep the discussions threaded.

___

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1132)* status: all files reviewed, 13 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @ying2liu)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1132#-NMnAgk51AJReYk-mJCa:-NN0dFWDEKfpCTuJmb1W:b7tmnhe) (no related file):*

> @mkow There is use case that requires Unix domain socket connection between two different gramine instances. This is the first patch. There will be another patch to address the security concern.

@ying2liu: We won't merge something which is not secure, and an existence of a use case for it doesn't justify making Gramine insecure.
Also, again, please use Reviewable when replying to keep the discussions threaded.
</blockquote></details>

I discussed with Ying privately, there will be a quite different approach to this. This PR should be closed and Ying will come up with a new PR.

___
___
*[`-- commits` line 6 at r1](https://reviewable.io/reviews/gramineproject/gramine/1132#-NM8PTBB5TahYS4RpRnZ):*
We'll need to change this text completely, giving more technical details and removing `customer requirement`.
___
*[`libos/src/libos_init.c` line 537 at r1](https://reviewable.io/reviews/gramineproject/gramine/1132#-NM8PhGd9-llZSP_o8oi:-NM8PhGd9-llZSP_o8oj:b38vfae) ([raw file](https://github.com/gramineproject/gramine/blob/627b82a7b6675949672a68de328fce1185aede80/libos/src/libos_init.c#L537)):*
> ```C
> }
> 
> int vmid_to_uri(IDTYPE vmid, char* uri, size_t uri_len) {
> ```

This func is used only in one place, so I would make it static and move close to the caller.
___
*[`libos/src/libos_init.c` line 545 at r1](https://reviewable.io/reviews/gramineproject/gramine/1132#-NM8Peeh8YJLYZEsYk48:-NM8Peeh8YJLYZEsYk49:bif1pdt) ([raw file](https://github.com/gramineproject/gramine/blob/627b82a7b6675949672a68de328fce1185aede80/libos/src/libos_init.c#L545)):*
> ```C
> }
> 
> int create_pipe(char* name, char* uri, size_t size, PAL_HANDLE* hdl, bool use_vmid_for_name) {
> ```

Meh, I'd like to refactor and rename this function finally.
___
*[`libos/src/libos_init.c` line 557 at r1](https://reviewable.io/reviews/gramineproject/gramine/1132#-NM8PZ0640gpu7gx3x5o:-NM8PZ07307szagcIinQ:b6iggzn) ([raw file](https://github.com/gramineproject/gramine/blob/627b82a7b6675949672a68de328fce1185aede80/libos/src/libos_init.c#L557)):*
> ```C
>     while (true) {
>         if (use_vmid_for_name) {
>             len = snprintf(pipename, sizeof(pipename), "%u/%lu",  g_process_ipc_ids.self_vmid, g_pal_public_state->instance_id);
> ```

It makes more sense to swap `instance_id` (aka `namespace_id`) and `vmid` places.
___
*[`libos/src/ipc/libos_ipc.c` line 125 at r1](https://reviewable.io/reviews/gramineproject/gramine/1132#-NM8NHcg8LBXQG3TR8Mj:-NM8NHchCw-wWVRayHrC:b-zidkvi) ([raw file](https://github.com/gramineproject/gramine/blob/627b82a7b6675949672a68de328fce1185aede80/libos/src/ipc/libos_ipc.c#L125)):*
> ```C
>         if (len >= sizeof(uri_new)) {
>                 return -ERANGE;
>         }
> ```

This logic should be moved into `vmid_to_uri()` function
___
*[`pal/include/pal/pal.h` line 177 at r1](https://reviewable.io/reviews/gramineproject/gramine/1132#-NM8NkIJDludxeAlX3CE:-NM8NkIJDludxeAlX3CF:b-ddz2xx) ([raw file](https://github.com/gramineproject/gramine/blob/627b82a7b6675949672a68de328fce1185aede80/pal/include/pal/pal.h#L177)):*
> ```objc
>     bool extra_runtime_domain_names_conf;
>     struct pal_dns_host_conf dns_host;
>     uint64_t instance_id;
> ```

Since you move `instance_id` here, you should:
- remove `instance_id` from `struct pal_common_state`
- modify all `instance_id` usages accordingly

Actually, I would also rename `instance_id` to `namespace_id` -- it would make more sense given this PR.
___
*[`pal/src/host/linux-common/gramine_unix_socket_addr.c` line 20 at r1](https://reviewable.io/reviews/gramineproject/gramine/1132#-NM8O6UPA9yWzHNcpb1H:-NM8O6UPA9yWzHNcpb1I:bj0lvpq) ([raw file](https://github.com/gramineproject/gramine/blob/627b82a7b6675949672a68de328fce1185aede80/pal/src/host/linux-common/gramine_unix_socket_addr.c#L20)):*
> ```C
>      * a fixed length byte array). */
>     memset(out_addr->sun_path, 0, sizeof(out_addr->sun_path));
>     int ret = snprintf(out_addr->sun_path + 1, sizeof(out_addr->sun_path) - 1, "/gramine/%s", name);
> ```

FYI: I actually like this change, it makes PAL more generic and moves the decision on "what is a Gramine namespace?" to the LibOS component.
___
*[`pal/src/host/linux-sgx/pal_main.c` line 547 at r1](https://reviewable.io/reviews/gramineproject/gramine/1132#-NM8OFkBEEC9g5TUASrD:-NM8OFkBEEC9g5TUASrE:b-tk9m29) ([raw file](https://github.com/gramineproject/gramine/blob/627b82a7b6675949672a68de328fce1185aede80/pal/src/host/linux-sgx/pal_main.c#L547)):*
> ```C
>     /* All our arguments are coming directly from the host. We are responsible to check them. */
>     int ret;
>     sgx_key_128bit_t seal_key;
> ```

Please move this declaration close to the usage.
___
*[`pal/src/host/linux-sgx/pal_main.c` line 665 at r1](https://reviewable.io/reviews/gramineproject/gramine/1132#-NM8OZHI5YFOM0Xkj6yF:-NM8OZHI5YFOM0Xkj6yG:b-42seb7) ([raw file](https://github.com/gramineproject/gramine/blob/627b82a7b6675949672a68de328fce1185aede80/pal/src/host/linux-sgx/pal_main.c#L665)):*
> ```C
>     ret = lib_HKDF_SHA256((uint8_t*)&seal_key, sizeof(seal_key), /*salt=*/NULL,
>                            /*salt_size=*/0, NULL, 0,
>                            (uint8_t*)&g_master_key, sizeof(g_master_key));
> ```

This should be behind some new manifest option -- we should have the ability to use the old logic (with totally random pipe-encryption key, i.e., "one Gramine instance = one namespace") and use the new logic (with MRSIGNER-derived pipe-encryption key, i.e. "several Gramine instances with same MRSIGNER = one namespace").

Actually, there should be something more than just MRSIGNER thrown into the mix... Well, I think `sgx.isvprodid` manifest option (which is already accounted for in getting seal key) is good for this.

PR #1144, opened 2023-01-24
Add utimensat system call

ying2liu:
Signed-off-by: Ying Liu <ying2.liu@intel.com>

## Description of the changes 

Add utimensat system call.

## How to test this PR? 
[utimenat.tar.gz](https://github.com/gramineproject/gramine/files/10487278/utimenat.tar.gz)

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1144)


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1144)* status: 0 of 13 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @ying2liu)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1144#-NMYAaks-uYCk9oQJ1kH:-NMYAaks-uYCk9oQJ1kI:b-a2bnfl) (no related file):*
What is the usage of `utimensat()` syscall in your application? Why exactly is it needed by the app? What happens if files' timestamps are **not** updated?

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1144#-NMYAj2G9algT2l7VlH6:-NMYAj2HDIWujB3VmlnG:b-wgcz8s) (no related file):*
This solution is partial: we want to add the whole family of "modify file timestamp" syscalls:
- `utime`
- `utimes`
- `futimesat`
- `utimensat`

And this family of syscalls must be implemented by single logic.

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1144#-NMYB0pKAMBhRCYKwBFS:-NMYB0pKAMBhRCYKwBFT:bcge1k6) (no related file):*
Why do you need to passthrough to the host? Could you just emulate this functionality inside of Gramine itself? Do you really need to reflect timestamps on the host?

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1144#-NMYBI0f12Xzho6DKOXy:-NMYBI0f12Xzho6DKOXz:bqbi8qm) (no related file):*
This implementation is wrong: it completely bypasses Gramine's file system component. In reality, the timestamps must be updated on the LibOS inode object, which means that we first need to lookup the file (dentry), and then get the corresponding inode.

There is a reason why we call Gramine a "Library OS" and not a "passthrough everything to the host, what can go wrong".

___

ying2liu:
@dimakuv 
When image-rs uses tar-rs crate to unpack the image, it needs preserve the timestamps for regular files, link files and directories, which is the reason why these three system calls are needed. We discussed about utimensat system call during a community meeting. I thought we had agreed to use pass-through to implement this system call. If you agree, I could also add other utime related system calls.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1144)* status: 0 of 13 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @ying2liu)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1144#-NMYAaks-uYCk9oQJ1kH:-NMhCyST-V9xyi3SCGYF:brnav0s) (no related file):*
> When image-rs uses tar-rs crate to unpack the image, it needs preserve the timestamps for regular files, link files and directories, which is the reason why these three system calls are needed. 

So what happens when the timestamps are not preserved/updated? How does the application fail?

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1144#-NMYBI0f12Xzho6DKOXy:-NMhD6Tl1Xs9P-jNn5S7:bj2il4e) (no related file):*
> I thought we had agreed to use pass-through to implement this system call.

We could, but clearly not in the way you did it now.

I mean, you're literally passing through `dirfd` that is one number (FD) inside Gramine and a completely different number on the host -- how can this work? Similarly, you're passing through `pathname` which may mean one file inside Gramine and mean a completely different file on the host.

This PR simply doesn't work at all. How could it work in your tests?

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1144#-NMYAj2G9algT2l7VlH6:-NMhDh9C1wNKTndBGB2g:bya4zb7) (no related file):*
> If you agree, I could also add other utime related system calls.

Yes, you should, but first you need to rework the whole logic of this PR.

___

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1144)* status: 0 of 13 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @ying2liu)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1144#-NMYBI0f12Xzho6DKOXy:-NMn8PgI0yfvPnxL5qN7:bw1aksu) (no related file):*

> I thought we had agreed to use pass-through to implement this system call.

We could, but clearly not in the way you did it now.

I mean, you're literally passing through `dirfd` that is one number (FD) inside Gramine and a completely different number on the host -- how can this work? Similarly, you're passing through `pathname` which may mean one file inside Gramine and mean a completely different file on the host.

This PR simply doesn't work at all. How could it work in your tests?
</blockquote></details>

I agree with @dimakuv, this PR is completely wrong and it's not something we could merge, so I'm closing it.
@ying2liu: If you want to have this feature then please implement it correctly and submit a new PR. Also, please use Reviewable for the review process.

___

PR #1012, opened 2022-10-26
[PAL/Linux-SGX] Add Invariant TSC query fallback for hypervisors

dimakuv:
## Description of the changes 

Some hypervisors (like QEMU/KVM) do not expose CPUID leaves 0x15 and 0x16 (Core Crystal Clock/Process Frequency). Instead, hypervisor-specific CPUID leaf 0x40000010 shows TSC frequency.

Note that QEMU/VKM must start the VM with CPU flags `+invtsc,+vmware-cpuid-freq` to expose required CPUID leaves.

Fixes #1009.

See also https://lwn.net/Articles/301888/.

## How to test this PR? 

Need a VM. See https://github.com/gramineproject/gramine/issues/1009.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1012)

PR #1154, opened 2023-02-01
[PAL/Linux-SGX] Use `clock_gettime()` from untrusted vDSO

### LINKED ISSUE #1009: Gramine always use ocall_gettime() instead of rdtsc in QEMU VMs

When trying performance tuning in a QEMU VM, I found that Gramine always goes to the slow path to get current time. But in the host environment on the same physical machine, Gramine uses RDTSC.

The root cause is QEMU does not expose CPUID 0x15 and 0x16 to the guest:
```
$ cpuid -1 -l 0x15 -r
CPU:
   0x00000015 0x00: eax=0x00000000 ebx=0x00000000 ecx=0x00000000 edx=0x00000000
$ cpuid -1 -l 0x16 -r
CPU:
   0x00000016 0x00: eax=0x00000000 ebx=0x00000000 ecx=0x00000000 edx=0x00000000
```

According to [this patch](https://patchwork.kernel.org/project/qemu-devel/patch/1484749457-87117-1-git-send-email-phil@philjordan.eu/) the CPU frequency is exposed to the guest from a different CPUID leaf 0x40000010. So, I think we can detect the CPU frequency like this:
- CPUID[0x1].ECX & (1<<31)  we are in a VM.
- CPUID[0x40000000].EAX >= 0x40000010 CPU frequency is reported
- CPUID[0x40000010].EAX * 1000 is the CPU frequency.

Also, when starting the VM, QEMU needs these CPU feature flag "+invtsc,+vmware-cpuid-freq" to report Invariant TSC and CPU frequency through CPUID.

### Steps to reproduce

Run this C program both in the host environment and in the VM:
```
#include <time.h>

int main()
{
        int i;
        for (i = 0; i < 1000*1000; ++i) {
                time(NULL);
        }
        return 0;
}
```

### Expected results

Gramine running inside the VM should have similar performance as what in the host.

### Actual results

On the host, only a few EEXITS are observed:
```
----- Total SGX stats for process 49 -----
  # of EENTERs:        398
  # of EEXITs:         324
  # of AEXs:           2078
  # of sync signals:   70
  # of async signals:  0
```

Inside the VM, a lot of EEXITS are observed:
```
----- Total SGX stats for process 2067 -----
  # of EENTERs:        1000328
  # of EEXITs:         1000287
  # of AEXs:           3669
  # of sync signals:   37
  # of async signals:  0
```

### Gramine commit hash

8b4ebdf

dimakuv:
## Description of the changes 

Previously, Gramine-SGX implemented `PalSystemTimeQuery()` in two ways:
- slow path: perform `gettime()` OCALL;
- fast path: use RDTSC with scaling when "Invariant TSC" hardware feature is available.

"Invariant TSC" feature availability was determined via CPUID leaves. However, these CPUID leaves are not exposed/have dummy values if Gramine runs inside a VM. In particular, MS Azure VMs run on Hyper-V VMM which doesn't expose these CPUID leaves (instead it exposes kernel-only MSR). Thus, Gramine-SGX on MS Azure always took the slow path.

To fix this, this PR implements an untrusted helper thread that periodically calls into `clock_gettime()` function in untrusted vDSO and puts the resulting time value in shared untrusted memory, and enclave threads read this time value. This has the same security properties as previous version, because RDTSC and CPUID leaves can be anyway trapped and controlled by a malicious VMM.

**As a side effect, the time resolution is lowered to around 100ms.**

Fixes #1009. Closes #1012.

## How to test this PR? 

Manually in different VMs from different cloud providers.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1154)


llly (commented):
> Reviewed 12 of 22 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1154)* status: 12 of 22 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`pal/src/host/linux-sgx/host_ocalls.c` line 907 at r1](https://reviewable.io/reviews/gramineproject/gramine/1154#-NNG075G1TvV9QueswP4:-NNG075G1TvV9QueswP5:b5x86gd) ([raw file](https://github.com/gramineproject/gramine/blob/6db58bf09f10f459326a3492f593ce151bd71881/pal/src/host/linux-sgx/host_ocalls.c#L907)):*
> ```C
> bool g_gettime_helper_started = false;
> 
> static int gettime_helper_loop(void* arg) {
> ```

These functions are not ocalls now, can be moved out of `host_ocalls.c`.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1154)* status: 11 of 24 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @llly)

___
*[`pal/regression/Event.c` line 64 at r2](https://reviewable.io/reviews/gramineproject/gramine/1154#-NNGHpdO4Kwk4PrGgG6L:-NNGHpdO4Kwk4PrGgG6M:b-lzfqpx) ([raw file](https://github.com/gramineproject/gramine/blob/72f72cc1536f3e434219c1c190796b84983e5115/pal/regression/Event.c#L64)):*
> ```C
>         CHECK(-1);
>     }
>     if (end - start < TIME_US_IN_S / 2) {
> ```

FYI: Old code slept for 1s and checked for exactly 1s. But with our new design time diff has a jitter of 100ms, we could end up with a time difference of 0.9s here. So this test failed.
___
*[`pal/regression/Misc.c` line 36 at r2](https://reviewable.io/reviews/gramineproject/gramine/1154#-NNGHjJHEN_4_dzLhnIH:-NNGHjJHEN_4_dzLhnII:b-plk370) ([raw file](https://github.com/gramineproject/gramine/blob/5fb315426a3bcc306969d5ace203513265844c38/pal/regression/Misc.c#L36)):*
> ```C
>     }
> 
>     uint64_t timeout = 10000;
> ```

FYI: Old code slept for 10ms, which is less than our new time-value resolution, so this test failed.
___
*[`pal/src/meson.build` line 9 at r2](https://reviewable.io/reviews/gramineproject/gramine/1154#-NNGGPnt4lE64oN-sA4F:-NNGGPnt4lE64oN-sA4G:b-15v4wo) ([raw file](https://github.com/gramineproject/gramine/blob/72f72cc1536f3e434219c1c190796b84983e5115/pal/src/meson.build#L9)):*
> ```Meson
> pal_elf_common_sources = files(
>     'pal_elf.c',
> )
> ```

FYI: This is a new "type" of common sources. The functions from `pal_elf.c` are used in three PAL components:
- Common PAL code (under `pal/src/`)
- Untrusted PAL code (under `pal/src/host/linux/` and `pal/src/host/linux-sgx/host_*.c`)
- Trusted PAL code (under `pal/src/host/linux-sgx/pal_*.c`)

Previously we didn't have a need for this new type. So I introduced it in this PR, and currently limited it to only two ELF functions (the ones I need in this PR).

Technically, this change could be split into a separate PR, but then it won't be clear why we need it...
___
*[`pal/src/host/linux-sgx/host_framework.c` line 536 at r2](https://reviewable.io/reviews/gramineproject/gramine/1154#-NNGHYHlDu2n-NPxWCUw:-NNGHYHlDu2n-NPxWCUx:bt9taxh) ([raw file](https://github.com/gramineproject/gramine/blob/72f72cc1536f3e434219c1c190796b84983e5115/pal/src/host/linux-sgx/host_framework.c#L536)):*
> ```C
> 
>         /* NOTE: This helper thread sleeps for 100ms, which limits the timestamp resolution. This
>          *       duration was chosen to balance timestamp resolution and CPU load. */
> ```

Important note! Unfortunately, no idea how to balance the CPU load of this helper thread in any other way.

See also my other comment on impossibility of an alternative design.
___
*[`pal/src/host/linux-sgx/host_framework.c` line 572 at r2](https://reviewable.io/reviews/gramineproject/gramine/1154#-NNGH390CU_f1GAWfxEb:-NNGH391A9ImH1GXIYlq:b-xsubsi) ([raw file](https://github.com/gramineproject/gramine/blob/72f72cc1536f3e434219c1c190796b84983e5115/pal/src/host/linux-sgx/host_framework.c#L572)):*
> ```C
>      *       must access vvar memory region (which resides in untrusted mem) and call into this code
>      *       from within the enclave. However, vDSO/vvar interactions are internal Linux details and
>      *       don't have stable APIs. */
> ```

Important note!

If anyone knows how to safely and future-proofly re-implement the Linux vDSO + vvar logic inside Gramine, please tell. I failed to find a robust way of doing this.

(You can look at the Linux implementation under `lib/vdso/` in Linux sources.)
___
*[`pal/src/host/linux-sgx/host_ocalls.c` line 907 at r1](https://reviewable.io/reviews/gramineproject/gramine/1154#-NNG075G1TvV9QueswP4:-NNGGFne7N76DCheWqs2:bb6un1j) ([raw file](https://github.com/gramineproject/gramine/blob/6db58bf09f10f459326a3492f593ce151bd71881/pal/src/host/linux-sgx/host_ocalls.c#L907)):*

These functions are not ocalls now, can be moved out of `host_ocalls.c`.
</blockquote></details>

Done

llly (commented):
> Reviewed 10 of 22 files at r1.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1154)* status: 20 of 24 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners

dimakuv:
Jenkins, retest this please (downloading packages failed for some time)

dimakuv:
Jenkins, retest Jenkins-Direct-Sanitizers please (Blender download timed out)

llly (commented):
> Reviewed 3 of 4 files at r2, 1 of 1 files at r3, 5 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1154)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1154)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1154#-NNYkVeu41_GXJU-wrNS:-NNYkVeu41_GXJU-wrNT:b-qaxk3z) (no related file):*
I have mixed feelings about this... It's burning system resources constantly (not much of them though, but even when the app isn't doing anything) regardless of the user workload, just to optimize a specific set of workloads, in specific environments :/ Also, it makes our timers imprecise in the order of 100ms, which is a lot?

Alternatives we have: (please add/comment on these if you have other ideas)
- Implement separate invariant TSC detection for each hypervisor.
- Talk to hypervisor devs and ask why this isn't emulated in their CPUID, maybe they could add it easily.
- Add a "force using TSC for time" manifest switch? But we'd need to think if this has security implications if such an enclave would be run on a system without invariant TSC.

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1154)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1154#-NNYkVeu41_GXJU-wrNS:-NN_sp45AB6QV2JJ48wj:bp5pb2n) (no related file):*
> It's burning system resources constantly (not much of them though, but even when the app isn't doing anything) regardless of the user workload, just to optimize a specific set of workloads, in specific environments :/ Also, it makes our timers imprecise in the order of 100ms, which is a lot?

To be honest, I have the same feelings after I implemented this PR. And especially after I found out that I need to change the clock resolution (clock granularity) in some PAL tests -- this may really bite Gramine users, since this is a breaking and quite unexpected change. So I guess I'm blocking myself on this, and I don't want to merge this PR (consider it my first attempt at tackling this problem).

---

> - Implement separate invariant TSC detection for each hypervisor.

Yes, and this would be the most straight-forward option. However, it doesn't look like MS Hyper-V even **allows userland processes** to learn Invariant-TSC characteristics: https://learn.microsoft.com/en-us/virtualization/hyper-v-on-windows/tlfs/timers#partition-reference-tsc-mechanism (the only way to get these characteristics is to get the physical memory page from an MSR and read from that page, which is only possible for kernel space)

> - Talk to hypervisor devs and ask why this isn't emulated in their CPUID, maybe they could add it easily.

Yes, I want to ask Microsoft devs about this behavior. I'll inform how that discussion goes.

> - Add a "force using TSC for time" manifest switch? 

Yeah, could do... Though I dislike that we'll have two flows, which will be like some black magic for regular Gramine users...

---

One more alternative:

- Re-implement Linux's `vdso` logic inside Gramine, and let this logic access the `vvar` page.

This way we'll do 100% the same as what normal apps do, but there is a huge problem: vDSO -> vvar logic is not standardized and is considered an internal Linux detail (after all, that's why vDSO is implemented in Linux, not in Glibc or something).

I did look into the vDSO code of Linux, and it's... black magic. Hoping that one vDSO re-implementation in Gramine will work on any Linux (i.e., with any `vvar` layout), past and future, is dangereous and brittle.

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1154)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1154#-NOEXOdB3rmSlq784rl-:-NOEXOdB3rmSlq784rl0:b-owgpa5) (no related file):*
Unfortunately, my PR's side effect of **time resolution is lowered to around 100ms** breaks applications. One example is MariaDB: https://github.com/MariaDB/server/blob/b1856aff37557e82b0e53ddbd89fc41f86df07e6/mysys/my_rdtsc.c#L362

Basically, during init, MariaDB tries to calibrate the clock and gets a series timestamps, and at some point performs a division: https://github.com/MariaDB/server/blob/b1856aff37557e82b0e53ddbd89fc41f86df07e6/mysys/my_rdtsc.c#L618. This results in **division by zero exception**, and the application dies.

Unfortunately, my PR is too much of a hack. This PR should not be merged, and we need to go with other solutions (e.g. https://github.com/gramineproject/gramine/pull/1012).

___

dimakuv:
Closing this PR. It is superseded by https://github.com/gramineproject/gramine/pull/1180.

Issue #58, opened 2020-09-09
[LibOS] Missing support for flock syscall


trusty-fox:
Hello,
I don't know if there is a better place to track this, but would it be possible to add support for the `flock()` system call in Graphene LibOS?

At the moment, in [`shim_syscalls.c`](https://github.com/oscarlab/graphene/blob/master/LibOS/shim/src/shim_syscalls.c#L349) this is declared as unsupported.

dimakuv:
Looking at this from August 2022, some things are irrelevant/were done already:

> I think this is blocked on filesystem rewrite (the current filesystem code is really terrible :) ).

The FS rewrite was done.

> At the LibOS level, there is no "protected fs" (it is implemented purely at the Linux-SGX PAL layer).

Protected FS (renamed to a more logical Encrypted FS) was lifted from the PAL layer to the LibOS level.

> For tmpfs sure. We need tmpfs first though :)

Tmpfs was implemented.

aep:
would everyone be ok with just ignoring the syscall and returning 0 ?

i don't think there's value in actually implementing it inside tmpfs for example because there's no second process that contests it  anyway.

could be made a config option, i guess.

boryspoplawski:
> would everyone be ok with just ignoring the syscall and returning 0 ?

Definitely not, if something depends on it, it would break horribly.

> there's no second process that contests it anyway.

There could be multiple threads (which share virtual memory and hence tmpfs)

dimakuv:
+1 to @boryspoplawski, we should implement `flock()` properly.

lead4good:
We've run into this problem graminizing `geth`. As this issue has seen some activity recently, I am wondering if a patch is already being developed now or in the near future?

dimakuv:
@lead4good We'll discuss whether we have the manpower to work on `flock()`. We just finished v1.3, so maybe we can include it in v1.4... No promises though.

lead4good:
@dimakuv Great news! Please let us know once there is a final decision on whether a flock implementation will be included in v1.4 or not.

lead4good:
For our use case, implementing flock for tmpfs would be sufficient for now. As tmpfs is per process only my understanding would be we could always return success?

lead4good:
@dimakuv @boryspoplawski Would you accept a PR that implements flock for tmpfs only by always returning success?

boryspoplawski:
@lead4good No, as it was already discussed above.

lead4good:
[...]

PR #1171, opened 2023-02-13
[LibOS] Add support for `flock` system call

billionairiam:
## Description of the changes 
We implemented it with the OCALL mechanism to make it compatible with NFS.
```
int flock(int fd, int operation);
```
The function applies or removes an advisory lock on the open file specified by `fd`.
In Linux, We can use `flock` to place an advisory lock in file `fd`, socket `fd` , pipe `fd`, and event `fd`.

## How to test this PR? 
To test the function we should notice the following situation

1. A call to flock() may block if another process holds an incompatible lock. To make a nonblocking request, include LOCK_NB (by ORing) with any of the above operations.
2. A single file may not simultaneously have both shared and exclusive locks.
3. Locks created by flock() are associated with an open file table entry. This means that duplicate file descriptors (created by, for example, [fork](https://linux.die.net/man/2/fork)(2) or [dup](https://linux.die.net/man/2/dup)(2)) refer to the same lock, and this lock may be modified or released using any of these descriptors. Furthermore, the lock is released either by an explicit LOCK_UN operation on any of these duplicate descriptors, or when all such descriptors have been closed.
4. If a process uses [open](https://linux.die.net/man/2/open)(2) (or similar) to obtain more than one descriptor for the same file, these descriptors are treated independently by flock(). An attempt to lock the file using one of these file descriptors may be denied by a lock that the calling process has already placed via another descriptor.

Below is the test code.
```
#include <stdio.h>
#include <unistd.h>
#include <sys/file.h>
#include <sys/socket.h>
#include <sys/wait.h>
#include <errno.h>

static struct tcase {
	int operation;
	char *opt;
} tcases[] = {
	{LOCK_SH, "Shared Lock" },
	{LOCK_UN, "Unlock"},
	{LOCK_EX, "Exclusive Lock"},
};

int open_file(int perm) {
    int fd = open("test_flock.txt",perm);
    if (fd == -1)
        perror("open error!\n");

    return fd;
}

static void verify_flock_1(int index, int count) {
    int fd = open_file(O_RDONLY);

    for (size_t i =0; i < sizeof(tcases)/sizeof(tcases[0]);i++) {
        if(flock(fd,tcases[i].operation)) {
            printf("flock() with %s error\n", tcases[i].opt);
            printf("[%d/%d verify_flock_1 FAILED!]\n", index, count);
            return;
        }
    }

    close(fd);
    printf("[%d/%d verify_flock_1 SUCCESS!]\n", index, count);
}

static int badfd = -1;
static int fd;
static struct tcase_2 {
	int *fd;
	int operation;
	int exp_err;
} tcases_2[] = {
	{&badfd, LOCK_SH, EBADF},
	{&fd, LOCK_NB, EINVAL},
	{&fd, LOCK_SH | LOCK_EX, EINVAL},
};

static void verify_flock_2(int index, int count) {
    int tmpfd = open_file(O_RDONLY);
    tcases_2[1].fd = &tmpfd;
    tcases_2[2].fd = &tmpfd;

    for (size_t i =0; i < sizeof(tcases_2)/sizeof(tcases[0]);i++) {
        if(flock(*tcases_2[i].fd, tcases_2[i].operation) == 0) {
            printf("flock() succeeded unexpectedly\n");
            printf("FAILED PARAMETERS: [fd:%d] [operation:%d] [exp_err:%d]\n",*tcases_2[i].fd, tcases_2[i].operation, tcases_2[i].exp_err);
            printf("[%d/%d verify_flock_2 FAILED!]\n", index, count);
            return;
        } else {
            if (errno != tcases_2[i].exp_err) {
                printf("flock() failed unexpectedly\n");
                printf("EXPECTED errno %d\n",tcases_2[i].exp_err);
                return;
            }
        }
    }
    close(tmpfd);
    printf("[%d/%d verify_flock_2 SUCCESS!]\n", index, count);
}

static void childfunc(int fd, int index, int count) {
    sleep(1);
    int fd2 = open_file(O_RDONLY);
    if (flock(fd2, LOCK_EX | LOCK_NB) != -1) {
        printf("CHILD: The file was not already locked!\n");
        goto out;
    }


    if (flock(fd, LOCK_UN) == -1) {
        printf("CHILD: Unable to unlock file locked by parent!\n");
        goto out;
    }

    if (flock(fd2, LOCK_EX | LOCK_NB) == -1) {
        printf("CHILD: Unable to lock file after unlock!\n");
        goto out;
    }

    close(fd2);
    printf("[%d/%d verify_flock_3 SUCCESS!]\n", index, count);
    return;

out:
    printf("[%d/%d verify_flock_3 FAILED!]\n", index, count);
    return;
}

static void verify_flock_3(int index, int count) {
    int fd1;
    pid_t pid;

    fd1 =  open_file(O_RDONLY);

    pid = fork();
    if (pid == 0)
        childfunc(fd1, index, count);
    else {
        if (flock(fd1, LOCK_EX | LOCK_NB) == -1) {
            printf("Parent: Initial attempt to flock() failed\n");
        }
        int status;
        wait(&status);
        close(fd1);
    }
}

static void verify_flock_4(int index, int count) {
    int fd1, fd2;

    fd1 = open_file(O_RDONLY);
    if (flock(fd1, LOCK_EX | LOCK_NB)) {
        printf("FAILED: First attempt to flock() failed\n");
    }

    fd2 = open_file(O_RDONLY);
    if (flock(fd2, LOCK_EX | LOCK_NB)) {
        printf("SUCCESS: Second attempt to flock() denied\n");
    } else {
        printf("FAILED: Second attempt to flock() SUCCESS UNEXPECTLY\n");
    }

    if (flock(fd1, LOCK_UN)) {
        printf("FAILED: Failed to unlock fd1\n");
    }

    if (flock(fd2,  LOCK_EX | LOCK_NB)) {
        printf("FAILED: Third attempt to flock() denied!\n");
    }

    printf("[%d/%d verify_flock_4 SUCCESS!]\n", index, count);
}

int main(void) {
    verify_flock_1(1,4);
    verify_flock_2(2,4);
    verify_flock_3(3,4);
    verify_flock_4(4,4);
}
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1171)


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1171)* status: 0 of 28 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @billionairiam)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1171#-NO9Ozz9AdfVZU811ojh:-NO9Ozz9AdfVZU811oji:b-b0pxmv) (no related file):*
1. This PR is insecure. It is a pass-through emulation of `flock()`. Did you perform any security analysis?
2. `to make it compatible with NFS` -- what does it mean exactly, and is this relevant for Gramine?

Why was this PR implemented? I don't remember discussions around such a pass-through implementation of `flock`. I don't remember any Gramine developer/maintainer thinking it would be a good idea to have a pass-through implementation.

___

jkr0103 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1171)* status: 0 of 28 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @billionairiam)

___
*[`libos/src/meson.build` line 82 at r1](https://reviewable.io/reviews/gramineproject/gramine/1171#-NONycO01lCKA3QOx2po:-NONycO01lCKA3QOx2pp:bgj0wmp) ([raw file](https://github.com/gramineproject/gramine/blob/3422a1836fef09d1aefef2d44fde9ce083f20d85/libos/src/meson.build#L82)):*
> ```Meson
>     'sys/libos_exit.c',
>     'sys/libos_fcntl.c',
>     'sys/libos_flock.c',
> ```

Does gramine even build? I see file  `sys/libos_flock.c` is missing in PR.

libos/src/sys/libos_flock.c
```diff
@@ -0,0 +1,35 @@
...
+    struct libos_handle* hdl = get_fd_handle(fd, NULL, handle_map);
+    if (!hdl)
+        return -EBADF;
+
+    if(!hdl->dentry || !hdl->fs || !hdl->fs->fs_ops || !hdl->fs->fs_ops->flock) {
+        ret = -EINVAL;
+        goto out;
+    }
+
+    ret = hdl->fs->fs_ops->flock(hdl, operation);
```
BraunPhilipp, line 30: Wanted to finish a similar PR next week. Based on my understanding this is missing various semantics of `flock(...)`. I have opted in my implementation to map `flock(...)` to `fcntl(...)`. This also allows for NFS support and should avoid making any calls out of the enclave. Maybe missing something since this is not my forte.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1171)* status: 0 of 29 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @billionairiam)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1171#-NOgFu5NCjsHhI-rjVwE:-NOgFu5O2Su9sLGN7PhT:b-f9zh6s) (no related file):*
@BraunPhilipp: If you could, please use Reviewable for reviewing, not GitHub reviews (that's what we use in Gramine).

> Wanted to finish a similar PR next week. Based on my understanding this is missing various semantics of `flock(...)`. I have opted in my implementation to map `flock(...)` to `fcntl(...)`.

This should be a good solution, assuming you can actually map `flock(...)` semantics to `fcntl(F_{SET,GET}LK, ...)`, including all corner cases and undocumented quirks :)

> This also [...] should avoid making any calls out of the enclave.

Yes and no - it won't do any locking on the host (i.e. no passthrough and relying on the untrusted host), but it still needs to go out of the enclave to talk to other Gramine processes and exchange the locks state - but this will be done by Gramine's encrypted IPC.

___

libos/src/sys/libos_flock.c
billionairiam: `fcntl` and `flock` use different file locks in Linux or it will cause deadlock.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1171)* status: 0 of 29 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @billionairiam, @BraunPhilipp, and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1171#-NOgFu5NCjsHhI-rjVwE:-NOhHYFv549ghr8YVsjO:b-1q43qc) (no related file):*
> This should be a good solution, assuming you can actually map flock(...) semantics to fcntl(F_{SET,GET}LK, ...), including all corner cases and undocumented quirks :)

Well, `flock()` and `fcntl()` locks have different semantics. It's hard to map them all correct and can possibly incur some side effects.

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1171)* status: 0 of 29 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @billionairiam, @BraunPhilipp, and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1171#-NOgFu5NCjsHhI-rjVwE:-NOiB7GKAcCQtzj3JMsC:bo3go3s) (no related file):*
@mkow @BraunPhilipp `flock()` has very different semantics than `fcntl()`. The fact that Linux (more specifically, Glibc) used to implement `flock()` via `fcntl()` is **not** a matter of correctness, it's a matter of total insanity of file-locking implementations. Please read through the articles mentioned here: https://github.com/gramineproject/gramine/issues/58#issuecomment-1434794634. Just one citation:

> That behaviour is certifiably insane, and there's no possible justification for why it should work that way. But it's how it's always worked, and POSIX standardized it, and now you're stuck with it.

One of the main differences between `flock()` and `fcntl()`, in Gramine terms:
- `fcntl()` works on the `process -> inode` level, i.e. the relationship is like this `(process, inode) -> lock state`
- `flock()` works on the `process -> handle` level, i.e. the relationship is like this `(process, handle) -> lock state`

___

dimakuv:
Let me close this PR, we agreed that the `flock()` implementation should be very different from this. See also https://github.com/gramineproject/gramine/issues/58

PR #1093, opened 2023-01-03
Do `gramine-sgx-get-token` automatically

woju:
## Description of the changes 

There's no EINITTOKEN on upstream driver (nor DCAP), and most people use that driver (it is/will be the only packaged flavour). It's bad UX to require them to do `gramine-sgx-get-token` for nothing (like, 384 * `\0`).

This PR is the first of series of two. It changes UX (`gramine-sgx-get-token` gets called automatically), but doesn't change behaviour of loader (the `.token` file is still required).

## How to test this PR? 

Run `gramine-sgx` without doing `gramine-sgx-get-token` first.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1093)

PR #1175, opened 2023-02-14
[python] Do not fetch tokens for regression tests (during `pytest`)

dimakuv:
## Description of the changes 

This was a leftover after https://github.com/gramineproject/gramine/pull/1093

## How to test this PR? 

CI.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1175)

### DIFF
--- a/python/gramine-test
+++ b/python/gramine-test
@@ -63,6 +63,11 @@ def clean(ctx):
         if os.path.exists(name):
             print(f'deleting {name}')
             os.unlink(name)
+    for name in os.listdir(os.getcwd()):
+        # token files are created by Gramine itself, not known to Ninja build system
+        if name.endswith('.token'):
+            print(f'deleting {name}')
+            os.unlink(name)
 
 
 @main.command(
--- a/python/graminelibos/util_tests.py
+++ b/python/graminelibos/util_tests.py
@@ -39,7 +39,7 @@ class TestConfig:
 
     Ninja handles the following targets:
 
-    - `NAME.manifest`, `NAME.manifest.sgx`, `NAME.sig`, `NAME.token`
+    - `NAME.manifest`, `NAME.manifest.sgx`, `NAME.sig`
     - `direct`, `sgx`: all files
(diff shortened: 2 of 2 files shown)


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1175)* status: 1 of 2 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @woju)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1175#-NOERgKJE4rjJJwgs3uG:-NOEkD6P7NFAPesjhISO:byvdyc3) (no related file):*

How those files are cleaned? Because when you start the app, they're created anyway.
</blockquote></details>

Done. There is a problem that Ninja build system doesn't know about these `.token` files since they are generated by Gramine itself. So had to add an explicit removal, not sure it's a "clean" approach.

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1175#-NOEjuAKCGfRH1tDrMpt:-NOEjuAKCGfRH1tDrMpu:b-30dq3t) (no related file):*
It turns out that our regression-test Python scripts do not use the `gramine-sgx` wrapper (that generates token if needed), but instead use `loader` binary directly. So the token files are not generated in that case.

See: 
- https://github.com/gramineproject/gramine/blob/32504a91010df1bcab3b8fdf0e1fdb8160059771/python/graminelibos/regression.py#L186
- https://github.com/gramineproject/gramine/blob/32504a91010df1bcab3b8fdf0e1fdb8160059771/python/graminelibos/regression.py#L222

I thought this PR would be simple, but it turns out too complicated. I don't want to spend time on this non-problem anymore. If noone minds, I'll just close it (after some timeout).

___

dimakuv:
Closing, see my other comment. Too complicated.

Issue #1174, opened 2023-02-14
app crashed for " Failed to reserve virtual memory for CodeRange"


mccoysc:
### Description of the problem

after enabled edmm, app crashed as the title.
my app is nodejs.
my gramine env is from the docker image version 1.4.

### Gramine commit hash

i don'tknow

mccoysc:
my linux kernel version is 6.1.11

mccoysc:
the sgx.enclave_size is omitted in template file.

dimakuv:
> after enabled edmm, app crashed as the title.

And with EDMM disabled, the app works?

Does `gramine-direct` work?

mccoysc:
yes,work well without edmm enabled.

mccoysc:
> > after enabled edmm, app crashed as the title.
> 
> 
> 
> And with EDMM disabled, the app works?
> 
> 
> 
> Does `gramine-direct` work?

i never used the gramine-direct.

dimakuv:
Could you:

1. Show the `manifest.template` file.
2. Add `loader.log_level = "all"` (and keep EDMM enabled), rebuild the app, rerun it, and attach the resulting log.

mccoysc:
this is the working template:
```

[loader]
entrypoint = "file:{{ gramine.libos }}"
log_level = "{{ log_level }}"
insecure__use_cmdline_argv = true
python_entrypoint="{{python_entrypoint}}"

[loader.env]
LD_LIBRARY_PATH = "{{ python.stdlib }}/lib:/lib:{{ arch_libdir }}:/usr/lib:/usr{{ arch_libdir }}"


[libos]
entrypoint="{{ nodejs_dir }}/node"

# Node.js requires eventfd2() emulation otherwise fails on `(uv_loop_init(&tracing_loop_)) == (0)'
[sys]
insecure__allow_eventfd = true
[sys.stack]
size="4M"


[fs]
mounts=[
{type = "chroot",path = "{{ nodejs_dir }}",uri = "file:{{ nodejs_dir }}"},
{type = "chroot",path = "/lib",uri = "file:{{ gramine.runtimedir()}}"},
{type = "chroot",path = "{{ arch_libdir }}",uri = "file:{{ arch_libdir }}"},
{type = "chroot",path = "/etc/",uri = "file:/etc/"},
{ path = "/usr/", uri = "file:/usr/" },
{% for path in python.get_sys_path(python_entrypoint) %}
[...]

PR #1178, opened 2023-02-15
[Docs] Mention `sgx.enclave_size` in EDMM enablement section

dimakuv:
## Description of the changes 

Many people have a feeling that when `sgx.edmm_enable = true`, they can remove `sgx.enclave_size`. Which is not true (the enclave size becomes the "maximum limit on the available enclave-memory space"). Let's make it more explicit in our documentation.

See #1174.

## How to test this PR? 

N/A.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1178)

### DIFF
--- a/Documentation/manifest-syntax.rst
+++ b/Documentation/manifest-syntax.rst
@@ -459,6 +459,10 @@ as RWX). Unfortunately it can negatively impact performance, as adding a page
 to the enclave at runtime is a more expensive operation than adding the page
 before enclave creation (because it involves more enclave exits and syscalls).
 
+Even when EDMM is enabled, it is still necessary to specify ``sgx.enclave_size``
+-- it specifies the *maximal* size that the enclave can grow to. In case of
+doubt, set this size to some large value like ``128G``.
+
 .. note::
    Support for EDMM first appeared in Linux 6.0.
 


mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1178)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/1178#-NOKpnna2yINx7_r6a5r):*
Sounds weird to me



_Suggestion:_
```
in "EDMM" section
```
___
*[`Documentation/manifest-syntax.rst` line 464 at r1](https://reviewable.io/reviews/gramineproject/gramine/1178#-NOKpukB5XnkkdyBVZHr:-NOKpukB5XnkkdyBVZHs:b-mr5f9e) ([raw file](https://github.com/gramineproject/gramine/blob/27bb3af2b9fd13d868c1920e20c972d8f8a2e045/Documentation/manifest-syntax.rst#L464)):*
> ```reStructuredText
> Even when EDMM is enabled, it is still necessary to specify ``sgx.enclave_size``
> -- it specifies the *maximal* size that the enclave can grow to. In case of
> doubt, set this size to some large value like ``128G``.
> ```

What my proposal from this discussion under EDMM PR review? https://reviewable.io/reviews/gramineproject/gramine/1054#-NJuhiaXDbWKSYpitfwO (point 3)

Would be simpler for users and more intuitive.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1178)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv and @mkow)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/1178#-NOKpnna2yINx7_r6a5r):*

Sounds weird to me
</blockquote></details>

Will fix during final rebase
___
*[`Documentation/manifest-syntax.rst` line 464 at r1](https://reviewable.io/reviews/gramineproject/gramine/1178#-NOKpukB5XnkkdyBVZHr:-NOKrZsDDJ-5gH39lvzC:b5le6jj) ([raw file](https://github.com/gramineproject/gramine/blob/27bb3af2b9fd13d868c1920e20c972d8f8a2e045/Documentation/manifest-syntax.rst#L464)):*
Do you mean this proposal:

> maybe the default for sgx.enclave_size should depend on EDMM being used?

We had a discussion with Borys on this (I suggested the same), but he argued that this is too much "magic". Actually, I'm even thinking -- maybe we should force users to specify `sgx.enclave_size` always, because the default `256MB` is also "magic" that only confuses people.

Also, what would be the default in case EDMM is enabled? How do we choose and justify this number? Won't it also be "magic" that will bite us on some unexpected workload?

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1178)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`Documentation/manifest-syntax.rst` line 464 at r1](https://reviewable.io/reviews/gramineproject/gramine/1178#-NOKpukB5XnkkdyBVZHr:-NOMVkHIC47eTBXDX5bd:b-z54ud5) ([raw file](https://github.com/gramineproject/gramine/blob/27bb3af2b9fd13d868c1920e20c972d8f8a2e045/Documentation/manifest-syntax.rst#L464)):*
> maybe we should force users to specify `sgx.enclave_size` always, because the default `256MB` is also "magic" that only confuses people.

I could agree with this, but only for the non-EDMM case.

> Also, what would be the default in case EDMM is enabled? How do we choose and justify this number? Won't it also be "magic" that will bite us on some unexpected workload?

Yeah, but I'm not sure you got my idea right. I'd like to set it to something really big, which should be enough for 99% of workloads. On x64 address space is cheap, and it's the address space of our own process, which we control. 1 TB maybe? That's 1/140 of the address space, so it sounds safe.

I just want to make configuring Gramine as simple as possible for majority of workloads.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1178)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

___
*[`Documentation/manifest-syntax.rst` line 464 at r1](https://reviewable.io/reviews/gramineproject/gramine/1178#-NOKpukB5XnkkdyBVZHr:-NOOsR5u9smrLDpWdEAC:bq4i950) ([raw file](https://github.com/gramineproject/gramine/blob/27bb3af2b9fd13d868c1920e20c972d8f8a2e045/Documentation/manifest-syntax.rst#L464)):*
> Yeah, but I'm not sure you got my idea right. I'd like to set it to something really big, which should be enough for 99% of workloads. On x64 address space is cheap, and it's the address space of our own process, which we control. 1 TB maybe? That's 1/140 of the address space, so it sounds safe.

We need to be careful with specifying such memory:
- https://github.com/gramineproject/gramine/blob/51f81f6f295fa4c33273029d505169c1a0534a4c/common/include/asan.h#L133
- https://github.com/gramineproject/gramine/blob/51f81f6f295fa4c33273029d505169c1a0534a4c/pal/src/host/linux-sgx/gdb_integration/sgx_gdb.h#L7-L14

(These are the ones I remember about, I think there should be no other hard-coded memory areas.)

It actually may work fine with your suggested 1TB enclave size -- nothing will overlap. The only thing is that we need to make sure that in this case, we never perform this legacy hack: 
- https://github.com/gramineproject/gramine/blob/51f81f6f295fa4c33273029d505169c1a0534a4c/pal/src/host/linux-sgx/host_main.c#L277-L280
- https://github.com/gramineproject/gramine/blob/51f81f6f295fa4c33273029d505169c1a0534a4c/python/graminelibos/sgx_sign.py#L487-L488

@mkow Do you think we should go this route? I can create a PR then, instead of this one.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1178)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`Documentation/manifest-syntax.rst` line 464 at r1](https://reviewable.io/reviews/gramineproject/gramine/1178#-NOKpukB5XnkkdyBVZHr:-NOPW16E7c3xzfnkuVg-:bo433na) ([raw file](https://github.com/gramineproject/gramine/blob/27bb3af2b9fd13d868c1920e20c972d8f8a2e045/Documentation/manifest-syntax.rst#L464)):*
> We need to be careful with specifying such memory:

We can just change this constant if it collides.

> @mkow Do you think we should go this route? I can create a PR then, instead of this one.

I don't see why not :)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1178)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

___
*[`Documentation/manifest-syntax.rst` line 464 at r1](https://reviewable.io/reviews/gramineproject/gramine/1178#-NOKpukB5XnkkdyBVZHr:-NOPlx8j7v7BINqCCeSc:bb0x0xl) ([raw file](https://github.com/gramineproject/gramine/blob/27bb3af2b9fd13d868c1920e20c972d8f8a2e045/Documentation/manifest-syntax.rst#L464)):*
> We can just change this constant if it collides.

I wouldn't want to touch the ASan constant (1.5TB). It's nontrivial. You can read about the calculations in the comments here: https://github.com/gramineproject/gramine/blob/51f81f6f295fa4c33273029d505169c1a0534a4c/common/include/asan.h

That's why I wrote that whole comment before -- it's not as trivial as bumping these constants to some higher addresses.

dimakuv:
Superseded by another approach in https://github.com/gramineproject/gramine/pull/1187 and 
https://github.com/gramineproject/gramine/pull/1188

PR #1205, opened 2023-02-28
WIP: Implement 'flock' using POSIX locks

BraunPhilipp:
As discussed with @dimakuv I started working this week on an initial draft for `flock` using POSIX locks. Just noticed that @billionairiam also started working on this according to #58. Since this is primarily educational, I don't mind too much. 

The PR implements `flock` using POSIX locks similar to `fcntl`. We introduce `handle_id` as discussed in #58 to easily track forks and dups.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1205)


dimakuv:
@BraunPhilipp Quick answers to the (unrelated to flock) issues you asked:

> `/home/gramine/libos/test/ltp/src/lib/tst_test.c:108: TBROK: mmap((nil),4096,3,1,3,0) failed: EACCES (13)`

You're trying to run Gramine with the SGX backend (`gramine-sgx`) on LTP tests. This is currently unsupported. LTP tests are **only** tested with Gramine on Linux (aka `gramine-direct`). That's why you see these mmap failures.

> `error: PAL failed setting personality failed: Operation not permitted (EPERM)`

Are you running it inside a Docker container? If yes, then there is a clash it the seccomp policy used by Docker by default and the seccomp policy required by Gramine. Please check the README here: https://hub.docker.com/r/gramineproject/gramine.

Depending on the version of Docker/Linux you use, you may need one of two different seccomp policies that we ship with Gramine: https://github.com/gramineproject/gramine/tree/master/scripts, `docker_seccomp_aug_2022.json` and `docker_seccomp_mar_2021.json`.

For some more context and an example on how to run Docker with a specific seccomp policy, you can also check this: https://gramine.readthedocs.io/projects/gsc/en/latest/#execute-with-linux-pal-instead-of-linux-sgx-pal

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1205)* status: 0 of 9 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @BraunPhilipp)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1205#-NPP0JXH8K9T2yIVBg26:-NPP0JXJ5fJsDGY3OI8Y:bze17yj) (no related file):*
One thing that is definitely missing -- when the LibOS handle is closed, the corresponding flock (if any) should be unlocked. Similarly, when the process is terminated, all its corresponding flocks should be unlocked.

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1205#-NPP12ZG6bPrN-HNsIKL:-NPP12ZHAkJVXhb6C7OF:b-ha1mgp) (no related file):*
Ok, another thing seems to be missing -- I don't think your implementation supports **flock inheritance on fork**. I suggest you to write a LibOS regression test with one process grabbing a lock and then forking the other (child) process, and the child process should be able to e.g. unlock this same lock.

___

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1205)* status: 0 of 9 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @BraunPhilipp and @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1205#-NPP0JXH8K9T2yIVBg26:-NPPgUEc6BlohFf5myte:b-9sdwkx) (no related file):*

One thing that is definitely missing -- when the LibOS handle is closed, the corresponding flock (if any) should be unlocked. Similarly, when the process is terminated, all its corresponding flocks should be unlocked.
</blockquote></details>

@BraunPhilipp, for detailed reference, pls kindly take another look at the proposal by @dimakuv: https://github.com/gramineproject/gramine/issues/58#issuecomment-1437237020 which I also copied below.

We must be careful to not remove flock locks on [closing the file descriptor](https://github.com/gramineproject/gramine/blob/da7ea3f0f4e72033ccbea877e5599ee65564a516/libos/src/bookkeep/libos_handle.c#L294)!
* flock locks should be removed on handle close, somewhere around [here](https://github.com/gramineproject/gramine/blob/da7ea3f0f4e72033ccbea877e5599ee65564a516/libos/src/bookkeep/libos_handle.c#L502).
* flock locks should also be removed on process termination, somewhere around [here](https://github.com/gramineproject/gramine/blob/da7ea3f0f4e72033ccbea877e5599ee65564a516/libos/src/sys/libos_exit.c#L103).

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1205#-NPP12ZG6bPrN-HNsIKL:-NPPiNCY8hZH-wPeKW1M:b4lxmvp) (no related file):*

Ok, another thing seems to be missing -- I don't think your implementation supports **flock inheritance on fork**. I suggest you to write a LibOS regression test with one process grabbing a lock and then forking the other (child) process, and the child process should be able to e.g. unlock this same lock.
</blockquote></details>

For reference, pls kindly take a look at [libos_ipc_fs_lock.c](https://github.com/gramineproject/gramine/blob/ca2b66e4a8a388f65e2d224ed555c1a04052b026/libos/src/ipc/libos_ipc_fs_lock.c). WRT tests, there are some other tests suggested in the original thread of discussion that you may be interested.

___

billionairiam:
1. If I `dup` 10 times and got 10 different `handle_id`, every time I close the `handle_id`, the `posix_lock` shouldn't be unlocked. Only all the duped `handle_id` and original `handle_id` are closed, and the `posix_lock` can be finally closed. Maybe we should have a `ref_count`, every time when we `flock` the duped `handle_id` we should make `ref_count++`, every time we close the `handle_id` make the `ref_count--`, only when the `ref_count == 0`, we can finally close unlock it.
2. When we `mmap` a file with `flock`. when we closed the file but with `mmap`, so we shouldn't unlock the `posix_lock`.

dimakuv:
> 1. If I `dup` 10 times and got 10 different `handle_id`, every time I close the `handle_id`, the `posix_lock` shouldn't be unlocked. Only all the duped `handle_id` and original `handle_id` are closed, and the `posix_lock` can be finally closed. Maybe we should have a `ref_count`, every time when we `flock` the duped `handle_id` we should make `ref_count++`, every time we close the `handle_id` make the `ref_count--`, only when the `ref_count == 0`, we can finally close unlock it.

@billionairiam But we already have this functionality implemented in Gramine. That's why Gramine has a separation on `libos_handle` objects and `libos_fd_handle` helper objects. Every time a new FD is taken for the same `libos_handle` (e.g. via `dup()`), the refcount of the `libos_handle` object is increased. And on every `close(fd)`, the refcount of the `libos_handle` object is decreased.

To check out this logic, you can start analyzing from this place: https://github.com/gramineproject/gramine/blob/ca2b66e4a8a388f65e2d224ed555c1a04052b026/libos/src/bookkeep/libos_handle.c#L369

So the "unlock posix_lock" logic for flocks should be placed somewhere here: https://github.com/gramineproject/gramine/blob/ca2b66e4a8a388f65e2d224ed555c1a04052b026/libos/src/bookkeep/libos_handle.c#L495-L518

> 2. When we `mmap` a file with `flock`. when we closed the file but with `mmap`, so we shouldn't unlock the `posix_lock`.

@billionairiam This is a good question what happens on `mmap(..., <file-fd>)` and `flock(file-fd)`. To be honest, I don't know how Linux kernel behaves in this case. Please double-check and implement in the same way.

BraunPhilipp:
@dimakuv Thank you for the feedback I got my docker deployment successfully working. The LTP tests are now also partially passing.

BraunPhilipp (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1205)* status: 0 of 15 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1205#-NPP0JXH8K9T2yIVBg26:-NPh5vpE7u6HWuOEQiGR:b58go2r) (no related file):*

@BraunPhilipp, for detailed reference, pls kindly take another look at the proposal by @dimakuv: https://github.com/gramineproject/gramine/issues/58#issuecomment-1437237020 which I also copied below.

We must be careful to not remove flock locks on [closing the file descriptor](https://github.com/gramineproject/gramine/blob/da7ea3f0f4e72033ccbea877e5599ee65564a516/libos/src/bookkeep/libos_handle.c#L294)!
* flock locks should be removed on handle close, somewhere around [here](https://github.com/gramineproject/gramine/blob/da7ea3f0f4e72033ccbea877e5599ee65564a516/libos/src/bookkeep/libos_handle.c#L502).
* flock locks should also be removed on process termination, somewhere around [here](https://github.com/gramineproject/gramine/blob/da7ea3f0f4e72033ccbea877e5599ee65564a516/libos/src/sys/libos_exit.c#L103).

</blockquote></details>

* Was addressed in a recent commit using `handle_id`. Assuming that `clear_posix_locks` and the additional `LISTP_DEL` on `flock` close does the trick.
* For `flock` termination I am assuming that the call through `pid` is sufficient.

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1205#-NPP12ZG6bPrN-HNsIKL:-NPh7O9A2NPP7fRSJzCZ:b-qw9uqo) (no related file):*

For reference, pls kindly take a look at [libos_ipc_fs_lock.c](https://github.com/gramineproject/gramine/blob/ca2b66e4a8a388f65e2d224ed555c1a04052b026/libos/src/ipc/libos_ipc_fs_lock.c). WRT tests, there are some other tests suggested in the original thread of discussion that you may be interested.
</blockquote></details>

Added an appropriate test. Will investigate further in the coming days.

___

BraunPhilipp:
@billionairiam did you investigate `mmap` further? I did not find any special treatment for `fcntl`.

BraunPhilipp:
Will close in favor of https://github.com/gramineproject/gramine/pull/1212 after reviewing some testcases.

BraunPhilipp:
I think @billionairiam has the leg up on this. For me this was also just a great learning experience.

dimakuv:
I agree to proceed with @billionairiam's PR. I hope that @BraunPhilipp will review that other PR, given his gathered experience with `flock` now (which he already does, as I can see).

I also hope that @BraunPhilipp will contribute other features to Gramine. I'm impressed by the quality of the code!

PR #684, opened 2022-06-24
[LibOS] Support `FIONREAD` ioctl on sockets

boryspoplawski:
## Description of the changes 
Fixes https://github.com/gramineproject/gramine/issues/673

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/684)

PR #1210, opened 2023-03-03
[LibOS] Add dummy ancillary data support for sockets

dimakuv:
## Description of the changes 

Previously, Gramine returned `-ENOSYS` on `recvmsg()`/`sendmsg()` and similar syscalls if it detected the use of `struct msghdr` ancillary data (`msg_control` field). This PR is the first step in adding proper ancillary data support.

Currently no ancillary data type is truly supported:
- on `sendmsg()`:
  - unknown/unsupported types force an error,
  - SCM_RIGHTS/SCM_CREDENTIALS are ignored in TCP/UDP sockets;
- on `recvmsg()`: `msg_controllen` is set to zero to indicate there is no ancillary data added to the received network packet.

## How to test this PR? 

I added a LibOS regression test + I disabled one misbehaving LTP test.

One can try this example: https://github.com/gramineproject/examples/pull/57. Without this PR, it doesn't work properly (see test 2 in the README there).

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1210)

### DIFF
--- a/common/include/linux_socket.h
+++ b/common/include/linux_socket.h
@@ -55,6 +55,8 @@ struct cmsghdr {
 #define CMSG_LEN(len)   (CMSG_ALIGN(sizeof(struct cmsghdr)) + (len))
 
 #define SCM_RIGHTS 1
+#define SCM_CREDENTIALS 2
+#define SCM_SECURITY 3
 
 #define AF_UNSPEC 0
 #define AF_UNIX 1
@@ -96,8 +98,13 @@ struct cmsghdr {
 #define SO_RCVTIMEO 20
 #define SO_SNDTIMEO 21
 #define SO_ACCEPTCONN 30
+#define SO_MARK 36
+#define SO_TIMESTAMPING_OLD 37
 #define SO_PROTOCOL 38
 #define SO_DOMAIN 39
+
+#define SO_TXTIME 61
+#define SCM_TXTIME SO_TXTIME
 
 /* TCP options. */
 #define TCP_NODELAY 1       /* Turn off Nagle's algorithm */
--- a/libos/include/libos_socket.h
+++ b/libos/include/libos_socket.h
@@ -80,6 +80,8 @@ struct libos_sock_ops {
      * \param      handle             A handle.
      * \param      iov                An array of buffers to write from.
      * \param      iov_len            The length of \p iov.
+     * \param      msg_control        An ancillary data buffer.
+     * \param      msg_controllen     The length of \p msg_control.
      * \param[out] out_size           On success contains the number of bytes sent.
      * \param      addr               An address to send to. May be NULL. It's up to
      *                                the implementation to decide what to do with it (which might
@@ -88,8 +90,9 @@ struct libos_sock_ops {
      * \param      force_nonblocking  If `true` this request should not block. Otherwise just use
      *                                whatever mode the handle is in.
      */
-    int (*send)(struct libos_handle* handle, struct iovec* iov, size_t iov_len, size_t* out_size,
-                void* addr, size_t addrlen, bool force_nonblocking);
+    int (*send)(struct libos_handle* handle, struct iovec* iov, size_t iov_len, void* msg_control,
+                size_t msg_controllen, size_t* out_size, void* addr, size_t addrlen,
+                bool force_nonblocking);
 
     /*!
      * \brief Receive continuous data into an array of buffers.
@@ -97,19 +100,23 @@ struct libos_sock_ops {
      * \param         handle             A handle.
      * \param         iov                An array of buffers to read to.
      * \param         iov_len            The length of \p iov.
+     * \param         msg_control        An ancillary data buffer to populate.
+     * \param[in,out] msg_controllen_ptr The length of \p msg_control. On success updated to the
+     *                                   actual length of the received ancillary data.
      * \param[out]    out_total_size     On success contains the number of bytes received (STREAM)
      *                                   or the datagram size (DGRAM), which might be bigger than
      *                                   the total size of buffers in \p iov array.
      * \param[out]    addr               On success contains the address data was received from. May
      *                                   be NULL.
-     * \param[in,out] addrlen            The length of \p addr. On success updated to the actual
+     * \param[in,out] addrlen_ptr        The length of \p addr. On success updated to the actual
      *                                   length of the address. Bigger than original value indicates
      *                                   that truncation has happened.
      * \param         force_nonblocking  If `true` this request should not block. Otherwise just use
      *                                   whatever mode the handle is in.
      */
-    int (*recv)(struct libos_handle* handle, struct iovec* iov, size_t iov_len,
-                size_t* out_total_size, void* addr, size_t* addrlen, bool force_nonblocking);
+    int (*recv)(struct libos_handle* handle, struct iovec* iov, size_t iov_len, void* msg_control,
+                size_t* msg_controllen_ptr, size_t* out_total_size, void* addr, size_t* addrlen_ptr,
+                bool force_nonblocking);
 };
 
 struct libos_handle* get_new_socket_handle(int family, int type, int protocol,
@@ -118,7 +125,9 @@ struct libos_handle* get_new_socket_handle(int family, int type, int protocol,
 extern struct libos_sock_ops sock_unix_ops;
 extern struct libos_sock_ops sock_ip_ops;
 
-ssize_t do_recvmsg(struct libos_handle* handle, struct iovec* iov, size_t iov_len, void* addr,
-                   size_t* addrlen, unsigned int* flags);
-ssize_t do_sendmsg(struct libos_handle* handle, struct iovec* iov, size_t iov_len, void* addr,
-                   size_t addrlen, unsigned int flags);
+ssize_t do_recvmsg(struct libos_handle* handle, struct iovec* iov, size_t iov_len,
+                   void* msg_control, size_t* msg_controllen_ptr, void* addr, size_t* addrlen_ptr,
+                   unsigned int* flags);
+ssize_t do_sendmsg(struct libos_handle* handle, struct iovec* iov, size_t iov_len,
+                   void* msg_control, size_t msg_controllen, void* addr, size_t addrlen,
+                   unsigned int flags);
--- a/libos/src/fs/socket/fs.c
+++ b/libos/src/fs/socket/fs.c
@@ -36,7 +36,8 @@ static ssize_t read(struct libos_handle* handle, void* buf, size_t size, file_of
         .iov_len = size,
     };
     unsigned int flags = 0;
-    return do_recvmsg(handle, &iov, /*iov_len=*/1, /*addr=*/NULL, /*addrlen=*/NULL, &flags);
+    return do_recvmsg(handle, &iov, /*iov_len=*/1, /*msg_control=*/NULL,
+                      /*msg_controllen_ptr=*/NULL, /*addr=*/NULL, /*addrlen_ptr=*/NULL, &flags);
 }
 
 static ssize_t write(struct libos_handle* handle, const void* buf, size_t size, file_off_t* pos) {
@@ -45,20 +46,23 @@ static ssize_t write(struct libos_handle* handle, const void* buf, size_t size, 
         .iov_base = (void*)buf,
         .iov_len = size,
     };
-    return do_sendmsg(handle, &iov, /*iov_len=*/1, /*addr=*/NULL, /*addrlen=*/0, /*flags=*/0);
+    return do_sendmsg(handle, &iov, /*iov_len=*/1, /*msg_control=*/NULL, /*msg_controllen=*/0,
+                      /*addr=*/NULL, /*addrlen=*/0, /*flags=*/0);
 }
 
 static ssize_t readv(struct libos_handle* handle, struct iovec* iov, size_t iov_len,
                      file_off_t* pos) {
     __UNUSED(pos);
     unsigned int flags = 0;
-    return do_recvmsg(handle, iov, iov_len, /*addr=*/NULL, /*addrlen=*/NULL, &flags);
+    return do_recvmsg(handle, iov, iov_len, /*msg_control=*/NULL, /*msg_controllen_ptr=*/NULL,
+                      /*addr=*/NULL, /*addrlen_ptr=*/NULL, &flags);
 }
 
 static ssize_t writev(struct libos_handle* handle, struct iovec* iov, size_t iov_len,
                       file_off_t* pos) {
     __UNUSED(pos);
-    return do_sendmsg(handle, iov, iov_len, /*addr=*/NULL, /*addrlen=*/0, /*flags=*/0);
+    return do_sendmsg(handle, iov, iov_len, /*msg_control=*/NULL, /*msg_controllen=*/0,
+                      /*addr=*/NULL, /*addrlen=*/0, /*flags=*/0);
 }
 
 static int hstat(struct libos_handle* handle, struct stat* stat) {
--- a/libos/src/net/ip.c
+++ b/libos/src/net/ip.c
@@ -11,14 +11,15 @@
 
 #include "libos_fs.h"
 #include "libos_socket.h"
+#include "linux_socket.h"
 #include "pal.h"
 #include "socket_utils.h"
 
-static int verify_sockaddr(int expected_family, void* addr, size_t* addrlen) {
+static int verify_sockaddr(int expected_family, void* addr, size_t* addrlen_ptr) {
     unsigned short family;
     switch (expected_family) {
         case AF_INET:
-            if (*addrlen < sizeof(struct sockaddr_in)) {
+            if (*addrlen_ptr < sizeof(struct sockaddr_in)) {
                 return -EINVAL;
             }
             memcpy(&family, (char*)addr + offsetof(struct sockaddr_in, sin_family), sizeof(family));
@@ -27,10 +28,10 @@ static int verify_sockaddr(int expected_family, void* addr, size_t* addrlen) {
             }
             /* Cap the address at the maximal possible size - rest of the input buffer (if any) is
              * ignored. */
-            *addrlen = sizeof(struct sockaddr_in);
+            *addrlen_ptr = sizeof(struct sockaddr_in);
             break;
         case AF_INET6:
-            if (*addrlen < sizeof(struct sockaddr_in6)) {
+            if (*addrlen_ptr < sizeof(struct sockaddr_in6)) {
                 return -EINVAL;
             }
             memcpy(&family, (char*)addr + offsetof(struct sockaddr_in6, sin6_family),
@@ -40,7 +41,7 @@ static int verify_sockaddr(int expected_family, void* addr, size_t* addrlen) {
             }
             /* Cap the address at the maximal possible size - rest of the input buffer (if any) is
              * ignored. */
-            *addrlen = sizeof(struct sockaddr_in6);
+            *addrlen_ptr = sizeof(struct sockaddr_in6);
             break;
         default:
             BUG();
@@ -662,12 +663,53 @@ static int getsockopt(struct libos_handle* handle, int level, int optname, void*
     }
 }
 
-static int send(struct libos_handle* handle, struct iovec* iov, size_t iov_len, size_t* out_size,
-                void* addr, size_t addrlen, bool force_nonblocking) {
+static int send(struct libos_handle* handle, struct iovec* iov, size_t iov_len, void* msg_control,
+                size_t msg_controllen, size_t* out_size, void* addr, size_t addrlen,
+                bool force_nonblocking) {
     assert(handle->type == TYPE_SOCK);
 
     struct libos_sock_handle* sock = &handle->info.sock;
     struct sockaddr_storage sock_addr;
+
+    struct cmsghdr* cmsg = (struct cmsghdr*)msg_control;
+    size_t rest_msg_controllen = msg_controllen;
+    while (cmsg && rest_msg_controllen >= sizeof(struct cmsghdr)) {
+        if (cmsg->cmsg_len < sizeof(struct cmsghdr) ||
+                CMSG_ALIGN(cmsg->cmsg_len) > rest_msg_controllen) {
+            return -EINVAL;
+        }
+
+        if (cmsg->cmsg_level != SOL_SOCKET) {
+            /*
+             * We currently don't support:
+             * - SOL_UDP:  UDP_SEGMENT
+             * - SOL_IPV6: IPV6_PKTINFO
+             * - SOL_IP:   IP_RETOPTS, IP_PKTINFO, IP_TTL, IP_TOS
+             *
+             * Note that there are no cmsgs for TCP (SOL_TCP) in Linux (as of v6.0).
+             */
+            return -EINVAL;
+        }
+
+        switch (cmsg->cmsg_type) {
+            /* We currently don't support below SOL_SOCKET types. */
+            case SO_MARK:
+            case SO_TIMESTAMPING_OLD:
+            case SCM_TXTIME:
+                return -EINVAL;
+
+            /* SCM_RIGHTS and SCM_CREDENTIALS are semantically in SOL_UNIX, simply ignored */
+            case SCM_RIGHTS:
+            case SCM_CREDENTIALS:
+                break;
+
+            default:
+                return -EINVAL;
+        }
+
+        rest_msg_controllen -= CMSG_ALIGN(cmsg->cmsg_len);
+        cmsg = (struct cmsghdr*)((char*)cmsg + CMSG_ALIGN(cmsg->cmsg_len));
+    }
 
     switch (sock->type) {
         case SOCK_STREAM:
@@ -709,15 +751,16 @@ static int send(struct libos_handle* handle, struct iovec* iov, size_t iov_len, 
     return ret;
 }
 
-static int recv(struct libos_handle* handle, struct iovec* iov, size_t iov_len,
-                size_t* out_total_size, void* addr, size_t* addrlen, bool force_nonblocking) {
+static int recv(struct libos_handle* handle, struct iovec* iov, size_t iov_len, void* msg_control,
+                size_t* msg_controllen_ptr, size_t* out_total_size, void* addr, size_t* addrlen_ptr,
+                bool force_nonblocking) {
     assert(handle->type == TYPE_SOCK);
 
     switch (handle->info.sock.type) {
         case SOCK_STREAM:
             /* TCP - not interested in remote address (we know it already). */
             addr = NULL;
-            addrlen = NULL;
+            addrlen_ptr = NULL;
             break;
         case SOCK_DGRAM:
             break;
@@ -731,14 +774,29 @@ static int recv(struct libos_handle* handle, struct iovec* iov, size_t iov_len,
     if (ret < 0) {
         return pal_to_unix_errno(ret);
     }
+
+    if (msg_control && msg_controllen_ptr) {
+        /*
+         * We currently don't support:
+         * - SOL_TCP:    TCP_CM_INQ
+         * - SOL_SOCKET: SO_TIMESTAMPNS_NEW, SO_TIMESTAMPNS_OLD, SO_TIMESTAMP_NEW, SO_TIMESTAMP_OLD
+         * - SOL_IPV6:   IPV6_PKTINFO
+         * - SOL_IP:     IP_RETOPTS, IP_RECVOPTS, IP_PKTINFO, IP_TTL, IP_TOS, IP_RECVFRAGSIZE,
+         *               IP_CHECKSUM, SCM_SECURITY, IP_ORIGDSTADDR, IP_RECVERR
+         *
+         *  Note that SCM_RIGHTS and SCM_CREDENTIALS are not possible on TCP/UDP sockets.
+         */
+        *msg_controllen_ptr = 0;
+    }
+
     if (addr) {
         struct sockaddr_storage linux_addr;
         size_t linux_addr_len = sizeof(linux_addr);
         pal_to_linux_sockaddr(&pal_ip_addr, &linux_addr, &linux_addr_len);
         /* If the user provided buffer is too small, the address is truncated, but we report
-         * the actual address size in `addrlen`. */
-        memcpy(addr, &linux_addr, MIN(*addrlen, linux_addr_len));
-        *addrlen = linux_addr_len;
+         * the actual address size in `addrlen_ptr`. */
+        memcpy(addr, &linux_addr, MIN(*addrlen_ptr, linux_addr_len));
+        *addrlen_ptr = linux_addr_len;
     }
     return 0;
 }
--- a/libos/src/net/unix.c
+++ b/libos/src/net/unix.c
@@ -19,25 +19,27 @@
 #include "libos_fs.h"
 #include "libos_internal.h"
 #include "libos_socket.h"
+#include "linux_socket.h"
 #include "pal.h"
 
 /*!
  * \brief Verify UNIX socket address and convert it to a unique socket name.
  *
  * \param         addr            The socket address to convert.
- * \param[in,out] addrlen         Pointer to the size of \p addr. Always updated to the actual size
+ * \param[in,out] addrlen_ptr     Pointer to the size of \p addr. Always updated to the actual size
  *                                of the address (but it's never extended).
  * \param[out]    sock_name       Buffer for the output socket name. On success contains a null
  *                                terminated string.
  * \param         sock_name_size  Size of \p sock_name.
  */
-static int unaddr_to_sockname(void* addr, size_t* addrlen, char* sock_name, size_t sock_name_size) {
-    if (*addrlen > sizeof(struct sockaddr_un)) {
+static int unaddr_to_sockname(void* addr, size_t* addrlen_ptr, char* sock_name,
+                              size_t sock_name_size) {
+    if (*addrlen_ptr > sizeof(struct sockaddr_un)) {
         /* Cap the address at the maximal possible size - rest of the input buffer (if any) is
          * ignored. */
-        *addrlen = sizeof(struct sockaddr_un);
+        *addrlen_ptr = sizeof(struct sockaddr_un);
     }
-    if (*addrlen < offsetof(struct sockaddr_un, sun_path) + 1) {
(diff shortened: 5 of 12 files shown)


mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1210)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPbXO9NBD_PXNAU0i9h:-NPcS0Gk8pm4XEn-Lk7I:b-14e4v9) (no related file):*

This was found and debugged on the `swtpm` software, when I tried to enable it in Gramine.

Here are the relevant code snippets:
- https://github.com/stefanberger/swtpm/blob/185832c8d4be95fa9305279240ca33b7403cfc91/src/swtpm/ctrlchannel.c#L500-L506
- https://github.com/stefanberger/swtpm/blob/185832c8d4be95fa9305279240ca33b7403cfc91/src/swtpm/ctrlchannel.c#L345
- https://github.com/stefanberger/swtpm/blob/185832c8d4be95fa9305279240ca33b7403cfc91/src/swtpm/ctrlchannel.c#L810-L822

If you look closely at this code, you'll notice two things:
- This `recvmsg(...with ancillary data...)` is called **even on TCP/IP sockets**, where it is simply ignored.
- The ancillary data is used **only in one command case** (`CMD_SET_DATAFD`), and not used in all other cases (that make sense in Gramine-SGX).

You can even notice that `CMD_SET_DATAFD` is described as useful only for Unix Domain Sockets in the man page: https://github.com/stefanberger/swtpm/blob/e5fdd1c181d0414ad27a0f4f72c49020fbc1ac6e/man/man3/swtpm_ioctls.pod

So to workaround such flows, I think it is reasonable to merge this PR.
</blockquote></details>

This doesn't sound correct, you are silently dropping the data which the user app sent? I don't like it, it will make the apps which actually use ancillary data super annoying to debug.

If swtpm is open source, isn't it easier to just fix it to not send the data, which it doesn't even use?

___

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1210)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPbXO9NBD_PXNAU0i9h:-NPcSfGX5kV2Rs8wsQ3b:b4386fd) (no related file):*

This doesn't sound correct, you are silently dropping the data which the user app sent? I don't like it, it will make the apps which actually use ancillary data super annoying to debug.

If swtpm is open source, isn't it easier to just fix it to not send the data, which it doesn't even use?
</blockquote></details>

Ok, you have a warning, but still... I don't like making our emulation behave differently than Linux just to workaround one specific app.

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1210)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPbXO9NBD_PXNAU0i9h:-NPcUaWF5Wg1u-ewaRO_:b-vfxcyf) (no related file):*
> This doesn't sound correct, you are silently dropping the data which the user app sent? I don't like it, it will make the apps which actually use ancillary data super annoying to debug.

TCP/IP sockets cannot have ancillary data.

Well, but now I think -- why do we even need a warning message. It should be the same behavior as Linux, so doesn't require a warning.

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1210)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPbXO9NBD_PXNAU0i9h:-NPqZ31PCctkq8AI5h2s:bl6cvkq) (no related file):*

> TCP/IP sockets cannot have ancillary data.

Well, that's not exactly true. Here are the ancillary data types that could be on the socket during `sendmsg()`:

- https://elixir.bootlin.com/linux/latest/source/net/core/sock.c#L2757
- (so, `SO_MARK`, `SO_TIMESTAMPING_OLD` and `SCM_TXTIME` are applicable to TCP, it seems, but they are all clearly unsupported by Gramine so could be ignored)

Here are the ancillary data types that could be on the socket during `recvmsg()`:

- https://elixir.bootlin.com/linux/latest/source/net/ipv4/tcp.c#L2684
- (so, `SO_TIMESTAMPNS_NEW`, `TCP_CM_INQ` are applicable to TCP, but they are unsupported by Gramine so could be ignored)
- (btw, found an interesting example of this: https://github.com/torvalds/linux/blob/master/tools/testing/selftests/net/tcp_inq.c)

TLDR: Yeah, looks like Linux added support for ancillary data for TCP/IP sockets in recent years, though Gramine has no mechanics to support these data. So what about this:

1. I'll create another PR that has more specific handling of ancillary data,
2. `sendmsg` will fail on any ancillary data
3. `recmsg` will *not* add any ancillary data

This will be kinda similar to the current PR, but more explicit. @mkow ?
</blockquote></details>

I experimented with TCP and different ancillary data types:

### `sendmsg()`

1. If `msg_control` is set to an empty buffer, then `sendmsg()` returns EINVAL, triggering this case: https://elixir.bootlin.com/linux/v6.2/source/net/core/sock.c#L2794
2. If `msg_control` is set to `SCM_RIGHTS`, then `sendmsg()` sends the packet and **ignores** ancillary data, triggering this case: https://elixir.bootlin.com/linux/v6.2/source/net/core/sock.c#L2792
3. If `msg_control` is set to some non-socket level (i.e. `cmsg_level = IPPROTO_IP`), then **ignores** this ancillary data; this case: https://elixir.bootlin.com/linux/v6.2/source/net/core/sock.c#L2810
4. If `msg_control` is set to socket-specific `SO_TIMESTAMPING`, then `sendmsg()` uses this ancillary data, triggering this case: https://elixir.bootlin.com/linux/v6.2/source/net/core/sock.c#L2780

### `recvmsg()`

1. If `sendmsg()` sent no ancillary data, `recvmsg()` gets nothing (`msg_controllen = 0`)
2. If `sendmsg()` sent `SCM_RIGHTS`, `recvmsg()` gets nothing (`msg_controllen = 0`)
3. If `sendmsg()` sent `IPPROTO_IP`, `recvmsg()` gets nothing (`msg_controllen = 0`)
4. If client socket has `setsockopt(SOL_SOCKET, SO_TIMESTAMPING...)` with `SOF_TIMESTAMPING_RX_SOFTWARE | SOF_TIMESTAMPING_SOFTWARE`, `recvmsg()` gets `SO_TIMESTAMPING_OLD` ancillary data (with `msg_controllen = 24`)

### Conclusion

TCP/IP sockets in Linux:
- can send and receive ancillary data (I experimented with `SO_TIMESTAMPING`)
- `sendmsg()` ignores UDS-specific ancillary data or non-socket-level data and fails with EINVAL on unrecognized data
- `recvmsg()` returns `msg_controllen = 0` if no ancillary data was attached to this packet

Thus, I should implement the "ignore all ancillary data" in the similar manner in this PR. 

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1210)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPbXO9NBD_PXNAU0i9h:-NPqtX5qCYW6GScskPkX:b9fuo99) (no related file):*

I experimented with TCP and different ancillary data types:

### `sendmsg()`

1. If `msg_control` is set to an empty buffer, then `sendmsg()` returns EINVAL, triggering this case: https://elixir.bootlin.com/linux/v6.2/source/net/core/sock.c#L2794
2. If `msg_control` is set to `SCM_RIGHTS`, then `sendmsg()` sends the packet and **ignores** ancillary data, triggering this case: https://elixir.bootlin.com/linux/v6.2/source/net/core/sock.c#L2792
3. If `msg_control` is set to some non-socket level (i.e. `cmsg_level = IPPROTO_IP`), then **ignores** this ancillary data; this case: https://elixir.bootlin.com/linux/v6.2/source/net/core/sock.c#L2810
4. If `msg_control` is set to socket-specific `SO_TIMESTAMPING`, then `sendmsg()` uses this ancillary data, triggering this case: https://elixir.bootlin.com/linux/v6.2/source/net/core/sock.c#L2780

### `recvmsg()`

1. If `sendmsg()` sent no ancillary data, `recvmsg()` gets nothing (`msg_controllen = 0`)
2. If `sendmsg()` sent `SCM_RIGHTS`, `recvmsg()` gets nothing (`msg_controllen = 0`)
3. If `sendmsg()` sent `IPPROTO_IP`, `recvmsg()` gets nothing (`msg_controllen = 0`)
4. If client socket has `setsockopt(SOL_SOCKET, SO_TIMESTAMPING...)` with `SOF_TIMESTAMPING_RX_SOFTWARE | SOF_TIMESTAMPING_SOFTWARE`, `recvmsg()` gets `SO_TIMESTAMPING_OLD` ancillary data (with `msg_controllen = 24`)

### Conclusion

TCP/IP sockets in Linux:
- can send and receive ancillary data (I experimented with `SO_TIMESTAMPING`)
- `sendmsg()` ignores UDS-specific ancillary data or non-socket-level data and fails with EINVAL on unrecognized data
- `recvmsg()` returns `msg_controllen = 0` if no ancillary data was attached to this packet

Thus, I should implement the "ignore all ancillary data" in the similar manner in this PR. 
</blockquote></details>

Btw, a good resource for this timestamping functionality: https://www.kernel.org/doc/Documentation/networking/timestamping.txt

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1210)* status: 0 of 11 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPrauMR2nVJGeGveC7Q:-NPrauMR2nVJGeGveC7R:b-ro8pxc) (no related file):*
FYI: This PR is completely rewritten now (as of revision 3). It adds proper support (though everything is still resulting in error or is ignored).

I went deep into the Linux sources, and imitated what Linux does. Here are the relevant sources:
- https://elixir.bootlin.com/linux/latest/source/include/linux/socket.h
- https://elixir.bootlin.com/linux/latest/source/include/uapi/asm-generic/socket.h
- https://elixir.bootlin.com/linux/latest/source/include/net/scm.h
- https://elixir.bootlin.com/linux/latest/source/net/ipv4/tcp.c
- https://elixir.bootlin.com/linux/latest/source/net/ipv4/udp.c
- https://elixir.bootlin.com/linux/latest/source/net/ipv4/ip_sockglue.c
- https://elixir.bootlin.com/linux/latest/source/net/unix/af_unix.c
- https://elixir.bootlin.com/linux/v6.2/source/net/core/sock.c
- https://elixir.bootlin.com/linux/latest/source/net/core/scm.c

___

stefanberger (commented):
> Reviewed 11 of 11 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1210)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

stefanberger (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1210#-:-NPrk2ciAKccgRFc5HU6:bnfp4nl)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1210)* status: 10 of 12 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow and @stefanberger)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPvSUGIB0NuQ_whL7BQ:-NPvSUGIB0NuQ_whL7BR:b-oys7w8) (no related file):*
FYI: This PR is ready for review. Because the PR was reworked completely for revision 3, please review from scratch (ignore revisions 1 and 2).

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1210)* status: 10 of 12 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow and @stefanberger)

___
*[`libos/test/ltp/ltp.cfg` line 1666 at r4](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPvVc4CEFNZVWcsgJ_4:-NPvVc4CEFNZVWcsgJ_5:br1hspl) ([raw file](https://github.com/gramineproject/gramine/blob/749005e25802e51229576553065e74fb530a83e4/libos/test/ltp/ltp.cfg#L1666)):*
> ```INI
> # subtest 8 requires SCM_RIGHTS support (to send an FD from sender to receiver) and doesn't check
> # for errors (Gramine fails with -ENOSYS), leading to a hang on the receiver side because it tries
> # to receive something that was never sent.
> ```

Btw, I created an issue in the LTP github repo: https://github.com/linux-test-project/ltp/issues/1020

Not sure though this is a "real enough" issue, and that LTP folks will fix it

mkow (changes requested):
> Reviewed 10 of 11 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1210)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPbXO9NBD_PXNAU0i9h:-NPto1iWE2xL5AU9Fv4f:b-nvbqqg) (no related file):*
Heh, I see a lot happened in the meantime :)

Regarding my original comment: The message I got from this PR originally was that this is a hack to make swtpm work and it's not how Linux works, thus my original blocking comment.

> I experimented with TCP and different ancillary data types:
> [...]

Jeez, what an ugly interface...

___
___
*[`-- commits` line 12 at r4](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPtxC4e3GLlxocBIShX):*




_Suggestion:_
```
    - SCM_RIGHTS/SCM_CREDENTIALS are ignored in TCP/UDP sockets (same as on Linux);
```
___
*[`libos/src/fs/socket/fs.c` line 39 at r4](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPyYon--80U23FByNIb:-NPyYon--80U23FByNIc:b-c9e13y) ([raw file](https://github.com/gramineproject/gramine/blob/749005e25802e51229576553065e74fb530a83e4/libos/src/fs/socket/fs.c#L39)):*
> ```C
>     };
>     unsigned int flags = 0;
>     return do_recvmsg(handle, &iov, /*iov_len=*/1, /*msg_control=*/NULL, /*msg_controllen=*/NULL,
> ```

Could we rename it to `msg_controllen_ptr` or something in this style? It's quite confusing to see things like `msg_controllen = 0` almost next to `msg_controllen = NULL` when reading the socket code, because it looks like a bug. This is especially visible in this flie.
And maybe the same with `addrlen`?



_Code quote:_
```C
/*msg_controllen=*/NULL
```
___
*[`libos/src/net/ip.c` line 682 at r4](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPyxHnZ50NHKZCSrrQz:-NPyxHnZ50NHKZCSrrR-:b-dy6l41) ([raw file](https://github.com/gramineproject/gramine/blob/749005e25802e51229576553065e74fb530a83e4/libos/src/net/ip.c#L682)):*
> ```C
>         if (cmsg->cmsg_level != SOL_SOCKET) {
>             /*
>              * Currently don't support:
> ```

(it's in imperative mood otherwise)



_Suggestion:_
```C
We currently don't support:
```
___
*[`libos/src/net/ip.c` line 693 at r4](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPyxuKc-puLJ0DrcQ9u:-NPyxuKd6Sw47bZeVWOC:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/749005e25802e51229576553065e74fb530a83e4/libos/src/net/ip.c#L693)):*
> ```C
> 
>         switch (cmsg->cmsg_type) {
>             /* currently don't support below SOL_SOCKET types */
> ```

ditto
___
*[`libos/src/net/ip.c` line 778 at r4](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPyyKwo-AGphAlH3kbf:-NPyyKwo-AGphAlH3kbg:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/749005e25802e51229576553065e74fb530a83e4/libos/src/net/ip.c#L778)):*
> ```C
>     if (msg_control && msg_controllen) {
>         /*
>          * Currently don't support:
> ```

ditto
___
*[`libos/src/net/unix.c` line 435 at r4](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPyzpnlByA_NmF-Tyzd:-NPyzpnlByA_NmF-Tyze:bulc6r) ([raw file](https://github.com/gramineproject/gramine/blob/749005e25802e51229576553065e74fb530a83e4/libos/src/net/unix.c#L435)):*
> ```C
> 
>         rest_msg_controllen -= CMSG_ALIGN(cmsg->cmsg_len);
>         cmsg = (struct cmsghdr*)((char*)cmsg + CMSG_ALIGN(cmsg->cmsg_len));
> ```

Purely theoretical (I think) trivia: This cast to `struct cmsghdr*` is an UB in C, because it's possible that the resulting pointer won't point to neither a `struct cmsghdr` object nor right behind any such object if the `msg_control` array is not padded (I'm not sure if it has to be, probably not).

You could fix it by checking if the buffer we got is big enough to reach the `(char*)cmsg + CMSG_ALIGN(cmsg->cmsg_len)` address and executing this line only if it is, but probably not worth the effort.

I think C has this rule because lines like this one would be problematic if `msg_control` was allocated near the end of address space, because the addition could overflow the pointer then.
___
*[`libos/test/regression/tcp_ancillary.c` line 21 at r4](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPyeA7h9NxnBl0Ecofy:-NPyeA7h9NxnBl0Ecofz:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/749005e25802e51229576553065e74fb530a83e4/libos/test/regression/tcp_ancillary.c#L21)):*
> ```C
> #define PORT   11110
> 
> #define MSG_SPACE CMSG_SPACE(sizeof(int)) + CMSG_SPACE(sizeof(struct ucred))
> ```





_Suggestion:_
```C
#define MSG_SPACE (CMSG_SPACE(sizeof(int)) + CMSG_SPACE(sizeof(struct ucred)))
```
___
*[`libos/test/regression/tcp_ancillary.c` line 42 at r4](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPyf-N69uAQaAmnQ8oN:-NPyf-N69uAQaAmnQ8oO:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/749005e25802e51229576553065e74fb530a83e4/libos/test/regression/tcp_ancillary.c#L42)):*
> ```C
>     char c = 0;
>     ssize_t x = CHECK(write(pipefd, &c, sizeof(c)));
>     if (x != 1) {
> ```





_Suggestion:_
```C
if (x != sizeof(c))
```
___
*[`libos/test/regression/tcp_ancillary.c` line 117 at r4](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPys0j41Ttwm751oHHb:-NPys0j41Ttwm751oHHc:byuroe0) ([raw file](https://github.com/gramineproject/gramine/blob/749005e25802e51229576553065e74fb530a83e4/libos/test/regression/tcp_ancillary.c#L117)):*
> ```C
>         .sin_port = htons(PORT),
>         .sin_addr = {
>             /* TODO: remove this once Ubuntu 18.04 is deprecated. */
> ```

What exactly? The whole assignment of `.sin_addr`? Why? Could you explain this a bit more in the comment?

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1210)* status: all files reviewed, 10 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`libos/src/net/ip.c` line 677 at r4](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPzseyg5dJ-DYhx8X0K:-NPzseyg5dJ-DYhx8X0L:b-szk4fn) ([raw file](https://github.com/gramineproject/gramine/blob/749005e25802e51229576553065e74fb530a83e4/libos/src/net/ip.c#L677)):*
> ```C
>     size_t rest_msg_controllen = msg_controllen;
>     while (cmsg && rest_msg_controllen >= sizeof(struct cmsghdr)) {
>         if (cmsg->cmsg_len < sizeof(struct cmsghdr) || cmsg->cmsg_len > rest_msg_controllen)
> ```

Do we need to check against `CMSG_ALIGN(cmsg->cmsg_len)`?



_Code quote:_
```C
cmsg->cmsg_len
```

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1210)* status: 6 of 12 files reviewed, 10 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin, @mkow, and @stefanberger)

___
*[`-- commits` line 12 at r4](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPtxC4e3GLlxocBIShX):*
Yep, I'll add during final rebase
___
*[`libos/src/fs/socket/fs.c` line 39 at r4](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPyYon--80U23FByNIb:-NQ-y5EfAwAV7ow0Mhbs:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/749005e25802e51229576553065e74fb530a83e4/libos/src/fs/socket/fs.c#L39)):*

Could we rename it to `msg_controllen_ptr` or something in this style? It's quite confusing to see things like `msg_controllen = 0` almost next to `msg_controllen = NULL` when reading the socket code, because it looks like a bug. This is especially visible in this flie.
And maybe the same with `addrlen`?
</blockquote></details>

Done.
___
*[`libos/src/net/ip.c` line 677 at r4](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPzseyg5dJ-DYhx8X0K:-NQ-smuV4LZmi6WsA4iU:b-ryjfrq) ([raw file](https://github.com/gramineproject/gramine/blob/749005e25802e51229576553065e74fb530a83e4/libos/src/net/ip.c#L677)):*

Do we need to check against `CMSG_ALIGN(cmsg->cmsg_len)`?
</blockquote></details>

Done. You're right, otherwise we could have a broken corner case: `cmsg->cmsg_len == rest_msg_controllen` but `CMSG_ALIGN(cmsg->cmsg_len) > rest_msg_controllen` (e.g. if `cmsg->cmsg_len == 7`).

In that case, `rest_msg_controllen -= CMSG_ALIGN(cmsg->cmsg_len);` could result in integer underflow. Fixed now (by changing only the second part of this IF).
___
*[`libos/src/net/ip.c` line 682 at r4](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPyxHnZ50NHKZCSrrQz:-NQ-pIM919znTO4ivSwt:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/749005e25802e51229576553065e74fb530a83e4/libos/src/net/ip.c#L682)):*

(it's in imperative mood otherwise)
</blockquote></details>

Done.
___
*[`libos/src/net/ip.c` line 693 at r4](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPyxuKc-puLJ0DrcQ9u:-NQ-pL0IBO0dGRvyzGxp:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/749005e25802e51229576553065e74fb530a83e4/libos/src/net/ip.c#L693)):*

ditto
</blockquote></details>

Done.
___
*[`libos/src/net/ip.c` line 778 at r4](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPyyKwo-AGphAlH3kbf:-NQ-pO4b4B4pun_j3pRI:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/749005e25802e51229576553065e74fb530a83e4/libos/src/net/ip.c#L778)):*

ditto
</blockquote></details>

Done.
___
*[`libos/src/net/unix.c` line 435 at r4](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPyzpnlByA_NmF-Tyzd:-NQ-r-P8EXB5j-yJwbhm:b8m2i13) ([raw file](https://github.com/gramineproject/gramine/blob/749005e25802e51229576553065e74fb530a83e4/libos/src/net/unix.c#L435)):*

Purely theoretical (I think) trivia: This cast to `struct cmsghdr*` is an UB in C, because it's possible that the resulting pointer won't point to neither a `struct cmsghdr` object nor right behind any such object if the `msg_control` array is not padded (I'm not sure if it has to be, probably not).

You could fix it by checking if the buffer we got is big enough to reach the `(char*)cmsg + CMSG_ALIGN(cmsg->cmsg_len)` address and executing this line only if it is, but probably not worth the effort.

I think C has this rule because lines like this one would be problematic if `msg_control` was allocated near the end of address space, because the addition could overflow the pointer then.
</blockquote></details>

I agree with your assessment. I also stopped for a bit because of this `CMSG_ALIGN()` that could "shift" the `cmsg` pointer past the `msg_control` object boundary, which is a UB.

But @kailun-qin noticed above that there is still potential for integer overflow, see that comment. I fixed it now.
___
*[`libos/test/regression/tcp_ancillary.c` line 21 at r4](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPyeA7h9NxnBl0Ecofy:-NQ-pX7E7Hqz0ZpaPL-q:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/749005e25802e51229576553065e74fb530a83e4/libos/test/regression/tcp_ancillary.c#L21)):*
> ```C
> #define PORT   11110
> 
> #define MSG_SPACE CMSG_SPACE(sizeof(int)) + CMSG_SPACE(sizeof(struct ucred))
> ```

Done.
___
*[`libos/test/regression/tcp_ancillary.c` line 42 at r4](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPyf-N69uAQaAmnQ8oN:-NQ-pbbkDv8_ns3R2HZY:b-191a4p) ([raw file](https://github.com/gramineproject/gramine/blob/749005e25802e51229576553065e74fb530a83e4/libos/test/regression/tcp_ancillary.c#L42)):*
> ```C
>     char c = 0;
>     ssize_t x = CHECK(write(pipefd, &c, sizeof(c)));
>     if (x != 1) {
> ```

Done. Wow, that's some pedantry :)
___
*[`libos/test/regression/tcp_ancillary.c` line 117 at r4](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPys0j41Ttwm751oHHb:-NQ-q9DYDcGJOrwBzFlZ:bhosmk4) ([raw file](https://github.com/gramineproject/gramine/blob/749005e25802e51229576553065e74fb530a83e4/libos/test/regression/tcp_ancillary.c#L117)):*

What exactly? The whole assignment of `.sin_addr`? Why? Could you explain this a bit more in the comment?
</blockquote></details>

Yes, remove this whole assignment of `.sin_addr` (because it would be implicitly set to zero anyway).

There is some weird bug in Ubuntu 18.04, and nobody wanted to care about this corner case. See the discussion starting here: https://github.com/gramineproject/gramine/pull/684#pullrequestreview-1028245091

(Note that I based this new test on the already-existing `tcp_msg_peek.c` test, where the same TODO can be seen.)

If you want to, I can change the text of the TODO, but I'm not sure to what.

mkow (changes requested):
> Reviewed 6 of 6 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1210)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @kailun-qin)

___
*[`libos/src/net/unix.c` line 435 at r4](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPyzpnlByA_NmF-Tyzd:-NQ0R47xCC1iQmNBa_ID:b5qo5ff) ([raw file](https://github.com/gramineproject/gramine/blob/749005e25802e51229576553065e74fb530a83e4/libos/src/net/unix.c#L435)):*

I agree with your assessment. I also stopped for a bit because of this `CMSG_ALIGN()` that could "shift" the `cmsg` pointer past the `msg_control` object boundary, which is a UB.

But @kailun-qin noticed above that there is still potential for integer overflow, see that comment. I fixed it now.
</blockquote></details>

Yup, seems to be the same issue, but I didn't notice that it actually leads to underflowing `rest_msg_controllen`. So, turns out it wasn't purely theoretical after all, although because of a different part of the code :)
___
*[`libos/test/regression/tcp_ancillary.c` line 42 at r4](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPyf-N69uAQaAmnQ8oN:-NQ0PNj80PSbDz5WAydj:bbadedk) ([raw file](https://github.com/gramineproject/gramine/blob/749005e25802e51229576553065e74fb530a83e4/libos/test/regression/tcp_ancillary.c#L42)):*

Done. Wow, that's some pedantry :)
</blockquote></details>

I disagree ;) This is how I check such places for correctness when reviewing: I read such constructions and check if there's a comparison between the return value and the exact expression which was used as a size for I/O. And here I had to stop and think because it was different ;)
___
*[`libos/test/regression/tcp_ancillary.c` line 117 at r4](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPys0j41Ttwm751oHHb:-NQ0Q-Zm1ixpG6E-YX7x:bq21eva) ([raw file](https://github.com/gramineproject/gramine/blob/749005e25802e51229576553065e74fb530a83e4/libos/test/regression/tcp_ancillary.c#L117)):*

Yes, remove this whole assignment of `.sin_addr` (because it would be implicitly set to zero anyway).

There is some weird bug in Ubuntu 18.04, and nobody wanted to care about this corner case. See the discussion starting here: https://github.com/gramineproject/gramine/pull/684#pullrequestreview-1028245091

(Note that I based this new test on the already-existing `tcp_msg_peek.c` test, where the same TODO can be seen.)

If you want to, I can change the text of the TODO, but I'm not sure to what.
</blockquote></details>

Oh, so it's about GCC, not Ubuntu. Actually, we already started dropping 18.04, maybe just remove this line now?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1210)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @kailun-qin, and @mkow)

___
*[`libos/test/regression/tcp_ancillary.c` line 117 at r4](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPys0j41Ttwm751oHHb:-NQ0_a5JCnZ5DbIwGlPe:bulp1dt) ([raw file](https://github.com/gramineproject/gramine/blob/749005e25802e51229576553065e74fb530a83e4/libos/test/regression/tcp_ancillary.c#L117)):*

Oh, so it's about GCC, not Ubuntu. Actually, we already started dropping 18.04, maybe just remove this line now?
</blockquote></details>

What do you mean by `started dropping 18.04`? It's in the CI (e.g. `Jenkins-18.04`), so the current PR will fail if I remove this line.

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1210)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @kailun-qin)

___
*[`libos/test/regression/tcp_ancillary.c` line 117 at r4](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPys0j41Ttwm751oHHb:-NQ0c5uqEplzpo-MhUMr:bsyeh9u) ([raw file](https://github.com/gramineproject/gramine/blob/749005e25802e51229576553065e74fb530a83e4/libos/test/regression/tcp_ancillary.c#L117)):*

What do you mean by `started dropping 18.04`? It's in the CI (e.g. `Jenkins-18.04`), so the current PR will fail if I remove this line.
</blockquote></details>

We're after v1.4, which was the last version which supports it and I think we already started removing some compat. But yeah, we need to disable the CI pipelines first.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1210)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

___
*[`-- commits` line 12 at r4](https://reviewable.io/reviews/gramineproject/gramine/1210#-NPtxC4e3GLlxocBIShX):*

Yep, I'll add during final rebase
</blockquote></details>

Done

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1210)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #1211, opened 2023-03-06
Fix a bunch of syscall arguments checking

mkow:
## Description of the changes 

This PR fixes a bunch of unrelated checks which were missing from various syscalls. These cases were found by Jaewon Hur using an adapted version of syzkaller, thanks!

Just for the record: none of these bugs is security related, because both sides of the emulated syscall interface are in the same security domain.

## How to test this PR? 

See each commit for the specific case where it failed. No tests added to CI, as all of the cases were adversarial and shouldn't be possible in normal apps.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1211)

### DIFF
--- a/libos/src/sys/libos_getcwd.c
+++ b/libos/src/sys/libos_getcwd.c
@@ -88,6 +88,10 @@ long libos_syscall_fchdir(int fd) {
 
     struct libos_dentry* dent = hdl->dentry;
 
+    if (!dent) {
+        log_debug("FD=%d has no path in the filesystem", fd);
+        return -ENOTDIR;
+    }
     if (!dent->inode || dent->inode->type != S_IFDIR) {
         char* path = NULL;
         dentry_abs_path(dent, &path, /*size=*/NULL);
--- a/libos/src/sys/libos_mmap.c
+++ b/libos/src/sys/libos_mmap.c
@@ -44,10 +44,32 @@
                        | MAP_HUGE_2MB           \
                        | MAP_HUGE_1GB)
 
+static int check_prot(int prot) {
+    if (prot & ~(PROT_NONE | PROT_READ | PROT_WRITE | PROT_EXEC | PROT_GROWSDOWN | PROT_GROWSUP |
+                 PROT_SEM)) {
+        return -EINVAL;
+    }
+
+    if ((prot & (PROT_GROWSDOWN | PROT_GROWSUP)) == (PROT_GROWSDOWN | PROT_GROWSUP)) {
+        return -EINVAL;
+    }
+
+    /* We do not support these flags (at least yet). */
+    if (prot & (PROT_GROWSUP | PROT_SEM)) {
+        return -EOPNOTSUPP;
+    }
+
+    return 0;
+}
+
 void* libos_syscall_mmap(void* addr, size_t length, int prot, int flags, int fd,
                          unsigned long offset) {
     struct libos_handle* hdl = NULL;
     long ret = 0;
+
+    ret = check_prot(prot);
+    if (ret < 0)
+        return (void*)ret;
 
     if (!(flags & MAP_FIXED) && addr)
         addr = ALLOC_ALIGN_DOWN_PTR(addr);
@@ -242,19 +264,9 @@ out_handle:
 }
 
 long libos_syscall_mprotect(void* addr, size_t length, int prot) {
-    if (prot & ~(PROT_NONE | PROT_READ | PROT_WRITE | PROT_EXEC | PROT_GROWSDOWN | PROT_GROWSUP |
-                 PROT_SEM)) {
-        return -EINVAL;
-    }
-
-    if ((prot & (PROT_GROWSDOWN | PROT_GROWSUP)) == (PROT_GROWSDOWN | PROT_GROWSUP)) {
-        return -EINVAL;
-    }
-
-    /* We do not support these flags (at least yet). */
-    if (prot & (PROT_GROWSUP | PROT_SEM)) {
-        return -EOPNOTSUPP;
-    }
+    int ret = check_prot(prot);
+    if (ret < 0)
+        return ret;
 
     /*
      * According to the manpage, addr has to be page-aligned, but not the
@@ -277,7 +289,7 @@ long libos_syscall_mprotect(void* addr, size_t length, int prot) {
     /* `bkeep_mprotect` and then `PalVirtualMemoryProtect` is racy, but it's hard to do it properly.
      * On the other hand if this race happens, it means user app is buggy, so not a huge problem. */
 
-    int ret = bkeep_mprotect(addr, length, prot, /*is_internal=*/false);
+    ret = bkeep_mprotect(addr, length, prot, /*is_internal=*/false);
     if (ret < 0) {
         return ret;
     }
--- a/libos/src/sys/libos_open.c
+++ b/libos/src/sys/libos_open.c
@@ -106,6 +106,11 @@ long libos_syscall_openat(int dfd, const char* filename, int flags, int mode) {
              | O_LARGEFILE | O_NOATIME | O_NOCTTY | O_NOFOLLOW | O_NONBLOCK | O_PATH | O_SYNC
              | O_TMPFILE | O_TRUNC;
 
+    if (   (flags & O_ACCMODE) != O_RDONLY
+        && (flags & O_ACCMODE) != O_WRONLY
+        && (flags & O_ACCMODE) != O_RDWR)
+        return -EINVAL;
+
     /* TODO: fail explicitly on valid but unsupported flags. */
 
     if (flags & O_PATH) {
--- a/libos/src/sys/libos_sched.c
+++ b/libos/src/sys/libos_sched.c
@@ -49,7 +49,10 @@ long libos_syscall_getpriority(int which, int who) {
 
 /* dummy implementation: ignore user-supplied param and return success */
 long libos_syscall_sched_setparam(pid_t pid, struct __kernel_sched_param* param) {
-    if (pid < 0 || param == NULL)
+    if (!is_user_memory_readable(param, sizeof(*param)))
+        return -EFAULT;
+
+    if (pid < 0)
         return -EINVAL;
 
     return 0;
@@ -57,7 +60,10 @@ long libos_syscall_sched_setparam(pid_t pid, struct __kernel_sched_param* param)
 
 /* dummy implementation: always return sched_priority of 0 (implies non-real-time sched policy) */
 long libos_syscall_sched_getparam(pid_t pid, struct __kernel_sched_param* param) {
-    if (pid < 0 || param == NULL)
+    if (!is_user_memory_writable(param, sizeof(*param)))
+        return -EFAULT;
+
+    if (pid < 0)
         return -EINVAL;
 
     param->__sched_priority = 0;
@@ -68,7 +74,10 @@ long libos_syscall_sched_getparam(pid_t pid, struct __kernel_sched_param* param)
 long libos_syscall_sched_setscheduler(pid_t pid, int policy, struct __kernel_sched_param* param) {
     policy &= ~SCHED_RESET_ON_FORK; /* ignore reset-on-fork flag */
 
-    if (pid < 0 || param == NULL)
+    if (!is_user_memory_readable(param, sizeof(*param)))
+        return -EFAULT;
+
+    if (pid < 0)
         return -EINVAL;
 
     /* fail on unrecognized policies */
--- a/libos/src/sys/libos_wait.c
+++ b/libos/src/sys/libos_wait.c
@@ -105,17 +105,20 @@ static long do_waitid(int which, pid_t id, siginfo_t* infop, int options) {
     }
 
     if (options & WSTOPPED) {
-        log_warning("Ignoring unsupported WSTOPPED flag to wait4");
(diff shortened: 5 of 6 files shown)


dimakuv (changes requested):
> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, 1 of 1 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1211)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @mkow)

___
*[`-- commits` line 12 at r4](https://reviewable.io/reviews/gramineproject/gramine/1211#-NPqEeo3BsIsby3hv_Bs):*
This commit looks good
___
*[`-- commits` line 17 at r4](https://reviewable.io/reviews/gramineproject/gramine/1211#-NPqEfOq1_ueJX6dhRCm):*
This commit looks good
___
*[`-- commits` line 22 at r5](https://reviewable.io/reviews/gramineproject/gramine/1211#-NPqIC0V2OpHevmEvNFo):*
This commit looks good
___
*[`-- commits` line 27 at r6](https://reviewable.io/reviews/gramineproject/gramine/1211#-NPqImjR0taS1IDS-X7l):*
This commit looks good
___
*[`libos/src/fs/chroot/fs.c` line 214 at r2](https://reviewable.io/reviews/gramineproject/gramine/1211#-NPqCZtx1SXFTod7RokW:-NPqCZty0MNgGKQycZYv:b-v9l37t) ([raw file](https://github.com/gramineproject/gramine/blob/0904000539bc8cbdaa75f7a6ca9b86231a2d337c/libos/src/fs/chroot/fs.c#L214)):*
> ```C
>     if (   (flags & O_ACCMODE) != PAL_ACCESS_RDONLY
>         && (flags & O_ACCMODE) != PAL_ACCESS_WRONLY
>         && (flags & O_ACCMODE) != PAL_ACCESS_RDWR)
> ```

These `flags` has Linux values, not PAL values?

By the way, maybe we'll have to add `O_APPEND` check here as well -- we don't support it currently but we ignore it everywhere. I think if we'll not have this flag allowed, then some workloads may break. (This is of course bad that Gramine silently ignores this unsupported flag, but if some workloads currently depend on it, I would prefer to **not** break them, and fix this APPEND issues properly in a follow up.)
___
*[`libos/src/fs/chroot/fs.c` line 215 at r2](https://reviewable.io/reviews/gramineproject/gramine/1211#-NPqD-n14W8pEOPiWMFg:-NPqD-n22Qwzq-uiMdM_:b-gg1v3) ([raw file](https://github.com/gramineproject/gramine/blob/0904000539bc8cbdaa75f7a6ca9b86231a2d337c/libos/src/fs/chroot/fs.c#L215)):*
> ```C
>         && (flags & O_ACCMODE) != PAL_ACCESS_WRONLY
>         && (flags & O_ACCMODE) != PAL_ACCESS_RDWR)
>         return -EINVAL;
> ```

Why is this check here? This check should apply to all FS types, not only to `chroot`.

I think the check should be in either of these two places:
1. https://github.com/gramineproject/gramine/blob/43b182e7f5e4f0ff62e37954122241c1a4d4fa28/libos/src/sys/libos_open.c#L109
2. https://github.com/gramineproject/gramine/blob/0119fd23ebbc892b33868e9ea9296ba6e8061685/libos/src/fs/libos_namei.c#L428

---

By the way, I found this problematic case: https://github.com/gramineproject/gramine/blob/0119fd23ebbc892b33868e9ea9296ba6e8061685/libos/src/bookkeep/libos_handle.c#L35

`O_APPEND` is ignored in Gramine, so if you'll add your check at the more "generic" level, you'll hit this problem. The fix/workaround is to remove `O_APPEND` from this line and to add a comment that `O_APPEND` should be re-added once it is supported (or just ignore it? this is stdout/stderr, so it is always appended, regardless of flags).
___
*[`libos/src/sys/libos_wait.c` line 118 at r1](https://reviewable.io/reviews/gramineproject/gramine/1211#-NPqAwtw6ZepAEgLByLY:-NPqAwtw6ZepAEgLByLZ:bxloi2s) ([raw file](https://github.com/gramineproject/gramine/blob/0d2ca906ec9e4459aae984254f843840e38e0962/libos/src/sys/libos_wait.c#L118)):*
> ```C
>         log_error("Unsupported combination of flags passed to waitid/wait4");
>         return -EINVAL;
>     }
> ```

This check doesn't belong here. Here we should actually keep the assert. This is because this is a helper function that is not supposed to check the input syscall args like this -- these checks should be done in `libos_syscall_waitid` and `libos_syscall_wait4` functions.

See my other comment where this check should be implemented.
___
*[`libos/src/sys/libos_wait.c` line 228 at r1](https://reviewable.io/reviews/gramineproject/gramine/1211#-NPqB52kEHnLQgWUyTNS:-NPqB52kEHnLQgWUyTNT:bz41g34) ([raw file](https://github.com/gramineproject/gramine/blob/0d2ca906ec9e4459aae984254f843840e38e0962/libos/src/sys/libos_wait.c#L228)):*
> ```C
>     if (!(options & (WEXITED | WSTOPPED | WCONTINUED)))
>         return -EINVAL;
> ```

We should add a check here that `WEXITED` must be specified, since without it we don't support `waitid()`.

This way all checks will be grouped in the correct functions, and also there will be a cleaner distinction between different treatment of `options` in `waitid()` and `wait4()`.
___
*[`libos/src/sys/libos_wait.c` line 250 at r1](https://reviewable.io/reviews/gramineproject/gramine/1211#-NPqBSbY93MUy6ZILAby:-NPqBSbY93MUy6ZILAbz:b-1r9lbl) ([raw file](https://github.com/gramineproject/gramine/blob/0d2ca906ec9e4459aae984254f843840e38e0962/libos/src/sys/libos_wait.c#L250)):*
> ```C
> 
>     /* Prepare options for do_waitid(). */
>     options |= WEXITED;
> ```

Note that `wait4` implementation **always sets** WEXITED, so that's why (I think) Borys added an `assert()` in the helper function. But Borys forgot about a corner case in `waitid()` implementation.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1211)* status: 5 of 7 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "squash! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`libos/src/sys/libos_wait.c` line 118 at r1](https://reviewable.io/reviews/gramineproject/gramine/1211#-NPqAwtw6ZepAEgLByLY:-NPqo-wQEK0EG8USKELI:b-s0zzzn) ([raw file](https://github.com/gramineproject/gramine/blob/0d2ca906ec9e4459aae984254f843840e38e0962/libos/src/sys/libos_wait.c#L118)):*
> This is because this is a helper function that is not supposed to check the input syscall args like this

Hmm, why? This check is not part of Linux API, it's here only because our helper function `do_waitid()` doesn't yet work with other flags. So, it seems to me that it should be placed here?
___
*[`libos/src/sys/libos_wait.c` line 250 at r1](https://reviewable.io/reviews/gramineproject/gramine/1211#-NPqBSbY93MUy6ZILAby:-NPqoTlb0vHGaRYZhzHG:bbnr0kb) ([raw file](https://github.com/gramineproject/gramine/blob/0d2ca906ec9e4459aae984254f843840e38e0962/libos/src/sys/libos_wait.c#L250)):*

Note that `wait4` implementation **always sets** WEXITED, so that's why (I think) Borys added an `assert()` in the helper function. But Borys forgot about a corner case in `waitid()` implementation.
</blockquote></details>

Yes, I also think that's how it happened.
___
*[`libos/src/fs/chroot/fs.c` line 214 at r2](https://reviewable.io/reviews/gramineproject/gramine/1211#-NPqCZtx1SXFTod7RokW:-NPqeQ-D1m74PFbBDaBa:bhz6xt6) ([raw file](https://github.com/gramineproject/gramine/blob/0904000539bc8cbdaa75f7a6ca9b86231a2d337c/libos/src/fs/chroot/fs.c#L214)):*

These `flags` has Linux values, not PAL values?

By the way, maybe we'll have to add `O_APPEND` check here as well -- we don't support it currently but we ignore it everywhere. I think if we'll not have this flag allowed, then some workloads may break. (This is of course bad that Gramine silently ignores this unsupported flag, but if some workloads currently depend on it, I would prefer to **not** break them, and fix this APPEND issues properly in a follow up.)
</blockquote></details>

Yeah, of course, sorry. Copy paste error when moving the flags from `LINUX_OPEN_FLAGS_TO_PAL_ACCESS()`. Fixed.
I guess it passed the tests by accident - these PAL macros accidentally have the same values as `O_*` ones.
___
*[`libos/src/fs/chroot/fs.c` line 215 at r2](https://reviewable.io/reviews/gramineproject/gramine/1211#-NPqD-n14W8pEOPiWMFg:-NPqmayY37ZZGvMs3i6E:b8gnxsw) ([raw file](https://github.com/gramineproject/gramine/blob/0904000539bc8cbdaa75f7a6ca9b86231a2d337c/libos/src/fs/chroot/fs.c#L215)):*
> This check should apply to all FS types, not only to `chroot`.

Right, moved.

> By the way, I found this problematic case: 

Why is it problematic? I don't get it.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1211)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "squash! " found in commit messages' one-liners

___
*[`libos/src/sys/libos_wait.c` line 118 at r1](https://reviewable.io/reviews/gramineproject/gramine/1211#-NPqAwtw6ZepAEgLByLY:-NPqtMhMDwmXJSZ0-n5V:b-cftnp) ([raw file](https://github.com/gramineproject/gramine/blob/0d2ca906ec9e4459aae984254f843840e38e0962/libos/src/sys/libos_wait.c#L118)):*

> This is because this is a helper function that is not supposed to check the input syscall args like this

Hmm, why? This check is not part of Linux API, it's here only because our helper function `do_waitid()` doesn't yet work with other flags. So, it seems to me that it should be placed here?
</blockquote></details>

Ok, by this logic it's fine.
___
*[`libos/src/fs/chroot/fs.c` line 215 at r2](https://reviewable.io/reviews/gramineproject/gramine/1211#-NPqD-n14W8pEOPiWMFg:-NPqsdjU2JZ4Z-exie8t:b-fvjrim) ([raw file](https://github.com/gramineproject/gramine/blob/0904000539bc8cbdaa75f7a6ca9b86231a2d337c/libos/src/fs/chroot/fs.c#L215)):*
> Why is it problematic? I don't get it.

Because inside Gramine codebase, we use a flag (`O_APPEND`) that we know is not supported. This is an inconsistent state of code, which may bite us at some point. Like, if you would move your check to `open_namei()`, then you would hit the issue, because this line would fail: https://github.com/gramineproject/gramine/blob/0119fd23ebbc892b33868e9ea9296ba6e8061685/libos/src/bookkeep/libos_handle.c#L36

(But since you chose to move the check to `libos_syscall_openat()`, we're fine. Ok, I'll resolve though I don't like that we have a clearly unsupported flag without any comment.)

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1211)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "squash! " found in commit messages' one-liners

___
*[`libos/src/fs/chroot/fs.c` line 215 at r2](https://reviewable.io/reviews/gramineproject/gramine/1211#-NPqD-n14W8pEOPiWMFg:-NPtR2i0EUSM2EHp652z:blf1402) ([raw file](https://github.com/gramineproject/gramine/blob/0904000539bc8cbdaa75f7a6ca9b86231a2d337c/libos/src/fs/chroot/fs.c#L215)):*
> Like, if you would move your check to `open_namei()`, then you would hit the issue, because this line would fail

It wouldn't? I still don't see why anything would fail because of that `O_APPEND`. Are you sure you aren't misreading this PR? Notice that masking with `O_ACCMODE`, maybe that's what you're missing?

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1211)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "squash! " found in commit messages' one-liners

___
*[`libos/src/fs/chroot/fs.c` line 215 at r2](https://reviewable.io/reviews/gramineproject/gramine/1211#-NPqD-n14W8pEOPiWMFg:-NPvWG7mCeqP30U6n0Tt:b-yinnts) ([raw file](https://github.com/gramineproject/gramine/blob/0904000539bc8cbdaa75f7a6ca9b86231a2d337c/libos/src/fs/chroot/fs.c#L215)):*
> Notice that masking with O_ACCMODE, maybe that's what you're missing?

Yep, sorry, that's what I was missing... Resolving.

dimakuv (approved):
> Reviewed 5 of 5 files at r8, 1 of 1 files at r9, 2 of 2 files at r10, 2 of 2 files at r11, 2 of 2 files at r12, 2 of 2 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1211)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #1235, opened 2023-03-16
Mitigate INTEL-SA-00615


kailun-qin:
### Description of the problem

Processor MMIO Stale Data Vulnerabilities (incl. 4 CVEs) are a class of memory-mapped I/O (MMIO) vulnerabilities that can expose data [1][2]. A malicious OS could map MMIO memory into the untrusted memory space of an enclave application (which can be used for e.g. parameter passing to/from ECALLs and OCALLs, or any external buffers that an enclave might use to communicate with its application). Then when the enclave writes to this memory, it could unintentionally propagate the stale (secret) data in its fill buffers into the uncore, where it could later be extracted by malicious software.

The 4 related CVEs are CVE-2022-21123 (SBDR), CVE-2022-21125 (SBDS), CVE-2022-21127 (SRBDS Update) and CVE-2022-21166 (DRPW). Among them, CVE-2022-21127 does not require SW mitigations for SGX; the latest processor microcode alone is sufficient. The other CVEs require both processor microcode and SW to mitigate. And since Device Register Partial Write (DRPW) aka CVE-2022-21166 is the only one that affected Intel® Xeon® Scalable processor family, we mainly focus on mitigating it in Gramine.

Naturally aligned 8-byte writes are not affected by this vulnerability, so as SW mitigation, we should make all untrusted memory writes 8-byte wide and naturally aligned. Specifically in Gramine, we'll need to check and update all external writes (via e.g., https://github.com/gramineproject/gramine/blob/634d0392c3acec724dad5a6af8e6305f166eca57/pal/src/host/linux-sgx/enclave_api.h#L21, or https://github.com/gramineproject/gramine/blob/634d0392c3acec724dad5a6af8e6305f166eca57/common/include/api.h#L314) to instead use the safe version (in 8-byte chunks aligned to 8-bytes boundary).

[1] https://www.intel.com/content/www/us/en/security-center/advisory/intel-sa-00615.html
[2] https://www.intel.com/content/www/us/en/developer/articles/technical/software-security-guidance/technical-documentation/processor-mmio-stale-data-vulnerabilities.html

cc @mkow @dimakuv if anything missed or incorrect

### Gramine commit hash

634d0392c3acec724dad5a6af8e6305f166eca57

kailun-qin:
Let me work on this. Thanks!

dimakuv:
[...]

PR #1237, opened 2023-03-17
[PAL/Linux-SGX] Mitigate CVE-2022-21166 aka INTEL-SA-00615

kailun-qin:
A malicious OS could map MMIO memory into the untrusted memory space of an enclave application (which can be used for e.g. parameter passing to/from ECALLs and OCALLs, or any external buffers that an enclave might use to communicate with its application). Then when the enclave writes to this memory, it could unintentionally propagate the stale (secret) data in its core fill buffers into the uncore, where it could later be extracted by malicious software.

Naturally aligned 8-byte writes are not affected by this vulnerability, so as SW mitigation, this commit makes all untrusted memory writes 8-byte wide and naturally aligned.

There is a side effect introduced to the Exitless (aka RPC-threads aka Switchless) feature by the fix of CVE-2022-21233 aka INTEL-SA-00657, which marks the feature as insecure and didn't fix it due to its complex usage of the untrusted memory. We assume the same for this fix.

Fixes https://github.com/gramineproject/gramine/issues/1235.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1237)

### DIFF
--- a/Documentation/manifest-syntax.rst
+++ b/Documentation/manifest-syntax.rst
@@ -590,8 +590,10 @@ more CPU cores and burning more CPU cycles. For example, a single-threaded
 Redis instance on Linux becomes 5-threaded on Gramine with Exitless. Thus,
 Exitless may negatively impact throughput but may improve latency.
 
-This feature is currently marked as insecure, because it reads untrusted memory
-in potentially insecure manner - susceptible to CVE-2022-21233 (INTEL-SA-00657).
+This feature is currently marked as insecure, because it reads and writes to
+untrusted memory in potentially insecure manner - susceptible to
+CVE-2022-21233 (INTEL-SA-00657) and CVE-2022-21166 (INTEL-SA-00615)
+respectively.
 
 Optional CPU features (AVX, AVX512, MPX, PKRU, AMX, EXINFO)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/Documentation/performance.rst
+++ b/Documentation/performance.rst
@@ -164,7 +164,8 @@ Exitless feature
 ----------------
 
 Note this feature is currently insecure and not recommended for production
-usage (potentially susceptible to CVE-2022-21233 aka INTEL-SA-00657).
+usage (potentially susceptible to CVE-2022-21233 aka INTEL-SA-00657 and
+CVE-2022-21166 aka INTEL-SA-00615).
 
 Gramine supports the Exitless (or Switchless) feature – it trades off CPU cores
 for faster OCALL execution. More specifically, with Exitless, enclave threads do
--- a/pal/src/host/linux-sgx/enclave_api.h
+++ b/pal/src/host/linux-sgx/enclave_api.h
@@ -18,15 +18,21 @@ void sgx_reset_ustack(const void* old_ustack);
 
 void sgx_copy_to_enclave_verified(void* ptr, const void* uptr, size_t size);
 bool sgx_copy_to_enclave(void* ptr, size_t maxsize, const void* uptr, size_t usize);
-bool sgx_copy_from_enclave(void* urts_ptr, const void* enclave_ptr, size_t size);
+void sgx_copy_from_enclave_verified(void* uptr, const void* ptr, size_t size);
+bool sgx_copy_from_enclave(void* uptr, const void* ptr, size_t size);
 void* sgx_import_array_to_enclave(const void* uptr, size_t elem_size, size_t elem_cnt);
 void* sgx_import_array2d_to_enclave(const void* uptr, size_t elem_size, size_t elem_cnt1,
                                     size_t elem_cnt2);
 
 #define COPY_UNTRUSTED_VALUE(untrusted_ptr) ({                          \
     __typeof__(*(untrusted_ptr)) val;                                   \
-    sgx_copy_to_enclave_verified(&val, (untrusted_ptr), sizeof(val));   \
+    sgx_copy_to_enclave_verified(&val, untrusted_ptr, sizeof(val));     \
     val;                                                                \
+})
+
+#define COPY_VALUE_TO_UNTRUSTED(untrusted_ptr, val) ({                            \
+    __typeof__(*(untrusted_ptr)) src_val = (val);                                 \
+    sgx_copy_from_enclave_verified(untrusted_ptr, &src_val, sizeof(src_val));     \
 })
 
 /*!
--- a/pal/src/host/linux-sgx/enclave_framework.c
+++ b/pal/src/host/linux-sgx/enclave_framework.c
@@ -120,7 +120,7 @@ void* sgx_copy_to_ustack(const void* ptr, size_t size) {
     }
     void* uptr = sgx_alloc_on_ustack(size);
     if (uptr) {
-        memcpy(uptr, ptr, size);
+        sgx_copy_from_enclave_verified(uptr, ptr, size);
     }
     return uptr;
 }
@@ -130,14 +130,25 @@ void sgx_reset_ustack(const void* old_ustack) {
     UPDATE_USTACK(old_ustack);
 }
 
-static void copy_u64s(void* dst, const void* untrusted_src, size_t count) {
-    assert((uintptr_t)untrusted_src % 8 == 0);
+static void copy_u64s(void* dst, const void* src, size_t count) {
     __asm__ volatile (
         "rep movsq\n"
-        : "+D"(dst), "+S"(untrusted_src), "+c"(count)
+        : "+D"(dst), "+S"(src), "+c"(count)
         :
         : "memory", "cc"
     );
+}
+
+static void copy_u64s_from_untrusted(void* dst, const void* untrusted_src, size_t count) {
+    assert((uintptr_t)untrusted_src % 8 == 0);
+
+    copy_u64s(dst, untrusted_src, count);
+}
+
+static void copy_u64s_to_untrusted(void* untrusted_dst, const void* src, size_t count) {
+    assert((uintptr_t)untrusted_dst % 8 == 0);
+
+    copy_u64s(untrusted_dst, src, count);
 }
 
 void sgx_copy_to_enclave_verified(void* ptr, const void* uptr, size_t size) {
@@ -159,7 +170,7 @@ void sgx_copy_to_enclave_verified(void* ptr, const void* uptr, size_t size) {
     if (prefix_misalignment) {
         /* Beginning of the copied range is misaligned. */
         char prefix_val[8] = { 0 };
-        copy_u64s(prefix_val, (char*)uptr - prefix_misalignment, /*count=*/1);
+        copy_u64s_from_untrusted(prefix_val, (char*)uptr - prefix_misalignment, /*count=*/1);
 
         copy_len = MIN(sizeof(prefix_val) - prefix_misalignment, size);
         memcpy(ptr, prefix_val + prefix_misalignment, copy_len);
@@ -176,7 +187,7 @@ void sgx_copy_to_enclave_verified(void* ptr, const void* uptr, size_t size) {
     size_t suffix_misalignment = size & 7;
     copy_len = size - suffix_misalignment;
     assert(copy_len % 8 == 0);
-    copy_u64s(ptr, uptr, copy_len / 8);
+    copy_u64s_from_untrusted(ptr, uptr, copy_len / 8);
     ptr = (char*)ptr + copy_len;
     uptr = (const char*)uptr + copy_len;
     size -= copy_len;
@@ -185,7 +196,7 @@ void sgx_copy_to_enclave_verified(void* ptr, const void* uptr, size_t size) {
     if (suffix_misalignment) {
         /* End of the copied range is misaligned. */
         char suffix_val[8] = { 0 };
-        copy_u64s(suffix_val, uptr, /*count=*/1);
+        copy_u64s_from_untrusted(suffix_val, uptr, /*count=*/1);
         memcpy(ptr, suffix_val, suffix_misalignment);
     }
 }
@@ -201,12 +212,66 @@ bool sgx_copy_to_enclave(void* ptr, size_t maxsize, const void* uptr, size_t usi
     return true;
 }
 
-bool sgx_copy_from_enclave(void* urts_ptr, const void* enclave_ptr, size_t size) {
-    if (!sgx_is_valid_untrusted_ptr(urts_ptr, size, /*alignment=*/1)
-            || !sgx_is_completely_within_enclave(enclave_ptr, size)) {
+void sgx_copy_from_enclave_verified(void* uptr, const void* ptr, size_t size) {
+    assert(sgx_is_completely_within_enclave(ptr, size));
+    assert(sgx_is_valid_untrusted_ptr(uptr, size, /*alignment=*/1));
+
+    if (size == 0) {
+        return;
+    }
+
+    /*
+     * This should be simple `memcpy(uptr, ptr, size)`, but CVE-2022-21166 (INTEL-SA-00615).
+     * To mitigate this issue, all writes to untrusted memory from within the enclave must be done
+     * in 8-byte chunks aligned to 8-bytes boundary. Since x64 allocates memory in pages of
+     * (at least) 0x1000 in size, we can safely 8-align the pointer down and the size up.
+     */
+    size_t copy_len;
+    size_t prefix_misalignment = (uintptr_t)uptr & 7;
+    if (prefix_misalignment) {
+        /* Beginning of the range to copy is misaligned. */
+        char prefix_val[8] = { 0 };
+        copy_len = MIN(sizeof(prefix_val) - prefix_misalignment, size);
+
+        copy_u64s_from_untrusted(prefix_val, (char*)uptr - prefix_misalignment, /*count=*/1);
+        memcpy(prefix_val + prefix_misalignment, ptr, copy_len);
+        copy_u64s_to_untrusted((char*)uptr - prefix_misalignment, prefix_val, /*count=*/1);
+        uptr = (char*)uptr + copy_len;
+        ptr = (const char*)ptr + copy_len;
+        size -= copy_len;
+
+        if (size == 0) {
+            return;
+        }
+    }
+    assert((uintptr_t)uptr % 8 == 0);
+
+    size_t suffix_misalignment = size & 7;
+    copy_len = size - suffix_misalignment;
+    assert(copy_len % 8 == 0);
+    copy_u64s_to_untrusted(uptr, ptr, copy_len / 8);
+    uptr = (char*)uptr + copy_len;
+    ptr = (const char*)ptr + copy_len;
+    size -= copy_len;
+
+    assert(size == suffix_misalignment);
+    if (suffix_misalignment) {
+        /* End of the range to copy is misaligned. */
+        char suffix_val[8] = { 0 };
+
+        copy_u64s_from_untrusted(suffix_val, uptr, /*count=*/1);
+        memcpy(suffix_val, ptr, suffix_misalignment);
+        copy_u64s_to_untrusted(uptr, suffix_val, 1);
+    }
+}
+
+bool sgx_copy_from_enclave(void* uptr, const void* ptr, size_t size) {
+    if (!sgx_is_valid_untrusted_ptr(uptr, size, /*alignment=*/1)
+            || !sgx_is_completely_within_enclave(ptr, size)) {
         return false;
     }
-    memcpy(urts_ptr, enclave_ptr, size);
+
+    sgx_copy_from_enclave_verified(uptr, ptr, size);
     return true;
 }
 
--- a/pal/src/host/linux-sgx/enclave_ocalls.c
+++ b/pal/src/host/linux-sgx/enclave_ocalls.c
@@ -60,8 +60,8 @@ static long sgx_exitless_ocall(uint64_t code, void* ocall_args) {
         return -ENOMEM;
     }
 
-    WRITE_ONCE(req->ocall_index, code);
-    WRITE_ONCE(req->buffer, ocall_args);
+    COPY_VALUE_TO_UNTRUSTED(&req->ocall_index, code);
+    COPY_VALUE_TO_UNTRUSTED(&req->buffer, ocall_args);
     spinlock_init(&req->lock);
 
     /* grab the lock on this request (it is the responsibility of RPC thread to unlock it when
@@ -108,9 +108,9 @@ static long sgx_exitless_ocall(uint64_t code, void* ocall_args) {
                 return -ENOMEM;
             }
 
-            WRITE_ONCE(ocall_futex_args->futex, &req->lock.lock);
-            WRITE_ONCE(ocall_futex_args->op, FUTEX_WAIT_PRIVATE);
-            WRITE_ONCE(ocall_futex_args->timeout_us, (uint64_t)-1); /* never time out */
+            COPY_VALUE_TO_UNTRUSTED(&ocall_futex_args->futex, &req->lock.lock);
+            COPY_VALUE_TO_UNTRUSTED(&ocall_futex_args->op, FUTEX_WAIT_PRIVATE);
+            COPY_VALUE_TO_UNTRUSTED(&ocall_futex_args->timeout_us, (uint64_t)-1); /* never time out */
 
             do {
                 /* at this point lock = LOCKED_*; before waiting on futex, need to move lock to
@@ -121,7 +121,7 @@ static long sgx_exitless_ocall(uint64_t code, void* ocall_args) {
                     /* at this point, futex(wait) syscall expects lock to be in LOCKED_WITH_WAITERS
                      * set by enclave thread above; if RPC thread moved it back to UNLOCKED, futex()
                      * immediately returns */
-                    WRITE_ONCE(ocall_futex_args->val, SPINLOCK_LOCKED_WITH_WAITERS);
+                    COPY_VALUE_TO_UNTRUSTED(&ocall_futex_args->val, SPINLOCK_LOCKED_WITH_WAITERS);
                     int ret = sgx_ocall(OCALL_FUTEX, ocall_futex_args);
                     if (ret < 0 && ret != -EAGAIN) {
                         sgx_reset_ustack(old_ustack);
@@ -152,8 +152,8 @@ noreturn void ocall_exit(int exitcode, int is_exitgroup) {
          * things. */
         die_or_inf_loop();
     }
-    WRITE_ONCE(ocall_exit_args->exitcode, exitcode);
-    WRITE_ONCE(ocall_exit_args->is_exitgroup, is_exitgroup);
+    COPY_VALUE_TO_UNTRUSTED(&ocall_exit_args->exitcode, exitcode);
+    COPY_VALUE_TO_UNTRUSTED(&ocall_exit_args->is_exitgroup, is_exitgroup);
 
 #ifdef ASAN
     /* Unpoison the stacks allocated for this thread. They can be later used for a new thread. */
@@ -203,12 +203,12 @@ int ocall_mmap_untrusted(void** addrptr, size_t size, int prot, int flags, int f
         requested_addr = NULL; /* for sanity */
     }
 
-    WRITE_ONCE(ocall_mmap_args->addr, requested_addr);
-    WRITE_ONCE(ocall_mmap_args->size, size);
-    WRITE_ONCE(ocall_mmap_args->prot, prot);
-    WRITE_ONCE(ocall_mmap_args->flags, flags);
-    WRITE_ONCE(ocall_mmap_args->fd, fd);
-    WRITE_ONCE(ocall_mmap_args->offset, offset);
+    COPY_VALUE_TO_UNTRUSTED(&ocall_mmap_args->addr, requested_addr);
+    COPY_VALUE_TO_UNTRUSTED(&ocall_mmap_args->size, size);
+    COPY_VALUE_TO_UNTRUSTED(&ocall_mmap_args->prot, prot);
+    COPY_VALUE_TO_UNTRUSTED(&ocall_mmap_args->flags, flags);
+    COPY_VALUE_TO_UNTRUSTED(&ocall_mmap_args->fd, fd);
+    COPY_VALUE_TO_UNTRUSTED(&ocall_mmap_args->offset, offset);
 
     do {
         retval = sgx_exitless_ocall(OCALL_MMAP_UNTRUSTED, ocall_mmap_args);
@@ -258,8 +258,8 @@ int ocall_munmap_untrusted(const void* addr, size_t size) {
         return -EPERM;
     }
 
-    WRITE_ONCE(ocall_munmap_args->addr, addr);
-    WRITE_ONCE(ocall_munmap_args->size, size);
+    COPY_VALUE_TO_UNTRUSTED(&ocall_munmap_args->addr, addr);
+    COPY_VALUE_TO_UNTRUSTED(&ocall_munmap_args->size, size);
 
     do {
         retval = sgx_exitless_ocall(OCALL_MUNMAP_UNTRUSTED, ocall_munmap_args);
@@ -358,8 +358,8 @@ int ocall_cpuid(unsigned int leaf, unsigned int subleaf, unsigned int values[sta
         return -EPERM;
     }
 
-    WRITE_ONCE(ocall_cpuid_args->leaf, leaf);
-    WRITE_ONCE(ocall_cpuid_args->subleaf, subleaf);
+    COPY_VALUE_TO_UNTRUSTED(&ocall_cpuid_args->leaf, leaf);
+    COPY_VALUE_TO_UNTRUSTED(&ocall_cpuid_args->subleaf, subleaf);
 
     do {
         /* cpuid must be retrieved in the context of current logical core, cannot use exitless */
@@ -395,14 +395,14 @@ int ocall_open(const char* pathname, int flags, unsigned short mode) {
         return -EPERM;
     }
 
-    WRITE_ONCE(ocall_open_args->flags, flags);
-    WRITE_ONCE(ocall_open_args->mode, mode);
+    COPY_VALUE_TO_UNTRUSTED(&ocall_open_args->flags, flags);
+    COPY_VALUE_TO_UNTRUSTED(&ocall_open_args->mode, mode);
     void* untrusted_pathname = sgx_copy_to_ustack(pathname, path_size);
     if (!untrusted_pathname) {
         sgx_reset_ustack(old_ustack);
         return -EPERM;
     }
-    WRITE_ONCE(ocall_open_args->pathname, untrusted_pathname);
+    COPY_VALUE_TO_UNTRUSTED(&ocall_open_args->pathname, untrusted_pathname);
 
     do {
         retval = sgx_exitless_ocall(OCALL_OPEN, ocall_open_args);
@@ -432,7 +432,7 @@ int ocall_close(int fd) {
         return -EPERM;
     }
 
-    WRITE_ONCE(ocall_close_args->fd, fd);
+    COPY_VALUE_TO_UNTRUSTED(&ocall_close_args->fd, fd);
 
     /* We should never restart host-level `close` syscall on errors (including `EINTR`), but
      * `sgx_ocall_close` does not forward any errors (always returns `0`), so this can only be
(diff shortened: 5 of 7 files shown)


kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1237)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv)

___
*[`pal/src/host/linux-sgx/enclave_api.h` line 33 at r1](https://reviewable.io/reviews/gramineproject/gramine/1237#-NQiwmwP6DQfEXbFriW3:-NQj8aXMBoJuE7nMujwN:b-gke1hm) ([raw file](https://github.com/gramineproject/gramine/blob/a79a60895adacc9e56c1df081537a819b4e0b026/pal/src/host/linux-sgx/enclave_api.h#L33)):*

I'm not blocking, but this name is hard to understand (it should be read as "copy to untrusted memory the following value").

My suggestion would be `COPY_VALUE_TO_UNTRUSTED()`, but let's wait if anyone else has a strong opinion. If not, then I suggest to keep the current name.
</blockquote></details>

The proposed name looks better, will update.
___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 165 at r1](https://reviewable.io/reviews/gramineproject/gramine/1237#-NQizTDl70AK6lXN2iUi:-NQj8Hi2EbHdkKx3Awma:bnjh14n) ([raw file](https://github.com/gramineproject/gramine/blob/a79a60895adacc9e56c1df081537a819b4e0b026/pal/src/host/linux-sgx/enclave_framework.c#L165)):*

Unfortunately, this whole PR looks wrong.

Consider what happened in the previous code, where we protected reads of untrusted memory from within the enclave:
1. `src` was the untrusted memory
2. `dest` was the enclave memory
3. The previous code correctly did the copying in two steps:
   1. First copy an 8B chunk from aligned-down `src` (untrusted memory) into temp var `prefix_val` (in-enclave) -- for this we must use the safe `copy_u64s()` func
   2. Then copy enclave-to-enclave, from `prefix_val` to `dest` -- for this we can use the unsafe `memcpy()` func

Now consider what happens with your new code, and the case of protected writes from enclave to untrusted memory:
1. `src` is the enclave memory
2. `dest` is the untrusted memory
3. Your new code does the copying in two steps:
   1. First copy an 8B chunk from aligned-down `src` (in-enclave) into temp var `prefix_val` (in-enclave) -- you could use an unsafe `memcpy()` here but you're using `copy_u64s()`
    2. Then copy enclave-to-untrusted, from `prefix_val` to `dest` -- for this you **must use the safe `copy_u64s()` func, but you're **wrongly using unsafe `memcpy()`**

---

I think you have no way but to write a similar function as `sgx_copy_to_enclave_verified()`, but with the order of `memcpy()` and `copy_u64s()` flipped.

Also, I would some quick comments about the usage of these two functions -- so that people in the future won't do the same mistake as you did now.
</blockquote></details>

Ah... So stupid! I did not audit the original function carefully (I made blind assumptions that it works safely as a whole).

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1237)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv)

___
*[`pal/src/host/linux-sgx/enclave_api.h` line 35 at r1](https://reviewable.io/reviews/gramineproject/gramine/1237#-NQixJrH8EODGlezmDRQ:-NQk1xh33ygTMGxIVQMz:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/a79a60895adacc9e56c1df081537a819b4e0b026/pal/src/host/linux-sgx/enclave_api.h#L35)):*

There is no reason to take `src_val` in parentheses, so you can just write the second arg as `&src_val`.
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 125 at r1](https://reviewable.io/reviews/gramineproject/gramine/1237#-NQiy5z-9w3Z4xLhzaM7:-NQk1zsdA7DSxkI1-bF3:bwd8vv8) ([raw file](https://github.com/gramineproject/gramine/blob/a79a60895adacc9e56c1df081537a819b4e0b026/pal/src/host/linux-sgx/enclave_framework.c#L125)):*

Rewrap to 100 chars per line
</blockquote></details>

Not relevant any more.
___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 165 at r1](https://reviewable.io/reviews/gramineproject/gramine/1237#-NQizTDl70AK6lXN2iUi:-NQk21Kr-kuZXbKi3goX:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/a79a60895adacc9e56c1df081537a819b4e0b026/pal/src/host/linux-sgx/enclave_framework.c#L165)):*

Ah... So stupid! I did not audit the original function carefully (I made blind assumptions that it works safely as a whole).
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 189 at r1](https://reviewable.io/reviews/gramineproject/gramine/1237#-NQj-Zo-1NPHv81G4vHU:-NQk22LuDVD7WhFYnZNb:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/a79a60895adacc9e56c1df081537a819b4e0b026/pal/src/host/linux-sgx/enclave_framework.c#L189)):*

ditto, same **unsafe problem** as above
</blockquote></details>

Done.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1237)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1237#-NQyMN7t-5NBPoRq09Qf:-NQyMN7t-5NBPoRq09Qg:b-c1sv9c) (no related file):*
Looks good to me now!

___
___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 238 at r2](https://reviewable.io/reviews/gramineproject/gramine/1237#-NQyL_Py2GL35AlwFDu3:-NQyL_Py2GL35AlwFDu4:b-3iu2nb) ([raw file](https://github.com/gramineproject/gramine/blob/d092e08451e57c9c18eec0a962c9ac5ed699a0e3/pal/src/host/linux-sgx/enclave_framework.c#L238)):*
> ```C
>         copy_u64s_from_untrusted(prefix_val, (char*)uptr - prefix_misalignment, /*count=*/1);
>         memcpy(prefix_val + prefix_misalignment, ptr, copy_len);
>         copy_u64s_to_untrusted((char*)uptr - prefix_misalignment, prefix_val, /*count=*/1);
> ```

Ok, this part is a bit tricky: to not overwrite the first bytes in untrusted memory (which are not covered by the copied-from-enclave data), we must first copy these untrusted must-not-be-overwritten bytes together with the to-be-overwritten bytes inside of the temporary var `prefix_val`, then do a `memcpy()` on the latter bytes, and then write back from enclave to untrusted the resulting temporary var.

I first thought of adding a comment about this trick, but then I couldn't come up with a good simple wording. So I think we should keep the code as-is -- a thoughtful reader will understand what this code does.
___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 264 at r2](https://reviewable.io/reviews/gramineproject/gramine/1237#-NQyMEQk02c-fCnrx6bz:-NQyMEQk02c-fCnrx6c-:bmi4j9b) ([raw file](https://github.com/gramineproject/gramine/blob/d092e08451e57c9c18eec0a962c9ac5ed699a0e3/pal/src/host/linux-sgx/enclave_framework.c#L264)):*
> ```C
>         copy_u64s_from_untrusted(suffix_val, uptr, /*count=*/1);
>         memcpy(suffix_val, ptr, suffix_misalignment);
>         copy_u64s_to_untrusted(uptr, suffix_val, 1);
> ```

Here we have the same situation as I described above. Just a note for other reviewers, no need to modify anything.

mgild:
Thank you for the work on this issue. How soon will this reach production and are there similar mitigations in gramine for INTEL-SA-00334

mgild:
I am still looking into the security scope of all syscalls but could the runtime check `sched_getaffinity` be securely used to ensure hyperthreading is disabled for the runtime? Perhaps a manifest setting to check this with documentation pointing to the advisory would be helpful.

mkow (changes requested):
> Reviewed 2 of 6 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1237)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

___
*[`pal/src/host/linux-sgx/enclave_api.h` line 35 at r2](https://reviewable.io/reviews/gramineproject/gramine/1237#-NRYCCfX2ELaDV5wJRJa:-NRYCCfYDMNK97OMiahb:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/d092e08451e57c9c18eec0a962c9ac5ed699a0e3/pal/src/host/linux-sgx/enclave_api.h#L35)):*
> ```objc
> #define COPY_VALUE_TO_UNTRUSTED(untrusted_ptr, val) ({                            \
>     __typeof__(*(untrusted_ptr)) src_val = (val);                                 \
>     sgx_copy_from_enclave_verified((untrusted_ptr), &src_val, sizeof(src_val));   \
> ```





_Suggestion:_
```objc
sgx_copy_from_enclave_verified(untrusted_ptr, &src_val, sizeof(src_val));
```
___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 238 at r2](https://reviewable.io/reviews/gramineproject/gramine/1237#-NQyL_Py2GL35AlwFDu3:-NRYL5_b7LXx_2CMCsfw:b-rkbe2j) ([raw file](https://github.com/gramineproject/gramine/blob/d092e08451e57c9c18eec0a962c9ac5ed699a0e3/pal/src/host/linux-sgx/enclave_framework.c#L238)):*

Ok, this part is a bit tricky: to not overwrite the first bytes in untrusted memory (which are not covered by the copied-from-enclave data), we must first copy these untrusted must-not-be-overwritten bytes together with the to-be-overwritten bytes inside of the temporary var `prefix_val`, then do a `memcpy()` on the latter bytes, and then write back from enclave to untrusted the resulting temporary var.

I first thought of adding a comment about this trick, but then I couldn't come up with a good simple wording. So I think we should keep the code as-is -- a thoughtful reader will understand what this code does.
</blockquote></details>

Can we actually do that? This could cause a race-condition in the untrusted part: we first read some data placed before the buffer and then a moment later write them back. But this is some other memory and there could be concurrent writes there, which we just undid.
___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 264 at r2](https://reviewable.io/reviews/gramineproject/gramine/1237#-NRYM_Rq8CjArGplfg_D:-NRYM_Rq8CjArGplfg_E:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/d092e08451e57c9c18eec0a962c9ac5ed699a0e3/pal/src/host/linux-sgx/enclave_framework.c#L264)):*
> ```C
>         copy_u64s_from_untrusted(suffix_val, uptr, /*count=*/1);
>         memcpy(suffix_val, ptr, suffix_misalignment);
>         copy_u64s_to_untrusted(uptr, suffix_val, 1);
> ```

ditto

mkow (commented):
It can't, both `sched_getaffinity` and the CPU topology as seen from the inside of the enclave can be lied to by the host. But AFAIR HT state is reflected in the attestation, so you can check it there.

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1237#-:-NRYQO4oCV5rDpFVp4mv:b-5clpzc)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

dimakuv (dismissed):
> But AFAIR HT state is reflected in the attestation, so you can check it there.

Yes, exactly: https://github.com/intel/linux-sgx/issues/340

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1237#-:-NRYWEou8KBogZX5Eqi9:bpg7xxi)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 238 at r2](https://reviewable.io/reviews/gramineproject/gramine/1237#-NQyL_Py2GL35AlwFDu3:-NRYWQPR-QJTanNwUU5Q:bd3jrzy) ([raw file](https://github.com/gramineproject/gramine/blob/d092e08451e57c9c18eec0a962c9ac5ed699a0e3/pal/src/host/linux-sgx/enclave_framework.c#L238)):*

Can we actually do that? This could cause a race-condition in the untrusted part: we first read some data placed before the buffer and then a moment later write them back. But this is some other memory and there could be concurrent writes there, which we just undid.
</blockquote></details>

@mkow The only time this function is called is to copy from the enclave to the untrusted stack (afair, actually @kailun-qin could we check this assumption). So since the untrusted stack is per-thread (and signals use a dedicated alternate stack), we are free from data race in the benign case.

kailun-qin (commented):
> I think the official mitigation for LVI was µC update + disabling hyper-threading?

Not exactly. CVE-2020-0551 aka INTEL-SA-00334 aka LVI may require SW-level hardening.

In short, Intel/Microsoft/upstream changed tools for LVI: GCC, Microsoft and LLVM, which include adding an LFENCE after every load. For example in Linux, we'll need GCC 7.3 or later and the patched GNU assembler (the fix was upstreamed to GBU Binutils). For convenience, SGX SDK provided the updated assembler binary along with all the "matching" tools/utilities needed from Binutils (ld, objdump etc.): https://download.01.org/intel-sgx/latest/linux-latest/as.ld.objdump.r4.tar.gz.

Additionally, on top of the above tools, SGX SDK provides several mitigation levels ("no mitigation", "indirects-only mitigation", "full mitigation"). Pls kindly see https://github.com/intel/linux-sgx/blob/d3fd8b45117bb656546cd088acfa16d353b322b4/buildenv.mk#L202-L216 and https://github.com/intel/linux-sgx/blob/d3fd8b45117bb656546cd088acfa16d353b322b4/buildenv.mk#L226-L248 for details.

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1237#-:-NR_wwoeD-b_dk0WKeKJ:b-1d0wh3)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`pal/src/host/linux-sgx/enclave_api.h` line 35 at r2](https://reviewable.io/reviews/gramineproject/gramine/1237#-NRYCCfX2ELaDV5wJRJa:-NR_ykis0QRsO5yiheIO:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/d092e08451e57c9c18eec0a962c9ac5ed699a0e3/pal/src/host/linux-sgx/enclave_api.h#L35)):*
> ```objc
> #define COPY_VALUE_TO_UNTRUSTED(untrusted_ptr, val) ({                            \
>     __typeof__(*(untrusted_ptr)) src_val = (val);                                 \
>     sgx_copy_from_enclave_verified((untrusted_ptr), &src_val, sizeof(src_val));   \
> ```

Done.
___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 238 at r2](https://reviewable.io/reviews/gramineproject/gramine/1237#-NQyL_Py2GL35AlwFDu3:-NR_yqlq8Wl834WU4aER:b6v1tfw) ([raw file](https://github.com/gramineproject/gramine/blob/d092e08451e57c9c18eec0a962c9ac5ed699a0e3/pal/src/host/linux-sgx/enclave_framework.c#L238)):*
> The only time this function is called is to copy from the enclave to the untrusted stack (afair, actually @kailun-qin could we check this assumption)

Yes, most of the references of this function are copying to the untrusted stack for OCALL args (e.g. allocated via `sgx_alloc_on_ustack_aligned()`). The only concern might be here: https://github.com/gramineproject/gramine/blob/master/pal/src/host/linux-sgx/pal_events.c#L27-L59. But since it's guaranteed with atomicity by the lock. I think it should be fine.

Pls kindly help double check.

> Can we actually do that? This could cause a race-condition in the untrusted part: we first read some data placed before the buffer and then a moment later write them back. But this is some other memory and there could be concurrent writes there, which we just undid.

Yeah, I had exatly the same concerns while making these changes. But I made the same assumption of the per-thread ustack as @dimakuv mentioned.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1237)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin and @mkow)

___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 238 at r2](https://reviewable.io/reviews/gramineproject/gramine/1237#-NQyL_Py2GL35AlwFDu3:-NRb7ZrAC4wFbyKAbm8x:b-p9sc4u) ([raw file](https://github.com/gramineproject/gramine/blob/d092e08451e57c9c18eec0a962c9ac5ed699a0e3/pal/src/host/linux-sgx/enclave_framework.c#L238)):*

> The only time this function is called is to copy from the enclave to the untrusted stack (afair, actually @kailun-qin could we check this assumption)

Yes, most of the references of this function are copying to the untrusted stack for OCALL args (e.g. allocated via `sgx_alloc_on_ustack_aligned()`). The only concern might be here: https://github.com/gramineproject/gramine/blob/master/pal/src/host/linux-sgx/pal_events.c#L27-L59. But since it's guaranteed with atomicity by the lock. I think it should be fine.

Pls kindly help double check.

> Can we actually do that? This could cause a race-condition in the untrusted part: we first read some data placed before the buffer and then a moment later write them back. But this is some other memory and there could be concurrent writes there, which we just undid.

Yeah, I had exatly the same concerns while making these changes. But I made the same assumption of the per-thread ustack as @dimakuv mentioned.


</blockquote></details>

Agreed with @kailun-qin that these two locations should be safe:
1. Copying to the untrusted stack (as the stack is per-thread, so no data races with other threads/signals)
2. `alloc_untrusted_futex_word()` and `free_untrusted_futex_word()` -- these are protected by a global lock `g_untrusted_page_lock`, so no data races possible

But I found one other vulnerability -- in the `pal_events.c` logic, with futex ints. See my other comment.
___
*[`pal/src/host/linux-sgx/pal_events.c` line 118 at r3](https://reviewable.io/reviews/gramineproject/gramine/1237#-NRb5fZW5mQ6bz9GKe5y:-NRb5fZW5mQ6bz9GKe5z:b-sswtxq) ([raw file](https://github.com/gramineproject/gramine/blob/bda2dc8ef28dcbc3334380dbf5aeda753e462597/pal/src/host/linux-sgx/pal_events.c#L118)):*
> ```C
>     handle->event.signaled = init_signaled;
>     handle->event.auto_clear = auto_clear;
>     __atomic_store_n(handle->event.signaled_untrusted, init_signaled ? 1 : 0, __ATOMIC_RELEASE);
> ```

Oops, looks like we still have a vulnerability in `signaled_untrusted` atomic accesses, like this one.

This is because `signaled_untrusted` is a 4-byte int, and so we perform 4-byte stores. I think we can more or less easily fix it as follows:
1. Change `signaled_untrusted` type to an 8-byte int
2. Change `alloc_untrusted_futex_word()` and `free_untrusted_futex_word()` to allocate 8-byte integers -- note that this will actually simplify the ASan logic (and will be uniform with ASan assumptions)
3. Now the ugly part is that we'll need to change `ocall_futex()` to accept an 8-byte integer, and in the untrusted part, we'll cast this 8-byte integer to a 4-byte int (to be compatible with `futex()` syscall signature).
   - Alternatively, we could keep `ocall_futex()` signature as is and just case the 8-byte integer to a 4-byte int here -- I actually prefer this approach, to touch less parts in the code.
4. We'll need to add VERY VISIBLE comments on why we use 8-byte ints instead of a classic 4-byte int for futexes.

Note that 8-byte ints returned by `alloc_untrusted_futex_word()` are guaranteed to be 8-byte aligned, so we don't need to care about alignment.

@kailun-qin Does that sound reasonable to you? Could you apply such a fix here?

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1237)* status: 5 of 7 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`pal/src/host/linux-sgx/pal_events.c` line 118 at r3](https://reviewable.io/reviews/gramineproject/gramine/1237#-NRb5fZW5mQ6bz9GKe5y:-NRbzU7R9mGUnWPlzMCK:b-juelf2) ([raw file](https://github.com/gramineproject/gramine/blob/bda2dc8ef28dcbc3334380dbf5aeda753e462597/pal/src/host/linux-sgx/pal_events.c#L118)):*

Oops, looks like we still have a vulnerability in `signaled_untrusted` atomic accesses, like this one.

This is because `signaled_untrusted` is a 4-byte int, and so we perform 4-byte stores. I think we can more or less easily fix it as follows:
1. Change `signaled_untrusted` type to an 8-byte int
2. Change `alloc_untrusted_futex_word()` and `free_untrusted_futex_word()` to allocate 8-byte integers -- note that this will actually simplify the ASan logic (and will be uniform with ASan assumptions)
3. Now the ugly part is that we'll need to change `ocall_futex()` to accept an 8-byte integer, and in the untrusted part, we'll cast this 8-byte integer to a 4-byte int (to be compatible with `futex()` syscall signature).
   - Alternatively, we could keep `ocall_futex()` signature as is and just case the 8-byte integer to a 4-byte int here -- I actually prefer this approach, to touch less parts in the code.
4. We'll need to add VERY VISIBLE comments on why we use 8-byte ints instead of a classic 4-byte int for futexes.

Note that 8-byte ints returned by `alloc_untrusted_futex_word()` are guaranteed to be 8-byte aligned, so we don't need to care about alignment.

@kailun-qin Does that sound reasonable to you? Could you apply such a fix here?
</blockquote></details>

Good catch! Yes, make sense to me, updated!

mkow (commented):
> Not exactly. CVE-2020-0551 aka INTEL-SA-00334 aka LVI may require SW-level hardening.

> include adding an LFENCE after every load

Ok, now I remember that mitigation. As explained on the call, I think the history behind our LVI mitigation state is that when LVI went public Gramine still wasn't used in production anywhere and we weren't sure when that would happen, so we didn't implement such mitigations (which are inherently temporal, because new CPUs won't have such bugs).
This strategy seems to have worked, because server cores >= ICX were fixed and aren't susceptible to LVI (@kailun-qin: could you post a link here to that table with vulnerable CPUs you posted on the call?).

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1237#-:-NRcckTs3NxMJPQTS0A5:bbbgasq)* status: 5 of 7 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @kailun-qin)

mkow (changes requested):
> Reviewed 1 of 1 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1237)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @kailun-qin)

___
*[`pal/src/host/linux-sgx/enclave_framework.c` line 238 at r2](https://reviewable.io/reviews/gramineproject/gramine/1237#-NQyL_Py2GL35AlwFDu3:-NRdpQAUEiN_w9HGVOQz:b-auowcv) ([raw file](https://github.com/gramineproject/gramine/blob/d092e08451e57c9c18eec0a962c9ac5ed699a0e3/pal/src/host/linux-sgx/enclave_framework.c#L238)):*

Agreed with @kailun-qin that these two locations should be safe:
1. Copying to the untrusted stack (as the stack is per-thread, so no data races with other threads/signals)
2. `alloc_untrusted_futex_word()` and `free_untrusted_futex_word()` -- these are protected by a global lock `g_untrusted_page_lock`, so no data races possible

But I found one other vulnerability -- in the `pal_events.c` logic, with futex ints. See my other comment.
</blockquote></details>

Heh, I just wanted to point exactly this thing out, but then read @dimakuv's message :) These CPU bugs are really annoying to workaround...
___
*[`pal/src/host/linux-sgx/pal_events.c` line 50 at r2](https://reviewable.io/reviews/gramineproject/gramine/1237#-NRdpsPp-6XaQgMtp-hv:-NRdpsPqDFOtmIzrQ313:b-y3va5k) ([raw file](https://github.com/gramineproject/gramine/blob/d092e08451e57c9c18eec0a962c9ac5ed699a0e3/pal/src/host/linux-sgx/pal_events.c#L50)):*
> ```C
> #ifdef ASAN
>     /* ASAN requires 8-byte aligned addresses, so we need to pad. */
>     g_untrusted_page_next_entry += sizeof(uint32_t);
> ```

Hmm, maybe I'd leave this comment and code, just comment it out or something? Some day in the future we'll be removing this painful mitigation (after all then-supported CPUs will be shipped without this bug) and I can easily imagine breaking ASan here accidentally when doing so :)

kailun-qin (commented):
> This strategy seems to have worked, because server cores >= ICX were fixed and aren't susceptible to LVI (@kailun-qin: could you post a link here to that table with vulnerable CPUs you posted on the call?).

Sure, pls kindly find below the list of affected processors for LVI (2018-2021 tab, Load Value Injection column):
https://www.intel.cn/content/www/cn/zh/developer/topic-technology/software-security-guidance/processors-affected-consolidated-product-cpu-model.html#tab-blade-1-1.

Note that this is an **updated** version of the original list of LVI affected processors (https://www.intel.com/content/www/us/en/developer/articles/technical/software-security-guidance/resources/processors-affected-load-value-injection.html) that I posted on the call.

In this latest list, we can clearly find that all series of Ice Lake (i.e., not only Ice Lake client cores in the original list as questioned on the call) and Sapphire Rapids microarchitectures are stated to be **not affected** by CVE-2020-0551/INTEL-SA-00334/LVI.

@mkow @dimakuv We can also follow this latest list to check whether Gramine is susceptible to other vulnerabilites that we may've missed for recent cores (>= ICX).

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1237#-:-NRf1YfwBWxyWO4txxPp:b-jpkphr)* status: 6 of 7 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`pal/src/host/linux-sgx/pal_events.c` line 50 at r2](https://reviewable.io/reviews/gramineproject/gramine/1237#-NRdpsPp-6XaQgMtp-hv:-NRf1WGX-aKcxuzuiLAY:ba2wsx5) ([raw file](https://github.com/gramineproject/gramine/blob/d092e08451e57c9c18eec0a962c9ac5ed699a0e3/pal/src/host/linux-sgx/pal_events.c#L50)):*

Hmm, maybe I'd leave this comment and code, just comment it out or something? Some day in the future we'll be removing this painful mitigation (after all then-supported CPUs will be shipped without this bug) and I can easily imagine breaking ASan here accidentally when doing so :)
</blockquote></details>

Added back and commented out, done.

dimakuv (changes requested):
> Reviewed 1 of 2 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1237)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin and @mkow)

___
*[`pal/src/host/linux-sgx/pal_events.c` line 50 at r2](https://reviewable.io/reviews/gramineproject/gramine/1237#-NRdpsPp-6XaQgMtp-hv:-NRgMUqW8FHFdIAYbjym:b-1gqge5) ([raw file](https://github.com/gramineproject/gramine/blob/d092e08451e57c9c18eec0a962c9ac5ed699a0e3/pal/src/host/linux-sgx/pal_events.c#L50)):*

Added back and commented out, done.
</blockquote></details>

I love the `painful mitigation` in the comment :)
___
*[`pal/src/host/linux-sgx/pal_host.h` line 139 at r5](https://reviewable.io/reviews/gramineproject/gramine/1237#-NRgNjosCQUePBdyhLBH:-NRgNjosCQUePBdyhLBI:bf8x6oj) ([raw file](https://github.com/gramineproject/gramine/blob/d95795372a8ab2a340299e6b87e8d65af7430369/pal/src/host/linux-sgx/pal_host.h#L139)):*
> ```objc
>              * this futex word. This is to mitigate CVE-2022-21166 (INTEL-SA-00615) which requires
>              * all writes to untrusted memory from within the enclave to be done in 8-byte chunks
>              * aligned to 8-bytes boundary.*/
> ```

Please also add a sentence at the end is that "we can safely typecast this 8-byte int to a 4-byte futex word because Intel SGX implies a little-endian CPU".

Basically, we're using this property of little-endian machines: https://en.wikipedia.org/wiki/Endianness#Simplified_access_to_part_of_a_field. And we need to comment about this trick.

Also, add a space before the closing `*/`

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1237)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`pal/src/host/linux-sgx/pal_host.h` line 139 at r5](https://reviewable.io/reviews/gramineproject/gramine/1237#-NRgNjosCQUePBdyhLBH:-NRgQI-R0SfWD043dvv9:b-7cqyw6) ([raw file](https://github.com/gramineproject/gramine/blob/d95795372a8ab2a340299e6b87e8d65af7430369/pal/src/host/linux-sgx/pal_host.h#L139)):*

Please also add a sentence at the end is that "we can safely typecast this 8-byte int to a 4-byte futex word because Intel SGX implies a little-endian CPU".

Basically, we're using this property of little-endian machines: https://en.wikipedia.org/wiki/Endianness#Simplified_access_to_part_of_a_field. And we need to comment about this trick.

Also, add a space before the closing `*/`
</blockquote></details>

Thanks for the reference! Done.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1237)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1237#-NRgUEsz-YPyOU0jOQtA:-NRgUEsz-YPyOU0jOQtB:b-t08i66) (no related file):*
The change now looks good to me.

___

mkow (commented):
> We can also follow this latest list to check whether Gramine is susceptible to other vulnerabilites that we may've missed for recent cores (>= ICX).

@dimakuv: You're our attestation expert, could you check this? :) I remember that ITL's ICX had some troubles with it.

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1237#-:-NRhRPZXDR2F48wb_lBN:bk10oaj)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

dimakuv (approved):
@mkow Not sure I understood. What exactly do you mean by "check this"?

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1237#-:-NRhTppr7gg8ZL3s0AQW:b-x00moi)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1237)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

___
*[`pal/src/host/linux-sgx/pal_events.c` line 55 at r6](https://reviewable.io/reviews/gramineproject/gramine/1237#-NRhQNrp5LOpxY-IBmAO:-NRhe88YEzY-ZlHvulEM:bb6un1j) ([raw file](https://github.com/gramineproject/gramine/blob/84b7c29bac1e6d072f6c3ed50e4013bde07e89ba/pal/src/host/linux-sgx/pal_events.c#L55)):*

we can fix this during the final rebase
</blockquote></details>

Done

mkow (approved):
Run an attestation on ICX and SPR and check what the attestation server says about it (i.e. whether it said "software mitigations required for <lits of advisories>".

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1237#-:-NRheiIW25SSBKU0D95Q:b-kuqc8b)* status: :shipit: complete! all files reviewed, all discussions resolved

dimakuv:
Jenkins, retest Jenkins-SGX-18.04 please (Jenkins job actually finished successfully, but this was result wasn't propagated to GitHub for some reason, so let's try to kick the flows again)

dimakuv:
Jenkins, retest Jenkins-SGX-Sanitizers please (same as above)

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1237)* status: :shipit: complete! all files reviewed, all discussions resolved

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1237#-NRhPX8j2TFnCLTdg23K:-NRk6uInDtwU_TToO9zH:bneogdu) (no related file):*
Thanks for the performance results!

> I think the main reason is using 8-byte-copy instead of rep movsb or AVX.

Maybe we can check this as a follow-up.

___

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1237)* status: :shipit: complete! all files reviewed, all discussions resolved

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1237#-NRhPX8j2TFnCLTdg23K:-NRnKA-CEJtmSJ2uViqM:bq35wop) (no related file):*

Thanks for the performance results!

> I think the main reason is using 8-byte-copy instead of rep movsb or AVX.

Maybe we can check this as a follow-up.

</blockquote></details>

The problem is that AFAIR the advisory didn't say if `rep movsb` is ok or not assuming the addresses are aligned (despite it being the recommended implementation for memcpy) and I'm also not sure about 16 and 32 byte long accesses - I think they are fine, but implementing them would require a non-trivial effort. I.e. it's not that trivial to even check, we'd need to actually implement `memcpy()` using 8-byte and 16/32-byte accesses in assembly.

___

mgild:
how soon might we see an official release tag including this PR?

mkow:
We plan to do the next release at the end of May, see https://github.com/gramineproject/gramine/discussions/1278. If you think we should do it earlier then please argue there.

mgild:
@mkow Was checking in here on the 1.5 release. The benefit of an announced stable release is user confidence in having an azure supported machine with all current advisories mitigated (DCSv3 machines specifically) while still using Gramine.

mkow:
What do you mean exactly by "announced stable release"? Overall, I'm not sure what you wanted to say :)

btw.: Generally, I'd recommend using Gramine from +/- `master`, it's rather stable (in terms of regressions, we test each PR) and you'll get all the bugfixes sooner.

kailun-qin:
> Was checking in here on the 1.5 release

@mgild Yes, this fix will be included in the v1.5 release (which will happen soon).

> The benefit of an announced stable release is user confidence in having an azure supported machine with all current advisories mitigated (DCSv3 machines specifically) while still using Gramine.

I think you wanted to state that having a machine with all security advisories applied plus a Gramine release including the security fixes would increase users' confidence from a security perspective? Anyway, as stated, v1.5 will have it.

mgild:
Great to hear, thanks for the great product

Issue #804, opened 2022-07-29
More flexible she bang in python scripts


rafariossaa:
## Description of the problem

The she bang in the python scripts is pointing to the python installation at system level and makes it difficult to try to run the scripts in different python environments, or if you have a version at system level but wants to run it with other python in your system, other case could be with venvs.

A solution for this could be the following.
```diff
> #!/usr/bin/python3
---
< #!/usr/bin/env python3
```
This will allow the scripts to use the python in the `PATH`

## Expected results

To be able to run the python scripts with other pythons

boryspoplawski:
https://github.com/gramineproject/gramine/pull/429

dimakuv:
I'd say the most relevant and concise explanation is here: https://gramine.readthedocs.io/en/latest/devel/coding-style.html#python

> The she bang in the python scripts is pointing to the python installation at system level and makes it difficult to try to run the scripts in different python environments, or if you have a version at system level but wants to run it with other python in your system, other case could be with venvs.

I believe there is a real problem here (with "virtual environments" of Python, e.g., `venv` and `conda`), but @rafariossaa could you explain a particular scenario where you hit this issue? Gramine probably *should* have some modifications to work with virtual environments, we just didn't have a concrete example for this.

rafariossaa:
Hi,
  Sorry for the delay.
  In this case, I have a python installation at system level but for the user that runs gramine I have a newer version of python. I am not using venv.

dimakuv:
> In this case, I have a python installation at system level but for the user that runs gramine I have a newer version of python.

Why do you need a newer version of Python for the user? What I'm looking for is a specific scenario where this is necessary, and our current reliance on `/usr/bin/python3` breaks Gramine workflows.

Issue #431, opened 2022-02-25
Packaging issues to be fixed in 1.2


woju:
- [ ] `-Ddcap=enabled`: requires Intel's repo with additional dependencies
- [ ] sed shebang to `#!/usr/bin/python3`, because using `env` is broken for custom (non-distro) python installs (see #429)
- [ ] consider separate packaging libratls verify libraries, maybe for other architectures (thread about arm64: https://groups.google.com/d/msgid/gramine-users/9f0bbc15-6e10-44e5-aada-6ea47c3423d9n%40googlegroups.com)
- [ ] missing dependency against `libprotobuf-c1`

dimakuv:
@woju What do you mean by `sed shebang`?

woju:
> @woju What do you mean by `sed shebang`?

When change the shebangs, we'll need some mechanism of controlling them when
packaging for different distros. Most likely in packaging (not in our meson),
because those are different on different distros, some packaging systems
already do it by default:
https://src.fedoraproject.org/rpms/redhat-rpm-config/blob/rawhide/f/brp-mangle-shebangs.
When packaging to different distros, I shall check if other also need and/or
support it or do we have to do it manually (hence sed).

dimakuv:
@woju I see your point. Is it something that we want to do in the future, or do you want to see this change now?

If now, then what exactly should I do in https://github.com/gramineproject/gramine/pull/429?

woju:
> @woju I see your point. Is it something that we want to do in the future, or
> do you want to see this change now?

This isn't either "now" or "in the future", this is something that I need to
adjust whenever I package for another distro. We need to do it now (gsc+python
image is also kinda distro, because it's different python), and will need do
again when new distro with new convention arrives (e.g. BSD).

Basically, we need a knob, whether it's meson option or just a sed
postprocessing in packaging scripts (most likely the latter). 

> If now, then what exactly should I do in
> https://github.com/gramineproject/gramine/pull/429?

We should do whatever is convenient for gsc. As long as GSC compiles from
source, I think it should be `/usr/bin/python3` (as the PR does). If we
switched GSC to `apt install gramine`, then we would be in position to keep
the "most correct" shebang of `/usr/bin/env python3` in the source and
postprocess it in debian/rules. But this is a big change, we can always do
that later, and for now I think it's OK to do `/usr/bin/python3`.

mkow:
[...]

PR #429, opened 2022-02-24
[Python] Change Python scripts' shebang to `/usr/bin/python3`

dimakuv:
## Description of the changes 

Gramine Python utilities should use the system-wide Python interpreter instead of the local version of Python (which may not have system-wide-installed Python packages). This was detected in GSC when running the default Python Ubuntu Docker image.

Corresponding GSC PR: https://github.com/gramineproject/gsc/pull/49

## How to test this PR? 

Test GSC with our Python example: https://gramine.readthedocs.io/projects/gsc/en/latest/#example

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/429)

PR #1250, opened 2023-03-24
Use python according to env

sbellem:
## Description of the changes 
This pull request changes the shebang from `#!/usr/bin/python3` to `#!/usr/bin/env python3` in order to pick up the python executable that is configured on the host system where the python scripts are invoked.

### Motivation
For python scripts (`gramine-gen-depend`, `gramine-manifest`, etc) the `python` executable under `/usr/bin/` is picked up even though it may not be the "active" one as per the `PATH` or some other system configuration.

The motivation for this change came through attempting to build `gramine` from source, in a docker image based on [`python:3.10-bullseye`](https://github.com/docker-library/python/blob/8a8d6baac38dcd208f699ae2eb10f0893a764035/3.10/bullseye/Dockerfile) (debian) and running into issues with `graminelibos` or some dependencies such as `click` not being found. 

## How to test this PR? 
One may run one of the python scripts, such as `gramine-sgx-gen-private-key`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1250)

### DIFF
--- a/python/gramine-gen-depend
+++ b/python/gramine-gen-depend
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
 # SPDX-License-Identifier: LGPL-3.0-or-later
 # Copyright (C) 2021 Intel Corporation
 #                    Borys Popławski <borysp@invisiblethingslab.com>
--- a/python/gramine-manifest
+++ b/python/gramine-manifest
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
 # SPDX-License-Identifier: LGPL-3.0-or-later
 # Copyright (C) 2021 Intel Corporation
 #                    Borys Popławski <borysp@invisiblethingslab.com>
--- a/python/gramine-sgx-gen-private-key
+++ b/python/gramine-sgx-gen-private-key
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
 # SPDX-License-Identifier: LGPL-3.0-or-later
 # Copyright (c) 2021 Intel Corporation
 #                    Wojtek Porczyk <woju@invisiblethingslab.com>
--- a/python/gramine-sgx-get-token
+++ b/python/gramine-sgx-get-token
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
 # SPDX-License-Identifier: LGPL-3.0-or-later
 # Copyright (C) 2021 Intel Corporation
 #                    Borys Popławski <borysp@invisiblethingslab.com>
--- a/python/gramine-sgx-sign
+++ b/python/gramine-sgx-sign
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
 # SPDX-License-Identifier: LGPL-3.0-or-later
 # Copyright (C) 2021 Intel Corporation
 #                    Borys Popławski <borysp@invisiblethingslab.com>
--- a/python/gramine-sgx-sigstruct-view
+++ b/python/gramine-sgx-sigstruct-view
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
 # SPDX-License-Identifier: LGPL-3.0-or-later
 # Copyright (C) 2023 Intel Corporation
 #                    Dmitrii Kuvaiskii <dmitrii.kuvaiskii@intel.com>
--- a/python/gramine-test
+++ b/python/gramine-test
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
 # SPDX-License-Identifier: LGPL-3.0-or-later
 # Copyright (C) 2021 Intel Corporation
 #                    Paweł Marczewski <pawel@invisiblethingslab.com>


mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1250)* status: 0 of 7 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @sbellem)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1250#-NRVhHM8C8QdSO553aap:-NRVhHM8C8QdSO553aaq:b-j5wr3n) (no related file):*
Please check:
- Discussion under https://github.com/gramineproject/gramine/issues/431
- https://gramine.readthedocs.io/en/latest/devel/coding-style.html#python
- https://github.com/gramineproject/gramine/pull/429
- https://github.com/gramineproject/gramine/issues/804

I'm not a fan of this solution, it's a hack to make some Docker images work under GSC which made core Gramine worse. But others preferred to just change the shebang everywhere and we went this way.
Maybe we will change this when moving GSC to Gramine-from-packages, if that happens.

___

sbellem:
Ok, I see ... How do you use a python executable different than the system one?

dimakuv:
> Ok, I see ... How do you use a python executable different than the system one?

@woju Do we have a good answer to this question?

woju:
For what do you want to use different python that system one? If you just want to run `gramine-manifest` and `gramine-sgx-sign`, then python interpreter is just implementation detail and for all intents and purposes those tools could be written in any other language and it shouldn't change much. You shouldn't mess with this because of `_offsets` module, which is tightly coupled to the particular gramine build (theoretically it can differ even from build to build of the same source). For this reason, this PR is NAK.

If you want to run different python version inside gramine, then you just write a manifest, specify a proper entrypoint and make sure the relevant platlib and purelib directories are measured as `sgx.trusted_files` (like any other application). There are some helpers in `gramine-manifest` (https://gramine.readthedocs.io/en/stable/manpages/gramine-manifest.html) to get those paths, but they're for system's python, so probably it's hard to use them, if at all. I could use some suggestion how to fix this API to make it usable for other interpreters.

Furthermore, if you have `bullseye`-based images, you can just install gramine from apt repositories. The package will pull proper dependencies (incl. `python3-click` for `click`, which you've mentioned) and will just work (unless you screw `PATH` and/or `PYTHONPATH` in some major way).

woju:
> He said he's trying to manually build Gramine. See the PR description.

Yeah, but for what exactly? What is to be gained by building gramine inside venv?

Our building instructions (the part where you install dependencies) are not written for exotic environments like python's venvs. They're general guidelines, but if you have some not supported build environment, then you're on your own to get all the dependencies.

sbellem:
Hey, thanks for taking the time to review the PR. I am not deep enough into gramine to argue for or against anything here.

As mentioned in the description of the PR, I was building gramine from source in a python-(debian) based image and I ran into the issue of `/usr/bin/python3` (lower precedence on `$PATH`) being picked up instead of `/usr/local/bin/python3` (higher precedence on the `$PATH`).

As for python dependencies, I feel that is a different issue, but as far as my little experience goes, not all of them are available via the distro, and sometimes must be obtained via PyPI, and this can also be a source of troubles if one is constrained with the system python of the distro. But again, I think this is a separate concern from controlling which python version is used to build gramine.

woju:
> not all of them are available via the distro, and sometimes must be obtained via PyPI, and this can also be a source of troubles if one is constrained with the system python of the distro.

We've been careful to use only those available from distro. There's not much of them, distro coverage in this area ranges from very good (click and jinja) to poor but workable (tomli), and we got away with this relatively easy. WRT to source of problems, it's the PyPI that's the source of problems, or more precisely, the people that upload there. Distro ensures that the package set that is available from distro works together fine. In this regard Debian is not that different from Conda, it just caters to slightly different population. We don't need newest features, for us it's enough that CLI arguments parsing works solid.

> But again, I think this is a separate concern from controlling which python version is used to build gramine.

Yes and no. The problem is exactly that it's not that easy to *separate* those concerns. It's very hard to use different python versions/interpreters/envs in a single project, because you need to keeping track of what version and package goes where.

If you really want to use some alternative venv, I suggest you apply this patch yourself, then go to our building manual (https://gramine.readthedocs.io/en/stable/devel/building.html) and whenever it says to install `python3-something`, you install that something with pip. This might or might not work, maybe you'll get too new packages and something will break. Or not. You're on your own. :)

I'm not willing to rewrite the build instructions nor shebangs, simply because if you stick to system's python (shebang and `apt-get install python3-*`), it 1) just works for most people, 2) it's easy to package. Otherwise you get into that area where you mix python versions and packages, which very fast gets very hard to untangle and would be a significant support burden on us.

Last but not least, I still didn't got the answer: why would you build gramine from source in an exotic env? Are you trying to run inside gramine a python app that requires newer python? For that you can use packaged version just fine. You don't need the same interpreter to run `gramine-sgx` and your app.

If you have a good answer why this is something that would be valuable, we could think about maybe some meson option for changing shebangs to `#!/usr/bin/env python3`, and fixing them to system's python in packaging (because it's packaging where this is really important). But I didn't get the motivation behind this effort.

dimakuv:
Based on @woju's feedback (and no activity since), I would like to close this PR.

Issue #1255, opened 2023-03-28
Update mbedtls to >=3.4.0 before 1.5 release


woju:
mbedtls 3.4.0 was released earlier today: https://github.com/Mbed-TLS/mbedtls/releases/tag/v3.4.0

It has fixed a vulnerability which we have mitigated with the patch: https://github.com/gramineproject/gramine/blob/43101b8fa3fd4c1cd6c9304c2094ce009627efb3/subprojects/packagefiles/mbedtls/enforce-aes-ni.patch

woju:
Cc: @BFuhry

PR #796, opened 2022-07-26
[mbedtls,tools/sgx,CI-Examples,common] Upgrade mbedTLS to version 3.2.1

kailun-qin:
Mbed TLS should be (periodically) updated to include security (also bug) fixes.

The initial PR and discussions are from https://github.com/gramineproject/gramine/pull/426 and another patch https://github.com/gramineproject/gramine/pull/730 proposes to upgrade it to 2.28.0 (LTS version) instead of 3.1.0 for smaller concern/controversy.

As the blocker https://github.com/gramineproject/gramine/pull/740 was resolved recently, we should be able to proceed w/ the latest mbedTLS version - which is v3.2.1.

This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/796)

PR #1256, opened 2023-03-28
[mbedtls] Upgrade mbedTLS to version 3.4.0

kailun-qin:
This commit introduces the following changes along with the upgrade of mbedTLS to version 3.4.0:
* updating the sonames of mbedTLS-produced libs;
* aligning all Gramine patches to conform with the new coding style of mbedTLS;
* explicitly specifying the Gramine-provided secure zeroization function `erase_memory()` and updating it to a more performant version to cope with the updated `mbedtls_platform_zeroize()` detection logic;
* dropping `MBEDTLS_SHA224_C` in the mbedTLS config which is now made independent from `MBEDTLS_SHA256_C` (it is actually not used in Gramine but was required to be present together with `MBEDTLS_SHA256_C` in the previous mbedTLS versions).

Fixes: https://github.com/gramineproject/gramine/issues/1255
Fixes: https://github.com/gramineproject/gramine/issues/1257

This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1256)


kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1256)* status: 0 of 10 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: )

___
*[`subprojects/mbedtls-mbedtls-3.4.0.wrap` line 4 at r1](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRd1HP11dFDdUs8o-on:-NRd1HP11dFDdUs8o-oo:blcz1y8) ([raw file](https://github.com/gramineproject/gramine/blob/bf0159cdd4c57d25c72894da2283321bbc1fdb47/subprojects/mbedtls-mbedtls-3.4.0.wrap#L4)):*
> ```
> directory = mbedtls-mbedtls-3.4.0
> source_url = https://github.com/ARMmbed/mbedtls/archive/mbedtls-3.4.0.tar.gz
> source_fallback_url = https://packages.gramineproject.io/distfiles/mbedtls-3.4.0.tar.gz
> ```

@woju: Please kindly help mirror this file. Thanks!

I'll test against it afterwards.

woju (changes requested):
> Reviewed 8 of 10 files at r1, 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1256)* status: 11 of 12 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin)

___
*[`-- commits` line 2 at r2](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRdDD-G7M9ezKA0xgfX):*
Commit message lacks description of the changes that were done together with this update. I've noticed at least:
- that the AES-NI patch was dropped wrt. upstream fix (and was the upstream fix reviewed to be sufficient for us?)
- sonames were changed
- zeroize
- sha224

Also please add "Fixes: #1255" either to PR description or here to the commit.
___
*[`subprojects/mbedtls-mbedtls-3.4.0.wrap` line 4 at r1](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRd1HP11dFDdUs8o-on:-NRdCY4_1b_vyiL72cfD:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/bf0159cdd4c57d25c72894da2283321bbc1fdb47/subprojects/mbedtls-mbedtls-3.4.0.wrap#L4)):*

@woju: Please kindly help mirror this file. Thanks!

I'll test against it afterwards.
</blockquote></details>

Done.
___
*[`subprojects/packagefiles/mbedtls/include/mbedtls/config-pal.h` line 38 at r2](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRdEoid7xYYTb32BxEk:-NRdEoid7xYYTb32BxEl:b-4scqbw) ([raw file](https://github.com/gramineproject/gramine/blob/43101b8fa3fd4c1cd6c9304c2094ce009627efb3/subprojects/packagefiles/mbedtls/include/mbedtls/config-pal.h#L38)):*
> ```objc
> #define MBEDTLS_PLATFORM_C
> #define MBEDTLS_RSA_C
> #define MBEDTLS_SHA224_C
> ```

Why this was removed, and if there's a reason, why isn't it written either it commit message, or at least in the PR?

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1256)* status: 8 of 12 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @woju)

___
*[`-- commits` line 2 at r2](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRdDD-G7M9ezKA0xgfX):*
> From https://github.com/gramineproject/gramine/issues/1255:
> mbedtls 3.4.0 was released earlier today: https://github.com/Mbed-TLS/mbedtls/releases/tag/v3.4.0
> It has fixed a vulnerability which we have mitigated with the patch: https://github.com/gramineproject/gramine/blob/43101b8fa3fd4c1cd6c9304c2094ce009627efb3/subprojects/packagefiles/mbedtls/enforce-aes-ni.patch

I saw your above comment when I submitted the first version of PR. However, after taking a deeper look at the changes in mbedtls, I don't think it's sufficient for us as it still does not enforce AES-NI at run time. Pls kindly double check https://github.com/Mbed-TLS/mbedtls/blob/v3.4.0/library/aes.c#L581-L585 and https://github.com/Mbed-TLS/mbedtls/blob/v3.4.0/library/aesni.c for details.

> that the AES-NI patch was dropped wrt. upstream fix (and was the upstream fix reviewed to be sufficient for us?)

So dropped this change.

> Also please add "Fixes: #1255" either to PR description or here to the commit.

Done and updated the PR description.
___
*[`subprojects/mbedtls-mbedtls-3.4.0.wrap` line 4 at r1](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRd1HP11dFDdUs8o-on:-NRdH99wCxrBE5cLHKqo:b-9hunvz) ([raw file](https://github.com/gramineproject/gramine/blob/bf0159cdd4c57d25c72894da2283321bbc1fdb47/subprojects/mbedtls-mbedtls-3.4.0.wrap#L4)):*

Done.
</blockquote></details>

Tested it manually. It works fine.
___
*[`subprojects/packagefiles/mbedtls/include/mbedtls/config-pal.h` line 38 at r2](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRdEoid7xYYTb32BxEk:-NRdHFNa0ePPamXXuJ0W:b-3t3ndb) ([raw file](https://github.com/gramineproject/gramine/blob/43101b8fa3fd4c1cd6c9304c2094ce009627efb3/subprojects/packagefiles/mbedtls/include/mbedtls/config-pal.h#L38)):*

Why this was removed, and if there's a reason, why isn't it written either it commit message, or at least in the PR?
</blockquote></details>

We actually did not use `MBEDTLS_SHA224_C` at all. Updated the commit message.

woju (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1256)* status: 11 of 12 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @mkow)

___
*[`-- commits` line 2 at r2](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRdDD-G7M9ezKA0xgfX):*

> From https://github.com/gramineproject/gramine/issues/1255:
> mbedtls 3.4.0 was released earlier today: https://github.com/Mbed-TLS/mbedtls/releases/tag/v3.4.0
> It has fixed a vulnerability which we have mitigated with the patch: https://github.com/gramineproject/gramine/blob/43101b8fa3fd4c1cd6c9304c2094ce009627efb3/subprojects/packagefiles/mbedtls/enforce-aes-ni.patch

I saw your above comment when I submitted the first version of PR. However, after taking a deeper look at the changes in mbedtls, I don't think it's sufficient for us as it still does not enforce AES-NI at run time. Pls kindly double check https://github.com/Mbed-TLS/mbedtls/blob/v3.4.0/library/aes.c#L581-L585 and https://github.com/Mbed-TLS/mbedtls/blob/v3.4.0/library/aesni.c for details.

> that the AES-NI patch was dropped wrt. upstream fix (and was the upstream fix reviewed to be sufficient for us?)

So dropped this change.

> Also please add "Fixes: #1255" either to PR description or here to the commit.

Done and updated the PR description.
</blockquote></details>

Right. I think if you can fail `mbedtsl_aesni_has_support` (https://github.com/Mbed-TLS/mbedtls/blob/v3.4.0/library/aesni.c#L45), then you can still fallthrough into software AES implementation.

@mkow ^ can you read those mbedtls fixes? You did this patch.

If so, do we have a reason to update to 3.4.0? If we don't, then we probably shouldn't, because this change breaks ABI (sonames) for no good reason.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1256)* status: 11 of 12 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @mkow)

___
*[`-- commits` line 2 at r2](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRdDD-G7M9ezKA0xgfX):*
> If so, do we have a reason to update to 3.4.0? If we don't, then we probably shouldn't, because this change breaks ABI (sonames) for no good reason.

Pesonally, I think we still have some reasons to update to 3.4.0, for instance:
1. It still introduced some security feautre updates (e.g., zeroize SSL cache entries when they are freed) along with normal bug fixes (e.g., integer overflow), as the update is recommaned by mbedTLS.
2. `SHA224_C` is now independent from `SHA256_C`. This helps in saving code size and resolving confusions (e.g., discussed somewhere here in https://github.com/gramineproject/gramine/pull/796#pullrequestreview-1055698811) since we do not use it at all in Gramine.
3. It changed the default `MBEDTLS_ECP_WINDOW_SIZE` from 6 to 2. As stated by mbedTLS, the correlation between this define and RSA decryption performance has changed lately due to security fixes. To fix the performance degradation when using default values the window was reduced from 6 to 2, a value that gives the best or close to best results when tested on Cortex-M4 and Intel i7.
4. We made some decisions previously to keep up to the latest version for mbedTLS (as discussed e.g., in https://github.com/gramineproject/gramine/pull/426#pullrequestreview-893552175 and https://github.com/gramineproject/gramine/pull/426#pullrequestreview-893719381).

Anyway, I'm open to it.

dimakuv (changes requested):
> Reviewed 9 of 10 files at r1, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1256)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin and @mkow)

___
*[`-- commits` line 2 at r2](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRdDD-G7M9ezKA0xgfX):*

> If so, do we have a reason to update to 3.4.0? If we don't, then we probably shouldn't, because this change breaks ABI (sonames) for no good reason.

Pesonally, I think we still have some reasons to update to 3.4.0, for instance:
1. It still introduced some security feautre updates (e.g., zeroize SSL cache entries when they are freed) along with normal bug fixes (e.g., integer overflow), as the update is recommaned by mbedTLS.
2. `SHA224_C` is now independent from `SHA256_C`. This helps in saving code size and resolving confusions (e.g., discussed somewhere here in https://github.com/gramineproject/gramine/pull/796#pullrequestreview-1055698811) since we do not use it at all in Gramine.
3. It changed the default `MBEDTLS_ECP_WINDOW_SIZE` from 6 to 2. As stated by mbedTLS, the correlation between this define and RSA decryption performance has changed lately due to security fixes. To fix the performance degradation when using default values the window was reduced from 6 to 2, a value that gives the best or close to best results when tested on Cortex-M4 and Intel i7.
4. We made some decisions previously to keep up to the latest version for mbedTLS (as discussed e.g., in https://github.com/gramineproject/gramine/pull/426#pullrequestreview-893552175 and https://github.com/gramineproject/gramine/pull/426#pullrequestreview-893719381).

Anyway, I'm open to it.
</blockquote></details>

I also think that we should keep mbedTLS to the latest version. Like, why not? Clearly a newer version has bug fixes, so we should use it.
___
*[`-- commits` line 4 at r4](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRgQTe5980Hs8Mjwwjq):*
`upgradation` -> `upgrade`
___
*[`common/src/crypto/adapters/mbedtls_adapter.c` line 512 at r4](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRdStZF15iocbXt20kp:-NRgR65FCL_S_en7u6ts:b6zbgi1) ([raw file](https://github.com/gramineproject/gramine/blob/0dac96274302ab4eb944013858a08f2e15242aeb/common/src/crypto/adapters/mbedtls_adapter.c#L512)):*

In fact I'm not sure about this - previously we were implicitly using simple `memset()`: https://github.com/Mbed-TLS/mbedtls/blob/v3.3.0/library/platform_util.c#L65. Do we require this in mbedTLS PAL to securely scrub sensitive memory bufs?
</blockquote></details>

I vote for removing this. Let's use mbedTLS's default implementation, which should survive compiler optimizations and is also performant (because it uses `memset()`).

The `erase_memory()` function that we have (https://github.com/gramineproject/gramine/blob/0119fd23ebbc892b33868e9ea9296ba6e8061685/common/include/api.h#L403-L410) is highly dubious and not performant at all. So I wouldn't want to use it in any way.

I created an issue about `erase_memory()` here: https://github.com/gramineproject/gramine/issues/1257
___
*[`subprojects/packagefiles/mbedtls/fcntl.patch` line 27 at r4](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRgScys6LIjruCVFEl1:-NRgScys6LIjruCVFEl2:bw5ahic) ([raw file](https://github.com/gramineproject/gramine/blob/0dac96274302ab4eb944013858a08f2e15242aeb/subprojects/packagefiles/mbedtls/fcntl.patch#L27)):*
> ```Diff
>       * Never return 'WOULD BLOCK' on a blocking socket
>       */
> -    if ((fcntl(ctx->fd, F_GETFL) & O_NONBLOCK) != O_NONBLOCK) {
> ```

Wow, mbedTLS finally uses adequate formatting :) I was always surprised when I looked at their code, how unreadable this formatting was.
___
*[`subprojects/packagefiles/mbedtls/include/mbedtls/config-pal.h` line 37 at r4](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRgT1-a9TQsN7_wFjXX:-NRgT1-a9TQsN7_wFjXY:b-zhnmor) ([raw file](https://github.com/gramineproject/gramine/blob/0dac96274302ab4eb944013858a08f2e15242aeb/subprojects/packagefiles/mbedtls/include/mbedtls/config-pal.h#L37)):*
> ```objc
> #define MBEDTLS_PKCS1_V15
> #define MBEDTLS_PLATFORM_C
> #define MBEDTLS_PLATFORM_ZEROIZE_ALT
> ```

I'd remove this, as discussed in the other comment

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1256)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "squash! " found in commit messages' one-liners (waiting on @kailun-qin and @mkow)

___
*[`-- commits` line 32 at r5](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRgtAi4DaYblp1JCRiP):*
space before `in the mbedTLS ...`
___
*[`common/src/crypto/adapters/mbedtls_adapter.c` line 512 at r4](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRdStZF15iocbXt20kp:-NRgtTx-1VBmmxJk4upA:bn71xz) ([raw file](https://github.com/gramineproject/gramine/blob/0dac96274302ab4eb944013858a08f2e15242aeb/common/src/crypto/adapters/mbedtls_adapter.c#L512)):*

Switched to mbedTLS's built-in implementation using the trick of a volatile function pointer to `memset()` instead of our `erase_memory`.
</blockquote></details>

Thanks, now it makes sense!
___
*[`subprojects/packagefiles/mbedtls/include/mbedtls/config-pal.h` line 37 at r4](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRgT1-a9TQsN7_wFjXX:-NRgt_Ek8EgiXBhDPp6b:b-wsvas3) ([raw file](https://github.com/gramineproject/gramine/blob/0dac96274302ab4eb944013858a08f2e15242aeb/subprojects/packagefiles/mbedtls/include/mbedtls/config-pal.h#L37)):*

The reason I explicitly set this is due to that mbedTLS 3.4.0 has added a logic of platform-provided secure memset auto detection at compile time. Pls kindly see https://github.com/Mbed-TLS/mbedtls/blob/v3.4.0/library/platform_util.c#L52-L57 and https://github.com/Mbed-TLS/mbedtls/blob/v3.4.0/library/platform_util.c#L69-L130.

However, at the linking phase, we may then encounter some issues like
```
/usr/local/bin/ld: subprojects/mbedtls-mbedtls-3.4.0/libmbedcrypto_pal.a(platform_util.o): in function `mbedtls_platform_zeroize':
platform_util.c:(.text+0xd8): undefined reference to `__explicit_bzero_chk'
```

I stated the reason in the PR description and now in the comment as well.
</blockquote></details>

Thanks for the explanation.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1256)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "squash! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`-- commits` line 32 at r5](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRgtAi4DaYblp1JCRiP):*

space before `in the mbedTLS ...`
</blockquote></details>

Fixed, thanks!

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1256)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "squash! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`-- commits` line 2 at r2](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRdDD-G7M9ezKA0xgfX):*
OK, this seems to be reasonable. Unblocking, but I'd still be grateful if @mkow read through this upstream (non-)fix.

> Like, why not?

Because this is incompatible update: 2 out of 3 `mbed*_gramine.so` libraries will change SONAME? This means that if someone uses our distribution of mbedtls (and users should at least because of our mitigation of AES-NI problem), s/he'll need to at least recompile the app. Not fun, but obviously can be done if the reason is compelling.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1256)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "squash! " found in commit messages' one-liners (waiting on @mkow)

___
*[`-- commits` line 2 at r2](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRdDD-G7M9ezKA0xgfX):*
> This means that if someone uses our distribution of mbedtls (and users should at least because of our mitigation of AES-NI problem), s/he'll need to at least recompile the app. Not fun, but obviously can be done if the reason is compelling.

Do SONAMEs really prevent the loader to load the newer-version shared lib? Ok, then I agree that it's bad...

mkow (changes requested):
> Reviewed 8 of 10 files at r1, 3 of 3 files at r3, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1256)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "squash! " found in commit messages' one-liners (waiting on @dimakuv, @kailun-qin, and @woju)

___
*[`-- commits` line 2 at r2](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRdDD-G7M9ezKA0xgfX):*
> I saw your above comment when I submitted the first version of PR. However, after taking a deeper look at the changes in mbedtls, I don't think it's sufficient for us as it still does not enforce AES-NI at run time. Pls kindly double check https://github.com/Mbed-TLS/mbedtls/blob/v3.4.0/library/aes.c#L581-L585 and https://github.com/Mbed-TLS/mbedtls/blob/v3.4.0/library/aesni.c for details.

I see some changes there, but none of them seem to be related to the problem. It still looks vulnerable.

> Also please add "Fixes: #1255" either to PR description or here to the commit.

@woju: we only add these to PR descriptions.

Also +1 from my side to updating it even if it doesn't fix the AES-NI problem. I don't think we have too many users which would be affected by this and the update includes fixes related to security and performance.
___
*[`common/src/crypto/adapters/mbedtls_adapter.c` line 512 at r4](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRdStZF15iocbXt20kp:-NRiPCC-1785XIaGzrkV:btd555b) ([raw file](https://github.com/gramineproject/gramine/blob/0dac96274302ab4eb944013858a08f2e15242aeb/common/src/crypto/adapters/mbedtls_adapter.c#L512)):*

Thanks, now it makes sense!
</blockquote></details>

I'm against this change, see https://github.com/gramineproject/gramine/issues/1257.
___
*[`subprojects/packagefiles/mbedtls/fcntl.patch` line 27 at r4](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRgScys6LIjruCVFEl1:-NRjyn6x55qxfg3H2Fof:bqm82qq) ([raw file](https://github.com/gramineproject/gramine/blob/0dac96274302ab4eb944013858a08f2e15242aeb/subprojects/packagefiles/mbedtls/fcntl.patch#L27)):*

Wow, mbedTLS finally uses adequate formatting :) I was always surprised when I looked at their code, how unreadable this formatting was.
</blockquote></details>

Yeah. But it was still nothing compared to GNU style :)

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1256)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "squash! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @woju)

___
*[`common/src/crypto/adapters/mbedtls_adapter.c` line 512 at r4](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRdStZF15iocbXt20kp:-NRk4sl15G1c9BX6gS6T:b-dnrjln) ([raw file](https://github.com/gramineproject/gramine/blob/0dac96274302ab4eb944013858a08f2e15242aeb/common/src/crypto/adapters/mbedtls_adapter.c#L512)):*

I'm against this change, see https://github.com/gramineproject/gramine/issues/1257.
</blockquote></details>

For this change here (orthogonal to replace `erase_memory()` with more performant alternative in #1257), since we'll anyway need `mbedtls_platform_zeroize()` to skip mbedTLS's auto detection, I'll just remove the volatile function pointer trick to avoid any potential misleading and simply use our memset instead. WDYT?

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1256)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "squash! " found in commit messages' one-liners (waiting on @mkow)

___
*[`common/src/crypto/adapters/mbedtls_adapter.c` line 512 at r4](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRdStZF15iocbXt20kp:-NRlP7WK4JNuatbK4-ns:b-2mzq0u) ([raw file](https://github.com/gramineproject/gramine/blob/0dac96274302ab4eb944013858a08f2e15242aeb/common/src/crypto/adapters/mbedtls_adapter.c#L512)):*
> I'll just remove the volatile function pointer trick to avoid any potential misleading and simply use our memset instead. WDYT?

Yes please.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1256)* status: 11 of 12 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`common/src/crypto/adapters/mbedtls_adapter.c` line 512 at r4](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRdStZF15iocbXt20kp:-NRlTYyj-sTaEuWiynQP:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/0dac96274302ab4eb944013858a08f2e15242aeb/common/src/crypto/adapters/mbedtls_adapter.c#L512)):*

> I'll just remove the volatile function pointer trick to avoid any potential misleading and simply use our memset instead. WDYT?

Yes please.
</blockquote></details>

Done.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1256)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

___
*[`common/src/crypto/adapters/mbedtls_adapter.c` line 512 at r4](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRdStZF15iocbXt20kp:-NRnPmFy8Vm3Cu8PEJ4H:bavgs57) ([raw file](https://github.com/gramineproject/gramine/blob/0dac96274302ab4eb944013858a08f2e15242aeb/common/src/crypto/adapters/mbedtls_adapter.c#L512)):*

Done.
</blockquote></details>

But why `memset()` instead of `erase_memory()`? `memset()` is very likely to be removed by the optimizer.
We'll take care of its performance when resolving #1257.

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1256)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`common/src/crypto/adapters/mbedtls_adapter.c` line 512 at r4](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRdStZF15iocbXt20kp:-NRq_AOb5dNDy623rL1M:brwe8ua) ([raw file](https://github.com/gramineproject/gramine/blob/0dac96274302ab4eb944013858a08f2e15242aeb/common/src/crypto/adapters/mbedtls_adapter.c#L512)):*

But why `memset()` instead of `erase_memory()`? `memset()` is very likely to be removed by the optimizer.
We'll take care of its performance when resolving #1257.
</blockquote></details>

`memset()` is written in volatile inline asm in Gramine for x86-64 (as @kailun-qin mentioned in #1257): https://github.com/gramineproject/gramine/blob/25ddc0effbf84c4a6b1a2d4bce93f3113f1ec04b/common/src/string/memset.c#L27

So it's fine. I would add a comment above memset invocation that it's written in pure asm so won't be optimized away (and say smth like `see common/src/string/memset.c`)

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1256)* status: 11 of 12 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`common/src/crypto/adapters/mbedtls_adapter.c` line 512 at r4](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRdStZF15iocbXt20kp:-NS4B4rTDTJApwO_UHba:b-v9vxsn) ([raw file](https://github.com/gramineproject/gramine/blob/0dac96274302ab4eb944013858a08f2e15242aeb/common/src/crypto/adapters/mbedtls_adapter.c#L512)):*

`memset()` is written in volatile inline asm in Gramine for x86-64 (as @kailun-qin mentioned in #1257): https://github.com/gramineproject/gramine/blob/25ddc0effbf84c4a6b1a2d4bce93f3113f1ec04b/common/src/string/memset.c#L27

So it's fine. I would add a comment above memset invocation that it's written in pure asm so won't be optimized away (and say smth like `see common/src/string/memset.c`)
</blockquote></details>

Added a comment, thanks!

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1256)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1256)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

___
*[`common/src/crypto/adapters/mbedtls_adapter.c` line 512 at r4](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRdStZF15iocbXt20kp:-NS6ovfgE4uiiYmcIiNq:b-gv5km3) ([raw file](https://github.com/gramineproject/gramine/blob/0dac96274302ab4eb944013858a08f2e15242aeb/common/src/crypto/adapters/mbedtls_adapter.c#L512)):*
> `memset()` is written in volatile inline asm in Gramine for x86-64 [...] So it's fine.

It's not, see https://github.com/gramineproject/gramine/issues/1257#issuecomment-1494515912.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1256)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`common/src/crypto/adapters/mbedtls_adapter.c` line 512 at r4](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRdStZF15iocbXt20kp:-NS75fxDCMHlpPOKtuPK:bkzed52) ([raw file](https://github.com/gramineproject/gramine/blob/0dac96274302ab4eb944013858a08f2e15242aeb/common/src/crypto/adapters/mbedtls_adapter.c#L512)):*

> `memset()` is written in volatile inline asm in Gramine for x86-64 [...] So it's fine.

It's not, see https://github.com/gramineproject/gramine/issues/1257#issuecomment-1494515912.
</blockquote></details>

Ok, my last attempt: just directly insert the inline asm: https://github.com/gramineproject/gramine/blob/25ddc0effbf84c4a6b1a2d4bce93f3113f1ec04b/common/src/string/memset.c#L27?plain=1

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1256)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`common/src/crypto/adapters/mbedtls_adapter.c` line 512 at r4](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRdStZF15iocbXt20kp:-NS8oIRK-yTxzrmVxffV:b-pojd44) ([raw file](https://github.com/gramineproject/gramine/blob/0dac96274302ab4eb944013858a08f2e15242aeb/common/src/crypto/adapters/mbedtls_adapter.c#L512)):*
> Ok, my last attempt: just directly insert the inline asm: https://github.com/gramineproject/gramine/blob/25ddc0effbf84c4a6b1a2d4bce93f3113f1ec04b/common/src/string/memset.c#L27?plain=1

Shouldn't we use `erase_memory()` here and insert the inline asm in `erase_memory()` (i.e., resolve https://github.com/gramineproject/gramine/issues/1257)? Just as @mkow suggested above.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1256)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`-- commits` line 2 at r2](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRdDD-G7M9ezKA0xgfX):*

> I saw your above comment when I submitted the first version of PR. However, after taking a deeper look at the changes in mbedtls, I don't think it's sufficient for us as it still does not enforce AES-NI at run time. Pls kindly double check https://github.com/Mbed-TLS/mbedtls/blob/v3.4.0/library/aes.c#L581-L585 and https://github.com/Mbed-TLS/mbedtls/blob/v3.4.0/library/aesni.c for details.

I see some changes there, but none of them seem to be related to the problem. It still looks vulnerable.

> Also please add "Fixes: #1255" either to PR description or here to the commit.

@woju: we only add these to PR descriptions.

Also +1 from my side to updating it even if it doesn't fix the AES-NI problem. I don't think we have too many users which would be affected by this and the update includes fixes related to security and performance.
</blockquote></details>

FYI, a PR WIP: https://github.com/Mbed-TLS/mbedtls/pull/7384 may serve our purpose of the AES-NI patch.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1256)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`common/src/crypto/adapters/mbedtls_adapter.c` line 512 at r4](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRdStZF15iocbXt20kp:-NSA6BlD4TS8JKtSFzgA:b-9g3o61) ([raw file](https://github.com/gramineproject/gramine/blob/0dac96274302ab4eb944013858a08f2e15242aeb/common/src/crypto/adapters/mbedtls_adapter.c#L512)):*

> Ok, my last attempt: just directly insert the inline asm: https://github.com/gramineproject/gramine/blob/25ddc0effbf84c4a6b1a2d4bce93f3113f1ec04b/common/src/string/memset.c#L27?plain=1

Shouldn't we use `erase_memory()` here and insert the inline asm in `erase_memory()` (i.e., resolve https://github.com/gramineproject/gramine/issues/1257)? Just as @mkow suggested above.

</blockquote></details>

Ah, yes, please! Even better, we'll solve two problems then.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1256)* status: 11 of 13 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "squash! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`common/src/crypto/adapters/mbedtls_adapter.c` line 512 at r4](https://reviewable.io/reviews/gramineproject/gramine/1256#-NRdStZF15iocbXt20kp:-NSAerHA3bmeRTQVqA0F:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/0dac96274302ab4eb944013858a08f2e15242aeb/common/src/crypto/adapters/mbedtls_adapter.c#L512)):*

Ah, yes, please! Even better, we'll solve two problems then.
</blockquote></details>

Done.

mkow (dismissed):
> Reviewed 2 of 2 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1256)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`common/include/api.h` line 408 at r10](https://reviewable.io/reviews/gramineproject/gramine/1256#-NSBUlD-1NUtVZSxaEqI:-NSBku7qDi4SCNRj0G3M:bszqclz) ([raw file](https://github.com/gramineproject/gramine/blob/7ba366f0faaa1a5873b86c0667a1ced862cb48f0/common/include/api.h#L408)):*

It was simply made for casting purpose. Otherwise casting in the inline asm and building w/ clang will fail with "invalid use of a cast in a inline asm context requiring an l-value". Anyway, I think it's fine that we remove the casting.
</blockquote></details>

Sounds weird, `buffer` is already an l-value, why would you need a cast/another variable there?
Anyways, looks good now.

dimakuv (dismissed):
> Reviewed 2 of 2 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1256)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

woju (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1256)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1256)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1256)* status: :shipit: complete! all files reviewed, all discussions resolved

woju (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1256)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #1268, opened 2023-04-04
[LibOS] Keep main thread alive in case of non-main thread `execve()`

### LINKED ISSUE #1265: Host process becomes a zombie when calling execve from another than the initial thread

I have the following program:

```c
#include <pthread.h>
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>

void* f(void* p) {
    char* argv[] = {p, "", NULL};
    execve(p, argv, NULL);
    abort();
}

int main(int argc, char** argv) {
    if (argc == 1) {
        pthread_t t;
        pthread_create(&t, NULL, f, argv[0]);
        pthread_join(t, NULL);
        abort();
    }

    for (;;) {
        // fputs("a\n", stderr);
        sleep(1);
    }
}
```

When I run this with Gramine, the host process becomes a zombie:

```
$ gramine-direct helloworld &
[1] 1457
$ ps
    PID TTY          TIME CMD
   1436 pts/0    00:00:00 bash
   1457 pts/0    00:00:00 exe <defunct>
   1461 pts/0    00:00:00 ps
```

I think this is because the execve implementation kills all other threads and the initial thread is gone.

### Steps to reproduce

_No response_

### Expected results

_No response_

### Actual results

_No response_

### Gramine commit hash

v1.4

dimakuv:
## Description of the changes 

There is a corner case of a non-main thread performing `execve()`. Linux does the following: the main (leader) thread is terminated, and the non-main thread assumes its identity: https://elixir.bootlin.com/linux/v6.0/source/fs/exec.c#L1078

Before this PR, Gramine simply terminated the main thread. This led to the process being marked as a zombie (even though the non-main thread runs correctly and responds to signals). This in turn confused tools like `docker kill`, which couldn't find the Gramine process anymore. See for details #1265.

This PR fixes this corner case. Gramine can't do as Linux because there is no way to ask the host OS to "rewire" the identity of one thread to the other thread. Thus we introduce a workaround of keeping alive (but in a "parked" state) the main thread -- it sleeps infinitely.

In this case, the main thread and its associated resources (like the thread's LibOS stack) are not freed -- this leaks memory, but only once per process, as there is only one main thread per process, even after several execve invocations.

Fixes #1265.

## How to test this PR? 

I don't know how to test it in our CI reliably. We already have the test `exec_same` which does `execve()` in the non-main thread.

Anyther way to test it manually is to build and run the program in #1265 -- it will show as `defunct` when doing `ps` or similar without this PR. With this PR, it shows correctly.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1268)

### DIFF
--- a/libos/include/libos_thread.h
+++ b/libos/include/libos_thread.h
@@ -20,6 +20,8 @@
 #include "list.h"
 #include "pal.h"
 
+extern bool g_execve_happenning;
+
 #define WAKE_QUEUE_TAIL ((void*)1)
 /* If next is NULL, then this node is not on any queue.
  * Otherwise it is a valid pointer to the next node or WAKE_QUEUE_TAIL. */
--- a/libos/src/sys/libos_exec.c
+++ b/libos/src/sys/libos_exec.c
@@ -15,6 +15,8 @@
 #include "libos_vma.h"
 #include "pal.h"
 
+bool g_execve_happenning;
+
 /* new_argp: pointer to beginning of first stack frame (argc, argv[0], ...)
  * new_auxv: pointer inside first stack frame (auxv[0], auxv[1], ...) */
 noreturn static void __libos_syscall_execve_rtld(void* new_argp, elf_auxv_t* new_auxv) {
@@ -174,6 +176,8 @@ long libos_syscall_execve(const char* file, const char* const* argv, const char*
         return ret;
     }
 
+    __atomic_store_n(&g_execve_happenning, true, __ATOMIC_RELAXED);
+
     /* If `execve` is invoked concurrently by multiple threads, let only one succeed. From this
      * point errors are fatal. */
     static unsigned int first = 0;
@@ -183,9 +187,10 @@ long libos_syscall_execve(const char* file, const char* const* argv, const char*
     }
     (void)kill_other_threads();
 
-    /* All other threads are dead. Restoring initial value in case we stay inside same process
+    /* All other threads are dead. Restoring initial values in case we stay inside same process
      * instance and call execve again. */
     __atomic_store_n(&first, 0, __ATOMIC_RELAXED);
+    __atomic_store_n(&g_execve_happenning, false, __ATOMIC_RELAXED);
 
     /* Passing ownership of `exec` and `new_argv`. */
     ret = libos_syscall_execve_rtld(exec, new_argv, envp);
--- a/libos/src/sys/libos_exit.c
+++ b/libos/src/sys/libos_exit.c
@@ -69,8 +69,64 @@ noreturn void thread_exit(int error_code, int term_signal) {
         cur_thread->robust_list = NULL;
     }
 
+    bool keep_alive_main_thread = false;
+    if (__atomic_load_n(&g_execve_happenning, __ATOMIC_RELAXED)
+            && cur_thread->pal_handle == g_pal_public_state->first_thread) {
+        /*
+         * The main thread ends up here in three cases:
+         *
+         *   1. Common case: a non-main thread issues `execve()`, which leads to killing all other
+         *      threads (including this main thread). The main thread is kicked into signal handling
+         *      and ends up here. At this point, the non-main thread is waiting on the
+         *      `check_last_thread()` loop, so there is no data race (until the main thread calls
+         *      `check_last_thread(true)` below).
+         *
+         *   2. Corner-case data race: main thread issues `exit()` while a non-main thread issues
+         *      `execve()`. In this case, the data race is always resolved as if the non-main thread
+         *      issued `execve()` before the main-thread's `exit()`.
+         *
+         *   3. Corner-case data race: both main thread and a non-main thread issue `execve()`. If
+         *      the non-main thread wins the race, then the main thread calls `thread_exit()` and
+         *      ends up here; see `libos_syscall_execve()` for details. (If the main thread wins,
+         *      then we cannot land here, as discussed below.)
+         *
+         *   Note that if the main thread itself issues `execve()`, we cannot land here because only
+         *   non-main threads will be kicked into signal handling.
+         */
+        keep_alive_main_thread = true;
+    }
+
     /* Remove current thread from the threads list. */
     if (!check_last_thread(/*mark_self_dead=*/true)) {
+        if (keep_alive_main_thread) {
+            /*
+             * This is a corner case of a non-main thread performing `execve()`.
+             *
+             * Do not exit the main thread and do not free thread resources -- this leaks memory,
+             * but only once per process (as there is only one main thread per process, even after
+             * several execve invocations). Instead, wait forever so that the host OS doesn't "lose
+             * track" of this Gramine process. E.g. on Linux, if the main thread (aka leader thread)
+             * terminates, then the process becomes a zombie, which may confuse some tools like
+             * `docker kill`.
+             *
+             * Linux solves this corner case differently: the leader thread is terminated, and the
+             * non-main thread assumes its identity (in particular, its PID):
+             *
+             *   https://elixir.bootlin.com/linux/v6.0/source/fs/exec.c#L1078
+             *
+             * Gramine can't do the same because there is no way to ask the host OS to "rewire" the
+             * identity of one thread to the other thread. Thus this workaround of infinite wait.
+             * Note that because the main thread was removed from the list of threads (thanks to
+             * `mark_self_dead=true` above), the still-alive main thread will not prevent the
+             * Gramine process from terminating later on. Also note that because this thread never
+             * leaves LibOS/PAL context, it will not receive signals.
+             */
+            thread_prepare_wait();
+            while (true)
+                thread_wait(/*timeout_us=*/NULL, /*ignore_pending_signals=*/true);
+            __builtin_unreachable();
+        }
+
         /* ask async worker thread to cleanup this thread */
         cur_thread->clear_child_tid_pal = 1; /* any non-zero value suffices */
         /* We pass this ownership to `cleanup_thread`. */


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1268)* status: 0 of 3 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

___
*[`libos/src/sys/libos_exec.c` line 179 at r1](https://reviewable.io/reviews/gramineproject/gramine/1268#-NSCEdLj4ThDjcBijOn2:-NSCEdLj4ThDjcBijOn3:bx6rzg2) ([raw file](https://github.com/gramineproject/gramine/blob/4072a61dca69d9d7f98a2f56ffee4de0e8f35098/libos/src/sys/libos_exec.c#L179)):*
> ```C
>     }
> 
>     __atomic_store_n(&g_execve_happenning, true, __ATOMIC_RELAXED);
> ```

This global variable is atomic only because of this place -- two threads can race on `execve()` syscall. We don't care which one of the threads will set this variable, we just want to make sure that this assignment is atomic.

Other places where we touch this variable are guaranteed to be synchronized (on the `first` var below and on `kill_other_threads()`/`check_last_thread()`). E.g., when we set `g_execve_happenning` to `false`, there are no other threads so no data races. And e.g., when we check `g_execve_happenning`, then we have benign data races which I described in another file.
___
*[`libos/src/sys/libos_exit.c` line 74 at r1](https://reviewable.io/reviews/gramineproject/gramine/1268#-NSCEEfF5xJ1bdXZzWSg:-NSCEEfF5xJ1bdXZzWSh:brl6g1n) ([raw file](https://github.com/gramineproject/gramine/blob/4072a61dca69d9d7f98a2f56ffee4de0e8f35098/libos/src/sys/libos_exit.c#L74)):*
> ```C
>     bool keep_alive_main_thread = false;
>     if (__atomic_load_n(&g_execve_happenning, __ATOMIC_RELAXED)
>             && cur_thread->pal_handle == g_pal_public_state->first_thread) {
> ```

@mkow suggested to introduce a new field like `libos_thread::is_main_thread_on_host`.

However, we already have `g_pal_public_state->first_thread`, so we can just use this to check whether this is the main-host-thread. We do it also in a couple other places.
___
*[`libos/src/sys/libos_exit.c` line 94 at r1](https://reviewable.io/reviews/gramineproject/gramine/1268#-NSCEXl54p2e57Dh24gr:-NSCEXl54p2e57Dh24gs:b-4zo1jn) ([raw file](https://github.com/gramineproject/gramine/blob/4072a61dca69d9d7f98a2f56ffee4de0e8f35098/libos/src/sys/libos_exit.c#L94)):*
> ```C
>          *
>          *   Note that if the main thread itself issues `execve()`, we cannot land here because only
>          *   non-main threads will be kicked into signal handling.
> ```

I feel like these comments are way too detailed and useless. Please tell me where/how to cut these.
___
*[`libos/src/sys/libos_exit.c` line 122 at r1](https://reviewable.io/reviews/gramineproject/gramine/1268#-NSCFp0JB_LlEEhIQGT3:-NSCFp0JB_LlEEhIQGT4:b-pr80ra) ([raw file](https://github.com/gramineproject/gramine/blob/4072a61dca69d9d7f98a2f56ffee4de0e8f35098/libos/src/sys/libos_exit.c#L122)):*
> ```C
>              * `mark_self_dead=true` above), the still-alive main thread will not prevent the
>              * Gramine process from terminating later on. Also note that because this thread never
>              * leaves LibOS/PAL context, it will not receive signals.
> ```

The "parking" part is pretty complex to reason about. I tried to capture the two most important aspects (and hopefully the only ones): the thread "disappears" from Gramine process because it is taken off the threads list, and the thread doesn't react to signals (like SIGTERM) because it never goes back to the app context (or in other words, it never leaves the LibOS/PAL context), so signal handling has no change to happen on this thread.

Would be good if maybe @boryspoplawski could chime in here?

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1268)* status: 0 of 3 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`libos/src/sys/libos_exec.c` line 179 at r1](https://reviewable.io/reviews/gramineproject/gramine/1268#-NSCrh3uDjEi1vCVhUBx:-NSCrh3uDjEi1vCVhUBy:b-puuodm) ([raw file](https://github.com/gramineproject/gramine/blob/4072a61dca69d9d7f98a2f56ffee4de0e8f35098/libos/src/sys/libos_exec.c#L179)):*
> ```C
>     }
> 
>     __atomic_store_n(&g_execve_happenning, true, __ATOMIC_RELAXED);
> ```

Wait, how does this prevent any races? What are you trying to accomplish with this variable? I.e. what interlacing of execution is impossible because of it?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1268)* status: 0 of 3 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

___
*[`libos/src/sys/libos_exec.c` line 179 at r1](https://reviewable.io/reviews/gramineproject/gramine/1268#-NSCrh3uDjEi1vCVhUBx:-NSFPg9TEjCbaHXItX6e:b-xuj7pt) ([raw file](https://github.com/gramineproject/gramine/blob/4072a61dca69d9d7f98a2f56ffee4de0e8f35098/libos/src/sys/libos_exec.c#L179)):*

Wait, how does this prevent any races? What are you trying to accomplish with this variable? I.e. what interlacing of execution is impossible because of it?
</blockquote></details>

Are you asking why (1) I introduce this variable at all, or (2) why is this variable atomic?

**Why I introduce this variable**. Main thread may end up in `thread_exit()` for reasons not related to `execve()` -- for example, some thread of the application issued a `exit_group()` system call. In this case, the main thread should just terminate as usual.

**Why this variable is atomic**. Simply so that when two threads invoke `execve()` concurrently, this variable should be atomically updated by each of these threads. I could technically keep it non-atomic because I don't think it's possible to have this variable is some weird partially-updated state afterwards anyway, but I thought I should be more specific and make it atomic here. UPDATE: now that I think of it, I could remove `first` variable below and use my new variable `g_execve_happenning` to serve as a gate.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1268)* status: 0 of 3 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1268#-NSGOC5xBBro-xeq_ZSr:-NSGOC5y911th4OHwzaY:b-g58i5v) (no related file):*
What if I do
```
clone();
if (child) {
    sleep(10);
    execve();
} else {
    // parent
    exit_thread();
}
```
IIUC it would not work

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1268)* status: 0 of 3 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1268#-NSGOC5xBBro-xeq_ZSr:-NSGV_9g79VUTxomAgu_:b-fkgfs9) (no related file):*

What if I do
```
clone();
if (child) {
    sleep(10);
    execve();
} else {
    // parent
    exit_thread();
}
```
IIUC it would not work
</blockquote></details>

@boryspoplawski Instead of `clone()`, did you mean `fork()`?

Or `clone()` in the sense of creating a thread? Then the `child` is a child *thread*? If not and it's a process, then where is `fork()`?

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1268)* status: 0 of 3 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1268#-NSGOC5xBBro-xeq_ZSr:-NSGZyvE4Z13MCm2rsbL:b7umms4) (no related file):*

@boryspoplawski Instead of `clone()`, did you mean `fork()`?

Or `clone()` in the sense of creating a thread? Then the `child` is a child *thread*? If not and it's a process, then where is `fork()`?
</blockquote></details>

Ok, I understand what you mean. You meant threads by `child` and `parent`, and spawning a thread by `clone()`.

Yes, this is an interesting corner case which works correctly on Linux and is not solved in Gramine (not by my current PR, I mean). What happens on Linux is like this:
1. The parent thread exits, but the child thread is still alive (and sleeping for 10 sec) -- at this point Linux considers this process as `defunct`
2. Then the child wakes up and performs `execve()` at which point it assumes the main-thread's identity (even though the main thread is already terminated!) -- at this point Linux re-considers this process to be "normal" again

This is a pretty insane behavior of Linux. Do we want to support such case in Gramine? This would require **always parking** the main thread instead of terminating it -- who knows if some child thread will decide to perform `execve()` in the future...

**Unless people suggest a good way of imitating Linux in this particular case, I would argue that we shouldn't try to mimic this particular Linux quirk.**

---

For those who want to play with this case:
```diff
diff --git a/CI-Examples/helloworld/Makefile b/CI-Examples/helloworld/Makefile
index cefca5f0..57aa926d 100644
--- a/CI-Examples/helloworld/Makefile
+++ b/CI-Examples/helloworld/Makefile
@@ -1,10 +1,11 @@
-CFLAGS = -Wall -Wextra
+CFLAGS = -Wall -Wextra -pthread
+LDFLAGS = -pthread

 ifeq ($(DEBUG),1)

diff --git a/CI-Examples/helloworld/helloworld.c b/CI-Examples/helloworld/helloworld.c
index 2efc7765..70ab5f2b 100644
--- a/CI-Examples/helloworld/helloworld.c
+++ b/CI-Examples/helloworld/helloworld.c
@@ -1,6 +1,25 @@
+#include <pthread.h>
 #include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>

-int main(void) {
-    printf("Hello, world\n");
-    return 0;
+void* f(void* p) {
+    sleep(10);
+    char* argv[] = {p, "", NULL};
+    execve(p, argv, NULL);
+    abort();
+}
+
+int main(int argc, char** argv) {
+    if (argc == 1) {
+        pthread_t t;
+        pthread_create(&t, NULL, f, argv[0]);
+        sleep(1);
+        pthread_exit(NULL);
+    }
+
+    for (;;) {
+        // fputs("a\n", stderr);
+        sleep(1);
+    }
 }

diff --git a/CI-Examples/helloworld/helloworld.manifest.template b/CI-Examples/helloworld/helloworld.manifest.template
index e15f0d2b..10763cc4 100644
--- a/CI-Examples/helloworld/helloworld.manifest.template
+++ b/CI-Examples/helloworld/helloworld.manifest.template
@@ -4,13 +4,16 @@ loader.entrypoint = "file:{{ gramine.libos }}"
 libos.entrypoint = "/helloworld"
 loader.log_level = "{{ log_level }}"

-loader.env.LD_LIBRARY_PATH = "/lib"
+loader.env.LD_LIBRARY_PATH = "/lib:/lib/x86_64-linux-gnu"

 fs.mounts = [
   { path = "/lib", uri = "file:{{ gramine.runtimedir() }}" },
   { path = "/helloworld", uri = "file:helloworld" },
+  { path = "/lib/x86_64-linux-gnu/libgcc_s.so.1", uri = "file:/lib/x86_64-linux-gnu/libgcc_s.so.1" },
 ]

+sys.enable_sigterm_injection = true
+
 sgx.debug = true
 sgx.edmm_enable = {{ 'true' if env.get('EDMM', '0') == '1' else 'false' }}

@@ -18,4 +21,5 @@ sgx.trusted_files = [
   "file:{{ gramine.libos }}",
   "file:helloworld",
   "file:{{ gramine.runtimedir() }}/",
+  "file:/lib/x86_64-linux-gnu/libgcc_s.so.1",
 ]
```

Then `make SGX=1` and run `./helloworld`, `gramine-direct helloworld` and `gramine-sgx helloworld` (each for at least 10 seconds, because the child thread sleeps for 10 seconds). Look at what Linux's `ps` thinks about the process like this: `ps -aux | grep -E "helloworld|loader|\[exe\]"`.

P.S. The addition of `libgcc_s.so.1` is required for `pthread_exit()` to work -- no idea why.

___

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1268)* status: 0 of 3 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski and @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1268#-NSGOC5xBBro-xeq_ZSr:-NSGtZW1BFPNhN1C5frK:b65j4i2) (no related file):*
> Yes, this is an interesting corner case which works correctly on Linux and is not solved in Gramine

Wait, I thought that's the very thing this PR is supposed to fix? Why not to always park the main thread if it gets killed too early?

___
___
*[`libos/src/sys/libos_exec.c` line 179 at r1](https://reviewable.io/reviews/gramineproject/gramine/1268#-NSCrh3uDjEi1vCVhUBx:-NSHMva8AyJnJzWkD5xK:b-34o1q1) ([raw file](https://github.com/gramineproject/gramine/blob/4072a61dca69d9d7f98a2f56ffee4de0e8f35098/libos/src/sys/libos_exec.c#L179)):*
> In this case, the main thread should just terminate as usual.

But this code doesn't ensure this? If the main thread goes through `__atomic_load_n(&g_execve_happenning, __ATOMIC_RELAXED)` and then we go through this line we end up in a state you say you are trying to prevent?

So, I have no idea what these checks here actually ensure...

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1268)* status: 0 of 3 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1268#-NSGOC5xBBro-xeq_ZSr:-NSKACV5AdLWy6X2LGKs:b-qco2g3) (no related file):*
> Wait, I thought that's the very thing this PR is supposed to fix?

No. This PR is supposed to fix the following scenario:
- **Two threads run, a main thread does its own thing, a non-main thread calls execve**.
- Linux kernel **immediately reroutes the identity of the process** to this non-main thread (it becomes main in the new exec image).
- This PR ensures that Gramine keeps the main thread alive but parked, so that the host Linux doesn't mark the process as defunct.

The scenario described by Borys is trickier:
- **Two threads run, main thread exits, then after some time non-main thread calls execve**.
- Linux kernel **first marks the process as defunct, and after some time reroutes the identity of the process** (after the non-main thread calls execve).
- My point in this comment is that Gramine can't support such tricky scenario.

> Why not to always park the main thread if it gets killed too early?

How do you determine what constitutes "killed too early"? How do you decide whether to park the main thread or to terminate it?

___
___
*[`libos/src/sys/libos_exec.c` line 179 at r1](https://reviewable.io/reviews/gramineproject/gramine/1268#-NSCrh3uDjEi1vCVhUBx:-NSKBsFeB8jbytKaF-NG:b-hkvwy2) ([raw file](https://github.com/gramineproject/gramine/blob/4072a61dca69d9d7f98a2f56ffee4de0e8f35098/libos/src/sys/libos_exec.c#L179)):*
> then we go through this line

Who exactly goes through which line exactly? The line that is highlighted by this Reviewable comment is `__atomic_store_n(&g_execve_happenning, true ...)` -- but this line is only triggered if some app thread issues `execve()` syscall.

So are you talking about the race like this: main thread ends up in `thread_exit()` for some reason (e.g. by issuing `exit()` syscall), and some non-main thread issues `execve()` syscall at the same time? In your scenario, if the main thread goes through `__atomic_load_n(&g_execve_happenning ...)` **earlier**, then the main thread simply terminates, yes. But I don't see what else can we do in this case -- this is the scenario Borys described. In this scenario, the main thread wants to terminate before other thread can do `execve()` -- even if the main thread is earlier only by a nanosecond.
___
*[`libos/src/sys/libos_exit.c` line 86 at r1](https://reviewable.io/reviews/gramineproject/gramine/1268#-NSKDFom5HWHRyyOH96v:-NSKDFom5HWHRyyOH96w:b9zepa) ([raw file](https://github.com/gramineproject/gramine/blob/4072a61dca69d9d7f98a2f56ffee4de0e8f35098/libos/src/sys/libos_exit.c#L86)):*
> ```C
>          *   2. Corner-case data race: main thread issues `exit()` while a non-main thread issues
>          *      `execve()`. In this case, the data race is always resolved as if the non-main thread
>          *      issued `execve()` before the main-thread's `exit()`.
> ```

I realize now that this description is not entirely correct. I should describe the two options here: if the main thread is faster to reach `load(g_execve_happenning)` (then it terminates, and the Gramine process becomes defunct, as Borys described). Or if the non-main thread is faster to reach `store(g_execve_happenning)`, then the text in the code comment applies.

I should probably modify this code comment.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1268)* status: 0 of 3 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski and @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1268#-NSGOC5xBBro-xeq_ZSr:-NSbQNr85rcXkB7OKghV:bnf9a2r) (no related file):*
Ok, I see, but:
- Why not also make it work for the more generic case?
- Are real apps actually doing what you described? It seems inherently racy, because you kill the main app thread while it's still doing something. I'd expect apps to first terminate it gracefully and only then execve from another thread?

> - Linux kernel first marks the process as defunct, and after some time reroutes the identity of the process (after the non-main thread calls execve).
> - My point in this comment is that Gramine can't support such tricky scenario.

We don't have to make the host process defunct in this case, just prevent becoming defunct when it shouldn't (i.e. after execve).

> How do you determine what constitutes "killed too early"? How do you decide whether to park the main thread or to terminate it?

Just always park it if it's not the last exiting thread? I don't see why it would be harder than your current change.

___
___
*[`libos/src/sys/libos_exec.c` line 179 at r1](https://reviewable.io/reviews/gramineproject/gramine/1268#-NSCrh3uDjEi1vCVhUBx:-NSbS5iP0hykVzwClgWZ:briypcz) ([raw file](https://github.com/gramineproject/gramine/blob/4072a61dca69d9d7f98a2f56ffee4de0e8f35098/libos/src/sys/libos_exec.c#L179)):*

> then we go through this line

Who exactly goes through which line exactly? The line that is highlighted by this Reviewable comment is `__atomic_store_n(&g_execve_happenning, true ...)` -- but this line is only triggered if some app thread issues `execve()` syscall.

So are you talking about the race like this: main thread ends up in `thread_exit()` for some reason (e.g. by issuing `exit()` syscall), and some non-main thread issues `execve()` syscall at the same time? In your scenario, if the main thread goes through `__atomic_load_n(&g_execve_happenning ...)` **earlier**, then the main thread simply terminates, yes. But I don't see what else can we do in this case -- this is the scenario Borys described. In this scenario, the main thread wants to terminate before other thread can do `execve()` -- even if the main thread is earlier only by a nanosecond.
</blockquote></details>

Yes, that's the case. Would it just work if you made locking here correct? I.e. use a semaphore or something like that.
___
*[`libos/src/sys/libos_exec.c` line 179 at r1](https://reviewable.io/reviews/gramineproject/gramine/1268#-NSbSlXHAhI1ijpGRbwC:-NSbSlXHAhI1ijpGRbwD:bvonnkn) ([raw file](https://github.com/gramineproject/gramine/blob/4072a61dca69d9d7f98a2f56ffee4de0e8f35098/libos/src/sys/libos_exec.c#L179)):*
> ```C
>     }
> 
>     __atomic_store_n(&g_execve_happenning, true, __ATOMIC_RELAXED);
> ```

Regardless of the race mentioned above, is `relaxed` correct here? Especially that we do other stores here and are firing another thread's killing in between.
If it's correct, then why?



_Code quote:_
```C
__ATOMIC_RELAXED
```

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1268)* status: 0 of 3 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1268#-NSGOC5xBBro-xeq_ZSr:-NSjgLzF7JqXthAtDqiY:b-wo1wde) (no related file):*
> Are real apps actually doing what you described? It seems inherently racy, because you kill the main app thread while it's still doing something. I'd expect apps to first terminate it gracefully and only then execve from another thread?

The app from #1265 seems to be doing what I described (the first, racy, scenario).

> We don't have to make the host process defunct in this case, just prevent becoming defunct when it shouldn't (i.e. after execve).

Hm, ok, I guess you're right. We don't need to emulate that Linux quirk. We can just always keep the main thread alive from the host-OS perspective.

> Just always park it if it's not the last exiting thread? I don't see why it would be harder than your current change.

It is not harder. Ok, I agree with you, there seems to be no harm in keeping the main thread always.

**I created a new PR to also support this case.** See #1277. The current PR is supersede by that one.

___
___
*[`libos/src/sys/libos_exec.c` line 179 at r1](https://reviewable.io/reviews/gramineproject/gramine/1268#-NSbSlXHAhI1ijpGRbwC:-NSjh6UM-BfjiuYz9i10:btty9qw) ([raw file](https://github.com/gramineproject/gramine/blob/4072a61dca69d9d7f98a2f56ffee4de0e8f35098/libos/src/sys/libos_exec.c#L179)):*

Regardless of the race mentioned above, is `relaxed` correct here? Especially that we do other stores here and are firing another thread's killing in between.
If it's correct, then why?
</blockquote></details>

It is correct because this `g_execve_happenning` variable doesn't serve as the barrier of any sort here.

What makes the "thread killing" logic thread-safe are the following implementations of functions:
- https://github.com/gramineproject/gramine/blob/0119fd23ebbc892b33868e9ea9296ba6e8061685/libos/src/sys/libos_exit.c#L136
- https://github.com/gramineproject/gramine/blob/0119fd23ebbc892b33868e9ea9296ba6e8061685/libos/src/bookkeep/libos_thread.c#L442

In particular, the `g_thread_list_lock` lock is the one that serves as the barrier:
- https://github.com/gramineproject/gramine/blob/0119fd23ebbc892b33868e9ea9296ba6e8061685/libos/src/bookkeep/libos_thread.c#L446
- https://github.com/gramineproject/gramine/blob/0119fd23ebbc892b33868e9ea9296ba6e8061685/libos/src/bookkeep/libos_thread.c#L460

So the `g_thread_list_lock` will force that all stores are done, even though `g_execve_happenning` is stored in a `relaxed` manner.

UPDATE: Not relevant anymore.
___
*[`libos/src/sys/libos_exec.c` line 179 at r1](https://reviewable.io/reviews/gramineproject/gramine/1268#-NSCrh3uDjEi1vCVhUBx:-NSk5Lfm4fdu7Z43BKfq:b-w5ckpu) ([raw file](https://github.com/gramineproject/gramine/blob/4072a61dca69d9d7f98a2f56ffee4de0e8f35098/libos/src/sys/libos_exec.c#L179)):*

Yes, that's the case. Would it just work if you made locking here correct? I.e. use a semaphore or something like that.
</blockquote></details>

Not relevant anymore.

dimakuv:
This PR is superseded by https://github.com/gramineproject/gramine/pull/1277, since the suggestions by Borys/Michal led to a drastically simplified change.

PR #1270, opened 2023-04-06
[tools] Add `gramine-direct --version` and `gramine-sgx --version`

dimakuv:
## Description of the changes 

Versions, by popular request.

The output follows the `git describe` [format](https://git-scm.com/docs/git-describe).

## How to test this PR? 

Manually (on this PR's commit):
```
$ gramine-direct --version
Version: v1.4-39-gcd6a9cca

$ gramine-sgx --version
Version: v1.4-39-gcd6a9cca
```

In the future releases will look like this:
```
$ git checkout v1.5

$ gramine-direct --version
Version: v1.5

$ gramine-sgx --version
Version: v1.5
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1270)

### DIFF
--- a/tools/gramine-sed.sh
+++ b/tools/gramine-sed.sh
@@ -1,0 +1,25 @@
+#!/bin/sh
+
+set -e
+
+INPUT="$1"
+OUTPUT="$2"
+
+SGX="$3"
+CONFIG_SGX_DRIVER="$4"
+PREFIX="$5"
+BINDIR="$6"
+HOST_PAL_PATH="$7"
+LIBPAL_PATH="$8"
+PAL_CMD="$9"
+
+sed -e "s|@SGX@|$SGX|g" \
+    -e "s|@CONFIG_SGX_DRIVER@|\"$CONFIG_SGX_DRIVER\"|g" \
+    -e "s|@PREFIX@|\"$PREFIX\"|g" \
+    -e "s|@BINDIR@|\"$BINDIR\"|g" \
+    -e "s|@HOST_PAL_PATH@|\"$HOST_PAL_PATH\"|g" \
+    -e "s|@LIBPAL_PATH@|\"$LIBPAL_PATH\"|g" \
+    -e "s|@PAL_CMD@|\"$PAL_CMD\"|g" \
+    "$INPUT" > "$OUTPUT"
+
+chmod +x "$OUTPUT"
--- a/tools/gramine.in
+++ b/tools/gramine.in
@@ -37,6 +37,11 @@ if [ "$PERF" == "1" ]; then
     PREFIX=(perf stat)
 fi
 
+if [ "$1" == "--version" ]; then
+    echo "Version: @GRAMINE_VERSION@"
+    exit 0
+fi
+
 while [ "$1" != "" ];
 do
     if [ "$APPLICATION" == "" ]; then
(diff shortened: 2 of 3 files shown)


oshogbo (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1270#-:-NSMDa8Z-Cj_tow_6ubV:bnfp4nl)* status: 0 of 3 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

mkow (commented):
> Reviewed 1 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1270)* status: 1 of 3 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`tools/meson.build` line 11 at r1](https://reviewable.io/reviews/gramineproject/gramine/1270#-NSO3sfUBL2kTU_JstVj:-NSO3sfUBL2kTU_JstVk:b4hdw2f) ([raw file](https://github.com/gramineproject/gramine/blob/2d44e0bf3679d0ba7adb1ca62e7cac272900c3b7/tools/meson.build#L11)):*
> ```Meson
>     hostpalpath_direct = prefix / pkglibdir / 'direct'
> 
>     custom_target('gramine-direct',
> ```

@woju, @eli-schwartz: Any ideas how to do this without resorting to `custom_target`? Seems quite hacky and ugly, and at the same time this looks like a very popular usecase (making `--version` in the resulting binary work).

eli-schwartz:
> FYI: I didn't find a way to mark the output/installed file as executable in Meson, so doing it here manually.

Targets are responsible for producing and fully managing the output file themselves -- meson doesn't run post-build commands to modify built outputs. So if you want the builddir output to be executable, it is correct to chmod it in the script.

Distinct from that, the `install_mode` kwarg is relevant here. The default is to preserve the mode of the built file while copying it into the install tree, but you can manually override that. So it's fully possible to have a non-executable built output but have `meson install` mark the installed version as executable. (It's more frequently used for advanced use cases like having customized permission bits for group/other, as well as of course to set owner/group ID to something other than the default "root".)

In my personal opinion I prefer to chmod it in the build rule, which also allows you to do things like set up a devenv to run the uninstalled version of the software. But both options exist, so you can choose whatever feels better for your workflow.

eli-schwartz:
> @woju, @eli-schwartz: Any ideas how to do this without resorting to `custom_target`? Seems quite hacky and ugly, and at the same time this looks like a very popular usecase (making `--version` in the resulting binary work).

Well, for compiled binaries this isn't usually a problem, you just have `version.h` be a vcs_tag and let the compiler handle this.

For generated scripts it's a bit more interesting. You have two sources of information you'd like to format into the file:
- configure-time data
- build-time data

The current solution seems to be to make both into build-time data, by configure-time creation of an edit script, and build-time formatting of the file via a double pass: one with vcs_tag (that intetnally makes a custom_target) and one with a custom_target.

But you could do it with one pass: make the edit script be processed by vcs_tag, and then have it do all replacements at the same time.

dimakuv:
Superseded by #1632, closing this one.

PR #1271, opened 2023-04-06
[LibOS] Add support for capget and capset system calls

NirjharRoyiitk:
The current implementation only supports getting and setting capabilities for the current thread of the current process.

## Description of the changes 
This PR adds support for capget and capset system calls. However, for now getting and setting
capabilities can be done only for the current thread of the current process. 

## How to test this PR? 

gramine/libos/test/regression
gramine-test --sgx    pytest  -v -k test_123_capget_capset_values_check
gramine-test  pytest  -v -k test_123_capget_capset_values_check
gramine-test --sgx    pytest  -v -k test_122_capget_bytes_check
gramine-test  pytest  -v -k test_122_capget_bytes_check


mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1271)* status: 0 of 23 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @NirjharRoyiitk)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSO-0ydCfZplwucibQE:-NSO-0ydCfZplwucibQF:bsdc352) (no related file):*
This seems to be just a naive passthrough to the host?

This goes against Gramine design and additionally introduces security problems on the SGX backend. Gramine is supposed to emulate Linux kernel in usermode and PAL API should be used just for basic I/O. How this implementation would work if we introduced Windows or FreeBSD PAL?

___

llly (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1271)* status: 0 of 23 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @NirjharRoyiitk)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSP2-kk2aH-jGpOo0eb:-NSP2-kk2aH-jGpOo0ec:b-92husg) (no related file):*
Do you want to enable a specific workload in Gramine?

___

NirjharRoyiitk (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1271)* status: 0 of 23 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @llly)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSP2-kk2aH-jGpOo0eb:-NSPDF54D9DEEJMZWPeo:b7yzv1k) (no related file):*

Do you want to enable a specific workload in Gramine?
</blockquote></details>

We had implemented this system in Fortanix EnclaveOs because we had one of customers/POCs that required free radius server and that required capget.

___

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1271)* status: 0 of 23 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @llly and @NirjharRoyiitk)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSP_NeWDCMU43H1zmIT:-NSP_NeWDCMU43H1zmIU:bujj7pc) (no related file):*
Pls kindly note that this was discussed in one of the previous meetings - see the whole discussions and notes here: https://github.com/gramineproject/gramine/discussions/1072.

I met w/ workloads using these capabilities-related syscalls but dummy implementations work fine in those cases.

___

NirjharRoyiitk (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1271)* status: 0 of 23 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin and @llly)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSP_NeWDCMU43H1zmIT:-NSQlnLSDatIuC6aCuVG:b74nepz) (no related file):*

Pls kindly note that this was discussed in one of the previous meetings - see the whole discussions and notes here: https://github.com/gramineproject/gramine/discussions/1072.

I met w/ workloads using these capabilities-related syscalls but dummy implementations work fine in those cases.


</blockquote></details>

Okay I will go through the thread. Do you want me to update the PR with a dummy implementation and remove the host syscalls?

___

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1271)* status: 0 of 23 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin, @llly, and @NirjharRoyiitk)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSP_NeWDCMU43H1zmIT:-NSTOoMCCIv0eJrPxoyJ:b-6yand) (no related file):*
<details><summary><i>Previously, NirjharRoyiitk wrote…</i></summary><blockquote>

Okay I will go through the thread. Do you want me to update the PR with a dummy implementation and remove the host syscalls?
</blockquote></details>

Depends what makes sense, but definitely a passthrough doesn't :) Would be easier to say if you said why this workload actually needs to manipulate capabilities, they are usually for things which won't work securely in an enclave anyways.

___

NirjharRoyiitk (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1271)* status: 0 of 23 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin and @llly)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSP_NeWDCMU43H1zmIT:-NSdn9Fr01MwXP9Lbcdq:b-n3oif1) (no related file):*

Depends what makes sense, but definitely a passthrough doesn't :) Would be easier to say if you said why this workload actually needs to manipulate capabilities, they are usually for things which won't work securely in an enclave anyways.
</blockquote></details>

So my question now: What approach do we usually take in Gramine if an application requires a functionality that can't be securely implemented inside enclaves? Do we have the insecure implementation (and try to make it as secure as possible) or do we simply reject running the application? 

Regarding what the workload does with capablities: I haven't gone through the source code but instead went through the strace logs and the application simply crashed and stopped when capget returned -ENOSYS. 

___

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1271)* status: 0 of 23 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin, @llly, and @NirjharRoyiitk)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSP_NeWDCMU43H1zmIT:-NSfT3WX7mT9ZRmwKGFt:b-94hnyv) (no related file):*
<details><summary><i>Previously, NirjharRoyiitk wrote…</i></summary><blockquote>

So my question now: What approach do we usually take in Gramine if an application requires a functionality that can't be securely implemented inside enclaves? Do we have the insecure implementation (and try to make it as secure as possible) or do we simply reject running the application? 

Regarding what the workload does with capablities: I haven't gone through the source code but instead went through the strace logs and the application simply crashed and stopped when capget returned -ENOSYS. 
</blockquote></details>

That's a good question :)
This is really case-by-case, but the first and the most important step is examining the real workload which is trying to use this functionality. This way you usually get into one of two cases:
a) The app is doing something which is impossible to do securely in SGX / under Gramine (e.g. tries to run untrusted code inside a sandbox or use shared memory).
b) The app is calling this Linux API, but doesn't really care about it, i.e. it doesn't matter for its correctness. Examples include performance optimizations like `madvise()` calls or using e.g. seccomp as an additional later of anti-exploit mitigations.

What's important in this distinction is the purpose of using these APIs in particular apps, not the APIs itself, because seccomp can fall in both categories depending on the app design.

b) is a simpler case - if the API is not semantically visible to the app (like almost all `madvise` sub-calls) then you can just implement a dummy version of it. If its effects are visible to the app, then a possible solution would be a manifest switch which switches on and off the dummy version (this is because we prefer to fail loudly if an app uses an unsupported API than corrupting its state, which could happen with different apps using this API and actually relying on its behavior).

a) is hard, and usually the only way to go forward is to change the app to not do that. But in specific cases it may be possible to create some workarounds.

___

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1271)* status: 0 of 23 files reviewed, 8 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin, @llly, and @NirjharRoyiitk)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSP_NeWDCMU43H1zmIT:-NSkMLcv1B8F5cdu0Fc3:b882d1e) (no related file):*

That's a good question :)
This is really case-by-case, but the first and the most important step is examining the real workload which is trying to use this functionality. This way you usually get into one of two cases:
a) The app is doing something which is impossible to do securely in SGX / under Gramine (e.g. tries to run untrusted code inside a sandbox or use shared memory).
b) The app is calling this Linux API, but doesn't really care about it, i.e. it doesn't matter for its correctness. Examples include performance optimizations like `madvise()` calls or using e.g. seccomp as an additional later of anti-exploit mitigations.

What's important in this distinction is the purpose of using these APIs in particular apps, not the APIs itself, because seccomp can fall in both categories depending on the app design.

b) is a simpler case - if the API is not semantically visible to the app (like almost all `madvise` sub-calls) then you can just implement a dummy version of it. If its effects are visible to the app, then a possible solution would be a manifest switch which switches on and off the dummy version (this is because we prefer to fail loudly if an app uses an unsupported API than corrupting its state, which could happen with different apps using this API and actually relying on its behavior).

a) is hard, and usually the only way to go forward is to change the app to not do that. But in specific cases it may be possible to create some workarounds.
</blockquote></details>

I second everything that @mkow wrote.

Also, I briefly looked at the code and I find it of good quality. Some comments if/when you re-implement this PR to get rid of new OCALLs (i.e., dummy implementations inside Gramine LibOS layer):
- LibOS and tests code looks normal, though I would add more empty lines between logical code blocks -- it's hard to read when all lines are lumped together. Plus see my comments in those files.
- PAL code must be removed completely -- we don't want to do a passthrough, as discussed in other comments.
- Please check LTP tests that test capget/capset and enable those that can be enabled (and explain in comments why the other tests cannot be enabled).

___
___
*[`common/include/linux_capabilities.h` line 11 at r1](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSkN7S0CK-r_yMmn538:-NSkN7S0CK-r_yMmn539:bbgunzk) ([raw file](https://github.com/gramineproject/gramine/blob/0fdff36d180204490204d0581312a6d0ae8368ff/common/include/linux_capabilities.h#L11)):*
> ```objc
> #define GRAMINE_LINUX_CAPABILITY_VERSION_1  0x19980330
> #define GRAMINE_LINUX_CAPABILITY_VERSION_2  0x20071026
> #define GRAMINE_LINUX_CAPABILITY_VERSION_3  0x20080522
> ```

These macros need some explanatory comment. Preferably also with a link to Elixir website where these macros are defined.
___
*[`libos/include/libos_table.h` line 15 at r1](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSkNK9QBB6O5HSwCqR8:-NSkNK9R1OWPZTK55sMJ:b-6r3is2) ([raw file](https://github.com/gramineproject/gramine/blob/0fdff36d180204490204d0581312a6d0ae8368ff/libos/include/libos_table.h#L15)):*
> ```objc
> 
> #include "libos_types.h"
> #include <linux/capability.h>
> ```

This should have been `linux_capabilities.h`?
___
*[`libos/src/bookkeep/libos_thread.c` line 154 at r1](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSkNPY-DrMBvHMYtRah:-NSkNPY-DrMBvHMYtRai:blofw47) ([raw file](https://github.com/gramineproject/gramine/blob/0fdff36d180204490204d0581312a6d0ae8368ff/libos/src/bookkeep/libos_thread.c#L154)):*
> ```C
> 
>     int64_t uid_int64;
>     ret = toml_int_in(g_manifest_root, "loader.uid", /*defaultval=*/1, &uid_int64);
> ```

Why? This is clearly wrong to silently change the default value in this code.
___
*[`libos/src/bookkeep/libos_thread.c` line 162 at r1](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSkNWFbBNcUpogkUmz0:-NSkNWFbBNcUpogkUmz1:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/0fdff36d180204490204d0581312a6d0ae8368ff/libos/src/bookkeep/libos_thread.c#L162)):*
> ```C
> 
>     int64_t gid_int64;
>     ret = toml_int_in(g_manifest_root, "loader.gid", /*defaultval=*/1, &gid_int64);
> ```

ditto
___
*[`libos/src/sys/libos_access.c` line 10 at r1](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSkNXMHB3kImU8kgcZ7:-NSkNXMHB3kImU8kgcZ8:b4c769h) ([raw file](https://github.com/gramineproject/gramine/blob/0fdff36d180204490204d0581312a6d0ae8368ff/libos/src/sys/libos_access.c#L10)):*
> ```C
> #include <errno.h>
> #include <linux/fcntl.h>
> #include <linux/capability.h>
> ```

Why do you have two headers? You should copy-paste all relevant structs/macros from the Linux header file into your new `linux_capabilities.h` file.

NirjharRoyiitk (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1271)* status: 0 of 23 files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @kailun-qin, and @llly)

___
*[`libos/src/sys/libos_access.c` line 10 at r1](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSkNXMHB3kImU8kgcZ7:-NTDu5VZD3S7W6g7ARXW:blj4a4j) ([raw file](https://github.com/gramineproject/gramine/blob/0fdff36d180204490204d0581312a6d0ae8368ff/libos/src/sys/libos_access.c#L10)):*

Why do you have two headers? You should copy-paste all relevant structs/macros from the Linux header file into your new `linux_capabilities.h` file.
</blockquote></details>

removed this file. Not required with the new changes
___
*[`common/include/linux_capabilities.h` line 11 at r1](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSkN7S0CK-r_yMmn538:-NTDu1G1BaJEaWOpkhQ2:blj4a4j) ([raw file](https://github.com/gramineproject/gramine/blob/0fdff36d180204490204d0581312a6d0ae8368ff/common/include/linux_capabilities.h#L11)):*

These macros need some explanatory comment. Preferably also with a link to Elixir website where these macros are defined.
</blockquote></details>

removed this file. Not required with the new changes

NirjharRoyiitk (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1271)* status: 0 of 23 files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @kailun-qin, and @llly)

___
*[`libos/include/libos_table.h` line 15 at r1](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSkNK9QBB6O5HSwCqR8:-NTDuABC98tilcwyw328:bvlav3d) ([raw file](https://github.com/gramineproject/gramine/blob/0fdff36d180204490204d0581312a6d0ae8368ff/libos/include/libos_table.h#L15)):*

This should have been `linux_capabilities.h`?
</blockquote></details>

removed this file linux_capabilities.h. Not required with my new changes

NirjharRoyiitk (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1271)* status: 0 of 23 files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @kailun-qin, and @llly)

___
*[`libos/src/bookkeep/libos_thread.c` line 154 at r1](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSkNPY-DrMBvHMYtRah:-NTDuIdqCUYgKKQNpOkW:b-vzn3b4) ([raw file](https://github.com/gramineproject/gramine/blob/0fdff36d180204490204d0581312a6d0ae8368ff/libos/src/bookkeep/libos_thread.c#L154)):*

Why? This is clearly wrong to silently change the default value in this code.
</blockquote></details>

If the default value is 0, it kind of means that the application is running as root even if the application isn't running with root privileges. Shouldn't we have the default value to be some non-zero value like 1? One of the tests that I wrote in this PR was failing because Gramine was incorrectly reporting the euid to be 0 (root).

mkow (changes requested):
> Reviewed 2 of 23 files at r1, 14 of 16 files at r2.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1271)* status: 16 of 23 files reviewed, 22 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @kailun-qin, @llly, and @NirjharRoyiitk)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSO-0ydCfZplwucibQE:-NTGaQ036u3_A0H2ubPg:b-rx4pxc) (no related file):*

This seems to be just a naive passthrough to the host?

This goes against Gramine design and additionally introduces security problems on the SGX backend. Gramine is supposed to emulate Linux kernel in usermode and PAL API should be used just for basic I/O. How this implementation would work if we introduced Windows or FreeBSD PAL?
</blockquote></details>

Seems you changes it, but now it always ignores `capset()`, which IMO shouldn't be the default, because apps may use it for sandboxing (to drop privileges), thinking that capabilities actually work in Gramine. If we want such implementation then it should be controlled by a non-default manifest switch with good documentation about possible security problems with enabling it.

___
___
*[`libos/include/libos_table.h` line 15 at r1](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSkNK9QBB6O5HSwCqR8:-NTGUmKP76fDkVRc2o7i:b-bu3vd7) ([raw file](https://github.com/gramineproject/gramine/blob/0fdff36d180204490204d0581312a6d0ae8368ff/libos/include/libos_table.h#L15)):*
<details><summary><i>Previously, NirjharRoyiitk wrote…</i></summary><blockquote>

removed this file linux_capabilities.h. Not required with my new changes
</blockquote></details>

See the other discussion.
___
*[`libos/include/libos_table.h` line 210 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTGUrpn9cqksL4k4NNZ:-NTGUrpn9cqksL4k4NN_:b-w7dkpm) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/libos/include/libos_table.h#L210)):*
> ```objc
> long libos_syscall_mlock2(unsigned long start, size_t len, int flags);
> long libos_syscall_sysinfo(struct sysinfo* info);
> long libos_syscall_capget(cap_user_header_t hdrp, const cap_user_data_t datap);
> ```

`const` here is meaningless.



_Code quote:_
```objc
const cap_user_data_t datap
```
___
*[`libos/include/libos_table.h` line 211 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTGUtkh4LmkMdRFEuJY:-NTGUtki-dhgBUYWN3Js:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/libos/include/libos_table.h#L211)):*
> ```objc
> long libos_syscall_sysinfo(struct sysinfo* info);
> long libos_syscall_capget(cap_user_header_t hdrp, const cap_user_data_t datap);
> long libos_syscall_capset(cap_user_header_t hdrp, const cap_user_data_t datap);
> ```

ditto
___
*[`libos/include/libos_thread.h` line 139 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTGWTyH-4X5HRXuvdYm:-NTGWTyH-4X5HRXuvdYn:b20cn6e) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/libos/include/libos_thread.h#L139)):*
> ```objc
>     refcount_t ref_count;
>     struct libos_lock lock;
>     struct __user_cap_data_struct capabilities[2];
> ```

Why is this an array, not a single struct? What are the indices? This code is quite confusing if you don't know upfront what it's doing.
___
*[`libos/src/bookkeep/libos_thread.c` line 154 at r1](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSkNPY-DrMBvHMYtRah:-NTGTBtI6oe3CYSpBu4H:b-61ah1w) ([raw file](https://github.com/gramineproject/gramine/blob/0fdff36d180204490204d0581312a6d0ae8368ff/libos/src/bookkeep/libos_thread.c#L154)):*
> even if the application isn't running with root privileges

But it _is_?
___
*[`libos/src/bookkeep/libos_thread.c` line 125 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTGVln6-pWF51flTmIQ:-NTGVln6-pWF51flTmIR:bwhjb2l) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/libos/src/bookkeep/libos_thread.c#L125)):*
> ```C
> }
> 
> static void set_cur_thread_capabilities(struct libos_thread* cur_thread) {
> ```

Usually when you prefix a function with `set_` the function takes an object and the input to set/put into the struct. This one is breaking this convention, I think it should be `init_cur_thread_capabilities`?



_Code quote:_
```C
set_cur_thread_capabilities(struct libos_thread* cur_thread)
```
___
*[`libos/src/bookkeep/libos_thread.c` line 133 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTGWDHn-2YJqaTseciY:-NTGWDHn-2YJqaTseciZ:b17ukcj) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/libos/src/bookkeep/libos_thread.c#L133)):*
> ```C
> 
>     cur_thread->capabilities[0].permitted = is_root ? 0xffffffff : 0;
>     cur_thread->capabilities[1].permitted = is_root ? 0x1ff : 0;
> ```

What are these magic constants?
___
*[`libos/src/sys/libos_access.c` line 10 at r1](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSkNXMHB3kImU8kgcZ7:-NTGTdv2CF1bzbKTIwYY:b-mt4kgy) ([raw file](https://github.com/gramineproject/gramine/blob/0fdff36d180204490204d0581312a6d0ae8368ff/libos/src/sys/libos_access.c#L10)):*
<details><summary><i>Previously, NirjharRoyiitk wrote…</i></summary><blockquote>

removed this file. Not required with the new changes
</blockquote></details>

But we shouldn't take the headers from the OS _where Gramine is built_, we should use headers which match the Linux version Gramine is emulating. So, they should be rewritten to Gramine.

btw. there's a pending effort to clean this up, see https://github.com/gramineproject/gramine/pull/1275.
___
*[`libos/src/sys/libos_access.c` line 54 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTGWsg09dMlYoAnjXu9:-NTGWsg09dMlYoAnjXuA:b-168cwg) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/libos/src/sys/libos_access.c#L54)):*
> ```C
> }
> 
> long libos_syscall_capget(cap_user_header_t _hdrp, const cap_user_data_t _datap) {
> ```

Why these underscores? I don't think we're using any such convention in Gramine anywhere.



_Code quote:_
```C
_hdrp
```
___
*[`libos/src/sys/libos_access.c` line 84 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTGYHys5NGdjANtnYl1:-NTGYHytCFnm1Nbp2tV0:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/libos/src/sys/libos_access.c#L84)):*
> ```C
>             else {
>                 memcpy(_hdrp, &hdrp, sizeof(*_hdrp));
>                 ret = _datap == NULL ? 0 : -EINVAL;
> ```





_Suggestion:_
```C
            if (!_hdrp || !is_user_memory_writable(_hdrp, sizeof(hdrp))) {
                ret = -EFAULT;
            } else {
                memcpy(_hdrp, &hdrp, sizeof(*_hdrp));
                ret = _datap == NULL ? 0 : -EINVAL;
            }
```
___
*[`libos/src/sys/libos_access.c` line 92 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTGYiYz1KTc-MJyNyeh:-NTGYiYz1KTc-MJyNyei:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/libos/src/sys/libos_access.c#L92)):*
> ```C
>         goto out;
>     }
>     /* For now we can get and set capabilties for current thread.
> ```





_Suggestion:_
```C
capabilities
```
___
*[`libos/src/sys/libos_access.c` line 142 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTG_U-O5zZYo1DXHHGC:-NTG_U-O5zZYo1DXHHGD:b-9es1do) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/libos/src/sys/libos_access.c#L142)):*
> ```C
>             return -EINVAL;
>     }
>     if (!_datap || !is_user_memory_readable(_datap, size * sizeof(datap[0]))) {
> ```

What's wrong with `_datap == 0`?



_Code quote:_
```C
!_datap
```
___
*[`pal/src/pal_misc.c` line 49 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTGTTcO0ncK6aYJQEyU:-NTGTTcO0ncK6aYJQEyV:b3kre) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/pal/src/pal_misc.c#L49)):*
> ```C
> int PalGetSpecialKey(const char* name, void* key, size_t* key_size) {
>     return _PalGetSpecialKey(name, key, key_size);
> }
> ```

?
___
*[`pal/src/pal_symbols` line 54 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTGTUpX4p-RNir6Bh9Z:-NTGTUpX4p-RNir6Bh9_:b3kre) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/pal/src/pal_symbols#L54)):*
> ```
> PalGetSpecialKey
> PalDebugLog
> PalGetPalPublicState
> ```

?
___
*[`pal/src/host/linux/pal_misc.c` line 10 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTG_zw6-xJdV9YKN13f:-NTG_zw6-xJdV9YKN13g:b3kre) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/pal/src/host/linux/pal_misc.c#L10)):*
> ```C
> #include <asm/fcntl.h>
> #include <linux/time.h>
> #include <sys/capability.h>
> ```

?
___
*[`pal/src/host/linux-sgx/host_ocalls.c` line 11 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTGa0Lk1HqHWnzqgA70:-NTGa0Lk1HqHWnzqgA71:b-cncphx) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/pal/src/host/linux-sgx/host_ocalls.c#L11)):*
> ```C
> #include <asm/poll.h>
> #include <limits.h>
> #include <linux/capability.h>
> ```

?
What are all these random changes in unrelated files?
___
*[`pal/src/host/linux-sgx/pal_ocall_types.h` line 363 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTGa9Lm72AzmJyH45dr:-NTGa9Lm72AzmJyH45ds:b3kre) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/pal/src/host/linux-sgx/pal_ocall_types.h#L363)):*
> ```objc
>     uint32_t version;
>     void *cap_data;
> };
> ```

?

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1271)* status: 16 of 23 files reviewed, 23 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @llly, and @NirjharRoyiitk)

___
*[`libos/test/regression/capget_capset_values.c` line 1 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTHJ_9s0trn9dw2HchJ:-NTHJ_9s0trn9dw2HchK:bxkhrit) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/libos/test/regression/capget_capset_values.c#L1)):*
> ```C
> ```

I assume some of the tests have already been covered by LTP? Then why don't we check and enable the related capget/capset LTP tests (and reduce the potential overlaps here) then?

NirjharRoyiitk (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1271)* status: 16 of 23 files reviewed, 23 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @kailun-qin, @llly, and @mkow)

___
*[`libos/include/libos_thread.h` line 139 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTGWTyH-4X5HRXuvdYm:-NTIgNpeA9Me0UTFMGOQ:bcazm65) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/libos/include/libos_thread.h#L139)):*

Why is this an array, not a single struct? What are the indices? This code is quite confusing if you don't know upfront what it's doing.
</blockquote></details>

https://man7.org/linux/man-pages/man2/capset.2.html
THis page says the following:

Note that 64-bit capabilities use datap[0] and datap[1], whereas
       32-bit capabilities use only datap[0].
That is reason this is an array with 2 elements
___
*[`libos/src/bookkeep/libos_thread.c` line 154 at r1](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSkNPY-DrMBvHMYtRah:-NTIglbKBz3CIw-tvU-N:bwcqnsc) ([raw file](https://github.com/gramineproject/gramine/blob/0fdff36d180204490204d0581312a6d0ae8368ff/libos/src/bookkeep/libos_thread.c#L154)):*

> even if the application isn't running with root privileges

But it _is_?
</blockquote></details>

So you mean apps in Gramine have root priv by default? When I ran the regression test with capset by making it passthrough, the host OS call failed with EPERM. If it is running as root, then it shouldn't fail with EPERM isn't it?
___
*[`libos/src/bookkeep/libos_thread.c` line 133 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTGWDHn-2YJqaTseciY:-NTIgZN79avoJSUFHJnd:bots2mh) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/libos/src/bookkeep/libos_thread.c#L133)):*

What are these magic constants?
</blockquote></details>

These are the default dummy values taken from the host behavior of the application. I ran it outside Gramine and got the values and used here.
___
*[`libos/src/sys/libos_access.c` line 10 at r1](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSkNXMHB3kImU8kgcZ7:-NTIhVbsBjy9kvW-fFTS:brsa0ew) ([raw file](https://github.com/gramineproject/gramine/blob/0fdff36d180204490204d0581312a6d0ae8368ff/libos/src/sys/libos_access.c#L10)):*

But we shouldn't take the headers from the OS _where Gramine is built_, we should use headers which match the Linux version Gramine is emulating. So, they should be rewritten to Gramine.

btw. there's a pending effort to clean this up, see https://github.com/gramineproject/gramine/pull/1275.
</blockquote></details>

okay but then it should be applicable for all the header files included right? I see other files being included like <linux/fcntl.h>
___
*[`libos/src/sys/libos_access.c` line 54 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTGWsg09dMlYoAnjXu9:-NTIhj6w79JrXZTTOeqJ:bbithir) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/libos/src/sys/libos_access.c#L54)):*

Why these underscores? I don't think we're using any such convention in Gramine anywhere.
</blockquote></details>

usually variables beginning with _ denote that they shouldn't be used directly and are kind of internal. THat is the reason I used it. 
___
*[`libos/src/sys/libos_access.c` line 142 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTG_U-O5zZYo1DXHHGC:-NTIiLnD9eOeIVNC-rfj:b-1l91hp) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/libos/src/sys/libos_access.c#L142)):*

What's wrong with `_datap == 0`?
</blockquote></details>

Didn;t get this. Do you mean why have I used !_datap instead of _datap == 0 ?

NirjharRoyiitk (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1271)* status: 16 of 23 files reviewed, 23 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @kailun-qin, @llly, and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSO-0ydCfZplwucibQE:-NTJH_neA-L5qk6YxDDv:bjazci8) (no related file):*

Seems you changes it, but now it always ignores `capset()`, which IMO shouldn't be the default, because apps may use it for sandboxing (to drop privileges), thinking that capabilities actually work in Gramine. If we want such implementation then it should be controlled by a non-default manifest switch with good documentation about possible security problems with enabling it.
</blockquote></details>

So do you mean the following:
We should have some kind of boolean parameter in the manifest file (like ```is_capability_set_allowed```) and we should check if this boolean is true. If this is set to true, then we will allow the passthrough (and maybe print some warning logs), else we will only restrict the flow of the system call within Gramine. Is that right?

___

NirjharRoyiitk (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1271)* status: 16 of 23 files reviewed, 23 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @kailun-qin, @llly, and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSO-0ydCfZplwucibQE:-NTNWu2d4C7KhIzrwMZ7:b-l4hb8m) (no related file):*

> We should have some kind of boolean parameter in the manifest file

Yes.

> If this is set to true, then we will allow the passthrough

No. How would this work on Windows or FreeBSD hosts? Think of Gramine as of a virtual machine running it's own OS and apps inside.

> else we will only restrict the flow of the system call within Gramine

No. Else we will fail, like we do now (and this should be the default, because capabilities are not implemented inside Gramine).
</blockquote></details>

Okay so summarizing the discussion so far:
1. Have a boolean (is_capability_set_allowed) in a non-default manifest file
2. If is_capability_set_allowed is true --> Then allow the the capset to work the way it is working now (fake it just by changing the permissions only inside Gramine but NOT  passthrough)
3. If is_capability_set_allowed is false  --> fail with -ENOSYS

Is that right?

___

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1271)* status: 16 of 23 files reviewed, 22 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin, @llly, @mkow, and @NirjharRoyiitk)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSO-0ydCfZplwucibQE:-NTNe2XcE8KhvEXbtznT:bc5n75j) (no related file):*
<details><summary><i>Previously, NirjharRoyiitk wrote…</i></summary><blockquote>

Okay so summarizing the discussion so far:
1. Have a boolean (is_capability_set_allowed) in a non-default manifest file
2. If is_capability_set_allowed is true --> Then allow the the capset to work the way it is working now (fake it just by changing the permissions only inside Gramine but NOT  passthrough)
3. If is_capability_set_allowed is false  --> fail with -ENOSYS

Is that right?
</blockquote></details>

@NirjharRoyiitk Your summary is correct to me. You can proceed with this, and I think this is also exactly what @mkow meant.

However, I don't like the name `is_capability_set_allowed`. I suggest to name the manifest option `sys.enable_capabilities = [true|false]`. And in the description of this new manifest option, mention that it is NOT a passthrough but instead it's emulated entirely inside Gramine.

___
___
*[`libos/include/libos_thread.h` line 139 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTGWTyH-4X5HRXuvdYm:-NTNep2N2K_LxwwKnQb1:bibpo0j) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/libos/include/libos_thread.h#L139)):*

But you can't assume that someone reading this code will open a man page for each field in this struct when reading it. From looking at the code alone you have no idea why this is an array and what's the difference between `capabilities[0]` and `capabilities[1]`, it seems very counter-intuitive that this is a 2-element array.
</blockquote></details>

Yes, I agree with Michal. Inside of Gramine, we can declare whatever data structs we want with whatever names. Linux kernel has notoriously bad/convoluted names and data types, and Gramine shouldn't blindly follow Linux in this regard.
___
*[`libos/src/bookkeep/libos_thread.c` line 133 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTGWDHn-2YJqaTseciY:-NTNf8Ae61ZlND_RgiPY:b4xb9as) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/libos/src/bookkeep/libos_thread.c#L133)):*
<details><summary><i>Previously, NirjharRoyiitk wrote…</i></summary><blockquote>

These are the default dummy values taken from the host behavior of the application. I ran it outside Gramine and got the values and used here.
</blockquote></details>

Could you find these constants in the Linux kernel source code, and add a comment with a link to these constants here? Otherwise it's truly magic.
___
*[`libos/src/sys/libos_access.c` line 10 at r1](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSkNXMHB3kImU8kgcZ7:-NTNfPEd3kd_jEvgzHTG:b-fjvh81) ([raw file](https://github.com/gramineproject/gramine/blob/0fdff36d180204490204d0581312a6d0ae8368ff/libos/src/sys/libos_access.c#L10)):*

Yes, it should and that's the work I'm doing right now (because old parts of the codebase are not doing this yet).
</blockquote></details>

Yes, as @mkow said, the fact that we're including host-OS files is totally wrong (other than some common and generic headers like `stdint.h`). This is a legacy of Graphene... Michal is removing these wrong includes currently.
___
*[`libos/src/sys/libos_access.c` line 54 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTGWsg09dMlYoAnjXu9:-NTNfevx3o0QkdaYKTm3:b3fxuro) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/libos/src/sys/libos_access.c#L54)):*

But we don't use such convention anywhere in Gramine codebase. Just suffix it with `_unaligned`, this way it will be clear why you do this copy.
</blockquote></details>

+1 to Michal. In our codebase, we tend to use longer names, but with very explicit meanings.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1271)* status: 16 of 23 files reviewed, 22 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @kailun-qin, @llly, and @NirjharRoyiitk)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSO-0ydCfZplwucibQE:-NTOBE4s95LZPJ6VnDUZ:b-sudff1) (no related file):*

@NirjharRoyiitk Your summary is correct to me. You can proceed with this, and I think this is also exactly what @mkow meant.

However, I don't like the name `is_capability_set_allowed`. I suggest to name the manifest option `sys.enable_capabilities = [true|false]`. And in the description of this new manifest option, mention that it is NOT a passthrough but instead it's emulated entirely inside Gramine.
</blockquote></details>

Yes, and I also don't like the proposed name. `is_capability_set_allowed` sounds like it was actually implemented in Gramine but for some reason disallowed. I'd go for something like `sys.enable_dummy_capabilities = [true|false]`.

___

NirjharRoyiitk (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1271)* status: 16 of 23 files reviewed, 22 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @kailun-qin, @llly, and @mkow)

___
*[`libos/include/libos_thread.h` line 139 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTGWTyH-4X5HRXuvdYm:-NTOLCsKE4y_OGYmjzfE:bmdbg6g) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/libos/include/libos_thread.h#L139)):*

Yes, I agree with Michal. Inside of Gramine, we can declare whatever data structs we want with whatever names. Linux kernel has notoriously bad/convoluted names and data types, and Gramine shouldn't blindly follow Linux in this regard.
</blockquote></details>

Okay I will have a comment in the code explaining these. Regarding naming, I can think of an alternative. Instead of having an array of size 2, I can do the following:


```struct gramine_user_cap_data_32 {
    uint32_t effective;
    uint32_t permitted;
    uint32_t inheritable;
}; 

struct gramine_user_cap_data_64 {
    struct gramine_user_cap_data_32 lower_bytes;
    struct gramine_user_cap_data_32 upper_bytes;
};
```
and then I can use ```struct gramine_user_cap_data_64``` 
How does the above look? Do you find it better than the using 1 array of size 2?

NirjharRoyiitk (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1271)* status: 16 of 23 files reviewed, 22 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @kailun-qin, @llly, and @mkow)

___
*[`libos/include/libos_thread.h` line 139 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTGWTyH-4X5HRXuvdYm:-NTOMBDE4pYKHEWDboi5:b-zffg4a) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/libos/include/libos_thread.h#L139)):*
<details><summary><i>Previously, NirjharRoyiitk wrote…</i></summary><blockquote>

Okay I will have a comment in the code explaining these. Regarding naming, I can think of an alternative. Instead of having an array of size 2, I can do the following:


```struct gramine_user_cap_data_32 {
    uint32_t effective;
    uint32_t permitted;
    uint32_t inheritable;
}; 

struct gramine_user_cap_data_64 {
    struct gramine_user_cap_data_32 lower_bytes;
    struct gramine_user_cap_data_32 upper_bytes;
};
```
and then I can use ```struct gramine_user_cap_data_64``` 
How does the above look? Do you find it better than the using 1 array of size 2?


</blockquote></details>

Sorry the first structure will look like 

```
struct gramine_user_cap_data_32 {
    uint32_t effective;
    uint32_t permitted;
    uint32_t inheritable;
}; 
```

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1271)* status: 16 of 23 files reviewed, 22 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin, @llly, @mkow, and @NirjharRoyiitk)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSO-0ydCfZplwucibQE:-NTOQQ3H-_xew1VLiJiV:b-wezfvf) (no related file):*

Yes, and I also don't like the proposed name. `is_capability_set_allowed` sounds like it was actually implemented in Gramine but for some reason disallowed. I'd go for something like `sys.enable_dummy_capabilities = [true|false]`.
</blockquote></details>

`sys.enable_dummy_capabilities` sounds good to me.

___
___
*[`libos/include/libos_thread.h` line 139 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTGWTyH-4X5HRXuvdYm:-NTOQXXq3m_B6DbIxcl8:bq7giss) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/libos/include/libos_thread.h#L139)):*
<details><summary><i>Previously, NirjharRoyiitk wrote…</i></summary><blockquote>

Sorry the first structure will look like 

```
struct gramine_user_cap_data_32 {
    uint32_t effective;
    uint32_t permitted;
    uint32_t inheritable;
}; 
```
</blockquote></details>

What's the point of having this two-32-bit-ints-as-one-64-bit ?

Why not just:
```
struct gramine_user_cap_data {
    uint32_t effective;
    uint32_t permitted;
    uint32_t inheritable;
};
```

Or do you need the Linux ABI (the syscall argument coming from application/glibc)? In that case yes, you'll need to conform to this Linux ABI... Well, I won't mind your proposed solution then, but please add the comment that this weird format is to comply with Linux ABI, plus add a link to the Elixir web site with this struct definition.

NirjharRoyiitk (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1271)* status: 16 of 23 files reviewed, 22 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @kailun-qin, @llly, and @mkow)

___
*[`libos/include/libos_thread.h` line 139 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTGWTyH-4X5HRXuvdYm:-NTOa0A_6lvXOLS_dI4L:bq1xpkk) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/libos/include/libos_thread.h#L139)):*

What's the point of having this two-32-bit-ints-as-one-64-bit ?

Why not just:
```
struct gramine_user_cap_data {
    uint32_t effective;
    uint32_t permitted;
    uint32_t inheritable;
};
```

Or do you need the Linux ABI (the syscall argument coming from application/glibc)? In that case yes, you'll need to conform to this Linux ABI... Well, I won't mind your proposed solution then, but please add the comment that this weird format is to comply with Linux ABI, plus add a link to the Elixir web site with this struct definition.
</blockquote></details>

I think I should stick to the array style. The system call abi(capget/capset) specifies that it should be a pointer to a structure like ```struct gramine_user_cap_data``` . This pointer can either be pointer to one structure variable or a an array of 2 (2 32 bits) depending on the header version. I looked into the source code and it has the same definition. https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/capability.h#L44

NirjharRoyiitk (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1271)* status: 5 of 25 files reviewed, 23 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @kailun-qin, @llly, and @mkow)

___
*[`common/include/linux_capabilities.h` line 30 at r3](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTObsqt8lmbiHJ0-AcT:-NTObsqt8lmbiHJ0-AcU:bbdvs66) ([raw file](https://github.com/gramineproject/gramine/blob/0ea8bf68ee2b9376c6beee8317a96c0c2f8ba659/common/include/linux_capabilities.h#L30)):*
> ```objc
>     uint32_t permitted;
>     uint32_t inheritable;
> };
> ```

todo: Add one new line
___
*[`libos/include/libos_table.h` line 15 at r1](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSkNK9QBB6O5HSwCqR8:-NTObwWn0Gz77do4ZEDZ:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/0fdff36d180204490204d0581312a6d0ae8368ff/libos/include/libos_table.h#L15)):*

See the other discussion.
</blockquote></details>

Done.
___
*[`libos/include/libos_table.h` line 210 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTGUrpn9cqksL4k4NNZ:-NTObxFhAf2VRepzld8r:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/libos/include/libos_table.h#L210)):*

`const` here is meaningless.
</blockquote></details>

Done.
___
*[`libos/include/libos_table.h` line 211 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTGUtkh4LmkMdRFEuJY:-NTObyOJDmw65cBBtp_c:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/libos/include/libos_table.h#L211)):*

ditto
</blockquote></details>

Done.
___
*[`libos/src/bookkeep/libos_thread.c` line 154 at r1](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSkNPY-DrMBvHMYtRah:-NTOcWEw6zfREu8h8-Bo:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/0fdff36d180204490204d0581312a6d0ae8368ff/libos/src/bookkeep/libos_thread.c#L154)):*

> So you mean apps in Gramine have root priv by default?

Yes.

> When I ran the regression test with capset by making it passthrough

But such change didn't make any sense, you broke Gramine design with it and such results can tell you nothing.
</blockquote></details>

Done.
___
*[`libos/src/bookkeep/libos_thread.c` line 162 at r1](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSkNWFbBNcUpogkUmz0:-NTOcX5m1DxrMpkjjV0n:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/0fdff36d180204490204d0581312a6d0ae8368ff/libos/src/bookkeep/libos_thread.c#L162)):*

ditto
</blockquote></details>

Done.
___
*[`libos/src/bookkeep/libos_thread.c` line 125 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTGVln6-pWF51flTmIQ:-NTOcTJ18bJTwwvvzLVC:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/libos/src/bookkeep/libos_thread.c#L125)):*

Usually when you prefix a function with `set_` the function takes an object and the input to set/put into the struct. This one is breaking this convention, I think it should be `init_cur_thread_capabilities`?
</blockquote></details>

Done.
___
*[`libos/src/sys/libos_access.c` line 10 at r1](https://reviewable.io/reviews/gramineproject/gramine/1271#-NSkNXMHB3kImU8kgcZ7:-NTOcYSMEN69fq8bfMQE:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/0fdff36d180204490204d0581312a6d0ae8368ff/libos/src/sys/libos_access.c#L10)):*

Yes, as @mkow said, the fact that we're including host-OS files is totally wrong (other than some common and generic headers like `stdint.h`). This is a legacy of Graphene... Michal is removing these wrong includes currently.
</blockquote></details>

Done.
___
*[`libos/src/sys/libos_access.c` line 54 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTGWsg09dMlYoAnjXu9:-NTOcZPqB9iTXNoIeHjb:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/libos/src/sys/libos_access.c#L54)):*

+1 to Michal. In our codebase, we tend to use longer names, but with very explicit meanings.
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/pal_ocall_types.h` line 363 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTGa9Lm72AzmJyH45dr:-NTOcpWZ9zu-Za0oL0Kv:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/pal/src/host/linux-sgx/pal_ocall_types.h#L363)):*

?
</blockquote></details>

Done.
___
*[`pal/src/pal_misc.c` line 49 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTGTTcO0ncK6aYJQEyU:-NTOcq0vDSVU3IXstWwC:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/pal/src/pal_misc.c#L49)):*

?
</blockquote></details>

Done.
___
*[`pal/src/pal_symbols` line 54 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTGTUpX4p-RNir6Bh9Z:-NTOcquy8gd4RywVd4s8:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/pal/src/pal_symbols#L54)):*

?
</blockquote></details>

Done.
___
*[`pal/src/host/linux/pal_misc.c` line 10 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTG_zw6-xJdV9YKN13f:-NTOcrhl4VJAxTmZqePY:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/pal/src/host/linux/pal_misc.c#L10)):*

?
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/host_ocalls.c` line 11 at r2](https://reviewable.io/reviews/gramineproject/gramine/1271#-NTGa0Lk1HqHWnzqgA70:-NTOcsiG7DMvaV0lRHNA:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/4bb825b227284ac1744acc6c501e8bd3c52e459c/pal/src/host/linux-sgx/host_ocalls.c#L11)):*

?
What are all these random changes in unrelated files?
</blockquote></details>

Done.

Issue #82, opened 2021-09-23
LTP test gettimeofday02 failing intermittently with GSGX


aniket-intelx:
## Description of the problem

The testcase gettimeofday02 is failing intermittently on OOT (kernel : 5.0.0-23-generic) and DCAP (kernel : 5.9.0-intel-next+) with Ubuntu 18.04 machines with SGX.
Log files for both the instances, pass and fail are attached. For convenience purpose, removed duplicate lines `shim_gettimeofday(0x7adac80, 0x0) = 0x0`
[gettimeofday02_TFAIL.txt](https://github.com/gramineproject/gramine/files/7215223/gettimeofday02_TFAIL.txt)
[gettimeofday02_TPASS.txt](https://github.com/gramineproject/gramine/files/7215224/gettimeofday02_TPASS.txt)

## Steps to reproduce
Run below mentioned command
```graphene-sgx gettimeofday02```

## Expected results
TPASS: gettimeofday() monotonous in 10 seconds
## Actual results
TFAIL: Time is going backwards: old 1632371655.582390 vs new 1632371655.582387!

**Comment by dimakuv** [comment]
Why do you say "on OOT and DCAP"? Does it matter? Have you tested on other systems, and there this test never fails at all?

**Comment by jinengandhi-intel** [comment]
Dmitrii, these machines were enabled recently in the CI and we see the issues only on them. On all other 5.12, 5.13 configurations we don't see any issues with this test. Also, to add the test consistently passes with Graphene native and Linux native and the intermittent failures are only seen with GSGX.

**Comment by jinengandhi-intel** [comment]
Let us know if you need access to the machine for a quick check. We can walk through the issue and the workspace.

**Comment by dimakuv** [comment]
@aneessahib You changed the "recalibrate" time PR recently, right?

I think what happens is that your PR triggers the "recalibration" (== don't calculate the time from RDTSC but instead read from the Linux host) more often now, every 50ms (or smth like this). And we **don't have** the logic to verify the read-from-Linux-host timestamp against the previous timestamp. So it is possible periodically that the previous timestamp (calculated via RDTSC) is greater than the current timestamp (read from Linux host), and this LTP test fails.

I suggest we add a code snippet like this:
```
current_timestamp = <from RDTSC or from Linux-host>;
if (current_timestamp < previous_timestamp);
    current_timestamp = previous_timestamp; /* can't go back in time! */
previous_timestamp = current_timestamp;
```

jinengandhi-intel:
[...]

PR #1276, opened 2023-04-11
[PAL/Linux-SGX] Verify OCALL-obtained time against last RDTSC time

kailun-qin:
## Description of the changes 
Previously, Gramine did not verify the read-from-Linux-host timestamp against the last RDTSC time. So it is possible periodically that the previous timestamp (calculated via RDTSC) is greater than the current timestamp (read from Linux host via OCALLs), which can lead to the intermittent failure of LTP test `gettimeofday02` on some configurations with indication that "time is going backwards". This could be even security-relevant since the app running inside the enclave may behave unexpectedly if time teleports back.

This commit adds a sanity adjustment which guarantees no rewinding time is applied when the OCALL-obtained timestamp is "back in time" than the last recorded RDTSC timestamp.

Fixes https://github.com/gramineproject/gramine/issues/82.

## How to test this PR? 
CI + LTP tests on specific configs mentioned in the original issue.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1276)

### DIFF
--- a/pal/src/host/linux-sgx/enclave_ocalls.c
+++ b/pal/src/host/linux-sgx/enclave_ocalls.c
@@ -1789,7 +1789,7 @@ int ocall_gettime(uint64_t* microsec_ptr) {
                 break;
             }
         }
-        *microsec_ptr = microsec;
+        *microsec_ptr = MAX(microsec, expected_microsec);
     }
 
     sgx_reset_ustack(old_ustack);
--- a/pal/src/host/linux-sgx/pal_misc.c
+++ b/pal/src/host/linux-sgx/pal_misc.c
@@ -43,7 +43,6 @@ void init_tsc(void) {
     }
 }
 
-/* TODO: result comes from the untrusted host, introduce some schielding */
 int _PalSystemTimeQuery(uint64_t* out_usec) {
     int ret;
 
@@ -62,6 +61,8 @@ int _PalSystemTimeQuery(uint64_t* out_usec) {
     } while (read_seqretry(&g_tsc_lock, seq));
 
     uint64_t usec = 0;
+    /* Last seen RDTSC-calculated time value. This guards against time rewinding. */
+    static uint64_t last_usec = 0;
     if (start_tsc > 0 && start_usec > 0) {
         /* baseline TSC/usec pair was initialized, can calculate time via RDTSC (but should be
          * careful with integer overflow during calculations) */
@@ -74,13 +75,21 @@ int _PalSystemTimeQuery(uint64_t* out_usec) {
                 usec = start_usec + diff_usec;
                 if (usec < start_usec)
                     return -PAL_ERROR_OVERFLOW;
+
+                /* It's simply `last_usec = max(last_usec, usec)`, but executed atomically. */
+                uint64_t expected_usec = __atomic_load_n(&last_usec, __ATOMIC_ACQUIRE);
+                while (expected_usec < usec) {
+                    if (__atomic_compare_exchange_n(&last_usec, &expected_usec, usec,
+                                                    /*weak=*/true, __ATOMIC_RELEASE,
+                                                    __ATOMIC_ACQUIRE)) {
+                        break;
+                    }
+                }
+
+                *out_usec = MAX(usec, expected_usec);
+                return 0;
             }
         }
-    }
-
-    if (usec) {
-        *out_usec = usec;
-        return 0;
     }
 
     /* if we are here, either the baseline TSC/usec pair was not yet initialized or too much time
@@ -90,6 +99,14 @@ int _PalSystemTimeQuery(uint64_t* out_usec) {
     if (ret < 0)
         return -PAL_ERROR_DENIED;
     uint64_t tsc_cyc2 = get_tsc();
+
+    uint64_t last_recorded_rdtsc = __atomic_load_n(&last_usec, __ATOMIC_ACQUIRE);
+    if (usec < last_recorded_rdtsc) {
+        /* new OCALL-obtained timestamp (`usec`) is "back in time" than the last recorded timestamp
+         * from RDTSC (`last_recorded_rdtsc`); this can happen if the actual host time drifted
+         * backwards compared to the RDTSC time. */
+         usec = last_recorded_rdtsc;
+    }
 
     /* we need to match the OCALL-obtained timestamp (`usec`) with the RDTSC-obtained number of
      * cycles (`tsc_cyc`); since OCALL is a time-consuming operation, we estimate `tsc_cyc` as a


kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1276)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv)

___
*[`pal/src/host/linux-sgx/pal_misc.c` line 104 at r1](https://reviewable.io/reviews/gramineproject/gramine/1276#-NSjITEd508rLtl0LLTy:-NSjYLxJDjNnnee2GpHG:bb6xxxw) ([raw file](https://github.com/gramineproject/gramine/blob/1f6149f9a45b969921890e256feb09bda6152772/pal/src/host/linux-sgx/pal_misc.c#L104)):*
> Why did you put the code under the else branch?

I was indeed hesitating when making this change. But otherwise, we would update `g_start_tsc -> tsc_cyc`but `g_start_usec -> last_usec_rdtsc`?

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1276)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin)

___
*[`pal/src/host/linux-sgx/pal_misc.c` line 104 at r1](https://reviewable.io/reviews/gramineproject/gramine/1276#-NSjITEd508rLtl0LLTy:-NSjdGwg2VSEJ8T81aNS:b9bo4w4) ([raw file](https://github.com/gramineproject/gramine/blob/1f6149f9a45b969921890e256feb09bda6152772/pal/src/host/linux-sgx/pal_misc.c#L104)):*

> Why did you put the code under the else branch?

I was indeed hesitating when making this change. But otherwise, we would update `g_start_tsc -> tsc_cyc`but `g_start_usec -> last_usec_rdtsc`?
</blockquote></details>

Yep, but it doesn't seem to be a problem? In fact, it's what we want I think.

- `g_start_usec` contains the last-baseline-timestamp, kept in microseconds
- `g_start_tsc` contains the last-baseline-timestamp, kept in TSC (cycles)

These two variables are used to calculate the "current timestamp in microseconds" from these two baseline values (by taking `g_start_usec`, and adding to it the "approx time passed in microseconds" based on `g_start_tsc` plus current TSC value).

So in your code, we update `g_start_tsc` to some timestamp in microseconds, and we must also associate a timestamp in TSC to it. We can think of your code as "update `g_start_tsc` to some kinda random value", and TSC value will be roughly the value at this update/assignment time. And `tsc_cyc` is exactly such a rough value.

(Hm, I have a hard time explaining this in words. I hope my text above makes sense to you.)

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1276)* status: 0 of 1 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`pal/src/host/linux-sgx/pal_misc.c` line 104 at r1](https://reviewable.io/reviews/gramineproject/gramine/1276#-NSjITEd508rLtl0LLTy:-NSji3S29IiByEiE9XYY:b-l142ej) ([raw file](https://github.com/gramineproject/gramine/blob/1f6149f9a45b969921890e256feb09bda6152772/pal/src/host/linux-sgx/pal_misc.c#L104)):*
> So in your code, we update g_start_tsc to some timestamp in microseconds, and we must also associate a timestamp in TSC to it. 

I think it's a typo here -- should be `g_start_usec`?

> We can think of your code as "update g_start_tsc to some kinda random value", and TSC value will be roughly the value at this update/assignment time. And tsc_cyc is exactly such a rough value.

Thanks for the explanation! It resolved my confusions. Updated.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1276)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

___
*[`pal/src/host/linux-sgx/pal_misc.c` line 104 at r1](https://reviewable.io/reviews/gramineproject/gramine/1276#-NSjITEd508rLtl0LLTy:-NSjxUQy1iTcHQXGGe6y:bwbdiwo) ([raw file](https://github.com/gramineproject/gramine/blob/1f6149f9a45b969921890e256feb09bda6152772/pal/src/host/linux-sgx/pal_misc.c#L104)):*
> I think it's a typo here -- should be `g_start_usec`?

Yes, sorry, this was a typo.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1276)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

___
*[`pal/src/host/linux-sgx/pal_misc.c` line 80 at r1](https://reviewable.io/reviews/gramineproject/gramine/1276#-NSjIuY-D9RsMhzQX8NY:-NSraa0N-4X444ZWAjzv:b-4vj20t) ([raw file](https://github.com/gramineproject/gramine/blob/1f6149f9a45b969921890e256feb09bda6152772/pal/src/host/linux-sgx/pal_misc.c#L80)):*

This looks brittle in terms of data races. But I thought about it a bit and couldn't come up with a multi-threaded scenario where something could go wrong in these store/load of `last_usec_rdtsc`. (One important thing to note is that if we hit this `atomic_store` line, then we will return from this function and won't go further into the OCALL invocation -- this seems to prevent data races I was thinking of.)

So I'm not blocking, but somehow I feel uneasy about this.
</blockquote></details>

Sample race:
1. thread 1 executes up until this store, but gets interrupted right before it
2. time passes and TSC for some reason gets ahead of host time
3. thread 2 calls `_PalSystemTimeQuery()`, gets time ahead of the host time
4. thread 2 hits `_PalSystemTimeQuery()` again, but this time `diff_usec >= TSC_REFINE_INIT_TIMEOUT_USECS`, so it has to recalculate.
5. thread 1 gets rescheduled and writes super-old value into `last_usec_rdtsc`
6. thread 2 hits #82 problem like this PR never existed :) (`_PalSystemTimeQuery()` returns a lower value than in the previous call)

@kailun-qin: Please take a look at how anti-rewinding protection is written in `ocall_gettime`, I think it's done correctly there (modulo that `*microsec_ptr = microsec` which should be `*microsec_ptr = max(microsec, expected_microsec)` ideally, but it's not incorrect).
___
*[`pal/src/host/linux-sgx/pal_misc.c` line 46 at r2](https://reviewable.io/reviews/gramineproject/gramine/1276#-NSrKr1607IYOty4RDTF:-NSrKr1607IYOty4RDTG:b-ie6lmn) ([raw file](https://github.com/gramineproject/gramine/blob/d5c2ce422c81f256ca866fd9a151334521513fc3/pal/src/host/linux-sgx/pal_misc.c#L46)):*
> ```C
> }
> 
> /* TODO: result comes from the untrusted host, introduce some schielding */
> ```

I guess this is done now? (`ocall_gettime` has an anti-time-rewinding check inside)
___
*[`pal/src/host/linux-sgx/pal_misc.c` line 66 at r2](https://reviewable.io/reviews/gramineproject/gramine/1276#-NSrZz6n5_su8-AAG7xw:-NSrZz6n5_su8-AAG7xx:b-8wnqdp) ([raw file](https://github.com/gramineproject/gramine/blob/d5c2ce422c81f256ca866fd9a151334521513fc3/pal/src/host/linux-sgx/pal_misc.c#L66)):*
> ```C
>     uint64_t usec = 0;
>     /* Last seen RDTSC time value. This guards against time rewinding. */
>     static uint64_t last_usec_rdtsc = 0;
> ```

You mean `last_used_rdtsc` or `last_usec`?



_Code quote:_
```C
last_usec_rdtsc
```
___
*[`pal/src/host/linux-sgx/pal_misc.c` line 80 at r2](https://reviewable.io/reviews/gramineproject/gramine/1276#-NSrYH2JCRTXUsKjrIHB:-NSrYH2JCRTXUsKjrIHC:b-1dbwv9) ([raw file](https://github.com/gramineproject/gramine/blob/d5c2ce422c81f256ca866fd9a151334521513fc3/pal/src/host/linux-sgx/pal_misc.c#L80)):*
> ```C
>                     return -PAL_ERROR_OVERFLOW;
> 
>                 __atomic_store_n(&last_usec_rdtsc, usec, __ATOMIC_RELEASE);
> ```

I'd move this to `if (usec) {` below, it would be easier to understand IMO. Or that `IF` moved here.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1276)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`pal/src/host/linux-sgx/pal_misc.c` line 80 at r1](https://reviewable.io/reviews/gramineproject/gramine/1276#-NSjIuY-D9RsMhzQX8NY:-NSsY4W38tR2eRuFLcuh:b-9wg2jw) ([raw file](https://github.com/gramineproject/gramine/blob/1f6149f9a45b969921890e256feb09bda6152772/pal/src/host/linux-sgx/pal_misc.c#L80)):*
Good sample! Make sense to me, updated.

> I think it's done correctly there (modulo that *microsec_ptr = microsec which should be *microsec_ptr = max(microsec, expected_microsec) ideally, but it's not incorrect).

Seems better to have `*microsec_ptr = max(microsec, expected_microsec)`. I'm also updating it.
___
*[`pal/src/host/linux-sgx/pal_misc.c` line 46 at r2](https://reviewable.io/reviews/gramineproject/gramine/1276#-NSrKr1607IYOty4RDTF:-NSsXeX7C5XEXuNm5QKV:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/d5c2ce422c81f256ca866fd9a151334521513fc3/pal/src/host/linux-sgx/pal_misc.c#L46)):*

I guess this is done now? (`ocall_gettime` has an anti-time-rewinding check inside)
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/pal_misc.c` line 66 at r2](https://reviewable.io/reviews/gramineproject/gramine/1276#-NSrZz6n5_su8-AAG7xw:-NSsXf7c0BlFstNt1RgD:b-tjyflp) ([raw file](https://github.com/gramineproject/gramine/blob/d5c2ce422c81f256ca866fd9a151334521513fc3/pal/src/host/linux-sgx/pal_misc.c#L66)):*

You mean `last_used_rdtsc` or `last_usec`?
</blockquote></details>

I intended to say "last recorded timestamp" in usec, calculated via RDTSC. Anyway, I updated the naming a bit -- see if it makes sense to me.
___
*[`pal/src/host/linux-sgx/pal_misc.c` line 80 at r2](https://reviewable.io/reviews/gramineproject/gramine/1276#-NSrYH2JCRTXUsKjrIHB:-NSsYJC01YsKAO6Ck_LP:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/d5c2ce422c81f256ca866fd9a151334521513fc3/pal/src/host/linux-sgx/pal_misc.c#L80)):*

I'd move this to `if (usec) {` below, it would be easier to understand IMO. Or that `IF` moved here.
</blockquote></details>

Done.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1276)* status: 0 of 2 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`pal/src/host/linux-sgx/pal_misc.c` line 66 at r2](https://reviewable.io/reviews/gramineproject/gramine/1276#-NSrZz6n5_su8-AAG7xw:-NSsayyLEL--lKu12g_q:b-rwpncp) ([raw file](https://github.com/gramineproject/gramine/blob/d5c2ce422c81f256ca866fd9a151334521513fc3/pal/src/host/linux-sgx/pal_misc.c#L66)):*
> see if it makes sense to me

to you ;) sorry

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1276)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin and @mkow)

___
*[`pal/src/host/linux-sgx/pal_misc.c` line 80 at r2](https://reviewable.io/reviews/gramineproject/gramine/1276#-NSrYH2JCRTXUsKjrIHB:-NStY7R71GGOvybRUQcR:b-kq1hgu) ([raw file](https://github.com/gramineproject/gramine/blob/d5c2ce422c81f256ca866fd9a151334521513fc3/pal/src/host/linux-sgx/pal_misc.c#L80)):*

Done.
</blockquote></details>

It's not wrong now, but it's superfluous check (see my other comment).

Previously, this `if (usec)` check was to make a decision -- did we calculate the time using RDTSC, or do we need to perform an OCALL. Now that you moved this `return 0`, there's no need in the check at all.
___
*[`pal/src/host/linux-sgx/pal_misc.c` line 79 at r3](https://reviewable.io/reviews/gramineproject/gramine/1276#-NStXrqQ2kdzgNoEnXDM:-NStXrqQ2kdzgNoEnXDN:b-yhf97j) ([raw file](https://github.com/gramineproject/gramine/blob/68621a567f1bd254dccfa63aff524eec7d7920e0/pal/src/host/linux-sgx/pal_misc.c#L79)):*
> ```C
>                     return -PAL_ERROR_OVERFLOW;
> 
>                 if (usec) {
> ```

What's the point in this `if`? `usec` seems to be guaranteed to have some value (plus we protected against integer overflow above).

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1276)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`pal/src/host/linux-sgx/pal_misc.c` line 79 at r3](https://reviewable.io/reviews/gramineproject/gramine/1276#-NStXrqQ2kdzgNoEnXDM:-NStdPrvBhtSRThHDBEJ:b-xgpj37) ([raw file](https://github.com/gramineproject/gramine/blob/68621a567f1bd254dccfa63aff524eec7d7920e0/pal/src/host/linux-sgx/pal_misc.c#L79)):*
> usec seems to be guaranteed to have some value (plus we protected against integer overflow above).

Yep, I agree. But I did not make such assumption -- if `start_tsc = 0`, `get_tsc() = 0` (due to some issues w/ the TSC of the core), we may get `diff_tsc = 0` then `usec = 0`? In that case, we may want to go to recalculate instead of return sth.? But this seems to be theoretical, I'm fine w/ removing this IF.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1276)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`pal/src/host/linux-sgx/pal_misc.c` line 79 at r3](https://reviewable.io/reviews/gramineproject/gramine/1276#-NStXrqQ2kdzgNoEnXDM:-NStr_zzEWoeUmGYMLbl:b-ikfrrb) ([raw file](https://github.com/gramineproject/gramine/blob/68621a567f1bd254dccfa63aff524eec7d7920e0/pal/src/host/linux-sgx/pal_misc.c#L79)):*

> usec seems to be guaranteed to have some value (plus we protected against integer overflow above).

Yep, I agree. But I did not make such assumption -- if `start_tsc = 0`, `get_tsc() = 0` (due to some issues w/ the TSC of the core), we may get `diff_tsc = 0` then `usec = 0`? In that case, we may want to go to recalculate instead of return sth.? But this seems to be theoretical, I'm fine w/ removing this IF.
</blockquote></details>

@kailun-qin But `usec = start_usec + diff_usec`, and `start_usec` is always a positive number (never a zero, because we have a check in the very beginning). So I don't think this would be even theoretically possible.

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1276)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

mkow (dismissed):
> Reviewed 1 of 2 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1276)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

___
*[`pal/src/host/linux-sgx/pal_misc.c` line 66 at r2](https://reviewable.io/reviews/gramineproject/gramine/1276#-NSrZz6n5_su8-AAG7xw:-NSusUU-20i01AxFVSAV:b-8bicdh) ([raw file](https://github.com/gramineproject/gramine/blob/d5c2ce422c81f256ca866fd9a151334521513fc3/pal/src/host/linux-sgx/pal_misc.c#L66)):*

> see if it makes sense to me

to you ;) sorry
</blockquote></details>

Oh, I see, the comment previously suggested that this variable holds raw RDTSC value and I was confused how can it be in μs.
___
*[`pal/src/host/linux-sgx/pal_misc.c` line 79 at r4](https://reviewable.io/reviews/gramineproject/gramine/1276#-NSuy9707ft9phjx4eAZ:-NSuy9707ft9phjx4eA_:bzgrfph) ([raw file](https://github.com/gramineproject/gramine/blob/c9e2adf7fb19ea3803aa87b2d16a87783971556f/pal/src/host/linux-sgx/pal_misc.c#L79)):*
> ```C
>                     return -PAL_ERROR_OVERFLOW;
> 
>                 uint64_t expected_usec = __atomic_load_n(&last_usec, __ATOMIC_ACQUIRE);
> ```

I'd add a comment here stating the intention of these complications: that it's simply `last_usec = max(last_usec, usec)`, but executed atomically.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1276)* status: 1 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`pal/src/host/linux-sgx/pal_misc.c` line 79 at r4](https://reviewable.io/reviews/gramineproject/gramine/1276#-NSuy9707ft9phjx4eAZ:-NTHDmkC9pwB-dUX6EYs:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/c9e2adf7fb19ea3803aa87b2d16a87783971556f/pal/src/host/linux-sgx/pal_misc.c#L79)):*

I'd add a comment here stating the intention of these complications: that it's simply `last_usec = max(last_usec, usec)`, but executed atomically.
</blockquote></details>

Done.

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1276)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1276)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #1277, opened 2023-04-11
[LibOS] Keep main thread alive if it's not the last one

### LINKED ISSUE #1265: Host process becomes a zombie when calling execve from another than the initial thread

dimakuv:
## Description of the changes 

This change is to cover a corner case of a non-main thread performing `execve()`. Linux does the following: the main (leader) thread is terminated, and the non-main thread assumes its identity: https://elixir.bootlin.com/linux/v6.0/source/fs/exec.c#L1078

Before this PR, Gramine simply terminated the main thread. This led to the process being marked as a zombie (even though the non-main thread runs correctly and responds to signals). This in turn confused tools like `docker kill`, which couldn't find the Gramine process anymore.

This PR fixes this corner case. Gramine can't do as Linux because there is no way to ask the host OS to "rewire" the identity of one thread to the other thread. Thus we introduce a workaround of keeping alive (but in a "parked" state) the main thread -- it sleeps infinitely.

In this case, the main thread and its associated resources (like the thread's LibOS stack) are not freed -- this leaks memory, but only once per process, as there is only one main thread per process, even after several execve invocations.

Fixes #1265.

Closes #1268.

## How to test this PR? 

I don't know how to test it in our CI reliably. We already have the test `exec_same` which does `execve()` in the non-main thread.

Anyther way to test it manually is to build and run the program in #1265 -- it will show as `defunct` when doing `ps` or similar without this PR. With this PR, it shows correctly.

Finally, the program described here also works: https://github.com/gramineproject/gramine/pull/1268#pullrequestreview-1372917185

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1277)

### DIFF
--- a/libos/src/sys/libos_exit.c
+++ b/libos/src/sys/libos_exit.c
@@ -71,6 +71,39 @@ noreturn void thread_exit(int error_code, int term_signal) {
 
     /* Remove current thread from the threads list. */
     if (!check_last_thread(/*mark_self_dead=*/true)) {
+        if (cur_thread->pal_handle == g_pal_public_state->first_thread) {
+            /*
+             * Do not exit the main thread (and do not free its resources) if the main thread is not
+             * the last one in the process. This is added to correctly handle the case of a non-main
+             * thread performing `execve()`, even after the main thread is considered terminated.
+             *
+             * The main thread waits forever so that the host OS doesn't "lose track" of this
+             * Gramine process. E.g. on Linux, if the main thread (aka leader thread) terminates,
+             * then the process becomes a zombie, which may confuse some tools like `docker kill`.
+             * This "waiting forever" leaks memory, but only once per process (as there is only one
+             * main thread per process, even after several execve invocations).
+             *
+             * Linux solves this corner case differently: the leader thread is terminated, and the
+             * non-main thread assumes its identity (in particular, its PID):
+             *
+             *   https://elixir.bootlin.com/linux/v6.0/source/fs/exec.c#L1078
+             *
+             * Gramine can't do the same because there is no way to ask the host OS to "rewire" the
+             * identity of one thread to another thread. Thus this workaround of infinite wait.
+             * Note that because the main thread was removed from the list of threads (thanks to
+             * `mark_self_dead=true` above), the still-alive main thread will not prevent the
+             * Gramine process from terminating later on. Also note that because this thread never
+             * leaves LibOS/PAL context, it will not receive signals.
+             *
+             * TODO: "Rewire" the identity of the non-main thread inside Gramine, similarly to how
+             *       Linux does it.
+             */
+            thread_prepare_wait();
+            while (true)
+                thread_wait(/*timeout_us=*/NULL, /*ignore_pending_signals=*/true);
+            __builtin_unreachable();
+        }
+
         /* ask async worker thread to cleanup this thread */
         cur_thread->clear_child_tid_pal = 1; /* any non-zero value suffices */
         /* We pass this ownership to `cleanup_thread`. */


kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1277)* status: 0 of 1 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`libos/src/sys/libos_exit.c` line 77 at r1](https://reviewable.io/reviews/gramineproject/gramine/1277#-NSn9dIZ5qFhZjtyBCjo:-NSn9dIZ5qFhZjtyBCjp:b6pjf1l) ([raw file](https://github.com/gramineproject/gramine/blob/3fe60f11bf4605e73a70a90eabad6748d085cf91/libos/src/sys/libos_exit.c#L77)):*
> ```C
>             /*
>              * Do not exit the main thread (and do not free its resources) if the main thread is not
>              * the last one in the process. This corner case is added to correctly handle the case
> ```

This is added to correctly handle the corner case



_Code quote:_
```C
This corner case is added to correctly handle the case
```
___
*[`libos/src/sys/libos_exit.c` line 82 at r1](https://reviewable.io/reviews/gramineproject/gramine/1277#-NSnBKM_C66JIcZmFTbO:-NSnBKM_C66JIcZmFTbP:bbq01h2) ([raw file](https://github.com/gramineproject/gramine/blob/3fe60f11bf4605e73a70a90eabad6748d085cf91/libos/src/sys/libos_exit.c#L82)):*
> ```C
>              *
>              * This leaks memory, but only once per process (as there is only one main thread per
>              * process, even after several execve invocations). Instead, wait forever so that the
> ```

Sorry but this "Instead" reads a bit weird to me here (after "this leaks memory, but only once per process...").



_Code quote:_
```C
Instead
```

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1277)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv and @kailun-qin)

___
*[`-- commits` line 17 at r1](https://reviewable.io/reviews/gramineproject/gramine/1277#-NSnFfHH6fTpPJnmrOlA):*




_Suggestion:_
```
to another thread
```
___
*[`libos/src/sys/libos_exit.c` line 77 at r1](https://reviewable.io/reviews/gramineproject/gramine/1277#-NSn9dIZ5qFhZjtyBCjo:-NSnBjlY0CWrA2Ho7RS7:bgskz2) ([raw file](https://github.com/gramineproject/gramine/blob/3fe60f11bf4605e73a70a90eabad6748d085cf91/libos/src/sys/libos_exit.c#L77)):*
> This corner case is added

heh :)
___
*[`libos/src/sys/libos_exit.c` line 92 at r1](https://reviewable.io/reviews/gramineproject/gramine/1277#-NSnEM8y7wKJJ2SakJlH:-NSnEM8y7wKJJ2SakJlI:bcfmxp9) ([raw file](https://github.com/gramineproject/gramine/blob/3fe60f11bf4605e73a70a90eabad6748d085cf91/libos/src/sys/libos_exit.c#L92)):*
> ```C
>              *   https://elixir.bootlin.com/linux/v6.0/source/fs/exec.c#L1078
>              *
>              * Gramine can't do the same because there is no way to ask the host OS to "rewire" the
> ```

Actually, shouldn't we do this _inside_ Gramine? Not sure if that matters for anything, but Linux for some reason tries to ensure that after execve the main thread has tid == pid.
And it's unrelated to the original issue which was about the state of the host process on PAL/Linux-*, but maybe we should still emulate this behavior inside Gramine? (but not necessarily via this PR)



_Code quote:_
```C
Gramine can't do the same
```
___
*[`libos/src/sys/libos_exit.c` line 93 at r1](https://reviewable.io/reviews/gramineproject/gramine/1277#-NSnC6yL6EPooJhbkGQb:-NSnC6yL6EPooJhbkGQc:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/3fe60f11bf4605e73a70a90eabad6748d085cf91/libos/src/sys/libos_exit.c#L93)):*
> ```C
>              *
>              * Gramine can't do the same because there is no way to ask the host OS to "rewire" the
>              * identity of one thread to the other thread. Thus this workaround of infinite wait.
> ```





_Suggestion:_
```C
to another thread
```

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1277)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv, @kailun-qin, and @mkow)

___
*[`-- commits` line 17 at r1](https://reviewable.io/reviews/gramineproject/gramine/1277#-NSnFfHH6fTpPJnmrOlA):*
Will do during final rebase
___
*[`libos/src/sys/libos_exit.c` line 77 at r1](https://reviewable.io/reviews/gramineproject/gramine/1277#-NSn9dIZ5qFhZjtyBCjo:-NSoeku71doVSiaOEJHv:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/3fe60f11bf4605e73a70a90eabad6748d085cf91/libos/src/sys/libos_exit.c#L77)):*

> This corner case is added

heh :)
</blockquote></details>

Done.
___
*[`libos/src/sys/libos_exit.c` line 82 at r1](https://reviewable.io/reviews/gramineproject/gramine/1277#-NSnBKM_C66JIcZmFTbO:-NSof3zO2Su7IHeTrca5:b-219gqf) ([raw file](https://github.com/gramineproject/gramine/blob/3fe60f11bf4605e73a70a90eabad6748d085cf91/libos/src/sys/libos_exit.c#L82)):*

Sorry but this "Instead" reads a bit weird to me here (after "this leaks memory, but only once per process...").
</blockquote></details>

Done. Rearranged the sentences.
___
*[`libos/src/sys/libos_exit.c` line 92 at r1](https://reviewable.io/reviews/gramineproject/gramine/1277#-NSnEM8y7wKJJ2SakJlH:-NSofTRZ7amfrmVtIqJ7:buypfi5) ([raw file](https://github.com/gramineproject/gramine/blob/3fe60f11bf4605e73a70a90eabad6748d085cf91/libos/src/sys/libos_exit.c#L92)):*

Actually, shouldn't we do this _inside_ Gramine? Not sure if that matters for anything, but Linux for some reason tries to ensure that after execve the main thread has tid == pid.
And it's unrelated to the original issue which was about the state of the host process on PAL/Linux-*, but maybe we should still emulate this behavior inside Gramine? (but not necessarily via this PR)
</blockquote></details>

I added a TODO comment and I created a GitHub issue: https://github.com/gramineproject/gramine/issues/1283
___
*[`libos/src/sys/libos_exit.c` line 93 at r1](https://reviewable.io/reviews/gramineproject/gramine/1277#-NSnC6yL6EPooJhbkGQb:-NSoh7bv3wB9uVRammOI:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/3fe60f11bf4605e73a70a90eabad6748d085cf91/libos/src/sys/libos_exit.c#L93)):*
> ```C
>              *
>              * Gramine can't do the same because there is no way to ask the host OS to "rewire" the
>              * identity of one thread to the other thread. Thus this workaround of infinite wait.
> ```

Done.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1277)* status: 0 of 1 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1277#-NSopowd6OS9V2Alam3a:-NSopowd6OS9V2Alam3b:b-ilupkq) (no related file):*
I was informed by the customer that this PR fixes their real-world case.

(The customer also informed me that the previous attempt at the fix -- PR #1268 -- also fixed their real-world case.)

___

kailun-qin (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1277#-:-NSoya_7AEJhANPCyCgI:bnfp4nl)* status: 0 of 1 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1277)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1277)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

___
*[`-- commits` line 17 at r1](https://reviewable.io/reviews/gramineproject/gramine/1277#-NSnFfHH6fTpPJnmrOlA):*

Will do during final rebase
</blockquote></details>

Done

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1277)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #1325, opened 2023-05-03
[Pal/Linux-SGX] Fix memory leak

mtcherni95:
## Description of the changes 
freeing `manifest_path` pointer before the app returns. 

## How to test this PR? 

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1325)

### DIFF
--- a/pal/src/host/linux-sgx/host_main.c
+++ b/pal/src/host/linux-sgx/host_main.c
@@ -1230,6 +1230,7 @@ int main(int argc, char* argv[], char* envp[]) {
         ret = read_text_file_to_cstr(manifest_path, &manifest);
         if (ret < 0) {
             log_error("Reading manifest failed");
+            free(manifest_path);
             return ret;
         }
         free(manifest_path);


oshogbo (commented):
I believe we simply allow OS to clear the free space (especially thats the main function).

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1325#-:-NUVsoQPA0GDJzSOHgOt:b-7itlhu)* status: 0 of 1 files reviewed, all discussions resolved, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: )

oshogbo:
In case somebody else has a different view:

Jenkins, please test this

oshogbo:
Jenkins, please test this

oshogbo:
In case somebody else has a different view:

Jenkins, test this please

kailun-qin (commented):
Yeah, I think we basically consider memory'll be reclaimed by Linux in such init failures. I agree that doing explicit `free()`s can be good practice but I do not really care in this case. (BTW, if we really want to manage this, there are more places to look into.)

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1325#-:-NUZGIzLDaIEy0ePbgGs:bpk84iv)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: )

dimakuv (commented):
> The leak is before returning from main function, so it does not really affect the system. However, theoretically it is a leak and I think it's good practice to fix them whenever possible.

I am negative towards this PR. I would prefer to close it.

Yes, this is a memory leak but:
1. Why fix exactly this leak and not the others? There are tons of such at-init-time leaks that we don't care about. I see no reason to special-case this single leak.
2. This leak happens in the untrusted-PAL code (on the untrusted host and **not** inside the SGX enclave). So I care even less about this leak...

We can wait for @mkow's opinion, but I would close this PR without merging.

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1325#-:-NUa95nJ-qGyA3-3daM1:b-ksmg6p)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: )

mkow (commented):
I don't consider such things to be memory leaks. The memory is used until the end of the process' life and the path you're changing immediately kills the whole process.
We had a discussion about this in the past and decided that it's not useful to manually free everything before exiting the process, it just bloats the code and gives nothing in return.

It seems that the maintainers are unanimous here, so I'm closing this PR without merging. But feel free to provide more arguments if you think there's some reason why we should do this.

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1325#-:-NUdYFtQ6dI-iPF_EHW1:bzf0zp)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: )

PR #1328, opened 2023-05-04
[PAL/Linux] Log a message on `munmap()` returning `-ENOMEM`

### LINKED ISSUE #1327: Redis Server fails with Gramine Direct when run for data size higher than 32768

We have a setup where Redis server is run on a server which is connected to a client via a 10GB P2P cable and the client runs the memtier benchmark. For performance benchmarking we are running this setup for request size starting from 48 to 4096. 

When we tried to enable a higher request size (upwards of 32768) from memtier benchmark, redis server running under gramine-direct fails with the following error and crashes:

```
(libos_mmap.c:191:libos_syscall_mmap) [P1:T5:redis-server] error: BUG() ../libos/src/sys/libos_mmap.c:191
(libos_signal.c:312:internal_fault) [P1:T5:redis-server] error: Illegal instruction during Gramine internal execution at 0x7fffffd460b0 (0x7fffffd460b0, VMID = 1, TID = 5)
```
whereas on the memtier benchmark side we see `Connection rest by peer` error.

I have uploaded a trimmed down log file which log level set to trace.

[redis_gramine_direct_failure_32768_data_size.txt](https://github.com/gramineproject/gramine/files/11396816/redis_gramine_direct_failure_32768_data_size.txt)

### Steps to reproduce

Redis server command:

```
numactl -C 1 gramine-direct redis-server --port 9001 --maxmemory 4294967296 --maxmemory-policy allkeys-lru --appendonly no --protected-mode no --save ''
```

Memtier benchmark command:

```
numactl -m 0 taskset -c 0 /home/intel/perf_benchmarking/memtier_benchmark/memtier_benchmark -s 192.168.0.10 -p 9001 --key-maximum=100000 -d 32768 --randomize --test-time=180 --ratio=1:1 --pipeline=6 -c 6 -t 7 --distinct-client-seed --out-file=file_name.csv
```

### Expected results

_No response_

### Actual results

_No response_

### Gramine commit hash

Master

dimakuv:
## Description of the changes 

Linux can return `-ENOMEM` if a request to `munmap()` will exceed the process's maximum number of mappings (this is possible in case of partial memory unmap, when Linux would have to create two mappings out of one). Gramine's LibOS is not ready for such failures, and also it is specific to the Linux PAL, so this commit simply makes this failure more visible and provides a hint for the user what can be done on the host.

Fixes #1327 (though not a real fix, more like a verbose failure).

## How to test this PR? 

Manually test like described in https://github.com/gramineproject/gramine/issues/1327. This can be hit only on specific and memory-hungry workloads.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1328)

### DIFF
--- a/pal/src/host/linux/pal_memory.c
+++ b/pal/src/host/linux/pal_memory.c
@@ -46,6 +46,11 @@ int _PalVirtualMemoryAlloc(void* addr, size_t size, pal_prot_flags_t prot) {
 
 int _PalVirtualMemoryFree(void* addr, size_t size) {
     int ret = DO_SYSCALL(munmap, addr, size);
+    if (ret == -ENOMEM && FIRST_TIME()) {
+        log_error("Host Linux returned -ENOMEM on munmap(); this may happen because process's "
+                  "maximum number of mappings is exceeded. Gramine cannot handle this case. "
+                  "You may want to increase the value in /proc/sys/vm/max_map_count.");
+    }
     return ret < 0 ? unix_to_pal_error(ret) : 0;
 }
 


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1328)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1328#-NUb5R519NbesMCcXN96:-NUb5R519NbesMCcXN97:b3gfxa) (no related file):*
We could discuss if we need a proper fix, rather than just an error log like in this PR.

First of all, this problem is specific **only** to the Linux PAL. On Linux-SGX, memory management happens entirely inside the enclave (for classic non-EDMM flows) or entirely via the `SGX_IOC_ENCLAVE_REMOVE_PAGES` ioctl (for EDMM flows). Moreover, afaik, Windows doesn't have this problem as well (as it doesn't allow partial unmap of memory regions).

Second of all, the logic in `libos_mmap.c` relies on the fact that unmapping memory never fails. Otherwise undoing the operations would become very complex (possibly impossible to do right?).

Well, what we could do is to **ignore errors** of `PalVirtualMemoryFree()`. This way Gramine would have memory leaks on unmaps, but there are two scenarios afterwards:
1. The application continues doing memory management. As soon as the app will issue `mmap()`, it will fail with `ENOMEM` (for the same reason -- max number of mappings is hit).
2. The application doesn't do any more memory management. Then that memory leak is at least limited.

What do people think?

___

dimakuv:
Jenkins, retest Jenkins-Debug-20.04 please (networking issues)

dimakuv:
Jenkins, retest Jenkins-Direct-Sanitizers please (networking issues)

dimakuv:
Jenkins, retest Jenkins-Direct-Sanitizers please (networking issues)

dimakuv:
Jenkins, retest Jenkins-Debug-20.04 please (networking issues)

dimakuv:
Jenkins, retest Jenkins-20.04 please

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1328)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1328#-NUb5R519NbesMCcXN96:-NUd_rj92C4U0tHIlIAS:bdbdaab) (no related file):*
> Well, what we could do is to **ignore errors** of `PalVirtualMemoryFree()`.

I think you mean ignore only `ENOMEM` here, right? `munmap` can fail w/ e.g., `EINVAL` where the elaboration above doesn't seem to apply any more.

> The application continues doing memory management. As soon as the app will issue `mmap()`, it will fail with `ENOMEM` (for the same reason -- max number of mappings is hit).

Is this assumption always right? What if the app continues by issuing several "legitime" (i.e., not partial) `munmap()` followed by a `mmap()`, wil it also fail w/ `ENOMEM`? Then is it possible that Gramine will have more memory leaks on the following failed `munmap()`s?

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1328)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1328#-NUb5R519NbesMCcXN96:-NUeT3w535JhzkIN19qq:b8timl) (no related file):*
> I think you mean ignore only `ENOMEM` here, right? `munmap` can fail w/ e.g., `EINVAL` where the elaboration above doesn't seem to apply any more.

Yes, only ignore `ENOMEM`, fail on all other errors.

> Is this assumption always right? What if the app continues by issuing several "legitime" (i.e., not partial) `munmap()` followed by a `mmap()`, wil it also fail w/ `ENOMEM`? Then is it possible that Gramine will have more memory leaks on the following failed `munmap()`s?

Yes, true, in the case of multiple munmaps, it is possible that the current number of mappings decreases, so that the process will be able to successfully do `mmap()` later. So yes, definitely memory leaks are possible; in the worst case, the leak will be unbounded (though very improbable).

___

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1328)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1328#-NUb5R519NbesMCcXN96:-NUrjnof33n6tU7k0N8Y:bg5spgf) (no related file):*

> I think you mean ignore only `ENOMEM` here, right? `munmap` can fail w/ e.g., `EINVAL` where the elaboration above doesn't seem to apply any more.

Yes, only ignore `ENOMEM`, fail on all other errors.

> Is this assumption always right? What if the app continues by issuing several "legitime" (i.e., not partial) `munmap()` followed by a `mmap()`, wil it also fail w/ `ENOMEM`? Then is it possible that Gramine will have more memory leaks on the following failed `munmap()`s?

Yes, true, in the case of multiple munmaps, it is possible that the current number of mappings decreases, so that the process will be able to successfully do `mmap()` later. So yes, definitely memory leaks are possible; in the worst case, the leak will be unbounded (though very improbable).
</blockquote></details>

I wouldn't try do of these workarounds and just tell the user to change the system settings (as this PR currently does). It's not really a Gramine problem, such workloads would have similar issues natively on Linux.

___

kailun-qin (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1328)* status: :shipit: complete! all files reviewed, all discussions resolved

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1328)* status: :shipit: complete! all files reviewed, all discussions resolved

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1328#-NUb5R519NbesMCcXN96:-NUuAGSA8K1uJqqecWGn:b2zarz6) (no related file):*

I wouldn't try do of these workarounds and just tell the user to change the system settings (as this PR currently does). It's not really a Gramine problem, such workloads would have similar issues natively on Linux.
</blockquote></details>

Ok, Michal has a point. Let's use this PR as is.

___

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1328)* status: :shipit: complete! all files reviewed, all discussions resolved

kailun-qin (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1328)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #1351, opened 2023-05-16
[tests] Fix compiler optimization in RW lock test

kailun-qin:
## Description of the changes 
The issue was found on GCC >= 11 with `-O3` optimization level where the compiler generated the test binary unexpectedly.

Fixes https://github.com/gramineproject/gramine/issues/1352.

## How to test this PR? 
Manual testing since we don't have a Ubuntu22.04 CI pipeline (w/ GCC >= 11) setup yet

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1351)

### DIFF
--- a/libos/test/regression/rwlock.c
+++ b/libos/test/regression/rwlock.c
@@ -86,25 +86,26 @@ static int writer_(void* args_) {
 static void run_test(size_t iterations, size_t readers_num, size_t writers_num,
                      size_t writers_delay_us) {
     struct shared_state m = {1, 0, 0, 1};
-    void* lock;
-    if (!gramine_rwlock_create(&lock))
-        errx(1, "gramine_rwlock_create failed");
-
-    thrd_t* threads = calloc(sizeof(*threads), readers_num + writers_num);
-    if (!threads)
-        errx(1, "calloc failed");
-
     struct reader_args reader_args = {
-        .lock = lock,
         .m = &m,
         .total_iterations = iterations * writers_num,
     };
     struct writer_args writer_args = {
-        .lock = lock,
         .m = &m,
         .iterations = iterations,
         .writers_delay_us = writers_delay_us,
     };
+
+    void* lock;
+    if (!gramine_rwlock_create(&lock))
+        errx(1, "gramine_rwlock_create failed");
+
+    reader_args.lock = lock;
+    writer_args.lock = lock;
+
+    thrd_t* threads = calloc(sizeof(*threads), readers_num + writers_num);
+    if (!threads)
+        errx(1, "calloc failed");
 
     /* Spawn readers */
     for (size_t i = 0; i < readers_num; i++) {


kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1351)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

___
*[`-- commits` line 5 at r1](https://reviewable.io/reviews/gramineproject/gramine/1351#-NVXwzvn3TJdsqX4Mbl4):*
I should dig further, just haven't got a chance. Let me update then.



_Code quote:_
```
generated the test binary unexpectedly
```

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1351)* status: 0 of 1 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @kailun-qin)

___
*[`libos/test/regression/rwlock.c` line 100 at r1](https://reviewable.io/reviews/gramineproject/gramine/1351#-NVYa2Ze9yJCRZWWWBKt:-NVYa2Ze9yJCRZWWWBKu:b-vcteb8) ([raw file](https://github.com/gramineproject/gramine/blob/a25da06b61cf6e933e313cd17a5db047d6cba988/libos/test/regression/rwlock.c#L100)):*
> ```C
> 
>     void* lock;
>     if (!gramine_rwlock_create(&lock))
> ```

My theory why this happens: we have some mismatch in memory allocation routines inside Gramine LibOS and in the application.

The problem seems to stem from the fact that we create a RWLock internal-Gramine object during `gramine_rwlock_create()`: https://github.com/gramineproject/gramine/blob/ede26cdcfc07a92716aaea44c588ab7d886fc1fd/libos/src/libos_call.c#L112

So the application suddenly gets a reference to the Gramine-internal memory region. And then the app passes this reference (which is what `lock` variable points to) to other threads. So it feels like a dangerous thing to perform.

@kailun-qin Could you try to modify this function: https://github.com/gramineproject/gramine/blob/ede26cdcfc07a92716aaea44c588ab7d886fc1fd/libos/src/libos_call.c#L109-L120

Such that the **application** provides the memory for the lock object, and this function won't need to perform a `malloc()`.

Similarly, the destroy of the lock should not use `free()`: https://github.com/gramineproject/gramine/blob/ede26cdcfc07a92716aaea44c588ab7d886fc1fd/libos/src/libos_call.c#L124

I have a strong suspicion this is the root cause of our problems. Probably the memory regions of the LibOS and the application clash, and we have buffers overridden.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1351)* status: 0 of 1 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @kailun-qin)

___
*[`libos/test/regression/rwlock.c` line 100 at r1](https://reviewable.io/reviews/gramineproject/gramine/1351#-NVYa2Ze9yJCRZWWWBKt:-NVYc3pj5ZH4Ap4f4x6y:b-irhzz8) ([raw file](https://github.com/gramineproject/gramine/blob/a25da06b61cf6e933e313cd17a5db047d6cba988/libos/test/regression/rwlock.c#L100)):*

My theory why this happens: we have some mismatch in memory allocation routines inside Gramine LibOS and in the application.

The problem seems to stem from the fact that we create a RWLock internal-Gramine object during `gramine_rwlock_create()`: https://github.com/gramineproject/gramine/blob/ede26cdcfc07a92716aaea44c588ab7d886fc1fd/libos/src/libos_call.c#L112

So the application suddenly gets a reference to the Gramine-internal memory region. And then the app passes this reference (which is what `lock` variable points to) to other threads. So it feels like a dangerous thing to perform.

@kailun-qin Could you try to modify this function: https://github.com/gramineproject/gramine/blob/ede26cdcfc07a92716aaea44c588ab7d886fc1fd/libos/src/libos_call.c#L109-L120

Such that the **application** provides the memory for the lock object, and this function won't need to perform a `malloc()`.

Similarly, the destroy of the lock should not use `free()`: https://github.com/gramineproject/gramine/blob/ede26cdcfc07a92716aaea44c588ab7d886fc1fd/libos/src/libos_call.c#L124

I have a strong suspicion this is the root cause of our problems. Probably the memory regions of the LibOS and the application clash, and we have buffers overridden.
</blockquote></details>

One caveat here is that the app doesn't know the size of `struct libos_rwlock`. Since this is just a test, I think it's enough to generously allocate 4KB of memory and pass this (together with size) to `gramine_rwlock_create(mem_for_lock, size_of_lock)`, and Gramine should check that `sizeof(struct libos_rwlock) < size_of_lock`.

PR #1355, opened 2023-05-16
[LibOS] Add SSE registers to clobber list in gramine_call

### LINKED ISSUE #1352: rwlock Libos Regression test fails on Ubuntu 22.04, Almalinux, Rockylinux, CentOS 9

rwlock Libos Regression test fails on Ubuntu 22.04, Almalinux, Rockylinux, CentOS 9. The test is just stuck.

### Steps to reproduce

1. Build Gramine
2. cd libos/test/regression
3. gramine-test build (Build Libos Regression Test)
4. python3 -m pytest -s -v -k test_001_rwlock (Run test)

### Expected results

Test OK

### Actual results

```
python3 -m pytest -s -v -k test_001_rwlock
=========================================================================== test session starts ============================================================================
platform linux -- Python 3.10.6, pytest-6.2.5, py-1.10.0, pluggy-0.13.0 -- /usr/bin/python3
cachedir: .pytest_cache
rootdir: /home/intel/anjali/gramine, configfile: pytest.ini
plugins: xdist-2.5.0, forked-1.4.0
collected 151 items / 150 deselected / 1 selected                                                                                                                          

test_libos.py::TC_00_Unittests::test_001_rwlock FAILED

================================================================================= FAILURES =================================================================================
_____________________________________________________________________ TC_00_Unittests.test_001_rwlock ______________________________________________________________________
test_libos.py:40: in test_001_rwlock
    stdout, _ = self.run_binary(['rwlock', str(instances), str(iterations), str(readers_num),
/home/intel/gramine_install/usr/lib/python3.10/site-packages/graminelibos/regression.py:224: in run_binary
    _returncode, stdout, stderr = run_command(cmd, timeout=timeout, **kwds)
/home/intel/gramine_install/usr/lib/python3.10/site-packages/graminelibos/regression.py:157: in run_command
    raise AssertionError('Command {} timed out after {} s'.format(cmd, timeout))
E   AssertionError: Command ['/home/intel/gramine_install/usr/lib/x86_64-linux-gnu/gramine/direct/loader', '/home/intel/gramine_install/usr/lib/x86_64-linux-gnu/gramine/direct/libpal.so', 'init', 'rwlock', '5', '100', '10', '3', '100'] timed out after 45 s
========================================================================= short test summary info ==========================================================================
FAILED test_libos.py::TC_00_Unittests::test_001_rwlock - AssertionError: Command ['/home/intel/gramine_install/usr/lib/x86_64-linux-gnu/gramine/direct/loader', '/home/in...
==================================================================== 1 failed, 150 deselected in 45.54s ====================================================================
```

### Gramine commit hash

9ead920f7c44455b0cef3c27dcd496d0a908bfec

boryspoplawski:
## Description of the changes 
`gramine_call` is a custom interface to test Gramine internal functions. It uses Gramine syscall emulation and hence does not save/restore SEE registers (see "libos/src/arch/x86_64/syscallas.S" for more details). This turned out to be problematic, since test code calling `gramine_call` might assume SSE registers are not clobbered, since `gramine_call` is in reality just an inline asm block. This commit fixed that by explicitly adding all SSE registers to the clobber list.

Fixes #1352
Closes #1354

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1355)

### DIFF
--- a/libos/include/arch/x86_64/gramine_entry_api.h
+++ b/libos/include/arch/x86_64/gramine_entry_api.h
@@ -43,6 +43,20 @@ __asm__(
     ".endm\n"
 );
 
+/* Magic syscall number for Gramine custom calls */
+#define GRAMINE_CUSTOM_SYSCALL_NR 0x100000
+
+/* Some GCC versions warn about unused parameters in naked functions. Bug? */
+__attribute__((naked)) static int gramine_call(int number __attribute__((unused)),
+                                               unsigned long arg1 __attribute__((unused)),
+                                               unsigned long arg2 __attribute__((unused))) {
+    __asm__ (
+        "mov $" GRAMINE_XSTR(GRAMINE_CUSTOM_SYSCALL_NR) ", %eax\n"
+        "GRAMINE_SYSCALL\n"
+        "ret\n"
+    );
+}
+
 #undef GRAMINE_XSTR
 #undef GRAMINE_STR
 
@@ -60,19 +74,6 @@ enum {
     GRAMINE_CALL_RWLOCK_WRITE_LOCK,
     GRAMINE_CALL_RWLOCK_WRITE_UNLOCK,
 };
-
-/* Magic syscall number for Gramine custom calls */
-#define GRAMINE_CUSTOM_SYSCALL_NR 0x100000
-
-static inline int gramine_call(int number, unsigned long arg1, unsigned long arg2) {
-    int ret;
-    __asm__ volatile(
-        "GRAMINE_SYSCALL"
-        : "=a"(ret)
-        : "0"(GRAMINE_CUSTOM_SYSCALL_NR), "D"(number), "S"(arg1), "d"(arg2)
-        : "rcx", "r11", "memory");
-    return ret;
-}
 
 static inline int gramine_register_library(const char* name, unsigned long load_address) {
     return gramine_call(GRAMINE_CALL_REGISTER_LIBRARY, (unsigned long)name, load_address);


kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1355)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1355#-NVZvRSs744FKIcUFZ1S:-NVZzXtvA5ZqZ7s_rIpC:bxj33ti) (no related file):*

@dimakuv please verify that it indeed fixes the issue.
</blockquote></details>

I verified it manually, it fixes the issue.

___

kailun-qin (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1355)* status: 0 of 1 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski and @dimakuv)

___
*[`libos/include/arch/x86_64/gramine_entry_api.h` line 75 at r1](https://reviewable.io/reviews/gramineproject/gramine/1355#-NV_9KTz-6gUtdSQuI_I:-NV_9KTz-6gUtdSQuI_J:b-smcbrp) ([raw file](https://github.com/gramineproject/gramine/blob/72122616438a5141fa4a4530d303338bca76eae5/libos/include/arch/x86_64/gramine_entry_api.h#L75)):*
> ```objc
>         : "rcx", "r11", "memory",
>           "xmm0", "xmm1", "xmm2", "xmm3", "xmm4", "xmm5", "xmm6", "xmm7", "xmm8", "xmm9", "xmm10",
>           "xmm11", "xmm12", "xmm13", "xmm14", "xmm15");
> ```

What about AVX registers then?



_Code quote:_
```objc
          "xmm0", "xmm1", "xmm2", "xmm3", "xmm4", "xmm5", "xmm6", "xmm7", "xmm8", "xmm9", "xmm10",
          "xmm11", "xmm12", "xmm13", "xmm14", "xmm15");
```

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1355)* status: 0 of 1 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski and @kailun-qin)

___
*[`libos/include/arch/x86_64/gramine_entry_api.h` line 75 at r1](https://reviewable.io/reviews/gramineproject/gramine/1355#-NV_9KTz-6gUtdSQuI_I:-NV_AoRQ1lSvPhWfaBwy:bt6agtt) ([raw file](https://github.com/gramineproject/gramine/blob/72122616438a5141fa4a4530d303338bca76eae5/libos/include/arch/x86_64/gramine_entry_api.h#L75)):*

What about AVX registers then?
</blockquote></details>

+1, should we care about other types of extended-state regs?

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1355)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski, @dimakuv, and @kailun-qin)

___
*[`libos/include/arch/x86_64/gramine_entry_api.h` line 75 at r1](https://reviewable.io/reviews/gramineproject/gramine/1355#-NV_9KTz-6gUtdSQuI_I:-NVaRBW3BWu4eRzoLJ_g:bo3t6u) ([raw file](https://github.com/gramineproject/gramine/blob/72122616438a5141fa4a4530d303338bca76eae5/libos/include/arch/x86_64/gramine_entry_api.h#L75)):*

+1, should we care about other types of extended-state regs?
</blockquote></details>

Probably yes, if we compile LibOS with support for them.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1355)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv, @kailun-qin, and @mkow)

___
*[`libos/include/arch/x86_64/gramine_entry_api.h` line 75 at r1](https://reviewable.io/reviews/gramineproject/gramine/1355#-NV_9KTz-6gUtdSQuI_I:-NV_Oqs8BONot4jCDbAI:b1qtl0m) ([raw file](https://github.com/gramineproject/gramine/blob/72122616438a5141fa4a4530d303338bca76eae5/libos/include/arch/x86_64/gramine_entry_api.h#L75)):*

Probably yes, if we compile LibOS with support for them.
</blockquote></details>

Can compiler assume upper parts of ymm/zmm are not changed while xmm changes? I have no idea what are semantics about that.
Also there are xmm16-xmm31, but they are not available on the most of the targets (so gcc won't compile code with such clobber registers). What do we do about that?

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1355)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @kailun-qin)

___
*[`libos/include/arch/x86_64/gramine_entry_api.h` line 75 at r1](https://reviewable.io/reviews/gramineproject/gramine/1355#-NV_9KTz-6gUtdSQuI_I:-NVahs1I6KaDRfZrR742:b9esxbz) ([raw file](https://github.com/gramineproject/gramine/blob/72122616438a5141fa4a4530d303338bca76eae5/libos/include/arch/x86_64/gramine_entry_api.h#L75)):*

Can compiler assume upper parts of ymm/zmm are not changed while xmm changes? I have no idea what are semantics about that.
Also there are xmm16-xmm31, but they are not available on the most of the targets (so gcc won't compile code with such clobber registers). What do we do about that?
</blockquote></details>

Ah, good point. But I'd add a comment explaining why we don't have to list the bigger regs.

And indeed, this seem to work as expected in GCC, YMMs are inferred to be clobbered by this inline asm: https://godbolt.org/z/Tqzrh8PK8 (try removing XMMs from the list there and see what happens).

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1355)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @kailun-qin and @mkow)

___
*[`libos/include/arch/x86_64/gramine_entry_api.h` line 75 at r1](https://reviewable.io/reviews/gramineproject/gramine/1355#-NV_9KTz-6gUtdSQuI_I:-NVc_OieBDfnQMVn8kUF:bsqy9u2) ([raw file](https://github.com/gramineproject/gramine/blob/72122616438a5141fa4a4530d303338bca76eae5/libos/include/arch/x86_64/gramine_entry_api.h#L75)):*

Ah, good point. But I'd add a comment explaining why we don't have to list the bigger regs.

And indeed, this seem to work as expected in GCC, YMMs are inferred to be clobbered by this inline asm: https://godbolt.org/z/Tqzrh8PK8 (try removing XMMs from the list there and see what happens).
</blockquote></details>

I think we should go with this Borys's PR rather than my hacky one (https://github.com/gramineproject/gramine/pull/1354). @boryspoplawski Could you then add the relevant comments in this code?

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1355)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv, @kailun-qin, and @mkow)

___
*[`libos/include/arch/x86_64/gramine_entry_api.h` line 75 at r1](https://reviewable.io/reviews/gramineproject/gramine/1355#-NV_9KTz-6gUtdSQuI_I:-NVqlHG78qbl346_asd1:b-ssorny) ([raw file](https://github.com/gramineproject/gramine/blob/72122616438a5141fa4a4530d303338bca76eae5/libos/include/arch/x86_64/gramine_entry_api.h#L75)):*
> And indeed, this seem to work as expected in GCC, YMMs are inferred to be clobbered by this inline asm: [https://godbolt.org/z/Tqzrh8PK8](https://godbolt.org/z/Tqzrh8PK8) (try removing XMMs from the list there and see what happens).

This actually doesn't show that fully. It only shows that GCC understands that lower half of ymm0 is clobbered (which it must assume). I wonder if it also assumes same about the upper part (which in theory is not clobbered). Also we should do something about `mm0`. And FPU (`st0`).
IMO it's too many assumptions. This kind of stuff should not be done form an inline asm, so we either should add saving these registers to Gramine entry code (could be expensive, so probably a no-go) or use a proper asm here (either separate file, or at least global scope `asm` statement).
Or we can add a long list of clobbered regs and hope it's enough, but we all should know how such wishful thinking ends...

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1355)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski, @dimakuv, and @kailun-qin)

___
*[`libos/include/arch/x86_64/gramine_entry_api.h` line 75 at r1](https://reviewable.io/reviews/gramineproject/gramine/1355#-NV_9KTz-6gUtdSQuI_I:-NVtyIJZ9fzdqBMN0sHl:b-vyfg1y) ([raw file](https://github.com/gramineproject/gramine/blob/72122616438a5141fa4a4530d303338bca76eae5/libos/include/arch/x86_64/gramine_entry_api.h#L75)):*

> And indeed, this seem to work as expected in GCC, YMMs are inferred to be clobbered by this inline asm: [https://godbolt.org/z/Tqzrh8PK8](https://godbolt.org/z/Tqzrh8PK8) (try removing XMMs from the list there and see what happens).

This actually doesn't show that fully. It only shows that GCC understands that lower half of ymm0 is clobbered (which it must assume). I wonder if it also assumes same about the upper part (which in theory is not clobbered). Also we should do something about `mm0`. And FPU (`st0`).
IMO it's too many assumptions. This kind of stuff should not be done form an inline asm, so we either should add saving these registers to Gramine entry code (could be expensive, so probably a no-go) or use a proper asm here (either separate file, or at least global scope `asm` statement).
Or we can add a long list of clobbered regs and hope it's enough, but we all should know how such wishful thinking ends...
</blockquote></details>

Hmm, maybe just write this as a non-inline assembly function and call it? Then it will be an external symbol and the compiler would be forbidden from assuming anything beyond what ABI says.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1355)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv, @kailun-qin, and @mkow)

___
*[`libos/include/arch/x86_64/gramine_entry_api.h` line 75 at r1](https://reviewable.io/reviews/gramineproject/gramine/1355#-NV_9KTz-6gUtdSQuI_I:-NVu2r2nEUgVKiCYoxez:b-ta834r) ([raw file](https://github.com/gramineproject/gramine/blob/72122616438a5141fa4a4530d303338bca76eae5/libos/include/arch/x86_64/gramine_entry_api.h#L75)):*
This is exactly what I proposed:

> or use a proper asm here (either separate file, or at least global scope `asm` statement).

The problem with it though is that it would complicate things a lot, as this is exposing Gramine API to outside users (like glibc) and for now it was simply a `.h` file. Making this a proper asm function would mean creating a `.o`/`.a` and somehow linking that into glibc.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1355)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski, @dimakuv, and @kailun-qin)

___
*[`libos/include/arch/x86_64/gramine_entry_api.h` line 75 at r1](https://reviewable.io/reviews/gramineproject/gramine/1355#-NV_9KTz-6gUtdSQuI_I:-NVuJ7Me09SS44hH605E:bmqyuv4) ([raw file](https://github.com/gramineproject/gramine/blob/72122616438a5141fa4a4530d303338bca76eae5/libos/include/arch/x86_64/gramine_entry_api.h#L75)):*

This is exactly what I proposed:

> or use a proper asm here (either separate file, or at least global scope `asm` statement).

The problem with it though is that it would complicate things a lot, as this is exposing Gramine API to outside users (like glibc) and for now it was simply a `.h` file. Making this a proper asm function would mean creating a `.o`/`.a` and somehow linking that into glibc.
</blockquote></details>

Ah, right. But that wouldn't be a problem if this was a global, inline assembly with local visibility? I.e. no additional `.o` files.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1355)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv, @kailun-qin, and @mkow)

___
*[`libos/include/arch/x86_64/gramine_entry_api.h` line 75 at r1](https://reviewable.io/reviews/gramineproject/gramine/1355#-NV_9KTz-6gUtdSQuI_I:-NVvi5prB8AcUeVw2txm:biwie4b) ([raw file](https://github.com/gramineproject/gramine/blob/72122616438a5141fa4a4530d303338bca76eae5/libos/include/arch/x86_64/gramine_entry_api.h#L75)):*

Ah, right. But that wouldn't be a problem if this was a global, inline assembly with local visibility? I.e. no additional `.o` files.
</blockquote></details>

I tried something like:
```
__asm__(
".pushsection .text\n"
".type gramine_call, @function\n"
"gramine_call:\n"
    "mov $" GRAMINE_XSTR(GRAMINE_CUSTOM_SYSCALL_NR) ", %eax\n"
    "GRAMINE_SYSCALL\n"
    "ret\n"
".popsection\n"
);
 
__attribute__((visibility("hidden")))
int gramine_call(int number, unsigned long arg1, unsigned long arg2);
```

and it works only in release builds. In debug builds it fails with:
```
errlist-data-gen.c: Assembler messages:
errlist-data-gen.c:482: Error: file number 1 already allocated
```
on building glibc. My suspicion is that there's something wrong with DWARF and this symbol (`gramine_call`) gets info it multiple times, while it still works correctly on C and ELF levels (where each `gramine_call` is indeed local to given module). Maybe. idk

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1355)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski, @kailun-qin, and @mkow)

___
*[`libos/include/arch/x86_64/gramine_entry_api.h` line 75 at r1](https://reviewable.io/reviews/gramineproject/gramine/1355#-NV_9KTz-6gUtdSQuI_I:-NW13_Io81IT5AwwTWDS:b-4vw3d0) ([raw file](https://github.com/gramineproject/gramine/blob/72122616438a5141fa4a4530d303338bca76eae5/libos/include/arch/x86_64/gramine_entry_api.h#L75)):*

I tried something like:
```
__asm__(
".pushsection .text\n"
".type gramine_call, @function\n"
"gramine_call:\n"
    "mov $" GRAMINE_XSTR(GRAMINE_CUSTOM_SYSCALL_NR) ", %eax\n"
    "GRAMINE_SYSCALL\n"
    "ret\n"
".popsection\n"
);
 
__attribute__((visibility("hidden")))
int gramine_call(int number, unsigned long arg1, unsigned long arg2);
```

and it works only in release builds. In debug builds it fails with:
```
errlist-data-gen.c: Assembler messages:
errlist-data-gen.c:482: Error: file number 1 already allocated
```
on building glibc. My suspicion is that there's something wrong with DWARF and this symbol (`gramine_call`) gets info it multiple times, while it still works correctly on C and ELF levels (where each `gramine_call` is indeed local to given module). Maybe. idk
</blockquote></details>

@mkow @boryspoplawski If all our solutions have limitations, why not go with my approach in #1354? It's a very simple approach, relying on `noinline` behavior of the compiler (to my knowledge, this is decently portable and achieves our purposes).

I think we're spending too much time on making this perfect, given that this feature is purely for running a couple tests in Gramine...

kailun-qin (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1355)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski, @dimakuv, and @mkow)

___
*[`libos/include/arch/x86_64/gramine_entry_api.h` line 75 at r1](https://reviewable.io/reviews/gramineproject/gramine/1355#-NV_9KTz-6gUtdSQuI_I:-NW17UQeEOzx0Gnkp2dw:b-yzaeip) ([raw file](https://github.com/gramineproject/gramine/blob/72122616438a5141fa4a4530d303338bca76eae5/libos/include/arch/x86_64/gramine_entry_api.h#L75)):*
> why not go with my approach in [#1354](https://github.com/gramineproject/gramine/pull/1354)

+1, looks like it's a simpler solution and should be fine w/ our current usage

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1355)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski, @dimakuv, and @kailun-qin)

___
*[`libos/include/arch/x86_64/gramine_entry_api.h` line 75 at r1](https://reviewable.io/reviews/gramineproject/gramine/1355#-NV_9KTz-6gUtdSQuI_I:-NW1elU14nUviyoDD8Em:b-w5sphp) ([raw file](https://github.com/gramineproject/gramine/blob/72122616438a5141fa4a4530d303338bca76eae5/libos/include/arch/x86_64/gramine_entry_api.h#L75)):*

> why not go with my approach in [#1354](https://github.com/gramineproject/gramine/pull/1354)

+1, looks like it's a simpler solution and should be fine w/ our current usage
</blockquote></details>

Because #1354 also doesn't solve the issue, it only masks it hoping that the optimizer won't be too smart.

We have an idea to test which should work: naked functions (`__attribute__((naked))`), just need to test it.

kailun-qin (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1355)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski, @dimakuv, and @mkow)

___
*[`libos/include/arch/x86_64/gramine_entry_api.h` line 75 at r1](https://reviewable.io/reviews/gramineproject/gramine/1355#-NV_9KTz-6gUtdSQuI_I:-NW1j89e94hpX643jdQD:b-qruz6p) ([raw file](https://github.com/gramineproject/gramine/blob/72122616438a5141fa4a4530d303338bca76eae5/libos/include/arch/x86_64/gramine_entry_api.h#L75)):*
> it only masks it hoping that the optimizer won't be too smart.

Ah OK, I see the problem. #1354 is based on the assumption that we'll always produce `call` instead of `jmp` w/ `__attribute__ ((noinline))`, but it's actually not guaranteed.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1355)* status: 0 of 1 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @kailun-qin, and @mkow)

___
*[`-- commits` line 11 at r2](https://reviewable.io/reviews/gramineproject/gramine/1355#-NWSSD771xB8ANrUn8ag):*




_Suggestion:_
```
- 7212261: [LibOS] Force `gramine_call` not to be inlined

  `gramine_call` is a custom interface to test Gramine internal functions.
  It uses Gramine syscall emulation and hence does not save/restore SEE
  registers (see "libos/src/arch/x86_64/syscallas.S" for more details).
  This turned out to be problematic, since test code calling
  `gramine_call` might assume SSE registers are not clobbered, since
  `gramine_call` is in reality just an inline asm block.
  This commit fixed that by making `gramine_call` a naked function, which
  cannot be inlined.
```
___
*[`libos/include/arch/x86_64/gramine_entry_api.h` line 75 at r1](https://reviewable.io/reviews/gramineproject/gramine/1355#-NV_9KTz-6gUtdSQuI_I:-NWSSa0q3KKbglL1VuAh:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/72122616438a5141fa4a4530d303338bca76eae5/libos/include/arch/x86_64/gramine_entry_api.h#L75)):*

> it only masks it hoping that the optimizer won't be too smart.

Ah OK, I see the problem. #1354 is based on the assumption that we'll always produce `call` instead of `jmp` w/ `__attribute__ ((noinline))`, but it's actually not guaranteed.
</blockquote></details>

Done.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1355)* status: 0 of 1 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @kailun-qin, and @mkow)

___
*[`-- commits` line 11 at r2](https://reviewable.io/reviews/gramineproject/gramine/1355#-NWSSD771xB8ANrUn8ag):*
I would also add:
```
and the compiler must assume SSE registers are clobbered.
```

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1355)* status: all files reviewed, 2 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @kailun-qin)

___
*[`-- commits` line 11 at r2](https://reviewable.io/reviews/gramineproject/gramine/1355#-NWSSD771xB8ANrUn8ag):*
> Force `gramine_call` not to be inlined

It's not actually about inlining itself, it was the inline asm which was incorrectly written, not the function. The no-inline version (https://github.com/gramineproject/gramine/pull/1354) is still incorrect, because the compiler is free to use SSE inside `gramine_call` around the inline asm.

So, I'd rather say: "Fix clobbers in `gramine_call`".

kailun-qin (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1355)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

___
*[`-- commits` line 11 at r2](https://reviewable.io/reviews/gramineproject/gramine/1355#-NWSSD771xB8ANrUn8ag):*
> Force `gramine_call` not to be inlined

but it can still be inlined w/ only `__attribute__((naked))`?

> and the compiler must assume SSE registers are clobbered.

`SSE registers` -> `the caller-saved registers`?

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1355)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[`libos/include/arch/x86_64/gramine_entry_api.h` line 75 at r1](https://reviewable.io/reviews/gramineproject/gramine/1355#-NV_9KTz-6gUtdSQuI_I:-NWbVyXu8A_k4C16L1v3:bkz6mhk) ([raw file](https://github.com/gramineproject/gramine/blob/72122616438a5141fa4a4530d303338bca76eae5/libos/include/arch/x86_64/gramine_entry_api.h#L75)):*

Done.
</blockquote></details>

Ok, I agree that my #1354 was just hiding the problem. I'm ok with this PR.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1355)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1355#-NWbXzar-xhzoEk9whRa:-NWbXzas4iDUAaNHXGpB:bjvkhyc) (no related file):*
@boryspoplawski Could you rebase this PR for final merge? With fixing the commit message.

___

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1355)* status: all files reviewed, 1 unresolved discussion (waiting on @mkow)

___
*[`-- commits` line 11 at r2](https://reviewable.io/reviews/gramineproject/gramine/1355#-NWSSD771xB8ANrUn8ag):*
> but it can still be inlined w/ only `__attribute__((naked))`?

No. It's literally an asm blob, which must be called.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1355)* status: :shipit: complete! all files reviewed, all discussions resolved

kailun-qin (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1355)* status: all files reviewed, 1 unresolved discussion (waiting on @boryspoplawski)

___
*[`-- commits` line 11 at r2](https://reviewable.io/reviews/gramineproject/gramine/1355#-NWSSD771xB8ANrUn8ag):*
> No. It's literally an asm blob, which must be called.

Thanks, I see -- it doesn't make any sense to be inlined (i.e., it's implicitly never-inlined).
___
*[`-- commits` line 5 at r3](https://reviewable.io/reviews/gramineproject/gramine/1355#-NWe2cXC3xZs0AD3zXch):*
SSE



_Code quote:_
```
SEE
```

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1355)* status: all files reviewed, 1 unresolved discussion (waiting on @kailun-qin)

___
*[`-- commits` line 5 at r3](https://reviewable.io/reviews/gramineproject/gramine/1355#-NWe2cXC3xZs0AD3zXch):*

SSE
</blockquote></details>

Done

kailun-qin (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1355)* status: all files reviewed, 1 unresolved discussion

kailun-qin (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1355)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #1379, opened 2023-06-01
[PAL/Linux-SGX] sgx_gdb: return const string for `str_ptrace_request()`

kailun-qin:
## Description of the changes 

An implicit conversion from const-qualified pointer types to non-const-qualified ones in C can generate a compilation warning. This happened for `str_ptrace_request()` when `DEBUG_GDB_PTRACE` was enabled, where "return discards `const` qualifier from pointer target type".

## How to test this PR? 

Manual testing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1379)

### DIFF
--- a/pal/src/host/linux-sgx/gdb_integration/sgx_gdb.c
+++ b/pal/src/host/linux-sgx/gdb_integration/sgx_gdb.c
@@ -52,7 +52,7 @@ static struct {
 } g_memdevs[32];
 
 #if DEBUG_GDB_PTRACE == 1
-static char* str_ptrace_request(gramine_ptrace_request request) {
+static const char* str_ptrace_request(gramine_ptrace_request request) {
     static char buf[50];
     int prev_errno;
 
@@ -95,7 +95,7 @@ static char* str_ptrace_request(gramine_ptrace_request request) {
     prev_errno = errno; /* snprintf can contaminate errno */
     snprintf(buf, sizeof(buf), "0x%x", request);
     errno = prev_errno;
-    return buf;
+    return (const char*)buf;
 }
 #endif
 


dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1379)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

___
*[`pal/src/host/linux-sgx/gdb_integration/sgx_gdb.c` line 98 at r1](https://reviewable.io/reviews/gramineproject/gramine/1379#-NWpoR4_2qUIRAvE_DtQ:-NWpoR4_2qUIRAvE_DtR:b-z1i5nh) ([raw file](https://github.com/gramineproject/gramine/blob/e9174f9b7fbf146c480fdcae22c2ab2c92128eea/pal/src/host/linux-sgx/gdb_integration/sgx_gdb.c#L98)):*
> ```C
>     snprintf(buf, sizeof(buf), "0x%x", request);
>     errno = prev_errno;
>     return (const char*)buf;
> ```

That's a bit wrong conversion, but since this is under the `DEBUG_GDB_PTRACE` macro (which is unset by default, and only used by hardcore SGX-GDB-plugin developers) and not part of core Gramine, I don't mind this hack.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1379)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`pal/src/host/linux-sgx/gdb_integration/sgx_gdb.c` line 98 at r1](https://reviewable.io/reviews/gramineproject/gramine/1379#-NWpoR4_2qUIRAvE_DtQ:-NX2x73Z8Zm8AHrdFTFz:bprhluq) ([raw file](https://github.com/gramineproject/gramine/blob/e9174f9b7fbf146c480fdcae22c2ab2c92128eea/pal/src/host/linux-sgx/gdb_integration/sgx_gdb.c#L98)):*

That's a bit wrong conversion, but since this is under the `DEBUG_GDB_PTRACE` macro (which is unset by default, and only used by hardcore SGX-GDB-plugin developers) and not part of core Gramine, I don't mind this hack.
</blockquote></details>

Why is it "wrong"? (and in what sense?)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1379)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`pal/src/host/linux-sgx/gdb_integration/sgx_gdb.c` line 98 at r1](https://reviewable.io/reviews/gramineproject/gramine/1379#-NWpoR4_2qUIRAvE_DtQ:-NX9KApI1Y6WvdTOyNOE:b-fqd9pv) ([raw file](https://github.com/gramineproject/gramine/blob/e9174f9b7fbf146c480fdcae22c2ab2c92128eea/pal/src/host/linux-sgx/gdb_integration/sgx_gdb.c#L98)):*

Why is it "wrong"? (and in what sense?)
</blockquote></details>

The `buf` variable is global, and a multi-threaded execution of GDB would "garble" this variable if two threads would call this func (with a `request` value not in the known-constant range).

Notice that `buf` is populated with a string representation of the `request` int value, via `snprintf()`. So basically this function returns a pointer that it is supposed to point to an immutable string, whereas in reality (in multi-threaded case), this string can surely be modified.

So technically this looks wrong. But as I said, iirc GDB always ptraces in single-threaded mode, and this is hidden behind the DEBUG macro, and this is purely for testing purposes. So I don't care.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1379)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`pal/src/host/linux-sgx/gdb_integration/sgx_gdb.c` line 98 at r1](https://reviewable.io/reviews/gramineproject/gramine/1379#-NWpoR4_2qUIRAvE_DtQ:-NXSBoKTEmgqUXAi0Mau:b-wjzlt9) ([raw file](https://github.com/gramineproject/gramine/blob/db98e155f9214eea43bf29b5eb17a47c642e9222/pal/src/host/linux-sgx/gdb_integration/sgx_gdb.c#L98)):*
> The `buf` variable is global, and a multi-threaded execution of GDB would "garble" this variable if two threads would call this func (with a `request` value not in the known-constant range).

What does this have to do with `const` specifier? `const` says that you cannot modify the variable through this pointer and nothing more.

dimakuv:
*[`pal/src/host/linux-sgx/gdb_integration/sgx_gdb.c` line 98 at r1](https://reviewable.io/reviews/gramineproject/gramine/1379#-NWpoR4_2qUIRAvE_DtQ:-NXUO_5K0GtQlJXceE-k:bxdxanr) ([raw file](https://github.com/gramineproject/gramine/blob/db98e155f9214eea43bf29b5eb17a47c642e9222/pal/src/host/linux-sgx/gdb_integration/sgx_gdb.c#L98)):*

> The `buf` variable is global, and a multi-threaded execution of GDB would "garble" this variable if two threads would call this func (with a `request` value not in the known-constant range).

What does this have to do with `const` specifier? `const` says that you cannot modify the variable through this pointer and nothing more.
</blockquote></details>

Ok, looks like you're right. My understanding was wrong.

Issue #513, opened 2022-04-10
The same MRENCLAVE for both trusted and protected files


AgentRX:
Hello,
My name is Andrei Pogoreltsev,
I'm CTO of Super Protocol Team and we glad to use Gramine in our project. Thank you for the product you're developing!

So, we want to protect our clients application code which run inside TEE (enclave). The TEE hardware is not owned by application owners. To do it we want to use protected files mechanism by this way:

**Prepare:**
1. We separate "base image" with some entrypoint and the "application". For example we can use Python base image and run /entrypoint/start.py script at start. Base image is no need to be protected and can be reusable by different applications.

2. The application owner combine base image and the application together to create Gramine manifest and get mrenclave. **Also the "/entrypoint/start.py" path is both trusted and protected file.** The main idea - it allows to check application before run and be sure nobody replaced the code because Gramine can check trusted files based on manifest hashes.

3. After it we can protect (encrypt by pf files mechanism) application using some encryption key.

So the idea is simple:
1. Separate protected/unprotected code
2. Prepare Gramine manifest for base image and not crypted application together
3. Encrypt application by some encryption key

Also, using the base image we can create different protected applications with **different** mrenclaves (see step 2)

To initialize Gramine PF encryption key we want to use Secret Provisioning mechanism which allows Gramine to get the key by connecting to our local server (which also runs inside enclave). The server attests enclave and send protection key matched to mrenclave.

**One of using mechanism example:**
1. First we run our local server
2. We attest the server remotely
3. Then all keys for different protected applications are transferred inside the server
4. We copy all base images and protected applications to the server side
5. Now we can run enclaves and send them keys using Secret Provisioning mechanism

**The problem**
The main problem is that Gramine doesn't check trusted file which also is protected:
https://github.com/gramineproject/gramine/blob/master/Pal/src/host/Linux-SGX/db_files.c#L112-L168

So in this case the intruder (for example hardware owner) can use his own malware to get other protected application encryption key by spoofing attack:

**Prepare:**
1. The intruder steals attacked application Gramine manifest because it's not encrypted
2. Then the intruder creates own protected application with the same base image BUT replace his own manifest by the attacked application
3. Then the intruder send the malware encryption key to the trusted server

**So now we have two protected applications with the same not encrypted base image and the same manifest but the applications are different**

**Using:**
1. Now the intruder runs his malware application by standard usage mechanism
[...]

PR #1390, opened 2023-06-14
RFC: [PAL/Linux-SGX] Add encrypted-and-trusted files

dimakuv:
## Description of the changes 

Previously, Gramine allowed encrypted files (aka protected files), which do not have a corresponding hash in the manifest file and thus their contents are not reflected in MRENCLAVE. Also, Gramine had trusted files whose contents are reflected in MRENCLAVE but these files are not encrypted.

This PR allows to combine two functionalities and get encrypted-and-trusted files as a result. The manifest syntax doesn't change and looks like this:
```toml
  fs.mounts = [
    { type = "encrypted", path = "/enc_file", uri = "file:enc_file" },
  ]
  sgx.trusted_files = [
    "file:enc_file", # or with {uri = ..., sha256 = ...} syntax
  ]
```

Note that previously Gramine had unintuitive behavior of *ignoring* the entry in `sgx.trusted_files` if it was already marked as encrypted file.

### Discussion

See https://github.com/gramineproject/gramine/issues/513.

Generally, there were three ideas for encrypted + hashed files:
1. Hash the file first, then encrypt it.
   - This seems to have weak security guarantees (e.g., weak resistance against known-plaintext attacks).
   - Implementation would be complicated, because it "reverses" the current Gramine flow of "encrypted at LibOS layer -> hashed at PAL layer".
   - Manifest syntax could probably stay the same as now (reusing `fs.mounts` and `sgx.trusted_files`).
2. Encrypt the file, then hash the first 4KB of the file.
   - This has high security guarantees, and also fast -- no need to hash and verify the whole file.
   - Implementation would be complicated, because we would introduce a new hashing logic, completely bypassing the existing PAL hashing logic.
   - Manifest syntax needs to be changed, and it would be non-trivial (e.g. `fs.mounts = { { type = "encrypted", sha256 = ... }`). I tried it as my first attempt, but it quickly got ugly as this `sha256` field belongs to the FS mount, not to the file.
3. Encrypt the file, then hash the whole file.
   - This has high security guarantees, but slower than 2. On the other hand, the overhead should be negligible as hashed files are typically read completely anyway.
   - Implementation is trivial, see this PR.
   - Manifest syntax stays the same as now, and it is actually logical.

This PR implements idea 3.

### TODOs

- [ ] Fix documentation

## How to test this PR? 

Two tests are added.

### DIFF
--- a/libos/test/regression/encrypt_file.sh
+++ b/libos/test/regression/encrypt_file.sh
@@ -1,0 +1,27 @@
+#!/bin/sh
+
+# Usage:
+#
+#    encrypt_file file_to_encrypt file_encrypted
+#
+# Encrypts the file using `gramine-sgx-pf-crypt encrypt`.
+
+set -e
+
+SRC_FILE="$1"
+DEST_FILE="$2"
+
+# The hard-coded key must correspond to `fs.insecure__keys.default` in test manifests
+# (e.g. in `helloworld_enc.manifest.template`). Note that POSIX printf doesn't understand hex so we
+# must use octal.
+WRAPKEY_FILE=$(mktemp /tmp/gramine-wrap-key.XXXXXX)
+printf "\377\356\335\314\273\252\231\210" > ${WRAPKEY_FILE}
+printf "\167\146\125\104\63\42\21\0"     >> ${WRAPKEY_FILE}
+
+gramine-sgx-pf-crypt encrypt --input ${SRC_FILE} --output ${DEST_FILE} --wrap-key ${WRAPKEY_FILE}
+
+# Copy mode bits from input file; required to e.g. mark the output file as executable (Gramine
+# enforces the check for executable bit on entrypoint files).
+chmod --reference=${SRC_FILE} ${DEST_FILE}
+
+rm ${WRAPKEY_FILE}
--- a/libos/test/regression/helloworld_enc.manifest.template
+++ b/libos/test/regression/helloworld_enc.manifest.template
@@ -1,0 +1,20 @@
+loader.entrypoint = "file:{{ gramine.libos }}"
+libos.entrypoint = "{{ entrypoint }}"
+
+loader.env.LD_LIBRARY_PATH = "/lib"
+
+fs.mounts = [
+  { path = "/lib", uri = "file:{{ gramine.runtimedir(libc) }}" },
+  { type = "encrypted", path = "/{{ entrypoint }}", uri = "file:{{ binary_dir }}/{{ entrypoint }}", key_name = "default" },
+]
+
+sgx.debug = true
+sgx.edmm_enable = {{ 'true' if env.get('EDMM', '0') == '1' else 'false' }}
+
+sgx.trusted_files = [
+  "file:{{ gramine.libos }}",
+  "file:{{ gramine.runtimedir(libc) }}/",
+  "file:{{ binary_dir }}/{{ entrypoint }}",
+]
+
+fs.insecure__keys.default = "ffeeddccbbaa99887766554433221100"
--- a/libos/test/regression/helloworld_enc_fail.manifest.template
+++ b/libos/test/regression/helloworld_enc_fail.manifest.template
@@ -1,0 +1,26 @@
+{% set entrypoint = "helloworld_enc" -%}
+
+loader.entrypoint = "file:{{ gramine.libos }}"
+libos.entrypoint = "{{ entrypoint }}"
+
+loader.log_level = "warning"
+
+loader.env.LD_LIBRARY_PATH = "/lib"
+
+fs.mounts = [
+  { path = "/lib", uri = "file:{{ gramine.runtimedir(libc) }}" },
+  { type = "encrypted", path = "/{{ entrypoint }}", uri = "file:{{ binary_dir }}/{{ entrypoint }}", key_name = "default" },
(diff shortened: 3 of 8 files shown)


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1390)* status: 0 of 8 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "DONTMERGE" found in commit messages' one-liners

___
*[`libos/test/regression/encrypt_file.sh` line 24 at r1](https://reviewable.io/reviews/gramineproject/gramine/1390#-NXtaduuCVdOoz8OzEvc:-NXtaduuCVdOoz8OzEvd:b-6pk887) ([raw file](https://github.com/gramineproject/gramine/blob/fbc130bda9669cdc424e50364160cdd34b0548ab/libos/test/regression/encrypt_file.sh#L24)):*
> ```Shell
> 
> # Copy mode bits from input file; required to e.g. mark the output file as executable (Gramine
> # enforces the check for executable bit on entrypoint files).
> ```

FYI: I hit this case in Gramine: https://github.com/gramineproject/gramine/blob/ab03e6f0c3457758984455212e5f2cb7171c1ee9/libos/src/bookkeep/libos_handle.c#L53-L56

I was contemplating removing this check (why wouldn't Gramine disallow specifying non-executable-bit files as entrypoints?), but in the end I think this is meaningful. Though this may be surprising to end users.
___
*[`libos/test/regression/helloworld_enc.manifest.template` line 8 at r1](https://reviewable.io/reviews/gramineproject/gramine/1390#-NXtb5ex2xQNj_ArwGkS:-NXtb5ex2xQNj_ArwGkT:b-tmmux9) ([raw file](https://github.com/gramineproject/gramine/blob/fbc130bda9669cdc424e50364160cdd34b0548ab/libos/test/regression/helloworld_enc.manifest.template#L8)):*
> ```
> fs.mounts = [
>   { path = "/lib", uri = "file:{{ gramine.runtimedir(libc) }}" },
>   { type = "encrypted", path = "/{{ entrypoint }}", uri = "file:{{ binary_dir }}/{{ entrypoint }}", key_name = "default" },
> ```

FYI: Note that I can't fix this over-100-chars-per-line issue because TOML syntax doesn't allow this. Here and below too.
___
*[`libos/test/regression/test_libos.py` line 86 at r1](https://reviewable.io/reviews/gramineproject/gramine/1390#-NXtbEcE9EvCgog60Ni-:-NXtbEcE9EvCgog60Ni0:b-jl7ie9) ([raw file](https://github.com/gramineproject/gramine/blob/fbc130bda9669cdc424e50364160cdd34b0548ab/libos/test/regression/test_libos.py#L86)):*
> ```Python
>         self.assertIn('Hello world!', stdout)
> 
>     @unittest.skipUnless(HAS_SGX, 'Required tool gramine-sgx-pf-crypt only available with SGX')
> ```

FYI: This is actually an artificial limitation. It was our mistake to install this tool only when SGX was specified in Meson. And the name is also wrong, it shouldn't have `-sgx-`.

Maybe I'll fix this in another PR, and then we can remove this `skipUnless` here.
___
*[`pal/src/host/linux-sgx/pal_files.c` line 99 at r1](https://reviewable.io/reviews/gramineproject/gramine/1390#-NXtbl6O3_59-A55_vGh:-NXtbl6O3_59-A55_vGi:b-5q5e0l) ([raw file](https://github.com/gramineproject/gramine/blob/fbc130bda9669cdc424e50364160cdd34b0548ab/pal/src/host/linux-sgx/pal_files.c#L99)):*
> ```C
>      * Allowed and trusted files are affected by the `file_check_policy` manifest option, whereas
>      * protected and protected-and-trusted files are not.
>      */
> ```

FYI: I admit that our internal implementation of protected/trusted/allowed files is a mess... Especially with regard to e.g. `file_check_policy` manifest option.

But this is a broader discussion about files -- we need a major overhaul of all that stuff, including changes in manifest syntax.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1390)* status: 0 of 8 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "DONTMERGE" found in commit messages' one-liners

___
*[`pal/src/host/linux-sgx/pal_files.c` line 139 at r1](https://reviewable.io/reviews/gramineproject/gramine/1390#-NXtcMO969wx_UZpxNW7:-NXtcMO969wx_UZpxNW8:b4kzvpt) ([raw file](https://github.com/gramineproject/gramine/blob/fbc130bda9669cdc424e50364160cdd34b0548ab/pal/src/host/linux-sgx/pal_files.c#L139)):*
> ```C
>      * always opens protected files in this mode; it's safe to ignore RDWR in this case because
>      * LibOS controls the permissions of protected files, and also because `file_write()` and
>      * `file_map()` PAL callbacks won't allow writing on protected-and-trusted files */
> ```

FYI: This is the problematic place that necessitated this change: https://github.com/gramineproject/gramine/blob/ab03e6f0c3457758984455212e5f2cb7171c1ee9/libos/src/fs/libos_fs_encrypted.c#L159-L160

PR #1412, opened 2023-06-22
[LibOS] Allow creation of encryption-key pseudo-files

dimakuv:
## Description of the changes 

Previously, Gramine disallowed creation of pseudo-files under `/dev/attestation/keys/[KEY_NAME]`, if the corresponding key name was not specified as `fs.insecure__keys.[KEY_NAME]` in the manifest. This happened because Gramine's pseudo FS didn't have the concept of file creation (all files were assumed to be created at init time).

This commit fixes it by adding the `creat()` callback in pseudo FS. Currently, only the `/dev/attestation/keys/` dir implements this callback. A LibOS test `keys` is updated.

Reported by @mythi. Thanks to him!

Notice that our documentation implies that such key files **can be created**. So this was a bug in Gramine, and no changes to the documentation are required: https://gramine.readthedocs.io/en/stable/manifest-syntax.html#encrypted-files.

## How to test this PR? 

Added a LibOS regression test.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1412)

### DIFF
--- a/libos/include/libos_fs_pseudo.h
+++ b/libos/include/libos_fs_pseudo.h
@@ -61,7 +61,8 @@ struct libos_dev_ops {
     int (*truncate)(struct libos_handle* hdl, uint64_t len);
 };
 
-#define PSEUDO_PERM_DIR     PERM_r_xr_xr_x  /* default for directories */
+#define PSEUDO_PERM_DIR_R   PERM_r_xr_xr_x  /* default for directories */
+#define PSEUDO_PERM_DIR_RW  PERM_rwxrwxrwx
 #define PSEUDO_PERM_LINK    PERM_rwxrwxrwx  /* default for links */
 #define PSEUDO_PERM_FILE_R  PERM_r__r__r__  /* default for all other files */
 #define PSEUDO_PERM_FILE_RW PERM_rw_rw_rw_
@@ -110,6 +111,9 @@ struct pseudo_node {
 
     /* Retrieves all file names for this node. Works the same as `readdir`. */
     int (*list_names)(struct libos_dentry* parent, readdir_callback_t callback, void* arg);
+
+    /* Creates a file with a given name. */
+    int (*creat)(struct libos_dentry* parent, const char* name, struct pseudo_node** out_node);
 
     /* File permissions. See `PSEUDO_PERM_*` above for defaults. */
     mode_t perm;
--- a/libos/src/fs/dev/attestation.c
+++ b/libos/src/fs/dev/attestation.c
@@ -323,6 +323,32 @@ static int key_save(struct libos_dentry* dent, const char* data, size_t size) {
     return 0;
 }
 
+static int keys_creat(struct libos_dentry* parent, const char* name,
+                      struct pseudo_node** out_node) {
+    assert(parent && parent->inode && parent->inode->data && strcmp(parent->name, "keys") == 0);
+
+    if (name[0] == '_') {
+        /* key names beginning with `_` are special keys provided by Gramine; cannot be created */
+        return -EACCES;
+    }
+
+    struct libos_encrypted_files_key* encrypted_files_key;
+    int ret = get_or_create_encrypted_files_key(name, &encrypted_files_key);
+    if (ret < 0) {
+        log_error("Cannot create encryption key `%s`", name);
+        return ret;
+    }
+
+    /* mount the newly created key over other /dev/attestation/keys/ */
+    struct pseudo_node* keys = parent->inode->data;
+    struct pseudo_node* key = pseudo_add_str(keys, name, &key_load);
+    key->perm = PSEUDO_PERM_FILE_RW;
+    key->str.save = &key_save;
+
+    *out_node = key;
+    return 0;
+}
+
 static int init_sgx_attestation(struct pseudo_node* attestation, struct pseudo_node* keys) {
     if (strcmp(g_pal_public_state->host_type, "Linux-SGX"))
         return 0;
@@ -376,6 +402,9 @@ int init_attestation(struct pseudo_node* dev) {
     struct pseudo_node* attestation = pseudo_add_dir(dev, "attestation");
 
     struct pseudo_node* keys = pseudo_add_dir(attestation, "keys");
+    keys->creat = &keys_creat;
+    keys->perm = PSEUDO_PERM_DIR_RW;
+
     struct pseudo_node* key = pseudo_add_str(keys, /*name=*/NULL, &key_load);
     key->name_exists = &key_name_exists;
     key->list_names = &key_list_names;
--- a/libos/src/fs/libos_fs_pseudo.c
+++ b/libos/src/fs/libos_fs_pseudo.c
@@ -158,6 +158,40 @@ static int pseudo_open(struct libos_handle* hdl, struct libos_dentry* dent, int 
     return 0;
 }
 
+static int pseudo_creat(struct libos_handle* hdl, struct libos_dentry* dent, int flags,
+                        mode_t perm) {
+    assert(locked(&g_dcache_lock));
+    assert(!dent->inode);
+
+    __UNUSED(flags);
+    __UNUSED(perm);
+
+    if (!dent->parent)
+        return -ENOENT;
+
+    struct pseudo_node* parent_node = pseudo_find(dent->parent);
+    if (!parent_node)
+        return -ENOENT;
+
+    if (!parent_node->creat)
+        return -EACCES;
+
+    /* note that we currently never unlink created pseudo-files, for simplicity of implementation */
+    struct pseudo_node* created_node;
+    int ret = parent_node->creat(dent->parent, dent->name, &created_node);
+    if (ret < 0)
+        return ret;
+
+    struct libos_inode* inode = get_new_inode(dent->mount, TYPE_STR, created_node->perm);
+    if (!inode)
+        return -ENOMEM;
+
+    inode->data = created_node;
+    dent->inode = inode;
+
+    return pseudo_open(hdl, dent, /*flags=*/0);
+}
+
 static int pseudo_lookup(struct libos_dentry* dent) {
     assert(locked(&g_dcache_lock));
     assert(!dent->inode);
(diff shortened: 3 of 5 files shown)


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1412)* status: 0 of 4 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1412#-NYYNOx12kNm_dQmTkKO:-NYYNOx12kNm_dQmTkKP:b-uhcdwj) (no related file):*
Without this PR, my modified `keys.c` test fails like this:
```
$ gramine-direct keys
keys: error at open(NONEXISTING_KEY_PATH, O_WRONLY | O_CREAT | O_TRUNC, 0666) (line 121): Permission denied: Permission denied
```

___

kailun-qin (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1412)* status: all files reviewed, 4 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel)

mythi (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1412)* status: all files reviewed, 4 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1412#-NYYLao_Bf_IvabA5yMG:-NaBt9IA8W2s_yNDbfhC:b-1phrc2) (no related file):*

@mythi Can you check this PR on your workload?
</blockquote></details>

I can confirm it works for me. I needed to add `.create(true)` to my rust OpenOptions so perhaps `O_CREAT` is needed for https://github.com/gramineproject/gramine/blob/51d915ded30aa2ca55ad5c2e86075e95fe938ca2/tools/sgx/ra-tls/secret_prov_attest.c#L423 too?

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1412)* status: 4 of 5 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1412#-NYYLao_Bf_IvabA5yMG:-NaC3tnn0Y0zn0Cd2EM2:b-mhvfuo) (no related file):*
<details><summary><i>Previously, mythi (Mikko Ylinen) wrote…</i></summary><blockquote>

I can confirm it works for me. I needed to add `.create(true)` to my rust OpenOptions so perhaps `O_CREAT` is needed for https://github.com/gramineproject/gramine/blob/51d915ded30aa2ca55ad5c2e86075e95fe938ca2/tools/sgx/ra-tls/secret_prov_attest.c#L423 too?
</blockquote></details>

Done. Good point. I also tested that it works by specifying my own key name in the manifest template for `CI-Examples/ra-tls-secret-prov/secret_prov_pf`.

___

mythi (commented):
Looks good to me, thanks for the fix!

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1412#-:-NaC5AIf5s6rgAwx4iOB:bwi80sq)* status: 4 of 5 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners

kailun-qin (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1412)* status: all files reviewed, 3 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners

kailun-qin (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1412)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`-- commits` line 6 at r3](https://reviewable.io/reviews/gramineproject/gramine/1412#-NbmJxbM24jSEAGBHmvU):*
Looking at this PR again -- I have a question regarding the below statement and the context/intention of this PR.

If the `key_name` was specified in the mount params of encrypted files (or default to the name `default`, pls see: https://gramine.readthedocs.io/en/latest/manifest-syntax.html#encrypted-files) which is the common usage of encryption keys, then even w/o `fs.insecure__keys.[KEY_NAME]`, shouldn't the corresponding keys be created during `chroot_encrypted_mount()` anyway (pls see:  https://github.com/gramineproject/gramine/blob/92636ae0bc8187d48dd40536d2a1901cb8eb6cd6/libos/src/fs/chroot/encrypted.c#L63)?

In this case, I gueses nothing will fail w/ the corresponding keys? I.e., why would we need to support creating (arbitrary) `/dev/attestation/keys/[KEY_NAME]`if it's not specified in encrypted files?



_Code quote:_
```
  Previously, Gramine disallowed creation of pseudo-files under
  `/dev/attestation/keys/[KEY_NAME]`, if the corresponding key name was
  not specified as `fs.insecure__keys.[KEY_NAME]` in the manifest. This
```

mythi (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1412)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

___
*[`-- commits` line 6 at r3](https://reviewable.io/reviews/gramineproject/gramine/1412#-NbmJxbM24jSEAGBHmvU):*
> shouldn't the corresponding keys be created during `chroot_encrypted_mount()` anyway (pls see:

To me it looks this just adds the key name to the list but the key does not get set (`is_set = false`).

> I.e., why would we need to support creating (arbitrary) `/dev/attestation/keys/[KEY_NAME]`if it's not specified in encrypted files?

This PR fixes the case where `[KEY_NAME]` is used as a `key_name` value as an encrypted files parameter but is  *not* added to `fs.insecure__keys.[KEY_NAME]` (nor is a special key/`default`). The use case is valid: if I set `loader.env.SECRET_PROVISION_SET_KEY=mythi_key` that key should be usable as `key_name` for the encrypted files (i.e., after the pseudo fs file creation `mythi_key`can be found from the list and `is_set = true`) .

mythi (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1412)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

___
*[`-- commits` line 6 at r3](https://reviewable.io/reviews/gramineproject/gramine/1412#-NbmJxbM24jSEAGBHmvU):*
<details><summary><i>Previously, mythi (Mikko Ylinen) wrote…</i></summary><blockquote>

> shouldn't the corresponding keys be created during `chroot_encrypted_mount()` anyway (pls see:

To me it looks this just adds the key name to the list but the key does not get set (`is_set = false`).

> I.e., why would we need to support creating (arbitrary) `/dev/attestation/keys/[KEY_NAME]`if it's not specified in encrypted files?

This PR fixes the case where `[KEY_NAME]` is used as a `key_name` value as an encrypted files parameter but is  *not* added to `fs.insecure__keys.[KEY_NAME]` (nor is a special key/`default`). The use case is valid: if I set `loader.env.SECRET_PROVISION_SET_KEY=mythi_key` that key should be usable as `key_name` for the encrypted files (i.e., after the pseudo fs file creation `mythi_key`can be found from the list and `is_set = true`) .
</blockquote></details>

My earlier comment does not make sense and @kailun-qin is correct. The original problem triggers only if `key_name` is not specified anywhere in the manifest. I guess I was blindly debugging why my `LD_PRELOAD`lib fails to write to `/dev/attestation/keys/key_name` that I missed that I did not have it specified for any of my mounts.

This opens the question is there a use case to have `/dev/attestation/keys/` writable outside of encrypted files which would justify this PR.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1412)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

___
*[`-- commits` line 6 at r3](https://reviewable.io/reviews/gramineproject/gramine/1412#-NbmJxbM24jSEAGBHmvU):*
<details><summary><i>Previously, mythi (Mikko Ylinen) wrote…</i></summary><blockquote>

My earlier comment does not make sense and @kailun-qin is correct. The original problem triggers only if `key_name` is not specified anywhere in the manifest. I guess I was blindly debugging why my `LD_PRELOAD`lib fails to write to `/dev/attestation/keys/key_name` that I missed that I did not have it specified for any of my mounts.

This opens the question is there a use case to have `/dev/attestation/keys/` writable outside of encrypted files which would justify this PR.
</blockquote></details>

Thanks @kailun-qin for raising this point.

Yes, Kailun is right that if the key **was specified** anywhere in the manifest file, then it will be visible and writable.

Given that this PR was created in request from @mythi, and now it turned out to be a bug in his manifest file, I'm leaning towards closing this PR. In other words, I feel like we don't have a valid use case to create keys outside of encrypted files.

If we'll ever have the need for such unspecified-in-manifest-keys in the future, we can resurrect this PR.

dimakuv:
There seems to be no valid use case for such unspecified-in-manifest `/dev/attestation/` keys, so I'm closing this PR. Maybe we will resurrect it in the future.

PR #1400, opened 2023-06-16
[Meson] Allow building with `-Dglibc=disabled`

dimakuv:
## Description of the changes 

Previously mbedTLS and RA-TLS libs depended on Glibc being enabled (even if building on musl-only OS distro like Alpine). With this commit, if Glibc is disabled, then mbedTLS and RA-TLS installation scripts install musl-linked versions of these libs.

Fixes #1399. Closes #1417.

## How to test this PR? 

Manually on Alpine. Also, CI (glibc-based) must not break.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1400)

PR #1417, opened 2023-06-23
[Meson,CI-Examples] Don't create symlinks in Gramine runtime dir

### LINKED ISSUE #1399: [Meson] Build fails when `meson setup -Dglibc=disabled`

This is because of the following snippets:

- https://github.com/gramineproject/gramine/blob/514df189a6867c92dac83923649cd2a124866e62/subprojects/packagefiles/mbedtls/meson.build#L58-L63

- https://github.com/gramineproject/gramine/blob/514df189a6867c92dac83923649cd2a124866e62/tools/sgx/ra-tls/meson.build#L31-L34 (and several more in this file)

### Steps to reproduce

This was detected on Alpine 3.18.0, on which there is no Glibc (https://ariadne.space/2021/08/26/there-is-no-such-thing-as-a-glibc-based-alpine-image/). So we have no choice but to use `-Dglibc=disabled`.

And anyway, it would be totally misleading for Gramine to have Glibc-based RA-TLS and mbedTLS libs by default on something like Alpine OS.

### Expected results

Bla

### Actual results

Bla

### Gramine commit hash

514df189a6867c92dac83923649cd2a124866e62

dimakuv:
## Description of the changes 

Gramine runtime dir is used for libc libraries (glibc or musl). Previously, we also added symlinks to Gramine-specific mbedTLS, RA-TLS and SecretProv libraries into this runtime dir. However, these latter libraries are built using the toolchain of the OS distro and thus may have a different libc than specified in the Gramine runtime dir. Therefore, the two kinds of libraries should be separated, and no symlinks should be created.

This is a *breaking* change for applications, because if the app uses Gramine-specific mbedTLS, RA-TLS or SecretProv libs it is required to add corresponding entries in the manifest. This commit updates our CI examples accordingly.

Fixes #1399. Closes #1400.

## How to test this PR? 

CI.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1417)

### DIFF
--- a/CI-Examples/ra-tls-mbedtls/Makefile
+++ b/CI-Examples/ra-tls-mbedtls/Makefile
@@ -57,6 +57,7 @@ server.manifest: server.manifest.template
 	gramine-manifest \
 		-Dlog_level=$(GRAMINE_LOG_LEVEL) \
 		-Darch_libdir=$(ARCH_LIBDIR) \
+		-Dmbedtls_gramine_dir=$(shell pkg-config --variable=libdir mbedtls_gramine) \
 		-Dra_type=$(RA_TYPE) \
 		-Dra_client_spid=$(RA_CLIENT_SPID) \
 		-Dra_client_linkable=$(RA_CLIENT_LINKABLE) \
@@ -79,6 +80,7 @@ client_dcap.manifest: client.manifest.template
 	gramine-manifest \
 		-Dlog_level=$(GRAMINE_LOG_LEVEL) \
 		-Darch_libdir=$(ARCH_LIBDIR) \
+		-Dmbedtls_gramine_dir=$(shell pkg-config --variable=libdir mbedtls_gramine) \
 		$< >$@
 
 client_dcap.manifest.sgx client_dcap.sig: sgx_sign_client_dcap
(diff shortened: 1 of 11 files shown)


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1417)* status: 0 of 11 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1417#-NYce8CD3TW9MyVyzpOH:-NYce8CD3TW9MyVyzpOI:bmc5fgs) (no related file):*
This PR is an alternative to https://github.com/gramineproject/gramine/pull/1400. This PR brings us closer to Alpine compatibility (in the sense of building Gramine on Alpine distro).

That PR simply assumed in which Gramine-runtime-dir to symlink mbedTLS, RA-TLS and SecretProv libs based on the `-Dglibc=... -Dmusl=...` meson options.

This PR doesn't assume anything about the OS distro toolchain used for building these libs, and instead removes the symlinks from Gramine-runtime-dir. The rationale is that these libs are conceptually different from Glibc/Musl libs, so they shouldn't be in the same directory.

___

dimakuv:
Superseded by #1427

PR #1414, opened 2023-06-23
[Meson] Replace `combo` with `feature` where possible

dimakuv:
## Description of the changes 

The `combo` options were left from the times when Meson was v0.47 (default on Ubuntu 18.04). Since those times, we moved to Meson at least v0.56 which has better-suited `feature` options.

## How to test this PR? 

CI.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1414)

PR #1420, opened 2023-06-26
[meson,tools] Introduce `epid` Meson option

dimakuv:
## Description of the changes 

Previously, Gramine always built EPID-based tools (e.g. IAS ones) and libraries (RA-TLS, SecreProv). However, in many deployments EPID is not required at all. This commit introduces the `epid` Meson option; `-Depid=disabled` can be used to disable EPID tools and libs. To preserve backward-compatibility, this option is enabled by default.

Also, if both EPID and DCAP options are disabled, then more tools and libs are not built (e.g. generic RA-TLS libs).

This is yet another attempt/proposal at circumventing the problem of `Dglibc=disabled`. It circumvents this problem in a hopefully nicer way, by removing the need for RA-TLS and SecretProv libraries inside Gramine runtime dir when the user builds Gramine without SGX attestation at all: `-Depid=disabled -Ddcap=disabled`.

See my other attempts:
- #1400 
- #1417 

Also read #1399 for context.

## How to test this PR? 

CI tests the backwards-compatibility.

Manual test is smth like this, to generate the SGX-attestation-less version of Gramine:
```
meson setup build/ -Dskeleton=enabled -Ddirect=enabled -Dsgx=enabled \
    -Depid=disabled -Ddcap=disabled -Dtests=disabled
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1420)

### DIFF
--- a/Documentation/devel/building.rst
+++ b/Documentation/devel/building.rst
@@ -192,7 +192,7 @@ Additional build options
 
   .. note::
      EPID version of RA-TLS library (``ra_tls_verify_epid.so``) is built by
-     default.
+     default. You can disable the build via  :command:`meson -Depid=disabled`.
 
 - To create a debug build, run :command:`meson --buildtype=debug`. This adds
   debug symbols in all Gramine components, builds them without optimizations,
--- a/meson.build
+++ b/meson.build
@@ -32,6 +32,7 @@ direct = get_option('direct') == 'enabled'
 sgx = get_option('sgx') == 'enabled'
 skeleton = get_option('skeleton') == 'enabled'
 
+epid = get_option('epid') == 'enabled'
 dcap = get_option('dcap') == 'enabled'
 ubsan = get_option('ubsan') == 'enabled'
 asan = get_option('asan') == 'enabled'
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -14,8 +14,11 @@ option('musl', type: 'combo', choices: ['disabled', 'enabled'],
 option('tests', type: 'combo', choices: ['disabled', 'enabled'],
     description: 'Build test binaries')
 
+option('epid', type: 'combo', choices: ['disabled', 'enabled'],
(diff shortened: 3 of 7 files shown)


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1420)* status: 0 of 7 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

___
*[`subprojects/packagefiles/mbedtls/meson.build` line 58 at r1](https://reviewable.io/reviews/gramineproject/gramine/1420#-NYrhBtk2vr4m1_6erRi:-NYrhBtk2vr4m1_6erRj:b-sh708h) ([raw file](https://github.com/gramineproject/gramine/blob/1376f02d74268d5c090f974338a7f16d4c6c2e3b/subprojects/packagefiles/mbedtls/meson.build#L58)):*
> ```Meson
> )
> 
> # install mbedTLS symlinks into Gramine runtime dir only if attestation was requested
> ```

I was hesitant what else to not install. Probably we shouldn't install mbedTLS headers too (see lines below)?
___
*[`subprojects/packagefiles/mbedtls/meson_options.txt` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/1420#-NYrh1-CEzvlFKsEqIsS:-NYrh1-CEzvlFKsEqIsT:b81r0e2) ([raw file](https://github.com/gramineproject/gramine/blob/1376f02d74268d5c090f974338a7f16d4c6c2e3b/subprojects/packagefiles/mbedtls/meson_options.txt#L2)):*
> ```Meson
> # TODO: after deprecating 18.04/bionic, change these to type: 'feature'
> #       see also top-level meson_options.txt
> ```

See https://github.com/gramineproject/gramine/pull/1414. If that PR is merged first, this will need to be changed to `feature`.

dimakuv:
Superseded by #1427

PR #1423, opened 2023-06-27
[LibOS] Check existence of file during `chroot_stat()`

### LINKED ISSUE #1422: Problem reading the FS using rust code

Description of the problem:
I put the rust program into the docker container upgraded based on the official gramine docker container and run it with sgx. The program loops to detect whether a file under the file path `DEFAULT_FILE_SAVE_PATH` exists, and `DEFAULT_FILE_SAVE_PATH` is mapped to the root directory of the host by docker on a certain path `A`, when an untrusted program on the host computer deletes a file in path `A` , my rust program running in gramine can still detect the existence of the file, but it cannot open the file ( Error No such file or directory (os error 2))

### Steps to reproduce

```rust
const DEFAULT_FILE_SAVE_PATH: &str = "/sgx/";
	let mut file_path = DEFAULT_FILE_SAVE_PATH.to_string() + &bs58::encode(&request_body.peer_id).into_string();
	let mut duplicate_file = 0;
	loop {
		let tmp_file_path = file_path.clone() + &format!(".{}", duplicate_file );
		let path = std::path::Path::new(&tmp_file_path);
		if path.try_exists().unwrap() && path.is_file() {
			warn!("File {:?} already exists ,it means rotator still not finished",&tmp_file_path);
			duplicate_file+=1;
			let tmp_file= match File::open(path){
				Ok(f) => {f}
				Err(e) => {
					error!("error when read file {:?},because {:?}",&tmp_file_path,e.to_string());
					continue
				}
			};
			warn!("the file {:?} in gramine system length is {}",&tmp_file_path,tmp_file.metadata().unwrap().len());
			continue
		}
		break
	};
```
You can try to run this code, after adding a file such as "/sgx/12D3KooWQHuPnuXCqR3c5JhfsM2K6rQUKXwWdoeBuPNV3kQjj3mr.0" on the host host delete (by the way "/sgx/12D3KooWQHuPnuXCqR3c5JhfsM2K6rQUKXwWdoeBuPNV3kQ jj3mr.0" file is generated from within sgx), there will be an error,The following is the content of the manifest file.

```sh
[fs]
mounts = [
  { path = "{{ arch_libdir }}", uri = "file:{{ arch_libdir }}" },
  { path = "/usr{{ arch_libdir }}", uri = "file:/usr{{ arch_libdir }}" },
  { path = "/sgx", uri = "file:/sgx" },
  { path = "/kaleido", uri = "file:/kaleido"},
  { path = "/kaleido/encrypted", uri = "file:/kaleido/encrypted", type = "encrypted", key_name = "_sgx_mrsigner" },
  { path = "/etc", uri = "file:/etc", type = "chroot"},
]

[sgx]
debug = true
nonpie_binary = true
enclave_size = "2G"
max_threads = 128
trusted_files = [
  "file:{{ gramine.libos }}",
  "file:{{ self_exe }}",
  "file:/usr{{ arch_libdir }}/",
  "file:{{ arch_libdir }}/",
]
allowed_files = [
  "file:/kaleido",
  "file:/sgx",
  "file:/etc/hosts",
  "file:/etc/resolv.conf",
]
```

### Expected results

```
[2023-06-27T02:25:19Z WARN  xxx::api] File "/sgx/12D3KooWQHuPnuXCqR3c5JhfsM2K6rQUKXwWdoeBuPNV3kQjj3mr.0" already exists ,it means rotator still not finished
[2023-06-27T02:25:19Z WARN xxx::api] the file "/sgx/12D3KooWQHuPnuXCqR3c5JhfsM2K6rQUKXwWdoeBuPNV3kQjj3mr.0" in gramine system length is 8388608
```

### Actual results

```
[2023-06-27T02:25:19Z WARN  xxx::api] File "/sgx/12D3KooWQHuPnuXCqR3c5JhfsM2K6rQUKXwWdoeBuPNV3kQjj3mr.0" already exists ,it means rotator still not finished
[2023-06-27T02:25:19Z ERROR xxx::api] error when read file "/sgx/12D3KooWQHuPnuXCqR3c5JhfsM2K6rQUKXwWdoeBuPNV3kQjj3mr.0",because "No such file or directory (os error 2)"
```

### Gramine commit hash

gramineproject/gramine digest:7e5b8b6111a6

dimakuv:
## Description of the changes 

This PR temporarily opens the file during `stat()` callback to check if the file still exists on the host. Without this open and if the same file was opened before, Gramine would serve the `chroot_stat()` callback from the cached `dent->inode` and wouldn't notice that the file was removed on the host. Note that such check is relevant only to chroot files, because all other file types are either pseudo in-memory files (dev, etc, sys, tmpfs) or fully controlled by Gramine files (encrypted).

Fixes #1422.

## How to test this PR? 

I don't know how to add such a test to our CI, that will start the Gramine process and then remove the host file under it.

Here is the manual test:
1. Change `large_file.c` LibOS regression test:
```diff
diff --git a/libos/test/regression/large_file.c b/libos/test/regression/large_file.c
@@ -91,6 +91,13 @@ int main(void) {
     if (close(fd) < 0)
         err(1, "close");

+    struct stat st;
+    if (stat(TEST_FILE, &st) < 0)
+        err(1, "stat");
+    sleep(5);
+    if (stat(TEST_FILE, &st) < 0)
+        err(1, "stat");
+
     printf("TEST OK\n");
     return 0;
 }
```

2. Check native:
   a. `<gramine install dir>/lib/x86_64-linux-gnu/gramine/tests/libos/regression/large_file` in one terminal
   b. `rm -rf libos/test/regression/tmp/large_file` in another terminal
   c. native app must complain in first terminal: `large_file: stat: No such file or directory`

3. Check Gramine:
   a. `gramine-direct large_file` in one terminal
   b. `rm -rf libos/test/regression/tmp/large_file` in another terminal
   c. Gramine app must complain in first terminal: `large_file: stat: No such file or directory`
   d. (without this PR, Gramine app incorrectly succeeds: `TEST OK`)

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1423)

### DIFF
--- a/libos/src/fs/chroot/fs.c
+++ b/libos/src/fs/chroot/fs.c
@@ -354,6 +354,27 @@ static int chroot_truncate(struct libos_handle* hdl, file_off_t size) {
     return ret;
 }
 
+static int chroot_stat(struct libos_dentry* dent, struct stat* buf) {
+    assert(locked(&g_dcache_lock));
+    assert(dent->inode);
+
+    int ret;
+
+    /* Temporary open of the file is required only to check if the file still exists. Without the
+     * open and if the same file was opened before, Gramine would serve this `stat()` callback from
+     * the cached `dent->inode` and wouldn't notice that the file was removed on the host. Note that
+     * such check is relevant only to chroot files, because all other file types are either pseudo
+     * in-memory files (dev, etc, sys, tmpfs) or fully controlled by Gramine files (encrypted). */
+    PAL_HANDLE palhdl;
+    ret = chroot_temp_open(dent, dent->inode->type, &palhdl);
+    if (ret < 0)
+        return ret;
+
+    ret = generic_inode_stat(dent, buf);
+    PalObjectClose(palhdl);
+    return ret < 0 ? pal_to_unix_errno(ret) : 0;
+}
+
 int chroot_readdir(struct libos_dentry* dent, readdir_callback_t callback, void* arg) {
     int ret;
     PAL_HANDLE palhdl;
@@ -512,7 +533,7 @@ struct libos_d_ops chroot_d_ops = {
     .lookup  = &chroot_lookup,
     .creat   = &chroot_creat,
     .mkdir   = &chroot_mkdir,
-    .stat    = &generic_inode_stat,
+    .stat    = &chroot_stat,
     .readdir = &chroot_readdir,
     .unlink  = &chroot_unlink,
     .rename  = &chroot_rename,


kailun-qin (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1423)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1423#-NYwFSHvAmUYKfeNS4RA:-NYw7qDpCKAPijnmlb_f:bcq1tg1) (no related file):*
I know this addresses the reported issue and helps us preserve the expected Linux behavior. But it's kind of like a ad-hoc fix since the cache can still be out of sync in some other cases (e.g., not remove the file but change the size of it)?

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1423)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1423#-NYwFSHvAmUYKfeNS4RA:-NYwec7UB4IOoG6ugjDM:b7p0ox4) (no related file):*
> But it's kind of like a ad-hoc fix since the cache can still be out of sync in some other cases (e.g., not remove the file but change the size of it)?

This is a fair comment. Several solutions then:
1. Do not allow such logic at all. Explain to people that Gramine doesn't expect the host to modify Gramine-used files, and that it's not possible to use Gramine as a passive "listener" of what happens on the host.
2. Keep the partial logic, as in this PR, and solve new use cases as/if people come and complain. I.e., grow the supported use cases organically and on-demand.
3. Make the logic more comprehensive -- e.g. update the size on `stat()` also, if it was modified on the host. And possibly other fields, see [here](https://github.com/gramineproject/gramine/blob/0684091aa1c95497f24dab9a060307b86fe3a713/libos/src/fs/libos_fs_util.c#L59).

I would prefer solution 3.

But I also don't mind solution 2, since currently Gramine has both things wrong (file existence and file size), and with this PR Gramine will have only one thing wrong (file size). So this PR still sounds like an improvement over the current state, even if an ad-hoc one.

I don't like solution 1. I see no reason to forbid such use cases.

___

kailun-qin (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1423)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1423#-NYwFSHvAmUYKfeNS4RA:-NZ9BcRr8ySjcA1waL1y:bxjfh3n) (no related file):*

> But it's kind of like a ad-hoc fix since the cache can still be out of sync in some other cases (e.g., not remove the file but change the size of it)?

This is a fair comment. Several solutions then:
1. Do not allow such logic at all. Explain to people that Gramine doesn't expect the host to modify Gramine-used files, and that it's not possible to use Gramine as a passive "listener" of what happens on the host.
2. Keep the partial logic, as in this PR, and solve new use cases as/if people come and complain. I.e., grow the supported use cases organically and on-demand.
3. Make the logic more comprehensive -- e.g. update the size on `stat()` also, if it was modified on the host. And possibly other fields, see [here](https://github.com/gramineproject/gramine/blob/0684091aa1c95497f24dab9a060307b86fe3a713/libos/src/fs/libos_fs_util.c#L59).

I would prefer solution 3.

But I also don't mind solution 2, since currently Gramine has both things wrong (file existence and file size), and with this PR Gramine will have only one thing wrong (file size). So this PR still sounds like an improvement over the current state, even if an ad-hoc one.

I don't like solution 1. I see no reason to forbid such use cases.
</blockquote></details>

OK, make sense. I also vote for solution 3, as it's more correct and seems not complex.

___

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1423)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1423#-NYwFSHvAmUYKfeNS4RA:-NkNHuKp14jN8Oh6RQyz:bturjm6) (no related file):*
I agree that we don't really have the semantics of chroot mounts well defined if the host changes the files when the enclave is still running. We should decide on that before merging this PR.

> I would prefer solution 3.

> I also vote for solution 3, as it's more correct and seems not complex.

Hmm, I'd rather say that it _is_ actually complex? We'd need to remove all caches for chroot FS and always passthrough everything. But this may behave weirdly with things like realpath(), where we need to traverse a path. And no cache probably means a rather huge perf penalty for SGX?

___

mwkmwkmwk (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1423)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv and @kailun-qin)

___
*[`libos/src/fs/chroot/fs.c` line 369 at r1](https://reviewable.io/reviews/gramineproject/gramine/1423#-NkNIozw0xrA9NOf8pTn:-NkNIozw0xrA9NOf8pTo:b-ucbgo1) ([raw file](https://github.com/gramineproject/gramine/blob/7433eb1d9befb7fa2a932b29003674c3553c8d51/libos/src/fs/chroot/fs.c#L369)):*
> ```C
>      * in-memory files (dev, etc, sys, tmpfs) or fully controlled by Gramine files (encrypted). */
>     PAL_HANDLE palhdl;
>     ret = chroot_temp_open(dent, dent->inode->type, &palhdl);
> ```

I'd suggest calling `PalStreamAttributesQuery` instead; it avoids the need for a second PAL call to close the file afterwards, and as a bonus can be used to actually update the attribute cache.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1423)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv, @kailun-qin, and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1423#-NYwFSHvAmUYKfeNS4RA:-NkP6RYL-iwOzq7HQS83:b-4c17as) (no related file):*

I agree that we don't really have the semantics of chroot mounts well defined if the host changes the files when the enclave is still running. We should decide on that before merging this PR.

> I would prefer solution 3.

> I also vote for solution 3, as it's more correct and seems not complex.

Hmm, I'd rather say that it _is_ actually complex? We'd need to remove all caches for chroot FS and always passthrough everything. But this may behave weirdly with things like realpath(), where we need to traverse a path. And no cache probably means a rather huge perf penalty for SGX?
</blockquote></details>

Ok, it's true that it will be a design decision that we should clearly state and agree upon. Created a discussion item for the next meeting: https://github.com/gramineproject/gramine/discussions/1660

___
___
*[`libos/src/fs/chroot/fs.c` line 369 at r1](https://reviewable.io/reviews/gramineproject/gramine/1423#-NkNIozw0xrA9NOf8pTn:-NkP6fSdD4Vf1ty16KsC:b-mum97x) ([raw file](https://github.com/gramineproject/gramine/blob/7433eb1d9befb7fa2a932b29003674c3553c8d51/libos/src/fs/chroot/fs.c#L369)):*

I'd suggest calling `PalStreamAttributesQuery` instead; it avoids the need for a second PAL call to close the file afterwards, and as a bonus can be used to actually update the attribute cache.
</blockquote></details>

Currently putting this comment on hold, as we need to discuss the general direction for these "host modifies file while Gramine is running"

PR #1421, opened 2023-06-27
[packaging] add Alpine Linux build scripts

amicrity:
## Description of the changes
This PR adds build scripts with patches needed to run Gramine on Alpine Linux.

## Features
Everything works*

*YMMV, some test might still error you out since they usually have to assume a glibc-based host, SGX works only through Docker (for now), glibc apps don't work (and are not going to work; we should respect Alpine community's position on mixing glibc with musl; it is possible to build Gramine with glibc, but it's way more hacky than just running a different distro through chroot or Docker)

## How to test this PR?
NOTE: in order to use SGX, you have to run Docker on a host supported by SGX SDK (such as Ubuntu or Alma) and mount its SGX device to Docker, this changes only if PSW/DCAP gets officially released for musl-based distros.
- create a Docker instance of Alpine 3.18/edge with following flags: `--cap-add SYS_ADMIN --cap-add SYS_PTRACE --security-opt seccomp=unconfined --device=/dev/sgx_enclave`
- add Alpine's main, community and testing repos to /etc/apk/repositories
- install doas, wget, alpine-sdk and perl-utils (and optionally git) through apk
- create a new user and assign it to the `wheel` and the `abuild` group
- remove -Wl,-z,pack-relative-relocs from /usr/share/abuild/default.conf
- pull the /alpine/ folder from this PR
- run `abuild-keygen -a -i`
- run `abuild -r checksum all` - this will compile Gramine from the main branch
- you might see an error `gramine*: ld-linux-x86-64.so.2: path not found`, ignore it
- packages got compiled to `~/packages/`, cd to that directory and install them using `apk`
For tests:
- compile Gramine repo and generate SGX privkey
- modify the example manifest.template from `gramine.runtimedir()` to `gramine.runtimedir(libc='musl')`

For bare metal running Alpine, you should set `-Dsgx=disabled` in the APKBUILD file.

Signed-off-by: Amie Raine <amie@invisiblethingslab.com>

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1421)

PR #1439, opened 2023-07-04
[packaging] Alpine

woju:
Taken over #1421, with fixes and CI.

## How to test

CI (esp. `pkg-apk-alpine3.17`)

## How to test yourself

Quick and dirty, but always works. Leaves root-owned files in you repo (you can clean with `sudo git clean -dxf` afterwards). Proper testing would require proper alpine packaging setup (read more here: https://wiki.alpinelinux.org/wiki/Creating_an_Alpine_package), which is not in scope here.

```
docker run -it --rm alpine:latest --volume $PWD:/gramine
cd /gramine/packaging/alpine
apk add alpine-sdk jq meson nasm
abuild-keygen -an
abuild -F -r snapshot checksum all
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1439)

### DIFF
--- a/.ci/alpine3.17.dockerfile
+++ b/.ci/alpine3.17.dockerfile
@@ -1,0 +1,12 @@
+FROM alpine:3.17
+
+RUN apk add \
+    alpine-sdk \
+    doas \
+    jq \
+    meson \
+    nasm
+
+COPY packaging/alpine/APKBUILD /APKBUILD
+RUN abuild -F -r builddeps
+RUN rm -f /APKBUILD
--- a/.ci/pkg-apk-alpine3.17.jenkinsfile
+++ b/.ci/pkg-apk-alpine3.17.jenkinsfile
@@ -1,0 +1,41 @@
+pipeline {
+    agent any
+    stages {
+        stage('makedist') {
+            agent {
+                dockerfile {
+                    filename '.ci/alpine3.17.dockerfile'
+                    reuseNode true
+                }
+            }
+            steps {
+                sh '''
+                    ABUILD_USERDIR=$PWD/abuild
+                    export ABUILD_USERDIR
+                    cd packaging/alpine
+                    abuild snapshot
+                '''
+            }
+        }
+        stage('build') {
+            agent {
+                dockerfile {
+                    filename '.ci/alpine3.17.dockerfile'
+                    args '--network=none'
+                    reuseNode true
+                }
+            }
+            steps {
+                sh '''
+                    ABUILD_USERDIR=$PWD/abuild
+                    REPODEST=$PWD/packages
+                    export ABUILD_USERDIR REPODEST
+                    cd packaging/alpine
+                    abuild-keygen -an
+                    abuild checksum all
+                '''
+                archiveArtifacts 'packages/'
+            }
+        }
+    }
+}
--- a/packaging/alpine/.gitignore
+++ b/packaging/alpine/.gitignore
@@ -1,0 +1,3 @@
+/build-dist/
+/pkg/
+/src/
--- a/packaging/alpine/APKBUILD
+++ b/packaging/alpine/APKBUILD
@@ -1,0 +1,85 @@
+# Contributor: Amie Raine <amie@invisiblethingslab.com>
+# Maintainer: Amie Raine <amie@invisiblethingslab.com>
+pkgname=gramine
+_real_pkgver=1.4post~UNRELEASED
+pkgver=$(printf %s "$_real_pkgver" | sed \
+    -e "s:post~UNRELEASED:_git$(printf %d 0x"$(git rev-parse HEAD 2>/dev/null | cut -c1-8)"):" \
+    -e 's:~:_:' \
+)
+pkgrel=0
+pkgdesc="A lightweight usermode guest OS designed to run a single Linux application"
+url="https://github.com/gramineproject/gramine"
+arch="x86_64"
+license="LGPL-3.0-or-later"
+makedepends="
+    autoconf
+    binutils-dev
+    bison
+    coreutils
+    findutils
+    gawk
+    gettext
+    grep
+    jq
+    libunwind
+    linux-headers
+    meson
+    musl-dev
+    nasm
+    openssl1.1-compat-dev
+    protobuf-c-compiler
+    protobuf-c-dev
+    py3-click
+    py3-elftools
+    py3-jinja2
+    py3-pytest
+    "
+depends="
+    py3-click
+    py3-cryptography
+    py3-elftools
+    py3-jinja2
+    py3-tomli
+    py3-tomli-w
+    "
+source="$pkgname-$pkgver.tar.gz"
+builddir="$srcdir/$pkgname-$_real_pkgver/"
+options="!check" # tests assume a glibc environment
+
+prepare() {
+    meson setup \
+        --wrap-mode=nodownload \
+        --prefix=/usr \
+        --buildtype=release \
+        -Ddirect=enabled \
+        -Dsgx=enabled \
+        -Dlibc=musl \
+        . build/
+
+    # sanity check for version number
+    test "$(meson introspect build/ --projectinfo | jq -r '.version')" = "$_real_pkgver" \
+        || die '$_real_pkgver does not match what is in meson.build'
+
+    default_prepare
+}
+
+build() {
+    meson compile -C build
+}
+
+package() {
+    DESTDIR="$pkgdir" meson install --no-rebuild -C build
+
+    for pc in \
+        "$pkgdir"/usr/lib/pkgconfig/ra_tls_gramine.pc \
+        "$pkgdir"/usr/lib/pkgconfig/secret_prov_gramine.pc \
+    ; do
+        sed -i -e "s:$_real_pkgver:$pkgver:" "$pc"
+    done
+}
+
+snapshot() {
+    meson setup -Dskeleton=enabled -Dlibc=musl build-dist/ ../..
+    meson dist -C build-dist/ --no-test --include-subprojects --format=gztar
+    install -D build-dist/meson-dist/"$pkgname-$_real_pkgver".tar.gz "$pkgname-$pkgver.tar.gz"
+}


woju:
Jenkins, test pkg-apk-alpine3.17 please

dimakuv (approved):
> Reviewed 2 of 4 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1439)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

kailun-qin (approved):
> Reviewed 2 of 4 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1439)* status: :shipit: complete! all files reviewed, all discussions resolved

mkow (approved):
> Reviewed 2 of 4 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1439)* status: :shipit: complete! all files reviewed, all discussions resolved

amicrity (commented):
I hope the .apk has been properly tested before, since this thing going through seems like a miracle

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1439#-:-NZa12dp73ELb-IN0o3w:b-ijrpjn)* status: all files reviewed, 9 unresolved discussions (waiting on @woju)

___
*[`.ci/alpine3.17.dockerfile` line 3 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZa1UiW6fyIicKNKJcs:-NZa1UiX8xlCzobIYkEb:by4wcdg) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/.ci/alpine3.17.dockerfile#L3)):*
> ```Dockerfile
> FROM alpine:3.17
> 
> RUN apk add \
> ```

FYI, please remember to mention that those packages should be installed prior to building on non-docker platforms (otherwise abuild will crash leaving lots of mess)
___
*[`packaging/alpine/.gitignore` line 1 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZ_wH24B4CngWVp9vpV:-NZ_wH24B4CngWVp9vpW:b-nvg05u) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/packaging/alpine/.gitignore#L1)):*
> ```ignore
> /src/
> ```

what is the .gitignore file doing here? and if it really has to, build-dist should be included too
___
*[`packaging/alpine/APKBUILD` line 4 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZa2Zpa2deW9shIRVXp:-NZa2Zpa2deW9shIRVXq:b-ii46he) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/packaging/alpine/APKBUILD#L4)):*
> ```abuild
> # Maintainer: Amie Raine <amie@invisiblethingslab.com>
> pkgname=gramine
> _real_pkgver=1.4post~UNRELEASED
> ```

this exact same apkbuild won't ever make its way to aports
___
*[`packaging/alpine/APKBUILD` line 8 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZ_xT6g5tFw99CCJtES:-NZ_xT6g5tFw99CCJtET:b-gg4q9z) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/packaging/alpine/APKBUILD#L8)):*
> ```abuild
>     -e 's:post~UNRELEASED:.u0n0r0e0l0e0a0s0e0d_git0:' \
>     -e 's:~:_:' \
> )
> ```

what on earth is this, call it 0-pre<version> or something like that, you can do more than one letter at once ;p

also about the sed, you can literally substitute it with ${_real_pkgver/~UNRELEASED/} instead of making an unreadable mess

i know, we cannot use a tilde, a commit hash and "something something alpine community bad", but we have way more options of doing it properly
"properly", as in with good practices
please refrain from perceiving the sexual orientation and/or gender identity of Alpine's core devs as more important (enough to not care about any good practices) than the fact they're more or less right on how the packaging should be done (friendly reminder, abuild is very not equal to meson)



_Code quote:_
```abuild
_real_pkgver=1.4post~UNRELEASED
pkgver=$(printf %s "$_real_pkgver" | sed \
    -e 's:post~UNRELEASED:.u0n0r0e0l0e0a0s0e0d_git0:' \
    -e 's:~:_:' \
)
```
___
*[`packaging/alpine/APKBUILD` line 44 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZ_zhdJEEJsXQDr5aAJ:-NZ_zhdK5u5_Lsi4KiqQ:b-erns39) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/packaging/alpine/APKBUILD#L44)):*
> ```abuild
> 	py3-tomli-w
> 	"
> source="$pkgname-$pkgver.tar.gz"
> ```

uhhhhh? this won't work outside of a CI, because this file won't exist; we should change that so it'll work both on CIs and using abuild on a clean system, I've proposed some changes which you didn't read mostly because you didn't like the "version" part
___
*[`packaging/alpine/APKBUILD` line 58 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZ_wfJA13u_XLgT3S2G:-NZ_wfJBBUKAq-GfuYGw:b-zfy3vf) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/packaging/alpine/APKBUILD#L58)):*
> ```abuild
> 
>     # sanity check for version number
>     test "$(meson introspect build/ --projectinfo | jq -r '.version')" = "$_real_pkgver" \
> ```

please refrain from using `test`, we don't do that for at least 10 years, it's a better practice to do `if <command>; then`
___
*[`packaging/alpine/APKBUILD` line 59 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZ_x22gBbz7B6_Gy1Vc:-NZ_x22gBbz7B6_Gy1Vd:bb7hmjd) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/packaging/alpine/APKBUILD#L59)):*
> ```abuild
>     # sanity check for version number
>     test "$(meson introspect build/ --projectinfo | jq -r '.version')" = "$_real_pkgver" \
>         || die '$_real_pkgver does not match what is in meson.build'
> ```

holy
___
*[`packaging/alpine/APKBUILD` line 80 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZ_yEPc2jPDRnVWVJC1:-NZ_yEPc2jPDRnVWVJC2:b-n6w7q1) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/packaging/alpine/APKBUILD#L80)):*
> ```abuild
> 
> snapshot() {
>     meson setup -Dskeleton=enabled -Dlibc=musl build-dist/ ../..
> ```

please don't use `meson` directly, we have `script/makedist.sh`

why are we copying the .tar.gz into the unknown anyways?
___
*[`packaging/alpine/APKBUILD` line 82 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZ_yPNC4tXdV5YvdySL:-NZ_yPNDAz2Y_cuyTL0s:b-pq0u9v) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/packaging/alpine/APKBUILD#L82)):*
> ```abuild
>     meson setup -Dskeleton=enabled -Dlibc=musl build-dist/ ../..
>     meson dist -C build-dist/ --no-test --include-subprojects --format=gztar
>     install -D build-dist/meson-dist/"$pkgname-$_real_pkgver".tar.gz "$pkgname-$pkgver.tar.gz"
> ```

it doesn't make any sense to use `install`, just do `mv`

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1439)* status: all files reviewed, 9 unresolved discussions (waiting on @amicrity)

___
*[`.ci/alpine3.17.dockerfile` line 3 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZa1UiW6fyIicKNKJcs:-NZaLu1H-fmICaIVY3-a:bhn71wa) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/.ci/alpine3.17.dockerfile#L3)):*
<details><summary><i>Previously, amicrity (Amie Raine) wrote…</i></summary><blockquote>

please remember to mention that those packages should be installed prior to building on non-docker platforms (otherwise abuild will crash leaving lots of mess)
</blockquote></details>

Where exactly should I mention this? It's already mentioned in https://github.com/gramineproject/gramine/blob/7c9f0dbd19384cc13cade296a4fa2906eaae9793/Documentation/devel/building.rst#install-dependencies, though that is Debian/Ubuntu specific, but relatable to other distros. Or is there any other place this should be added?
___
*[`packaging/alpine/.gitignore` line 1 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZ_wH24B4CngWVp9vpV:-NZa9F6nDhBHg4oDxZol:bl9pjlf) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/packaging/alpine/.gitignore#L1)):*
<details><summary><i>Previously, amicrity (Amie Raine) wrote…</i></summary><blockquote>

what is the .gitignore file doing here? and if it really has to, build-dist should be included too
</blockquote></details>

Done (added `build-dist/`), thanks for pointing this out.

This `.gitignore` is making sure that no-one commits any temporary files, like the content of `$srcdir` or `$pkgdir` (which is those two).
___
*[`packaging/alpine/APKBUILD` line 4 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZa2Zpa2deW9shIRVXp:-NZa9T3ZCopStHK6eSqd:b-bzimc1) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/packaging/alpine/APKBUILD#L4)):*
<details><summary><i>Previously, amicrity (Amie Raine) wrote…</i></summary><blockquote>

this exact same apkbuild won't ever make its way to aports
</blockquote></details>

Yes, it won't. It will require rework before someone submits it to aports. However, you're certainly aware that this APKBUILD's primary reason is to test alpine builds in our CI, not as preparation for submission to aports.
___
*[`packaging/alpine/APKBUILD` line 8 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZ_xT6g5tFw99CCJtES:-NZa9ccMDIFMTZaHQ6pQ:bfmmlcc) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/packaging/alpine/APKBUILD#L8)):*
> also about the sed, you can literally substitute it with ${\_real\_pkgver/~UNRELEASED/} instead of making an unreadable mess

This unreadable mess is a direct consequence of apk requiring a digit after a letter. I want to include the string "UNRELEASED" for when people will file issues. I don't want a situation where people report problems with package versioned like something released, in such cases it's hard to guess that this package was self-built from random commit after this release. "UNRELEASED" in version makes this unambiguous.

I can't make it ` ${\_real\_pkgver/~UNRELEASED/}`, because it would contain `post` and therefore would be rejected by `apk version --check`.

> please refrain from perceiving the sexual orientation and/or gender identity of Alpine's core devs

Whaaat? I don't think this has any bearing on technical issues, so I don't understand why are you bringing this subject up, it's offtopic here.

> they're more or less right on how the packaging should be done (friendly reminder, abuild is very not equal to meson)

I'm aware of differences between abuild and meson, thank you. I agree with some technical choices by Alpine devs and disagree with others, it's a simple as that. This version thingy is a consequence of limitations in support of perfectly valid version string imposed by them. Best I can do is to add a comment explaining the situation, with a link to this snippet: https://gitlab.alpinelinux.org/alpine/apk-tools/-/blob/23be7e321009b5c524ebb06746219760d9b40b6a/src/version.c#L37-38
___
*[`packaging/alpine/APKBUILD` line 44 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZ_zhdJEEJsXQDr5aAJ:-NZaFc302mk9QXytxase:bjnplte) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/packaging/alpine/APKBUILD#L44)):*
<details><summary><i>Previously, amicrity (Amie Raine) wrote…</i></summary><blockquote>

uhhhhh? this won't work outside of a CI, because this file won't exist; we should change that so it'll work both on CIs and using abuild on a clean system, I've proposed some changes which you didn't read mostly because you didn't like the "version" part
</blockquote></details>

This will work perfectly fine with "released" tarballs (will be called `gramine-1.5.tar.gz`). And yes, the main concern here is that this should work 1) in CI, 2) for manual builds of packages released directly by us, not via aports.

I understand this is not how Alpine operates their APKBUILDs, but they keep APKBUILDs separate from source code build. What I'm attempting to do here is similar to debian native packages. I haven't found a good explanation on https://wiki.alpinelinux.org/ how to do it best, the closest thing is https://wiki.alpinelinux.org/wiki/APKBUILD_examples:Git_checkout, and that example is incomplete at best. I acknowledge that this might be because this is very uncommon thing to do (if ever done).
___
*[`packaging/alpine/APKBUILD` line 58 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZ_wfJA13u_XLgT3S2G:-NZaFzLP--ocyvju5Xaz:b3c29p2) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/packaging/alpine/APKBUILD#L58)):*
<details><summary><i>Previously, amicrity (Amie Raine) wrote…</i></summary><blockquote>

please refrain from using `test`, we don't do that for at least 10 years, it's a better practice to do `if <command>; then`
</blockquote></details>

This project is only about 5 years old, so I don't understand how do "we don't do that for at least 10 years", and we're certainly doing that (https://github.com/gramineproject/gramine/blob/7c9f0dbd19384cc13cade296a4fa2906eaae9793/debian/get-linux-src-path.sh#L7-L12). The logic here is correct, I don't understand what's wrong with it.
___
*[`packaging/alpine/APKBUILD` line 59 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZ_x22gBbz7B6_Gy1Vc:-NZaGmlx-d7qbTJyXreG:bc9lmoh) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/packaging/alpine/APKBUILD#L59)):*
<details><summary><i>Previously, amicrity (Amie Raine) wrote…</i></summary><blockquote>

holy
</blockquote></details>

What's the concern here, exactly?
___
*[`packaging/alpine/APKBUILD` line 80 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZ_yEPc2jPDRnVWVJC1:-NZaGvhL0lp3dFSiy2y_:b8k4z3i) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/packaging/alpine/APKBUILD#L80)):*
<details><summary><i>Previously, amicrity (Amie Raine) wrote…</i></summary><blockquote>

please don't use `meson` directly, we have `script/makedist.sh`

why are we copying the .tar.gz into the unknown anyways?
</blockquote></details>

`script/makedist.sh` has `-Dlibc=glibc` hardcoded in `meson setup`, I would have to adjust that (otherwise there would be no musl in tarball), and creates two exactly same tarballs (differing only in name) which is useless here. It's easier to simply run `meson dist` directly here.

We're copying tarball into the current directory (`$startdir` to be precise). If I'm not mistaken, we will always have `is_remote` false (https://gitlab.alpinelinux.org/alpine/abuild/-/blob/d4b2415fb1d52208c2dcdfb4ec547b09a671eca2/abuild.in#L419), because having anything downloaded from e.g. github is not acceptable (you shouldn't be required to push to official repository before building from local branch), therefore we need tarball in `$startdir` because of https://gitlab.alpinelinux.org/alpine/abuild/-/blob/d4b2415fb1d52208c2dcdfb4ec547b09a671eca2/abuild.in#L458.
___
*[`packaging/alpine/APKBUILD` line 82 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZ_yPNC4tXdV5YvdySL:-NZaGsKe9BY0AKqsLMDb:bb6rlrq) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/packaging/alpine/APKBUILD#L82)):*
<details><summary><i>Previously, amicrity (Amie Raine) wrote…</i></summary><blockquote>

it doesn't make any sense to use `install`, just do `mv`
</blockquote></details>

`install` is equally correct, and `install -D` is for when this might be adapted to use `SRCDEST` somehow, which I'm still unsure about. In any case, the tarball needs to end up in pwd/`$startdir` (cf. above thread).

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1439)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (1 more required) (waiting on @amicrity)

amicrity (commented):
> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1439)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required) (waiting on @woju)

___
*[`.ci/alpine3.17.dockerfile` line 3 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZa1UiW6fyIicKNKJcs:-NZaPvgCCJVsq2OAuNSw:bbiqntd) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/.ci/alpine3.17.dockerfile#L3)):*

Where exactly should I mention this? It's already mentioned in https://github.com/gramineproject/gramine/blob/7c9f0dbd19384cc13cade296a4fa2906eaae9793/Documentation/devel/building.rst#install-dependencies, though that is Debian/Ubuntu specific, but relatable to other distros. Or is there any other place this should be added?
</blockquote></details>

it's okay as long as this PR's description will be added somewhere, if not, it's really worth mentioning (for example on the mentioned website) that `alpine-sdk jq meson nasm` are essential for the build process (if someone sets up abuild only, and doesn't install the dependencies and runs the build, it might break itself beyond repair)
___
*[`packaging/alpine/.gitignore` line 1 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZ_wH24B4CngWVp9vpV:-NZaQxb5-XZmFpR63D2H:bb7k0qh) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/packaging/alpine/.gitignore#L1)):*

Done (added `build-dist/`), thanks for pointing this out.

This `.gitignore` is making sure that no-one commits any temporary files, like the content of `$srcdir` or `$pkgdir` (which is those two).
</blockquote></details>

lgtm
___
*[`packaging/alpine/APKBUILD` line 4 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZa2Zpa2deW9shIRVXp:-NZaR4tREIRLua4hEjfp:b-eym0xq) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/packaging/alpine/APKBUILD#L4)):*

Yes, it won't. It will require rework before someone submits it to aports. However, you're certainly aware that this APKBUILD's primary reason is to test alpine builds in our CI, not as preparation for submission to aports.
</blockquote></details>

APKBUILDs aren't meant for CI (and that's also why they hurt so much when put in a CI), just copy the commands from prepare() and build() and here'll be your CI

and please stop assuming that my goal here is to upstream this exact APKBUILD to aports, I want it to meet at least some Alpine's standards (since we will probably support Alpine for a longer time; doing so will make maintaining it way easier) instead of using deprecated (or never actually supported) configurations
___
*[`packaging/alpine/APKBUILD` line 8 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZ_xT6g5tFw99CCJtES:-NZaSEnF4zjzfDWvyskg:b-342zo5) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/packaging/alpine/APKBUILD#L8)):*
nope, you just (for some reason) don't want to use semantic versioning
there are two solutions for that:
   - instead of UNRELEASED, do 1.4_gitX (with the X being the build number; we should go with the date but I know you really don't want to do it)
   https://wiki.alpinelinux.org/wiki/APKBUILD_Reference#pkgver; 1.4_pre420 - good, 1.4_pre - good, 1.4_.u0n0r0e0l0e0a0s0e0d\_git0 - very bad
   - overriding the sanitycheck() in abuild

from the link above: "To package a specific git commit, the date of the commit gets appended to the latest release, e.g. `1.0.0_git20180204`." It *is* a proper way of doing this, but if you really don't like it, we have other options available

> Whaaat? I don't think this has any bearing on technical issues, so I don't understand why are you bringing this subject up, it's offtopic here.
from what i recall that was your only explanation for making a huge mess (not going into details) and thinking that after what you will do with this APKBUILD, AL community will "fix" the issues you're trying to "circumvent"
however i don't think breaking the core principles of abuild will attract enough attention of just a few people who maintain Alpine and have better things to do (as seen above ;p)
___
*[`packaging/alpine/APKBUILD` line 44 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZ_zhdJEEJsXQDr5aAJ:-NZaWxEj22pp92dlvCLk:b-45ln0p) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/packaging/alpine/APKBUILD#L44)):*

This will work perfectly fine with "released" tarballs (will be called `gramine-1.5.tar.gz`). And yes, the main concern here is that this should work 1) in CI, 2) for manual builds of packages released directly by us, not via aports.

I understand this is not how Alpine operates their APKBUILDs, but they keep APKBUILDs separate from source code build. What I'm attempting to do here is similar to debian native packages. I haven't found a good explanation on https://wiki.alpinelinux.org/ how to do it best, the closest thing is https://wiki.alpinelinux.org/wiki/APKBUILD_examples:Git_checkout, and that example is incomplete at best. I acknowledge that this might be because this is very uncommon thing to do (if ever done).
</blockquote></details>

you know what also works perfectly fine with "released" tarballs? omitting the snapshot entirely

you didn't understand the example, so it makes sense why it looks "incomplete at best"
___
*[`packaging/alpine/APKBUILD` line 58 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZ_wfJA13u_XLgT3S2G:-NZaVY3y30BOU7PcaV6a:b-3imbva) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/packaging/alpine/APKBUILD#L58)):*

This project is only about 5 years old, so I don't understand how do "we don't do that for at least 10 years", and we're certainly doing that (https://github.com/gramineproject/gramine/blob/7c9f0dbd19384cc13cade296a4fa2906eaae9793/debian/get-linux-src-path.sh#L7-L12). The logic here is correct, I don't understand what's wrong with it.
</blockquote></details>

the fact that we (as in, Invisible Things Lab) do it doesn't mean that everyone else does it too
test in bash/sh is the equivalent of `[[`
it's a good practice to do `if [[ -d ... ]]; then`, alternatively `[[ ! -d ... ]] && exit 1`
___
*[`packaging/alpine/APKBUILD` line 59 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZ_x22gBbz7B6_Gy1Vc:-NZaW5Nh1nGxQMJdI21Z:bs8ccfd) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/packaging/alpine/APKBUILD#L59)):*

What's the concern here, exactly?
</blockquote></details>

there's no valid reason (at least to me) to perform that check
___
*[`packaging/alpine/APKBUILD` line 82 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZ_yPNC4tXdV5YvdySL:-NZaWGH7BkmTTCFSdwV3:b-6gjqsc) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/packaging/alpine/APKBUILD#L82)):*

`install` is equally correct, and `install -D` is for when this might be adapted to use `SRCDEST` somehow, which I'm still unsure about. In any case, the tarball needs to end up in pwd/`$startdir` (cf. above thread).
</blockquote></details>

yeah, right, https://www.man7.org/linux/man-pages/man1/install.1.html doesn't mention that you can't do that, but usually it's used for installing precompiled pkgs to /usr, while you use it with just the intention of copying a tarball; why?

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1439)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required) (waiting on @amicrity)

___
*[`.ci/alpine3.17.dockerfile` line 3 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZa1UiW6fyIicKNKJcs:-NZarbMs28NKGf7h3IVK:bmlczib) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/.ci/alpine3.17.dockerfile#L3)):*
<details><summary><i>Previously, amicrity (Amie Raine) wrote…</i></summary><blockquote>

it's okay as long as this PR's description will be added somewhere, if not, it's really worth mentioning (for example on the mentioned website) that `alpine-sdk jq meson nasm` are essential for the build process (if someone sets up abuild only, and doesn't install the dependencies and runs the build, it might break itself beyond repair)
</blockquote></details>

I'll leave this out of scope for now. Maybe in later PR, for now we're running out of time.
___
*[`packaging/alpine/APKBUILD` line 4 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZa2Zpa2deW9shIRVXp:-NZaYOAZCL-8gwogMCsS:b-mrib2j) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/packaging/alpine/APKBUILD#L4)):*
> APKBUILDs aren't meant for CI

Packaging is not only `prepare()` and `build()`, it's also dependencies, that might be or might not be present on all distros/versions. I don't want to copy instructions, copying is prone to missing something. The only sure way is to test end-to-end, and exactly the thing that will be used to package for release (and this APKBUILD is exactly it).

Yes, Alpine will be supported as long as practicable, and this APKBUILD will be living in our repo. As you can see there are several features that make it easier to maintain in our repo, like `_real_pkgver` that is meant to exactly reflect what is in `meson.build` and this version check. Optimally, abuild should support something like Debian "native packages", where `debian/` directory is meant to live together with the code packaged, in the same repo. Until and unless that happens, we are doing what we can to emulate that.
___
*[`packaging/alpine/APKBUILD` line 8 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZ_xT6g5tFw99CCJtES:-NZa_L165s9FbNdKuY81:b-82uggr) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/packaging/alpine/APKBUILD#L8)):*
Yes, and I have explained the reason why we're not doing "semantic versioning" as defined by javascript developers. This UNRELEASED string is based on experiences with past bug reports. I still haven't heard a credible explanation why you can't write consecutive letters in `pkgver=` (unless it's part of specific suffix) other than "just do $something_else", that something else being usually "semantic versioning", which I find lacking in necessary features. From the suggested solutions none are acceptable to us. `date` is non-starter, because if you checkout this same commit twice, on two different days, you should get exactly the same packages, bit for bit identical. For plain `$(date)` you won't get this. You can try getting timestamp from git commit, but then this gets problematic in other ways.

> from what i recall

You recall wrong, also not going into details, so EOT.
___
*[`packaging/alpine/APKBUILD` line 44 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZ_zhdJEEJsXQDr5aAJ:-NZafClH11--psqlbnbs:b-idx2v4) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/packaging/alpine/APKBUILD#L44)):*
<details><summary><i>Previously, amicrity (Amie Raine) wrote…</i></summary><blockquote>

you know what also works perfectly fine with "released" tarballs? omitting the snapshot entirely

you didn't understand the example, so it makes sense why it looks "incomplete at best"
</blockquote></details>

I'd be grateful if you pointed me to documentation of correct solution to the problem stated above (i.e. CI). Otherwise this discussion is not productive.
___
*[`packaging/alpine/APKBUILD` line 58 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZ_wfJA13u_XLgT3S2G:-NZadLjSC7IOW2eUGrrk:b-6fnglj) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/packaging/alpine/APKBUILD#L58)):*
<details><summary><i>Previously, amicrity (Amie Raine) wrote…</i></summary><blockquote>

the fact that we (as in, Invisible Things Lab) do it doesn't mean that everyone else does it too
test in bash/sh is the equivalent of `[[`
it's a good practice to do `if [[ -d ... ]]; then`, alternatively `[[ ! -d ... ]] && exit 1`
</blockquote></details>

In fact, `test` is not equivalent to `[[`. `[[` is bash-only *syntax* (not in POSIX, therefore not a good practice)` and `test` is a command. `test` is equivalent to `[`, apart from closing bracket.

`[ <something> ] && ...` is very wrong when coupled with `set -e`. Only `[ ] ||` (or `test ||`) is OK in the face of `set -e`.
___
*[`packaging/alpine/APKBUILD` line 59 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZ_x22gBbz7B6_Gy1Vc:-NZadzkRDG2Ot_yAKGc3:b-h25zuh) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/packaging/alpine/APKBUILD#L59)):*
<details><summary><i>Previously, amicrity (Amie Raine) wrote…</i></summary><blockquote>

there's no valid reason (at least to me) to perform that check
</blockquote></details>

Yes, in theory, theory matches practice. In practice, it's different. This sanity check is meant for when I would bump `meson.build` but forgot to change `_real_pkgver` here. Those changes are done manually, so we need to ensure they're synchronised. We do this check also for all other packaging:
https://github.com/gramineproject/gramine/blob/7c9f0dbd19384cc13cade296a4fa2906eaae9793/debian/rules#L44-L49
https://github.com/gramineproject/gramine/blob/7c9f0dbd19384cc13cade296a4fa2906eaae9793/gramine.spec#L63-L68
I see no reason why we wouldn't need to perform that check, but only on Alpine.
___
*[`packaging/alpine/APKBUILD` line 82 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZ_yPNC4tXdV5YvdySL:-NZaeCTx78zzwvCIzMjP:b5a672f) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/packaging/alpine/APKBUILD#L82)):*
<details><summary><i>Previously, amicrity (Amie Raine) wrote…</i></summary><blockquote>

yeah, right, https://www.man7.org/linux/man-pages/man1/install.1.html doesn't mention that you can't do that, but usually it's used for installing precompiled pkgs to /usr, while you use it with just the intention of copying a tarball; why?
</blockquote></details>

`install` doesn't care about what it's copying, and has some useful options that without it would require additional `mkdir`, `chmod` and/or `chown`. It's just very convenient. I don't care about what it's "usually" used for.

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1439)* status: 3 of 4 files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @amicrity, @dimakuv, @kailun-qin, and @mkow)

___
*[`packaging/alpine/APKBUILD` line 8 at r2](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZ_xT6g5tFw99CCJtES:-NZf3wJN67osyCz0Q_FC:bnu5wbw) ([raw file](https://github.com/gramineproject/gramine/blob/2df3dc044d3c2d4698532abd9dfe77dbad9d332c/packaging/alpine/APKBUILD#L8)):*

Yes, and I have explained the reason why we're not doing "semantic versioning" as defined by javascript developers. This UNRELEASED string is based on experiences with past bug reports. I still haven't heard a credible explanation why you can't write consecutive letters in `pkgver=` (unless it's part of specific suffix) other than "just do $something_else", that something else being usually "semantic versioning", which I find lacking in necessary features. From the suggested solutions none are acceptable to us. `date` is non-starter, because if you checkout this same commit twice, on two different days, you should get exactly the same packages, bit for bit identical. For plain `$(date)` you won't get this. You can try getting timestamp from git commit, but then this gets problematic in other ways.

> from what i recall

You recall wrong, also not going into details, so EOT.
</blockquote></details>

I've changed this so it outputs the commit ID of the build (in decimal...). There aren't really many other options. We can't use "build ID", because there's no such thing outside CI.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1439)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required) (waiting on @amicrity)

mkow (approved):
> Dismissed @amicrity from 8 discussions.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1439)* status: all files reviewed, 8 unresolved discussions

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1439#-NZjgELH6Wp1kK5Ew3Yx:-NZjgELH6Wp1kK5Ew3Yy:b-iyo1bb) (no related file):*
FYI: We (me, Amie and woju) reached an agreement on priv that (if I understand correctly) we'll merge the current state of the PR, make the release and then work on making these scripts more suitable with Alpine standards.
@amicrity: I'll dismiss you from the discussions, I hope you don't mind - it seems that you forgot to unblock and went AFK, and we need to finish this PR as soon as possible (and to my understanding you were fine with the current status of the PR as long as we'll work on fixing these issues after the release).

___

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1439)* status: :shipit: complete! all files reviewed, all discussions resolved

nekopsykose:
> then work on making these scripts more suitable with Alpine standards.

there's no need to, as i have no interest in accepting this software into aports itself. you should keep your downstream builds however you want to make them, and it's better for you that way too.

PR #1454, opened 2023-07-12
[LibOS,PAL] Fix wrong handling of sysfs NUMA nodes when nodes are offline

### LINKED ISSUE #1452: [LibOS,PAL] NUMA info is initialized on `possible` instead of `online`
NUMA nodes can be *possible* (memory can be plugged in for this node in the future) and *online* (node has the attached memory).

Gramine current code assumes that possible nodes set is the same as the online nodes set.

### Problem 1

Consider this flow: 
- https://github.com/gramineproject/gramine/blob/db2a92e4e37acd7e5023cd728fb78df00527d14d/pal/src/host/linux-common/topo_info.c#L282
- https://github.com/gramineproject/gramine/blob/db2a92e4e37acd7e5023cd728fb78df00527d14d/pal/src/host/linux-common/topo_info.c#L360-L368

We set `nodes_cnt` to the number of *possible* nodes, then we assume that each of these `1..nodes_cnt` nodes has the `cpulist` file. But this is not true on some systems:
```
# cat /sys/devices/system/node/possible
0-19
# cat /sys/devices/system/node/online
0-1
# ls /sys/devices/system/node/
node0  node1  online  possible  power  uevent
```

### Problem 2

The `distances` array in this case is wrongly initialized by Gramine. Gramine assumes that `distances` is a 2D array `[nodes_cnt][nodes_cnt]`, i.e. the distances are between all *possible* nodes:
- https://github.com/gramineproject/gramine/blob/db2a92e4e37acd7e5023cd728fb78df00527d14d/pal/src/host/linux-common/topo_info.c#L294
- https://github.com/gramineproject/gramine/blob/db2a92e4e37acd7e5023cd728fb78df00527d14d/pal/src/host/linux-common/topo_info.c#L370-L375

In reality, on Linux the `distances` are only between *online* nodes:
```
# cat /sys/devices/system/node/node0/distance
10 21
# cat /sys/devices/system/node/node1/distance
21 10
```

### How Linux deals with NUMA nodes

The Linux code to identify and to set up the NUMA nodes (actually, only ccNUMA nodes) is extremely complicated. To be honest, I didn't understand it fully -- for example, I still don't understand at which point and from which source the *possible* nodes are identified. However, the high-level logic goes like this:
1. Linux calls `x86_numa_init()`.
2. On Intel x86, the function `x86_acpi_numa_init()` is used -- i.e., the NUMA topology is extracted from ACPI tables (prepared by BIOS).
3. Then the function `acpi_numa_init()` is called -- it gets NUMA info from ACPI SRAT and SLIT tables.
   - `acpi_parse_x2apic_affinity()` is mainly used on modern CPUs, i.e., extracting CPU/NUMA info from x2APIC.
   - this info gives us PXMs which are finally mapped to NUMA nodes.
4. Anyway, `node_possible_map` and `node_online_map`, as well as `numa_distance` are initialized in the end.
   - Importantly, `numa_distance` is initialized **only from online nodes** (in the link, online nodes are the ones that are NOT marked as `NUMA_NO_NODE`): https://github.com/torvalds/linux/blob/v6.4/drivers/acpi/numa/srat.c#L215-L235
   - The above is also confirmed via `numa_distance_cnt`; see [this code](https://github.com/torvalds/linux/blob/v6.4/arch/x86/mm/numa.c#L363-L399); `dmesg` on this machine tells me: `NUMA: Initialized distance table, cnt=2`.

I looked at the SRAT and SLIT tables. These tables are a bit weird: The SLIT table shows me 10 localities, but the SRAT table shows 2 proximity domains (the latter makes sense). This still doesn't explain where 20 possible nodes are taken from, and why 18 of these nodes are "memoryless" (which means there are CPUs or Generic Initiators that don't report a corresponding memory node). I got lost in the Linux code for this, and my current assumption is that the SRAT and SLIT tables are hard-coded incorrectly in the BIOS on this particular machine.

### Relevant Linux sources and assorted references

Linux sources:

https://github.com/torvalds/linux/blob/master/include/linux/nodemask.h
https://github.com/torvalds/linux/blob/v6.4/mm/page_alloc.c#L217
https://github.com/torvalds/linux/blob/v6.4/mm/mm_init.c#L1854-L1876
https://github.com/torvalds/linux/blob/v6.4/arch/x86/mm/numa.c
https://github.com/torvalds/linux/blob/v6.4/arch/x86/mm/srat.c
https://github.com/torvalds/linux/blob/v6.4/drivers/acpi/numa/srat.c
https://github.com/torvalds/linux/blob/v6.4/drivers/acpi/numa/hmat.c

References to random NUMA/ACPI stuff:
https://uefi.org/htmlspecs/ACPI_Spec_6_4_html/17_NUMA_Architecture_Platforms/NUMA_Architecture_Platforms.html
https://uefi.org/htmlspecs/ACPI_Spec_6_4_html/05_ACPI_Software_Programming_Model/ACPI_Software_Programming_Model.html
https://www.codeblueprint.co.uk/2019/07/12/what-are-slit-tables.html
https://adam.younglogic.com/2022/07/looking-at-acpi-pptt-from-userspace/
https://wiki.osdev.org/SRAT
https://www.techarp.com/bios-guide/acpi-srat-table/
https://www.kernel.org/doc/html/v5.6/vm/numa.html
https://lwn.net/Articles/7177/
https://patchwork.kernel.org/project/linux-acpi/cover/20191216153809.105463-1-Jonathan.Cameron@huawei.com/
https://www.linuxjournal.com/content/handling-complex-memory-situations

dimakuv:
## Description of the changes 

Previously, Gramine iterated through all NUMA nodes within the range obtained from the host's `/sys/devices/system/node/possible` file to get `cpulist` and `distance`. It also used the count of possible nodes to parse the `distance` between nodes.

However, there is no `/sys/devices/system/node/nodeX` directory when a node is not online. Also, `/sys/devices/system/node/nodeX/distance` lists distances between online nodes only (jumping over offline nodes).

This can lead to failures when Gramine tries to read values of `cpulist` and `distance` pseudo-files from the directory. Plus a correct parsing of `distance` should consider only online nodes.

For context, see https://github.com/gramineproject/gramine/issues/1452.

Fixes #1452.

The main Linux source ref: https://github.com/torvalds/linux/blob/v6.4/drivers/base/node.c#L543-L564

## How to test this PR? 

Manually on weird NUMA platforms...

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1454)

### DIFF
--- a/libos/src/fs/sys/fs.c
+++ b/libos/src/fs/sys/fs.c
@@ -185,6 +185,9 @@ bool sys_resource_name_exists(struct libos_dentry* parent, const char* name) {
     if (pseudo_parse_ulong(&name[prefix_len], total - 1, &n) < 0)
         return false;
 
+    if (!strcmp(parent->name, "node") && !g_pal_public_state->topo_info.numa_nodes[n].is_online)
+        return false;
+
     return true;
 }
 
@@ -198,6 +201,9 @@ int sys_resource_list_names(struct libos_dentry* parent, readdir_callback_t call
     /* Generate "{prefix}N" names for all N less than total */
 
     for (size_t i = 0; i < total; i++) {
+        if (!strcmp(parent->name, "node") && !g_pal_public_state->topo_info.numa_nodes[i].is_online)
+            continue;
+
         char ent_name[strlen(prefix) + 22];
         snprintf(ent_name, sizeof(ent_name), "%s%zu", prefix, i);
         int ret = callback(ent_name, arg);
@@ -283,8 +289,6 @@ static void init_node_dir(struct pseudo_node* node) {
     pseudo_add_str(nodeX, "distance", &sys_node_load);
     pseudo_add_str(nodeX, "meminfo", &sys_node_meminfo_load);
 
-    // TODO(mkow): Does this show up for offline nodes? I never succeeded in shutting down one, even
-    // after shutting down all CPUs inside the node it shows up as online on `node/online` list.
     struct pseudo_node* hugepages = pseudo_add_dir(nodeX, "hugepages");
     struct pseudo_node* hugepages_2m = pseudo_add_dir(hugepages, "hugepages-2048kB");
     pseudo_add_str(hugepages_2m, "nr_hugepages", &sys_node_load);
--- a/libos/src/fs/sys/node_info.c
+++ b/libos/src/fs/sys/node_info.c
@@ -65,15 +65,27 @@ int sys_node_load(struct libos_dentry* dent, char** out_data, size_t* out_size) 
     const char* name = dent->name;
     const struct pal_topo_info* topo = &g_pal_public_state->topo_info;
     const struct pal_numa_node_info* numa_node = &topo->numa_nodes[node_id];
+    if (!numa_node->is_online)
+        return -ENOENT;
+
     char str[PAL_SYSFS_MAP_FILESZ] = {0};
     if (strcmp(name, "cpumap") == 0) {
         ret = sys_print_as_bitmask(str, sizeof(str), topo->threads_cnt, is_in_same_node, &node_id);
     } else if (strcmp(name, "distance") == 0) {
-        size_t* distances = topo->numa_distance_matrix + node_id * topo->numa_nodes_cnt;
-        size_t str_pos = 0;
+        size_t online_nodes_cnt = 0; /* `numa_distance_matrix` reflects only online nodes */
+        size_t i_online = 0;
         for (size_t i = 0; i < topo->numa_nodes_cnt; i++) {
+            struct pal_numa_node_info* node = &topo->numa_nodes[i];
+            if (i == node_id)
+                i_online = online_nodes_cnt;
+            if (node->is_online)
+                online_nodes_cnt++;
+        }
+        size_t* distances = topo->numa_distance_matrix + i_online * online_nodes_cnt;
+        size_t str_pos = 0;
+        for (size_t i = 0; i < online_nodes_cnt; i++) {
             ret = snprintf(str + str_pos, sizeof(str) - str_pos,
-                           "%zu%s", distances[i], i != (topo->numa_nodes_cnt - 1) ? " " : "\n");
+                           "%zu%s", distances[i], i != (online_nodes_cnt - 1) ? " " : "\n");
             if (ret < 0)
                 return ret;
             if ((size_t)ret >= sizeof(str) - str_pos)
@@ -114,6 +126,11 @@ int sys_node_meminfo_load(struct libos_dentry* dent, char** out_data, size_t* ou
     int ret = sys_resource_find(dent, "node", &node_id);
     if (ret < 0)
         return ret;
+
+    const struct pal_topo_info* topo = &g_pal_public_state->topo_info;
+    const struct pal_numa_node_info* numa_node = &topo->numa_nodes[node_id];
+    if (!numa_node->is_online)
+        return -ENOENT;
 
     size_t size = 0, max = 256;
     char* str = malloc(max);
--- a/pal/include/arch/x86_64/pal_topology.h
+++ b/pal/include/arch/x86_64/pal_topology.h
@@ -112,10 +112,13 @@ struct pal_topo_info {
 
     /*
      * Has `online_numa_nodes_cnt * online_numa_nodes_cnt` valid elements, where
-     * `online_numa_nodes_cnt` is the number of NUMA nodes that have `is_online == true`.
+     * `online_numa_nodes_cnt` is the number of NUMA nodes that have `is_online == true`. For
+     * simplicity, this array is allocated as `possible_numa_nodes_cnt * possible_numa_nodes_cnt`
+     * but only the first `online_numa_nodes_cnt` elements in each row/column are usable, and the
+     * rest are invalid.
      *
      * Value in `numa_distance_matrix[i*online_numa_nodes_cnt + j]` is a NUMA distance from online
-     * node i to online node j.
+     * node i to online node j (i.e. indexing only through the online nodes).
      *
      * It may be surprising that this array contains only online nodes, but this is what Linux does:
      *   https://elixir.bootlin.com/linux/v6.4/source/drivers/base/node.c#L543
--- a/pal/include/arch/x86_64/pal_topology.h
+++ b/pal/include/arch/x86_64/pal_topology.h
@@ -110,7 +110,15 @@ struct pal_topo_info {
     size_t numa_nodes_cnt;
     struct pal_numa_node_info* numa_nodes;
 
-    /* Has `numa_nodes_cnt * numa_nodes_cnt` elements.
-     * numa_distance_matrix[i*numa_nodes_cnt + j] is NUMA distance from node i to node j. */
+    /*
+     * Has `online_numa_nodes_cnt * online_numa_nodes_cnt` valid elements, where
+     * `online_numa_nodes_cnt` is the number of NUMA nodes that have `is_online == true`.
+     *
+     * Value in `numa_distance_matrix[i*online_numa_nodes_cnt + j]` is a NUMA distance from online
+     * node i to online node j.
+     *
+     * It may be surprising that this array contains only online nodes, but this is what Linux does:
+     *   https://elixir.bootlin.com/linux/v6.4/source/drivers/base/node.c#L543
+     */
     size_t* numa_distance_matrix;
 };
--- a/pal/src/host/linux-common/topo_info.c
+++ b/pal/src/host/linux-common/topo_info.c
@@ -291,7 +291,7 @@ int get_topology_info(struct pal_topo_info* topo_info) {
     size_t sockets_cnt = 0;
     struct pal_socket_info* sockets = malloc(threads_cnt * sizeof(*sockets)); // overapproximate the count
     struct pal_numa_node_info* numa_nodes = malloc(nodes_cnt * sizeof(*numa_nodes));
-    size_t* distances = malloc(nodes_cnt * nodes_cnt * sizeof(*distances));
+    size_t* distances = malloc(nodes_cnt * nodes_cnt * sizeof(*distances)); // overapproximate
     if (!threads || !caches || !cores || !sockets || !numa_nodes || !distances) {
         ret = -ENOMEM;
         goto fail;
@@ -357,7 +357,13 @@ int get_topology_info(struct pal_topo_info* topo_info) {
         }
     }
 
+    size_t online_nodes_cnt = 0; /* Required for `distance` which reflects only online nodes */
     for (size_t i = 0; i < nodes_cnt; i++) {
+        if (!numa_nodes[i].is_online)
+            continue;
+
+        online_nodes_cnt++;
+
         snprintf(path, sizeof(path), "/sys/devices/system/node/node%zu/cpulist", i);
         ret = iterate_ranges_from_file(path, set_node_id, &(struct set_node_id_args){
             .threads = threads,
@@ -367,17 +373,32 @@ int get_topology_info(struct pal_topo_info* topo_info) {
         if (ret < 0)
             goto fail;
 
-        ret = snprintf(path, sizeof(path), "/sys/devices/system/node/node%zu/distance", i);
-        if (ret < 0)
-            goto fail;
-        ret = read_numbers_from_file(path, distances + i * nodes_cnt, nodes_cnt);
-        if (ret < 0)
-            goto fail;
-
         /* Since our sysfs doesn't support writes, set persistent hugepages to their default value
          * of zero */
         numa_nodes[i].nr_hugepages[HUGEPAGES_2M] = 0;
         numa_nodes[i].nr_hugepages[HUGEPAGES_1G] = 0;
+    }
+
+    /*
+     * Be careful to reflect only online nodes in the `distances` array. E.g. if a system has node 0
+     * online, node 1 offline and node 2 online, then distances should look like this:
+     *
+     *   [ node 0 -> node 0, node 0 -> node 2
+     *     node 2 -> node 0, node 2 -> node 2 ]
+     */
+    for (size_t i = 0, i_online = 0; i < nodes_cnt; i++) {
+        if (!numa_nodes[i].is_online)
+            continue;
+
+        ret = snprintf(path, sizeof(path), "/sys/devices/system/node/node%zu/distance", i);
+        if (ret < 0)
+            goto fail;
+        ret = read_numbers_from_file(path, distances + i_online * online_nodes_cnt,
+                                     online_nodes_cnt);
+        if (ret < 0)
+            goto fail;
+
+        i_online++;
     }
 
     for (size_t i = 0; i < threads_cnt; i++) {
--- a/pal/src/host/linux-sgx/pal_main.c
+++ b/pal/src/host/linux-sgx/pal_main.c
@@ -183,10 +183,14 @@ static int sanitize_topo_info(struct pal_topo_info* topo_info) {
         return -PAL_ERROR_INVAL;
     }
 
+    /* Required for `numa_distance_matrix` which reflects only online nodes */
+    size_t online_nodes_cnt = 0;
+
     for (size_t i = 0; i < topo_info->numa_nodes_cnt; i++) {
         struct pal_numa_node_info* node = &topo_info->numa_nodes[i];
         coerce_untrusted_bool(&node->is_online);
         if (node->is_online) {
+            online_nodes_cnt++;
             for (size_t j = 0; j < HUGEPAGES_MAX; j++) {
                 size_t unused; // can't use __builtin_mul_overflow_p because clang doesn't have it.
                 if (__builtin_mul_overflow(node->nr_hugepages[j], hugepage_size[j], &unused))
@@ -200,12 +204,12 @@ static int sanitize_topo_info(struct pal_topo_info* topo_info) {
         }
     }
 
-    for (size_t i = 0; i < topo_info->numa_nodes_cnt; i++) {
-        /* Note: Linux doesn't guarantee that distance i -> i is 0, so we aren't checking this (it's
-         * actually non-zero on all machines we have). */
-        for (size_t j = 0; j < topo_info->numa_nodes_cnt; j++) {
-            if (   topo_info->numa_distance_matrix[i*topo_info->numa_nodes_cnt + j]
-                != topo_info->numa_distance_matrix[j*topo_info->numa_nodes_cnt + i])
+    for (size_t i = 0; i < online_nodes_cnt; i++) {
+        /* Note: distance i -> i is 10 according to the ACPI 2.0 SLIT spec, but to accomodate for
+         * weird BIOS settings we aren't checking this. */
+        for (size_t j = 0; j < online_nodes_cnt; j++) {
+            if (   topo_info->numa_distance_matrix[i*online_nodes_cnt + j]
+                != topo_info->numa_distance_matrix[j*online_nodes_cnt + i])
                 return -PAL_ERROR_INVAL;
         }
     }


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1454)* status: 0 of 4 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

___
*[`libos/src/fs/sys/node_info.c` line 81 at r1](https://reviewable.io/reviews/gramineproject/gramine/1454#-N_8KkgmBzt_jSVWNt0m:-N_8KkgmBzt_jSVWNt0n:bq1uf3r) ([raw file](https://github.com/gramineproject/gramine/blob/8f1b75ad9f9861c46992428d2e12c6f9e8318c40/libos/src/fs/sys/node_info.c#L81)):*
> ```C
>                 online_nodes_cnt++;
>         }
>         size_t* distances = topo->numa_distance_matrix + i_online * online_nodes_cnt;
> ```

When reviewing this code and this `i_online` variable, always keep in mind the following scenario:

- Node 0 is online
- Node 1 is offline
- Node 2 is online

In this case, the distances matrix "jumps over" Node 1, so e.g. `/sys/devices/system/node/node2/distance` will contain:
```
[at row 1 of numa_distance_matrix] node 2 -> node 0, node 2 -> node 2
```
___
*[`pal/src/host/linux-sgx/pal_main.c` line 209 at r1](https://reviewable.io/reviews/gramineproject/gramine/1454#-N_8LIG-63YfrV5TUcEN:-N_8LIG-63YfrV5TUcEO:bvevbmv) ([raw file](https://github.com/gramineproject/gramine/blob/8f1b75ad9f9861c46992428d2e12c6f9e8318c40/pal/src/host/linux-sgx/pal_main.c#L209)):*
> ```C
>     for (size_t i = 0; i < online_nodes_cnt; i++) {
>         /* Note: distance i -> i is 10 according to the ACPI 2.0 SLIT spec, but to accomodate for
>          * weird BIOS settings we aren't checking this. */
> ```

See e.g. https://github.com/torvalds/linux/blob/6995e2de6891c724bfeb2db33d7b87775f913ad1/include/linux/topology.h#L49

Or if you're brave enough, Chapter "5.2.17 System Locality Distance Information Table (SLIT)" in https://uefi.org/sites/default/files/resources/ACPI_6_3_final_Jan30.pdf

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1454)* status: 0 of 6 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

___
*[`libos/src/fs/sys/fs.c` line 257 at r2](https://reviewable.io/reviews/gramineproject/gramine/1454#-N_8sz0C4Qg4xc7Cp6NU:-N_8sz0C4Qg4xc7Cp6NV:bsy9h36) ([raw file](https://github.com/gramineproject/gramine/blob/db2a92e4e37acd7e5023cd728fb78df00527d14d/libos/src/fs/sys/fs.c#L257)):*
> ```C
> 
>     // TODO(mkow): Does this show up for offline nodes? I never succeeded in shutting down one, even
>     // after shutting down all CPUs inside the node it shows up as online on `node/online` list.
> ```

FYI: The answer to this question is "no", nothing shows up under offline nodes.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1454)* status: 0 of 6 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

___
*[`libos/src/fs/sys/node_info.c` line 69 at r3](https://reviewable.io/reviews/gramineproject/gramine/1454#-N_8yVoTBCUSDBBFthGA:-N_8yVoTBCUSDBBFthGB:bm3valp) ([raw file](https://github.com/gramineproject/gramine/blob/27d00c5e5f6b7b1afe9d258ab7dc7d98d05ebcb2/libos/src/fs/sys/node_info.c#L69)):*
> ```C
>     const struct pal_numa_node_info* numa_node = &topo->numa_nodes[node_id];
>     if (!numa_node->is_online)
>         return -ENOENT;
> ```

I think we actually can never hit this case, because we don't even allow `/sys/devices/system/nodeX/` directories if `nodeX` is offline. But I keep it just in case.
___
*[`libos/src/fs/sys/node_info.c` line 133 at r3](https://reviewable.io/reviews/gramineproject/gramine/1454#-N_8yful-Bge2A0l0pLZ:-N_8yful-Bge2A0l0pL_:bm3valp) ([raw file](https://github.com/gramineproject/gramine/blob/27d00c5e5f6b7b1afe9d258ab7dc7d98d05ebcb2/libos/src/fs/sys/node_info.c#L133)):*
> ```C
>     const struct pal_numa_node_info* numa_node = &topo->numa_nodes[node_id];
>     if (!numa_node->is_online)
>         return -ENOENT;
> ```

I think we actually can never hit this case, because we don't even allow `/sys/devices/system/nodeX/` directories if `nodeX` is offline. But I keep it just in case.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1454)* status: 0 of 8 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

___
*[`-- commits` line 22 at r4](https://reviewable.io/reviews/gramineproject/gramine/1454#-N_94Bre0wX_8-ZSb6Dd):*
This second commit is a bit unrelated, but I noticed that it's missing and it's very easy to add.

In particular, this broke the `lscpu` tool when put under Gramine. Here's a very simple diff to run `lscpu` under gramine-direct (on a RHEL-based system):
```diff
diff --git a/CI-Examples/helloworld/helloworld.manifest.template b/CI-Examples/helloworld/helloworld.manifest.template
index e15f0d2b..41a7f47f 100644
--- a/CI-Examples/helloworld/helloworld.manifest.template
+++ b/CI-Examples/helloworld/helloworld.manifest.template
@@ -1,14 +1,15 @@
 # Hello World manifest file example

 loader.entrypoint = "file:{{ gramine.libos }}"
-libos.entrypoint = "/helloworld"
+libos.entrypoint = "/lscpu"
 loader.log_level = "{{ log_level }}"

-loader.env.LD_LIBRARY_PATH = "/lib"
+loader.env.LD_LIBRARY_PATH = "/lib:/lib64"

 fs.mounts = [
   { path = "/lib", uri = "file:{{ gramine.runtimedir() }}" },
-  { path = "/helloworld", uri = "file:helloworld" },
+  { path = "/lscpu", uri = "file:/usr/bin/lscpu" },
+  { path = "/lib64", uri = "file:/lib64" },
 ]

 sgx.debug = true
```

---

For `lscpu` code that uses `present` file, see:
- https://github.com/util-linux/util-linux/blob/0fc338b0360a2479cc4144aba481aae1aa240c3c/sys-utils/lscpu.c#L1019
- https://github.com/util-linux/util-linux/blob/0fc338b0360a2479cc4144aba481aae1aa240c3c/sys-utils/lscpu-cputype.c#L716-L717
___
*[`libos/src/fs/sys/cpu_info.c` line 41 at r4](https://reviewable.io/reviews/gramineproject/gramine/1454#-N_94rDG5CL36bzbwtAP:-N_94rDG5CL36bzbwtAQ:bgfot2y) ([raw file](https://github.com/gramineproject/gramine/blob/2bbd9de9f991fee61bb9be523c4ad28bd444c61d/libos/src/fs/sys/cpu_info.c#L41)):*
> ```C
>         ret = sys_print_as_ranges(str, sizeof(str), topo->threads_cnt, is_offline, NULL);
>     } else if (strcmp(name, "possible") == 0 || strcmp(name, "present") == 0) {
>         /* we simplify and always output "present" CPUs same as "possible" CPUs */
> ```

If you wonder why we can assume that `present` is the same as `possible`: that's because `present` is always a subset of `possible`, and always a superset of `online`. See:
- https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-devices-system-cpu
- https://www.kernel.org/doc/Documentation/admin-guide/cputopology.rst

I only use `present` to fix `lscpu` tool output, see my other comment.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1454)* status: 0 of 8 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

___
*[`-- commits` line 31 at r5](https://reviewable.io/reviews/gramineproject/gramine/1454#-N_9C9pA5gWcHkInK0Fs):*
Very simple commit, adds a hard-coded value typically found on systems (~8K).

Some refs:
- https://github.com/torvalds/linux/blob/3f01e9fed8454dcd89727016c3e5b2fbb8f8e50c/drivers/base/cpu.c#L217
- https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-devices-system-cpu

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1454)* status: 0 of 9 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

___
*[`-- commits` line 36 at r6](https://reviewable.io/reviews/gramineproject/gramine/1454#-N_A5DACAIPwHBcXU9Bq):*
Another very simple commit.

Unfortunately, I cannot show the source codes of offending libraries. But I debugged this lib: https://packages.ubuntu.com/search?keywords=libmkl-gnu-thread.

I tested manually on the offending machine, and it works fine now.

kailun-qin (changes requested):
> Reviewed 3 of 4 files at r1, 2 of 3 files at r3, 1 of 3 files at r4, 2 of 2 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1454)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`libos/src/fs/sys/fs.c` line 225 at r6](https://reviewable.io/reviews/gramineproject/gramine/1454#-N_C7UJ77WQkq2p4DImF:-N_C7UJ77WQkq2p4DImG:bbk9xfe) ([raw file](https://github.com/gramineproject/gramine/blob/c194fc7ebfb9b5b03f67080fc03843a288bbf820/libos/src/fs/sys/fs.c#L225)):*
> ```C
>         log_warning("Actual number of CPUs (%lu) is greater than the value (%u) hard-coded into "
>                     "/sys/devices/system/cpu/kernel_max. This may confuse apps that rely on the "
>                     "value in this file.", g_pal_public_state->topo_info.threads_cnt,
> ```

This is not actionable for users. Besides, is this happening only w/ weird BIOS settings? If so, but we simply accomodate for them instead of checking/warning in the other places?



_Code quote:_
```C
                    "/sys/devices/system/cpu/kernel_max. This may confuse apps that rely on the "
                    "value in this file.", g_pal_public_state->topo_info.threads_cnt,
```
___
*[`pal/include/arch/x86_64/pal_topology.h` line 114 at r6](https://reviewable.io/reviews/gramineproject/gramine/1454#-N_C2tNU0pK9FTuIX1e0:-N_C2tNU0pK9FTuIX1e1:b-w2nchr) ([raw file](https://github.com/gramineproject/gramine/blob/c194fc7ebfb9b5b03f67080fc03843a288bbf820/pal/include/arch/x86_64/pal_topology.h#L114)):*
> ```objc
> 
>     /*
>      * Has `online_numa_nodes_cnt * online_numa_nodes_cnt` elements, where `online_numa_nodes_cnt`
> ```

-> `non-zero/valid elements` (since it's actually an overapproximate)?



_Code quote:_
```objc
elements
```
___
*[`pal/include/arch/x86_64/pal_topology.h` line 121 at r6](https://reviewable.io/reviews/gramineproject/gramine/1454#-N_C5Gf0-JYP54EQsn7F:-N_C5Gf0-JYP54EQsn7G:bmrz1pc) ([raw file](https://github.com/gramineproject/gramine/blob/c194fc7ebfb9b5b03f67080fc03843a288bbf820/pal/include/arch/x86_64/pal_topology.h#L121)):*
> ```objc
>      *
>      * It may be surprising that this array contains only online nodes, but this is what Linux does:
>      *   https://github.com/torvalds/linux/blob/v6.4/drivers/base/node.c#L543-L564
> ```

Are we more in favor of `bootlin` (https://elixir.bootlin.com/linux/v6.4/source/drivers/base/node.c#L543)?



_Code quote:_
```objc
https://github.com/torvalds/linux/blob/v6.4/drivers/base/node.c#L543-L564
```

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1454)* status: 7 of 9 files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

___
*[`libos/src/fs/sys/fs.c` line 225 at r6](https://reviewable.io/reviews/gramineproject/gramine/1454#-N_C7UJ77WQkq2p4DImF:-N_DUOcx11mP9sU7B2bH:b146cfg) ([raw file](https://github.com/gramineproject/gramine/blob/c194fc7ebfb9b5b03f67080fc03843a288bbf820/libos/src/fs/sys/fs.c#L225)):*
Done. I just update this `kernel_max` value, since Gramine doesn't really have an upper limit on the number of CPU cores.

> If so, but we simply accomodate for them instead of checking/warning in the other places?

I didn't understand your question.
___
*[`pal/include/arch/x86_64/pal_topology.h` line 114 at r6](https://reviewable.io/reviews/gramineproject/gramine/1454#-N_C2tNU0pK9FTuIX1e0:-N_DUq16CZC4MBK-JnpD:b2nl9ah) ([raw file](https://github.com/gramineproject/gramine/blob/c194fc7ebfb9b5b03f67080fc03843a288bbf820/pal/include/arch/x86_64/pal_topology.h#L114)):*

-> `non-zero/valid elements` (since it's actually an overapproximate)?
</blockquote></details>

Done. Added `valid` to make it more explicit (I hope, or is it confusing?)
___
*[`pal/include/arch/x86_64/pal_topology.h` line 121 at r6](https://reviewable.io/reviews/gramineproject/gramine/1454#-N_C5Gf0-JYP54EQsn7F:-N_DUzgiDdR3czMpUdgF:bs1vbdg) ([raw file](https://github.com/gramineproject/gramine/blob/c194fc7ebfb9b5b03f67080fc03843a288bbf820/pal/include/arch/x86_64/pal_topology.h#L121)):*

Are we more in favor of `bootlin` (https://elixir.bootlin.com/linux/v6.4/source/drivers/base/node.c#L543)?
</blockquote></details>

Done. Right, though I started using GitHub much more actively because it now has the cool feature of finding symbols.

kailun-qin (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1454)* status: 7 of 9 files reviewed, 10 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`libos/src/fs/sys/fs.c` line 225 at r6](https://reviewable.io/reviews/gramineproject/gramine/1454#-N_C7UJ77WQkq2p4DImF:-N_GjmYk-lCNgMxkX1QI:b-ft03xh) ([raw file](https://github.com/gramineproject/gramine/blob/c194fc7ebfb9b5b03f67080fc03843a288bbf820/libos/src/fs/sys/fs.c#L225)):*
> I didn't understand your question.

Forget about it, it's not relevant anymore. I'm fine w/ the current approach, thanks!
___
*[`libos/src/fs/sys/fs.c` line 16 at r7](https://reviewable.io/reviews/gramineproject/gramine/1454#-N_Gu-E4CI_8YZp5tZ0E:-N_Gu-E4CI_8YZp5tZ0F:b-yt0oz6) ([raw file](https://github.com/gramineproject/gramine/blob/244f0bdf436493dce3afb93e8332074ca84980cd/libos/src/fs/sys/fs.c#L16)):*
> ```C
> #include "libos_fs_pseudo.h"
> 
> #define DEFAULT_KERNEL_MAX_CPUS 8191
> ```

The name is a bit misleading as 8191 is actually the max **index**? I understand this follows Linux kernel and I don't insist.



_Code quote:_
```C
DEFAULT_KERNEL_MAX_CPUS
```
___
*[`libos/src/fs/sys/fs.c` line 229 at r7](https://reviewable.io/reviews/gramineproject/gramine/1454#-N_Gr5Bl2-JeoxYgA0I-:-N_Gr5Bl2-JeoxYgA0I0:bgmuymg) ([raw file](https://github.com/gramineproject/gramine/blob/244f0bdf436493dce3afb93e8332074ca84980cd/libos/src/fs/sys/fs.c#L229)):*
> ```C
> 
> static void init_cpu_dir(struct pseudo_node* cpu) {
>     if (g_pal_public_state->topo_info.threads_cnt > DEFAULT_KERNEL_MAX_CPUS) {
> ```

Here we compare `threads_cnt` w/ `kernel_max` which is the maximum cpu **index** allowed, shouldn't we compare w/ `max_index + 1`?
___
*[`libos/src/fs/sys/fs.c` line 229 at r7](https://reviewable.io/reviews/gramineproject/gramine/1454#-N_GtBYI9_zFlanxndhH:-N_GtBYI9_zFlanxndhI:b-i8g4aa) ([raw file](https://github.com/gramineproject/gramine/blob/244f0bdf436493dce3afb93e8332074ca84980cd/libos/src/fs/sys/fs.c#L229)):*
> ```C
> 
> static void init_cpu_dir(struct pseudo_node* cpu) {
>     if (g_pal_public_state->topo_info.threads_cnt > DEFAULT_KERNEL_MAX_CPUS) {
> ```

Actually, maybe we can even remove the IF?

Perhaps something like:
```
static void init_cpu_dir(struct pseudo_node* cpu) {
    uint32_t next_power_of_2 = DEFAULT_KERNEL_MAX_CPUS + 1;
    while (next_power_of_2 < g_pal_public_state->topo_info.threads_cnt) {
        /* this should never happen in real systems, so BUG() railguard is enough */
        if (__builtin_mul_overflow(next_power_of_2, 2, &next_power_of_2))
            BUG();
    }
    /* `- 1` because Linux always reports kernel_max like this */
    g_kernel_max_cpus = next_power_of_2 - 1;
}
```
___
*[`libos/src/fs/sys/fs.c` line 231 at r7](https://reviewable.io/reviews/gramineproject/gramine/1454#-N_Gucrk3fmil16l_NRU:-N_Gucrk3fmil16l_NRV:b-983ife) ([raw file](https://github.com/gramineproject/gramine/blob/244f0bdf436493dce3afb93e8332074ca84980cd/libos/src/fs/sys/fs.c#L231)):*
> ```C
>     if (g_pal_public_state->topo_info.threads_cnt > DEFAULT_KERNEL_MAX_CPUS) {
>         assert(g_kernel_max_cpus == DEFAULT_KERNEL_MAX_CPUS);
>         uint32_t next_power_of_2 = 1;
> ```

Can't we just start from `DEFAULT_KERNEL_MAX_CPUS + 1` (as we control it completely)?



_Code quote:_
```C
uint32_t next_power_of_2 = 1;
```
___
*[`pal/include/arch/x86_64/pal_topology.h` line 114 at r6](https://reviewable.io/reviews/gramineproject/gramine/1454#-N_C2tNU0pK9FTuIX1e0:-N_GhQ05-C5Npp5yYrSI:b1r9tjg) ([raw file](https://github.com/gramineproject/gramine/blob/c194fc7ebfb9b5b03f67080fc03843a288bbf820/pal/include/arch/x86_64/pal_topology.h#L114)):*

Done. Added `valid` to make it more explicit (I hope, or is it confusing?)
</blockquote></details>

It's more explicit to me, thanks!

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1454)* status: 7 of 9 files reviewed, 9 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

___
*[`libos/src/fs/sys/fs.c` line 16 at r7](https://reviewable.io/reviews/gramineproject/gramine/1454#-N_Gu-E4CI_8YZp5tZ0E:-N_YTPcM5EPiSFzj7jG_:b-hy76ue) ([raw file](https://github.com/gramineproject/gramine/blob/244f0bdf436493dce3afb93e8332074ca84980cd/libos/src/fs/sys/fs.c#L16)):*

The name is a bit misleading as 8191 is actually the max **index**? I understand this follows Linux kernel and I don't insist.
</blockquote></details>

Done. I'm keeping the name but it's not refering to the **index** now, but instead to the **max number**.
___
*[`libos/src/fs/sys/fs.c` line 229 at r7](https://reviewable.io/reviews/gramineproject/gramine/1454#-N_GtBYI9_zFlanxndhH:-N_YU5T1CVBxqUVf1NgX:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/244f0bdf436493dce3afb93e8332074ca84980cd/libos/src/fs/sys/fs.c#L229)):*

Actually, maybe we can even remove the IF?

Perhaps something like:
```
static void init_cpu_dir(struct pseudo_node* cpu) {
    uint32_t next_power_of_2 = DEFAULT_KERNEL_MAX_CPUS + 1;
    while (next_power_of_2 < g_pal_public_state->topo_info.threads_cnt) {
        /* this should never happen in real systems, so BUG() railguard is enough */
        if (__builtin_mul_overflow(next_power_of_2, 2, &next_power_of_2))
            BUG();
    }
    /* `- 1` because Linux always reports kernel_max like this */
    g_kernel_max_cpus = next_power_of_2 - 1;
}
```
</blockquote></details>

Done.
___
*[`libos/src/fs/sys/fs.c` line 229 at r7](https://reviewable.io/reviews/gramineproject/gramine/1454#-N_Gr5Bl2-JeoxYgA0I-:-N_YU6qODf_QcBsah2RR:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/244f0bdf436493dce3afb93e8332074ca84980cd/libos/src/fs/sys/fs.c#L229)):*

Here we compare `threads_cnt` w/ `kernel_max` which is the maximum cpu **index** allowed, shouldn't we compare w/ `max_index + 1`?
</blockquote></details>

Done.
___
*[`libos/src/fs/sys/fs.c` line 231 at r7](https://reviewable.io/reviews/gramineproject/gramine/1454#-N_Gucrk3fmil16l_NRU:-N_YU6AKDGVAmo45bfT9:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/244f0bdf436493dce3afb93e8332074ca84980cd/libos/src/fs/sys/fs.c#L231)):*

Can't we just start from `DEFAULT_KERNEL_MAX_CPUS + 1` (as we control it completely)?
</blockquote></details>

Done.

kailun-qin (dismissed):
> Reviewed 1 of 2 files at r7, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1454)* status: all files reviewed, 6 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1454)* status: all files reviewed, 7 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`libos/src/fs/sys/fs.c` line 257 at r2](https://reviewable.io/reviews/gramineproject/gramine/1454#-N_8sz0C4Qg4xc7Cp6NU:-NabPQcoCGINepxvwVzy:bmw8mtt) ([raw file](https://github.com/gramineproject/gramine/blob/db2a92e4e37acd7e5023cd728fb78df00527d14d/libos/src/fs/sys/fs.c#L257)):*

FYI: The answer to this question is "no", nothing shows up under offline nodes.
</blockquote></details>

So, how did you actually manage to shut down a NUMA node?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1454)* status: all files reviewed, 7 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`libos/src/fs/sys/fs.c` line 257 at r2](https://reviewable.io/reviews/gramineproject/gramine/1454#-N_8sz0C4Qg4xc7Cp6NU:-NaeeIPB6yOZ0f7EieCo:b-ysjkoj) ([raw file](https://github.com/gramineproject/gramine/blob/db2a92e4e37acd7e5023cd728fb78df00527d14d/libos/src/fs/sys/fs.c#L257)):*

So, how did you actually manage to shut down a NUMA node?
</blockquote></details>

I didn't shut down NUMA nodes, the Linux kernel simply started on this platform with some NUMA nodes being offline.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1454)* status: all files reviewed, 7 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`-- commits` line 22 at r4](https://reviewable.io/reviews/gramineproject/gramine/1454#-N_94Bre0wX_8-ZSb6Dd):*

This second commit is a bit unrelated, but I noticed that it's missing and it's very easy to add.

In particular, this broke the `lscpu` tool when put under Gramine. Here's a very simple diff to run `lscpu` under gramine-direct (on a RHEL-based system):
```diff
diff --git a/CI-Examples/helloworld/helloworld.manifest.template b/CI-Examples/helloworld/helloworld.manifest.template
index e15f0d2b..41a7f47f 100644
--- a/CI-Examples/helloworld/helloworld.manifest.template
+++ b/CI-Examples/helloworld/helloworld.manifest.template
@@ -1,14 +1,15 @@
 # Hello World manifest file example

 loader.entrypoint = "file:{{ gramine.libos }}"
-libos.entrypoint = "/helloworld"
+libos.entrypoint = "/lscpu"
 loader.log_level = "{{ log_level }}"

-loader.env.LD_LIBRARY_PATH = "/lib"
+loader.env.LD_LIBRARY_PATH = "/lib:/lib64"

 fs.mounts = [
   { path = "/lib", uri = "file:{{ gramine.runtimedir() }}" },
-  { path = "/helloworld", uri = "file:helloworld" },
+  { path = "/lscpu", uri = "file:/usr/bin/lscpu" },
+  { path = "/lib64", uri = "file:/lib64" },
 ]

 sgx.debug = true
```

---

For `lscpu` code that uses `present` file, see:
- https://github.com/util-linux/util-linux/blob/0fc338b0360a2479cc4144aba481aae1aa240c3c/sys-utils/lscpu.c#L1019
- https://github.com/util-linux/util-linux/blob/0fc338b0360a2479cc4144aba481aae1aa240c3c/sys-utils/lscpu-cputype.c#L716-L717
</blockquote></details>

This second commit was extracted in a separate PR: https://github.com/gramineproject/gramine/pull/1476

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1454)* status: all files reviewed, 7 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`-- commits` line 36 at r6](https://reviewable.io/reviews/gramineproject/gramine/1454#-N_A5DACAIPwHBcXU9Bq):*

Another very simple commit.

Unfortunately, I cannot show the source codes of offending libraries. But I debugged this lib: https://packages.ubuntu.com/search?keywords=libmkl-gnu-thread.

I tested manually on the offending machine, and it works fine now.
</blockquote></details>

This fourth commit was extracted in a separate PR: https://github.com/gramineproject/gramine/pull/1477

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1454)* status: all files reviewed, 7 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`-- commits` line 31 at r5](https://reviewable.io/reviews/gramineproject/gramine/1454#-N_9C9pA5gWcHkInK0Fs):*

Very simple commit, adds a hard-coded value typically found on systems (~8K).

Some refs:
- https://github.com/torvalds/linux/blob/3f01e9fed8454dcd89727016c3e5b2fbb8f8e50c/drivers/base/cpu.c#L217
- https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-devices-system-cpu
</blockquote></details>

Extracted into a separate PR: https://github.com/gramineproject/gramine/pull/1478

mkow (commented):
> Reviewed 1 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1454)* status: all files reviewed, 7 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`-- commits` line 22 at r4](https://reviewable.io/reviews/gramineproject/gramine/1454#-N_94Bre0wX_8-ZSb6Dd):*

This second commit was extracted in a separate PR: https://github.com/gramineproject/gramine/pull/1476
</blockquote></details>

@dimakuv: Could you rebase this PR to drop the commits which were already merged?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1454)* status: 4 of 9 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @kailun-qin and @mkow)

___
*[`-- commits` line 22 at r4](https://reviewable.io/reviews/gramineproject/gramine/1454#-N_94Bre0wX_8-ZSb6Dd):*

@dimakuv: Could you rebase this PR to drop the commits which were already merged?
</blockquote></details>

Done, rebased to latest master and dropped the unrelated commits.

mkow (changes requested):
> Reviewed 1 of 4 files at r1, 1 of 3 files at r3, 1 of 2 files at r7, 5 of 5 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1454)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1454#-NdXfw0Q4fY2TTPK3j6U:-NdXfw0Q4fY2TTPK3j6V:b-plhipe) (no related file):*
TODO for myself: test this on a system where this reproduces.

___
___
*[`pal/include/arch/x86_64/pal_topology.h` line 114 at r9](https://reviewable.io/reviews/gramineproject/gramine/1454#-NdbVfUnDeopJXzyiHqM:-NdbVfUnDeopJXzyiHqN:bb9mo71) ([raw file](https://github.com/gramineproject/gramine/blob/ccd4e359734f06ecf020c3bf8496ee7698a6a4c6/pal/include/arch/x86_64/pal_topology.h#L114)):*
> ```objc
> 
>     /*
>      * Has `online_numa_nodes_cnt * online_numa_nodes_cnt` valid elements, where
> ```

Valid? Does it also have some invalid elements?

// update: ok, it does, but this sentence is confusing (as it only mentions the _number_ of valid elements, not which are valid and which are not)



_Code quote:_
```objc
valid elements
```
___
*[`pal/include/arch/x86_64/pal_topology.h` line 118 at r9](https://reviewable.io/reviews/gramineproject/gramine/1454#-NdbW3T40rr5y8pb8HiE:-NdbW3T40rr5y8pb8HiF:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/ccd4e359734f06ecf020c3bf8496ee7698a6a4c6/pal/include/arch/x86_64/pal_topology.h#L118)):*
> ```objc
>      *
>      * Value in `numa_distance_matrix[i*online_numa_nodes_cnt + j]` is a NUMA distance from online
>      * node i to online node j.
> ```





_Suggestion:_
```objc
node i to online node j (i.e. indexing only through the online nodes).
```
___
*[`pal/src/host/linux-common/topo_info.c` line 387 at r9](https://reviewable.io/reviews/gramineproject/gramine/1454#-NdbYjvK9ISUitSz7Izv:-NdbYjvK9ISUitSz7Izw:brx2szp) ([raw file](https://github.com/gramineproject/gramine/blob/ccd4e359734f06ecf020c3bf8496ee7698a6a4c6/pal/src/host/linux-common/topo_info.c#L387)):*
> ```C
>      *
>      *   [ node 0 -> node 0, node 0 -> node 2
>      *     node 2 -> node 0, node 2 -> node 2 ]
> ```

Wouldn't the code here and at the printing site be much easier if we stored `distances` in a non-crazy way (just setting entries for offline nodes to 0) and would do this weird stuff _only_ when printing it to sysfs? This would also remove the need for most of the explanation comments, only the printing side would need them :)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1454)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

___
*[`pal/src/host/linux-common/topo_info.c` line 387 at r9](https://reviewable.io/reviews/gramineproject/gramine/1454#-NdbYjvK9ISUitSz7Izv:-Ndd_xR76rmdXEuVl_ye:bjg8nty) ([raw file](https://github.com/gramineproject/gramine/blob/ccd4e359734f06ecf020c3bf8496ee7698a6a4c6/pal/src/host/linux-common/topo_info.c#L387)):*

Wouldn't the code here and at the printing site be much easier if we stored `distances` in a non-crazy way (just setting entries for offline nodes to 0) and would do this weird stuff _only_ when printing it to sysfs? This would also remove the need for most of the explanation comments, only the printing side would need them :)
</blockquote></details>

Two counter-arguments to your proposal:
1. The host files `/sys/devices/system/node/nodeXXX/distance` have distances to online-only nodes. So in this code snippet, we'll have to (i) first read this file into a temporary string, (ii) then add `0`s to represent offline nodes, and (iii) save this constructed string into the `distances` array.
   - We will complicate this code snippet.
   - We can revert changes to `sanitize_topo_info()`. That's the only place that will get simplified.
   -  Is it really worth it? We exchange one complexity for another complexity.
2. With your proposed change, Gramine's `distance` is **not** the same as Linux's `distance`. Won't it be confusing for people that Gramine has a different representation of distances?

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1454)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv and @kailun-qin)

___
*[`pal/src/host/linux-common/topo_info.c` line 387 at r9](https://reviewable.io/reviews/gramineproject/gramine/1454#-NdbYjvK9ISUitSz7Izv:-NdgrIdv8sbMdBGiNk2w:b-awn103) ([raw file](https://github.com/gramineproject/gramine/blob/ccd4e359734f06ecf020c3bf8496ee7698a6a4c6/pal/src/host/linux-common/topo_info.c#L387)):*

Two counter-arguments to your proposal:
1. The host files `/sys/devices/system/node/nodeXXX/distance` have distances to online-only nodes. So in this code snippet, we'll have to (i) first read this file into a temporary string, (ii) then add `0`s to represent offline nodes, and (iii) save this constructed string into the `distances` array.
   - We will complicate this code snippet.
   - We can revert changes to `sanitize_topo_info()`. That's the only place that will get simplified.
   -  Is it really worth it? We exchange one complexity for another complexity.
2. With your proposed change, Gramine's `distance` is **not** the same as Linux's `distance`. Won't it be confusing for people that Gramine has a different representation of distances?
</blockquote></details>

1. Hmm, no, I'd just change `read_numbers_from_file()` to skip writes (and rename it).
2. I think it's much more intuitive to have it stored normally (i.e. that indices in it are `node_id`s). Also, the description of the array would be much simpler.
3. We'll need to keep the whole table anyways if we ever add support for hot-plugging.
4. There are more places which gets simpler:
   - the printing in node_info.c - you don't need any calculations anymore, just skip printing offline entries.
   - less explanatory comments needed everywhere

The only thing I don't like about this idea is that change to `read_numbers_from_file()`, which sounds ugly. But it's untrusted code, and it makes trusted code a bit simpler...

I won't be blocking on this, but let's maybe also ask @kailun-qin what he thinks about it.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1454)* status: 8 of 9 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin and @mkow)

___
*[`pal/include/arch/x86_64/pal_topology.h` line 114 at r9](https://reviewable.io/reviews/gramineproject/gramine/1454#-NdbVfUnDeopJXzyiHqM:-NdiVy6w4fYAKU2n_lfE:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/ccd4e359734f06ecf020c3bf8496ee7698a6a4c6/pal/include/arch/x86_64/pal_topology.h#L114)):*

Valid? Does it also have some invalid elements?

// update: ok, it does, but this sentence is confusing (as it only mentions the _number_ of valid elements, not which are valid and which are not)
</blockquote></details>

Done.
___
*[`pal/include/arch/x86_64/pal_topology.h` line 118 at r9](https://reviewable.io/reviews/gramineproject/gramine/1454#-NdbW3T40rr5y8pb8HiE:-NdiW59Y6CVDt2P_SMv0:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/ccd4e359734f06ecf020c3bf8496ee7698a6a4c6/pal/include/arch/x86_64/pal_topology.h#L118)):*
> ```objc
>      *
>      * Value in `numa_distance_matrix[i*online_numa_nodes_cnt + j]` is a NUMA distance from online
>      * node i to online node j.
> ```

Done.
___
*[`pal/src/host/linux-common/topo_info.c` line 387 at r9](https://reviewable.io/reviews/gramineproject/gramine/1454#-NdbYjvK9ISUitSz7Izv:-NdiUdfL9swqDiToNj3Q:btsq72j) ([raw file](https://github.com/gramineproject/gramine/blob/ccd4e359734f06ecf020c3bf8496ee7698a6a4c6/pal/src/host/linux-common/topo_info.c#L387)):*
> Hmm, no, I'd just change `read_numbers_from_file()` to skip writes (and rename it).

I don't understand what you mean by this.

Let's wait for @kailun-qin's opinion.

mkow (commented):
> Reviewed 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1454)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @kailun-qin)

___
*[`pal/src/host/linux-common/topo_info.c` line 387 at r9](https://reviewable.io/reviews/gramineproject/gramine/1454#-NdbYjvK9ISUitSz7Izv:-Ndj_v4n7t0LvZqJ5C6m:b-xx8lqd) ([raw file](https://github.com/gramineproject/gramine/blob/ccd4e359734f06ecf020c3bf8496ee7698a6a4c6/pal/src/host/linux-common/topo_info.c#L387)):*
> > Hmm, no, I'd just change `read_numbers_from_file()` to skip writes (and rename it).
> 
> I don't understand what you mean by this.

This was a response to "we'll have to (i) first read this file into a temporary string, (ii) then add `0`s to represent offline nodes, and (iii) save this constructed string into the `distances` array." - the alternative is to change the parsing function to write the results skipping offline nodes.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1454)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @kailun-qin)

___
*[`pal/src/host/linux-common/topo_info.c` line 387 at r9](https://reviewable.io/reviews/gramineproject/gramine/1454#-NdbYjvK9ISUitSz7Izv:-NdoGNe93mJHBGDcHesN:b-4loo3x) ([raw file](https://github.com/gramineproject/gramine/blob/ccd4e359734f06ecf020c3bf8496ee7698a6a4c6/pal/src/host/linux-common/topo_info.c#L387)):*

> the alternative is to change the parsing function to write the results skipping offline nodes.

Still doesn't make sense to me. Host files `/sys/devices/system/node/nodeXXX/distance` already do **not** have any offline nodes, so what's there to skip?
</blockquote></details>

I mean, you don't skip them when reading, but when writing the parsing results to the resulting array.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1454)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

___
*[`pal/src/host/linux-common/topo_info.c` line 387 at r9](https://reviewable.io/reviews/gramineproject/gramine/1454#-NdbYjvK9ISUitSz7Izv:-NdyUX5i7NSyaxo-Dy2V:b-sl6gmb) ([raw file](https://github.com/gramineproject/gramine/blob/ccd4e359734f06ecf020c3bf8496ee7698a6a4c6/pal/src/host/linux-common/topo_info.c#L387)):*

I mean, you don't skip them when reading, but when writing the parsing results to the resulting array.
</blockquote></details>

Done in separate PR: #1545. See there, if you like that definition of `distances` array better.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1454)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @kailun-qin)

___
*[`pal/src/host/linux-common/topo_info.c` line 387 at r9](https://reviewable.io/reviews/gramineproject/gramine/1454#-NdbYjvK9ISUitSz7Izv:-NdzzVa-1lpUrWn8vbrz:b-7rgnl8) ([raw file](https://github.com/gramineproject/gramine/blob/ccd4e359734f06ecf020c3bf8496ee7698a6a4c6/pal/src/host/linux-common/topo_info.c#L387)):*

Done in separate PR: #1545. See there, if you like that definition of `distances` array better.
</blockquote></details>

I prefer #1545, mostly because it has simpler data representation. But let's wait for @kailun-qin's opinion :)

kailun-qin (commented):
> Reviewed 1 of 5 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1454)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`pal/src/host/linux-common/topo_info.c` line 387 at r9](https://reviewable.io/reviews/gramineproject/gramine/1454#-NdbYjvK9ISUitSz7Izv:-Ne1JNrB1gxv3GaY7ebx:bs2svx7) ([raw file](https://github.com/gramineproject/gramine/blob/ccd4e359734f06ecf020c3bf8496ee7698a6a4c6/pal/src/host/linux-common/topo_info.c#L387)):*
I vote for https://github.com/gramineproject/gramine/pull/1545 because the two counter-arguments do not really hold up IMO:
> We will complicate this code snippet. Is it really worth it? We exchange one complexity for another complexity.

Based on the lataest impl. from #1545, it seems to be not complex (and slightly easier to understand).

> Gramine's `distance` is **not** the same as Linux's `distance`. Won't it be confusing for people that Gramine has a different representation of distances?

Does Gramine's `distance` has to follow the same as Linux's `distance`? I suppose it's rather implementation dependent for what's the `distance` representation between nodes that have one or both offline.

However, we can still get below benefits if we go w/ #1545:
> It's much more intuitive to have it stored normally (i.e. that indices in it are `node_id`s). We'll need to keep the whole table anyways if we ever add support for hot-plugging. There are more places which gets simpler.

dimakuv:
Everyone agreed to proceed with #1545, closing this PR

PR #1457, opened 2023-07-17
[common] Do not expand `__FILE_NAME__` to our func call

### LINKED ISSUE #1456: [common] Performance bug on older GCC/Clang because of always calling `truncate_file_name()`

We have this function: https://github.com/gramineproject/gramine/blob/db2a92e4e37acd7e5023cd728fb78df00527d14d/common/include/log.h#L17-L33

This function is used in the macro: https://github.com/gramineproject/gramine/blob/db2a92e4e37acd7e5023cd728fb78df00527d14d/common/include/log.h#L35

...and that macro is a function argument to our log functions: https://github.com/gramineproject/gramine/blob/db2a92e4e37acd7e5023cd728fb78df00527d14d/common/include/log.h#L39-L44

**Performance bug**: if your GCC/Clang version is too old and does *not* have the built-in `__FILE_NAME__` macro, then **any** log-function call, **even if logging is disabled in the manifest**, ends up calling `truncate_file_name()`.

### Steps to reproduce

Build Gramine on an older system (e.g. Ubuntu 20.04 which has GCC 10.3), and put a breakpoint on `truncate_file_name()`. You'll see how it's called many-many times.

### Expected results

`truncate_file_name()` must not be called on not-enabled-at-runtime log messages.

### Actual results

`truncate_file_name()` **is** called on not-enabled-at-runtime log messages.

Here's a way to quickly check the bug on Ubuntu 20.04 (after building Gramine manually):
```
CI-Examples/helloworld$ GDB=1 gramine-direct helloworld

(gdb) b truncate_file_name
(gdb) ignore 1 100000

(gdb) run
...
Hello, world
[LWP 3943654 exited]
[Inferior 1 (process 3943643) exited normally]

(gdb) info breakpoints
1       breakpoint     keep y   <MULTIPLE>
        breakpoint already hit 174 times          # OOPS!!!
```

### Gramine commit hash

db2a92e4e37acd7e5023cd728fb78df00527d14d

dimakuv:
## Description of the changes 

Previously we had a function that emulated `__FILE_NAME__` macro (which was introduced only recently in GCC12 and Clang9) by truncating the file name in a loop. It led to unnecessary calls to this func even when logs were disabled, ultimately resulting in perf overhead. We observed around 5% overhead because of this on file-I/O intensive benchmarks (because there we print a lot of log_debug info).

This PR simply falls back to a more clunky `__FILE__` macro when `__FILE_NAME__` is not defined.

Fixes #1456.

## How to test this PR? 

See how I test in https://github.com/gramineproject/gramine/issues/1456. With this PR, manual inspection. You may also want to run some simple FS benchmark (like calling `read()` syscall in a loop) and observe perf overhead, e.g. via `perf record`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1457)

### DIFF
--- a/common/include/log.h
+++ b/common/include/log.h
@@ -14,26 +14,45 @@ enum {
     LOG_LEVEL_ALL     = 5,
 };
 
-/*
- * __FILE_NAME__ was introduced in GCC12 and clang9.
- * If it's not defined we have to do our own magic.
- */
 #ifndef __FILE_NAME__
+#ifndef __OPTIMIZE__
+/* __FILE_NAME__ was introduced in GCC12 and clang9. If it's not defined we do our own magic. */
 static inline const char* truncate_file_name(const char* filename) {
     const char* ret = filename;
-
     while (*filename != '\0') {
         if (*filename == '/') {
             ret = filename + 1;
         }
         filename++;
     }
-
     return ret;
 }
-
 #define __FILE_NAME__ (truncate_file_name(__FILE__))
-#endif
+#else // ifndef __OPTIMIZE__
+/* Below macros will be optimized at compile-time, incurring no perf overhead at runtime. */
+#define TRY_TRUNCATE_FILE_NAME_AT_LOC(s, loc) \
+    (sizeof(s) >= (loc) && (s)[sizeof(s)-(loc)] == '/') ? ((s) + sizeof(s) - (loc) + 1)
+#define TRUNCATE_FILE_NAME(s) \
+    (TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 1)  : TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 2)  : \
+     TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 3)  : TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 4)  : \
+     TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 5)  : TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 6)  : \
+     TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 7)  : TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 8)  : \
+     TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 9)  : TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 10) : \
+     TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 11) : TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 12) : \
+     TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 13) : TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 14) : \
+     TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 15) : TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 16) : \
+     TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 17) : TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 18) : \
+     TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 19) : TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 20) : \
+     TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 21) : TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 22) : \
+     TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 23) : TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 24) : \
+     TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 25) : TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 26) : \
+     TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 27) : TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 28) : \
+     TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 29) : TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 30) : \
+     TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 31) : TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 32) : \
+     TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 33) : TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 34) : (s))
+#define __FILE_NAME__ (TRUNCATE_FILE_NAME(__FILE__))
+#endif // ifndef __OPTIMIZE__
+#endif // #ifndef __FILE_NAME__
 
 /* All of them implicitly append a newline at the end of the message. */
 #define log_always(fmt...)   _log(LOG_LEVEL_NONE, __FILE_NAME__, __func__, __LINE__, fmt)


ynonflumintel:
Thank you for this fix 👍

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1457)* status: 0 of 1 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1457#-N_YGhNe7TXmmE2U-pfs:-N_YGhNe7TXmmE2U-pft:b-e1kqj8) (no related file):*
I couldn't come up with a nicer approach, because `__FILE_NAME__` **must be** a macro, because it defines the source file. In other words, embedding this `truncate_file_name()` into e.g. `_log()` function would result in Gramine always showing `log.c` as the filename...

If anyone has ideas how to keep the macro AND not calculate it in every place, then we could modify the code.

Also note that the "log level" variables are different in PAL, LibOS, common code, regression tests, tools, so I can't embed the check on the log level inside `truncate_file_name()`.

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1457#-N_YHDII7fucwTgHxFtP:-N_YHDIJEnRlXf1rm8oi:bguyift) (no related file):*
Unfortunately, our Gramine packages v1.5 for Debian/Ubuntu were built on Debian 11, which has **GCC 10**.

Similarly, RHEL-like packages were built on AlmaLinux 8 (which has GCC 9) and AlmaLinux 9 (which has GCC 11).

So all our current packages have this performance regression...

This is according to @woju, please correct me if I'm wrong.

___

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1457)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`common/include/log.h` line 35 at r1](https://reviewable.io/reviews/gramineproject/gramine/1457#-N_Y_hIs0xl5agv23s0W:-N_Y_hIs0xl5agv23s0X:b4u9bz) ([raw file](https://github.com/gramineproject/gramine/blob/db2a92e4e37acd7e5023cd728fb78df00527d14d/common/include/log.h#L35)):*
> ```objc
> }
> 
> #define __FILE_NAME__ (truncate_file_name(__FILE__))
> ```

I don't like this fix, it makes the logs hard to read. Can't we just recompile our packages on a newer GCC?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1457)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @woju)

___
*[`common/include/log.h` line 35 at r1](https://reviewable.io/reviews/gramineproject/gramine/1457#-N_Y_hIs0xl5agv23s0W:-N_Ycup0BNYsdqQQuIIX:b-rczzwa) ([raw file](https://github.com/gramineproject/gramine/blob/db2a92e4e37acd7e5023cd728fb78df00527d14d/common/include/log.h#L35)):*

I don't like this fix, it makes the logs hard to read. Can't we just recompile our packages on a newer GCC?
</blockquote></details>

1. @woju needs to weigh on this possibility
2. We will need to explain this somehow to users? Like, have a loud warning? How to do that?

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1457)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1457#-N_YHDII7fucwTgHxFtP:-N_YlymI8ZsEbJ_AFiDJ:bd1mj8t) (no related file):*

Unfortunately, our Gramine packages v1.5 for Debian/Ubuntu were built on Debian 11, which has **GCC 10**.

Similarly, RHEL-like packages were built on AlmaLinux 8 (which has GCC 9) and AlmaLinux 9 (which has GCC 11).

So all our current packages have this performance regression...

This is according to @woju, please correct me if I'm wrong.
</blockquote></details>

Yes, this is all correct.

___
___
*[`common/include/log.h` line 35 at r1](https://reviewable.io/reviews/gramineproject/gramine/1457#-N_Y_hIs0xl5agv23s0W:-N_YnPv943X2I5gBArGq:b-x5aj4s) ([raw file](https://github.com/gramineproject/gramine/blob/db2a92e4e37acd7e5023cd728fb78df00527d14d/common/include/log.h#L35)):*

1. @woju needs to weigh on this possibility
2. We will need to explain this somehow to users? Like, have a loud warning? How to do that?
</blockquote></details>

At least EL8 packages we can't compile with GCC12 at all. On EL9 we might have fighting chance (GCC 12 is in EPEL9), but I'm not sure, I would have to check this. With Debian/Ubuntu, the situation is yet less clear, we would have to compile on newer distro (with newer glibc and libprotobuf) and try to run this on older distro. This is not generally viable because of possibly missing symbols in older libs. Might work, might not work. We don't test this. The solution that would work would be to release different package for Debian 12, but I'm not sure if such package would work even on Ubuntu 22.04.

I understand you're annoyed that you can't use a nice feature of a newest compiler, but that's the constraints we're working with.

If you want to have predictable prefix, maybe you can `#define` the log name in each file you care about, `pr_fmt`-style like Linux does?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1457)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

___
*[`common/include/log.h` line 35 at r1](https://reviewable.io/reviews/gramineproject/gramine/1457#-N_Y_hIs0xl5agv23s0W:-N_d60vo9F-PbX5EYEVb:b-3igzo4) ([raw file](https://github.com/gramineproject/gramine/blob/db2a92e4e37acd7e5023cd728fb78df00527d14d/common/include/log.h#L35)):*
@mkow Given Woju's comments, I think it's unreasonable to keep the perf overhead... Even Ubuntu 22.04 has GCC11, which still suffers from this problem.

> If you want to have predictable prefix, maybe you can `#define` the log name in each file you care about, `pr_fmt`\-style like Linux does?

@mkow This is not a bad idea. What do you think? If you're fine with this proposal, I can modify in this PR.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1457)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1457)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1457#-N_e0SKiESTuhhsPGxeC:-N_e0SKiESTuhhsPGxeD:bc2d3m4) (no related file):*
TODO(dimakuv): I will rework this into an always-inlined cascade-of-IFs function, such that the compiler in `-O1` mode should be able to extract the basename at compile time.

If this won't work out, then I'll fall back to another approach: change `_log()` function implementations into `_log()` macro implementations. Unfortunately, LibOS, PAL, regression and tools have 4 different implementations, so it will be a cascade of `#ifdef` snippets.

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1457)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1457#-N_e0SKiESTuhhsPGxeC:-N_e2Nzk8grsCPxooJYu:b-wogyt5) (no related file):*

TODO(dimakuv): I will rework this into an always-inlined cascade-of-IFs function, such that the compiler in `-O1` mode should be able to extract the basename at compile time.

If this won't work out, then I'll fall back to another approach: change `_log()` function implementations into `_log()` macro implementations. Unfortunately, LibOS, PAL, regression and tools have 4 different implementations, so it will be a cascade of `#ifdef` snippets.
</blockquote></details>

I think I can use this: https://stackoverflow.com/a/41367919

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1457)* status: 0 of 1 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @kailun-qin and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1457#-N_e0SKiESTuhhsPGxeC:-N_h4jrJ2RbLbQmXdIkQ:beah7p8) (no related file):*

I think I can use this: https://stackoverflow.com/a/41367919
</blockquote></details>

To use the cascade of IFs, I needed to learn the current max size of the basename in Gramine codebase. I did something like this in all relevant dirs of Gramine:
```
gramineproject/gramine/libos/include$ find . -type f | awk -F/ 'BEGIN {maxlength = 0; longest = "" } length( $NF ) > maxlength { maxlength = length( $NF ); longest = $NF } END { print "longest was", longest, "at", maxlength, "characters." }'
```

The results:
```
longest was libos_pollable_event.h at 22 characters.
longest was libos_ipc_process_info.c at 24 characters.
longest was protected_files_internal.h at 26 characters.
longest was etc_host_info_internal.h at 24 characters.
longest was common_manifest_sgx_parser.c at 28 characters.
```

So I'll currently overapproximate to 35 chars.

___
___
*[`common/include/log.h` line 35 at r1](https://reviewable.io/reviews/gramineproject/gramine/1457#-N_Y_hIs0xl5agv23s0W:-N_h85Qx2JS0w3dPSP0K:b-wh9bd7) ([raw file](https://github.com/gramineproject/gramine/blob/db2a92e4e37acd7e5023cd728fb78df00527d14d/common/include/log.h#L35)):*

@mkow Given Woju's comments, I think it's unreasonable to keep the perf overhead... Even Ubuntu 22.04 has GCC11, which still suffers from this problem.

> If you want to have predictable prefix, maybe you can `#define` the log name in each file you care about, `pr_fmt`\-style like Linux does?

@mkow This is not a bad idea. What do you think? If you're fine with this proposal, I can modify in this PR.
</blockquote></details>

I went with another idea. Works pretty nice.
___
*[`common/include/log.h` line 18 at r2](https://reviewable.io/reviews/gramineproject/gramine/1457#-N_h7y4fD66ct6lRnMlu:-N_h7y4fD66ct6lRnMlv:bnjckvh) ([raw file](https://github.com/gramineproject/gramine/blob/e67e8e8130a6fafc4a02899d3fdeaa7100f99ed3/common/include/log.h#L18)):*
> ```objc
> 
> #ifndef __FILE_NAME__
> #ifndef __OPTIMIZE__
> ```

I verified that `__OPTIMIZE__` is issued by both GCC and Clang on any non-O0 optimization level.
___
*[`common/include/log.h` line 67 at r2](https://reviewable.io/reviews/gramineproject/gramine/1457#-N_h898p25CJi8GCKXyb:-N_h898p25CJi8GCKXyc:b81lmfj) ([raw file](https://github.com/gramineproject/gramine/blob/e67e8e8130a6fafc4a02899d3fdeaa7100f99ed3/common/include/log.h#L67)):*
> ```objc
>      sizeof(s) > 33 && (s)[sizeof(s)-33] == '/' ? (s) + sizeof(s) - 32 : \
>      sizeof(s) > 34 && (s)[sizeof(s)-34] == '/' ? (s) + sizeof(s) - 33 : \
>      sizeof(s) > 35 && (s)[sizeof(s)-35] == '/' ? (s) + sizeof(s) - 34 : (s))
> ```

I could have an additional helper macro to simplify each line, but on the other hand... who cares?

woju (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1457)* status: 0 of 1 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv, @kailun-qin, and @mkow)

___
*[`common/include/log.h` line 18 at r2](https://reviewable.io/reviews/gramineproject/gramine/1457#-N_h7y4fD66ct6lRnMlu:-N_ia8A49Ra72805yAoZ:b-p9v2hk) ([raw file](https://github.com/gramineproject/gramine/blob/e67e8e8130a6fafc4a02899d3fdeaa7100f99ed3/common/include/log.h#L18)):*

I verified that `__OPTIMIZE__` is issued by both GCC and Clang on any non-O0 optimization level.
</blockquote></details>

Why do we need two implementations? If this cascade version is OK, why can't we use it also when not `__OPTIMIZE__`?
___
*[`common/include/log.h` line 67 at r2](https://reviewable.io/reviews/gramineproject/gramine/1457#-N_h898p25CJi8GCKXyb:-N_iaTqr11OgSS4ZV7kt:b29pe5a) ([raw file](https://github.com/gramineproject/gramine/blob/e67e8e8130a6fafc4a02899d3fdeaa7100f99ed3/common/include/log.h#L67)):*

I could have an additional helper macro to simplify each line, but on the other hand... who cares?
</blockquote></details>

Or you could generate a header with this device using a parametrised script...

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1457)* status: 0 of 1 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv, @kailun-qin, and @mkow)

___
*[`common/include/log.h` line 18 at r2](https://reviewable.io/reviews/gramineproject/gramine/1457#-N_h7y4fD66ct6lRnMlu:-N_ifwadEyxQhESj7Bub:bgadc6i) ([raw file](https://github.com/gramineproject/gramine/blob/e67e8e8130a6fafc4a02899d3fdeaa7100f99ed3/common/include/log.h#L18)):*

Because the cascade version is NOT optimized away, and debugging with GDB through 35 IFs every time there is a `log_debug()` func in the code is annoying.
</blockquote></details>

Ah, OK. Yes, that's fair.
___
*[`common/include/log.h` line 67 at r2](https://reviewable.io/reviews/gramineproject/gramine/1457#-N_h898p25CJi8GCKXyb:-N_ig-te38UztjxxkLPq:br4ssp7) ([raw file](https://github.com/gramineproject/gramine/blob/e67e8e8130a6fafc4a02899d3fdeaa7100f99ed3/common/include/log.h#L67)):*

Which device? Not sure what you mean.
</blockquote></details>

By "device" I mean this construct of `sizeof (s) ... ? ... : ` sequence. (Cf. https://en.wikipedia.org/wiki/Duff%27s_device)

Looks repetitive to me, and how do I check if you didn't make an error somewhere within those numbers. It can be generated by a script ran from meson's `generator()` or `custom_target()`.

woju (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1457)* status: 0 of 1 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @kailun-qin, and @mkow)

___
*[`common/include/log.h` line 67 at r2](https://reviewable.io/reviews/gramineproject/gramine/1457#-N_h898p25CJi8GCKXyb:-N_nVYNG2vH5rTHeBie_:byzmjvq) ([raw file](https://github.com/gramineproject/gramine/blob/e67e8e8130a6fafc4a02899d3fdeaa7100f99ed3/common/include/log.h#L67)):*

Ok, I added a helper macro, this should look more readable now.

If I understood correctly Woju's proposal, the script approach would require: adding the meson file into this dir, writing a simple Bash/Python script to generate the repeated macro sequence, add a `custom_target()` to meson, and specify the auto-generated header file as an include to this file. This whole change sounds like an overkill for this particular small problem.
</blockquote></details>

Yes. LGTM.

kailun-qin (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1457)* status: all files reviewed, 5 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`common/include/log.h` line 31 at r3](https://reviewable.io/reviews/gramineproject/gramine/1457#-N_pfveu2fop44hFNNkT:-N_pfveu2fop44hFNNkU:bzqwpe) ([raw file](https://github.com/gramineproject/gramine/blob/8b8da94742e1e4779f25152eddad65b0c7881362/common/include/log.h#L31)):*
> ```objc
> }
> #define __FILE_NAME__ (truncate_file_name(__FILE__))
> #else  // ifndef __OPTIMIZE__
> ```

-> `#else // ifndef __OPTIMIZE__` (single space before the comment)?



_Code quote:_
```objc
#else  // ifndef __OPTIMIZE__
```

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1457)* status: all files reviewed, 5 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`common/include/log.h` line 35 at r1](https://reviewable.io/reviews/gramineproject/gramine/1457#-N_Y_hIs0xl5agv23s0W:-Na5n3QR2ElTBeXrvhkJ:b3b23bw) ([raw file](https://github.com/gramineproject/gramine/blob/db2a92e4e37acd7e5023cd728fb78df00527d14d/common/include/log.h#L35)):*
> > If you want to have predictable prefix, maybe you can `#define` the log name in each file you care about, `pr_fmt`\-style like Linux does?
> 
> [@mkow](https://github.com/mkow) This is not a bad idea. What do you think? If you're fine with this proposal, I can modify in this PR.

This is definitely a bad idea, I don't want to modify and trash all our source files because of some issue with old compilers. Also, this would be very hard to keep correct when renaming files.
The current workaround is constrained to the place where it occurs, which is nice. What @woju proposed would basically spread the workaround through the whole codebase.
___
*[`common/include/log.h` line 67 at r2](https://reviewable.io/reviews/gramineproject/gramine/1457#-N_h898p25CJi8GCKXyb:-Na5oj4eEnY0a-nxsccI:b235kh4) ([raw file](https://github.com/gramineproject/gramine/blob/e67e8e8130a6fafc4a02899d3fdeaa7100f99ed3/common/include/log.h#L67)):*
> It can be generated by a script ran from meson's `generator()` or `custom_target()`.

> This whole change sounds like an overkill for this particular small problem.

Yes, this is a huge overkill for just a few lines of code which run only when old compilers are used. The current version is ok.
___
*[`common/include/log.h` line 34 at r3](https://reviewable.io/reviews/gramineproject/gramine/1457#-Na5oM1W-samroR893R6:-Na5oM1W-samroR893R7:b-k8my5y) ([raw file](https://github.com/gramineproject/gramine/blob/8b8da94742e1e4779f25152eddad65b0c7881362/common/include/log.h#L34)):*
> ```objc
> /* Below macros will be optimized at compile-time, incurring no perf overhead at runtime. */
> #define TRY_TRUNCATE_FILE_NAME_AT_LOC(s, loc) \
>     (sizeof(s) > (loc) && (s)[sizeof(s)-(loc)] == '/') ? ((s) + sizeof(s) - (loc) + 1)
> ```

What's the exact intention of this check? Isn't there an off-by-one?



_Code quote:_
```objc
sizeof(s) > (loc)
```
___
*[`common/include/log.h` line 34 at r3](https://reviewable.io/reviews/gramineproject/gramine/1457#-Na5pAwz9sU7gAiPxlid:-Na5pAwz9sU7gAiPxlie:birwe76) ([raw file](https://github.com/gramineproject/gramine/blob/8b8da94742e1e4779f25152eddad65b0c7881362/common/include/log.h#L34)):*
> ```objc
> /* Below macros will be optimized at compile-time, incurring no perf overhead at runtime. */
> #define TRY_TRUNCATE_FILE_NAME_AT_LOC(s, loc) \
>     (sizeof(s) > (loc) && (s)[sizeof(s)-(loc)] == '/') ? ((s) + sizeof(s) - (loc) + 1)
> ```

(to assert that `s` is a C-string literal, not a pointer to a string)



_Suggestion:_
```objc
("" s, sizeof(s) > (loc) && (s)[sizeof(s)-(loc)] == '/') ? ((s) + sizeof(s) - (loc) + 1)
```

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1457)* status: 0 of 1 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin and @mkow)

___
*[`common/include/log.h` line 35 at r1](https://reviewable.io/reviews/gramineproject/gramine/1457#-N_Y_hIs0xl5agv23s0W:-Na6UAuWCKYw_BNltb7T:bp5462d) ([raw file](https://github.com/gramineproject/gramine/blob/db2a92e4e37acd7e5023cd728fb78df00527d14d/common/include/log.h#L35)):*

> > If you want to have predictable prefix, maybe you can `#define` the log name in each file you care about, `pr_fmt`\-style like Linux does?
> 
> [@mkow](https://github.com/mkow) This is not a bad idea. What do you think? If you're fine with this proposal, I can modify in this PR.

This is definitely a bad idea, I don't want to modify and trash all our source files because of some issue with old compilers. Also, this would be very hard to keep correct when renaming files.
The current workaround is constrained to the place where it occurs, which is nice. What @woju proposed would basically spread the workaround through the whole codebase.
</blockquote></details>

Yes, agreed with Michal's counterargument.
___
*[`common/include/log.h` line 31 at r3](https://reviewable.io/reviews/gramineproject/gramine/1457#-N_pfveu2fop44hFNNkT:-Na6a3Qt7KNdyqj9NPQY:bb6un1j) ([raw file](https://github.com/gramineproject/gramine/blob/8b8da94742e1e4779f25152eddad65b0c7881362/common/include/log.h#L31)):*

-> `#else // ifndef __OPTIMIZE__` (single space before the comment)?
</blockquote></details>

Done
___
*[`common/include/log.h` line 34 at r3](https://reviewable.io/reviews/gramineproject/gramine/1457#-Na5oM1W-samroR893R6:-Na6VXcl0NRf1x3Jkvi3:bqjdk58) ([raw file](https://github.com/gramineproject/gramine/blob/8b8da94742e1e4779f25152eddad65b0c7881362/common/include/log.h#L34)):*

What's the exact intention of this check? Isn't there an off-by-one?
</blockquote></details>

Done.  This was intentional to fix the problem described below, but I just realized that this off-by-one incorrectly parses cases like `"/a\0"` (i.e. a string `/a` plus a NUL symbol). So I removed the off-by-one and just assume that `"/\0"` case should never happen in reality.

---

I originally tried to solve this problem with this off-by-one:

Consider the example `s = "/\0"` (a single `/` symbol followed by NUL). With current code, the only case that passes this first check is `TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 1)`. In this case, the second check fails (because `s[2-1] == '\0'` and not `/`). So we don't go into the true branch. In the end, `TRUNCATE_FILE_NAME(s)` ends up with `"/\0"` (the whole string), which is the reasonable output.

Now consider this example with the first check being replaced with `sizeof(s) >= (loc)` (that's how I understand your comment). There will be two cases that pass the first check:
1. `TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 1)` -- same as previously, it won't pass the second check.
2. `TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 2)` -- it passes the second check (because `s[2-2] == '/'`). So we end up in the true branch which gives us `s + 2 - 2 + 1`, or `s[1]` i.e. the NUL symbol -- not a reasonable output.

But now I see that it breaks other cases. Let's take another example and see how the current macro works with it: `s = "/a\0"` (i.e. string `/a`):
1. `TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 1)` -- passes first check, fails second check (because `s[3-1] == '\0'`).
2. `TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 2)` -- passes first check, fails second check (because `s[3-2] == 'a'`.
3. `TRY_TRUNCATE_FILE_NAME_AT_LOC(s, 3)` -- fails first check.
4. ... all the rest fail the first check.
5. End up with printing just `s`, i.e. the full name, not the basename.
___
*[`common/include/log.h` line 34 at r3](https://reviewable.io/reviews/gramineproject/gramine/1457#-Na5pAwz9sU7gAiPxlid:-Na6aanDA9e_zEPVIzL5:b-vl5tom) ([raw file](https://github.com/gramineproject/gramine/blob/8b8da94742e1e4779f25152eddad65b0c7881362/common/include/log.h#L34)):*

(to assert that `s` is a C-string literal, not a pointer to a string)
</blockquote></details>

I tried the exact line that you proposed (with a comma) but:
```
../common/include/log.h:34:10: warning: left-hand operand of comma expression has no effect [-Wunused-value]
   34 |     ("" s, sizeof(s) >= (loc) && (s)[sizeof(s)-(loc)] == '/') ? ((s) + sizeof(s) - (loc) + 1)
      |          ^
../common/include/log.h:51:45: note: in expansion of macro ‘TRY_TRUNCATE_FILE_NAME_AT_LOC’
```

I tried a couple other things, but they have false positives... Also, it looks non-trivial to determine whether something is a C-string literal: https://stackoverflow.com/questions/5691232/can-i-determine-if-an-argument-is-string-literal

So I don't know how to address your comment.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1457)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners

___
*[`common/include/log.h` line 34 at r3](https://reviewable.io/reviews/gramineproject/gramine/1457#-Na5pAwz9sU7gAiPxlid:-Na8G4nZCWEAag2VVR15:b-tvd4ax) ([raw file](https://github.com/gramineproject/gramine/blob/8b8da94742e1e4779f25152eddad65b0c7881362/common/include/log.h#L34)):*

I tried the exact line that you proposed (with a comma) but:
```
../common/include/log.h:34:10: warning: left-hand operand of comma expression has no effect [-Wunused-value]
   34 |     ("" s, sizeof(s) >= (loc) && (s)[sizeof(s)-(loc)] == '/') ? ((s) + sizeof(s) - (loc) + 1)
      |          ^
../common/include/log.h:51:45: note: in expansion of macro ‘TRY_TRUNCATE_FILE_NAME_AT_LOC’
```

I tried a couple other things, but they have false positives... Also, it looks non-trivial to determine whether something is a C-string literal: https://stackoverflow.com/questions/5691232/can-i-determine-if-an-argument-is-string-literal

So I don't know how to address your comment.
</blockquote></details>

Interesting. Ok, let's skip it then, I thought it will be simple to add.

kailun-qin (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1457)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners

Issue #1459, opened 2023-07-18
Gramine Build is failing with latest meson version 1.2.0


anjalirai-intel:
### Description of the problem

As part of Gramine Readme, any meson version above 0.56 is required to install for Gramine Build, But with latest meson version it is failing with **_"/usr/bin/ld: ias.c:(.text+0xdae): undefined reference to `curl_easy_strerror'"_**

**Gramine Readme**
sudo python3 -m pip install 'meson>=0.56'

As part of our ci dockerfile, we are restricting meson version to less than 0.57, This should also be in sync with readme and updated to meson>=0.56
```
RUN python3 -m pip install -U \
    'tomli>=1.1.0' \
    'tomli-w>=0.4.0' \
    'meson>=0.56,<0.57' \
```
https://github.com/gramineproject/gramine/blob/master/.ci/ubuntu20.04.dockerfile#L103

**Solution:**
Meson version-1.1.1 is compatible and gramine build goes through

### Steps to reproduce

meson setup build/ --buildtype=release -Ddirect=enabled -Dsgx=enabled
ninja -C build/

### Expected results

ninja build should go through

### Actual results

```
ninja build
.....
....
[381/389] Linking target tools/sgx/ra-tls/libra_tls_verify_epid.so
FAILED: tools/sgx/ra-tls/libra_tls_verify_epid.so 
[...]

PR #1461, opened 2023-07-18
[curl] Make curl dependency compatible with Meson 1.2.0

dimakuv:
## Description of the changes 

It is unclear why, but starting from v1.2.0 Meson refuses to see the static `libcurl.a` library unless it is more explicitly specified.

WIP.

## How to test this PR? 

Test with e.g. GSC on Ubuntu 20.04.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1461)


dimakuv:
See my analysis: https://github.com/gramineproject/gramine/issues/1459. This PR doesn't help at all.

PR #1473, opened 2023-07-25
[PAL] Fail if Gramine binaries are built with RELR relocs

### LINKED ISSUE #1446: LibOS and trusted-PAL binaries may be built with `RELR` relocations

Recently, a new type of relocations called `RELR` was added in many linkers. See details here:
- https://maskray.me/blog/2021-10-31-relative-relocations-and-relr
- https://github.com/gramineproject/gramine/pull/1393#pullrequestreview-1481456937

AFAIK, to date (July 2023) **no** build system and no compiler/linker emit `RELR` relocations by default.

However, the Alpine's `abuild` tool (used to build Alpine packages) extends LDFLAGS to force the compiler/linker to emit `RELR` relocations. E.g., this is the default abuild config file on Alpine 3.18.0:
```
alpine:~$ cat /usr/share/abuild/default.conf | fgrep LDFLAGS
export LDFLAGS="$LDFLAGS -Wl,-z,pack-relative-relocs"
```

("Relative relocs" = `RELR`)

At the same time, Gramine consists of two "special" self-relocating binaries: the LibOS binary and the trusted-PAL binary (only for `gramine-sgx`). The relocation code is written by ourselves and the relevant code snippets are here:
- https://github.com/gramineproject/gramine/blob/ab2d3217df4220578c607f8117b21d46c849859f/pal/src/pal_rtld.c#L561
- https://github.com/gramineproject/gramine/blob/ab2d3217df4220578c607f8117b21d46c849859f/pal/src/pal_rtld.c#L205
- https://github.com/gramineproject/gramine/blob/ab2d3217df4220578c607f8117b21d46c849859f/pal/src/pal_rtld.c#L219-L231

As you can see, Gramine doesn't know about `RELR` relocations, and what's worse Gramine **simply ignores** unknown relocation types. Thus, some symbols and addresses in Gramine's LibOS and trusted-PAL binaries are not relocated at all and thus contain bogus address. This leads to segfaults like this:
```
alpine:~$ gramine-sgx helloworld
error: Unexpected memory fault occurred inside PAL (0x13b05)
```

From the above error message it's clear that the problem is in trusted PAL (otherwise the message would say `... inside untrusted PAL`). And the address `0x13b05` is too low for a real meaningful address, so it's clearly a not-relocated offset value.

---

**Possible fixes**:
1. Currently we **unset `LDFLAGS`** in our APKBUILD script. This is a valid solution because we have our own loader, we're in position of a distro, and it's perfectly fine for us to dictate LDFLAGS.
2. The first solution works *now* but it may break in the future -- at some point emitting `RELR` relocations may be enabled by default by all compilers/linkers, and then Gramine's ignorance about `RELR` will lead to segfaults. A proper fix would be to **add parsing of `RELR` relocations to Gramine's RTLD code**.

Until we implement fix 2, we should at least loudly fail on `RELR` relocation types, with a hint of what went wrong.

dimakuv:
## Description of the changes 

RELR (relative relocs) is a new type of relocations recently added to compilers/linkers, e.g. glibc 2.36 and musl 1.2.4. These relocs are not enabled by default; however, if the system enables them during Gramine build, Gramine would segfault. This is because Gramine has its own self-relocation code that relocates the SGX PAL and the LibOS binaries. Before this commit, the code ignored RELR relocs, which led to some symbols being not relocated and ultimately to segfaults.

This commit makes Gramine aware of RELR relocs, though we simply loudly fail when we detect such type. Currently, Gramine build environment must disable such relocs.

Fixes #1446.

For the context, see #1446.

Good references:
- https://maskray.me/blog/2021-10-31-relative-relocations-and-relr
- https://git.musl-libc.org/cgit/musl/commit/?id=d32dadd60efb9d3b255351a3b532f8e4c3dd0db1
- https://reviews.llvm.org/D120701

The updates to `elf.h` were taken from this Glibc commit: https://github.com/bminor/glibc/commit/4610b24f5e4e6d2c4b769594efa6d460943163bb

## How to test this PR? 

Manually try building with `LDFLAGS="$LDFLAGS -Wl,-z,pack-relative-relocs"` on an OS with Glibc 2.36 (e.g. Ubuntu 23.04). I'll post my experiments soon.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1473)

### DIFF
--- a/pal/include/elf/elf.h
+++ b/pal/include/elf/elf.h
@@ -321,7 +321,8 @@ typedef struct {
 #define SHT_PREINIT_ARRAY  16         /* Array of pre-constructors */
 #define SHT_GROUP          17         /* Section group */
 #define SHT_SYMTAB_SHNDX   18         /* Extended section indeces */
-#define SHT_NUM            19         /* Number of defined types.  */
+#define SHT_RELR           19         /* RELR relative relocations */
+#define SHT_NUM            20         /* Number of defined types.  */
 #define SHT_LOOS           0x60000000 /* Start OS-specific.  */
 #define SHT_GNU_ATTRIBUTES 0x6ffffff5 /* Object attributes.  */
 #define SHT_GNU_HASH       0x6ffffff6 /* GNU-style hash table.  */
@@ -498,6 +499,11 @@ typedef struct {
     Elf64_Sxword r_addend; /* Addend */
 } Elf64_Rela;
 
+/* RELR relocation table entry */
+
+typedef Elf32_Word  Elf32_Relr;
+typedef Elf64_Xword Elf64_Relr;
+
 /* How to extract and insert information held in the r_info field.  */
 
 #define ELF32_R_SYM(val)        ((val) >> 8)
@@ -653,7 +659,11 @@ typedef struct {
 #define DT_ENCODING        32          /* Start of encoded range */
 #define DT_PREINIT_ARRAY   32          /* Array with addresses of preinit fct*/
 #define DT_PREINIT_ARRAYSZ 33          /* size in bytes of DT_PREINIT_ARRAY */
-#define DT_NUM             34          /* Number used */
+#define DT_SYMTAB_SHNDX    34          /* Address of SYMTAB_SHNDX section */
+#define DT_RELRSZ          35          /* Total size of RELR relative relocations */
+#define DT_RELR            36          /* Address of RELR relative relocations */
+#define DT_RELRENT         37          /* Size of one RELR relative relocaction */
+#define DT_NUM             38          /* Number used */
 #define DT_LOOS            0x6000000d  /* Start of OS-specific */
 #define DT_HIOS            0x6ffff000  /* End of OS-specific */
 #define DT_LOPROC          0x70000000  /* Start of processor-specific */
@@ -2655,7 +2665,9 @@ typedef ElfW(Ehdr)   elf_ehdr_t;
 typedef ElfW(Half)   elf_half_t;
 typedef ElfW(Off)    elf_off_t;
 typedef ElfW(Phdr)   elf_phdr_t;
+typedef ElfW(Rel)    elf_rel_t;
 typedef ElfW(Rela)   elf_rela_t;
+typedef ElfW(Relr)   elf_relr_t;
 typedef ElfW(Sym)    elf_sym_t;
 typedef ElfW(Word)   elf_word_t;
 typedef ElfW(Xword)  elf_xword_t;
--- a/pal/src/pal_rtld.c
+++ b/pal/src/pal_rtld.c
@@ -42,6 +42,7 @@
  * doesn't require relocation. */
 extern const elf_ehdr_t __ehdr_start __attribute__((visibility("hidden")));
 
+static const char* g_pal_soname = NULL;
 static struct link_map g_pal_map;
 static struct link_map g_entrypoint_map;
 
@@ -202,6 +203,130 @@ static int find_symbol_in_loaded_maps(struct link_map* map, elf_rela_t* rela,
     return -PAL_ERROR_DENIED;
 }
 
+static int verify_dynamic_entries(struct link_map* map) {
+    elf_addr_t base_diff = map->l_base_diff;
+    elf_dyn_t* dynamic_section_entry = map->l_ld;
+
+    const char* string_table = NULL;
+    elf_xword_t string_table_size = 0;
+
+    /* DT_NEEDED entries contain offsets into the string table (DT_STRTAB) */
+    elf_xword_t needed_offset = 0;
+    bool needed_offset_found  = false;
+
+    while (dynamic_section_entry->d_tag != DT_NULL) {
+        switch (dynamic_section_entry->d_tag) {
+            case DT_RELACOUNT:
+            case DT_RELA:
+            case DT_RELASZ:
+            case DT_JMPREL:
+            case DT_PLTRELSZ:
+                /* recognized relocation types, used in perform_relocations() */
+                break;
+            case DT_RELENT:
+                if (dynamic_section_entry->d_un.d_val != sizeof(elf_rel_t)) {
+                    log_error("Unexpected DT_RELENT (size of one Rel reloc)");
+                    return -PAL_ERROR_DENIED;
+                }
+                break;
+            case DT_RELAENT:
+                if (dynamic_section_entry->d_un.d_val != sizeof(elf_rela_t)) {
+                    log_error("Unexpected DT_RELAENT (size of one Rela reloc)");
+                    return -PAL_ERROR_DENIED;
+                }
+                break;
+            case DT_RELRENT:
+                if (dynamic_section_entry->d_un.d_val != sizeof(elf_relr_t)) {
+                    log_error("Unexpected DT_RELRENT (size of one Relr reloc)");
+                    return -PAL_ERROR_DENIED;
+                }
+                break;
+            case DT_SYMENT:
+                if (dynamic_section_entry->d_un.d_val != sizeof(elf_sym_t)) {
+                    log_error("Unexpected DT_SYMENT (size of one symbol table entry)");
+                    return -PAL_ERROR_DENIED;
+                }
+                break;
+            case DT_PLTREL:
+                if (dynamic_section_entry->d_un.d_val != DT_RELA) {
+                    log_error("Unexpected DT_PLTREL (type of relocs in PLT must be Rela)");
+                    return -PAL_ERROR_DENIED;
+                }
+                break;
+            case DT_NEEDED:
+                if (needed_offset_found) {
+                    log_error("Loaded binary has more than one dependency (must be only one "
+                              "dependency -- the PAL binary)");
+                    return -PAL_ERROR_DENIED;
+                }
+                needed_offset = dynamic_section_entry->d_un.d_val;
+                needed_offset_found = true;
+                break;
+            case DT_SONAME:
+                /* unused, semantically a no-op (for PAL binary, extracted in setup_pal_binary()) */
+                break;
+            case DT_DEBUG:
+                /* unused, semantically a no-op */
+                break;
+            case DT_PLTGOT:
+                /* address of PLT -- unused because Rela relocs include this addr in the offset */
+            case DT_FLAGS:
+            case DT_FLAGS_1:
+                /* additional linker flags -- unclear how to verify them so currently ignore */
+            case DT_VERDEF:
+            case DT_VERDEFNUM:
+            case DT_VERNEED:
+            case DT_VERNEEDNUM:
+            case DT_VERSYM:
+                /* versioning entries -- unclear how to verify them so currently ignore */
+                break;
+            case DT_INIT_ARRAY:
+            case DT_FINI_ARRAY:
+            case DT_INIT_ARRAYSZ:
+            case DT_FINI_ARRAYSZ:
+                /* init/fini routines -- used by AddressSanitizer; unclear how to verify them */
+                break;
+            case DT_STRTAB:
+                string_table = (const char*)(dynamic_section_entry->d_un.d_ptr + base_diff);
+                break;
+            case DT_STRSZ:
+                string_table_size = dynamic_section_entry->d_un.d_val;
+                break;
+            case DT_HASH:
+            case DT_SYMTAB:
+                /* used in find_string_and_symbol_tables() */
+                break;
+            case DT_RELR:
+            case DT_RELRSZ:
+                /* be explicit about unsupported RELR relocation type */
+                log_error("Unsupported relocation type DT_RELR; you may need to rebuild Gramine "
+                          "with `-Wl,-z,nopack-relative-relocs` linker option");
+                return -PAL_ERROR_DENIED;
+            default:
+                log_error("Unrecognized dynamic entry (DT_*) %ld", dynamic_section_entry->d_tag);
+                return -PAL_ERROR_DENIED;
+
+            }
+        dynamic_section_entry++;
+    }
+
+    if (!string_table || !string_table_size) {
+        log_error("Loaded binary doesn't have string table or it is empty (DT_STRTAB/DT_STRSZ)");
+        return -PAL_ERROR_DENIED;
+    }
+
+    /* verify DT_NEEDED: LibOS and PAL tests have PAL lib; PAL doesn't have this entry */
+    if (needed_offset_found) {
+        const char* needed = string_table + needed_offset;
+        if (strcmp(needed, g_pal_soname) != 0) {
+            log_error("Unexpected DT_NEEDED (must be name of the PAL library)");
+            return -PAL_ERROR_DENIED;
+        }
+    }
+
+    return 0;
+}
+
 static int perform_relocations(struct link_map* map) {
     int ret;
 
@@ -214,8 +339,14 @@ static int perform_relocations(struct link_map* map) {
     elf_rela_t* plt_relas_addr = NULL;
     elf_xword_t plt_relas_size = 0;
 
+    elf_xword_t relas_count = 0;
+    elf_xword_t expected_relas_count = 0;
+
     while (dynamic_section_entry->d_tag != DT_NULL) {
         switch (dynamic_section_entry->d_tag) {
+            case DT_RELACOUNT:
+                expected_relas_count = dynamic_section_entry->d_un.d_val;
+                break;
             case DT_RELA:
                 relas_addr = (elf_rela_t*)(dynamic_section_entry->d_un.d_ptr + base_diff);
                 break;
@@ -243,6 +374,7 @@ static int perform_relocations(struct link_map* map) {
         if (ELF_R_TYPE(rela->r_info) == R_X86_64_RELATIVE) {
             elf_addr_t* addr_to_relocate = (elf_addr_t*)(rela->r_offset + base_diff);
             *addr_to_relocate = *addr_to_relocate + base_diff;
+            relas_count++;
         } else if (ELF_R_TYPE(rela->r_info) == R_X86_64_GLOB_DAT) {
             elf_addr_t symbol_addr;
             ret = find_symbol_in_loaded_maps(map, rela, &symbol_addr);
@@ -256,7 +388,6 @@ static int perform_relocations(struct link_map* map) {
                       "R_X86_64_RELATIVE and R_X86_64_GLOB_DAT relocations");
             return -PAL_ERROR_DENIED;
         }
-
     }
 
     if (!plt_relas_addr && plt_relas_size) {
@@ -280,6 +411,11 @@ static int perform_relocations(struct link_map* map) {
 
         elf_addr_t* addr_to_relocate = (elf_addr_t*)(plt_rela->r_offset + base_diff);
         *addr_to_relocate = symbol_addr + plt_rela->r_addend;
+    }
+
+    if (relas_count != expected_relas_count) {
+        log_error("Expected %ld Rela relocs but got %ld", expected_relas_count, relas_count);
+        return -PAL_ERROR_DENIED;
     }
 
     return 0;
@@ -429,6 +565,10 @@ static int create_and_relocate_entrypoint(PAL_HANDLE handle, const char* uri,
     g_entrypoint_map.l_entry = g_entrypoint_map.l_entry + g_entrypoint_map.l_base_diff;
     g_entrypoint_map.l_ld = (elf_dyn_t*)((elf_addr_t)g_entrypoint_map.l_ld +
                                          g_entrypoint_map.l_base_diff);
+
+    ret = verify_dynamic_entries(&g_entrypoint_map);
+    if (ret < 0)
+        goto out;
 
     ret = find_string_and_symbol_tables(g_entrypoint_map.l_map_start, g_entrypoint_map.l_base_diff,
                                         &g_entrypoint_map.string_table,
@@ -584,11 +724,33 @@ int setup_pal_binary(void) {
     g_pal_map.l_base_diff = pal_base_diff;
     g_pal_map.l_ld = dynamic_section;
 
+    ret = verify_dynamic_entries(&g_pal_map);
+    if (ret < 0)
+        return ret;
+
     ret = find_string_and_symbol_tables(g_pal_map.l_map_start, g_pal_map.l_base_diff,
                                         &g_pal_map.string_table, &g_pal_map.symbol_table,
                                         &g_pal_map.symbol_table_cnt);
     if (ret < 0)
         return ret;
+
+    /* find soname of PAL binary -- will be used during DT_NEEDED verification of other binaries */
+    elf_dyn_t* dynamic_section_entry = dynamic_section;
+    elf_xword_t soname_offset = 0;
+    bool soname_offset_found  = false;
+    while (dynamic_section_entry->d_tag != DT_NULL) {
+        if (dynamic_section_entry->d_tag == DT_SONAME) {
+            soname_offset = dynamic_section_entry->d_un.d_val;
+            soname_offset_found = true;
+            break;
+        }
+        dynamic_section_entry++;
+    }
+    if (!soname_offset_found) {
+        log_error("Did not find DT_SONAME for PAL binary (name of the PAL library)");
+        return -PAL_ERROR_DENIED;
+    }
+    g_pal_soname = g_pal_map.string_table + soname_offset;
 
     ret = perform_relocations(&g_pal_map);
     return ret;


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1473)* status: 0 of 2 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1473#-NaByqwAC1huNyUzrfP3:-NaByqwAC1huNyUzrfP4:b-mj364t) (no related file):*
I tested this on Ubuntu 23.04 docker container (it has the required glibc/ld versions). I asked the linker to produce RELR relocs for the SGX PAL:
```diff
diff --git a/pal/src/host/linux-sgx/meson.build b/pal/src/host/linux-sgx/meson.build
@@ -110,6 +110,7 @@ libpal_sgx = shared_library('pal',
         # PAL is expected to be linked with RELRO (Relocation Read-Only)
         '-Wl,-zrelro',
         '-Wl,-znow',
+        '-Wl,-z,pack-relative-relocs',

         '-Wl,-zcombreloc',
         '-Wl,-zdefs',
```

Then I built Gramine and tried to run it:
```bash
root@b6db211304ab:/gramine/CI-Examples/helloworld# gramine-sgx helloworld

error: Unsupported relocation type DT_RELR; you may need to rebuild Gramine with `-Wl,-z,nopack-relative-relocs` linker option
error: Relocation of the PAL binary failed: -6
```

I even added `-Wl,-z,nopack-relative-relocs` to the linker options, and Gramine started to work fine again. So this PR works as expected.

___
___
*[`-- commits` line 14 at r1](https://reviewable.io/reviews/gramineproject/gramine/1473#-NaBubyg6aW9sN4Z9jGR):*
Actually, I don't think Gramine will ever need to take into account `RELR` relocations. I would simply force them to NOT be emitted, similarly how we do it with other linker configs e.g. here: https://github.com/gramineproject/gramine/blob/51d915ded30aa2ca55ad5c2e86075e95fe938ca2/libos/src/meson.build#L152-L165

These new RELR relocations can debloat the binaries' sizes about 5%. Given that this only concerns LibOS and Trusted-SGX PAL binaries, I just wouldn't care about this size improvement... We need to add more code (to work on these RELR relocs) for a dubious benefit of smaller Gramine executables. That's not the price I want to pay.

kailun-qin (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1473)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`pal/src/pal_rtld.c` line 232 at r1](https://reviewable.io/reviews/gramineproject/gramine/1473#-NaF7Nlg7Gmzvza2BhYl:-NaF7Nlg7Gmzvza2BhYm:b-3cnieq) ([raw file](https://github.com/gramineproject/gramine/blob/53883da309814762ae06a3cb926f2ac1ff6e8c24/pal/src/pal_rtld.c#L232)):*
> ```C
>                 break;
>             case DT_RELR:
>             case DT_RELRSZ:
> ```

Any special reason to not have `DT_RELRENT` here?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1473)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @kailun-qin)

___
*[`pal/src/pal_rtld.c` line 232 at r1](https://reviewable.io/reviews/gramineproject/gramine/1473#-NaF7Nlg7Gmzvza2BhYl:-NaFwoCBAS0J3Zufnh_P:be5avwm) ([raw file](https://github.com/gramineproject/gramine/blob/53883da309814762ae06a3cb926f2ac1ff6e8c24/pal/src/pal_rtld.c#L232)):*

Any special reason to not have `DT_RELRENT` here?
</blockquote></details>

I didn't want to add it because:
1. This is more like a metadata (size of the RELR entry), so even if this field exists in the binary, it doesn't affect any actual relocations.
2. Musl doesn't use it at all, glibc uses it only as a sanity check:
   - https://github.com/search?q=repo%3Abminor%2Fmusl%20DT_RELRENT%20&type=code
   - https://github.com/search?q=repo%3Abminor%2Fglibc%20DT_RELRENT%20&type=code

So I would be against failing on `DT_RELRENT`.

kailun-qin (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1473)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

___
*[`pal/src/pal_rtld.c` line 232 at r1](https://reviewable.io/reviews/gramineproject/gramine/1473#-NaF7Nlg7Gmzvza2BhYl:-NaG5KgC8ubKeDZIeejO:b-u7wv04) ([raw file](https://github.com/gramineproject/gramine/blob/53883da309814762ae06a3cb926f2ac1ff6e8c24/pal/src/pal_rtld.c#L232)):*

I didn't want to add it because:
1. This is more like a metadata (size of the RELR entry), so even if this field exists in the binary, it doesn't affect any actual relocations.
2. Musl doesn't use it at all, glibc uses it only as a sanity check:
   - https://github.com/search?q=repo%3Abminor%2Fmusl%20DT_RELRENT%20&type=code
   - https://github.com/search?q=repo%3Abminor%2Fglibc%20DT_RELRENT%20&type=code

So I would be against failing on `DT_RELRENT`.
</blockquote></details>

Great to know, thanks!

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1473)* status: all files reviewed, 3 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/1473#-NabLI3f7bN_TrD1v2B2):*
And please change the PR accordingly :)



_Suggestion:_
```
[PAL] Fail if Gramine binaries were built with unknown relocs types
```

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1473)* status: all files reviewed, 2 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`-- commits` line 14 at r1](https://reviewable.io/reviews/gramineproject/gramine/1473#-NaBubyg6aW9sN4Z9jGR):*

Actually, I don't think Gramine will ever need to take into account `RELR` relocations. I would simply force them to NOT be emitted, similarly how we do it with other linker configs e.g. here: https://github.com/gramineproject/gramine/blob/51d915ded30aa2ca55ad5c2e86075e95fe938ca2/libos/src/meson.build#L152-L165

These new RELR relocations can debloat the binaries' sizes about 5%. Given that this only concerns LibOS and Trusted-SGX PAL binaries, I just wouldn't care about this size improvement... We need to add more code (to work on these RELR relocs) for a dubious benefit of smaller Gramine executables. That's not the price I want to pay.
</blockquote></details>

Yeah, IMO it's not that important.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1473)* status: all files reviewed, 2 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/1473#-NabLI3f7bN_TrD1v2B2):*

I don't know how to do it. There is no list of relocations that I can allow.

I mean, relocations are mixed together with other information about "dynamic entries" (the `DT_` entries). See this list: https://github.com/gramineproject/gramine/blob/4de6aa5fb939a129d3498e8fee432d2873924b4c/pal/include/elf/elf.h#L622-L656

As you can see, there is **no** separate list of relocation types. So the only thing I can do is to whitelist all the not-breaking-Gramine `DT_` entries, but it will look ugly and surprising-to-readers:
```c
    while (dynamic_section_entry->d_tag != DT_NULL) {
        switch (dynamic_section_entry->d_tag) {
            case DT_RELA:
            case DT_RELASZ:
            case DT_JMPREL:
            case DT_PLTRELSZ:
                ... actual good relocation entries, do some job ...

            case DT_NEEDED:
            case DT_HASH:
            case <and so on>:
                ... whitelisted good not-relocation-type entries, ignore them ...
                break;

            default:
                log_error("Unsupported relocation type %d, terminating", dynamic_section_entry->d_tag);
                return -PAL_ERROR_DENIED;
            }
    }
```

Is this what'd you want? I'm hesitant to add a whitelist of ~30 entries...
</blockquote></details>

Yes, that's what I'm asking for, otherwise we're just silently ignoring semantically important entries and hoping for the best.

btw. `DT_NEEDED` probably can't be ignored?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1473)* status: 0 of 2 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " and "squash! " found in commit messages' one-liners (waiting on @dimakuv, @kailun-qin, and @mkow)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/1473#-NabLI3f7bN_TrD1v2B2):*

> Not sure what you mean by this comment. Please see my comment in the source code, maybe this will explain the situation.

> /* benign -- LibOS and PAL tests have PAL shared lib here, PAL has nothing */

But you could have said exactly the same about `DT_RELR` before the GCC update and justify ignoring it "because PAL currently is not using it for anything important".
What I mean is that `DT_NEEDED` is semantically important for ELF loading and we have to either implement it or error out. Ignoring it seems to be a bug in Gramine which just doesn't trigger currently.
</blockquote></details>

Done, good point. Now we are hardcore -- we check all `DT_` entries. Except for versioning ones and flags -- I don't know how to verify them, and if their verification would actually bring anything.
___
*[`-- commits` line 21 at r2](https://reviewable.io/reviews/gramineproject/gramine/1473#-NagKrob5KtH0IWHiNyN):*

Will do this: use this commit message at final rebase.
</blockquote></details>

I will also need to rephrase `unrecognized dyn entries` -> `unrecognized or unexpected/malformed dyn entries`

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1473)* status: 0 of 2 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " and "squash! " found in commit messages' one-liners (waiting on @kailun-qin and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1473#-NagLdVmBwTL1FkMuA5d:-NalFlyQE-eII2bHEjUt:b-f7mhvp) (no related file):*

Here is lists of all dynamic entries found in built Gramine binaries:

- LibOS:
```
  Tag        Type                         Name/Value
 0x0000000000000001 (NEEDED)             Shared library: [libpal.so]
 0x000000000000000e (SONAME)             Library soname: [libsysdb.so]
 0x0000000000000004 (HASH)               0x1f0
 0x0000000000000005 (STRTAB)             0x8b0
 0x0000000000000006 (SYMTAB)             0x370
 0x000000000000000a (STRSZ)              1112 (bytes)
 0x000000000000000b (SYMENT)             24 (bytes)
 0x0000000000000003 (PLTGOT)             0xe5da8
 0x0000000000000002 (PLTRELSZ)           1248 (bytes)
 0x0000000000000014 (PLTREL)             RELA
 0x0000000000000017 (JMPREL)             0xc440
 0x0000000000000007 (RELA)               0xdd0
 0x0000000000000008 (RELASZ)             46704 (bytes)
 0x0000000000000009 (RELAENT)            24 (bytes)
 0x000000006ffffffc (VERDEF)             0xd78
 0x000000006ffffffd (VERDEFNUM)          2
 0x000000000000001e (FLAGS)              BIND_NOW
 0x000000006ffffffb (FLAGS_1)            Flags: NOW
 0x000000006ffffffe (VERNEED)            0xdb0
 0x000000006fffffff (VERNEEDNUM)         1
 0x000000006ffffff0 (VERSYM)             0xd08
 0x000000006ffffff9 (RELACOUNT)          1944
 0x0000000000000000 (NULL)               0x0
```
- Linux PAL
```
  Tag        Type                         Name/Value
 0x000000000000000e (SONAME)             Library soname: [libpal.so]
 0x0000000000000004 (HASH)               0x1f0
 0x0000000000000005 (STRTAB)             0x8b0
 0x0000000000000006 (SYMTAB)             0x370
 0x000000000000000a (STRSZ)              1010 (bytes)
 0x000000000000000b (SYMENT)             24 (bytes)
 0x0000000000000007 (RELA)               0xd50
 0x0000000000000008 (RELASZ)             6168 (bytes)
 0x0000000000000009 (RELAENT)            24 (bytes)
 0x000000006ffffffc (VERDEF)             0xd18
 0x000000006ffffffd (VERDEFNUM)          2
 0x000000000000001e (FLAGS)              BIND_NOW
 0x000000006ffffffb (FLAGS_1)            Flags: NOW
 0x000000006ffffff0 (VERSYM)             0xca2
 0x000000006ffffff9 (RELACOUNT)          257
 0x0000000000000000 (NULL)               0x0
```
- Linux-SGX PAL (trusted part)
```
  Tag        Type                         Name/Value
 0x000000000000000e (SONAME)             Library soname: [libpal.so]
 0x0000000000000004 (HASH)               0x618c8
 0x0000000000000005 (STRTAB)             0x730
 0x0000000000000006 (SYMTAB)             0x1f0
 0x000000000000000a (STRSZ)              1064 (bytes)
 0x000000000000000b (SYMENT)             24 (bytes)
 0x0000000000000003 (PLTGOT)             0x62fe0
 0x0000000000000002 (PLTRELSZ)           24 (bytes)
 0x0000000000000014 (PLTREL)             RELA
 0x0000000000000017 (JMPREL)             0x2688
 0x0000000000000007 (RELA)               0xc00
 0x0000000000000008 (RELASZ)             6792 (bytes)
 0x0000000000000009 (RELAENT)            24 (bytes)
 0x000000006ffffffc (VERDEF)             0xbc8
 0x000000006ffffffd (VERDEFNUM)          2
 0x000000000000001e (FLAGS)              BIND_NOW
 0x000000006ffffffb (FLAGS_1)            Flags: NOW
 0x000000006ffffff0 (VERSYM)             0xb58
 0x000000006ffffff9 (RELACOUNT)          283
 0x0000000000000000 (NULL)               0x0
```
</blockquote></details>

When built with ASan, also `DT_INIT_ARRAY`, `DT_FINI_ARRAY`, `DT_INIT_ARRAYSZ`, `DT_FINI_ARRAYSZ` are added.

___

mkow (changes requested):
> Reviewed 1 of 2 files at r2, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1473)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " and "squash! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/1473#-NabLI3f7bN_TrD1v2B2):*

Done, good point. Now we are hardcore -- we check all `DT_` entries. Except for versioning ones and flags -- I don't know how to verify them, and if their verification would actually bring anything.
</blockquote></details>

Ok, now we stopped ignoring semantically meaningful entries, but also try to check unused entries (like debug info) for specific values we expect from our current PAL binary. What's the point of that? This may actually change between compilers and fail our builds.

What I originally meant is that we cannot just ignore some DT entries if they are supposed to be interpreted by the loader (like `DT_NEEDED`). But I don't really see a point in hardcoding in the loader how the PAL binary should look like.
___
*[`pal/src/pal_rtld.c` line 270 at r4](https://reviewable.io/reviews/gramineproject/gramine/1473#-NasqcgtB8aNWUskRBKg:-NasqcgtB8aNWUskRBKh:b47474j) ([raw file](https://github.com/gramineproject/gramine/blob/f5799e2591f04a9488225ad9ad5a684ae7f79978/pal/src/pal_rtld.c#L270)):*
> ```C
>                  * relevant for our linker */
>                 if (dynamic_section_entry->d_un.d_val != 0) {
>                     log_error("Unexpected DT_DEBUG (must be zero)");
> ```

What's the point of failing here? We don't use it for anything anyways and it's semantically a no-op.
___
*[`pal/src/pal_rtld.c` line 317 at r4](https://reviewable.io/reviews/gramineproject/gramine/1473#-NaspSTl7KPX9nrknvLs:-NaspSTm9CEtkjmoeumW:b-aajuou) ([raw file](https://github.com/gramineproject/gramine/blob/f5799e2591f04a9488225ad9ad5a684ae7f79978/pal/src/pal_rtld.c#L317)):*
> ```C
> 
>     if (!string_table || !string_table_size) {
>         log_error("Loaded binary doesn't have string table or its empty (DT_STRTAB and DT_STRSZ)");
> ```

Is this a problem? I mean, wouldn't the loader clearly fail about this later, when actually trying to use it? And if it doesn't use it, then why failing here?
___
*[`pal/src/pal_rtld.c` line 326 at r4](https://reviewable.io/reviews/gramineproject/gramine/1473#-Nasotqx8EpWjAtkXDuJ:-Nasotqx8EpWjAtkXDuK:b-po8ydd) ([raw file](https://github.com/gramineproject/gramine/blob/f5799e2591f04a9488225ad9ad5a684ae7f79978/pal/src/pal_rtld.c#L326)):*
> ```C
>         const char* soname = string_table + soname_offset;
>         if (strcmp(soname, "libsysdb.so") != 0 && strcmp(soname, "libpal.so") != 0) {
>             log_error("Unexpected DT_SONAME (must be name of the LibOS or PAL library)");
> ```

What's the reason for having these checks? Do we rely on `DT_SONAME` being exactly that anywhere?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1473)* status: 1 of 2 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " and "squash! " found in commit messages' one-liners (waiting on @mkow)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/1473#-NabLI3f7bN_TrD1v2B2):*

Ok, now we stopped ignoring semantically meaningful entries, but also try to check unused entries (like debug info) for specific values we expect from our current PAL binary. What's the point of that? This may actually change between compilers and fail our builds.

What I originally meant is that we cannot just ignore some DT entries if they are supposed to be interpreted by the loader (like `DT_NEEDED`). But I don't really see a point in hardcoding in the loader how the PAL binary should look like.
</blockquote></details>

Done.
___
*[`pal/src/pal_rtld.c` line 270 at r4](https://reviewable.io/reviews/gramineproject/gramine/1473#-NasqcgtB8aNWUskRBKg:-NauCl3g2M4gI_xwlUXr:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/f5799e2591f04a9488225ad9ad5a684ae7f79978/pal/src/pal_rtld.c#L270)):*

What's the point of failing here? We don't use it for anything anyways and it's semantically a no-op.
</blockquote></details>

Done.
___
*[`pal/src/pal_rtld.c` line 317 at r4](https://reviewable.io/reviews/gramineproject/gramine/1473#-NaspSTl7KPX9nrknvLs:-NauAx381ZQOlf5w9F7w:b-ayw2bz) ([raw file](https://github.com/gramineproject/gramine/blob/f5799e2591f04a9488225ad9ad5a684ae7f79978/pal/src/pal_rtld.c#L317)):*

Is this a problem? I mean, wouldn't the loader clearly fail about this later, when actually trying to use it? And if it doesn't use it, then why failing here?
</blockquote></details>

Yes, the loader would clearly fail about this later...

However, I added this check here because I am using `string_table` below, to verify `DT_NEEDED`.
___
*[`pal/src/pal_rtld.c` line 326 at r4](https://reviewable.io/reviews/gramineproject/gramine/1473#-Nasotqx8EpWjAtkXDuJ:-NauCvR11yaxaNfHGd9f:bd1j4ie) ([raw file](https://github.com/gramineproject/gramine/blob/f5799e2591f04a9488225ad9ad5a684ae7f79978/pal/src/pal_rtld.c#L326)):*

What's the reason for having these checks? Do we rely on `DT_SONAME` being exactly that anywhere?
</blockquote></details>

Done.

I don't hard-code the SONAMEs anymore, but I memorize the SONAME of the PAL binary now (`libpal.so`), and I compare NEEDED against this SONAME.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1473)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " and "squash! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`pal/src/pal_rtld.c` line 213 at r5](https://reviewable.io/reviews/gramineproject/gramine/1473#-Navgbby9HbmfdgEGr4E:-Navgbby9HbmfdgEGr4F:bu8gjq6) ([raw file](https://github.com/gramineproject/gramine/blob/e18f91f36e7da2c9148d7b459375672fdc0a2454/pal/src/pal_rtld.c#L213)):*
It should still be plural:

> The dynamic array can contain multiple entries of this type.



_Suggestion:_
```C
DT_NEEDED entries contain offsets
```
___
*[`pal/src/pal_rtld.c` line 732 at r5](https://reviewable.io/reviews/gramineproject/gramine/1473#-NavgyOe6qN0kXLDsCha:-NavgyOe6qN0kXLDsChb:bfutn6j) ([raw file](https://github.com/gramineproject/gramine/blob/e18f91f36e7da2c9148d7b459375672fdc0a2454/pal/src/pal_rtld.c#L732)):*
> ```C
>         return ret;
> 
>     /* find soname of the PAL binary -- will be used during DT_NEEDED verification of other binaries */
> ```

please wrap
___
*[`pal/src/pal_rtld.c` line 738 at r5](https://reviewable.io/reviews/gramineproject/gramine/1473#-NavhGNF9q9-5wuqajMb:-NavhGNGDYg1AjXiz3Ox:b-gpvs6m) ([raw file](https://github.com/gramineproject/gramine/blob/e18f91f36e7da2c9148d7b459375672fdc0a2454/pal/src/pal_rtld.c#L738)):*
> ```C
>     while (dynamic_section_entry->d_tag != DT_NULL) {
>         switch (dynamic_section_entry->d_tag) {
>             case DT_SONAME:
> ```

Why not just `if`?



_Code quote:_
```C
        switch (dynamic_section_entry->d_tag) {
            case DT_SONAME:
```

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1473)* status: 1 of 2 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " and "squash! " found in commit messages' one-liners (waiting on @mkow)

___
*[`pal/src/pal_rtld.c` line 213 at r5](https://reviewable.io/reviews/gramineproject/gramine/1473#-Navgbby9HbmfdgEGr4E:-NavosJ185qOjyQ9JeSg:bid0dub) ([raw file](https://github.com/gramineproject/gramine/blob/e18f91f36e7da2c9148d7b459375672fdc0a2454/pal/src/pal_rtld.c#L213)):*

It should still be plural:

> The dynamic array can contain multiple entries of this type.
</blockquote></details>

Done.

Ok, I added one more check to make sure that we only have one `DT_NEEDED` entry -- because in Gramine, we only have one dependency -- the PAL binary. If e.g. LibOS would have two dependencies, then this is not supported by our current loader (and generally I think it would go against the whole Gramine design).
___
*[`pal/src/pal_rtld.c` line 732 at r5](https://reviewable.io/reviews/gramineproject/gramine/1473#-NavgyOe6qN0kXLDsCha:-Navp_wt8kCcYdNxT70H:b-l3mqsu) ([raw file](https://github.com/gramineproject/gramine/blob/e18f91f36e7da2c9148d7b459375672fdc0a2454/pal/src/pal_rtld.c#L732)):*

please wrap
</blockquote></details>

Done. Removed `the` instead because I don't like wrapping.
___
*[`pal/src/pal_rtld.c` line 738 at r5](https://reviewable.io/reviews/gramineproject/gramine/1473#-NavhGNF9q9-5wuqajMb:-Navpmue9xoeP0z5VAnb:bc7c1r8) ([raw file](https://github.com/gramineproject/gramine/blob/e18f91f36e7da2c9148d7b459375672fdc0a2454/pal/src/pal_rtld.c#L738)):*

Why not just `if`?
</blockquote></details>

Done. Copy-paste fail.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1473)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " and "squash! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`pal/src/pal_rtld.c` line 738 at r5](https://reviewable.io/reviews/gramineproject/gramine/1473#-NavhGNF9q9-5wuqajMb:-NaxUj7rBmpULYzI_L9O:bln74ba) ([raw file](https://github.com/gramineproject/gramine/blob/e18f91f36e7da2c9148d7b459375672fdc0a2454/pal/src/pal_rtld.c#L738)):*

Done. Copy-paste fail.
</blockquote></details>

btw. the `break` also didn't work as intended there, it only jumped out of the switch, not the `while` loop.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1473)* status: all files reviewed, 3 unresolved discussions (waiting on @dimakuv)

___
*[`-- commits` line 21 at r2](https://reviewable.io/reviews/gramineproject/gramine/1473#-NagKrob5KtH0IWHiNyN):*

I will also need to rephrase `unrecognized dyn entries` -> `unrecognized or unexpected/malformed dyn entries`
</blockquote></details>

Done, with slight shortening (it still exceeds the limit, but I don't have any further ideas).

Issue #1099, opened 2023-01-04
[LibOS,PAL/Linux-SGX] EDMM: Introduce lazy allocation


dimakuv:
### Description of the feature

PR #1054 implemented an initial version of EDMM support.

In particular, every `mmap(addr, size)` request ends up allocating the range of enclave pages `[addr, addr+size)`, via the call to `pal_memory.c: sgx_edmm_add_pages()` which in turn does a loop `sgx_eaccept` + restrict/expand permissions over all pages in the range. In other words, the whole mmapped region is pre-accepted.

In some cases, applications may rely on lazy allocation of pages, where the VMAs are reserved but not actually committed to physical memory. In particular, `mmap(..., MAP_NORESERVE)` requests are used in such cases -- to mmap a huge chunk of memory (possibly never used in the future) at once and then commit pages on demand on page fault events.

So, our initial implementation of EDMM support doesn't have this concept of lazy allocation. Ideally, we would pre-accept the mmapped range only on some `mmap()` requests, and defer page accepts to page-fault events on other `mmap()` requests. One obvious heuristic to defer page accepts is when Gramine notices `MAP_NORESERVE` flag in the mmap request.

### Why Gramine should implement it?

Performance reasons, as well as to decrease the amount of required EPC (physical SGX memory). E.g., Java runtime may issue `mmap(64GB, MAP_NORESERVE)` -- current EDMM implementation will spend a lot of time and physical memory on allocating + accepting all 64GB of enclave pages. The improved implementation would not allocate these 64GB enclave pages at all (only the actually required subset of pages will be allocated + accepted during page fault handling).

vijaydhanraj:
Hi @dimakuv, please assign this to me.

kailun-qin:
I'm not quite familiar w/ the context and was just looking into this requirement. I have the following assumptions and would like to have more discussions/clarifications.
1. We'd like to support `mmap(..., MAP_NORESERVE)` only for Linux-SGX PAL and when EDMM is enabled, i.e., Linux PAL should work as-is which ignores `MAP_NORESERVE`.
2. I'd image some potential changes below:
* Extend `filter_saved_flags()`: https://github.com/gramineproject/gramine/blob/51d915ded30aa2ca55ad5c2e86075e95fe938ca2/libos/src/bookkeep/libos_vma.c#L35-L39 so that `MAP_NORESERVE` flag can be saved in VMAs.
[...]

PR #1502, opened 2023-08-22
[PAL/Linux-SGX] Disallow memfault handling if `sgx.require_exinfo = false`

kailun-qin:
## Description of the changes 
Previously, Linux-SGX PAL did not check `sgx.require_exinfo` and propogated the address anyway to LibOS/app. However, LibOS can't/shouldn't handle faulting addresses at all if `sgx.require_exinfo = false`, as in this case, no matter what enclave address LibOS/app tried to access, SGX would always report address 0.

This commit adds a check on whether `sgx.require_exinfo` is enabled when a memory fault occurred/to be handled and we loudly fail if it's not.

For more context, see discussions here: https://github.com/gramineproject/gramine/issues/1099#issuecomment-1663842018.

## How to test this PR? 

CI + manual testing w/ a purpose-built memfault app.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)

### DIFF
--- a/pal/regression/Exception.c
+++ b/pal/regression/Exception.c
@@ -33,27 +33,18 @@ static void handler2(bool is_in_pal, uintptr_t addr, PAL_CONTEXT* context) {
         context->rip++;
 }
 
+atomic_bool handler3_called = false;
+
 static void handler3(bool is_in_pal, uintptr_t addr, PAL_CONTEXT* context) {
     __UNUSED(is_in_pal);
 
-    pal_printf("Memory Fault Exception Handler: 0x%08lx, rip = 0x%08lx\n", addr, context->rip);
-
-    while (*(unsigned char*)context->rip != 0x90)
-        context->rip++;
-}
-
-atomic_bool handler4_called = false;
-
-static void handler4(bool is_in_pal, uintptr_t addr, PAL_CONTEXT* context) {
-    __UNUSED(is_in_pal);
-
-    pal_printf("Arithmetic Exception Handler 4: 0x%" PRIx64 ", rip = 0x%" PRIx64 "\n", addr,
+    pal_printf("Arithmetic Exception Handler 3: 0x%" PRIx64 ", rip = 0x%" PRIx64 "\n", addr,
                context->rip);
 
     while (*(unsigned char*)context->rip != 0x90)
         context->rip++;
 
-    handler4_called = true;
+    handler3_called = true;
 }
 
 static void red_zone_test(void) {
@@ -94,7 +85,7 @@ static void red_zone_test(void) {
         :
         : "rax", "rbx", "rcx", "rdx", "cc", "memory");
 
-    if (!handler4_called) {
+    if (!handler3_called) {
         pal_printf("Exception handler was not called!\n");
         return;
     }
@@ -105,6 +96,20 @@ static void red_zone_test(void) {
     }
 
     pal_printf("Red zone test ok.\n");
+}
+
+#define INVALID_ADDR 0x1000
+
+static void handler4(bool is_in_pal, uintptr_t addr, PAL_CONTEXT* context) {
+    __UNUSED(is_in_pal);
+
+    pal_printf("Memory Fault Exception Handler: 0x%08lx, rip = 0x%08lx\n", addr, context->rip);
+
+    if (addr != INVALID_ADDR)
+        pal_printf("Wrong faulting address: expected: 0x%08x, got: 0x%08lx\n", INVALID_ADDR, addr);
+
+    while (*(unsigned char*)context->rip != 0x90)
+        context->rip++;
 }
 
 int main(void) {
@@ -128,12 +133,15 @@ int main(void) {
             "nop\n"
             ::: "rax", "rbx", "rdx", "cc");
 
-    PalSetExceptionHandler(handler3, PAL_EVENT_MEMFAULT);
-    *(volatile long*)0x1000 = 0;
-    __asm__ volatile("nop");
-
-    PalSetExceptionHandler(handler4, PAL_EVENT_ARITHMETIC_ERROR);
+    PalSetExceptionHandler(handler3, PAL_EVENT_ARITHMETIC_ERROR);
     red_zone_test();
+
+    /* For Gramine-SGX, memory fault exception handling test will error out and fail on machines
+     * that do not support EXINFO. We thus put it as the last test so that we could still check on
+     * the other tests in such case. */
+    PalSetExceptionHandler(handler4, PAL_EVENT_MEMFAULT);
+    *(volatile long*)INVALID_ADDR = 0;
+    __asm__ volatile("nop");
 
     return 0;
 }
--- a/pal/regression/manifest.template
+++ b/pal/regression/manifest.template
@@ -5,6 +5,9 @@ loader.insecure__use_cmdline_argv = true
 sgx.debug = true
 sgx.edmm_enable = {{ 'true' if env.get('EDMM', '0') == '1' else 'false' }}
 
+# for Exception test (memfault handler)
+sgx.use_exinfo = {{ 'true' if env.get('EDMM', '0') == '1' else 'false' }}
+
 sgx.allowed_files = [
   "file:test.txt", # for File2 test
   "file:to_send.tmp", # for PalSendHandle test
--- a/pal/regression/test_pal.py
+++ b/pal/regression/test_pal.py
@@ -203,25 +203,31 @@ class TC_10_Exception(RegressionTestCase):
 
     @unittest.skipUnless(ON_X86, "x86-specific")
     def test_000_exception(self):
-        _, stderr = self.run_binary(['Exception'])
+        try:
+            _, stderr = self.run_binary(['Exception'])
+            if HAS_SGX and not HAS_EDMM:
+                self.fail('expected to return nonzero')
+        except subprocess.CalledProcessError as e:
+            if HAS_SGX and not HAS_EDMM:
+                self.assertNotEqual(e.returncode, 0)
+                stderr = e.stderr.decode()
+            else:
+                self.fail('expected to return zero')
 
         self.assertTrue(self.is_altstack_different_from_main_stack(stderr))
 
         # Exception Handling (Div-by-Zero)
-        self.assertIn('Arithmetic Exception Handler', stderr)
-
-        # Exception Handling (Memory Fault)
-        self.assertIn('Memory Fault Exception Handler', stderr)
-
-        # Exception Handler Swap
         self.assertIn('Arithmetic Exception Handler 1', stderr)
         self.assertIn('Arithmetic Exception Handler 2', stderr)
 
-        # Exception Handling (Set Context)
-        self.assertIn('Arithmetic Exception Handler 1', stderr)
-
         # Exception Handling (Red zone)
+        self.assertIn('Arithmetic Exception Handler 3', stderr)
         self.assertIn('Red zone test ok.', stderr)
+
+        if not HAS_SGX or HAS_EDMM:
+            # Exception Handling (Memory Fault)
+            self.assertIn('Memory Fault Exception Handler', stderr)
+            self.assertNotIn('Wrong faulting address', stderr)
 
 class TC_20_SingleProcess(RegressionTestCase):
     def test_000_exit_code(self):
--- a/pal/src/host/linux-sgx/pal_exception.c
+++ b/pal/src/host/linux-sgx/pal_exception.c
@@ -303,6 +303,8 @@ void _PalExceptionHandler(unsigned int exit_info, sgx_cpu_context_t* uc,
     PAL_CONTEXT ctx = { 0 };
     save_pal_context(&ctx, uc, xregs_state);
 
+    bool has_hw_fault_address = false;
+
     if (ei.info.valid) {
         ctx.trapno = ei.info.vector;
         /* Only these two exceptions save information in EXINFO. */
@@ -312,6 +314,7 @@ void _PalExceptionHandler(unsigned int exit_info, sgx_cpu_context_t* uc,
         }
         if (ei.info.vector == SGX_EXCEPTION_VECTOR_PF) {
             ctx.cr2 = exinfo->maddr;
+            has_hw_fault_address = true;
         }
     }
 
@@ -321,6 +324,11 @@ void _PalExceptionHandler(unsigned int exit_info, sgx_cpu_context_t* uc,
             addr = uc->rip;
             break;
         case PAL_EVENT_MEMFAULT:
+            if (!has_hw_fault_address) {
+                log_error("Tried to handle a memory fault with no faulting address reported by "
+                          "SGX. Please consider enabling 'sgx.use_exinfo' in the manifest.");
+                _PalProcessExit(1);
+            }
             addr = ctx.cr2;
             break;
         default:


dimakuv (changes requested):
> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @kailun-qin)

___
*[`pal/regression/manifest.template` line 7 at r2](https://reviewable.io/reviews/gramineproject/gramine/1502#-NcQNno80h7C8vxNrgGF:-NcQNno80h7C8vxNrgGG:bz033s5) ([raw file](https://github.com/gramineproject/gramine/blob/87b729c19a27a9e50634242176c2e29ed343abd1/pal/regression/manifest.template#L7)):*
> ```
> sgx.debug = true
> sgx.edmm_enable = {{ 'true' if env.get('EDMM', '0') == '1' else 'false' }}
> sgx.require_exinfo = true # for Exception test
> ```

You can't do this because not all SGX machines support EXINFO. You should do like this: https://github.com/gramineproject/examples/blob/957f0617e1ada4489874d33f620ff0f82ef3aadb/openjdk/java.manifest.template#L26

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1502#-NcQO7cgAAQdmMojcpSJ:-NcQO7cgAAQdmMojcpSK:b9hhhm4) (no related file):*
@boryspoplawski I wonder if you could take a quick look at the PR and say if the fix makes sense to you (you were the one working on EDMM).

___

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`pal/regression/manifest.template` line 7 at r2](https://reviewable.io/reviews/gramineproject/gramine/1502#-NcQNno80h7C8vxNrgGF:-NcQVGnwDQSF28srsfPR:b-os1ryl) ([raw file](https://github.com/gramineproject/gramine/blob/87b729c19a27a9e50634242176c2e29ed343abd1/pal/regression/manifest.template#L7)):*

You can't do this because not all SGX machines support EXINFO. You should do like this: https://github.com/gramineproject/examples/blob/957f0617e1ada4489874d33f620ff0f82ef3aadb/openjdk/java.manifest.template#L26
</blockquote></details>

Indeed, I overlooked it. But this PR actually makes [Exception test](https://github.com/gramineproject/gramine/blob/9ef75ebc18ab9b390e7c248d6068237d433a4b20/pal/regression/Exception.c) (specifically handler3) work only on machines that support EXINFO.

So looking at this PR again, I'm wondering what's the right behavior/fix for the apps that registered their own handlers, particularly when they run on machines that do not support EXINFO. Should they expect failure (w/ nothing they could actually do because EXINFO is not supported) even if they don't really use those faulting addresses?

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @kailun-qin)

___
*[`pal/regression/manifest.template` line 7 at r2](https://reviewable.io/reviews/gramineproject/gramine/1502#-NcQNno80h7C8vxNrgGF:-NcY5p0z8J2siCbji-Vi:b-4znz67) ([raw file](https://github.com/gramineproject/gramine/blob/87b729c19a27a9e50634242176c2e29ed343abd1/pal/regression/manifest.template#L7)):*
The `Exception` PAL test is terrible... My SGX machine without EXINFO shows this when I run it manually:
```bash
# note that direct version shows correct faulting addresses
$ gramine-direct Exception
Arithmetic Exception Handler 1: 0x004013f3, rip = 0x004013f3
Arithmetic Exception Handler 2: 0x00401416, rip = 0x00401416
Memory Fault Exception Handler: 0x00001000, rip = 0x0040142e
Arithmetic Exception Handler 4: 0x401314, rip = 0x401314

# note that SGX version always shows 0x00 faulting addresses
$ gramine-sgx Exception
Arithmetic Exception Handler 1: 0x00000000, rip = 0x004013f3
Arithmetic Exception Handler 2: 0x00000000, rip = 0x00401416
Memory Fault Exception Handler: 0x00000000, rip = 0x0040142e
Arithmetic Exception Handler 4: 0x0, rip = 0x401314
```

This `Exception` test doesn't really test the most important thing -- on which accessed address the exception happened.

So, the `Exception` test works even on no-EXINFO machines, and I don't see how your PR would change it. (Again, this test works on these older machines simply because the test is really bad and doesn't check the accessed addresses in the exception handlers.)

> So looking at this PR again, I'm wondering what's the right behavior/fix for the apps that registered their own handlers, particularly when they run on machines that do not support EXINFO. Should they expect failure (w/ nothing they could actually do because EXINFO is not supported) even if they don't really use those faulting addresses?

I think such apps should expect failure. Otherwise Gramine is just "pretending" that everything is fine, and hands `0x0` faulting address always to the application, so that the application will almost certainly will do a wrong thing!

Why do you say `even if they don't really use those faulting addresses`? In your PR, the termination happens only if the app tried to use the faulting address. If the app never accesses inaccessible addresses, we'll never hit the `PAL_EVENT_MEMFAULT`.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`pal/regression/manifest.template` line 7 at r2](https://reviewable.io/reviews/gramineproject/gramine/1502#-NcQNno80h7C8vxNrgGF:-Nc_N83VDdjpn2qg9fEY:b-5xn9m5) ([raw file](https://github.com/gramineproject/gramine/blob/87b729c19a27a9e50634242176c2e29ed343abd1/pal/regression/manifest.template#L7)):*
> So, the `Exception` test works even on no-EXINFO machines, and I don't see how your PR would change it.

But now that this PR proposed to disallow memfault handling (by error out and exit) if `sgx.require_exinfo = false` (that's the case on no-EXINFO machines), so the `Exception` test (handler3) would simply fail on no-EXINFO machines w/ this PR.

> Why do you say `even if they don't really use those faulting addresses`? In your PR, the termination happens only if the app tried to use the faulting address. If the app never accesses inaccessible addresses, we'll never hit the `PAL_EVENT_MEMFAULT`.

Yes, sorry for the confusions. When I said "use...", I actually meant it in the the registered handlers, i.e., the handler does not inspect/manipulate the faulting address but works on other context like rip etc.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @kailun-qin)

___
*[`pal/regression/manifest.template` line 7 at r2](https://reviewable.io/reviews/gramineproject/gramine/1502#-NcQNno80h7C8vxNrgGF:-Nc_YdMY4zkkHo0nSrtb:bnndk7g) ([raw file](https://github.com/gramineproject/gramine/blob/87b729c19a27a9e50634242176c2e29ed343abd1/pal/regression/manifest.template#L7)):*
> But now that this PR proposed to disallow memfault handling (by error out and exit) if `sgx.require_exinfo = false` (that's the case on no-EXINFO machines), so the `Exception` test (handler3) would simply fail on no-EXINFO machines w/ this PR.

Ah, right. I think you can modify the `Exception` test, so that Handler 3 is only checked if EDMM is enabled (you'll need to modify the `test_pal.py` for this). We have some similar checks for LibOS tests (see `libos/test/regression/test_libos.py` for examples).

Also, since we're here, could you also add the check on the faulting address in the `Exception` test? Otherwise the test of Handler 3 is useless.

> Yes, sorry for the confusions. When I said "use...", I actually meant it in the the registered handlers, i.e., the handler does not inspect/manipulate the faulting address but works on other context like rip etc.

The #PF handler that does **not** inspect the faulting address makes no sense. No real-world handler would do what the `Exception` test handler 3 does (blindly increment RIP until we are in the safe zone). Unless you find an example of apps that don't inspect the faulting address, I have a strong opinion of just terminating the whole process, as you do in this PR.

Btw, your PR may help detect issues like https://github.com/gramineproject/gramine/discussions/1507 (I'm not sure if that's their problem, but does look like it). Currently, we just have to guess what these `memory fault at 0x00000000` mean, but with your PR we'll have a clear error message.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @kailun-qin)

___
*[`pal/regression/manifest.template` line 7 at r2](https://reviewable.io/reviews/gramineproject/gramine/1502#-NcQNno80h7C8vxNrgGF:-Nc__8v5C-e3Xyxuy4mP:b-11ihc2) ([raw file](https://github.com/gramineproject/gramine/blob/87b729c19a27a9e50634242176c2e29ed343abd1/pal/regression/manifest.template#L7)):*

> But now that this PR proposed to disallow memfault handling (by error out and exit) if `sgx.require_exinfo = false` (that's the case on no-EXINFO machines), so the `Exception` test (handler3) would simply fail on no-EXINFO machines w/ this PR.

Ah, right. I think you can modify the `Exception` test, so that Handler 3 is only checked if EDMM is enabled (you'll need to modify the `test_pal.py` for this). We have some similar checks for LibOS tests (see `libos/test/regression/test_libos.py` for examples).

Also, since we're here, could you also add the check on the faulting address in the `Exception` test? Otherwise the test of Handler 3 is useless.

> Yes, sorry for the confusions. When I said "use...", I actually meant it in the the registered handlers, i.e., the handler does not inspect/manipulate the faulting address but works on other context like rip etc.

The #PF handler that does **not** inspect the faulting address makes no sense. No real-world handler would do what the `Exception` test handler 3 does (blindly increment RIP until we are in the safe zone). Unless you find an example of apps that don't inspect the faulting address, I have a strong opinion of just terminating the whole process, as you do in this PR.

Btw, your PR may help detect issues like https://github.com/gramineproject/gramine/discussions/1507 (I'm not sure if that's their problem, but does look like it). Currently, we just have to guess what these `memory fault at 0x00000000` mean, but with your PR we'll have a clear error message.
</blockquote></details>

[ Hm, the link to https://github.com/gramineproject/gramine/discussions/1507 didn't work in the comment above. Adding a new comment with this link. ]

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @kailun-qin)

___
*[`pal/regression/manifest.template` line 7 at r2](https://reviewable.io/reviews/gramineproject/gramine/1502#-NcQNno80h7C8vxNrgGF:-Nc__GkjBApvVxIQuKhX:b-4ui3df) ([raw file](https://github.com/gramineproject/gramine/blob/87b729c19a27a9e50634242176c2e29ed343abd1/pal/regression/manifest.template#L7)):*

[ Hm, the link to https://github.com/gramineproject/gramine/discussions/1507 didn't work in the comment above. Adding a new comment with this link. ]
</blockquote></details>

[ Ah, no, links work, but at least for me, they are not highlighted by color or in any other way. Weird, this is a new behavior in Reviewable. ]

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: 1 of 4 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`pal/regression/manifest.template` line 7 at r2](https://reviewable.io/reviews/gramineproject/gramine/1502#-NcQNno80h7C8vxNrgGF:-NcaxV0LDmpp2lGJRf-L:b-pz51d3) ([raw file](https://github.com/gramineproject/gramine/blob/87b729c19a27a9e50634242176c2e29ed343abd1/pal/regression/manifest.template#L7)):*
Done.

> Should default value of `sgx.require_exinfo` be `true`?

No. Setting the default value of `sgx.require_exinfo` to be `true` will make Gramine not work on no-EXINFO machines.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

___
*[`pal/regression/Exception.c` line 109 at r3](https://reviewable.io/reviews/gramineproject/gramine/1502#-NcbtOUy9_rwVV27QIzJ:-NcbtOUy9_rwVV27QIzK:b4pk7up) ([raw file](https://github.com/gramineproject/gramine/blob/d5338a677aad639aef6dd0599277f5cdb6abeef5/pal/regression/Exception.c#L109)):*
> ```C
> 
>     if (addr != INVALID_ADDR) {
>         pal_printf("Faulting address not match: expected: 0x%08x, got: 0x%08lx\n",
> ```

`Faulting address not match` -> `Wrong faulting address`
___
*[`pal/regression/manifest.template` line 7 at r3](https://reviewable.io/reviews/gramineproject/gramine/1502#-NcbtqZYACTvGDYRpYTj:-NcbtqZYACTvGDYRpYTk:b61cyyq) ([raw file](https://github.com/gramineproject/gramine/blob/d5338a677aad639aef6dd0599277f5cdb6abeef5/pal/regression/manifest.template#L7)):*
> ```
> sgx.debug = true
> sgx.edmm_enable = {{ 'true' if env.get('EDMM', '0') == '1' else 'false' }}
> sgx.require_exinfo = {{ 'true' if env.get('EDMM', '0') == '1' else 'false' }} # for Exception test (memfault handler)
> ```

Exceeds 100 char limit, hard to read. I suggest:
```
...
sgx.edmm_enable = {{ 'true' if env.get('EDMM', '0') == '1' else 'false' }}

# for Exception test (memfault handler)
sgx.require_exinfo = {{ 'true' if env.get('EDMM', '0') == '1' else 'false' }}
```
___
*[`pal/regression/test_pal.py` line 212 at r3](https://reviewable.io/reviews/gramineproject/gramine/1502#-NcbueKT9jD9B7JMtcOs:-NcbueKT9jD9B7JMtcOt:bhfwyj1) ([raw file](https://github.com/gramineproject/gramine/blob/d5338a677aad639aef6dd0599277f5cdb6abeef5/pal/regression/test_pal.py#L212)):*
> ```Python
>         except subprocess.CalledProcessError as e:
>             if HAS_SGX and not HAS_EDMM:
>                 self.assertEqual(e.returncode, 1)
> ```

Why are we sure the process will terminate with code `1`? I think it's an internal detail of Gramine's PAL... Maybe simply check `assertNotEqual(e.returncode, 0)`?
___
*[`pal/regression/test_pal.py` line 230 at r3](https://reviewable.io/reviews/gramineproject/gramine/1502#-Ncbu_MEB6A-SDt21eSg:-Ncbu_MEB6A-SDt21eSh:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/d5338a677aad639aef6dd0599277f5cdb6abeef5/pal/regression/test_pal.py#L230)):*
> ```Python
>                 # Exception Handling (Memory Fault)
>                 self.assertIn('Memory Fault Exception Handler', stderr)
>                 self.assertNotIn('Faulting address not match', stderr)
> ```

ditto

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1502#-Ndej1oT0R3eYTdc_4dL:-Ndej1oUCO4bHN8sN7EH:bwj8je) (no related file):*
Do you think it's better to completely disable workloads that use fault handling on non-SGX2/exinfo capable machines?
Doesn't some of them work correctly with `0`/`NULL` as faulting address (e.g. java worked with that)?

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1502#-Ndej1oT0R3eYTdc_4dL:-Ndekh09CTqIGge375mb:b-f62bz8) (no related file):*
> Do you think it's better to completely disable workloads that use fault handling on non-SGX2/exinfo capable machines?

Hard to say... But we are risking confusing applications, and creating hard-to-chase bugs. Isn't it better to fail early, rather than "return 0 and hope that it will be alright"?

Also, I do feel like non-SGX2 machines are a thing of the past. IIRC, all SGX machines starting from 2017 have exinfo.

___

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1502#-Ndej1oT0R3eYTdc_4dL:-Ndeumj5AcA7KTUA7474:b-zcgjlw) (no related file):*

> Do you think it's better to completely disable workloads that use fault handling on non-SGX2/exinfo capable machines?

Hard to say... But we are risking confusing applications, and creating hard-to-chase bugs. Isn't it better to fail early, rather than "return 0 and hope that it will be alright"?

Also, I do feel like non-SGX2 machines are a thing of the past. IIRC, all SGX machines starting from 2017 have exinfo.
</blockquote></details>

But IIRC `require_exinfo` was exactly for this case - if your app requires fault address then you must set it and refuse working on non compatible machines.

___

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @kailun-qin)

___
*[`pal/src/host/linux-sgx/pal_exception.c` line 315 at r4](https://reviewable.io/reviews/gramineproject/gramine/1502#-NdewBEg2SLAzhC6TcED:-NdewBEg2SLAzhC6TcEE:b1v8r0k) ([raw file](https://github.com/gramineproject/gramine/blob/3a218e4598f45c54a6a0277ba1466ab1f19eb4da/pal/src/host/linux-sgx/pal_exception.c#L315)):*
> ```C
>         }
>         if (ei.info.vector == SGX_EXCEPTION_VECTOR_PF) {
>             ctx.cr2 = exinfo->maddr;
> ```

Wouldn't it be easier to set `bool has_hw_fault_address = true` here and just check that below?
___
*[`pal/src/host/linux-sgx/pal_exception.c` line 327 at r4](https://reviewable.io/reviews/gramineproject/gramine/1502#-NdewbsA50J7G4nX1-C-:-NdewbsA50J7G4nX1-C0:b-lnzxtf) ([raw file](https://github.com/gramineproject/gramine/blob/3a218e4598f45c54a6a0277ba1466ab1f19eb4da/pal/src/host/linux-sgx/pal_exception.c#L327)):*
> ```C
>             if (FIRST_TIME()) {
>                 bool require_exinfo;
>                 int ret = toml_bool_in(g_pal_public_state.manifest_root, "sgx.require_exinfo",
> ```

Parsing toml in exception handling doesn't sound right

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1502#-Ndej1oT0R3eYTdc_4dL:-NdfOwU43PcgV2sE9i6T:b-tp295z) (no related file):*
> But IIRC `require_exinfo` was exactly for this case - if your app requires fault address then you must set it and refuse working on non compatible machines.

Yes, but how do you know what the app really requires? Like, how do you know if the app is satisfied with `0` as the faulting address, or the app actually depends on knowing the exact faulting address?

To me, this is almost impossible to answer for sure. It's not about what the Intel SGX technology provides, but about the sane and safe default behavior of Gramine.

___
___
*[`pal/src/host/linux-sgx/pal_exception.c` line 315 at r4](https://reviewable.io/reviews/gramineproject/gramine/1502#-NdewBEg2SLAzhC6TcED:-NdfPdms1OC4pXI-2MBK:b-avu5bt) ([raw file](https://github.com/gramineproject/gramine/blob/3a218e4598f45c54a6a0277ba1466ab1f19eb4da/pal/src/host/linux-sgx/pal_exception.c#L315)):*

Wouldn't it be easier to set `bool has_hw_fault_address = true` here and just check that below?
</blockquote></details>

+1, I would be fine with the current version, but for future having a special variable seems safer and cleaner.
___
*[`pal/src/host/linux-sgx/pal_exception.c` line 327 at r4](https://reviewable.io/reviews/gramineproject/gramine/1502#-NdewbsA50J7G4nX1-C-:-NdfPmgc-CEXywPDTfQd:b-qf5wmo) ([raw file](https://github.com/gramineproject/gramine/blob/3a218e4598f45c54a6a0277ba1466ab1f19eb4da/pal/src/host/linux-sgx/pal_exception.c#L327)):*

Parsing toml in exception handling doesn't sound right
</blockquote></details>

Well, this is not really *parsing*. This is searching for a specific key in an already-parsed in-memory structure. But I agree, the parsed manifest may be pretty huge.

So let's put this parsing somewhere in `pal_main.c`, where we parse all other things, and have a global variable `g_require_exinfo`.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1502#-Ndej1oT0R3eYTdc_4dL:-NdfeJH4DAAmk_KQHRNx:bpcpp0k) (no related file):*
> Yes, but how do you know what the app really requires? 

If you put apps into Gramine you'd better know how it works.

___
___
*[`pal/src/host/linux-sgx/pal_exception.c` line 327 at r4](https://reviewable.io/reviews/gramineproject/gramine/1502#-NdewbsA50J7G4nX1-C-:-Ndff6YJBHOLw4MgRYUc:b-ml2vr2) ([raw file](https://github.com/gramineproject/gramine/blob/3a218e4598f45c54a6a0277ba1466ab1f19eb4da/pal/src/host/linux-sgx/pal_exception.c#L327)):*

Well, this is not really *parsing*. This is searching for a specific key in an already-parsed in-memory structure. But I agree, the parsed manifest may be pretty huge.

So let's put this parsing somewhere in `pal_main.c`, where we parse all other things, and have a global variable `g_require_exinfo`.
</blockquote></details>

Or let's just use a local variable as mentioned above and not bother with toml.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @kailun-qin)

___
*[`pal/src/host/linux-sgx/pal_exception.c` line 327 at r4](https://reviewable.io/reviews/gramineproject/gramine/1502#-NdewbsA50J7G4nX1-C-:-NdiXdPV-O2fxOAXFLJM:bcvlciu) ([raw file](https://github.com/gramineproject/gramine/blob/3a218e4598f45c54a6a0277ba1466ab1f19eb4da/pal/src/host/linux-sgx/pal_exception.c#L327)):*

Or let's just use a local variable as mentioned above and not bother with toml.
</blockquote></details>

I guess what you're saying is that if we ever get to that line above (`ctx.cr2 = exinfo->maddr`), then we are guaranteed to have the EXINFO SGX feature, and there is no need to additionally check the manifest's `sgx.require_exinfo` value. Sounds good to me. Yes, we can remove searching through the manifest completely then.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1502#-NdpAd96AGNI9nrl9gmF:-NdpAd96AGNI9nrl9gmG:bexhsto) (no related file):*
From the manifest syntax docs:

> If ``require_exinfo`` is set, user application can retrieve faulting address in
> signal handler in case of a page fault. Otherwise (set to ``false``),
> the faulting address will always be provided as ``0``.

Slightly unrelated, but it seems that this slipped through the older review and the documentation for this flag is wrong - setting it to `false` doesn't force the address to be zero, it's only that setting it to `true` blocks the enclave from running on CPUs which don't support this feature.
And this PR also seems to misinterpret this flag - "However, LibOS can't/shouldn't handle faulting addresses at all if `sgx.require_exinfo = false`" - no, it shouldn't handle it if exinfo is not supported by CPU / not enabled by enclave config, not if it's not required by the enclave to run.

I maaay be wrong regarding the semantics of this option, but I guess then its name is wrong? The other `sgx.require_*` options are only for refusing to start on CPUs without the corresponding feature.

___
___
*[`pal/src/host/linux-sgx/pal_exception.c` line 327 at r4](https://reviewable.io/reviews/gramineproject/gramine/1502#-NdewbsA50J7G4nX1-C-:-NdpAHr6DjdLkxAXHeGJ:b2kf7rz) ([raw file](https://github.com/gramineproject/gramine/blob/3a218e4598f45c54a6a0277ba1466ab1f19eb4da/pal/src/host/linux-sgx/pal_exception.c#L327)):*

I guess what you're saying is that if we ever get to that line above (`ctx.cr2 = exinfo->maddr`), then we are guaranteed to have the EXINFO SGX feature, and there is no need to additionally check the manifest's `sgx.require_exinfo` value. Sounds good to me. Yes, we can remove searching through the manifest completely then.
</blockquote></details>

+1 to Borys.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @kailun-qin, and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1502#-Ndej1oT0R3eYTdc_4dL:-Ndv8iTO8AXvCMz5udFC:b-33m0j1) (no related file):*

> Yes, but how do you know what the app really requires? 

If you put apps into Gramine you'd better know how it works.
</blockquote></details>

@mkow @kailun-qin WDYT?

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1502#-NdpAd96AGNI9nrl9gmF:-Ndv2_Lm0QUoZ0wJHrh7:b-rbqinm) (no related file):*

From the manifest syntax docs:

> If ``require_exinfo`` is set, user application can retrieve faulting address in
> signal handler in case of a page fault. Otherwise (set to ``false``),
> the faulting address will always be provided as ``0``.

Slightly unrelated, but it seems that this slipped through the older review and the documentation for this flag is wrong - setting it to `false` doesn't force the address to be zero, it's only that setting it to `true` blocks the enclave from running on CPUs which don't support this feature.
And this PR also seems to misinterpret this flag - "However, LibOS can't/shouldn't handle faulting addresses at all if `sgx.require_exinfo = false`" - no, it shouldn't handle it if exinfo is not supported by CPU / not enabled by enclave config, not if it's not required by the enclave to run.

I maaay be wrong regarding the semantics of this option, but I guess then its name is wrong? The other `sgx.require_*` options are only for refusing to start on CPUs without the corresponding feature.
</blockquote></details>

Ok, so to resolve all confusions (including my own), I dived into the Intel SGX architecture and read Intel SDM carefully.

Here is the situation:
1. Enclave writer may set `SIGSTRUCT.MISCSELECT[0] = 1`, to "report information about page fault [...] that occurred inside an enclave" (bit 0 is also referred to as `EXINFO`). On our Gramine manifest level, this is `sgx.require_exinfo = true`.
2. Upon enclave startup, the enclave copies this value into SECS, i.e. `SECS.MISCSELECT.EXINFO = SIGSTRUCT.MISCSELECT.EXINFO`. At this enclave-startup point, if the hardware does **not** support EXINFO (can be detected via CPUID.EAX=12H,ECX=0.EBX[0] bit), then the enclave terminates with error.
3. Upon every #PF, an asynchronous enclave exit (AEX) is generated, and the SSA frame will contain `SSA.GPRSGX.EXITINFO.VECTOR = #PF` plus `SSA.MISC.EXINFO.MADDR` (the page fault linear address that caused the fault).

Note that Intel SDM is explicit about #PF events: "Page fault exception. Only reported if `SECS.MISCSELECT.EXINFO = 1`".

**Conclusion 1:** Indeed, when in the code we end up after line `ei.info.vector == SGX_EXCEPTION_VECTOR_PF`, we are guaranteed that EXINFO = 1, i.e. it is both supported by the machine and `sgx.require_exinfo = true` was set in the manifest.

Now what happens when we specify `sgx.require_exinfo = false` in the manifest?
1. Then `SIGSTRUCT.MISCSELECT[0] = 0` aka `SIGSTRUCT.MISCSELECT.EXINFO = 0`.
2. Upon enclave startup, the enclave copies this value into SECS: `SECS.MISCSELECT.EXINFO = 0`. It is irrelevant whether the hardware supports EXINFO or not.
3. Upon every #PF, an asynchronous enclave exit (AEX) is generated, and the SSA frame will contain `SSA.GPRSGX.EXITINFO.VALID = 0`. Also, there is **no** `SSA.MISC` region at all. (By the way, our exception-handling code in this case falls back to using the untrusted "exception vector number" from the host.)
4. When it sees `SSA.GPRSGX.EXITINFO.VALID = 0`, our code is like "well, I cannot extract any additional info anyway, I'll just create an all-zeros context" and passes it further to LibOS. So in the end, the LibOS gets a memory fault but with `cr2 = 0`.

**Conclusion 2**: `sgx.require_exinfo` leads to two states:
- if `sgx.require_exinfo = 0`, then LibOS gets an (untrusted) memory fault with `cr2 = 0` (faulting address is zero).
- if `sgx.require_exinfo = 1`, then LibOS gets a trusted memory fault (a proper #PF) with `cr2 = faulting address`. Also, the enclave will start only on a machine that supports EXINFO feature.

So, the description provided in https://gramine.readthedocs.io/en/stable/manifest-syntax.html#optional-cpu-features-avx-avx512-mpx-pkru-amx-exinfo seems exactly correct to me.

___
___
*[`pal/src/host/linux-sgx/pal_exception.c` line 327 at r4](https://reviewable.io/reviews/gramineproject/gramine/1502#-NdewbsA50J7G4nX1-C-:-Ndv8U36CTITV8GHf9D5:b-1bpzcx) ([raw file](https://github.com/gramineproject/gramine/blob/3a218e4598f45c54a6a0277ba1466ab1f19eb4da/pal/src/host/linux-sgx/pal_exception.c#L327)):*

+1 to Borys.
</blockquote></details>

+1 again, see my top-level long comment. I verified by reading the Intel SDM that this is indeed the case.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, @kailun-qin, and @mwkmwkmwk)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1502#-Ndej1oT0R3eYTdc_4dL:-NdyoCBX0KZzwslVr6Lf:b526h4x) (no related file):*

@mkow @kailun-qin WDYT?
</blockquote></details>

I talked with @mwkmwkmwk and she had some (IMO) good arguments that this actually doesn't matter, because the app can't have any reasonable use for memory fault exception notification without the exception address included:
- If the app has an exception handler for diagnostics / error reporting, then killing it instead or not doesn't make much difference. But also telling it that it's `NULL` doesn't break anything.
- If the app does actually handle it, then it's usually for lazy loading of memory. Which won't work without the address.
- Also, without SGX2 and EDMM the app can't really do remove-permissions-and-then-trap on memory, because you can't change memory permissions. And AFAIK CPUs without exinfo don't have EDMM.

I think I'd prefer to kill the app just to be sure that it doesn't use wrong data.

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1502#-NdpAd96AGNI9nrl9gmF:-NdymqmP111LbUuT3zav:b-v2i37a) (no related file):*

Ok, so to resolve all confusions (including my own), I dived into the Intel SGX architecture and read Intel SDM carefully.

Here is the situation:
1. Enclave writer may set `SIGSTRUCT.MISCSELECT[0] = 1`, to "report information about page fault [...] that occurred inside an enclave" (bit 0 is also referred to as `EXINFO`). On our Gramine manifest level, this is `sgx.require_exinfo = true`.
2. Upon enclave startup, the enclave copies this value into SECS, i.e. `SECS.MISCSELECT.EXINFO = SIGSTRUCT.MISCSELECT.EXINFO`. At this enclave-startup point, if the hardware does **not** support EXINFO (can be detected via CPUID.EAX=12H,ECX=0.EBX[0] bit), then the enclave terminates with error.
3. Upon every #PF, an asynchronous enclave exit (AEX) is generated, and the SSA frame will contain `SSA.GPRSGX.EXITINFO.VECTOR = #PF` plus `SSA.MISC.EXINFO.MADDR` (the page fault linear address that caused the fault).

Note that Intel SDM is explicit about #PF events: "Page fault exception. Only reported if `SECS.MISCSELECT.EXINFO = 1`".

**Conclusion 1:** Indeed, when in the code we end up after line `ei.info.vector == SGX_EXCEPTION_VECTOR_PF`, we are guaranteed that EXINFO = 1, i.e. it is both supported by the machine and `sgx.require_exinfo = true` was set in the manifest.

Now what happens when we specify `sgx.require_exinfo = false` in the manifest?
1. Then `SIGSTRUCT.MISCSELECT[0] = 0` aka `SIGSTRUCT.MISCSELECT.EXINFO = 0`.
2. Upon enclave startup, the enclave copies this value into SECS: `SECS.MISCSELECT.EXINFO = 0`. It is irrelevant whether the hardware supports EXINFO or not.
3. Upon every #PF, an asynchronous enclave exit (AEX) is generated, and the SSA frame will contain `SSA.GPRSGX.EXITINFO.VALID = 0`. Also, there is **no** `SSA.MISC` region at all. (By the way, our exception-handling code in this case falls back to using the untrusted "exception vector number" from the host.)
4. When it sees `SSA.GPRSGX.EXITINFO.VALID = 0`, our code is like "well, I cannot extract any additional info anyway, I'll just create an all-zeros context" and passes it further to LibOS. So in the end, the LibOS gets a memory fault but with `cr2 = 0`.

**Conclusion 2**: `sgx.require_exinfo` leads to two states:
- if `sgx.require_exinfo = 0`, then LibOS gets an (untrusted) memory fault with `cr2 = 0` (faulting address is zero).
- if `sgx.require_exinfo = 1`, then LibOS gets a trusted memory fault (a proper #PF) with `cr2 = faulting address`. Also, the enclave will start only on a machine that supports EXINFO feature.

So, the description provided in https://gramine.readthedocs.io/en/stable/manifest-syntax.html#optional-cpu-features-avx-avx512-mpx-pkru-amx-exinfo seems exactly correct to me.
</blockquote></details>

Thanks for checking!
So, it seems that this manifest option is implemented differently than all other `sgx.require_*` manifest switches? If so, I'd rename it to `sgx.use_exinfo` (the others change `MISCSELECT_MASK`, and for this option if I understand correctly the mask bit is always set and we change only `MISCSELECT`?).
Also, it seems that it's hard to find a good default for it? It's `false`, but that seems like a bad choice, we should use it when available. Maybe default to the same value as `sgx.edmm_enable`? Does EDMM imply support for exinfo?

___

mwkmwkmwk:

I talked with @mwkmwkmwk and she had some (IMO) good arguments that this actually doesn't matter, because the app can't have any reasonable use for memory fault exception notification without the exception address included:
- If the app has an exception handler for diagnostics / error reporting, then killing it instead or not doesn't make much difference. But also telling it that it's `NULL` doesn't break anything.
- If the app does actually handle it, then it's usually for lazy loading of memory. Which won't work without the address.
- Also, without SGX2 and EDMM the app can't really do remove-permissions-and-then-trap on memory, because you can't change memory permissions. And AFAIK CPUs without exinfo don't have EDMM.

I think I'd prefer to kill the app just to be sure that it doesn't use wrong data.
</blockquote></details>

After thinking about this a bit more, I'm coming to the conclusion that the default behavior should be to fail on any page fault, for security reasons — use cases requiring page fault handling in normal operation are rare, and should be explicitly marked.  The other common usage of SIGSEGV handlers is backtraces and other diagnostics, which we'd want to discourage for security.

In general, I think the reasonable possibility space contains 4 distinct behaviors:

1. Any page fault destroys the enclave
2. Always deliver SIGSEGV, use address 0 if not available
3. On fault, deliver SIGSEGV if address is available, otherwise destroy enclave
4. Like 3, but prevent starting the enclave in the first place if EXINFO is not available (the current require_exinfo behavior)

An application that has no use for page fault handling would use behavior 1, which should be the default. The other classes of applications are, as far as I can tell:

1. JVM etc. that use SIGSEGV handler specifically for null pointer handling. Behavior 2 would be fine for them.
2. Applications that use SIGSEGV handling to do VM tricks. Behavior 4 would be required for them.

I don't think behavior 3 is particularly useful at all.

This patch, as it stands, implements behaviors 1 and 4 as outlined above, which seems quite reasonable. IMO the only question left to ask here is if we'd like to additionally implement behavior 2 (via a three-value option in the manifest, perhaps?) in case someone wants to run a JVM or something on pre-SGX2 hardware?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @kailun-qin, and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1502#-Ndej1oT0R3eYTdc_4dL:-Ne26DNfDiSbyonzmtls:b3tusol) (no related file):*
> IMO the only question left to ask here is if we'd like to additionally implement behavior 2 (via a three-value option in the manifest, perhaps?) in case someone wants to run a JVM or something on pre-SGX2 hardware?

I wouldn't want to implement complex manifest options... I vote against implementing behavior 2.

Also, I'd like to stress again -- I think we should pay little attention to old CPUs (that don't have EXINFO). The only reason to allow behavior 2 is to accommodate CPUs from <2017, and I don't think it's reasonable.

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1502#-NdpAd96AGNI9nrl9gmF:-Ne26ech6Qp6sKNODlnJ:b-b5z512) (no related file):*
> So, it seems that this manifest option is implemented differently than all other `sgx.require_*` manifest switches? If so, I'd rename it to `sgx.use_exinfo` (the others change `MISCSELECT_MASK`, and for this option if I understand correctly the mask bit is always set and we change only `MISCSELECT`?).

We discussed it already a bit offline. I would like to point out this long-standing PR, which deals with similar issues and has a lot of context: https://github.com/gramineproject/gramine/pull/881. Since we're discussing all these problems, I suggest we review that other PR too.

But TLDR: yes, Michal is right about not modifying the MASK field.

> Also, it seems that it's hard to find a good default for it? It's `false`, but that seems like a bad choice, we should use it when available. Maybe default to the same value as `sgx.edmm_enable`? Does EDMM imply support for exinfo?

I vote against "the same value as `sgx.edmm_enable`". Two reasons:
1. EDMM and EXINFO may be independent of each other. E.g., enclave developer may want to disable EDMM (for performance reasons) but to enable EXINFO (because it requires #PF handling).
2. EDMM and EXINFO seem to go hand in hand on real CPUs. However, they are enumerated via different CPUID bits, and I see no reason why some CPUs wouldn't implement one without the other.

___

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1502#-Ndej1oT0R3eYTdc_4dL:-Ne2ZH_U79xM1GWK2ggw:b-fpwf6v) (no related file):*

> IMO the only question left to ask here is if we'd like to additionally implement behavior 2 (via a three-value option in the manifest, perhaps?) in case someone wants to run a JVM or something on pre-SGX2 hardware?

I wouldn't want to implement complex manifest options... I vote against implementing behavior 2.

Also, I'd like to stress again -- I think we should pay little attention to old CPUs (that don't have EXINFO). The only reason to allow behavior 2 is to accommodate CPUs from <2017, and I don't think it's reasonable.
</blockquote></details>

I agree w/ all points from @mwkmwkmwk (thanks for the summary!) and I also vote against implementing behavior 2 (same reasons as explained by @dimakuv).

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1502#-NdpAd96AGNI9nrl9gmF:-Ne2ZyCs1bw1sMlZIV9B:b-vw423p) (no related file):*
> I vote against "the same value as `sgx.edmm_enable`"

+1 for the same reasons

___
___
*[`pal/src/host/linux-sgx/pal_exception.c` line 315 at r4](https://reviewable.io/reviews/gramineproject/gramine/1502#-NdewBEg2SLAzhC6TcED:-Ne2CxQz0tCNLixi-5ih:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/3a218e4598f45c54a6a0277ba1466ab1f19eb4da/pal/src/host/linux-sgx/pal_exception.c#L315)):*

+1, I would be fine with the current version, but for future having a special variable seems safer and cleaner.
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/pal_exception.c` line 327 at r4](https://reviewable.io/reviews/gramineproject/gramine/1502#-NdewbsA50J7G4nX1-C-:-Ne2MOVl8a9g1eax4aIr:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/3a218e4598f45c54a6a0277ba1466ab1f19eb4da/pal/src/host/linux-sgx/pal_exception.c#L327)):*

+1 again, see my top-level long comment. I verified by reading the Intel SDM that this is indeed the case.
</blockquote></details>

Done.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 4 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "squash! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

mkow (changes requested):
> Reviewed 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 4 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "squash! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1502#-Ndej1oT0R3eYTdc_4dL:-NeAqNlBA_ZPyyzmB8_u:b-kshj73) (no related file):*

I agree w/ all points from @mwkmwkmwk (thanks for the summary!) and I also vote against implementing behavior 2 (same reasons as explained by @dimakuv).
</blockquote></details>

I'm fine with 1+4, as long as 1 is the default and we rename the manifest option.

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1502#-NdpAd96AGNI9nrl9gmF:-NeAqZz3Cr_JiH8CVO1j:b-guxqtr) (no related file):*
> EDMM and EXINFO may be independent of each other. E.g., enclave developer may want to disable EDMM (for performance reasons) but to enable EXINFO (because it requires #PF handling).

Yes, it would still be possible, I'm talking only about the default value. Not about removing this option.

But I don't care that much, and also the backtrace security argument convinced me that it's better to disable this by default. I'll take a look at #881.

___

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 4 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "squash! " found in commit messages' one-liners (waiting on @boryspoplawski, @kailun-qin, and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1502#-Ndej1oT0R3eYTdc_4dL:-NeChJTNB6o2K0nTcbpi:b-ixau8m) (no related file):*
> ... and we rename the manifest option.

I think the rename of the manifest option should be done as part of https://github.com/gramineproject/gramine/pull/881 (or as a follow up PR to that one). Not in this PR.

___

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 4 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "squash! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1502#-Ndej1oT0R3eYTdc_4dL:-NeCitT11EGTSf51ZV48:b-rmuolu) (no related file):*
> I think the rename of the manifest option should be done as part of https://github.com/gramineproject/gramine/pull/881 (or as a follow up PR to that one). Not in this PR.

+1

___

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 3 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "squash! " found in commit messages' one-liners (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1502#-Ndej1oT0R3eYTdc_4dL:-NeG87tIDDvASrLmtEA2:b-eh3xgy) (no related file):*
> I think the rename of the manifest option should be done as part of #881 (or as a follow up PR to that one). Not in this PR.

That's fine, but it doesn't include that currently :) Could you add it there?

___

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 3 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "squash! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1502#-Ndej1oT0R3eYTdc_4dL:-NeHtXsABumRgih0oeUJ:b-k5ajb6) (no related file):*

> I think the rename of the manifest option should be done as part of #881 (or as a follow up PR to that one). Not in this PR.

That's fine, but it doesn't include that currently :) Could you add it there?
</blockquote></details>

Done, see last commit in #881 

___

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 3 unresolved discussions, "squash! " found in commit messages' one-liners (waiting on @boryspoplawski)

boryspoplawski (approved):
> Reviewed 3 of 3 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 3 unresolved discussions, "squash! " found in commit messages' one-liners (waiting on @kailun-qin)

___
*[`-- commits` line 30 at r5](https://reviewable.io/reviews/gramineproject/gramine/1502#-NeYFTDc0DDkr9nTdZrf):*
I think it should be either:
- `with no faulting address reported`
- `if no faulting address was reported`

but not blocking



_Code quote:_
```
[PAL/Linux-SGX] Disallow memfault handling if no faulting address reported
```
___
*[`pal/regression/Exception.c` line 1 at r5](https://reviewable.io/reviews/gramineproject/gramine/1502#-NeYG5uO6iqK8eNT4zcO:-NeYG5uO6iqK8eNT4zcP:b-iljxr2) ([raw file](https://github.com/gramineproject/gramine/blob/84ef1976ba198eabc9a872f5803918f08228f8eb/pal/regression/Exception.c#L1)):*
> ```C
> /* TODO: Hans, get ze flammenwerfer */
> ```

FWIW: I did not review these changes. This comment should really be applied.
___
*[`pal/regression/test_pal.py` line 216 at r5](https://reviewable.io/reviews/gramineproject/gramine/1502#-NeYH_yb839VeMmvq5T4:-NeYH_yb839VeMmvq5T5:bsyklbb) ([raw file](https://github.com/gramineproject/gramine/blob/84ef1976ba198eabc9a872f5803918f08228f8eb/pal/regression/test_pal.py#L216)):*
> ```Python
>             else:
>                 self.fail('expected to return zero')
>         finally:
> ```

Why these checks are in `finally` block, not outside of `try` completely?

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: 2 of 4 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "squash! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

___
*[`-- commits` line 30 at r5](https://reviewable.io/reviews/gramineproject/gramine/1502#-NeYFTDc0DDkr9nTdZrf):*

I think it should be either:
- `with no faulting address reported`
- `if no faulting address was reported`

but not blocking
</blockquote></details>

Done.
___
*[`pal/regression/test_pal.py` line 216 at r5](https://reviewable.io/reviews/gramineproject/gramine/1502#-NeYH_yb839VeMmvq5T4:-Nea24zT7ZyyvP1tZPEw:b-bur6bd) ([raw file](https://github.com/gramineproject/gramine/blob/84ef1976ba198eabc9a872f5803918f08228f8eb/pal/regression/test_pal.py#L216)):*

Why these checks are in `finally` block, not outside of `try` completely?
</blockquote></details>

Yeah, looks more correct to be completely outside. Updated, thanks!

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "squash! " found in commit messages' one-liners (waiting on @boryspoplawski and @kailun-qin)

___
*[`-- commits` line 50 at r6](https://reviewable.io/reviews/gramineproject/gramine/1502#-Neb4uyy9k2Ba1439FL2):*
`propagated`, can be fixed during final rebase

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 2 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "squash! " found in commit messages' one-liners (waiting on @boryspoplawski and @kailun-qin)

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 2 unresolved discussions, "squash! " found in commit messages' one-liners (waiting on @boryspoplawski and @kailun-qin)

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 2 unresolved discussions (waiting on @boryspoplawski, @dimakuv, and @mkow)

___
*[`-- commits` line 50 at r6](https://reviewable.io/reviews/gramineproject/gramine/1502#-Neb4uyy9k2Ba1439FL2):*

`propagated`, can be fixed during final rebase
</blockquote></details>

Done.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: all files reviewed, 2 unresolved discussions (waiting on @boryspoplawski and @dimakuv)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1502)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #1545, opened 2023-09-10
[LibOS,PAL] Fix wrong handling of sysfs NUMA nodes when nodes are offline

dimakuv:
## Description of the changes 

Previously, Gramine iterated through all NUMA nodes within the range obtained from the host's `/sys/devices/system/node/possible` file to get `cpulist` and `distance`. It also used the count of possible nodes to parse the `distance` between nodes. However, there is no `/sys/devices/system/node/nodeX` directory when a node is not online. Also, `/sys/devices/system/node/nodeX/distance` lists distances between online nodes only (jumping over offline nodes). This can lead to failures when Gramine tries to read values of `cpulist` and `distance` pseudo-files from the directory. Plus a correct parsing of `distance` should consider only online nodes.

**This is version 2 of #1454.** @mkow asked to keep an original definition of the `distances` array for NUMA nodes inside Gramine, and this PR reflects this. Since the change is rather drastic from the original #1454 code, I chose to create it as a separate PR.

Fixes #1452. For more context and discussions, see #1452 and #1454.

## How to test this PR? 

Manually on weird NUMA platforms...

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1545)

PR #1556, opened 2023-09-15
[PAL] Adjust get_topopology_info for more general case

stefanberger:
## Description of the changes 

The recent changes in get_topology_info still have bugs that make it fail to work on some of my
ppc64 machines where one of them for example has possible NUMA nodes '0,8' and online
NUMA nodes are also '0,8'.

The proposed changes implement a more general solution than what is there right now.

## How to test this PR? 

I can test this on my ppc64 machine but on other machines it depends on the NUMA node configuration and what is shown in sysfs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1556)

### DIFF
--- a/pal/src/host/linux-common/topo_info.c
+++ b/pal/src/host/linux-common/topo_info.c
@@ -220,6 +220,11 @@ static int get_ranges_end(size_t ind, void* _arg) {
     return 0;
 }
 
+static int get_ranges_count(size_t ind, void* _arg) {
+    *(size_t*)_arg += 1;
+    return 0;
+}
+
 static int set_thread_online(size_t ind, void* _threads) {
     struct pal_cpu_thread_info* threads = (struct pal_cpu_thread_info*)_threads;
     threads[ind].is_online = true;
@@ -289,7 +294,11 @@ int get_topology_info(struct pal_topo_info* topo_info) {
     if (ret < 0)
         return ret;
     size_t nodes_cnt = 0;
-    ret = iterate_ranges_from_file("/sys/devices/system/node/possible", get_ranges_end, &nodes_cnt);
+    ret = iterate_ranges_from_file("/sys/devices/system/node/online", get_ranges_count, &nodes_cnt);
+    if (ret < 0)
+        return ret;
+    size_t nodes_max = 0;
+    ret = iterate_ranges_from_file("/sys/devices/system/node/possible", get_ranges_end, &nodes_max);
     if (ret < 0)
         return ret;
 
@@ -300,8 +309,8 @@ int get_topology_info(struct pal_topo_info* topo_info) {
     struct pal_cpu_core_info* cores = malloc(threads_cnt * sizeof(*cores)); // overapproximate the count
     size_t sockets_cnt = 0;
     struct pal_socket_info* sockets = malloc(threads_cnt * sizeof(*sockets)); // overapproximate the count
-    struct pal_numa_node_info* numa_nodes = malloc(nodes_cnt * sizeof(*numa_nodes));
-    size_t* distances = malloc(nodes_cnt * nodes_cnt * sizeof(*distances));
+    struct pal_numa_node_info* numa_nodes = malloc(nodes_max * sizeof(*numa_nodes));
+    size_t* distances = malloc(nodes_max * nodes_max * sizeof(*distances));
     if (!threads || !caches || !cores || !sockets || !numa_nodes || !distances) {
         ret = -ENOMEM;
         goto fail;
@@ -316,7 +325,7 @@ int get_topology_info(struct pal_topo_info* topo_info) {
             threads[i].ids_of_caches[j] = (size_t)-1;
         }
     }
-    for (size_t i = 0; i < nodes_cnt; i++)
+    for (size_t i = 0; i < nodes_max; i++)
         numa_nodes[i].is_online = false;
 
     ret = iterate_ranges_from_file("/sys/devices/system/cpu/online", set_thread_online, threads);
@@ -400,7 +409,7 @@ int get_topology_info(struct pal_topo_info* topo_info) {
      *            0        ,    0    ,        0
      *     node 2 -> node 0,    0    , node 2 -> node 2 ]
      */
-    memset(distances, 0, nodes_cnt * nodes_cnt * sizeof(*distances));
+    memset(distances, 0, nodes_max * nodes_max * sizeof(*distances));
     for (size_t i = 0; i < nodes_cnt; i++) {
         if (!numa_nodes[i].is_online)
             continue;
--- a/pal/src/host/linux-common/topo_info.c
+++ b/pal/src/host/linux-common/topo_info.c
@@ -60,11 +60,22 @@ static int get_hw_resource_value(const char* filename, size_t* out_value) {
     return 0;
 }
 
+static ssize_t get_next_online_numa_node(struct pal_numa_node_info* numa_nodes, size_t nodes_max,
+                                         size_t sidx) {
+    while (!numa_nodes[sidx].is_online) {
+        sidx++;
+        if (sidx >= nodes_max)
+            return -1;
+    }
+    return sidx;
+}
+
 /* Read a space-separated list of numbers in the format used by
  * `/sys/devices/system/node/node<i>/distance`, and write the result to the online nodes from
  * `numa_nodes`. */
 static int read_distances_from_file(const char* path, size_t* out_arr,
-                                    struct pal_numa_node_info* numa_nodes, size_t nodes_cnt) {
+                                    struct pal_numa_node_info* numa_nodes, size_t nodes_max,
+                                    size_t nodes_cnt) {
     char buf[PAL_SYSFS_BUF_FILESZ];
     int ret = read_file_buffer(path, buf, sizeof(buf) - 1);
     if (ret < 0)
@@ -75,10 +86,14 @@ static int read_distances_from_file(const char* path, size_t* out_arr,
     const char* end;
     char last_separator = ' ';
     size_t node_i = 0;
-    for (size_t input_i = 0; /* no condition */; input_i++) {
+    ssize_t sidx = 0;
+    for (size_t input_i = 0; /* no condition */; input_i++, sidx++) {
         /* Find next online node (only these are listed in `distance` file). */
-        while (node_i < nodes_cnt && !numa_nodes[node_i].is_online)
-            node_i++;
+        if (node_i < nodes_cnt) {
+            sidx = get_next_online_numa_node(numa_nodes, nodes_max, sidx);
+            if (sidx < 0)
+                return -EINVAL;
+        }
         if (node_i == nodes_cnt)
             break;
         if (last_separator != ' ')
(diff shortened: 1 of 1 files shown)


mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1556)* status: 0 of 1 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @stefanberger)

___
*[`-- commits` line 3 at r2](https://reviewable.io/reviews/gramineproject/gramine/1556#-NePotnK6zDqA_m6Zy8g):*
ditto below



_Suggestion:_
```
[PAL]
```
___
*[`-- commits` line 7 at r2](https://reviewable.io/reviews/gramineproject/gramine/1556#-NePp2YbAr358E4g2uVc):*
Please don't, the name of this function says you're returning the maximum value, but in fact you're returning max+1...



_Code quote:_
```
Implement get_ranges_max() to get the maximum value + 1
```
___
*[`-- commits` line 30 at r2](https://reviewable.io/reviews/gramineproject/gramine/1556#-NeQ0G63-R1gqxnnfzq8):*
>   The current implementation doesn't work if the numa nodes that are online are non-consecutive

I'm pretty sure that's not true, #1545 was tested on a system with ON,OFF,ON NUMA nodes.
What and how exactly failed on your system?

stefanberger (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1556)* status: 0 of 1 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @mkow)

___
*[`-- commits` line 30 at r2](https://reviewable.io/reviews/gramineproject/gramine/1556#-NeQ0G63-R1gqxnnfzq8):*

>   The current implementation doesn't work if the numa nodes that are online are non-consecutive

I'm pretty sure that's not true, #1545 was tested on a system with ON,OFF,ON NUMA nodes.
What and how exactly failed on your system?
</blockquote></details>

On one of my systems it was failing because it was creating the wrong path names, like trying to get a numa node1 where there is no node1 but noe0 and node8:

```
$ ls -l /sys/devices/system/node/
total 0
-r--r--r--. 1 root root 65536 Sep 10 12:14 has_cpu
-r--r--r--. 1 root root 65536 Sep 10 12:14 has_generic_initiator
-r--r--r--. 1 root root 65536 Sep 10 12:14 has_memory
-r--r--r--. 1 root root 65536 Sep 10 12:14 has_normal_memory
drwxr-xr-x. 5 root root     0 Sep 10 11:54 node0
drwxr-xr-x. 5 root root     0 Sep 10 11:54 node8
-r--r--r--. 1 root root 65536 Sep 10 12:14 online
-r--r--r--. 1 root root 65536 Sep 10 12:14 possible
drwxr-xr-x. 2 root root     0 Sep 10 12:14 power
-rw-r--r--. 1 root root 65536 Sep 10 11:54 uevent
```

Besides this I think it was only allocating 2 numa node entries for '1,8' but then when it set them online it set '1' and '8' online, which was an array-out-of-bounds access.

stefanberger (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1556)* status: 0 of 1 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @mkow)

___
*[`-- commits` line 30 at r2](https://reviewable.io/reviews/gramineproject/gramine/1556#-NeQ0G63-R1gqxnnfzq8):*

On one of my systems it was failing because it was creating the wrong path names, like trying to get a numa node1 where there is no node1 but noe0 and node8:

```
$ ls -l /sys/devices/system/node/
total 0
-r--r--r--. 1 root root 65536 Sep 10 12:14 has_cpu
-r--r--r--. 1 root root 65536 Sep 10 12:14 has_generic_initiator
-r--r--r--. 1 root root 65536 Sep 10 12:14 has_memory
-r--r--r--. 1 root root 65536 Sep 10 12:14 has_normal_memory
drwxr-xr-x. 5 root root     0 Sep 10 11:54 node0
drwxr-xr-x. 5 root root     0 Sep 10 11:54 node8
-r--r--r--. 1 root root 65536 Sep 10 12:14 online
-r--r--r--. 1 root root 65536 Sep 10 12:14 possible
drwxr-xr-x. 2 root root     0 Sep 10 12:14 power
-rw-r--r--. 1 root root 65536 Sep 10 11:54 uevent
```

Besides this I think it was only allocating 2 numa node entries for '1,8' but then when it set them online it set '1' and '8' online, which was an array-out-of-bounds access.
</blockquote></details>

What I need in this case is nodes_cnt = 2 because there are two numa nodes but nodes_max needs to be '9' and the two numa nodes are non-consecutive, so not node0 and node1.

I take back the statement about out-of-bounds access.

stefanberger (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1556)* status: 0 of 1 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @mkow)

___
*[`-- commits` line 3 at r2](https://reviewable.io/reviews/gramineproject/gramine/1556#-NePotnK6zDqA_m6Zy8g):*

ditto below
</blockquote></details>

Done.
___
*[`-- commits` line 7 at r2](https://reviewable.io/reviews/gramineproject/gramine/1556#-NePp2YbAr358E4g2uVc):*

Please don't, the name of this function says you're returning the maximum value, but in fact you're returning max+1...
</blockquote></details>

Squashed the first two patches and rewrote the title.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1556)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @mkow)

___
*[`-- commits` line 30 at r2](https://reviewable.io/reviews/gramineproject/gramine/1556#-NeQ0G63-R1gqxnnfzq8):*
Something doesn't add up here.

In our current implementation, after this line https://github.com/gramineproject/gramine/blob/dc9206760a562f3df6c5dedb471e2d46c96ecf1e/pal/src/host/linux-common/topo_info.c#L292

the `nodes_cnt` is equal to 9 (in your example). Then, after this line: https://github.com/gramineproject/gramine/blob/dc9206760a562f3df6c5dedb471e2d46c96ecf1e/pal/src/host/linux-common/topo_info.c#L325

we have an array of NUMA nodes of size 9, and only NUMA node 0 and NUMA node 8 are marked as online.

Then we create a `distances` array which will be 9x9, and we'll populate only the cells that are on the rows/cols 0 and 8 (the other cells will be equal to zero).

So where is the problem? The current code looks correct.

> What I need in this case is `nodes_cnt = 2` because there are two numa nodes but `nodes_max` needs to be '9' and the two numa nodes are non-consecutive, so not node0 and node1.

Why do you need `nodes_cnt` to be 2? Why can't it be 9, and only two of these nine will be online (as the current code does)?

stefanberger (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1556)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @mkow)

___
*[`-- commits` line 30 at r2](https://reviewable.io/reviews/gramineproject/gramine/1556#-NeQ0G63-R1gqxnnfzq8):*

Something doesn't add up here.

In our current implementation, after this line https://github.com/gramineproject/gramine/blob/dc9206760a562f3df6c5dedb471e2d46c96ecf1e/pal/src/host/linux-common/topo_info.c#L292

the `nodes_cnt` is equal to 9 (in your example). Then, after this line: https://github.com/gramineproject/gramine/blob/dc9206760a562f3df6c5dedb471e2d46c96ecf1e/pal/src/host/linux-common/topo_info.c#L325

we have an array of NUMA nodes of size 9, and only NUMA node 0 and NUMA node 8 are marked as online.

Then we create a `distances` array which will be 9x9, and we'll populate only the cells that are on the rows/cols 0 and 8 (the other cells will be equal to zero).

So where is the problem? The current code looks correct.

> What I need in this case is `nodes_cnt = 2` because there are two numa nodes but `nodes_max` needs to be '9' and the two numa nodes are non-consecutive, so not node0 and node1.

Why do you need `nodes_cnt` to be 2? Why can't it be 9, and only two of these nine will be online (as the current code does)?
</blockquote></details>

Because the current code creates wrong path names and if they don't work it bails out.

```
        snprintf(path, sizeof(path), "/sys/devices/system/node/node%zu/cpulist", sidx);
        ret = iterate_ranges_from_file(path, set_node_id, &(struct set_node_id_args){
            .threads = threads,
            .cores = cores,
            .id_to_set = i,
        });
        if (ret < 0)
            goto fail;

```

If the above path becomes /sys/devices/system/node/node1/cpulist, which doesn't exist, then iterate_ranges_from_file() will return an error code. That's why the 2nd patch only tries to find the numa nodes that are online and there are '2'.

stefanberger (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1556)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @mkow)

___
*[`-- commits` line 30 at r2](https://reviewable.io/reviews/gramineproject/gramine/1556#-NeQ0G63-R1gqxnnfzq8):*

Because the current code creates wrong path names and if they don't work it bails out.

```
        snprintf(path, sizeof(path), "/sys/devices/system/node/node%zu/cpulist", sidx);
        ret = iterate_ranges_from_file(path, set_node_id, &(struct set_node_id_args){
            .threads = threads,
            .cores = cores,
            .id_to_set = i,
        });
        if (ret < 0)
            goto fail;

```

If the above path becomes /sys/devices/system/node/node1/cpulist, which doesn't exist, then iterate_ranges_from_file() will return an error code. That's why the 2nd patch only tries to find the numa nodes that are online and there are '2'.
</blockquote></details>

Also, we don't want to loop trying to find '9' online nodes but only '2'.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1556)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @mkow)

___
*[`-- commits` line 30 at r2](https://reviewable.io/reviews/gramineproject/gramine/1556#-NeQ0G63-R1gqxnnfzq8):*
> Because the current code creates wrong path names and if they don't work it bails out.

Why do you copy-paste your own code, but talk about "the current code"? This is confusing.

> If the above path becomes /sys/devices/system/node/node1/cpulist, which doesn't exist, then iterate\_ranges\_from\_file() will return an error code.

Why would the path be `.../node/node1/...`? This is our current code:
```c
    for (size_t i = 0; i < nodes_cnt; i++) {
        if (!numa_nodes[i].is_online)
            continue;

        snprintf(path, sizeof(path), "/sys/devices/system/node/node%zu/cpulist", i);
        ret = iterate_ranges_from_file(path, set_node_id, &(struct set_node_id_args){
```

So node 1 is known to be offline (because we collected only node 0 and node 8 as online from the `.../node/online` file, the rest are offline). Thus, we will `continue;` in the snippet above, and we'll never try to open the `.../node/node1/...` file.

What am I missing in our current code? Maybe pasting a GDB session over our current code will help?

---

> Also, we don't want to loop trying to find '9' online nodes but only '2'.

What part of our current code you're talking about? This is the relevant current code:
```c
    for (size_t i = 0; i < nodes_cnt; i++)
        numa_nodes[i].is_online = false;
    ...
    ret = iterate_ranges_from_file("/sys/devices/system/node/online", set_numa_node_online,
                                   numa_nodes);
```

So we mark all NUMA nodes as offline first (i.e. 0-8 in your case). And then we mark NUMA nodes as online if they are present in the `.../node/online` file. So in the end, we'll have node 0 and node 8 as online, the rest (1-7) as offline. Which looks like correct behavior to me.

stefanberger:
While rebasing on the newer Gramine code I made the mistake that I did not realize that the correct code was now in Gramine and it would just work, so I got 'creative' instead. I am sorry for the confusion.

PR #1067, opened 2022-11-30
[LibOS] Clean up `select` and `poll` implementations

boryspoplawski:
## Description of the changes 
Fixed some bugs, hopefully the code easier to manage too now.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1067)

PR #332, opened 2022-01-12
[Pal] Support opening TTY devices with O_TRUNC flag

jkr0103:
## Description of the changes 
Write to `/proc/self/fd/2` (which is symlink to /dev/tty) fails with `O_TRUNC` flag which translates to `DkStreamSetLength` as `setlength` function is not implemented for dev:tty devices.

This PR adds dummy function `setlength` to support O_TRUNC flag.

Fixes #303 

## How to test this PR? 

Should be able to execute below commands successfully:
```
gramine-direct busybox sh -c 'echo foo>/proc/self/fd/2'
gramine-sgx busybox sh -c 'echo foo>/proc/self/fd/2'
```
as described in link  https://github.com/gramineproject/gramine/issues/303#issuecomment-1002591818

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/332)

PR #1401, opened 2023-06-19
[LibOS] Ignore O_TRUNC flag for FIFO files

sahason:
## Description of the changes 

During enablement of mxnet-model-server in gramine one issue is observed that for FIFO files opening it with `O_TRUNC` returns `EINVAL`. In the native `O_TRUNC` flag for FIFO files is ignored. See `man 2 open`. This PR also emulates the same behaviour for gramine.

## How to test this PR? 

Modify this line https://github.com/gramineproject/gramine/blob/702ac42f81452b288b66b00308763584c3d115d3/libos/test/regression/mkfifo.c#L30  by adding `O_TRUNC` flag and replacing `O_RDONLY` by `O_RDWR` and test this PR by running regression test.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1401)

PR #1563, opened 2023-09-20
[LibOS,PAL] Refactor LibOS's `/dev/tty` and PAL's corresponding `dev:tty`

dimakuv:
## Description of the changes 

In Gramine, similar to UNIX, the stdin, stdout and stderr streams are file descriptors on the `/dev/tty` device (controlling terminal):
- `stdin = open("/dev/tty", O_RDONLY)`
- `stdout = stderr = open("/dev/tty", O_WRONLY | O_APPEND)`

On the LibOS side, the previous implementation was confusing: it mounted `/dev/tty` device file in a special way: not via Gramine's devfs, but instead via chrootfs.

On the PAL side, the `/dev/tty` file was confusingly mapped to PAL's `dev:tty`, which was implemented via `PAL_TYPE_DEV` but in a special way: all device APIs in `pal_devices.c` had a special path for `dev:tty` (which was also erroneously checked by comparing the FD to 0 or 1, even though these host FDs could be different in the child).

This commit refactors this mess:
- In LibOS, `/dev/tty` becomes a pseudo-file mounted via devfs.
- In PAL, `/dev/tty` maps to `console:`, which is a new type of PAL handle `PAL_TYPE_CONSOLE`. This new type is removed from `pal_devices.c` and instead implemented in a new `pal_console.c`.

Closes #1096.

Helpers:
- https://man7.org/linux/man-pages/man4/tty.4.html
- https://man7.org/linux/man-pages/man3/stdout.3.html
- https://github.com/gramineproject/graphene/issues/2553 (this was fixed by Borys in his [Poll Rework](https://github.com/gramineproject/gramine/pull/1067) but in a hacky way, I modified this to be less hacky in this PR)
- https://github.com/gramineproject/gramine/pull/332 (just for info)
- https://github.com/gramineproject/graphene/pull/2413 (just for info)
- https://github.com/gramineproject/graphene/pull/1951 (just for info, previous attempt at refactoring)
- Closing/redirecting standard streams:
  - https://stackoverflow.com/questions/4972994/how-to-close-stdout-and-stderr-in-c
  - https://stackoverflow.com/questions/33267559/linux-close-or-redirect-standard-output-to-dev-null-after-detaching-from-term
  - https://stackoverflow.com/questions/9084099/re-opening-stdout-and-stdin-file-descriptors-after-closing-them

## How to test this PR? 

All CI tests should work. Plus a new LibOS test `console` is added to test the tty/console.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1563)


llly (commented):
> Reviewed 32 of 32 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1563)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`libos/src/sys/libos_poll.c` line 140 at r1](https://reviewable.io/reviews/gramineproject/gramine/1563#-NepK33zA_x62N-cglY8:-NepK33zA_x62N-cglY9:b-5xlt9j) ([raw file](https://github.com/gramineproject/gramine/blob/d97eff5ec2eb85dd547ddfc50afd3e5176eaeeb7/libos/src/sys/libos_poll.c#L140)):*
> ```C
>             continue;
> 
>             use_pal_polling:;
> ```

Label at the end of code block looks wired. Is it intended in order to get rid of some warning?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1563)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @llly)

___
*[`libos/src/sys/libos_poll.c` line 140 at r1](https://reviewable.io/reviews/gramineproject/gramine/1563#-NepK33zA_x62N-cglY8:-NeqbDXc4N9zbnfR7cBo:b-rr3oq5) ([raw file](https://github.com/gramineproject/gramine/blob/d97eff5ec2eb85dd547ddfc50afd3e5176eaeeb7/libos/src/sys/libos_poll.c#L140)):*

Label at the end of code block looks wired. Is it intended in order to get rid of some warning?
</blockquote></details>

This is just to skip the middle part of this code block, nothing else.

The logic in this code is like this:
```
for (each handle) {
    if (handle has a poll callback) {
        handle->poll();
        if (poll failed) {
            jump out of this if code block; // this is the `goto` line
        } else {
            do some handle-specific stuff;
            jump to next handle; // this is the `continue` line
        }
    }
}
```

I didn't want to modify already-existing control flow with the goto and the label. So I just kept them as-is, but only changed the names.

If you think there's a more readable way to describe this logic, I can change it.

llly (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1563)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`libos/src/sys/libos_poll.c` line 140 at r1](https://reviewable.io/reviews/gramineproject/gramine/1563#-NepK33zA_x62N-cglY8:-Neua7fV0VoI3u4l8WAb:b-hiauy) ([raw file](https://github.com/gramineproject/gramine/blob/d97eff5ec2eb85dd547ddfc50afd3e5176eaeeb7/libos/src/sys/libos_poll.c#L140)):*

This is just to skip the middle part of this code block, nothing else.

The logic in this code is like this:
```
for (each handle) {
    if (handle has a poll callback) {
        handle->poll();
        if (poll failed) {
            jump out of this if code block; // this is the `goto` line
        } else {
            do some handle-specific stuff;
            jump to next handle; // this is the `continue` line
        }
    }
}
```

I didn't want to modify already-existing control flow with the goto and the label. So I just kept them as-is, but only changed the names.

If you think there's a more readable way to describe this logic, I can change it.
</blockquote></details>

        }
        use_pal_polling:;

When I jump to the label, I need roll up to check the following `}` is the end of `if` or `for` to determine next sentence.



_Code snippet:_
```C
        use_pal_polling:;
    }
```

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1563)* status: 31 of 32 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @llly)

___
*[`libos/src/sys/libos_poll.c` line 140 at r1](https://reviewable.io/reviews/gramineproject/gramine/1563#-NepK33zA_x62N-cglY8:-NevaEZL5-_NvLr6ZwgM:b-mlqq2r) ([raw file](https://github.com/gramineproject/gramine/blob/d97eff5ec2eb85dd547ddfc50afd3e5176eaeeb7/libos/src/sys/libos_poll.c#L140)):*

}
        use_pal_polling:;

When I jump to the label, I need roll up to check the following `}` is the end of `if` or `for` to determine next sentence.
</blockquote></details>

Done, what about now?

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1563)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`-- commits` line 4 at r2](https://reviewable.io/reviews/gramineproject/gramine/1563#-Nex3qLM4E-8gsLgK2mZ):*
Hmm, no? They are whatever the shell passes there, which can be also e.g. a pipe or a regular file.
___
*[`-- commits` line 7 at r2](https://reviewable.io/reviews/gramineproject/gramine/1563#-NexFgEx2X9r6HYhxXYf):*
On Unixes libc doesn't open `/dev/tty`, just inherits FDs 0,1,2 from the parent.

mkow (changes requested):
> Reviewed 12 of 32 files at r1.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1563)* status: all files reviewed, 11 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`pal/include/pal/pal.h` line 348 at r2](https://reviewable.io/reviews/gramineproject/gramine/1563#-NexYXUJE4AI-NhVDqK_:-NexYXUJE4AI-NhVDqKa:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/e27e63d4a5cd04b7a20e453ed159a17e7b449ac7/pal/include/pal/pal.h#L348)):*
> ```objc
>  *   #PAL_CREATE_ALWAYS is given in `create` flags, the file/directory will be created.
>  * * `dev:...`: Open a device as a stream.
>  * * `console:`: Open a console (stdin/stdout/stderr) as a stream.
> ```





_Suggestion:_
```objc
(host process stdin/stdout/stderr) as a stream
```
___
*[`pal/src/host/linux/pal_console.c` line 5 at r2](https://reviewable.io/reviews/gramineproject/gramine/1563#-Neymo-61HRxz9HIwcXI:-Neymo-61HRxz9HIwcXJ:b-2jejb6) ([raw file](https://github.com/gramineproject/gramine/blob/e27e63d4a5cd04b7a20e453ed159a17e7b449ac7/pal/src/host/linux/pal_console.c#L5)):*
> ```C
> 
> /*
>  * Operations to handle the console (stdin/stdout/stderr). Note that some operations (like stat and
> ```

I don't like this merging of these terms. std{in,out,err} may be unrelated to any console.



_Code quote:_
```C
the console (stdin/stdout/stderr)
```
___
*[`pal/src/host/linux/pal_console.c` line 38 at r2](https://reviewable.io/reviews/gramineproject/gramine/1563#-NeyqAkf5wGgHjS5-BRX:-NexJ1IJEjhw_CyjI4BI:b-w87lg) ([raw file](https://github.com/gramineproject/gramine/blob/e27e63d4a5cd04b7a20e453ed159a17e7b449ac7/pal/src/host/linux/pal_console.c#L38)):*
> ```C
>     hdl->hdr.type = PAL_TYPE_CONSOLE;
>     hdl->flags = access == PAL_ACCESS_RDONLY ? PAL_HANDLE_FD_READABLE : PAL_HANDLE_FD_WRITABLE;
>     hdl->console.fd = access == PAL_ACCESS_RDONLY ? /*host stdin*/0 : /*host stdout*/1;
> ```

We should really finally split these handle opening functions into specialized ones, we keep abusing this interface for opening regular files over and over...
___
*[`pal/src/host/linux/pal_devices.c` line 27 at r2](https://reviewable.io/reviews/gramineproject/gramine/1563#-NeysM4G3mfsNMef72y3:-NeysM4G3mfsNMef72y4:b3gyf9a) ([raw file](https://github.com/gramineproject/gramine/blob/9f80c4245a8a852a2d6ac30c81840c94db447305/pal/src/host/linux/pal_devices.c#L27)):*
> ```C
>     int ret;
> 
>     if (strcmp(type, URI_TYPE_DEV))
> ```

Maybe an assert then?
___
*[`pal/src/host/linux/pal_streams.c` line 57 at r2](https://reviewable.io/reviews/gramineproject/gramine/1563#-NeyvPGA-yD4tgex4dHO:-NeyvPGA-yD4tgex4dHP:b-6be3pj) ([raw file](https://github.com/gramineproject/gramine/blob/e27e63d4a5cd04b7a20e453ed159a17e7b449ac7/pal/src/host/linux/pal_streams.c#L57)):*
> ```C
>             break;
>         case PAL_TYPE_CONSOLE:
>             /* console (stdin/stdout/stderr) has no fields to serialize */
> ```

What about the state whether the handle was already closed? Shouldn't be writable afterwards, seems that now migration re-opens the handle?

mkow (changes requested):
I'm very confused by the naming and the explanations here (see my review comments), but in terms of what it does, I think this is a great change :)

> Reviewed 19 of 32 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1563#-:-Nf7nutxEFKBke_srzDW:bxupjsd)* status: all files reviewed, 12 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`libos/src/sys/libos_poll.c` line 123 at r1](https://reviewable.io/reviews/gramineproject/gramine/1563#-NelpClD6yC2Z2INnOGh:-Nf7pszUAnyNl8PZOeVL:bxwm3nn) ([raw file](https://github.com/gramineproject/gramine/blob/d97eff5ec2eb85dd547ddfc50afd3e5176eaeeb7/libos/src/sys/libos_poll.c#L123)):*

Yep, this is still ugly, but that's because we have two levels of indirection in our FS code...
</blockquote></details>

What's different in `console:` handles from others which also exist on the host (e.g. pipes or normal files) that they need this special treatment?
___
*[`libos/test/regression/console.c` line 31 at r2](https://reviewable.io/reviews/gramineproject/gramine/1563#-Nf7qv2O-VudKsD0M9i0:-Nf7qv2O-VudKsD0M9i1:b5b2pmp) ([raw file](https://github.com/gramineproject/gramine/blob/e27e63d4a5cd04b7a20e453ed159a17e7b449ac7/libos/test/regression/console.c#L31)):*
> ```C
> int main(void) {
>     /* initialization -- write some messages and save stdout/stderr (for further restore) */
>     ssize_t x = CHECK(write(STDOUT_FILENO, FIRST_HELLO_STDOUT, sizeof(FIRST_HELLO_STDOUT)));
> ```

Not `strlen()`? We rather don't want to print the null byte to the terminal.
ditto below



_Code quote:_
```C
sizeof(FIRST_HELLO_STDOUT)
```
___
*[`pal/src/host/linux/pal_devices.c` line 27 at r2](https://reviewable.io/reviews/gramineproject/gramine/1563#-NeysM4G3mfsNMef72y3:-Nf7kEr900wlN-6liLBn:bv0oanj) ([raw file](https://github.com/gramineproject/gramine/blob/9f80c4245a8a852a2d6ac30c81840c94db447305/pal/src/host/linux/pal_devices.c#L27)):*

Maybe an assert then?
</blockquote></details>

(ditto for all these comments to SGX PAL)
___
*[`pal/src/host/linux-sgx/pal_devices.c` line 201 at r2](https://reviewable.io/reviews/gramineproject/gramine/1563#-Nf7lN_H1htWaPZJr8PU:-Nf7lN_H1htWaPZJr8PV:b3b5k90) ([raw file](https://github.com/gramineproject/gramine/blob/9f80c4245a8a852a2d6ac30c81840c94db447305/pal/src/host/linux-sgx/pal_devices.c#L201)):*
> ```C
> }
> 
> /* this dummy function is implemented to support opening TTY devices with O_TRUNC flag */
> ```

I already managed to forget how bad this whole thing was...

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1563)* status: all files reviewed, 11 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`-- commits` line 4 at r2](https://reviewable.io/reviews/gramineproject/gramine/1563#-Nex3qLM4E-8gsLgK2mZ):*

Hmm, no? They are whatever the shell passes there, which can be also e.g. a pipe or a regular file.
</blockquote></details>

Let me expand on my reasoning.

From the point of view of the applications running inside Gramine:
- Gramine environment can be considered a VM.
- Gramine itself can be considered a Linux kernel.
- Gramine's first process (aka main process aka master process) performs initialization steps that can be considered similar to what `init` (and its helper utilities) do in Linux.

From this perspective, Gramine initialization must start with finding devices (like the teletypewriter aka TTY), building abstractions on top of these devices, and passing control to the user programs.

What does `init` do with respect to initializing stdin/stdout/stderr (aka FDs 0, 1, 2), which then will be inherited by the `init` child processes? The following:
1. Find all TTY devices and create `/dev/ttyXXX` pseudo-files corresponding to these devices
2. Call `getty` utility
3. The `getty` utility opens each `/dev/ttyXXX` pseudo-file and connects its read end to stdin (FD 0), and its write ends to stdout + stderr (FDs 1 and 2).
4. This abstraction/association will be used by all children of `init` (I'm fuzzy on how `getty` sends back to `init` this association).

So Gramine's initialization logic must do a similar thing. In Gramine, we don't have special `init` and `getty` processes, so Gramine's LibOS code does this itself. Also, we don't have multiple TTYs, and we don't add the complexity of pseudo TTYs (PTYs), so we just set up a single `/dev/tty` device and associate FDs 0,1,2 with it.

A bunch of references:
- https://stackoverflow.com/questions/63394927/what-does-the-tty-do-when-a-process-asks-for-input-or-outputs-to-the-screen
- https://unix.stackexchange.com/questions/4126/what-is-the-exact-difference-between-a-terminal-a-shell-a-tty-and-a-con
- https://www.linusakesson.net/programming/tty/ (a great article, recommend)
- https://tldp.org/LDP/intro-linux/html/sect_04_02.html
- https://tldp.org/HOWTO/Text-Terminal-HOWTO-15.html

I hope I didn't screw up too much in my descriptions :)

P.S. If I understand correctly, the above description works both all UNIX-based operating systems. (That's why I say `UNIX` in the commit message instead of `Linux`.) I may be wrong here, but I read the same docs about OpenBSD, so I assume that it's the same for all popular UNIX-based OSes.
___
*[`-- commits` line 7 at r2](https://reviewable.io/reviews/gramineproject/gramine/1563#-NexFgEx2X9r6HYhxXYf):*

On Unixes libc doesn't open `/dev/tty`, just inherits FDs 0,1,2 from the parent.
</blockquote></details>

But Gramine is not libc? Gramine's initialization code is akin to the `init` process in Linux. Gramine **prepares** FDs 0,1,2 for the use by the app's first process (which contains libc). Gramine **also prepares** the `/dev/tty` pseudo-file, because some applications may want to use it directly.
___
*[`libos/src/sys/libos_poll.c` line 123 at r1](https://reviewable.io/reviews/gramineproject/gramine/1563#-NelpClD6yC2Z2INnOGh:-NfAm0Q673vnT8j47nNj:b79ll3i) ([raw file](https://github.com/gramineproject/gramine/blob/d97eff5ec2eb85dd547ddfc50afd3e5176eaeeb7/libos/src/sys/libos_poll.c#L123)):*

What's different in `console:` handles from others which also exist on the host (e.g. pipes or normal files) that they need this special treatment?
</blockquote></details>

The `console:` handle:
- is backed by the pseudo-file `/dev/tty`, which is located in the [`pseudo` FS](https://github.com/gramineproject/gramine/blob/2ad54dd52426da115261a26244c10110840f9c83/libos/src/fs/libos_fs.c#L173-L177)
- thus, `handle->fs` points to [`pseudo_builtin_fs` struct](https://github.com/gramineproject/gramine/blob/2ad54dd52426da115261a26244c10110840f9c83/libos/src/fs/libos_fs.c#L32)
- this struct has a [generic `poll()` callback](https://github.com/gramineproject/gramine/blob/2ad54dd52426da115261a26244c10110840f9c83/libos/src/fs/libos_fs_pseudo.c#L480)
- This generic `poll()` callback can call a pseudo-file-type-specific callback, e.g. for `PSEUDO_MEM` files it looks [like this](https://github.com/gramineproject/gramine/blob/2ad54dd52426da115261a26244c10110840f9c83/libos/src/fs/libos_fs_pseudo.c#L487).
- In this PR, I add a similar logic with a "pseudo-file-type-specific callback" for `PSEUDO_DEV` files, to accommodate for my `/dev/tty` logic.

So you can see that there are **two levels of indirection**, and a simple check `if (handle->fs->fs_ops->poll)` is not enough. We need to actually call the first level of a `poll` callback to learn whether the second level of a `poll` callback exists.

Compare with the `pipe:` handle:
- is backed by the [`pipe` FS directly](https://github.com/gramineproject/gramine/blob/2ad54dd52426da115261a26244c10110840f9c83/libos/src/fs/libos_fs.c#L27)
- This `pipe` FS struct [doesn't have a `poll` callback](https://github.com/gramineproject/gramine/blob/2ad54dd52426da115261a26244c10110840f9c83/libos/src/fs/pipe/fs.c#L279-L284)
- Thus, a simple check `if (handle->fs->fs_ops->poll)` is enough.

---

Could we modify `console:` to be its own FS type? Then we'd have just one level of indirection, and the `if (...->poll)` check would be good enough. Yes, we could do this, but I'm not sure it will be cleaner...

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1563)* status: all files reviewed, 11 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`-- commits` line 4 at r2](https://reviewable.io/reviews/gramineproject/gramine/1563#-Nex3qLM4E-8gsLgK2mZ):*

Let me expand on my reasoning.

From the point of view of the applications running inside Gramine:
- Gramine environment can be considered a VM.
- Gramine itself can be considered a Linux kernel.
- Gramine's first process (aka main process aka master process) performs initialization steps that can be considered similar to what `init` (and its helper utilities) do in Linux.

From this perspective, Gramine initialization must start with finding devices (like the teletypewriter aka TTY), building abstractions on top of these devices, and passing control to the user programs.

What does `init` do with respect to initializing stdin/stdout/stderr (aka FDs 0, 1, 2), which then will be inherited by the `init` child processes? The following:
1. Find all TTY devices and create `/dev/ttyXXX` pseudo-files corresponding to these devices
2. Call `getty` utility
3. The `getty` utility opens each `/dev/ttyXXX` pseudo-file and connects its read end to stdin (FD 0), and its write ends to stdout + stderr (FDs 1 and 2).
4. This abstraction/association will be used by all children of `init` (I'm fuzzy on how `getty` sends back to `init` this association).

So Gramine's initialization logic must do a similar thing. In Gramine, we don't have special `init` and `getty` processes, so Gramine's LibOS code does this itself. Also, we don't have multiple TTYs, and we don't add the complexity of pseudo TTYs (PTYs), so we just set up a single `/dev/tty` device and associate FDs 0,1,2 with it.

A bunch of references:
- https://stackoverflow.com/questions/63394927/what-does-the-tty-do-when-a-process-asks-for-input-or-outputs-to-the-screen
- https://unix.stackexchange.com/questions/4126/what-is-the-exact-difference-between-a-terminal-a-shell-a-tty-and-a-con
- https://www.linusakesson.net/programming/tty/ (a great article, recommend)
- https://tldp.org/LDP/intro-linux/html/sect_04_02.html
- https://tldp.org/HOWTO/Text-Terminal-HOWTO-15.html

I hope I didn't screw up too much in my descriptions :)

P.S. If I understand correctly, the above description works both all UNIX-based operating systems. (That's why I say `UNIX` in the commit message instead of `Linux`.) I may be wrong here, but I read the same docs about OpenBSD, so I assume that it's the same for all popular UNIX-based OSes.
</blockquote></details>

Yes, that I agree with, but this is true only for `init` and this sentence seem to speak in general. Also, regarding the rest of this PR - these statements are not true in children processes.
___
*[`-- commits` line 7 at r2](https://reviewable.io/reviews/gramineproject/gramine/1563#-NexFgEx2X9r6HYhxXYf):*

But Gramine is not libc? Gramine's initialization code is akin to the `init` process in Linux. Gramine **prepares** FDs 0,1,2 for the use by the app's first process (which contains libc). Gramine **also prepares** the `/dev/tty` pseudo-file, because some applications may want to use it directly.
</blockquote></details>

I thought this sentence describes how this works in each process, while it's only valid for `init`. Please make it clear that this describes only how the initial process in Unix behaves and how this relates to Gramine.
___
*[`libos/src/sys/libos_poll.c` line 123 at r1](https://reviewable.io/reviews/gramineproject/gramine/1563#-NelpClD6yC2Z2INnOGh:-NfB5VOnCZyqM-dwLaPb:b2miuad) ([raw file](https://github.com/gramineproject/gramine/blob/d97eff5ec2eb85dd547ddfc50afd3e5176eaeeb7/libos/src/sys/libos_poll.c#L123)):*
> We need to actually call the first level of a `poll` callback to learn whether the second level of a `poll` callback exists.

Ahhh, that's how it works. The control flow here is quite obfuscated, I didn't notice that it ends up in the same code path.

I'm fine with how we check whether that nested `poll()` callback exists, I'm just confused by the control flow here. Any idea how to make it more readable?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1563)* status: 25 of 32 files reviewed, 11 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "squash! " found in commit messages' one-liners (waiting on @llly and @mkow)

___
*[`-- commits` line 4 at r2](https://reviewable.io/reviews/gramineproject/gramine/1563#-Nex3qLM4E-8gsLgK2mZ):*
> these statements are not true in children processes

Nothing happens in the children processes, because of this check: https://github.com/gramineproject/gramine/blob/2ad54dd52426da115261a26244c10110840f9c83/libos/src/bookkeep/libos_handle.c#L155

I.e., we don't modify/init anything if the handles array was already inited (which is true for children processes).
___
*[`-- commits` line 7 at r2](https://reviewable.io/reviews/gramineproject/gramine/1563#-NexFgEx2X9r6HYhxXYf):*

I thought this sentence describes how this works in each process, while it's only valid for `init`. Please make it clear that this describes only how the initial process in Unix behaves and how this relates to Gramine.
</blockquote></details>

Done. Check the new commit message.
___
*[`libos/src/sys/libos_poll.c` line 123 at r1](https://reviewable.io/reviews/gramineproject/gramine/1563#-NelpClD6yC2Z2INnOGh:-NfBAXc7CjK7-p_b5YQr:bxrgsdn) ([raw file](https://github.com/gramineproject/gramine/blob/d97eff5ec2eb85dd547ddfc50afd3e5176eaeeb7/libos/src/sys/libos_poll.c#L123)):*

> We need to actually call the first level of a `poll` callback to learn whether the second level of a `poll` callback exists.

Ahhh, that's how it works. The control flow here is quite obfuscated, I didn't notice that it ends up in the same code path.

I'm fine with how we check whether that nested `poll()` callback exists, I'm just confused by the control flow here. Any idea how to make it more readable?
</blockquote></details>

Done. What about this refactoring?
___
*[`libos/test/regression/console.c` line 31 at r2](https://reviewable.io/reviews/gramineproject/gramine/1563#-Nf7qv2O-VudKsD0M9i0:-NfApb6v32EGf6eeJIJN:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/e27e63d4a5cd04b7a20e453ed159a17e7b449ac7/libos/test/regression/console.c#L31)):*

Not `strlen()`? We rather don't want to print the null byte to the terminal.
ditto below
</blockquote></details>

Done.
___
*[`pal/include/pal/pal.h` line 348 at r2](https://reviewable.io/reviews/gramineproject/gramine/1563#-NexYXUJE4AI-NhVDqK_:-NfAphIb2qn8gxlD9uTg:bvpqkw9) ([raw file](https://github.com/gramineproject/gramine/blob/e27e63d4a5cd04b7a20e453ed159a17e7b449ac7/pal/include/pal/pal.h#L348)):*
> ```objc
>  *   #PAL_CREATE_ALWAYS is given in `create` flags, the file/directory will be created.
>  * * `dev:...`: Open a device as a stream.
>  * * `console:`: Open a console (stdin/stdout/stderr) as a stream.
> ```

Done.

By the way, I just realized that this description applies only to process-based PAL of Gramine. For example, a VM-based PAL won't have stdin/stdout/stderr streams; instead such PAL would have e.g. a `virtio-console` driver. Do we want to generalize the description, or it will only complicate matters, and we should stick to the current process-based PALs?
___
*[`pal/src/host/linux/pal_console.c` line 5 at r2](https://reviewable.io/reviews/gramineproject/gramine/1563#-Neymo-61HRxz9HIwcXI:-NfAr-EO9slu1Q2Bsgat:b-6rsa1n) ([raw file](https://github.com/gramineproject/gramine/blob/e27e63d4a5cd04b7a20e453ed159a17e7b449ac7/pal/src/host/linux/pal_console.c#L5)):*

I don't like this merging of these terms. std{in,out,err} may be unrelated to any console.
</blockquote></details>

Done. I rewrote the description, does it make more sense now?
___
*[`pal/src/host/linux/pal_console.c` line 38 at r2](https://reviewable.io/reviews/gramineproject/gramine/1563#-NeyqAkf5wGgHjS5-BRX:-NfAqHSc-rnRiYppmcOy:bb77ucv) ([raw file](https://github.com/gramineproject/gramine/blob/e27e63d4a5cd04b7a20e453ed159a17e7b449ac7/pal/src/host/linux/pal_console.c#L38)):*

We should really finally split these handle opening functions into specialized ones, we keep abusing this interface for opening regular files over and over...
</blockquote></details>

Yep.
___
*[`pal/src/host/linux/pal_devices.c` line 27 at r2](https://reviewable.io/reviews/gramineproject/gramine/1563#-NeysM4G3mfsNMef72y3:-NfAsc5kEg8hIlOfOxCN:brc4b94) ([raw file](https://github.com/gramineproject/gramine/blob/9f80c4245a8a852a2d6ac30c81840c94db447305/pal/src/host/linux/pal_devices.c#L27)):*

(ditto for all these comments to SGX PAL)
</blockquote></details>

Ok, though I find this particular check (on the type of the PAL handle) redundant these days, as I can't imagine how Gramine developers could abuse/mix it.

UPDATE: Actually, in `dev_open()`, I re-instated the IF check, as it makes some sense to verify during the opening of the object. In other functions, I used the assert.
___
*[`pal/src/host/linux/pal_streams.c` line 57 at r2](https://reviewable.io/reviews/gramineproject/gramine/1563#-NeyvPGA-yD4tgex4dHO:-NfArNOj0S6gbGob5cLH:b-sljy67) ([raw file](https://github.com/gramineproject/gramine/blob/e27e63d4a5cd04b7a20e453ed159a17e7b449ac7/pal/src/host/linux/pal_streams.c#L57)):*

What about the state whether the handle was already closed? Shouldn't be writable afterwards, seems that now migration re-opens the handle?
</blockquote></details>

Sorry, I'm not sure about which exact case you talk?

Note that PAL objects of type `PAL_TYPE_CONSOLE` have the `fd` field, and this field is marked as `PAL_IDX_POISON` during PAL-object close. Thus, any `write()` afterwards will return `-PAL_ERROR_DENIED` because the host fd is not real now.

Also note that PAL object checkpoint-and-restore logic serializes all the natively-typed fields of the object. In particular, `pal_handle::console.fd` is serialized and sent as an integer value to the child process (enclave), and the child deserializes this field. So in the end, the child will also have `fd = PAL_IDX_POISON`, and `write()` in the child will also return `-PAL_ERROR_DENIED`.

To see this logic, scroll a bit down in this file, and you'll see `/* copy into buffer all handle fields and then serialized fields */` code snippet.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1563)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "squash! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`-- commits` line 4 at r2](https://reviewable.io/reviews/gramineproject/gramine/1563#-Nex3qLM4E-8gsLgK2mZ):*

> these statements are not true in children processes

Nothing happens in the children processes, because of this check: https://github.com/gramineproject/gramine/blob/2ad54dd52426da115261a26244c10110840f9c83/libos/src/bookkeep/libos_handle.c#L155

I.e., we don't modify/init anything if the handles array was already inited (which is true for children processes).
</blockquote></details>

Looks good now.
___
*[`-- commits` line 51 at r3](https://reviewable.io/reviews/gramineproject/gramine/1563#-NfBdLBT4hY3U1jiX4Ka):*
They may not inherit all of 0,1,2, some may be closed.



_Suggestion:_
```
inherit all the FDs
```
___
*[`-- commits` line 52 at r3](https://reviewable.io/reviews/gramineproject/gramine/1563#-NfBdGgO3kFSNH9HGJzV):*




_Suggestion:_
```
parent process.
```
___
*[`libos/src/sys/libos_poll.c` line 123 at r1](https://reviewable.io/reviews/gramineproject/gramine/1563#-NelpClD6yC2Z2INnOGh:-NfBx5m03HybYo_Gkq3I:by2s728) ([raw file](https://github.com/gramineproject/gramine/blob/d97eff5ec2eb85dd547ddfc50afd3e5176eaeeb7/libos/src/sys/libos_poll.c#L123)):*

Done. What about this refactoring?
</blockquote></details>

Looks good now, thanks!
___
*[`libos/src/sys/libos_poll.c` line 149 at r3](https://reviewable.io/reviews/gramineproject/gramine/1563#-NfBvAVl0dgqLbg-PfER:-NfBvAVl0dgqLbg-PfES:bttvwn7) ([raw file](https://github.com/gramineproject/gramine/blob/b6b4cf08525c4a85786b1a32dd9632a5fc4759f0/libos/src/sys/libos_poll.c#L149)):*
> ```C
>         }
> 
>         PAL_HANDLE pal_handle;
> ```

Now that we don't have that "use_pal_polling" label here maybe we should add a comment that this code path is for PAL polling?
___
*[`pal/include/pal/pal.h` line 348 at r2](https://reviewable.io/reviews/gramineproject/gramine/1563#-NexYXUJE4AI-NhVDqK_:-NfBvTRQ0QhUbj_vXWfx:b-4yjygx) ([raw file](https://github.com/gramineproject/gramine/blob/e27e63d4a5cd04b7a20e453ed159a17e7b449ac7/pal/include/pal/pal.h#L348)):*

Done.

By the way, I just realized that this description applies only to process-based PAL of Gramine. For example, a VM-based PAL won't have stdin/stdout/stderr streams; instead such PAL would have e.g. a `virtio-console` driver. Do we want to generalize the description, or it will only complicate matters, and we should stick to the current process-based PALs?
</blockquote></details>

Yes, that's also a part of my disliking of the current descriptions in this PR :) I'd generalize it more. Gramine just maps first process' 0/1/2 descriptors to some "log output" of Gramine, which may depend on the PAL.
___
*[`pal/src/host/linux/pal_console.c` line 5 at r2](https://reviewable.io/reviews/gramineproject/gramine/1563#-Neymo-61HRxz9HIwcXI:-NfBvnjg-UVw6bz-vrzg:b-ku1u9q) ([raw file](https://github.com/gramineproject/gramine/blob/e27e63d4a5cd04b7a20e453ed159a17e7b449ac7/pal/src/host/linux/pal_console.c#L5)):*

Done. I rewrote the description, does it make more sense now?
</blockquote></details>

Yup, it's better now.
___
*[`pal/src/host/linux/pal_streams.c` line 57 at r2](https://reviewable.io/reviews/gramineproject/gramine/1563#-NeyvPGA-yD4tgex4dHO:-NfBwZf28C0xYq2Jzr9v:b-td51qx) ([raw file](https://github.com/gramineproject/gramine/blob/e27e63d4a5cd04b7a20e453ed159a17e7b449ac7/pal/src/host/linux/pal_streams.c#L57)):*

Sorry, I'm not sure about which exact case you talk?

Note that PAL objects of type `PAL_TYPE_CONSOLE` have the `fd` field, and this field is marked as `PAL_IDX_POISON` during PAL-object close. Thus, any `write()` afterwards will return `-PAL_ERROR_DENIED` because the host fd is not real now.

Also note that PAL object checkpoint-and-restore logic serializes all the natively-typed fields of the object. In particular, `pal_handle::console.fd` is serialized and sent as an integer value to the child process (enclave), and the child deserializes this field. So in the end, the child will also have `fd = PAL_IDX_POISON`, and `write()` in the child will also return `-PAL_ERROR_DENIED`.

To see this logic, scroll a bit down in this file, and you'll see `/* copy into buffer all handle fields and then serialized fields */` code snippet.
</blockquote></details>

Yup, I missed that the native fields are copied normally, didn't look at this code in a while. Looks good then.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1563)* status: 29 of 32 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @llly, and @mkow)

___
*[`-- commits` line 51 at r3](https://reviewable.io/reviews/gramineproject/gramine/1563#-NfBdLBT4hY3U1jiX4Ka):*

They may not inherit all of 0,1,2, some may be closed.
</blockquote></details>

I'll fix during final rebase
___
*[`-- commits` line 52 at r3](https://reviewable.io/reviews/gramineproject/gramine/1563#-NfBdGgO3kFSNH9HGJzV):*
I'll fix during final rebase
___
*[`libos/src/sys/libos_poll.c` line 149 at r3](https://reviewable.io/reviews/gramineproject/gramine/1563#-NfBvAVl0dgqLbg-PfER:-NfFw1OHDM_JXBUGFDyA:bdpsena) ([raw file](https://github.com/gramineproject/gramine/blob/b6b4cf08525c4a85786b1a32dd9632a5fc4759f0/libos/src/sys/libos_poll.c#L149)):*

Now that we don't have that "use_pal_polling" label here maybe we should add a comment that this code path is for PAL polling?
</blockquote></details>

Done.

The path is not exactly for PAL polling (this sounds misleading), but to *add the handle to the array of handles to poll in PAL*.

The actual polling happens outside of this large for-loop, see `PalStreamsWaitEvents()` invocation about 30 lines below.
___
*[`pal/include/pal/pal.h` line 348 at r2](https://reviewable.io/reviews/gramineproject/gramine/1563#-NexYXUJE4AI-NhVDqK_:-NfFwwYS9g_eMErbuGCL:bothqm7) ([raw file](https://github.com/gramineproject/gramine/blob/e27e63d4a5cd04b7a20e453ed159a17e7b449ac7/pal/include/pal/pal.h#L348)):*
Done. Hopefully this reads better.

> to some "log output" of Gramine

I don't like your `log output`, this is misleading. It's not only output, it's only the input channel. Also, it's not about the log (the Gramine log), but about whatever output the application is sending.

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1563)* status: all files reviewed, 6 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`libos/src/sys/libos_poll.c` line 149 at r4](https://reviewable.io/reviews/gramineproject/gramine/1563#-NfIhj21Enkxa7LgSCmY:-NfIhj21Enkxa7LgSCmZ:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/33ff713e187963e64211841cbdbea2838625a7a6/libos/src/sys/libos_poll.c#L149)):*
> ```C
>         }
> 
>         /* below code path adds the handle for PAL polling */
> ```





_Suggestion:_
```C
/* add the handle for PAL polling */
```
___
*[`pal/include/pal/pal.h` line 348 at r2](https://reviewable.io/reviews/gramineproject/gramine/1563#-NexYXUJE4AI-NhVDqK_:-NfIhvJoEzqAm3OslxQW:b-q6djyj) ([raw file](https://github.com/gramineproject/gramine/blob/e27e63d4a5cd04b7a20e453ed159a17e7b449ac7/pal/include/pal/pal.h#L348)):*

Done. Hopefully this reads better.

> to some "log output" of Gramine

I don't like your `log output`, this is misleading. It's not only output, it's only the input channel. Also, it's not about the log (the Gramine log), but about whatever output the application is sending.
</blockquote></details>

Yeah, now it's clear :)

kailun-qin (changes requested):
> Reviewed 24 of 32 files at r1, 5 of 7 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1563)* status: all files reviewed, 8 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`libos/src/fs/dev/fs.c` line 105 at r4](https://reviewable.io/reviews/gramineproject/gramine/1563#-NfJOB8dEb-G3yvEcMMQ:-NfJOB8dEb-G3yvEcMMR:bb13ay5) ([raw file](https://github.com/gramineproject/gramine/blob/33ff713e187963e64211841cbdbea2838625a7a6/libos/src/fs/dev/fs.c#L105)):*
> ```C
> }
> 
> /* this dummy function is implemented to support opening TTY (console) with O_TRUNC flag */
> ```

But shouldn't we ignore `O_TRUNC` flag for such case?



_Code quote:_
```C
/* this dummy function is implemented to support opening TTY (console) with O_TRUNC flag */
```
___
*[`libos/test/regression/console.c` line 2 at r4](https://reviewable.io/reviews/gramineproject/gramine/1563#-NfJEOXa0ReeqGcTPny5:-NfJEOXa0ReeqGcTPny6:b1o9b3q) ([raw file](https://github.com/gramineproject/gramine/blob/33ff713e187963e64211841cbdbea2838625a7a6/libos/test/regression/console.c#L2)):*
> ```C
> /* Copyright (C) 2023 Intel Corporation */
> ```

pls add an empty line between the copyright and the includes

kailun-qin (changes requested):
> Reviewed 2 of 2 files at r5.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1563)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`libos/src/fs/dev/fs.c` line 105 at r4](https://reviewable.io/reviews/gramineproject/gramine/1563#-NfJOB8dEb-G3yvEcMMQ:-NfKa8AZCvJr-Z-B3gAl:bnxpc9) ([raw file](https://github.com/gramineproject/gramine/blob/33ff713e187963e64211841cbdbea2838625a7a6/libos/src/fs/dev/fs.c#L105)):*

@kailun-qin Sorry, I don't understand your comment. Let me give some context:

LibOS generic FS code, when it sees `open(..., O_TRUNC)`, will call the file-specific `truncate()` callback. This is described in our headers, and this is implemented in our C code:
- https://github.com/gramineproject/gramine/blob/2ad54dd52426da115261a26244c10110840f9c83/libos/include/libos_fs.h#L687
- https://github.com/gramineproject/gramine/blob/2ad54dd52426da115261a26244c10110840f9c83/libos/src/fs/libos_namei.c#L408-L419

And because `/dev/tty` is represented as a pseudo-file, the above logic applies to this file as well. So the generic FS code will try to find the `truncate()` callback, and if won't find this callback, the generic FS code fails with `-EINVAL` (see C code above). We want to prevent this failure, but we also don't need to do anything in the truncate callback. That's the whole explanation.
</blockquote></details>

Yeah, I understand the flow. But from the manpage: https://man7.org/linux/man-pages/man2/open.2.html -- `O_TRUNC: ...If the file is a FIFO or terminal device file, the O_TRUNC flag is ignored.`?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1563)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin and @mkow)

___
*[`libos/src/fs/dev/fs.c` line 105 at r4](https://reviewable.io/reviews/gramineproject/gramine/1563#-NfJOB8dEb-G3yvEcMMQ:-NfKcJ8R8pHCk1rEJDi0:b-wz76a9) ([raw file](https://github.com/gramineproject/gramine/blob/33ff713e187963e64211841cbdbea2838625a7a6/libos/src/fs/dev/fs.c#L105)):*

Yeah, I understand the flow. But from the manpage: https://man7.org/linux/man-pages/man2/open.2.html -- `O_TRUNC: ...If the file is a FIFO or terminal device file, the O_TRUNC flag is ignored.`?
</blockquote></details>

Yep, but that's what we do in this `dev_tty_truncate()` callback? We just ignore the truncation request, which is the same as ignoring the `O_TRUNC` flag.

Or do you think that this ignore-O_TRUNC logic should be in `libos_namei.c` file? It can't be there because at that point we don't know which file we're working on -- there's no convenient way to determine that the current file is a "terminal device file".

So we move the logic into this callback, where we simply ignore.

kailun-qin (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1563)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`libos/src/fs/dev/fs.c` line 105 at r4](https://reviewable.io/reviews/gramineproject/gramine/1563#-NfJOB8dEb-G3yvEcMMQ:-NfKbVynB7wEslq8veVq:b-uj6p4g) ([raw file](https://github.com/gramineproject/gramine/blob/33ff713e187963e64211841cbdbea2838625a7a6/libos/src/fs/dev/fs.c#L105)):*
> Yep, but that's what we do in this `dev_tty_truncate()` callback? We just ignore the truncation request, which is the same as ignoring the `O_TRUNC` flag.

Yes, it's functionally equivalent for this `open(terminal device file, O_TRUNC)` case. But what if we directly call `truncate(, 0)`/`ftruncate(, 0)` on the terminal device file (descriptor)? Then Gramine would return success whereas Linux would return failure I guess.

> Or do you think that this ignore-O\_TRUNC logic should be in `libos_namei.c` file?

Yes this was what I thought.

>  It can't be there because at that point we don't know which file we're working on -- there's no convenient way to determine that the current file is a "terminal device file".

Ok, then. Maybe we can at least note somewhere if the above stated behavior is indeed inconsistent w/ Linux?

dimakuv:
*[`libos/src/fs/dev/fs.c` line 105 at r4](https://reviewable.io/reviews/gramineproject/gramine/1563#-NfJOB8dEb-G3yvEcMMQ:-NfKlfsg63vQDEZD5zUC:b-dldfnb) ([raw file](https://github.com/gramineproject/gramine/blob/33ff713e187963e64211841cbdbea2838625a7a6/libos/src/fs/dev/fs.c#L105)):*
@kailun-qin is right. I just did a quick test:
```
#include <errno.h>
#include <fcntl.h>
#include <stdio.h>
#include <sys/stat.h>
#include <sys/types.h>
#include <unistd.h>

int main(void) {
    int tty = open("/dev/tty", O_WRONLY | O_CLOEXEC);
    write(tty, "Hi to /dev/tty!\n", sizeof("Hi to /dev/tty!\n"));
    int ret = ftruncate(tty, /*length=*/0);
    printf("ftruncate returns %d (errno = %d)\n", ret, errno);
    close(tty);
    return 0;
}
```

Compile this file and run natively and under Gramine:
```
$ ./helloworld
Hello, world
Hi to /dev/tty!
ftruncate returns -1 (errno = 22)

$ gramine-direct helloworld
Hi to /dev/tty!
Hello, world
ftruncate returns 0 (errno = 0)
```

Native run correctly fails `ftruncate()` with -22 which is `EINVAL`. As the man page says:
> `EINVAL     fd does not reference a regular file or a POSIX shared memory object.`

Let me fix this...

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1563)* status: 30 of 33 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin, @llly, and @mkow)

___
*[`libos/src/fs/libos_fs_pseudo.c` line 439 at r6](https://reviewable.io/reviews/gramineproject/gramine/1563#-NfKrEnFBJf5PkXjadYq:-NfKrEnFBJf5PkXjadYr:b-7vmes6) ([raw file](https://github.com/gramineproject/gramine/blob/7179d102ccdd81dec0c6bb22b0aea8bdcbdf992c/libos/src/fs/libos_fs_pseudo.c#L439)):*
> ```C
>         case PSEUDO_DEV:
>             if (!node->dev.dev_ops.truncate)
>                 return -EINVAL;
> ```

FYI: Linux returns `EINVAL` on ftruncate on non regular files, see `man 2 ftruncate`
___
*[`libos/src/fs/libos_namei.c` line 410 at r6](https://reviewable.io/reviews/gramineproject/gramine/1563#-NfKrX7BA-i0ioJ-BnOT:-NfKrX7BA-i0ioJ-BnOU:br7krf5) ([raw file](https://github.com/gramineproject/gramine/blob/7179d102ccdd81dec0c6bb22b0aea8bdcbdf992c/libos/src/fs/libos_namei.c#L410)):*
> ```C
>     /* truncate regular writable file if O_TRUNC is given */
>     if ((flags & O_TRUNC) && ((flags & O_RDWR) | (flags & O_WRONLY))
>             && (dent->inode->type != S_IFCHR)
> ```

According to `man 2 open`, `O_TRUNC` is only applicable to regular files. So I ignore it for devices too now.
___
*[`libos/src/fs/dev/fs.c` line 105 at r4](https://reviewable.io/reviews/gramineproject/gramine/1563#-NfJOB8dEb-G3yvEcMMQ:-NfKrh9v3rTrYVhCD1F_:bge6v3e) ([raw file](https://github.com/gramineproject/gramine/blob/33ff713e187963e64211841cbdbea2838625a7a6/libos/src/fs/dev/fs.c#L105)):*

@kailun-qin is right. I just did a quick test:
```
#include <errno.h>
#include <fcntl.h>
#include <stdio.h>
#include <sys/stat.h>
#include <sys/types.h>
#include <unistd.h>

int main(void) {
    int tty = open("/dev/tty", O_WRONLY | O_CLOEXEC);
    write(tty, "Hi to /dev/tty!\n", sizeof("Hi to /dev/tty!\n"));
    int ret = ftruncate(tty, /*length=*/0);
    printf("ftruncate returns %d (errno = %d)\n", ret, errno);
    close(tty);
    return 0;
}
```

Compile this file and run natively and under Gramine:
```
$ ./helloworld
Hello, world
Hi to /dev/tty!
ftruncate returns -1 (errno = 22)

$ gramine-direct helloworld
Hi to /dev/tty!
Hello, world
ftruncate returns 0 (errno = 0)
```

Native run correctly fails `ftruncate()` with -22 which is `EINVAL`. As the man page says:
> `EINVAL     fd does not reference a regular file or a POSIX shared memory object.`

Let me fix this...
</blockquote></details>

Done.

Now the correct behavior is restored (same test I mention in the above comment):
```
$ ./helloworld
Hi to /dev/tty!
ftruncate returns -1 (errno = 22)

$ gramine-direct helloworld
Hi to /dev/tty!
ftruncate returns -1 (errno = 22)

$ gramine-sgx helloworld
Hi to /dev/tty!
ftruncate returns -1 (errno = 22)
```

Thanks @kailun-qin for noticing this subtle difference.

mkow (changes requested):
> Reviewed 2 of 2 files at r5, 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1563)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @kailun-qin)

___
*[`libos/src/fs/libos_namei.c` line 410 at r6](https://reviewable.io/reviews/gramineproject/gramine/1563#-NfKrX7BA-i0ioJ-BnOT:-NfLqAMW8CYdiRVQeOaI:bhh574m) ([raw file](https://github.com/gramineproject/gramine/blob/7179d102ccdd81dec0c6bb22b0aea8bdcbdf992c/libos/src/fs/libos_namei.c#L410)):*
> only applicable to regular files

Then why not do an "if type == regular" check instead of excluding all other types?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1563)* status: 32 of 33 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin and @mkow)

___
*[`libos/src/fs/libos_namei.c` line 410 at r6](https://reviewable.io/reviews/gramineproject/gramine/1563#-NfKrX7BA-i0ioJ-BnOT:-NfM0m0s4vvN2d93qCMA:birude7) ([raw file](https://github.com/gramineproject/gramine/blob/7179d102ccdd81dec0c6bb22b0aea8bdcbdf992c/libos/src/fs/libos_namei.c#L410)):*

Done. Heh, true. I don't know why Pawel wrote it like that at the time (~2 years ago).
</blockquote></details>

There was a similar PR for FIFOs: https://github.com/gramineproject/gramine/pull/1401

kailun-qin (approved):
> Reviewed 2 of 3 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1563)* status: all files reviewed, 3 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1563)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1563)* status: 32 of 33 files reviewed, 2 unresolved discussions (waiting on @kailun-qin, @llly, and @mkow)

___
*[`-- commits` line 51 at r3](https://reviewable.io/reviews/gramineproject/gramine/1563#-NfBdLBT4hY3U1jiX4Ka):*

I'll fix during final rebase
</blockquote></details>

Done.
___
*[`-- commits` line 52 at r3](https://reviewable.io/reviews/gramineproject/gramine/1563#-NfBdGgO3kFSNH9HGJzV):*

I'll fix during final rebase
</blockquote></details>

Done.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1563)* status: 32 of 33 files reviewed, 2 unresolved discussions (waiting on @kailun-qin, @llly, and @mkow)

___
*[`libos/src/fs/libos_namei.c` line 410 at r6](https://reviewable.io/reviews/gramineproject/gramine/1563#-NfKrX7BA-i0ioJ-BnOT:-NfPcwdUDXs9bR6dasjY:b-c4w922) ([raw file](https://github.com/gramineproject/gramine/blob/7179d102ccdd81dec0c6bb22b0aea8bdcbdf992c/libos/src/fs/libos_namei.c#L410)):*

There was a similar PR for FIFOs: https://github.com/gramineproject/gramine/pull/1401
</blockquote></details>

I verified that this PR (the final rebased version) fixes the issue in #1401. So I will close that PR.

kailun-qin (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1563)* status: all files reviewed, 2 unresolved discussions (waiting on @mkow)

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1563)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #1584, opened 2023-10-02
[python] Add gramine-sgx-sign --chroot option

woju:
## Description of the changes 
see commit message

## How to test this PR? 

1. Checkout this branch and build gramine.
2. Install this to your system.
3. `sed -i -e 's@file:helloworld@file:/app/helloworld@g' path/to/gramine/CI-Examples/helloworld`
4. Build `helloworld` example (only the manifest and binary, signing will come later).
5. Deboostrap an example chroot, e.g. using mmdebstrap, and install the same gramine into your chroot; copy an example into the chroot and measure it from outside. The following is a heavily modified mmdebstrap invocation I have lying aroun d (my version compiles a package using `debuild`, this version is hopefully simpler and builds using only meson):
```sh
mmdebstrap \
    --customize-hook 'DESTDIR="$1" meson install -C path/to/the/build/dir' \
    --customize-hook 'cp -r path/to/repo/of/gramine/CI-Examples/helloworld "$1"/app' \
    --customize-hook 'gramine-sgx-sign --chroot "$1" --manifest /app/helloworld.manifest --output /helloworld.manifest.sgx' \
    bookworm rootfs.tar 'deb http://deb.debian.org/debian main'
```
(this invocation is possibly missing some gramine dependencies, didn't check directly; maybe add `--include $package` for each missing package — my version uses `apt-get install ./gramine_1.5post~UNRELEASED_amd64.deb` and installs dependencies automatically).
6. run Gramine from this chroot, e.g. by building a docker image from it:
```Dockerfile
FROM scratch
ADD rootfs.tar /
WORKDIR /app
CMD ["gramine-sgx", "helloworld"]
```
```sh
docker build .
docker run --rm -it --device /dev/sgx_enclave --volume /var/run/aesmd/aesm.socket:/var/run/aesmd/aesm.socket sha256:...
```

## FAQ

1. Why can't you just `chroot "$1" gramine-sgx-sign ...`

   Because sgx signing plugins might not be installed inside the image/chroot.

2. What could possibly go wrong?

   Most of the CI-Examples won't work out of the box because `--chroot` option requires the paths to be absolute (to avoid ambiguity about relative paths, what directory those are related to, and how to translate that into chroot). And CI-Examples have relative paths e.g. to the current directory (like `file:helloworld`, hence `sed` in the testing instruction).

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1584)

### DIFF
--- a/Documentation/manpages/gramine-sgx-sign.rst
+++ b/Documentation/manpages/gramine-sgx-sign.rst
@@ -61,6 +61,17 @@ Command line arguments
     modified. The dependency file is in Makefile format, and is suitable for
     using in build systems (Make, Ninja).
 
+.. option:: --chroot <path>
+
+    When calculating cryptographic hashes of trusted files, measure files inside
+    a |~| chroot instead of paths in root of the file system. Requires that all
+    paths in manifest are absolute, and those will be interpreted as relative to
+    the directory specified as the value of the option.
+
+    Note you need to be very careful that the Gramine runtime binaries are
+    exactly the same inside chroot as the ones used to execute
+    :program:`gramine-sgx-sign`.
+
 .. option:: --verbose, -v
 
     Print details to standard output. This is the default.
--- a/Documentation/python/api.rst
+++ b/Documentation/python/api.rst
@@ -57,6 +57,8 @@ API Reference
 .. autoclass:: graminelibos.Manifest
    :members:
 
+.. autoclass:: graminelibos.manifest.TrustedFile
+   :members:
 ..
   TODO: enable this once we build Gramine on readthedocs
   .. autoclass:: graminelibos.Sigstruct
--- a/python/gramine-sgx-sign
+++ b/python/gramine-sgx-sign
@@ -82,6 +82,9 @@ class BCDDateParamType(click.ParamType):
 @click.option('--date', '-d',
     type=BCDDateParamType(), default='today',
     help='Set DATE field in SIGSTRUCT to this value (YYYY-MM-DD or "today")')
+@click.option('--chroot',
+    type=click.Path(exists=True, dir_okay=True, file_okay=False),
+    help='Measure a chroot directory, not the host filesystem')
 @click.option('--sigfile', '-s',
     help='Output .sig file')
 @click.option('--depfile',
@@ -94,7 +97,8 @@ class BCDDateParamType(click.ParamType):
     nargs=-1,
     type=click.UNPROCESSED)
 @click.pass_context
-def main(ctx, with_, output, libpal, manifest_file, date, sigfile, depfile, verbose, plugin_args):
+def main(ctx, with_, output, libpal, manifest_file, date, sigfile, depfile, verbose, plugin_args,
+         chroot):
     # pylint: disable=too-many-arguments, too-many-locals
 
     ret = get_sgx_sign_plugin(with_)(args=plugin_args, standalone_mode=False)
@@ -125,7 +129,7 @@ def main(ctx, with_, output, libpal, manifest_file, date, sigfile, depfile, verb
 
     manifest = Manifest.load(manifest_file)
 
-    expanded = manifest.expand_all_trusted_files()
+    expanded = manifest.expand_all_trusted_files(chroot=chroot)
 
     with open(output, 'wb') as f:
         manifest.dump(f)
--- a/python/graminelibos/manifest.py
+++ b/python/graminelibos/manifest.py
@@ -3,6 +3,8 @@
 # Copyright (C) 2022 Intel Corporation
 #                    Michał Kowalczyk <mkow@invisiblethingslab.com>
 #                    Borys Popławski <borysp@invisiblethingslab.com>
+# Copyright (C) 2023 Intel Corporation
+#                    Wojtek Porczyk <woju@invisiblethingslab.com>
 
 """
 Gramine manifest management and rendering
@@ -27,51 +29,186 @@ class ManifestError(Exception):
     Contains a string with error description.
     """
 
-def hash_file_contents(path):
-    with open(path, 'rb') as f:
-        sha = hashlib.sha256()
-        for chunk in iter(lambda: f.read(128 * sha.block_size), b''):
-            sha.update(chunk)
-        return sha.hexdigest()
-
 def uri2path(uri):
     if not uri.startswith('file:'):
         raise ManifestError(f'Unsupported URI type: {uri}')
     return pathlib.Path(uri[len('file:'):])
 
-def append_tf(trusted_files, path, hash_=None):
-    if path not in trusted_files:
-        trusted_files[path] = hash_ if hash_ is not None else hash_file_contents(path)
 
-def append_trusted_dir_or_file(trusted_files, val, expanded):
-    if isinstance(val, dict):
-        uri = val['uri']
-        if val.get('sha256'):
-            append_tf(trusted_files, uri2path(uri), val['sha256'])
-            return
-    elif isinstance(val, str):
-        uri = val
-    else:
-        raise ManifestError(f'Unknown trusted file format: {val!r}')
+class TrustedFile:
+    """Represents a single entry in sgx.trusted_files.
 
-    path = uri2path(uri)
-    if not path.exists():
-        raise ManifestError(f'Cannot resolve {path}')
-    if path.is_dir():
-        if not uri.endswith('/'):
-            raise ManifestError(f'Directory URI ({uri}) does not end with "/"')
+    Args:
+        uri (str): URI
+        sha256 (str or None): sha256
+        chroot (pathlib.Path or None): optional path to chroot, if being measured in chroot dir
 
-        expanded.append(path)
-        for sub_path in sorted(path.rglob('*')):
-            expanded.append(sub_path)
-            if sub_path.is_file():
-                # Skip inaccessible files
-                if os.access(sub_path, os.R_OK):
-                    append_tf(trusted_files, sub_path)
-    else:
-        assert path.is_file()
-        append_tf(trusted_files, path)
-        expanded.append(path)
+    Raises:
+        graminelibos.ManifestError: on invalid URI values, or when *chroot* is not None and realpath
+            is not absolute
+    """
+    def __init__(self, uri, sha256=None, *, chroot=None):
+        #: URI of the trusted file
+        self.uri = uri
+        #: sha256 of the trusted file as str of hex digits, or None if not measured
+        self.sha256 = sha256
+        #: optional chroot, if the file is to be measured in a subdirectory
+        self.chroot = pathlib.Path(chroot) if chroot is not None else chroot
+
+        #: real path to the file on disk, including chroot path if specified
+        self.realpath = None
+
+        path = pathlib.PurePosixPath(uri2path(uri))
+
+        if self.chroot is None:
+            self.realpath = pathlib.Path(path)
+        else:
+            if not path.is_absolute():
+                raise ManifestError('only absolute paths can be measured in chroot')
+            self.realpath = self.chroot / path.relative_to('/')
+
+    @classmethod
+    def from_manifest(cls, data, *, chroot=None):
+        """Create an instance from an entry in manifest.
+
+        Args:
+            data (str or dict): what is found in manifest data
+            chroot (pathlib.Path or None): optional path to chroot, if being measured in chroot dir
+
+        Returns:
+            TrustedFile: a single instance of TrustedFile
+
+        Raises:
+            graminelibos.ManifestError: on errors in data
+        """
+        if isinstance(data, str):
+            uri, sha256 = data, None
+
+        elif isinstance(data, dict):
+            uri, sha256 = data.pop('uri'), data.pop('sha256', None)
+            if data:
+                # there are some unknown keys left after two .pop()s above
+                raise ManifestError(f'Leftover trusted file items: {data!r}')
+
+        else:
+            raise ManifestError(f'Unknown trusted file format: {data!r}')
+
+        return cls(uri, sha256, chroot=chroot)
+
+    @classmethod
+    def from_realpath(cls, realpath, *, chroot=None):
+        """Create an instance from a realpath.
+
+        This is used for recursive expansion of directories.
+
+        Args:
+            realpath (pathlib.Path): path to the file
+            chroot (pathlib.Path or None): optional path to chroot, if being measured in chroot dir
+
+        Returns:
+            TrustedFile: a single instance of TrustedFile
+
+        Raises:
+            ValueError: when *chroot* is not None and realpath is not inside manifest
+        """
+        path = pathlib.PurePosixPath(realpath)
+        if chroot is not None:
+            # path.relative_to(chroot) will throw ValueError if the path is not relative to chroot
+            path = '/' / path.relative_to(chroot)
+        self = cls(f'file:{path}{"/" if realpath.is_dir() else ""}', chroot=chroot)
+        assert self.realpath == realpath
+        return self
+
+    def __repr__(self):
+        return (f'<{type(self).__name__}('
+                    f'uri={self.uri!r}, sha256={self.sha256!r}, chroot={self.chroot!r}'
+                f') realpath={self.realpath!r}>')
+
+
+    def to_manifest(self):
+        """Returns the representation of the current file for manifest.
+
+        Returns:
+            str or dict: To be included as element in ``sgx.trusted_files`` list.
+        """
+        if self.sha256 is None:
+            return self.uri
+        return {
+            'uri': self.uri,
+            'sha256': self.sha256,
+        }
+
+
+    def ensure_hash(self):
+        """Ensures that the trusted file carries the sha256 sum.
+
+        If not, this method will open the file and measure it.
+
+        Returns:
+            TrustedFile: self
+        """
+        if self.sha256 is None:
+            with open(self.realpath, 'rb') as file:
+                sha = hashlib.sha256()
+                for chunk in iter(lambda: file.read(128 * sha.block_size), b''):
+                    sha.update(chunk)
+                self.sha256 = sha.hexdigest()
+        return self
+
+
+    def expand_directory(self, *, recursive=True, skip_inaccessible=True):
+        """If this TrustedFile is a directory, iterate over its contents.
+
+        If the TrustedFile instance is referring to a regular file, yield self and stop iteration.
+
+        Args:
+            recursive (bool): If :py:obj:`False`, will iterate only over direct descendants,
+                yielding files and directories; if :py:obj:`True`, will recursively descend into all
+                directories, yielding only regular files.
+            skip_inaccessible (bool): If :py:obj:`True` (the default), will skip entries that are
+                neither directories nor regular files, or fail ``os.access(realpath, os.R_OK)``. If
+                :py:obj:`False`, will iterate over files that failed access test and will possibly
+                error out on while measuring. This argument applies only while recursing into
+                directory (if the instance is referring to a regular file, it will be yielded
+                regardless).
+
+        Yields:
+            :py:class:`TrustedFile`: one object for each entry in the directory
+
+        Raises:
+            graminelibos.ManifestError: On errors in URIs, e.g. when directory does not have ``/``
+                at the end or *vice versa*, or when directory has ``sha256`` value.
+        """
+        if self.uri.endswith('/'):
+            if not self.realpath.is_dir():
+                raise ManifestError(f'URI {self.uri!r} ends with "/" but is not a directory')
+            if self.sha256 is not None:
+                raise ManifestError(f'Directory URI ({self.uri!r}) has sha256 specified')
+
+            for realpath in sorted(self.realpath.glob('*')):
+                # this conditional could be one-lined, but please don't, it would be unreadable
+                if skip_inaccessible:
+                    if not realpath.is_file() and not realpath.is_dir():
+                        continue
+                    if not os.access(realpath, os.R_OK):
+                        continue
+
+                tf = type(self).from_realpath(realpath, chroot=self.chroot)
+
+                if not recursive:
+                    yield tf
+                else:
+                    if realpath.is_symlink() and realpath.is_dir():
+                        # do not descend into symlinked directories
+                        continue
+                    yield from tf.expand_directory(
+                        recursive=recursive, skip_inaccessible=skip_inaccessible)
+
+        else:
+            if self.realpath.is_dir():
+                raise ManifestError(f'Directory URI ({self.uri!r}) does not end with "/"')
+            yield self
+
 
 class Manifest:
     """Just a representation of a manifest.
@@ -171,29 +308,46 @@ class Manifest:
     def dump(self, f):
         tomli_w.dump(self._manifest, f)
 
-    def expand_all_trusted_files(self):
+    def expand_all_trusted_files(self, chroot=None):
         """Expand all trusted files entries.
 
         Collects all trusted files entries, hashes each of them (skipping these which already had a
         hash present) and updates ``sgx.trusted_files`` manifest entry with the result.
 
-        Returns a list of all expanded files, i.e. files that we need to hash, and directories that
-        we needed to list.
+        Returns a list of all expanded files, as included in the manifest.
+
+        Args:
+            chroot (pathlib.Path or None): Optional chroot directory. If specified, trusted files
+                are expected to be found inside this directory, not in root of filesystem.
 
         Raises:
-            ManifestError: There was an error with the format of some trusted files in the manifest
-                or some of them could not be loaded from the filesystem.
+            graminelibos.ManifestError: There was an error with the format of some trusted files in
+                the manifest or some of them could not be loaded from the filesystem.
 
         """
         trusted_files = {}
-        expanded = []
-        for tf in self['sgx']['trusted_files']:
-            append_trusted_dir_or_file(trusted_files, tf, expanded)
+        for data in self['sgx']['trusted_files']:
+            for tf in TrustedFile.from_manifest(data, chroot=chroot).expand_directory():
+                if tf.uri in trusted_files:
+                    # On duplicate entries, pick the one that is already measured, and if both don't
+                    # have hashes, prefer existing one, to avoid dict insertion. Accept double
+                    # (matching) sha256 deduplicating them, and error out on conflicting
+                    # measurement.
+                    tf_old = trusted_files[tf.uri]
+                    if tf_old.sha256 is not None:
+                        if tf.sha256 is not None and tf.sha256 != tf_old.sha256:
+                            raise ManifestError(
+                                f'Two different sha256 values ({tf_old.sha256} and {tf.sha256}) '
+                                f'for the same URI {tf.uri!r}')
+                        continue
 
-        self['sgx']['trusted_files'] = [
-            {'uri': f'file:{k}', 'sha256': v} for k, v in trusted_files.items()
-        ]
-        return expanded
+                trusted_files[tf.uri] = tf
+
+        for tf in trusted_files.values():
+            tf.ensure_hash()
+
+        self['sgx']['trusted_files'] = [tf.to_manifest() for tf in trusted_files.values()]
+        return [tf.realpath for tf in trusted_files.values()]
 
     def get_dependencies(self):
         """Generate list of files which this manifest depends on.
@@ -205,7 +359,7 @@ class Manifest:
             list(pathlib.Path): List of paths to the files this manifest depends on.
 
         Raises:
-            ManifestError: One of the found URIs is in an unsupported format.
+            graminelibos.ManifestError: One of the found URIs is in an unsupported format.
         """
         deps = set()
 


woju:
Jenkins, test check-python-platlib please

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1584)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @woju)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/1584#-NfoO1r20TbSM_wC2mRq):*
I'd prefer to have the command in backticks. Can be done at final rebase.
___
*[`python/graminelibos/manifest.py` line 42 at r1](https://reviewable.io/reviews/gramineproject/gramine/1584#-NfoS8K2-dNOGLYfAnHt:-NfoS8K2-dNOGLYfAnHu:bvspwy2) ([raw file](https://github.com/gramineproject/gramine/blob/94899398023e2dd2b9486aac03cc2996c2f1b13d/python/graminelibos/manifest.py#L42)):*
> ```Python
>     return pathlib.Path(uri[len('file:'):])
> 
> def append_tf(trusted_files, path, hash_=None):
> ```

I verified that these moved and inlined funcs are the same (bar my comments) in the new version as in the old version.
___
*[`python/graminelibos/manifest.py` line 146 at r1](https://reviewable.io/reviews/gramineproject/gramine/1584#-NfoRTmY5i-tnCpk1dVC:-NfoRTmY5i-tnCpk1dVD:bmf3hvp) ([raw file](https://github.com/gramineproject/gramine/blob/b1b03eb6f95b0be0ea7fc1f1dfec4b267ce37b23/python/graminelibos/manifest.py#L146)):*
> ```Python
>         hash present) and updates ``sgx.trusted_files`` manifest entry with the result.
> 
>         Returns a list of all expanded files, i.e. files that we need to hash, and directories that
> ```

`files that we need` -> `files that we needed`
___
*[`python/graminelibos/manifest.py` line 148 at r1](https://reviewable.io/reviews/gramineproject/gramine/1584#-NfoOZgT37vpSgVB9FQj:-NfoOZgT37vpSgVB9FQk:bx2qx4o) ([raw file](https://github.com/gramineproject/gramine/blob/b1b03eb6f95b0be0ea7fc1f1dfec4b267ce37b23/python/graminelibos/manifest.py#L148)):*
> ```Python
>         Returns a list of all expanded files, i.e. files that we need to hash, and directories that
>         we needed to list.
> ```

Don't you want to add the `Args:` section in this comment?
___
*[`python/graminelibos/manifest.py` line 173 at r1](https://reviewable.io/reviews/gramineproject/gramine/1584#-NfoPDgj7sOJHlUU4FDb:-NfoPDgj7sOJHlUU4FDc:b-c48qv4) ([raw file](https://github.com/gramineproject/gramine/blob/b1b03eb6f95b0be0ea7fc1f1dfec4b267ce37b23/python/graminelibos/manifest.py#L173)):*
> ```Python
>         def append_tf(chroot_path):
>             if path not in trusted_files:
>                 trusted_files[out_chroot(chroot_path)] = hash_file_contents(chroot_path)
> ```

How does this function work? It has a single arg `chroot_path`, but it uses both `path` and `chroot_path`?

UPDATE: Ok, I see the `path` variable in the caller function. Is this really how you want to write it? Very confusing.

UPDATE: No, something doesn't end up. `path` in the caller function (outer scope) is a **parent path**, not the current sub-path. How does all this work?

UPDATE: Ok, I think you wanted to do like this but something got missing during refactoring:
```python
        def append_tf(chroot_path):
            path = out_chroot(chroot_path)
            if path not in trusted_files:
                trusted_files[path] = hash_file_contents(chroot_path)
```

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1584)* status: 2 of 4 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @woju)

___
*[`python/graminelibos/manifest.py` line 42 at r1](https://reviewable.io/reviews/gramineproject/gramine/1584#-NfoS8K2-dNOGLYfAnHt:-NfpoMFB17LoLa83P33l:b-h5xis0) ([raw file](https://github.com/gramineproject/gramine/blob/94899398023e2dd2b9486aac03cc2996c2f1b13d/python/graminelibos/manifest.py#L42)):*

I verified that these moved and inlined funcs are the same (bar my comments) in the new version as in the old version.
</blockquote></details>

I'm sorry, I have rewritten this all yet again. :/
___
*[`python/graminelibos/manifest.py` line 146 at r1](https://reviewable.io/reviews/gramineproject/gramine/1584#-NfoRTmY5i-tnCpk1dVC:-NfpoZK5D0KEHYehDGYG:b-lc0zit) ([raw file](https://github.com/gramineproject/gramine/blob/b1b03eb6f95b0be0ea7fc1f1dfec4b267ce37b23/python/graminelibos/manifest.py#L146)):*

`files that we need` -> `files that we needed`
</blockquote></details>

I have changed this to list only the files themselves, as included in the manifest. Is this still OK, and if so, how do I reword this comment?
___
*[`python/graminelibos/manifest.py` line 148 at r1](https://reviewable.io/reviews/gramineproject/gramine/1584#-NfoOZgT37vpSgVB9FQj:-NfpoQtFEk13fkmmzUh_:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/b1b03eb6f95b0be0ea7fc1f1dfec4b267ce37b23/python/graminelibos/manifest.py#L148)):*

Don't you want to add the `Args:` section in this comment?
</blockquote></details>

Done.
___
*[`python/graminelibos/manifest.py` line 173 at r1](https://reviewable.io/reviews/gramineproject/gramine/1584#-NfoPDgj7sOJHlUU4FDb:-NfoVVUhAVi5Lfa43rVR:bk4aoht) ([raw file](https://github.com/gramineproject/gramine/blob/b1b03eb6f95b0be0ea7fc1f1dfec4b267ce37b23/python/graminelibos/manifest.py#L173)):*

How does this function work? It has a single arg `chroot_path`, but it uses both `path` and `chroot_path`?

UPDATE: Ok, I see the `path` variable in the caller function. Is this really how you want to write it? Very confusing.

UPDATE: No, something doesn't end up. `path` in the caller function (outer scope) is a **parent path**, not the current sub-path. How does all this work?

UPDATE: Ok, I think you wanted to do like this but something got missing during refactoring:
```python
        def append_tf(chroot_path):
            path = out_chroot(chroot_path)
            if path not in trusted_files:
                trusted_files[path] = hash_file_contents(chroot_path)
```
</blockquote></details>

Good catch. This was working correctly by accident. `path` ends up being a variable from closure (see [LOAD_DEREF](https://docs.python.org/3/library/dis.html#opcode-LOAD_DEREF) in line 172 in `dis` below), and at the time of the call, it is set in outer function, in `for tf in self['sgx']['trusted_files']`, as `path = uri2path(uri)`, about in the middle of the function.

Because of this bug, and ambiguity of those closures, I've rewritten all of trusted files measurement. Now every function fits one screen, doesn't rely on closured (instead uses explicit OOP) and I hope is generally easier to reason about.

Here's a disassembly for reference:

```
Disassembly of <code object append_tf at 0x7fecfa50ecd0, file "/usr/lib/python3/dist-packages/graminelibos/manifest.py", line 171>:
              0 COPY_FREE_VARS           3

171           2 RESUME                   0

172           4 LOAD_DEREF               2 (path)
              6 LOAD_DEREF               3 (trusted_files)
              8 CONTAINS_OP              1
             10 POP_JUMP_FORWARD_IF_FALSE    29 (to 70)

173          12 LOAD_GLOBAL              1 (NULL + hash_file_contents)
             24 LOAD_FAST                0 (chroot_path)
             26 PRECALL                  1
             30 CALL                     1
             40 LOAD_DEREF               3 (trusted_files)
             42 PUSH_NULL
             44 LOAD_DEREF               1 (out_chroot)
             46 LOAD_FAST                0 (chroot_path)
             48 PRECALL                  1
             52 CALL                     1
             62 STORE_SUBSCR
             66 LOAD_CONST               0 (None)
             68 RETURN_VALUE

172     >>   70 LOAD_CONST               0 (None)
             72 RETURN_VALUE
```

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1584)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @woju)

___
*[`python/graminelibos/manifest.py` line 146 at r1](https://reviewable.io/reviews/gramineproject/gramine/1584#-NfoRTmY5i-tnCpk1dVC:-Nfqi0W86jZw2DDHvKXl:b-30rrsr) ([raw file](https://github.com/gramineproject/gramine/blob/b1b03eb6f95b0be0ea7fc1f1dfec4b267ce37b23/python/graminelibos/manifest.py#L146)):*

I have changed this to list only the files themselves, as included in the manifest. Is this still OK, and if so, how do I reword this comment?
</blockquote></details>

Yes, this is ok to me. And it is used as a dep input, so looks legit to only have the files themselves, not directories.

Why not `Returns a list of all expanded files, as included in the manifest` ?
___
*[`python/graminelibos/manifest.py` line 173 at r1](https://reviewable.io/reviews/gramineproject/gramine/1584#-NfoPDgj7sOJHlUU4FDb:-NfqhYqWAzj8fWNSEevu:b-x5hyxb) ([raw file](https://github.com/gramineproject/gramine/blob/b1b03eb6f95b0be0ea7fc1f1dfec4b267ce37b23/python/graminelibos/manifest.py#L173)):*

Good catch. This was working correctly by accident. `path` ends up being a variable from closure (see [LOAD_DEREF](https://docs.python.org/3/library/dis.html#opcode-LOAD_DEREF) in line 172 in `dis` below), and at the time of the call, it is set in outer function, in `for tf in self['sgx']['trusted_files']`, as `path = uri2path(uri)`, about in the middle of the function.

Because of this bug, and ambiguity of those closures, I've rewritten all of trusted files measurement. Now every function fits one screen, doesn't rely on closured (instead uses explicit OOP) and I hope is generally easier to reason about.

Here's a disassembly for reference:

```
Disassembly of <code object append_tf at 0x7fecfa50ecd0, file "/usr/lib/python3/dist-packages/graminelibos/manifest.py", line 171>:
              0 COPY_FREE_VARS           3

171           2 RESUME                   0

172           4 LOAD_DEREF               2 (path)
              6 LOAD_DEREF               3 (trusted_files)
              8 CONTAINS_OP              1
             10 POP_JUMP_FORWARD_IF_FALSE    29 (to 70)

173          12 LOAD_GLOBAL              1 (NULL + hash_file_contents)
             24 LOAD_FAST                0 (chroot_path)
             26 PRECALL                  1
             30 CALL                     1
             40 LOAD_DEREF               3 (trusted_files)
             42 PUSH_NULL
             44 LOAD_DEREF               1 (out_chroot)
             46 LOAD_FAST                0 (chroot_path)
             48 PRECALL                  1
             52 CALL                     1
             62 STORE_SUBSCR
             66 LOAD_CONST               0 (None)
             68 RETURN_VALUE

172     >>   70 LOAD_CONST               0 (None)
             72 RETURN_VALUE
```
</blockquote></details>

Looks good to me. I'm not sure how the Python disassembly helps in reviewing :)
___
*[`python/graminelibos/manifest.py` line 105 at r2](https://reviewable.io/reviews/gramineproject/gramine/1584#-Nfqeqj05VEm9vEms7rX:-Nfqeqj05VEm9vEms7rY:b-5e7msj) ([raw file](https://github.com/gramineproject/gramine/blob/dad028217a0a435311b98577ae6eda2e06152e9a/python/graminelibos/manifest.py#L105)):*
> ```Python
> 
>         Returns:
>             TrustedFile: a single instance of TrustedFile
> ```

Why the comment doesn't have the `Raises:` part?
___
*[`python/graminelibos/manifest.py` line 118 at r2](https://reviewable.io/reviews/gramineproject/gramine/1584#-NfqfaUEDIYMshKn2S28:-NfqfaUEDIYMshKn2S29:b-d242bu) ([raw file](https://github.com/gramineproject/gramine/blob/dad028217a0a435311b98577ae6eda2e06152e9a/python/graminelibos/manifest.py#L118)):*
> ```Python
>     def __repr__(self):
>         return (f'{type(self).__name__}('
>                 f'uri={self.uri!r}, sha256={self.sha256!r}, chroot={self.chroot!r})')
> ```

IIUC, this is the replicate the existing object.

But is this used anywhere?
___
*[`python/graminelibos/manifest.py` line 122 at r2](https://reviewable.io/reviews/gramineproject/gramine/1584#-Nfqfh_h6kHytdMjFVW3:-Nfqfh_h6kHytdMjFVW4:btylzwu) ([raw file](https://github.com/gramineproject/gramine/blob/dad028217a0a435311b98577ae6eda2e06152e9a/python/graminelibos/manifest.py#L122)):*
> ```Python
> 
>     def to_manifest(self):
>         """Returns the representation of the current file for manifest
> ```

Please add a dot at the end of the sentence.
___
*[`python/graminelibos/manifest.py` line 135 at r2](https://reviewable.io/reviews/gramineproject/gramine/1584#-NfqfxST7PEnOeizQs42:-NfqfxST7PEnOeizQs43:b-y9iyk9) ([raw file](https://github.com/gramineproject/gramine/blob/dad028217a0a435311b98577ae6eda2e06152e9a/python/graminelibos/manifest.py#L135)):*
> ```Python
> 
> 
>     def ensure_hash(self):
> ```

This name is weird. Maybe `calculate_hash`? Or `calculate_hash_if_none`?

I mean, 99% of the time this function will actually compute the hash, not just ensure that it exists.
___
*[`python/graminelibos/manifest.py` line 311 at r2](https://reviewable.io/reviews/gramineproject/gramine/1584#-Nfqcwhz9m5bhj9LiBfw:-Nfqcwhz9m5bhj9LiBfx:b-3s1c37) ([raw file](https://github.com/gramineproject/gramine/blob/dad028217a0a435311b98577ae6eda2e06152e9a/python/graminelibos/manifest.py#L311)):*
> ```Python
> 
>         Raises:
>             graminelibos.ManifestError: There was an error with the format of some trusted files in the manifest
> ```

Doesn't fit into 100 chars per line now
___
*[`python/graminelibos/manifest.py` line 320 at r2](https://reviewable.io/reviews/gramineproject/gramine/1584#-NfqddDY8mn8X07Qjxf6:-NfqddDY8mn8X07Qjxf7:b-auxy2x) ([raw file](https://github.com/gramineproject/gramine/blob/dad028217a0a435311b98577ae6eda2e06152e9a/python/graminelibos/manifest.py#L320)):*
> ```Python
>                 if tf.uri in trusted_files:
>                     # On duplicate entries, pick the one that is already measured, preferring
>                     # existing one to avoid dict insertion. Accept double (matching) sha256
> ```

Don't you need a comma before `to avoid dict insertion`?
___
*[`python/graminelibos/manifest.py` line 326 at r2](https://reviewable.io/reviews/gramineproject/gramine/1584#-NfqdP-rEIagMAGEpyF0:-NfqdP-rEIagMAGEpyF1:bfx0bzm) ([raw file](https://github.com/gramineproject/gramine/blob/dad028217a0a435311b98577ae6eda2e06152e9a/python/graminelibos/manifest.py#L326)):*
> ```Python
>                         if tf.sha256 is not None and tf.sha256 != tf_old.sha256:
>                             raise ManifestError(
>                                 f'Two different sha256 values ({tf2.sha256} and {tf.sha256}) '
> ```

`tf2`? Should be `tf_old`

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1584)* status: 3 of 4 files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/1584#-NfoO1r20TbSM_wC2mRq):*

I'd prefer to have the command in backticks. Can be done at final rebase.
</blockquote></details>

Done.
___
*[`python/graminelibos/manifest.py` line 146 at r1](https://reviewable.io/reviews/gramineproject/gramine/1584#-NfoRTmY5i-tnCpk1dVC:-NfrTSDs4mU2N7vHMKzB:b-rz4yv2) ([raw file](https://github.com/gramineproject/gramine/blob/b1b03eb6f95b0be0ea7fc1f1dfec4b267ce37b23/python/graminelibos/manifest.py#L146)):*

Yes, this is ok to me. And it is used as a dep input, so looks legit to only have the files themselves, not directories.

Why not `Returns a list of all expanded files, as included in the manifest` ?
</blockquote></details>

Done. Just to be sure: what happens if you have say `trusted_files = ["file:/a/"]`, and directories `/a`, `/a/b`, and a file `/a/b/c`. You measure, you end up with `trusted_files = [{uri = "file:/a/b/c", sha256 = "..."}]` and makedep `manifest.sgx: /a/b/c`. Then you add file `/a/b/d`, which on the next `gramine-sgx-sign` will end up in trusted_files, but the directory `/a/b` is not part of dependencies anymore (it was before this change). Does that change anything?
___
*[`python/graminelibos/manifest.py` line 173 at r1](https://reviewable.io/reviews/gramineproject/gramine/1584#-NfoPDgj7sOJHlUU4FDb:-NfrUc3H5ndDjNWFgKGF:bdc83go) ([raw file](https://github.com/gramineproject/gramine/blob/b1b03eb6f95b0be0ea7fc1f1dfec4b267ce37b23/python/graminelibos/manifest.py#L173)):*

Looks good to me. I'm not sure how the Python disassembly helps in reviewing :)
</blockquote></details>

This is about the only sure way to know where a variable value is sourced when evaluating an expression in a python function. Python has multitude of possibilities, cf. [global](https://docs.python.org/3/reference/simple_stmts.html#the-global-statement), [nonlocal](https://docs.python.org/3/reference/simple_stmts.html#the-nonlocal-statement) and this monstrosity: https://docs.python.org/3/reference/datamodel.html#code-objects, which purports to explain various code object attributes that hold them.
___
*[`python/graminelibos/manifest.py` line 105 at r2](https://reviewable.io/reviews/gramineproject/gramine/1584#-Nfqeqj05VEm9vEms7rX:-NfrMJBLClN7pV48OBvk:b-effer) ([raw file](https://github.com/gramineproject/gramine/blob/dad028217a0a435311b98577ae6eda2e06152e9a/python/graminelibos/manifest.py#L105)):*

Why the comment doesn't have the `Raises:` part?
</blockquote></details>

Because I forgot, and nothing really checks this. Now that I think about this, I'm not sure if raising ManifestError would be the correct thing to do, because in normal operation this won't ever be a problem with manifest. I think I'll document this to ValueError, raised by `.relative_to()`, as this will only result from programming errors, not from manifest errors.
___
*[`python/graminelibos/manifest.py` line 118 at r2](https://reviewable.io/reviews/gramineproject/gramine/1584#-NfqfaUEDIYMshKn2S28:-Nfr69j8Al9oc4VTe-CM:b-8qr0z3) ([raw file](https://github.com/gramineproject/gramine/blob/dad028217a0a435311b98577ae6eda2e06152e9a/python/graminelibos/manifest.py#L118)):*

IIUC, this is the replicate the existing object.

But is this used anywhere?
</blockquote></details>

I used it for debugging (let's call them "strategically placed `print(repr(self))` statements"). It doesn't hurt to leave it as is.

It's a textual representation of the instance, to be used in string formatting (not for serialisation). Sometimes reprs are shown as constructor invocations (which this one is), but sometimes they're not, like `<MyClass instance at 0xadd8e55 attr1=123 attr2=456>` and that's up to class author. https://docs.python.org/3/reference/datamodel.html#object.__repr__

In this case I see one degree of freedom: `self.realpath`, which is not part of constructor invocation, but is a relevant attribute of this class. If you think it should be part of this repr, I'll change to `<TrustedFile(uri='file:/a/b/c', sha256='0123', chroot=None) realpath='/a/b/c'>`, to be both copy-paste-able to construct from REPL, and include this realpath.
___
*[`python/graminelibos/manifest.py` line 122 at r2](https://reviewable.io/reviews/gramineproject/gramine/1584#-Nfqfh_h6kHytdMjFVW3:-NfrPhup-a0-1U8NmEpc:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/dad028217a0a435311b98577ae6eda2e06152e9a/python/graminelibos/manifest.py#L122)):*

Please add a dot at the end of the sentence.
</blockquote></details>

Done.
___
*[`python/graminelibos/manifest.py` line 135 at r2](https://reviewable.io/reviews/gramineproject/gramine/1584#-NfqfxST7PEnOeizQs42:-NfrPk4F62Wt2a70F1CN:b-s386r3) ([raw file](https://github.com/gramineproject/gramine/blob/dad028217a0a435311b98577ae6eda2e06152e9a/python/graminelibos/manifest.py#L135)):*

This name is weird. Maybe `calculate_hash`? Or `calculate_hash_if_none`?

I mean, 99% of the time this function will actually compute the hash, not just ensure that it exists.
</blockquote></details>

`calculate_hash` would suggest to recalculate it unconditionally (with option to fail, e.g. if the file is gone from fs, but will end up in the image some other way). And `calculate_hash_if_none` can be shortened to `ensure_hash`: "ensure" is a word used for this behaviour (idempotent fix for something, noop if already done). There are already things in other Python code, like functions: [[1]](https://docs.python.org/3/library/asyncio-future.html#asyncio.ensure_future) [[2]](https://docs.python.org/3/library/venv.html#venv.EnvBuilder.ensure_directories) [[3]](https://docs.python.org/3/library/optparse.html#adding-new-actions), commands: [[4]](https://docs.python.org/3/library/ensurepip.html#module-ensurepip), or even arguments: [[5]](https://docs.python.org/3/library/json.html#json.dump).

IDK about usage, if you did specify some trusted files' hashes manually, it ensures that sha256 is there? I think it might be relevant when you'll be assembling that list from different places (currently not supported by our tooling, but using this API you can do it yourself). Then it will very much ensure that all hashes are OK, some already present.

Usage in `expand_all_trusted_files` follows this pattern.
___
*[`python/graminelibos/manifest.py` line 311 at r2](https://reviewable.io/reviews/gramineproject/gramine/1584#-Nfqcwhz9m5bhj9LiBfw:-NfrUMa95TcdyIaSBfs1:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/dad028217a0a435311b98577ae6eda2e06152e9a/python/graminelibos/manifest.py#L311)):*

Doesn't fit into 100 chars per line now
</blockquote></details>

Done.
___
*[`python/graminelibos/manifest.py` line 320 at r2](https://reviewable.io/reviews/gramineproject/gramine/1584#-NfqddDY8mn8X07Qjxf6:-NfrUXw0BFpSJWFGC8ji:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/dad028217a0a435311b98577ae6eda2e06152e9a/python/graminelibos/manifest.py#L320)):*

Don't you need a comma before `to avoid dict insertion`?
</blockquote></details>

Done.
___
*[`python/graminelibos/manifest.py` line 326 at r2](https://reviewable.io/reviews/gramineproject/gramine/1584#-NfqdP-rEIagMAGEpyF0:-NfrVw_L2bfgwCfhv6-t:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/dad028217a0a435311b98577ae6eda2e06152e9a/python/graminelibos/manifest.py#L326)):*

`tf2`? Should be `tf_old`
</blockquote></details>

Done.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1584)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @woju)

___
*[`python/graminelibos/manifest.py` line 146 at r1](https://reviewable.io/reviews/gramineproject/gramine/1584#-NfoRTmY5i-tnCpk1dVC:-NftP7e906AQ2ivWvcYX:b3jk81h) ([raw file](https://github.com/gramineproject/gramine/blob/b1b03eb6f95b0be0ea7fc1f1dfec4b267ce37b23/python/graminelibos/manifest.py#L146)):*
> Does that change anything?

Well, the re-build will not be triggered in this case, and running `gramine-sgx app` will result in an error "Cannot open this trusted file because it was not specified in the manifest". I think this is benign, as the user will be said what needs to be done.

Also, I'm not even sure if these deps work correctly. I always tend to do a clean re-build of applications/tests, because in many cases the deps are incomplete.
___
*[`python/graminelibos/manifest.py` line 118 at r2](https://reviewable.io/reviews/gramineproject/gramine/1584#-NfqfaUEDIYMshKn2S28:-NftOQx5-qiiMTgAuWhq:bq1lq8o) ([raw file](https://github.com/gramineproject/gramine/blob/dad028217a0a435311b98577ae6eda2e06152e9a/python/graminelibos/manifest.py#L118)):*

I used it for debugging (let's call them "strategically placed `print(repr(self))` statements"). It doesn't hurt to leave it as is.

It's a textual representation of the instance, to be used in string formatting (not for serialisation). Sometimes reprs are shown as constructor invocations (which this one is), but sometimes they're not, like `<MyClass instance at 0xadd8e55 attr1=123 attr2=456>` and that's up to class author. https://docs.python.org/3/reference/datamodel.html#object.__repr__

In this case I see one degree of freedom: `self.realpath`, which is not part of constructor invocation, but is a relevant attribute of this class. If you think it should be part of this repr, I'll change to `<TrustedFile(uri='file:/a/b/c', sha256='0123', chroot=None) realpath='/a/b/c'>`, to be both copy-paste-able to construct from REPL, and include this realpath.
</blockquote></details>

Ah, it was purely for debugging. I like your idea with the format that also has `realpath`, please add.
___
*[`python/graminelibos/manifest.py` line 135 at r2](https://reviewable.io/reviews/gramineproject/gramine/1584#-NfqfxST7PEnOeizQs42:-NftOl6x40qutoBxLeiA:b-1kg8pr) ([raw file](https://github.com/gramineproject/gramine/blob/dad028217a0a435311b98577ae6eda2e06152e9a/python/graminelibos/manifest.py#L135)):*

`calculate_hash` would suggest to recalculate it unconditionally (with option to fail, e.g. if the file is gone from fs, but will end up in the image some other way). And `calculate_hash_if_none` can be shortened to `ensure_hash`: "ensure" is a word used for this behaviour (idempotent fix for something, noop if already done). There are already things in other Python code, like functions: [[1]](https://docs.python.org/3/library/asyncio-future.html#asyncio.ensure_future) [[2]](https://docs.python.org/3/library/venv.html#venv.EnvBuilder.ensure_directories) [[3]](https://docs.python.org/3/library/optparse.html#adding-new-actions), commands: [[4]](https://docs.python.org/3/library/ensurepip.html#module-ensurepip), or even arguments: [[5]](https://docs.python.org/3/library/json.html#json.dump).

IDK about usage, if you did specify some trusted files' hashes manually, it ensures that sha256 is there? I think it might be relevant when you'll be assembling that list from different places (currently not supported by our tooling, but using this API you can do it yourself). Then it will very much ensure that all hashes are OK, some already present.

Usage in `expand_all_trusted_files` follows this pattern.
</blockquote></details>

LGTM, thanks for the references, interesting to see Python terminology.

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1584)* status: 3 of 4 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

___
*[`python/graminelibos/manifest.py` line 118 at r2](https://reviewable.io/reviews/gramineproject/gramine/1584#-NfqfaUEDIYMshKn2S28:-Nftje-B0tIEzvzv7f4D:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/dad028217a0a435311b98577ae6eda2e06152e9a/python/graminelibos/manifest.py#L118)):*

Ah, it was purely for debugging. I like your idea with the format that also has `realpath`, please add.
</blockquote></details>

Done.

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1584)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

kailun-qin (changes requested):
> Reviewed 2 of 3 files at r1, 1 of 2 files at r2, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1584)* status: all files reviewed, 5 unresolved discussions (waiting on @woju)

___
*[`Documentation/manpages/gramine-sgx-sign.rst` line 63 at r4](https://reviewable.io/reviews/gramineproject/gramine/1584#-NfuqTxl4Kb3CHVceTv8:-NfuqTxl4Kb3CHVceTv9:b-67rmj1) ([raw file](https://github.com/gramineproject/gramine/blob/33f2cced009ebddb94aed501326bdbb0694d72fd/Documentation/manpages/gramine-sgx-sign.rst#L63)):*
> ```reStructuredText
>     the directory specified as the value of the option.
> 
>     Note you need to be very careful that the gramine runtime binaries are
> ```

Gramine (to align w/ the other place inside this doc).



_Code quote:_
```reStructuredText
gramine
```
___
*[`Documentation/manpages/gramine-sgx-sign.rst` line 63 at r4](https://reviewable.io/reviews/gramineproject/gramine/1584#-Nfv9MRs-U-CPPMKg3Ag:-Nfv9MRs-U-CPPMKg3Ah:by9wus6) ([raw file](https://github.com/gramineproject/gramine/blob/33f2cced009ebddb94aed501326bdbb0694d72fd/Documentation/manpages/gramine-sgx-sign.rst#L63)):*
> ```reStructuredText
>     the directory specified as the value of the option.
> 
>     Note you need to be very careful that the gramine runtime binaries are
> ```

What if one is careless? Why this note specifically for this option?



_Code quote:_
```reStructuredText
Note you need to be very careful
```
___
*[`python/graminelibos/manifest.py` line 44 at r4](https://reviewable.io/reviews/gramineproject/gramine/1584#-NfuvSUeETghqROxg-G7:-NfuvSUeETghqROxg-G8:btamc2z) ([raw file](https://github.com/gramineproject/gramine/blob/33f2cced009ebddb94aed501326bdbb0694d72fd/python/graminelibos/manifest.py#L44)):*
> ```Python
>         uri (str): URI
>         sha256 (str or None): sha256
>         chroot (pathlib.Path or None): optional path to chroot, if being measured in chroot dir
> ```

Do we want to add `Raises:` section in this comment?
___
*[`python/graminelibos/manifest.py` line 172 at r4](https://reviewable.io/reviews/gramineproject/gramine/1584#-Nfv8Uie6GRkXJK9l3k6:-Nfv8Uie6GRkXJK9l3k7:btamc2z) ([raw file](https://github.com/gramineproject/gramine/blob/33f2cced009ebddb94aed501326bdbb0694d72fd/python/graminelibos/manifest.py#L172)):*
> ```Python
> 
>         Yields:
>             :py:class:`TrustedFile`: one object for each entry in the directory
> ```

Do we want to add `Raises:` section in this comment?
___
*[`python/graminelibos/manifest.py` line 321 at r4](https://reviewable.io/reviews/gramineproject/gramine/1584#-Nfv6YX09s60kvltR4fa:-Nfv6YX09s60kvltR4fb:b-l4gvmq) ([raw file](https://github.com/gramineproject/gramine/blob/33f2cced009ebddb94aed501326bdbb0694d72fd/python/graminelibos/manifest.py#L321)):*
> ```Python
>             for tf in TrustedFile.from_manifest(data, chroot=chroot).expand_directory():
>                 if tf.uri in trusted_files:
>                     # On duplicate entries, pick the one that is already measured, preferring
> ```

Should we also add some notes for the case where none of the duplicate entries is measured?

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1584)* status: 2 of 4 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv, @donporter, and @kailun-qin)

___
*[`Documentation/manpages/gramine-sgx-sign.rst` line 63 at r4](https://reviewable.io/reviews/gramineproject/gramine/1584#-Nfv9MRs-U-CPPMKg3Ag:-NfvCes0-bkDOWZiab_n:bt9ln7e) ([raw file](https://github.com/gramineproject/gramine/blob/33f2cced009ebddb94aed501326bdbb0694d72fd/Documentation/manpages/gramine-sgx-sign.rst#L63)):*

What if one is careless? Why this note specifically for this option?
</blockquote></details>

If one is careless, s/he gets this (`#` is because this is in docker):
```
# gramine-sgx app
error: Enclave initialization IOCTL failed: Operation not permitted (EPERM)
error: Initializing enclave failed: Operation not permitted (EPERM)
error: load_enclave() failed with error: Operation not permitted (EPERM)
```

And one need to figure out what is `-EPERM` on `ioctl(..., SGX_IOC_ENCLAVE_INIT)` on `/dev/sgx_enclave`. https://elixir.bootlin.com/linux/v6.5.5/source/arch/x86/kernel/cpu/sgx/ioctl.c#L588 says it's "Invalid SIGSTRUCT", but to find it, you already need to kinda know where to look. Now what's invalid with this SIGSTRUCT? It's MRENCLAVE, because you loaded different `libsysdb.so` and possibly `runtime/` libs, so hash doesn't match, but this needs to be guessed entirely.

So to answer the first question: because of poor UX on our side.
___
*[`Documentation/manpages/gramine-sgx-sign.rst` line 63 at r4](https://reviewable.io/reviews/gramineproject/gramine/1584#-NfuqTxl4Kb3CHVceTv8:-NfvPTJo8dDuiKrab2gP:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/33f2cced009ebddb94aed501326bdbb0694d72fd/Documentation/manpages/gramine-sgx-sign.rst#L63)):*

Gramine (to align w/ the other place inside this doc).
</blockquote></details>

Done.
___
*[`python/graminelibos/manifest.py` line 44 at r4](https://reviewable.io/reviews/gramineproject/gramine/1584#-NfuvSUeETghqROxg-G7:-NfvHj6E4gBeUYSRHJ6e:ba0g9mv) ([raw file](https://github.com/gramineproject/gramine/blob/33f2cced009ebddb94aed501326bdbb0694d72fd/python/graminelibos/manifest.py#L44)):*

Do we want to add `Raises:` section in this comment?
</blockquote></details>

Hmm, yes. `ManifestError` is raised. Done.
___
*[`python/graminelibos/manifest.py` line 172 at r4](https://reviewable.io/reviews/gramineproject/gramine/1584#-Nfv8Uie6GRkXJK9l3k6:-NfvOEm690rjKk--ZVZq:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/33f2cced009ebddb94aed501326bdbb0694d72fd/python/graminelibos/manifest.py#L172)):*

Do we want to add `Raises:` section in this comment?
</blockquote></details>

Done.
___
*[`python/graminelibos/manifest.py` line 321 at r4](https://reviewable.io/reviews/gramineproject/gramine/1584#-Nfv6YX09s60kvltR4fa:-NfvKyi2EHRNBCevMBb6:bv7xfjq) ([raw file](https://github.com/gramineproject/gramine/blob/33f2cced009ebddb94aed501326bdbb0694d72fd/python/graminelibos/manifest.py#L321)):*

Should we also add some notes for the case where none of the duplicate entries is measured?
</blockquote></details>

It said so already, but I have reworded it to be simpler and more explicit (hopefully more clear). Is this version better?

kailun-qin (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1584)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

___
*[`Documentation/manpages/gramine-sgx-sign.rst` line 63 at r4](https://reviewable.io/reviews/gramineproject/gramine/1584#-Nfv9MRs-U-CPPMKg3Ag:-NfvT0ys7wOL7KP279SC:b-lg05om) ([raw file](https://github.com/gramineproject/gramine/blob/33f2cced009ebddb94aed501326bdbb0694d72fd/Documentation/manpages/gramine-sgx-sign.rst#L63)):*

If one is careless, s/he gets this (`#` is because this is in docker):
```
# gramine-sgx app
error: Enclave initialization IOCTL failed: Operation not permitted (EPERM)
error: Initializing enclave failed: Operation not permitted (EPERM)
error: load_enclave() failed with error: Operation not permitted (EPERM)
```

And one need to figure out what is `-EPERM` on `ioctl(..., SGX_IOC_ENCLAVE_INIT)` on `/dev/sgx_enclave`. https://elixir.bootlin.com/linux/v6.5.5/source/arch/x86/kernel/cpu/sgx/ioctl.c#L588 says it's "Invalid SIGSTRUCT", but to find it, you already need to kinda know where to look. Now what's invalid with this SIGSTRUCT? It's MRENCLAVE, because you loaded different `libsysdb.so` and possibly `runtime/` libs, so hash doesn't match, but this needs to be guessed entirely.

So to answer the first question: because of poor UX on our side.
</blockquote></details>

I see, thanks for the explanation!
___
*[`python/graminelibos/manifest.py` line 321 at r4](https://reviewable.io/reviews/gramineproject/gramine/1584#-Nfv6YX09s60kvltR4fa:-NfvTi-BEJTWtPU-p7a7:b1uxrt8) ([raw file](https://github.com/gramineproject/gramine/blob/33f2cced009ebddb94aed501326bdbb0694d72fd/python/graminelibos/manifest.py#L321)):*

It said so already, but I have reworded it to be simpler and more explicit (hopefully more clear). Is this version better?
</blockquote></details>

Ah indeed (sorry for my poor English). Yes, this is clearer to me, thanks!

dimakuv (approved):
> Reviewed 1 of 2 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1584)* status: :shipit: complete! all files reviewed, all discussions resolved

kailun-qin (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1584)* status: :shipit: complete! all files reviewed, all discussions resolved

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1584)* status: all files reviewed, 1 unresolved discussion (waiting on @woju)

___
*[`-- commits` line 7 at r6](https://reviewable.io/reviews/gramineproject/gramine/1584#-Nh-6FW68jvZ-N0c3xTI):*
> Also rewrite trusted files measurement.

Would it be possible to split this off as a separate PR? Right now it's hard to say what is refactoring and what is the core change here.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1584)* status: all files reviewed, 1 unresolved discussion (waiting on @mkow)

___
*[`-- commits` line 7 at r6](https://reviewable.io/reviews/gramineproject/gramine/1584#-Nh-6FW68jvZ-N0c3xTI):*

> Also rewrite trusted files measurement.

Would it be possible to split this off as a separate PR? Right now it's hard to say what is refactoring and what is the core change here.
</blockquote></details>

+1, agree, this PR ended up modifying the trusted files measurement code heavily.

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1584)* status: all files reviewed, 1 unresolved discussion (waiting on @dimakuv and @mkow)

___
*[`-- commits` line 7 at r6](https://reviewable.io/reviews/gramineproject/gramine/1584#-Nh-6FW68jvZ-N0c3xTI):*

+1, agree, this PR ended up modifying the trusted files measurement code heavily.
</blockquote></details>

Not at this stage I'm afraid. The problem with the older implementation (which relied on closures) was that it wasn't easy to add this chroot option. Best I can do would be to split into two commits, first would rewrite to non-chroot version and the second one would add chroot. This would be much work, which I would instead like to spend on fixing CI.

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1584)* status: all files reviewed, 1 unresolved discussion (waiting on @mkow)

___
*[`-- commits` line 7 at r6](https://reviewable.io/reviews/gramineproject/gramine/1584#-Nh-6FW68jvZ-N0c3xTI):*

Not at this stage I'm afraid. The problem with the older implementation (which relied on closures) was that it wasn't easy to add this chroot option. Best I can do would be to split into two commits, first would rewrite to non-chroot version and the second one would add chroot. This would be much work, which I would instead like to spend on fixing CI.
</blockquote></details>

Ok.

mkow (commented):
> Reviewed 1 of 3 files at r1, 1 of 2 files at r2, 1 of 2 files at r5, 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1584)* status: all files reviewed, 1 unresolved discussion (waiting on @woju)

___
*[`python/graminelibos/manifest.py` line 155 at r6](https://reviewable.io/reviews/gramineproject/gramine/1584#-Nl638e-B1TGbgo3DiLL:-Nl638e-B1TGbgo3DiLM:b-9xutop) ([raw file](https://github.com/gramineproject/gramine/blob/8b5ecdca772002f18b62d42a8fc19439e47fcd18/python/graminelibos/manifest.py#L155)):*
> ```Python
>                 for chunk in iter(lambda: file.read(128 * sha.block_size), b''):
>                     sha.update(chunk)
>                 self.sha256 = sha.hexdigest()
> ```

Could you rename this field to `sha256_hex`? Up until this line I didn't notice it's in hex, not a byte string, I think it's a bit unintuitive. Or maybe store in bytes, and only convert when outputting it?



_Code quote:_
```Python
self.sha256
```

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1584)* status: 3 of 4 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv, @kailun-qin, and @mkow)

___
*[`python/graminelibos/manifest.py` line 155 at r6](https://reviewable.io/reviews/gramineproject/gramine/1584#-Nl638e-B1TGbgo3DiLL:-Nl7ye1D104dZDLfTwla:bpzgnnv) ([raw file](https://github.com/gramineproject/gramine/blob/8b5ecdca772002f18b62d42a8fc19439e47fcd18/python/graminelibos/manifest.py#L155)):*

Could you rename this field to `sha256_hex`? Up until this line I didn't notice it's in hex, not a byte string, I think it's a bit unintuitive. Or maybe store in bytes, and only convert when outputting it?
</blockquote></details>

I would prefer not to, because in manifest (and consequently in the dict that the manifest gets parsed to) the key is also `sha256` and carries hex string. This correspondence seems nice to me. I think there's no use anywhere in Python for digest as `bytes`, so I don't want to decode to bytes on load and encode to hex on save.

I can explicitly document it in attribute doc comment.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1584)* status: all files reviewed, 1 unresolved discussion (waiting on @mkow)

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1584)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

___
*[`python/graminelibos/manifest.py` line 155 at r6](https://reviewable.io/reviews/gramineproject/gramine/1584#-Nl638e-B1TGbgo3DiLL:-Nl9XT1M5J7SucEFRzr2:b5iuwn6) ([raw file](https://github.com/gramineproject/gramine/blob/8b5ecdca772002f18b62d42a8fc19439e47fcd18/python/graminelibos/manifest.py#L155)):*

I would prefer not to, because in manifest (and consequently in the dict that the manifest gets parsed to) the key is also `sha256` and carries hex string. This correspondence seems nice to me. I think there's no use anywhere in Python for digest as `bytes`, so I don't want to decode to bytes on load and encode to hex on save.

I can explicitly document it in attribute doc comment.
</blockquote></details>

Ok, I won't block, but my problem / what I mean by "unintuitiveness" here is that string representation is not the native way of storing bytes.

kailun-qin (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1584)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1584)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

kailun-qin (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1584)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

Issue #1525, opened 2023-08-30
Gramine-Direct Fuzzing: assert failed ../pal/src/pal_object.c:50 !handle_array[i] || handle_array[i]->hdr.type < PAL_HANDLE_TYPE_BOUND


jinengandhi-intel:
### Description of the problem

Syzkaller has been modified in order to run with Gramine (https://github.com/JaewonHur/syzkaller). We were able to run it with Gramine v1.5 for Gramine-Direct and below crash is reported by the tool.

**(pal_object.c:50:PalStreamsWaitEvents) assert failed ../pal/src/pal_object.c:50 !handle_array[i] || handle_array[i]->hdr.type < PAL_HANDLE_TYPE_BOUND**

### Steps to reproduce

This crash was reported 10 times by Syzkaller fuzz tool but we are not able to reproduce the crashes manually. Filing this bug to see if from the logs we can deduce if this is a real bug or not.

Attaching a zip of the crash logs from a few test programs along with their C source files, specifically look at `crash-ccc71c0aa0201433.log` and any other crash file.
[pal_handle_assert.zip](https://github.com/gramineproject/gramine/files/12472429/pal_handle_assert.zip)

### Actual results

...
(libos_parser.c:1628:buf_write_all) [P2:T2:syz-executor] trace: ---- return from futex(...) = 0x0
(libos_parser.c:1628:buf_write_all) [P2:T2:syz-executor] trace: ---- clock_gettime(1, 0x7fe6bfcbe530) = 0x0
(libos_parser.c:1628:buf_write_all) [P2:T2:syz-executor] trace: ---- write(2, 0x7fe6bfcbe3b0, 0x32) ...
(libos_parser.c:1628:buf_write_all) [P2:T2:syz-executor] trace: ---- return from write(...) = 0x32
(libos_parser.c:1628:buf_write_all) [P2:T2:syz-executor] trace: ---- write(2, 0x7fe6bfcbe3b0, 0x3) ...
(libos_parser.c:1628:buf_write_all) [P2:T2:syz-executor] trace: ---- return from write(...) = 0x3
(libos_parser.c:1628:buf_write_all) [P2:T2:syz-executor] trace: ---- write(2, 0x7fe6bfcbe3b0, 0x2) ...
(libos_parser.c:1628:buf_write_all) [P2:T2:syz-executor] trace: ---- return from write(...) = 0x2
[...]

PR #1585, opened 2023-10-03
[LibOS] Defer put_handle on thread exit for handles with async events

stefanberger:
## Description of the changes 

This PR fixes issue #1525 where a PAL_HANDLE was freed even though this PAL_HANDLE had an async event associated with it. Now all PAL_HANDLEs in the thread_exit -> detach_all_fds -> detach_fd path are checked for whether they have an async event associated and if so they are queued for the async worker thread to perform the put_handle on, otherwise the put_handle is called immediately.

## How to test this PR? 

A LibOS regression test case is provided.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1585)

### DIFF
--- a/libos/include/libos_handle.h
+++ b/libos/include/libos_handle.h
@@ -224,7 +224,7 @@ struct libos_handle {
 /* allocating / manage handle */
 struct libos_handle* get_new_handle(void);
 void get_handle(struct libos_handle* hdl);
-void put_handle(struct libos_handle* hdl);
+refcount_t put_handle(struct libos_handle* hdl);
 
 /* Set handle to non-blocking or blocking mode. */
 int set_handle_nonblocking(struct libos_handle* hdl, bool on);
--- a/libos/include/libos_utils.h
+++ b/libos/include/libos_utils.h
@@ -55,6 +55,7 @@ int create_pipe(char* name, char* uri, size_t size, PAL_HANDLE* hdl, bool use_vm
 int init_async_worker(void);
 int64_t install_async_event(PAL_HANDLE object, unsigned long time,
                             void (*callback)(IDTYPE caller, void* arg), void* arg);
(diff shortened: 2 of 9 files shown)


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1585)* status: 0 of 8 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @stefanberger)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1585#-Nndl45w1z2pTRVSGH0O:-Nndl45w1z2pTRVSGH0P:b-k53qd9) (no related file):*
This is wrong. This solves the symptom, not the root cause. See my root cause analysis: https://github.com/gramineproject/gramine/issues/1525#issuecomment-1881309948

In general, a slightly better fix to the problem would be to add correct reference counting to objects that are `FIOASYNC` tracked. But the really correct fix is to re-write that historical crappy code of `FIOASYNC` from scratch.

___

stefanberger:

This is wrong. This solves the symptom, not the root cause. See my root cause analysis: https://github.com/gramineproject/gramine/issues/1525#issuecomment-1881309948

In general, a slightly better fix to the problem would be to add correct reference counting to objects that are `FIOASYNC` tracked. But the really correct fix is to re-write that historical crappy code of `FIOASYNC` from scratch.
</blockquote></details>

It's been a while that I had looked at this. It feels like that the IPC thread must be told to call put_handle on a thread that exits and owned this handle otherwise when will it ever stop looking at it? I don't currently know whether a handle can be shared between different threads so that it may have to be kept around on the list until the put_handle reached 0.

Issue #1520, opened 2023-08-28
Gramine-Direct Fuzzing: IPC leader disconnected 


anjalirai-intel:
### Description of the problem

Syzkaller has been modified in order to run with Gramine (https://github.com/JaewonHur/syzkaller). We were able to run it with Gramine v1.5 for Gramine-Direct

The attached crash does not report any assert or crash in console level, but in debug mode error related to IPC leader disconnected has been observed and then IPC pid release failed

### Steps to reproduce

1. Extract the zip file and cd to the drive
2. Build the Gramine(v1.5) in DEBUG mode
3. make DEBUG=1 CRASH=6682fcdd5ac4cf13
4. gramine-direct crash
[crash-6682fcdd5ac4cf13.zip](https://github.com/gramineproject/gramine/files/12453341/crash-6682fcdd5ac4cf13.zip)

These issues are very rarely reproducible but it is reported by syzcaller fuzzer several times

### Actual results

```
(libos_parser.c:1628:buf_write_all) [P2:T2:syz-executor] trace: ---- getpgid(2) = 0x2
(libos_parser.c:1628:buf_write_all) [P2:T2:syz-executor] trace: ---- clock_gettime(1, 0x648926fe9530) = 0x0
(libos_parser.c:1628:buf_write_all) [P2:T2:syz-executor] trace: ---- write(2, 0x648926fe93b0, 0x18) ...
(libos_parser.c:1628:buf_write_all) [P2:T2:syz-executor] trace: ---- return from write(...) = 0x18
(libos_parser.c:1628:buf_write_all) [P2:T2:syz-executor] trace: ---- write(2, 0x648926fe93b0, 0x1) ...
(libos_parser.c:1628:buf_write_all) [P2:T2:syz-executor] trace: ---- return from write(...) = 0x1
(libos_parser.c:1628:buf_write_all) [P2:T2:syz-executor] trace: ---- futex(0x648926476f4c, FUTEX_PRIVATE|FUTEX_WAKE, 1000000, 0x1, 0, 0) ...
(libos_parser.c:1628:buf_write_all) [P2:T2:syz-executor] trace: ---- return from futex(...) = 0x0
(libos_parser.c:1628:buf_write_all) [P2:T2:syz-executor] trace: ---- write(248, 0x648926fe9550, 0x2c) ...
(libos_parser.c:1628:buf_write_all) [P2:T2:syz-executor] trace: ---- return from write(...) = 0x2c
[...]

Issue #21, opened 2021-09-10
IPC leader does not wait for all processes to finish


boryspoplawski:
## Description of the problem
If the first process, which is the IPC leader, exits, all other processes cannot use IPC any more. We need to decide on a first process exit policy and implement it. I see two options:
- IPC leader kills all other processes before exiting
- IPC leader just cleans up app resources, but stays alive as long as other processes are still running.

We already have a TODO about it:
https://github.com/oscarlab/graphene/blob/4f8d6fb8b323d62f00be2ac2c7239878b14afc1e/LibOS/shim/src/sys/shim_exit.c#L23-L27

jinengandhi-intel:
We continue to see the issue as of Feb 21st. Copied are a few console links from opensource-CI where the errors were seen:
https://localhost:8080/job/graphene-20.04/2237/testReport/LibOS.shim.test.ltp/test_ltp/test_direct___test_ltp_waitpid04_/
https://localhost:8080/job/graphene-20.04/2256/testReport/LibOS.shim.test.ltp/test_ltp/test_direct___test_ltp_waitpid04_/
https://localhost:8080/job/graphene-20.04/2251/testReport/LibOS.shim.test.ltp/test_ltp/test_direct___test_ltp_kill09_/

marcelamelara:
Hello! I wanted to bump this issue since an application I'm trying to run in Gramine (gramine-direct) is failing because the IPC leader exits before its child process is done (see log snippet below). Is there a workaround for this that I could test?

Gramine log snippet:
```
(libos_parser.c:1609:buf_write_all) [P2:T36:Listener] trace: ---- execve("/home/mmelara/app/bin/Worker", [/home/mmelara/app/bin/Worker,spawnclient,121,124,], [DOTNET_SYSTEM_NET_HTTP_USESOCKETSHTTPHANDLER=0,https_proxy=http://proxy:912/,]) ...
(libos_parser.c:1609:buf_write_all) [P1:T35:Listener] trace: ---- return from vfork(...) = 0x24
(libos_async.c:122:install_async_event) [P1:T35:Listener] debug: Installed async event at 1679016125655580
(libos_fs_lock.c:653:posix_lock_clear_pid) [P1:T23:Listener] debug: clearing POSIX locks for pid 1
(libos_async.c:327:libos_async_worker) [P1:libos] debug: Thread exited, cleaning up
[...]

PR #1588, opened 2023-10-03
WIP: [LibOS] Delay IPC leader notification until it has 0 connections

stefanberger:
To avoid killing the IPC leader before all the child processes have died, delay notifying the leader that it should terminate until it has 0 connections.

I think this could fix the following issues (famous last words?)
Fixes: Issue #21 
Fixes: Issue #1519 
Fixes: Issue #1520 

However, I know that this patch may trigger a lot more of these types of messages here:

```
(libos_ipc_child.c:23:ipc_child_disconnect_callback) [P1:libos] debug: Unknown process (vmid: 0x2) disconnected
```

However, these message with 'vmid' are following this type of message with 'pid: 2' and are harmless from what I can see. (matching pid 2 to vmid 2!)
```
(libos_ipc_child.c:63:ipc_cld_exit_callback) [P1:libos] debug: Child process (pid: 2) died
```

I think these messages are the lesser evil and could possibly be removed from the code.

## How to test this PR? 

Use the test cases provided in #1520 or #1519.


mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1588)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @stefanberger)

___
*[`libos/src/ipc/libos_ipc_worker.c` line 112 at r1](https://reviewable.io/reviews/gramineproject/gramine/1588#-Nfyo-rp0RTv8SD1pNVc:-Nfyo-rp0RTv8SD1pNVd:bpf48j0) ([raw file](https://github.com/gramineproject/gramine/blob/7e757816dc35f97d5b0fee67572b9f1be4bb02cb/libos/src/ipc/libos_ipc_worker.c#L112)):*
> ```C
> 
> static void del_ipc_connection(struct libos_ipc_connection* conn,
>                                PAL_HANDLE *notifier) {
> ```

ditto for the whole PR



_Suggestion:_
```C
PAL_HANDLE* notifier
```

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1588)* status: 0 of 1 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @stefanberger)

___
*[`libos/src/ipc/libos_ipc_worker.c` line 440 at r1](https://reviewable.io/reviews/gramineproject/gramine/1588#-NfyomVn0wtcopby1VGv:-NfyomVn0wtcopby1VGw:by38fjw) ([raw file](https://github.com/gramineproject/gramine/blob/7e757816dc35f97d5b0fee67572b9f1be4bb02cb/libos/src/ipc/libos_ipc_worker.c#L440)):*
> ```C
>         while (__atomic_load_n(&g_ipc_connections_cnt, __ATOMIC_ACQUIRE) > 0) {
>             PalEventWait(leader_notifier, &timeout_us);
>             if (!__atomic_load_n(&g_clear_on_worker_exit, __ATOMIC_ACQUIRE))
> ```

How can this if ever happen? Worker thread is informed that it should exit only 4 lines below (`set_pollable_event`)

stefanberger (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1588)* status: 0 of 1 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[`libos/src/ipc/libos_ipc_worker.c` line 112 at r1](https://reviewable.io/reviews/gramineproject/gramine/1588#-Nfyo-rp0RTv8SD1pNVc:-NfzpgpdAub96Kc9bcYy:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/7e757816dc35f97d5b0fee67572b9f1be4bb02cb/libos/src/ipc/libos_ipc_worker.c#L112)):*

ditto for the whole PR
</blockquote></details>

Done.
___
*[`libos/src/ipc/libos_ipc_worker.c` line 440 at r1](https://reviewable.io/reviews/gramineproject/gramine/1588#-NfyomVn0wtcopby1VGv:-NfzpigWEyBJDByxE6f4:b2wbray) ([raw file](https://github.com/gramineproject/gramine/blob/7e757816dc35f97d5b0fee67572b9f1be4bb02cb/libos/src/ipc/libos_ipc_worker.c#L440)):*

How can this if ever happen? Worker thread is informed that it should exit only 4 lines below (`set_pollable_event`)
</blockquote></details>

What is 'this' -- you mean 'this' is 'the problem' that this patch is solving? It can happen when the initial process (pid 1) exits before its children.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1588)* status: 0 of 1 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @mkow and @stefanberger)

___
*[`libos/src/ipc/libos_ipc_worker.c` line 440 at r1](https://reviewable.io/reviews/gramineproject/gramine/1588#-NfyomVn0wtcopby1VGv:-Nfzt15CCZEnnL1cZBDE:birmr7q) ([raw file](https://github.com/gramineproject/gramine/blob/7e757816dc35f97d5b0fee67572b9f1be4bb02cb/libos/src/ipc/libos_ipc_worker.c#L440)):*

What is 'this' -- you mean 'this' is 'the problem' that this patch is solving? It can happen when the initial process (pid 1) exits before its children.
</blockquote></details>

I meant how can this `if` statement ever be `true`

stefanberger (commented):
The iret_emulation test had caused and issue that I had not seen before. I now squashed the patches (to better see the code) and fixed this by initializing the notifier handle earlier.

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1588#-:-Ng-2-I03S6X5UISPELZ:bp60l7q)* status: 0 of 1 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski and @mkow)

stefanberger (commented):
> Now there's an issue in ltp test sched\_setaffinity01 where one child is stuck in the pause() system call. This is a test case that is not expected to work per ltp.cfg but we need it to work for the number of ipc clients to decrease to zero.

An implementation of setresuid() supporting setting `euid` would resolve the issue.

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1588#-:-Ng-G-oo3ojOPbeYhj63:b18qx04)* status: 0 of 1 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski and @mkow)

stefanberger (commented):
This PR is blocked by PR #1590 implementing setresuid() support.

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1588#-:-Ng4Kv535DCGertPSvvA:b-po9qdv)* status: 0 of 1 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski and @mkow)

stefanberger (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1588)* status: all files reviewed, 13 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

___
*[`libos/src/ipc/libos_ipc_worker.c` line 68 at r4](https://reviewable.io/reviews/gramineproject/gramine/1588#-NndZ_Z5CfK2db_RucFw:-NndikFYCMhw5i9qdGe7:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/284118e8eaa9b74ff97e197765b19b4c953e2468/libos/src/ipc/libos_ipc_worker.c#L68)):*

All global-scope variables must be prepended with `g_`. This includes even `static` keyword variables.
</blockquote></details>

Done.
___
*[`libos/src/ipc/libos_ipc_worker.c` line 77 at r4](https://reviewable.io/reviews/gramineproject/gramine/1588#-NndZnmPBX86vwcZ8zM4:-Nndiqvp33Azq3kUBsBT:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/284118e8eaa9b74ff97e197765b19b4c953e2468/libos/src/ipc/libos_ipc_worker.c#L77)):*

`bool`
</blockquote></details>

Done.
___
*[`libos/src/ipc/libos_ipc_worker.c` line 116 at r4](https://reviewable.io/reviews/gramineproject/gramine/1588#-NndZhiD1YxCw1TmrxpT:-NndkCxu5MgMO6Kw2s3t:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/284118e8eaa9b74ff97e197765b19b4c953e2468/libos/src/ipc/libos_ipc_worker.c#L116)):*

Why do you need this `notifier` argument?
- You are either the leader (`is_ipc_leader() == true`), then you know that the event is `g_leader_notifier`
- Or you are the child (`is_ipc_leader() == false`), then you know that there's no event so you never call `PalEventSet()`
</blockquote></details>

Done.
___
*[`libos/src/ipc/libos_ipc_worker.c` line 125 at r4](https://reviewable.io/reviews/gramineproject/gramine/1588#-Nndb5c_86AJAFHvrvjJ:-NndpJLg5iD_BGAGL95N:b20pf5v) ([raw file](https://github.com/gramineproject/gramine/blob/284118e8eaa9b74ff97e197765b19b4c953e2468/libos/src/ipc/libos_ipc_worker.c#L125)):*

This feels wrong. What prevents the compiler/CPU to reorder `PalEventSet()` and `g_ipc_connections_cnt`? I mean, this func can execute on CPU0 and the waiting func (`terminate_ipc_worker()`) can execute on CPU1. Isn't it possible that CPU1 will notice the event and wake up, but won't notice `g_ipc_connections_cnt` being updated?

I feel like accessing `g_ipc_connections_cnt` should be either made atomic or surrounded by a spinlock. I would prefer a spinlock.
</blockquote></details>

g_ipc_connections_cnt is incremented/decremented by add/del_ipc_connection, which are both called by the ipc_worker_main(). Besides that it's read then only in terminate_ipc_worker where I read it like this:

```
    if (is_ipc_leader()) {
        PalEventClear(g_leader_notifier);
        while (__atomic_load_n(&g_ipc_connections_cnt, __ATOMIC_ACQUIRE) > 0 &&
               PalEventWait(g_leader_notifier, NULL) < 0);
    }
    set_pollable_event(&g_worker_thread->pollable_event);

```

So I thought that making it atomic or surrounding it with a spinlock wouldn't help much.
___
*[`libos/src/ipc/libos_ipc_worker.c` line 226 at r4](https://reviewable.io/reviews/gramineproject/gramine/1588#-Nnd_-9v-YS8EoChtDAU:-Nndmn9XEeZaEdqrg2Ja:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/284118e8eaa9b74ff97e197765b19b4c953e2468/libos/src/ipc/libos_ipc_worker.c#L226)):*

ditto (no need for `notifier` arg)
</blockquote></details>

Done.
___
*[`libos/src/ipc/libos_ipc_worker.c` line 373 at r4](https://reviewable.io/reviews/gramineproject/gramine/1588#-Nnd_sr4EkAhEeAfM-cX:-NndnvpcCWSoukET08Ap:b-hqvnrb) ([raw file](https://github.com/gramineproject/gramine/blob/284118e8eaa9b74ff97e197765b19b4c953e2468/libos/src/ipc/libos_ipc_worker.c#L373)):*

What's the point in these lines? We'll perform `PalProcessExit(1)` anyway, which is basically "die process, die immediately, I don't care about anything".
</blockquote></details>

Done. Removed. 
___
*[`libos/src/ipc/libos_ipc_worker.c` line 389 at r4](https://reviewable.io/reviews/gramineproject/gramine/1588#-Nnd__Ri2-YB9xnMZcj_:-Nndo7Uc6CYRO1eSsxyH:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/284118e8eaa9b74ff97e197765b19b4c953e2468/libos/src/ipc/libos_ipc_worker.c#L389)):*

This line won't be needed if you'll just do it more explicitly, as I described in another comment.
</blockquote></details>

Done.
___
*[`libos/src/ipc/libos_ipc_worker.c` line 410 at r4](https://reviewable.io/reviews/gramineproject/gramine/1588#-NndaoqjBdzXBFk04fSw:-NndooRs8k-U2RZH8Nah:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/284118e8eaa9b74ff97e197765b19b4c953e2468/libos/src/ipc/libos_ipc_worker.c#L410)):*

This should actually be the result of `PalEventCreate()` converted from PAL error code to UNIX error code, not a hardcoded ENOMEM.
</blockquote></details>

Done.
___
*[`libos/src/ipc/libos_ipc_worker.c` line 437 at r4](https://reviewable.io/reviews/gramineproject/gramine/1588#-NndaZyWCB3abvm89qn0:-NndoyVD-VDCFCbj8XRq:b38ybxj) ([raw file](https://github.com/gramineproject/gramine/blob/284118e8eaa9b74ff97e197765b19b4c953e2468/libos/src/ipc/libos_ipc_worker.c#L437)):*

Why is this line needed?
</blockquote></details>

In case the g_ipc_connections_cnt had dropped down to 0 before at some point, and the event was set agt that point, but now g_ipc_connection_cnt is > 0.

stefanberger (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1588)* status: all files reviewed, 13 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

___
*[`libos/src/ipc/libos_ipc_worker.c` line 408 at r4](https://reviewable.io/reviews/gramineproject/gramine/1588#-NndacsC69WG0udQ0eWC:-NndqLmH60enVdoAkv7g:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/284118e8eaa9b74ff97e197765b19b4c953e2468/libos/src/ipc/libos_ipc_worker.c#L408)):*

Can you specify what these `false` args are? Using the pattern:
```
PalEventCreate(&leader_notifier, /*argname=*/false, /*argname=*/false);
```
</blockquote></details>

Done.

stefanberger:

What happens when the leader process (`P0`) is sent a SIGTERM (request to terminate) while its child process (`P1`) is still running?

In normal Linux, P0 should terminate and P1 will terminate with it.

In your PR, it looks like SIGTERM will be ignored? Or, more specifically, P0 will terminate everything **except** the IPC helper thread. So we'll get a weird state when SIGTERM was partially ignored, and Gramine still runs even though the user asked it to die.

(To test this scenario, you need to specify `sys.enable_sigterm_injection = true` in the manifest. Otherwise Gramine processes will not react to SIGTERM at all.)
</blockquote></details>

Thanks, I did not know about this case. I now added a `force` parameter to `terminate_ipc_worker` so that the IPC leader can be forced to be terminated rather than having it wait for 0 connections. It now reacts to `SIGTERM` sent to the 'root' process. I tested this with bash and 2 child bash'es and it behaves as before in this case.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1588)* status: all files reviewed, 14 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski, @mkow, and @stefanberger)

___
*[`-- commits` line 13 at r5](https://reviewable.io/reviews/gramineproject/gramine/1588#-NoWF-m35no2JnZXSDLL):*
We don't use the term "root", we just use the term "main process" or "leader process". I'd prefer "main" here.
___
*[`-- commits` line 14 at r5](https://reviewable.io/reviews/gramineproject/gramine/1588#-NoWF65D4uwgHmksj1m-):*
The "leader" term applies to the process. We have one "main" process who is the parent of all "child" processes. We use the term "leader" to denote that the main process decides on some global-state events on behalf of its child processes.

Next, there is an "IPC worker/helper thread", which is a thread in each of the processes. There is no such thing as the "IPC leader", because this would imply "a leader among threads of the same process", which is incorrect.

So, the correct phrasing is "to terminate the IPC worker thread of the leader process". I would like this more explicit sentence rather than a vague "IPC leader".
___
*[`libos/include/libos_ipc.h` line 67 at r5](https://reviewable.io/reviews/gramineproject/gramine/1588#-NoWFoar8F4z13NA3XFl:-NoWFoar8F4z13NA3XFm:b-9940yw) ([raw file](https://github.com/gramineproject/gramine/blob/76ebbc12865ec0534b0e265bb6f5979f51599a07/libos/include/libos_ipc.h#L67)):*
> ```objc
>  * \brief Terminate the IPC worker thread.
>  *
>  * \param force  Whether to force the termination even if the IPC thread is the leader
> ```

Why not simply `Whether to force the termination of the IPC thread`

Then a curious reader may learn why we don't always force the termination by looking at the source code of this function.
___
*[`libos/src/ipc/libos_ipc_worker.c` line 125 at r4](https://reviewable.io/reviews/gramineproject/gramine/1588#-Nndb5c_86AJAFHvrvjJ:-NoWHfWl9yxq24UWXHPy:b2fhxsv) ([raw file](https://github.com/gramineproject/gramine/blob/284118e8eaa9b74ff97e197765b19b4c953e2468/libos/src/ipc/libos_ipc_worker.c#L125)):*

g_ipc_connections_cnt is incremented/decremented by add/del_ipc_connection, which are both called by the ipc_worker_main(). Besides that it's read then only in terminate_ipc_worker where I read it like this:

```
    if (is_ipc_leader()) {
        PalEventClear(g_leader_notifier);
        while (__atomic_load_n(&g_ipc_connections_cnt, __ATOMIC_ACQUIRE) > 0 &&
               PalEventWait(g_leader_notifier, NULL) < 0);
    }
    set_pollable_event(&g_worker_thread->pollable_event);

```

So I thought that making it atomic or surrounding it with a spinlock wouldn't help much.
</blockquote></details>

TODO(myself): think about it more.
___
*[`libos/src/ipc/libos_ipc_worker.c` line 437 at r4](https://reviewable.io/reviews/gramineproject/gramine/1588#-NndaZyWCB3abvm89qn0:-NoWHiofDf1gYhvPgB6k:b2fhxsv) ([raw file](https://github.com/gramineproject/gramine/blob/284118e8eaa9b74ff97e197765b19b4c953e2468/libos/src/ipc/libos_ipc_worker.c#L437)):*

In case the g_ipc_connections_cnt had dropped down to 0 before at some point, and the event was set agt that point, but now g_ipc_connection_cnt is > 0.
</blockquote></details>

TODO(myself): think about it more.
___
*[`libos/src/ipc/libos_ipc_worker.c` line 401 at r5](https://reviewable.io/reviews/gramineproject/gramine/1588#-NoWGS1vEK7jCU1trvQB:-NoWGS1vEK7jCU1trvQC:brv9ue7) ([raw file](https://github.com/gramineproject/gramine/blob/76ebbc12865ec0534b0e265bb6f5979f51599a07/libos/src/ipc/libos_ipc_worker.c#L401)):*
> ```C
>     /* IPC leader gets a notifier used in terminate_ipc_leader */
>     if (is_ipc_leader() &&
>         (ret = PalEventCreate(&g_leader_notifier, /*init_signaled=*/false, /*auto_clear=*/false)) < 0) {
> ```

Sorry, this now looks ugly. Can we make it more readable by splitting into two embedded IFs?
___
*[`libos/src/ipc/libos_ipc_worker.c` line 403 at r5](https://reviewable.io/reviews/gramineproject/gramine/1588#-NoWGgnf7XPts6bEeqqA:-NoWGgnf7XPts6bEeqqB:b-caiiee) ([raw file](https://github.com/gramineproject/gramine/blob/76ebbc12865ec0534b0e265bb6f5979f51599a07/libos/src/ipc/libos_ipc_worker.c#L403)):*
> ```C
>         (ret = PalEventCreate(&g_leader_notifier, /*init_signaled=*/false, /*auto_clear=*/false)) < 0) {
>         log_error("IPC leader: PalEventCreate() failed");
>         return pal_to_unix_errno(ret);;
> ```

Double `;;` typo
___
*[`libos/src/ipc/libos_ipc_worker.c` line 429 at r5](https://reviewable.io/reviews/gramineproject/gramine/1588#-NoWGouE-E2AdWMnJXwS:-NoWGouE-E2AdWMnJXwT:b-rng93t) ([raw file](https://github.com/gramineproject/gramine/blob/76ebbc12865ec0534b0e265bb6f5979f51599a07/libos/src/ipc/libos_ipc_worker.c#L429)):*
> ```C
> 
> /* Terminate the IPC worker. If 'force' is set, the IPC leader will not wait
>  * until it has 0 connections but terminate immediatel, otherwise it will
> ```

`immediately` typo
___
*[`libos/src/ipc/libos_ipc_worker.c` line 430 at r5](https://reviewable.io/reviews/gramineproject/gramine/1588#-NoWGrSMDZkhWrUZyEf1:-NoWGrSMDZkhWrUZyEf2:bn5a2c5) ([raw file](https://github.com/gramineproject/gramine/blob/76ebbc12865ec0534b0e265bb6f5979f51599a07/libos/src/ipc/libos_ipc_worker.c#L430)):*
> ```C
> /* Terminate the IPC worker. If 'force' is set, the IPC leader will not wait
>  * until it has 0 connections but terminate immediatel, otherwise it will
>  * wait until all child processes have died and there are 0 connections.
> ```

I would also mention that the `force` case is used when the user sends an explicit SIGTERM signal to the Gramine process.
___
*[`libos/src/ipc/libos_ipc_worker.c` line 431 at r5](https://reviewable.io/reviews/gramineproject/gramine/1588#-NoWGl2XCbcRc34nFdis:-NoWGl2XCbcRc34nFdit:b-hq8cnt) ([raw file](https://github.com/gramineproject/gramine/blob/76ebbc12865ec0534b0e265bb6f5979f51599a07/libos/src/ipc/libos_ipc_worker.c#L431)):*
> ```C
>  * until it has 0 connections but terminate immediatel, otherwise it will
>  * wait until all child processes have died and there are 0 connections.
>  */
> ```

We have a 100 chars per line limit, not 80
___
*[`libos/src/sys/libos_exit.c` line 25 at r5](https://reviewable.io/reviews/gramineproject/gramine/1588#-NoWHL785kZvoWYXh8Qv:-NoWHL785kZvoWYXh8Qw:ba49y8) ([raw file](https://github.com/gramineproject/gramine/blob/76ebbc12865ec0534b0e265bb6f5979f51599a07/libos/src/sys/libos_exit.c#L25)):*
> ```C
>      * 1) kill all other Gramine processes
>      * 2) wait for them to exit here, before we terminate the IPC helper
>      */
> ```

Don't you resolve this TODO comment (as described in option 2)?

Please remove this comment and instead add a comment before `terminate_ipc_worker()` call below, explaining the option 2.
___
*[`libos/src/sys/libos_exit.c` line 150 at r5](https://reviewable.io/reviews/gramineproject/gramine/1588#-NoWH8gH3hzQALEBEP0W:-NoWH8gH3hzQALEBEP0X:b-i951yt) ([raw file](https://github.com/gramineproject/gramine/blob/76ebbc12865ec0534b0e265bb6f5979f51599a07/libos/src/sys/libos_exit.c#L150)):*
> ```C
>     /* At this point other threads might be still in the middle of an exit routine, but we don't
>      * care since the below will call `exit_group` eventually. */
>     libos_clean_and_exit(term_signal ? 128 + (term_signal & ~__WCOREDUMP_BIT) : error_code, term_signal);
> ```

Please split into two lines; you're exceeding the 100 chars per line limit.

stefanberger (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1588)* status: 0 of 3 files reviewed, 14 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

___
*[`libos/src/sys/libos_exit.c` line 150 at r5](https://reviewable.io/reviews/gramineproject/gramine/1588#-NoWH8gH3hzQALEBEP0W:-NoWxpDe6SQxVJ7SH4KI:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/76ebbc12865ec0534b0e265bb6f5979f51599a07/libos/src/sys/libos_exit.c#L150)):*

Please split into two lines; you're exceeding the 100 chars per line limit.
</blockquote></details>

Done.

Issue #1515, opened 2023-08-28
Gramine Fuzzing: assert failed IS_ALIGNED_PTR(xstate, LIBOS_XSTATE_ALIGN) at libos_context.c:113


anjalirai-intel:
### Description of the problem

Syzkaller has been modified in order to run with Gramine (https://github.com/JaewonHur/syzkaller). We were able to run it with Gramine v1.5 for Gramine-Direct and below crash is reported by the tool.

**[P2:T2:crash-68b52d938b74b3a5] assert failed ../libos/src/arch/x86_64/libos_context.c:113 IS_ALIGNED_PTR(xstate, LIBOS_XSTATE_ALIGN)**

### Steps to reproduce

1. Extract the zip file and cd to the drive
2. Build the Gramine(v1.5) in DEBUG mode
3. make DEBUG=1 CRASH=68b52d938b74b3a5
4. gramine-direct crash
[crash-68b52d938b74b3a5.zip](https://github.com/gramineproject/gramine/files/12451746/crash-68b52d938b74b3a5.zip)

### Actual results

```
(libos_init.c:505:libos_init) [P2:T2:crash-68b52d938b74b3a5] debug: LibOS initialized
(libos_parser.c:1628:buf_write_all) [P2:T2:crash-68b52d938b74b3a5] trace: ---- set_robust_list(0x381e24056a20, 0x18) = 0x0
(libos_parser.c:1628:buf_write_all) [P2:T2:crash-68b52d938b74b3a5] trace: ---- write(1, 0x381e24270004, 0x12) ...
executing program
(libos_parser.c:1628:buf_write_all) [P2:T2:crash-68b52d938b74b3a5] trace: ---- return from write(...) = 0x12
(libos_parser.c:1628:buf_write_all) [P2:T2:crash-68b52d938b74b3a5] trace: ---- write(2, 0x381e24270017, 0xa) ...
### start
(libos_parser.c:1628:buf_write_all) [P2:T2:crash-68b52d938b74b3a5] trace: ---- return from write(...) = 0xa
(libos_context.c:113:is_xstate_extended) [P2:T2:crash-68b52d938b74b3a5] assert failed ../libos/src/arch/x86_64/libos_context.c:113 IS_ALIGNED_PTR(xstate, LIBOS_XSTATE_ALIGN)
(libos_parser.c:1628:buf_write_all) [P1:T1:crash-68b52d938b74b3a5] trace: ---- return from clone(...) = 0x2
[...]

PR #1589, opened 2023-10-04
[LibOS] Allow any context restore and handle SI_KERNEL

stefanberger:
Fix issue in #1515.

The question with the changes is whether it is more helpful for us to have these asserts in there or to get the correct result. 

It now shows this in debug mode:

```
(libos_signal.c:317:internal_fault) [P2:T2:fail_with_rt_sigreturn] debug: process in kernel: Internal memory fault at 0x00000000 (libsysdb.so+0x9632a (addr = 0x7ffffff4932a), VMID = 2, TID = 2)
(libos_signal.c:58:sighandler_kill) [P2:T2:fail_with_rt_sigreturn] debug: killed by signal 11
```

Here we try to restore some context from address 0x0 and this fails as expected... We should let it restore any context and fail on bad pointers.

With this patch we now also get the correct result in `status` from `wait(...,&status,...);`. It is now `status=139` and without this patch it would be `status = 30`.

## How to test this PR? 

Use the provided test case.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1589)

### DIFF
--- a/libos/src/arch/x86_64/libos_context.c
+++ b/libos/src/arch/x86_64/libos_context.c
@@ -110,7 +110,6 @@ void libos_xstate_save(void* xstate_extended) {
 #endif
 
 __attribute__((used)) static int is_xstate_extended(const struct libos_xstate* xstate) {
-    assert(IS_ALIGNED_PTR(xstate, LIBOS_XSTATE_ALIGN));
 
     if (!g_libos_xsave_enabled) {
         return 0;
--- a/libos/src/bookkeep/libos_signal.c
+++ b/libos/src/bookkeep/libos_signal.c
@@ -313,8 +313,21 @@ static noreturn void internal_fault(const char* errstr, uintptr_t addr, PAL_CONT
     char buf[LOCATION_BUF_SIZE];
     libos_describe_location(ip, buf, sizeof(buf));
 
-    log_error("%s at 0x%08lx (%s, VMID = %u, TID = %u)", errstr, addr, buf,
-              g_process_ipc_ids.self_vmid, tid);
+    if (context_is_libos(context) && !is_internal(get_cur_thread())) {
+        log_debug("process in kernel: %s at 0x%08lx (%s, VMID = %u, TID = %u)", errstr, addr, buf,
+                  g_process_ipc_ids.self_vmid, tid);
+        siginfo_t info = {
+            .si_signo = SIGSEGV,
+            .si_code = SI_KERNEL,
+            .si_addr = (void*)addr,
+        };
+        force_signal(&info);
+        handle_signal(context);
+        __builtin_unreachable();
+    } else {
+        log_error("%s at 0x%08lx (%s, VMID = %u, TID = %u)", errstr, addr, buf,
+                  g_process_ipc_ids.self_vmid, tid);
+    }
 
     DEBUG_BREAK_ON_FAILURE();
     PalProcessExit(1);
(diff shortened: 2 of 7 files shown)


mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1589)* status: 0 of 7 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "WIP" and "fixup! " found in commit messages' one-liners (waiting on @stefanberger)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1589#-Nje3lzY1Rje4AOy8SZk:-Nje3lzZCnVUCe3QgsZ7:b13lm8p) (no related file):*
I'm sorry, but this PR seems completely wrong, both the change itself and the test. The test is just one huge UB (calling a syscall with uninitialized arguments and then asserting the process crashed with SIGSEGV...), and the change itself is just removing an assert which was ok.

___

stefanberger:

I'm sorry, but this PR seems completely wrong, both the change itself and the test. The test is just one huge UB (calling a syscall with uninitialized arguments and then asserting the process crashed with SIGSEGV...), and the change itself is just removing an assert which was ok.
</blockquote></details>

The test at least lead to the same result on the host and on Gramine, so I don't think it's that wrong.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1589)* status: 0 of 7 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "WIP" and "fixup! " found in commit messages' one-liners (waiting on @stefanberger)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1589#-Nje3lzY1Rje4AOy8SZk:-NjgQjzt07ycB_0aMdyw:b-u1jl7c) (no related file):*

The test at least lead to the same result on the host and on Gramine, so I don't think it's that wrong.
</blockquote></details>

It just used uninitialized variables and it accidentally happened that in your configuration it led to the reproduction, but it could also just succeed or fail in a different way, it's completely unreliable. And also the compiler can just remove this call (or actually do anything with it) with optimizations enabled, as it's a UB.

___

stefanberger:

It just used uninitialized variables and it accidentally happened that in your configuration it led to the reproduction, but it could also just succeed or fail in a different way, it's completely unreliable. And also the compiler can just remove this call (or actually do anything with it) with optimizations enabled, as it's a UB.
</blockquote></details>

What speaks for the fix to let the user restore the context from any address is that it results in the same behavior when used on host and with Gramine. It would be the developer's mistake to provide a bad pointer. I understand the part with the test case and uninitialized stack data, though 50000 test runs on the host and 135000 test runs with Gramine lead to the same test result. Never mind, though.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1589)* status: 0 of 7 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "WIP" and "fixup! " found in commit messages' one-liners (waiting on @stefanberger)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1589#-Nje3lzY1Rje4AOy8SZk:-NjiwMEW7iD87qO2ItTI:b-bm7csv) (no related file):*
> What speaks for the fix to let the user restore the context from any address is that it results in the same behavior when used on host and with Gramine. It would be the developer's mistake to provide a bad pointer.

1. At the same time, it causes all internal Gramine crashes (which may result in corrupted state) to send the user app a signal instead of forcefully exiting.
2. We don't care about apps which restore context from garbage - first, this just doesn't work and means the app is terribly buggy, second, we don't know any app which would do that.
3. Officially supporting Gramine crashing inside its own code and then trying to recover the state is just a bad idea overall, we don't want to support any such scenario. If we decided to change anything here it would be to change the assert into a hard failure, so it also fails in release builds (assuming we cared about this scenario, but we don't, as apps don't load garbage contexts).

> I understand the part with the test case and uninitialized stack data, though 50000 test runs on the host and 135000 test runs with Gramine lead to the same test result. Never mind, though.

But you ran it compiled in just one way. Relying on such stuff results in a random GCC update suddenly breaking the test. We don't want to have such tests in our CI.

___

PR #1590, opened 2023-10-05
[LibOS] Implement setresuid and getresuid syscalls

stefanberger:
## Description of the changes 

To resolve an issue found in LTP test sched_setaffinity01 implement the setresuid() syscall that is used by this test case. Otherwise the test case terminates early without terminating the child process that is stuck in the pause() syscall. Even though the test case will then still not pass (sched_setaffinity() cannot find child process via lookup_thead(pid)], it will then properly kill the child process that was created and enable progress in PR #1588 where we have to wait until the last IPC client has terminated. 

Note: The sched_setaffinity01 in PR #1588 is currently stuck due to the child process being stuck in pause(), which holds up the IPC leader from terminating.

## How to test this PR? 

Enable tracing in LTP sched_setaffinity01 to set that setresuid() is working properly.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1590)


stefanberger (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1590)* status: 0 of 6 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: )

___
*[`libos/src/fs/libos_namei.c` line 35 at r1](https://reviewable.io/reviews/gramineproject/gramine/1590#-Ng4LJmR7q5ez91H0TR7:-Ng4LJmR7q5ez91H0TR8:bchgr2d) ([raw file](https://github.com/gramineproject/gramine/blob/17b2f6a67cb50bb9807d290f0d2d493189e203de/libos/src/fs/libos_namei.c#L35)):*
> ```C
>         /* fall back to 'other' */
>         shift = 0;
>     }
> ```

If this part here is too controversal then I'd take it out and mark one of the test cases as expecting to fail. Though the existing implementation of setuid() would trigger the issue as well since a process that dropped privileges can open a file that it shouldn't be allowed to open.

dimakuv:
It was decided to close this PR as we haven't observed any real world applications that would rely on `setresuid` and `getresuid`. Given the complexity of implementing these syscalls (even if they are mocked), and no apps using them, let's just close.

PR #1600, opened 2023-10-10
Optimize the buffer allocation in libos_syscall_sendfile for NGINX

TejaswineeL:
Optimizing libos_syscall_sendfile by statically allocating buffer. This removes the malloc bottleneck and improves performance of  sendfile heavy workloads like NGINX by 20% in gramine-direct.

The PR can be tested by setting sendfile to 'on' in 'http' section in CI-Examples/nginx/nginx-gramine.conf.template and then executing nginx.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1600)

### DIFF
--- a/libos/src/sys/libos_file.c
+++ b/libos/src/sys/libos_file.c
@@ -410,7 +410,8 @@ out:
 
 long libos_syscall_sendfile(int out_fd, int in_fd, off_t* offset, size_t count) {
     long ret;
-    char* buf = NULL;
+    /* Statically declared buffer allocated such that it is reused on every sendfile syscall */
+    static char buf[BUF_SIZE];
 
     size_t read_from_in  = 0;
     size_t copied_to_out = 0;
@@ -442,12 +443,6 @@ long libos_syscall_sendfile(int out_fd, int in_fd, off_t* offset, size_t count) 
     /* FIXME: This sendfile() emulation is very simple and not particularly efficient: it reads from
      *        input FD in BUF_SIZE chunks and writes into output FD. Mmap-based emulation may be
      *        more efficient but adds complexity (not all handle types provide mmap callback). */
-    buf = malloc(BUF_SIZE);
-    if (!buf) {
-        ret = -ENOMEM;
-        goto out;
-    }
-
     if (!count) {
         ret = 0;
         goto out;
@@ -533,7 +528,6 @@ out_update:
     }
 
 out:
-    free(buf);
     put_handle(in_hdl);
     put_handle(out_hdl);
     return copied_to_out ? (long)copied_to_out : ret;


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1600)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @TejaswineeL)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1600#-NgOTylj1oWj29oxCZrQ:-NgOTylj1oWj29oxCZrR:bq4bocq) (no related file):*
I like the idea of this PR, and a 20% boost on Nginx is great.

___
___
*[`libos/src/sys/libos_file.c` line 414 at r1](https://reviewable.io/reviews/gramineproject/gramine/1600#-NgOSTGJCt1jx-XfH7h-:-NgOSTGJCt1jx-XfH7h0:b-hmgipz) ([raw file](https://github.com/gramineproject/gramine/blob/5295c1a09adea4a2538278ecf8b5647248a62539/libos/src/sys/libos_file.c#L414)):*
> ```C
>     long ret;
>     /* Statically declared buffer allocated such that it is reused on every sendfile syscall */
>     static char buf[BUF_SIZE];
> ```

This makes this `sendfile()` syscall thread-unsafe.

If two threads will call `sendfile()` simultaneously, then they both will start overwriting the contents of `buf`, leading to corrupted contents.

We must do something about it:
1. Revert to as it was (with malloc) -- clearly not a good idea, as Nginx shows real improvement with this PR.
2. Put this whole sendfile logic under a new global lock -- this will make all `sendfile()` invocations linearized, but then `sendfile()` in multi-threaded apps will suffer (well, maybe no apps do sendfile from multiple threads, I don't know).
3. Add this buffer to thread-local storage (TLS) so that it becomes per-thread -- I think this is a slippery slope, adding memory regions for threads to gain small optimizations.
4. Allocate on the stack, but then `BUF_SIZE` cannot be that big (64KB), we should decrease it to e.g. 8KB to not overflow the stack -- this sounds like a legit option but needs perf experiments.

@kailun-qin @mkow and others What option would we prefer? Or maybe there are other options that I didn't think about?

I'm currently leaning towards 4. In the worst case, towards 2.

stefanberger (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1600)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin and @mkow)

___
*[`libos/src/sys/libos_file.c` line 414 at r1](https://reviewable.io/reviews/gramineproject/gramine/1600#-NgOSTGJCt1jx-XfH7h-:-NgQ5O8H8h7OgtfWViG-:b-tyscpu) ([raw file](https://github.com/gramineproject/gramine/blob/5295c1a09adea4a2538278ecf8b5647248a62539/libos/src/sys/libos_file.c#L414)):*

This makes this `sendfile()` syscall thread-unsafe.

If two threads will call `sendfile()` simultaneously, then they both will start overwriting the contents of `buf`, leading to corrupted contents.

We must do something about it:
1. Revert to as it was (with malloc) -- clearly not a good idea, as Nginx shows real improvement with this PR.
2. Put this whole sendfile logic under a new global lock -- this will make all `sendfile()` invocations linearized, but then `sendfile()` in multi-threaded apps will suffer (well, maybe no apps do sendfile from multiple threads, I don't know).
3. Add this buffer to thread-local storage (TLS) so that it becomes per-thread -- I think this is a slippery slope, adding memory regions for threads to gain small optimizations.
4. Allocate on the stack, but then `BUF_SIZE` cannot be that big (64KB), we should decrease it to e.g. 8KB to not overflow the stack -- this sounds like a legit option but needs perf experiments.

@kailun-qin @mkow and others What option would we prefer? Or maybe there are other options that I didn't think about?

I'm currently leaning towards 4. In the worst case, towards 2.
</blockquote></details>

5. This seems to work for me: char *buf = __alloca(count < BUF_SIZE ? count : BUF_SIZE);  [ disable the free(buf) but leave at BUF_SIZE at 64kb)

stefanberger (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1600)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin and @mkow)

___
*[`libos/src/sys/libos_file.c` line 414 at r1](https://reviewable.io/reviews/gramineproject/gramine/1600#-NgOSTGJCt1jx-XfH7h-:-NgQLraV6XdkH879Zrnh:b44q6un) ([raw file](https://github.com/gramineproject/gramine/blob/5295c1a09adea4a2538278ecf8b5647248a62539/libos/src/sys/libos_file.c#L414)):*

5. This seems to work for me: char *buf = __alloca(count < BUF_SIZE ? count : BUF_SIZE);  [ disable the free(buf) but leave at BUF_SIZE at 64kb)
</blockquote></details>

Here are some numbers (x86_64 VM with 128 vCPUs):

```

../common_tools/benchmark-http.sh http://127.0.0.1:8002   [ without wrk's -R option since missing ]



BUF_SIZE = 4KB   && char *buf = __alloca(count < BUF_SIZE ? count : BUF_SIZE);

Concurrency =   1: Per Thread Median Througput (bytes) = 41520.000, Latency (ms) =     3.910
Concurrency =   2: Per Thread Median Througput (bytes) = 34700.000, Latency (ms) =     4.360
Concurrency =   4: Per Thread Median Througput (bytes) = 14590.000, Latency (ms) =     5.170
Concurrency =   8: Per Thread Median Througput (bytes) =  7590.000, Latency (ms) =     4.920
Concurrency =  16: Per Thread Median Througput (bytes) =  3610.000, Latency (ms) =     5.080
Concurrency =  32: Per Thread Median Througput (bytes) =  1780.000, Latency (ms) =     5.140
Concurrency =  64: Per Thread Median Througput (bytes) =   880.000, Latency (ms) =     4.620
Concurrency = 128: Per Thread Median Througput (bytes) =   438.230, Latency (ms) =     4.610
Concurrency = 256: Per Thread Median Througput (bytes) =   216.820, Latency (ms) =     4.640


BUF_SIZE = 8KB   && char *buf = __alloca(count < BUF_SIZE ? count : BUF_SIZE);

Concurrency =   1: Per Thread Median Througput (bytes) = 43980.000, Latency (ms) =     4.000
Concurrency =   2: Per Thread Median Througput (bytes) = 37490.000, Latency (ms) =     4.000
Concurrency =   4: Per Thread Median Througput (bytes) = 17730.000, Latency (ms) =     4.250
Concurrency =   8: Per Thread Median Througput (bytes) =  9280.000, Latency (ms) =     4.030
Concurrency =  16: Per Thread Median Througput (bytes) =  4510.000, Latency (ms) =     4.050
Concurrency =  32: Per Thread Median Througput (bytes) =  2160.000, Latency (ms) =     4.230
Concurrency =  64: Per Thread Median Througput (bytes) =  1030.000, Latency (ms) =     3.940
Concurrency = 128: Per Thread Median Througput (bytes) =   511.610, Latency (ms) =     3.940
Concurrency = 256: Per Thread Median Througput (bytes) =   252.540, Latency (ms) =     3.980



BUF_SIZE = 16KB   && char *buf = __alloca(count < BUF_SIZE ? count : BUF_SIZE);

Concurrency =   1: Per Thread Median Througput (bytes) = 49830.000, Latency (ms) =     3.220
Concurrency =   2: Per Thread Median Througput (bytes) = 45730.000, Latency (ms) =     3.290
Concurrency =   4: Per Thread Median Througput (bytes) = 22200.000, Latency (ms) =     3.400
Concurrency =   8: Per Thread Median Througput (bytes) = 10640.000, Latency (ms) =     3.510
Concurrency =  16: Per Thread Median Througput (bytes) =  5260.000, Latency (ms) =     3.470
Concurrency =  32: Per Thread Median Througput (bytes) =  2370.000, Latency (ms) =     3.860
Concurrency =  64: Per Thread Median Througput (bytes) =  1100.000, Latency (ms) =     3.660
Concurrency = 128: Per Thread Median Througput (bytes) =   539.100, Latency (ms) =     3.730
Concurrency = 256: Per Thread Median Througput (bytes) =   269.180, Latency (ms) =     3.730


BUF_SIZE = 64KB   && char *buf = __alloca(count < BUF_SIZE ? count : BUF_SIZE);           <===== BEST

Concurrency =   1: Per Thread Median Througput (bytes) = 50920.000, Latency (ms) =     3.320
Concurrency =   2: Per Thread Median Througput (bytes) = 46790.000, Latency (ms) =     3.220
Concurrency =   4: Per Thread Median Througput (bytes) = 21670.000, Latency (ms) =     3.480
Concurrency =   8: Per Thread Median Througput (bytes) = 10380.000, Latency (ms) =     3.610
Concurrency =  16: Per Thread Median Througput (bytes) =  5150.000, Latency (ms) =     3.540
Concurrency =  32: Per Thread Median Througput (bytes) =  2450.000, Latency (ms) =     3.720
Concurrency =  64: Per Thread Median Througput (bytes) =  1150.000, Latency (ms) =     3.490
Concurrency = 128: Per Thread Median Througput (bytes) =   562.790, Latency (ms) =     3.580
Concurrency = 256: Per Thread Median Througput (bytes) =   277.680, Latency (ms) =     3.620




BUF_SIZE = 4kb   &&    char buf[BUF_SIZE];

Concurrency =   1: Per Thread Median Througput (bytes) = 42770.000, Latency (ms) =     3.990
Concurrency =   2: Per Thread Median Througput (bytes) = 38250.000, Latency (ms) =     3.940
Concurrency =   4: Per Thread Median Througput (bytes) = 16540.000, Latency (ms) =     4.560
Concurrency =   8: Per Thread Median Througput (bytes) =  7770.000, Latency (ms) =     4.830
Concurrency =  16: Per Thread Median Througput (bytes) =  4020.000, Latency (ms) =     4.560
Concurrency =  32: Per Thread Median Througput (bytes) =  1990.000, Latency (ms) =     4.610
Concurrency =  64: Per Thread Median Througput (bytes) =   940.000, Latency (ms) =     4.290
Concurrency = 128: Per Thread Median Througput (bytes) =   473.980, Latency (ms) =     4.250
Concurrency = 256: Per Thread Median Througput (bytes) =   234.110, Latency (ms) =     4.290


BUF_SIZE = 8kb   &&    char buf[BUF_SIZE];

Concurrency =   1: Per Thread Median Througput (bytes) = 46510.000, Latency (ms) =     3.560
Concurrency =   2: Per Thread Median Througput (bytes) = 35990.000, Latency (ms) =     4.170
Concurrency =   4: Per Thread Median Througput (bytes) = 17290.000, Latency (ms) =     4.370
Concurrency =   8: Per Thread Median Througput (bytes) =  8180.000, Latency (ms) =     4.610
Concurrency =  16: Per Thread Median Througput (bytes) =  4100.000, Latency (ms) =     4.460
Concurrency =  32: Per Thread Median Througput (bytes) =  2050.000, Latency (ms) =     4.470
Concurrency =  64: Per Thread Median Througput (bytes) =  1000.000, Latency (ms) =     4.020
Concurrency = 128: Per Thread Median Througput (bytes) =   503.010, Latency (ms) =     4.010
Concurrency = 256: Per Thread Median Througput (bytes) =   247.430, Latency (ms) =     4.060


BUF_SIZE = 16kb   &&    char buf[BUF_SIZE];

Concurrency =   1: Per Thread Median Througput (bytes) = 40130.000, Latency (ms) =     3.890    <---- ???
Concurrency =   2: Per Thread Median Througput (bytes) = 43540.000, Latency (ms) =     3.450
Concurrency =   4: Per Thread Median Througput (bytes) = 18770.000, Latency (ms) =     4.020
Concurrency =   8: Per Thread Median Througput (bytes) =  9040.000, Latency (ms) =     4.130
Concurrency =  16: Per Thread Median Througput (bytes) =  4650.000, Latency (ms) =     3.930
Concurrency =  32: Per Thread Median Througput (bytes) =  2160.000, Latency (ms) =     4.230
Concurrency =  64: Per Thread Median Througput (bytes) =  1070.000, Latency (ms) =     3.780
Concurrency = 128: Per Thread Median Througput (bytes) =   543.130, Latency (ms) =     3.710
Concurrency = 256: Per Thread Median Througput (bytes) =   269.970, Latency (ms) =     3.720

repeat of   BUF_SIZE = 16kb   &&    char buf[BUF_SIZE]:

Concurrency =   1: Per Thread Median Througput (bytes) = 34980.000, Latency (ms) =     4.880    <----- again slow start
Concurrency =   2: Per Thread Median Througput (bytes) = 44120.000, Latency (ms) =     3.330
Concurrency =   4: Per Thread Median Througput (bytes) = 19920.000, Latency (ms) =     3.780
Concurrency =   8: Per Thread Median Througput (bytes) =  9810.000, Latency (ms) =     3.810
Concurrency =  16: Per Thread Median Througput (bytes) =  5140.000, Latency (ms) =     3.550
Concurrency =  32: Per Thread Median Througput (bytes) =  2390.000, Latency (ms) =     3.820
Concurrency =  64: Per Thread Median Througput (bytes) =  1180.000, Latency (ms) =     3.430
Concurrency = 128: Per Thread Median Througput (bytes) =   568.100, Latency (ms) =     3.540
Concurrency = 256: Per Thread Median Througput (bytes) =   283.630, Latency (ms) =     3.540

```

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1600)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @mkow)

___
*[`libos/src/sys/libos_file.c` line 414 at r1](https://reviewable.io/reviews/gramineproject/gramine/1600#-NgOSTGJCt1jx-XfH7h-:-NgRI0wIC6ZMCUg3QOhX:b2woljz) ([raw file](https://github.com/gramineproject/gramine/blob/5295c1a09adea4a2538278ecf8b5647248a62539/libos/src/sys/libos_file.c#L414)):*
> What option would we prefer?

Frankly, I'm not that much into 1, 2 and 3.
4 and 5 are basically the same essential to me. I slightly prefer 5, but `BUF_SIZE` at 64KB is kind of abusing `__alloca()`. And I think we'll need more perf experiments anyway.

What about:
6. to allocate memory on stack (`__alloca()`) if less than x KB (a safer value, also require more perf experiments) is requested, and to use heap (`malloc()`) otherwise.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1600)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @mkow)

___
*[`libos/src/sys/libos_file.c` line 414 at r1](https://reviewable.io/reviews/gramineproject/gramine/1600#-NgOSTGJCt1jx-XfH7h-:-NgSnq6LA0M1a8yLTzxh:b-ircijd) ([raw file](https://github.com/gramineproject/gramine/blob/5295c1a09adea4a2538278ecf8b5647248a62539/libos/src/sys/libos_file.c#L414)):*
> 6. to allocate memory on stack (`__alloca()`) if less than x KB (a safer value, also require more perf experiments) is requested, and to use heap (`malloc()`) otherwise.

I actually like @stefanberger approach (always allocate on the stack, and use the smallest of user-supplied `count` or our `BUF_SIZE`). With the additional requirement of changing `BUF_SIZE` to 8KB or 16KB (to not accidentally overflow the stack).

The `malloc()` approach will definitely be slower, and I see no specific benefit in this. I understand that with malloc we can allocate a larger buffer, which will probably make it faster to `write()` and `read()`, but I doubt that malloc will out-weigh the benefit of a smaller-size `alloca()`.

Ideally we need to make more micro-benchmarks, with all these different variants. But honestly, I doubt anyone has time for this...

TejaswineeL (commented):
https://reviewable.io/reviews/gramineproject/gramine/1600#-NgOSTGJCt1jx-XfH7h-

Team,

I tried option 4 suggested by @dimakuv , 
I see that with 8KB buf_size gramine-direct performance reduces by 12%.

If I set buf_size to 64KB with option 4 suggestion,  I get Internal Memory fault because of the overflow.
I also tried increasing 'stack size' parameter in the manifest upto 2048K with buf_size 64K to avoid the overflowing, but still hit the internal memory fault.

Should we try to get rid of internal memory fault through possible means or
would you suggest to go for option 2- global lock

Kindly, help  with you suggestions.

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1600#-:-NgxQnyz0m4Z5Cq3SEJY:b-qi3016)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @mkow)

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1600)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @TejaswineeL)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/1600#-NgzMJZ92zVP7MkNdJqb):*
double space
___
*[`libos/src/sys/libos_file.c` line 414 at r1](https://reviewable.io/reviews/gramineproject/gramine/1600#-NgOSTGJCt1jx-XfH7h-:-NgzMdvj8qhkNCM-35XO:b-582rm6) ([raw file](https://github.com/gramineproject/gramine/blob/5295c1a09adea4a2538278ecf8b5647248a62539/libos/src/sys/libos_file.c#L414)):*

> 6. to allocate memory on stack (`__alloca()`) if less than x KB (a safer value, also require more perf experiments) is requested, and to use heap (`malloc()`) otherwise.

I actually like @stefanberger approach (always allocate on the stack, and use the smallest of user-supplied `count` or our `BUF_SIZE`). With the additional requirement of changing `BUF_SIZE` to 8KB or 16KB (to not accidentally overflow the stack).

The `malloc()` approach will definitely be slower, and I see no specific benefit in this. I understand that with malloc we can allocate a larger buffer, which will probably make it faster to `write()` and `read()`, but I doubt that malloc will out-weigh the benefit of a smaller-size `alloca()`.

Ideally we need to make more micro-benchmarks, with all these different variants. But honestly, I doubt anyone has time for this...
</blockquote></details>

I prefer alloca with an IF and a malloc() fallback, otherwise for big buffers the app would need to call LibOS many times.
64KB is definitely too much for our small stacks, I'd say 8 KB is max to stay safe.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1600)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @TejaswineeL)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/1600#-NgzMJZ92zVP7MkNdJqb):*

double space
</blockquote></details>

Can be fixed during final rebase

I'm still amazed how Michal finds these double-space instances. My eyes don't notice them.
___
*[`libos/src/sys/libos_file.c` line 414 at r1](https://reviewable.io/reviews/gramineproject/gramine/1600#-NgOSTGJCt1jx-XfH7h-:-Nh0aPtJ7J5QH0yo1ytu:b-m0r9zk) ([raw file](https://github.com/gramineproject/gramine/blob/5295c1a09adea4a2538278ecf8b5647248a62539/libos/src/sys/libos_file.c#L414)):*
Replies to @TejaswineeL (please next time use Reviewable and reply in the corresponding thread):

> I see that with 8KB buf\_size gramine-direct performance reduces by 12%.

The performance **reduces**? So the performance gets worse?

Well, I guess this is expected because you now copy not in 64KB chunks, but in 8KB chunks...

> If I set buf\_size to 64KB with option 4 suggestion, I get Internal Memory fault because of the overflow.  

Yes, this is because the Gramine internal stack is only 28K in size (excluding the stack guard page): https://github.com/gramineproject/gramine/blob/c92355d929bf2df139ad0dbd1fc334b67be74c1a/libos/include/libos_defs.h#L5-L11

> I also tried increasing 'stack size' parameter in the manifest upto 2048K with buf\_size 64K to avoid the overflowing, but still hit the internal memory fault.

Changing the `stack size` parameter in the manifest doesn't help here, because this parameter is for the **application stack**, not for the internal Gramine stack.

Replies to @kailun-qin @mkow @stefanberger:

Looks like we converge on the "small-sized alloca and a current malloc fallback". If I understand correctly, that's what everyone agrees with:
```
#define BUF_SIZE_ON_STACK (8 * 1024)  // or maybe even 4KB
    size_t buf_size = count <= BUF_SIZE_ON_STACK ? count : BUF_SIZE;
    char* buf = count <= BUF_SIZE_ON_STACK ? __alloca(buf_size) : malloc(buf_size);

    /* ... replace all BUF_SIZE with buf_size in the main logic ... */

out:
    if (count > BUF_SIZE_ON_STACK) {
        /* otherwise allocated on the stack and no heap-freeing is required */
        free(buf);
    }
```

By the way, according to @stefanberger experiments with stack-allocated buffer, 16KB seems the best in terms of stack-size vs performance trade-off. However, 16KB is surely too big for our currently small stack sizes (28KB), so if we choose 16KB, we'll need to increase our LibOS internal stack.

Also, by looking at this function `libos_syscall_sendfile()`, I think that it's practically impossible to overflow the stack because there will be no more than like 5-6 stack frames (i.e., the stack will never grow too deep, as there are not too many nested funcs to be called). Just an observation.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1600)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @TejaswineeL)

___
*[`libos/src/sys/libos_file.c` line 414 at r1](https://reviewable.io/reviews/gramineproject/gramine/1600#-NgOSTGJCt1jx-XfH7h-:-Nh2cbL8-2ztJL-DCMYK:b-xuni7r) ([raw file](https://github.com/gramineproject/gramine/blob/5295c1a09adea4a2538278ecf8b5647248a62539/libos/src/sys/libos_file.c#L414)):*

Replies to @TejaswineeL (please next time use Reviewable and reply in the corresponding thread):

> I see that with 8KB buf\_size gramine-direct performance reduces by 12%.

The performance **reduces**? So the performance gets worse?

Well, I guess this is expected because you now copy not in 64KB chunks, but in 8KB chunks...

> If I set buf\_size to 64KB with option 4 suggestion, I get Internal Memory fault because of the overflow.  

Yes, this is because the Gramine internal stack is only 28K in size (excluding the stack guard page): https://github.com/gramineproject/gramine/blob/c92355d929bf2df139ad0dbd1fc334b67be74c1a/libos/include/libos_defs.h#L5-L11

> I also tried increasing 'stack size' parameter in the manifest upto 2048K with buf\_size 64K to avoid the overflowing, but still hit the internal memory fault.

Changing the `stack size` parameter in the manifest doesn't help here, because this parameter is for the **application stack**, not for the internal Gramine stack.

Replies to @kailun-qin @mkow @stefanberger:

Looks like we converge on the "small-sized alloca and a current malloc fallback". If I understand correctly, that's what everyone agrees with:
```
#define BUF_SIZE_ON_STACK (8 * 1024)  // or maybe even 4KB
    size_t buf_size = count <= BUF_SIZE_ON_STACK ? count : BUF_SIZE;
    char* buf = count <= BUF_SIZE_ON_STACK ? __alloca(buf_size) : malloc(buf_size);

    /* ... replace all BUF_SIZE with buf_size in the main logic ... */

out:
    if (count > BUF_SIZE_ON_STACK) {
        /* otherwise allocated on the stack and no heap-freeing is required */
        free(buf);
    }
```

By the way, according to @stefanberger experiments with stack-allocated buffer, 16KB seems the best in terms of stack-size vs performance trade-off. However, 16KB is surely too big for our currently small stack sizes (28KB), so if we choose 16KB, we'll need to increase our LibOS internal stack.

Also, by looking at this function `libos_syscall_sendfile()`, I think that it's practically impossible to overflow the stack because there will be no more than like 5-6 stack frames (i.e., the stack will never grow too deep, as there are not too many nested funcs to be called). Just an observation.
</blockquote></details>

Hmm, maybe I'd even go with 4KB, I thought our stack are bigger. 28K is really not much and allocating 8KB seems risky...

stefanberger (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1600)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @TejaswineeL)

___
*[`libos/src/sys/libos_file.c` line 414 at r1](https://reviewable.io/reviews/gramineproject/gramine/1600#-NgOSTGJCt1jx-XfH7h-:-Nh2obBzCvP9i4mpxDjX:bkz4qn8) ([raw file](https://github.com/gramineproject/gramine/blob/5295c1a09adea4a2538278ecf8b5647248a62539/libos/src/sys/libos_file.c#L414)):*

Hmm, maybe I'd even go with 4KB, I thought our stack are bigger. 28K is really not much and allocating 8KB seems risky...
</blockquote></details>

[[[ To add to (my) confusion: Iirc, __alloca(64*1024) works shown but char buf[64*1024] causes a fault. I thought that libos would extend the stack if it needed more stack due to stepping on the guard page (and would have to do this only once per thread), but maybe if one steps below the guard page the program crashes -- and maybe in this regards __alloca() was more careful with stepping on guard pages consecutively. ]]]

aneessahib (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1600)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @TejaswineeL)

___
*[`libos/src/sys/libos_file.c` line 414 at r1](https://reviewable.io/reviews/gramineproject/gramine/1600#-NgOSTGJCt1jx-XfH7h-:-Nh6ASLw9cc464fcPlCx:b-kj5ugj) ([raw file](https://github.com/gramineproject/gramine/blob/5295c1a09adea4a2538278ecf8b5647248a62539/libos/src/sys/libos_file.c#L414)):*

[[[ To add to (my) confusion: Iirc, __alloca(64*1024) works shown but char buf[64*1024] causes a fault. I thought that libos would extend the stack if it needed more stack due to stepping on the guard page (and would have to do this only once per thread), but maybe if one steps below the guard page the program crashes -- and maybe in this regards __alloca() was more careful with stepping on guard pages consecutively. ]]]
</blockquote></details>

Allocating small chunks on stack and bigger ones on heap does not solve the perf issue for us. Our benchmark is with >10K HTML files (which is not too big these days). How about we allocate on the heap every time, and dedicate a field in  fd handle `struct libos_handle`  (assuming the fd wont be shared with multiple threads, so this make it thread safe?)

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1600)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @TejaswineeL)

___
*[`libos/src/sys/libos_file.c` line 414 at r1](https://reviewable.io/reviews/gramineproject/gramine/1600#-NgOSTGJCt1jx-XfH7h-:-Nh6DICeBKLDxb6at_zQ:belcvsx) ([raw file](https://github.com/gramineproject/gramine/blob/5295c1a09adea4a2538278ecf8b5647248a62539/libos/src/sys/libos_file.c#L414)):*
<details><summary><i>Previously, aneessahib (Anees Sahib) wrote…</i></summary><blockquote>

Allocating small chunks on stack and bigger ones on heap does not solve the perf issue for us. Our benchmark is with >10K HTML files (which is not too big these days). How about we allocate on the heap every time, and dedicate a field in  fd handle `struct libos_handle`  (assuming the fd wont be shared with multiple threads, so this make it thread safe?)
</blockquote></details>

I agree with @aneessahib that our current approach (see my code snippet) won't bring performance benefit for "typical" files (e.g. 10K HTML files, which is pretty small).

I think we need not a **two-tier** approach but a **three-tier** one. I mean like this:
1. If `count` (size of the file) is less than 4KB, allocate `buf` on stack with size `count`
2. If `count` is less than e.g. 256KB, allocate `buf` on stack with size `4KB`
3. Otherwise (`count > 256KB`), fall-back to the current code: allocate `buf` on heap with size `64KB`

aneessahib (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1600)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @TejaswineeL)

___
*[`libos/src/sys/libos_file.c` line 414 at r1](https://reviewable.io/reviews/gramineproject/gramine/1600#-NgOSTGJCt1jx-XfH7h-:-Nh6LrRc8fauUR8ufiAx:b-kn4xwv) ([raw file](https://github.com/gramineproject/gramine/blob/5295c1a09adea4a2538278ecf8b5647248a62539/libos/src/sys/libos_file.c#L414)):*

I agree with @aneessahib that our current approach (see my code snippet) won't bring performance benefit for "typical" files (e.g. 10K HTML files, which is pretty small).

I think we need not a **two-tier** approach but a **three-tier** one. I mean like this:
1. If `count` (size of the file) is less than 4KB, allocate `buf` on stack with size `count`
2. If `count` is less than e.g. 256KB, allocate `buf` on stack with size `4KB`
3. Otherwise (`count > 256KB`), fall-back to the current code: allocate `buf` on heap with size `64KB`
</blockquote></details>

We would still have the perf issue with small 4KB chunks when the ask is 10K. (see @TejaswineeL results above)

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1600)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @TejaswineeL)

___
*[`libos/src/sys/libos_file.c` line 414 at r1](https://reviewable.io/reviews/gramineproject/gramine/1600#-NgOSTGJCt1jx-XfH7h-:-Nh6eUyCAYN11JIYjIld:bmmb731) ([raw file](https://github.com/gramineproject/gramine/blob/5295c1a09adea4a2538278ecf8b5647248a62539/libos/src/sys/libos_file.c#L414)):*
<details><summary><i>Previously, aneessahib (Anees Sahib) wrote…</i></summary><blockquote>

We would still have the perf issue with small 4KB chunks when the ask is 10K. (see @TejaswineeL results above)
</blockquote></details>

@aneessahib But in this case we side-step the overhead of `malloc()` (heap allocation, which is expensive) and instead do `__alloca()` (stack allocation, which is very cheap).

We just need to perf benchmark my suggestion.

TejaswineeL (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1600)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @mkow)

___
*[`libos/src/sys/libos_file.c` line 414 at r1](https://reviewable.io/reviews/gramineproject/gramine/1600#-NgOSTGJCt1jx-XfH7h-:-NhAg0w2ADJ0rR5jnTpN:b-4yipk7) ([raw file](https://github.com/gramineproject/gramine/blob/5295c1a09adea4a2538278ecf8b5647248a62539/libos/src/sys/libos_file.c#L414)):*

@aneessahib But in this case we side-step the overhead of `malloc()` (heap allocation, which is expensive) and instead do `__alloca()` (stack allocation, which is very cheap).

We just need to perf benchmark my suggestion.
</blockquote></details>

I have taken nginx performance results with 8KB. we could see 12% reduction in gramine-direct throughput when buf_size changed to 8KB from 64KB on stack. 
I think 4KB stack allocation can reduce it further.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1600)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @mkow)

___
*[`libos/src/sys/libos_file.c` line 414 at r1](https://reviewable.io/reviews/gramineproject/gramine/1600#-NgOSTGJCt1jx-XfH7h-:-NhAtVKT1d4pM12gUg2_:b-vytm4a) ([raw file](https://github.com/gramineproject/gramine/blob/5295c1a09adea4a2538278ecf8b5647248a62539/libos/src/sys/libos_file.c#L414)):*
<details><summary><i>Previously, TejaswineeL (Tejaswinee) wrote…</i></summary><blockquote>

I have taken nginx performance results with 8KB. we could see 12% reduction in gramine-direct throughput when buf_size changed to 8KB from 64KB on stack. 
I think 4KB stack allocation can reduce it further.
</blockquote></details>

@TejaswineeL Could you show the git diff of the modifications that you used? This is really surprising to me that the performance **gets worse** with my proposed three-tier approach (on the benchmark with 10KB web pages).

TejaswineeL (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1600)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @mkow)

___
*[`libos/src/sys/libos_file.c` line 414 at r1](https://reviewable.io/reviews/gramineproject/gramine/1600#-NgOSTGJCt1jx-XfH7h-:-NhBe87F7G7Se_6eTLyW:b-xf3qg1) ([raw file](https://github.com/gramineproject/gramine/blob/5295c1a09adea4a2538278ecf8b5647248a62539/libos/src/sys/libos_file.c#L414)):*

@TejaswineeL Could you show the git diff of the modifications that you used? This is really surprising to me that the performance **gets worse** with my proposed three-tier approach (on the benchmark with 10KB web pages).
</blockquote></details>

Here's the git diff with __alloca(4KB) [as the file size is 10K]

-#define BUF_SIZE (64 * 1024) /* read/write in 64KB chunks for sendfile() */
+#define BUF_SIZE (4 * 1024) /* read/write in 64KB chunks for sendfile() */


-    buf = malloc(BUF_SIZE);
+    char* buf = __alloca(BUF_SIZE);

-free(buf);

Observation: with alloca(4KB) option I see 10% decrease in the gramine degradation w.r.t. to current(malloc) allocation.
______________________________________________________________________________________________
Here's the git diff with static stack allocation of 64KB 

-char* buf = NULL;
+static char buf[BUF_SIZE];

-    buf = malloc(BUF_SIZE);

-free(buf);

Observation: with static stack allocation I see 25% decrease in gramine degradation w.r.t. current(malloc) allocation.

PR #1615, opened 2023-10-23
[LibOS] Use static buffer for first thread executing `sendfile()`

dimakuv:
## Description of the changes 

This performance optimization uses a statically allocated buffer instead of allocating on the heap in the `sendfile()` syscall. To prevent data races on a single static buffer, this optimization is applied only to the first thread. This optimization is useful for sendfile-heavy workloads, e.g. Nginx in HTTP mode (and with config `sendfile=on`).

The approach is suggested by @mkow privately.

Closes #1600. See there for preliminary discussions and approaches that don't work well.

Thanks to @aneessahib and @TejaswineeL for doing initial exploration.

## How to test this PR? 

Use Nginx in HTTP mode. The `sendfile=on` option is [already set in our CI-Examples config file](https://github.com/gramineproject/gramine/blob/2adc7d9bf22c0cb915657972bc49ddb653bda73d/CI-Examples/nginx/nginx-gramine.conf.template#L32).

Manual results below. Take them with grain of salt as I tested on my single developer machine, without bothering for proper experiment environment. But they give an idea of how the optimization helps:

### HOW I RAN

- First terminal:
```
CI-Examples/nginx$ make start-gramine-server [SGX=1]
```

- Second terminal:
```
CI-Examples/nginx$ LOOP=1 CONCURRENCY_LIST="256" ../common_tools/benchmark-http.sh http://127.0.0.1:8002
```

The above `benchmark-http.sh` script basically runs this:
```wrk -c 300 -d 30 -t 256 -R 10000 http://127.0.0.1:8002/random/10K.1.html```

### WITHOUT PR

- direct:
```
Run = 1 Concurrency = 256 Per thread Throughput (bytes) = 40.33, Latency (ms) = 1.00
Concurrency = 256: Per Thread Median Througput (bytes) =    40.330, Latency (ms) =     1.000
```

- SGX:
```
Run = 1 Concurrency = 256 Per thread Throughput (bytes) = 40.75, Latency (ms) = 3.24
Concurrency = 256: Per Thread Median Througput (bytes) =    40.750, Latency (ms) =     3.240
```

### WITH PR

- direct (`make start-gramine-server`):
```
Run = 1 Concurrency = 256 Per thread Throughput (bytes) = 39.89, Latency (ms) = .773
Concurrency = 256: Per Thread Median Througput (bytes) =    39.890, Latency (ms) =     0.773
```

- SGX (`make start-gramine-server SGX=1`):
```
Run = 1 Concurrency = 256 Per thread Throughput (bytes) = 40.74, Latency (ms) = 2.80
Concurrency = 256: Per Thread Median Througput (bytes) =    40.740, Latency (ms) =     2.800
```

### CONCLUSIONS

We have roughly the same throughput, but the latency is improved by 15-25%. Not bad.

### DIFF
--- a/libos/src/sys/libos_file.c
+++ b/libos/src/sys/libos_file.c
@@ -17,7 +17,17 @@
 #include "perm.h"
 #include "stat.h"
 
-#define BUF_SIZE (64 * 1024) /* read/write in 64KB chunks for sendfile() */
+/*
+ * Read/write in 64KB chunks in the sendfile() syscall. This syscall also has an optimization of
+ * using a statically allocated buffer instead of allocating on the heap (as our internal malloc()
+ * has a heavy mutex that guards all memory-allocation operations). To prevent data races of
+ * multiple threads executing sendfile() at the same time and thus potentially corrupting a single
+ * static buffer, we optimize for a common case: only the first thread uses the static buffer
+ * whereas other threads fall back to a slower heap allocation.
+ */
+#define BUF_SIZE (64 * 1024)
+static char g_sendfile_buf[BUF_SIZE];
+static bool g_sendfile_buf_in_use = false;
 
 /* The kernel would look up the parent directory, and remove the child from the inode. But we are
  * working with the PAL, so we open the file, truncate and close it. */
@@ -442,10 +452,17 @@ long libos_syscall_sendfile(int out_fd, int in_fd, off_t* offset, size_t count) 
     /* FIXME: This sendfile() emulation is very simple and not particularly efficient: it reads from
      *        input FD in BUF_SIZE chunks and writes into output FD. Mmap-based emulation may be
      *        more efficient but adds complexity (not all handle types provide mmap callback). */
-    buf = malloc(BUF_SIZE);
-    if (!buf) {
-        ret = -ENOMEM;
-        goto out;
+
+    bool buf_in_use = false;
+    if (__atomic_compare_exchange_n(&g_sendfile_buf_in_use, &buf_in_use, /*desired=*/true,
+                                    /*weak=*/true, __ATOMIC_RELEASE, __ATOMIC_ACQUIRE)) {
+        buf = g_sendfile_buf;
+    } else {
+        buf = malloc(BUF_SIZE);
+        if (!buf) {
+            ret = -ENOMEM;
+            goto out;
+        }
     }
 
     if (!count) {
@@ -533,7 +550,10 @@ out_update:
     }
 
 out:
-    free(buf);
+    if (buf == g_sendfile_buf)
+        __atomic_store_n(&g_sendfile_buf_in_use, 0, __ATOMIC_RELEASE);
+    else
+        free(buf);
     put_handle(in_hdl);
     put_handle(out_hdl);
     return copied_to_out ? (long)copied_to_out : ret;


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1615)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

___
*[`libos/src/sys/libos_file.c` line 23 at r1](https://reviewable.io/reviews/gramineproject/gramine/1615#-NhR-zma2kUMpADTRDzX:-NhR-zma2kUMpADTRDzY:b-vdudua) ([raw file](https://github.com/gramineproject/gramine/blob/b9ea93da66abf8ceecfe928ae3802d7f2af5a859/libos/src/sys/libos_file.c#L23)):*
> ```C
>  * Read/write in 64KB chunks in the sendfile() syscall. This syscall also has an optimization of
>  * using a statically allocated buffer instead of allocating on the heap (as our internal malloc()
>  * has a heavy mutex that guards all memory-allocation operations). To prevent data races of
> ```

Actually, maybe I should drop the `has a heavy mutex ...` explanation because I'm not sure it's correct in our case -- after all, Nginx is single-threaded so there shouldn't be really any mutex contention on malloc/free. Probably the overhead comes just from our internal allocator, which is not super-optimized...

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1615)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`libos/src/sys/libos_file.c` line 23 at r1](https://reviewable.io/reviews/gramineproject/gramine/1615#-NhR-zma2kUMpADTRDzX:-NhTxPfY56IakCD0lZwR:b32v57) ([raw file](https://github.com/gramineproject/gramine/blob/b9ea93da66abf8ceecfe928ae3802d7f2af5a859/libos/src/sys/libos_file.c#L23)):*

Actually, maybe I should drop the `has a heavy mutex ...` explanation because I'm not sure it's correct in our case -- after all, Nginx is single-threaded so there shouldn't be really any mutex contention on malloc/free. Probably the overhead comes just from our internal allocator, which is not super-optimized...
</blockquote></details>

+1
___
*[`libos/src/sys/libos_file.c` line 458 at r2](https://reviewable.io/reviews/gramineproject/gramine/1615#-NhTyHAO23h01GihWce4:-NhTyHAO23h01GihWce5:bgezegv) ([raw file](https://github.com/gramineproject/gramine/blob/8b46db0318edcb7fbb9edeffb932104ee6b0c619/libos/src/sys/libos_file.c#L458)):*
> ```C
>     bool buf_in_use = false;
>     if (__atomic_compare_exchange_n(&g_sendfile_buf_in_use, &buf_in_use, /*desired=*/true,
>                                     /*weak=*/true, __ATOMIC_RELEASE, __ATOMIC_ACQUIRE)) {
> ```

Actually, you can simplify (+ probably slightly speed up) this by using `__atomic_exchange_n()` instead.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1615)* status: 0 of 1 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`libos/src/sys/libos_file.c` line 23 at r1](https://reviewable.io/reviews/gramineproject/gramine/1615#-NhR-zma2kUMpADTRDzX:-NhV_RENEyS6Ggd6HPzb:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/b9ea93da66abf8ceecfe928ae3802d7f2af5a859/libos/src/sys/libos_file.c#L23)):*

+1
</blockquote></details>

Done.
___
*[`libos/src/sys/libos_file.c` line 458 at r2](https://reviewable.io/reviews/gramineproject/gramine/1615#-NhTyHAO23h01GihWce4:-NhV_RnJ7jODyW2r-eDx:bn81zal) ([raw file](https://github.com/gramineproject/gramine/blob/8b46db0318edcb7fbb9edeffb932104ee6b0c619/libos/src/sys/libos_file.c#L458)):*

Actually, you can simplify (+ probably slightly speed up) this by using `__atomic_exchange_n()` instead.
</blockquote></details>

Done.

According to my quick-and-dirty measurements, there is **no change in performance** (in comparison to the previous approach with `__atomic_compare_exchange_n`).

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1615)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1615)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1615#-NhWe-KE1siYNhPIyFwO:-NhWe-KE1siYNhPIyFwP:bwzofj8) (no related file):*
@aneessahib and @TejaswineeL: could you verify that this PR improves performance on your Nginx workload? The effect should be the same as in the original PR #1600. I will block until you ack/report the result.

___

kailun-qin (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1615)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners

TejaswineeL (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1615)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @aneessahib)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1615#-NhWe-KE1siYNhPIyFwO:-NhaH0r46Nc_F46xYMGf:b-djc9wj) (no related file):*

@aneessahib and @TejaswineeL: could you verify that this PR improves performance on your Nginx workload? The effect should be the same as in the original PR #1600. I will block until you ack/report the result.
</blockquote></details>

 @dimakuv 
yes the changes give the expected performance results for nginx (as that of with thread unsafe static buffer approach)
we are good to go with this change for nginx

Thanks

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1615)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1615#-NhWe-KE1siYNhPIyFwO:-NhaIuMB4F2KylcPr5yo:b-oh3yxv) (no related file):*
<details><summary><i>Previously, TejaswineeL (Tejaswinee) wrote…</i></summary><blockquote>

@dimakuv 
yes the changes give the expected performance results for nginx (as that of with thread unsafe static buffer approach)
we are good to go with this change for nginx

Thanks
</blockquote></details>

@TejaswineeL Awesome, thanks for running the experiment!

___

PR #1633, opened 2023-10-31
[LibOS] Implement `close_range` syscall

mwkmwkmwk:
## Description of the changes

This syscall is used by many language runtimes (Python included) to ensure only the explicitly specified file descriptors are actually passed to the newly created subprocess. Without it, the runtimes will generally fallback to slower methods like scanning `/proc/self/fd/`.

## How to test this PR?

Regression test covering the functionality is included in the PR.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1633)


kailun-qin (changes requested):
> Reviewed 12 of 12 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1633)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @mwkmwkmwk)

___
*[`libos/include/linux_abi/fs.h` line 14 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-Ni7g7_U6RrRek2q-SRT:-Ni7g7_U6RrRek2q-SRU:bht66qi) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/include/linux_abi/fs.h#L14)):*
> ```objc
> #include <asm/stat.h>
> #include <asm/statfs.h>
> #include <linux/close_range.h>
> ```

This was introduced since Linux 5.9. Is it guaranteed to be available?
___
*[`libos/test/fs/close_range.c` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-Ni7hkwE54n0hJNU3jLn:-Ni7hkwE54n0hJNU3jLo:bgx5t55) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L2)):*
> ```C
> #define _GNU_SOURCE
> ```

Any special reason that we put the test under `test/fs` rather than `test/regression`?

If not, I'd suggest to put it as a regression test and conform w/ the style there (e.g., using `CHECK()` macro, checking "TEST OK" etc.).

dimakuv (changes requested):
> Reviewed 12 of 12 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1633)* status: all files reviewed, 21 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin, @mkow, and @mwkmwkmwk)

___
*[`-- commits` line 7 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiE3y6s-L4mWrWqMYJg):*
Out of curiosity: why did you decide to implement this syscall exactly? Was it just a convenient syscall to start contributing to Gramine, or was there a particular need to implement it?
___
*[`libos/include/libos_table.h` line 209 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiE4YMs1aig6DW-DoSk:-NiE4YMs1aig6DW-DoSl:b-7lndga) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/include/libos_table.h#L209)):*
> ```objc
> long libos_syscall_mlock2(unsigned long start, size_t len, int flags);
> long libos_syscall_sysinfo(struct sysinfo* info);
> long libos_syscall_close_range(unsigned int first, unsigned int last, int flags);
> ```

Linux defines `flags` as unsigned int: https://elixir.bootlin.com/linux/v6.6-rc7/source/fs/open.c#L1595

Any reason you use `int`? Please note that we must conform to Linux kernel, not to Glibc wrappers. And even those use `unsigned int` -- https://man7.org/linux/man-pages/man2/close_range.2.html
___
*[`libos/include/linux_abi/fs.h` line 14 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-Ni7g7_U6RrRek2q-SRT:-NiE4jh77m0jE2E_Fo8n:b138bw5) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/include/linux_abi/fs.h#L14)):*

This was introduced since Linux 5.9. Is it guaranteed to be available?
</blockquote></details>

Please replace the Linux header with our own header. I guess you only need the flags, so please copy them from e.g. here: https://android.googlesource.com/kernel/common/+/refs/heads/android-mainline/include/uapi/linux/close_range.h

Gramine is in the process of removing all Linux headers taken from the host (where Gramine builds) system. There should be no dependency on the host in the LibOS headers. Check also with @mkow  on this effort, and see our GitHub issues and commit history.
___
*[`libos/src/sys/libos_open.c` line 177 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEB-Xj2Pj_cr7REnPD:-NiEB-Xj2Pj_cr7REnPE:b-1456qo) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/src/sys/libos_open.c#L177)):*
> ```C
> }
> 
> long libos_syscall_close_range(unsigned int first, unsigned int last, int flags) {
> ```

ditto (type of `flags`)
___
*[`libos/src/sys/libos_open.c` line 181 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEC7S2AOzaBSidlN4n:-NiEC7S2AOzaBSidlN4o:bejg6fo) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/src/sys/libos_open.c#L181)):*
> ```C
>         return -EINVAL;
>     }
>     if (flags & CLOSE_RANGE_UNSHARE) {
> ```

I wonder if this unsharing really works :) I don't think we ever tested this really, and probably normal apps don't use this at all.

But if it works, then looks like we have all the building blocks to implement `unshare(CLONE_FILES)`, which could be a good exercise for a Gramine dev.

UPDATE: Wow it works. I see now that you check it in your new test. Nice to know this. Definitely can implement `unshare(CLONE_FILES)` then.
___
*[`libos/src/sys/libos_open.c` line 190 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEBrys-p35uKzF7AYH:-NiEBrys-p35uKzF7AYI:b8khm3u) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/src/sys/libos_open.c#L190)):*
> ```C
>         }
>         set_handle_map(thread, new_map);
>         put_handle_map(new_map);
> ```

I had to double-check how refcounting works here. Everything is good:
- `dup_handle_map()` creates a new map with refcount = 1
- `set_handle_map()` increases new map's refcount, so refcount = 2
- we must decrease the refcount to one, as the only user/owner is this thread, so we do `put_handle_map()`
___
*[`libos/test/fs/close_range.c` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-Ni7hkwE54n0hJNU3jLn:-NiE9bCG-rO0eSQ8ZAAu:baggbvw) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L2)):*

Any special reason that we put the test under `test/fs` rather than `test/regression`?

If not, I'd suggest to put it as a regression test and conform w/ the style there (e.g., using `CHECK()` macro, checking "TEST OK" etc.).
</blockquote></details>

+1. Please move under `test/regression`.

I know the directory names may be misleading, but the idea is as follows:
- `test/fs/` contains tests on **regular files** and their contents. A special feature of these tests is that the same test verifies all FS types Gramine offers: allowed (aka passthrough) files, trusted (aka hashed) files, protected (aka encrypted) files, in-memory-only (aka tmpfs) files.
- `test/regression/` contains miscellaneous tests.

Since your test doesn't fall under the category of "regular files", it should go under `test/regression/`.
___
*[`libos/test/fs/close_range.c` line 4 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiECvek456T6sMHJxpx:-NiECrSd1J74eY1MWR4W:b-jkgf9u) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L4)):*
> ```C
> 
> #include <fcntl.h>
> #include <linux/close_range.h>
> ```

Hm, I'm a bit unsure -- I know there are systems that use e.g. Linux 5.4 (with SGX patches on top), and so this test will fail on them as there was no `close_range()` at that version.

Do we want to verify this, and run this test only on Linux 5.9+ ?
___
*[`libos/test/fs/close_range.c` line 11 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEDBs02vffADYBOZlh:-NiEDBs14k7s01dyDr0U:b-60otho) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L11)):*
> ```C
> 
> static void open_files(void) {
>     for (int i = 3; i <= 9; i += 1) {
> ```

`i += 1`?
___
*[`libos/test/fs/close_range.c` line 23 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEDw1s4Es-IyJqbNuM:-NiEDw1s4Es-IyJqbNuN:b-v3sm5a) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L23)):*
> ```C
> 
> static void close_files(void) {
>     if (close_range(3, ~0u, 0)) {
> ```

In other cases (e.g. `fcntl()`) you check against `-1` explicitly, maybe do the same here?
___
*[`libos/test/fs/close_range.c` line 39 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEDfQ-BZ4U7e6lAwz0:-NiEDfQ-BZ4U7e6lAwz1:b-2naxlr) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L39)):*
> ```C
>         fatal_error("File %d not open\n", fd);
>     }
>     if (res != FD_CLOEXEC) {
> ```

Theoretically the returned flags can have some other bits set, so maybe replace this check with a bit-set check?
___
*[`libos/test/fs/close_range.c` line 49 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEDlNFCnrjLxbl5gCQ:-NiEDlNFCnrjLxbl5gCR:b-xsbhph) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L49)):*
> ```C
>         fatal_error("File %d not open\n", fd);
>     }
>     if (res != 0) {
> ```

Theoretically the returned flags can have some other bits set, so maybe replace this check with a bit-not-set check?
___
*[`libos/test/fs/close_range.c` line 56 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEE-nvDNcyuwRAcJyO:-NiEE-nvDNcyuwRAcJyP:b-v3sm5a) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L56)):*
> ```C
> static void* close_range_thread(void* arg) {
>     int flags = *(int*)arg;
>     if (close_range(5, 7, flags)) {
> ```

In other cases (e.g. `fcntl()`) you check against `-1` explicitly, maybe do the same here?
___
*[`libos/test/fs/close_range.c` line 65 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEE5gyBjcNNsST3B5F:-NiEE5gyBjcNNsST3B5G:b-pb4q6m) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L65)):*
> ```C
>         check_cloexec(6);
>         check_cloexec(7);
> ```

Please remove empty line
___
*[`libos/test/fs/close_range.c` line 87 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEEFV06KO6uF-u_bzY:-NiEEFV06KO6uF-u_bzZ:b-ftzrke) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L87)):*
> ```C
> }
> 
> static void close_range_plain(void) {
> ```

Here and below: can you add the `test_` prefix? That's our typical convention, to distinguish where helper functions end and sub-test functions start.
___
*[`libos/test/fs/close_range.c` line 89 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEEWpGD8s8vAGSpKFM:-NiEEWpGD8s8vAGSpKFN:b-j8pr4t) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L89)):*
> ```C
> static void close_range_plain(void) {
>     open_files();
>     run_thread(0);
> ```

Please change to `run_thread(/*flags=*/0)`. Otherwise I had to pause here and remember the signature.
___
*[`libos/test/fs/close_range.c` line 90 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEEoln6t4wokDIxlBH:-NiEEoln6t4wokDIxlBI:b-tig19b) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L90)):*
> ```C
>     open_files();
>     run_thread(0);
>     check_open_nocloexec(3);
> ```

Just to make sure: Below checks are because you performed `close_range()` in a separate thread, and you want to make sure that the same FD map state is reflected in this (main) thread?
___
*[`libos/test/fs/close_range.c` line 97 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEFBKo7I_QwizeGfIm:-NiEFBKo7I_QwizeGfIn:bvqfbte) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L97)):*
> ```C
>     check_open_nocloexec(8);
>     check_open_nocloexec(9);
>     close_files();
> ```

This is confusing. `close_files()` seems to be used purely for cleanup of the test, but it still uses the under-test `close_range()` syscall. Isn't it contradictory? I think `close_files()` should have an explicit loop over FDs 3..9 and perform `close()` on each FD.
___
*[`libos/test/fs/close_range.c` line 116 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEFX3EBNPYe9s7oV0n:-NiEFX3EBNPYe9s7oV0o:b43w0i6) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L116)):*
> ```C
>     open_files();
>     run_thread(CLOSE_RANGE_UNSHARE);
>     for (int i = 3; i <= 9; i++) check_open_nocloexec(i);
> ```

Please split into two lines. We never have this one-liner coding style.
___
*[`libos/test/fs/close_range.c` line 123 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEF_fj5qnzI-HvVw3N:-NiEF_fj5qnzI-HvVw3O:b43w0i6) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L123)):*
> ```C
>     open_files();
>     run_thread(CLOSE_RANGE_UNSHARE | CLOSE_RANGE_CLOEXEC);
>     for (int i = 3; i <= 9; i++) check_open_nocloexec(i);
> ```

Please split into two lines. We never have this one-liner coding style.
___
*[`libos/test/fs/close_range.c` line 128 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEFp6N4Ydo85DlQejp:-NiEFp6N4Ydo85DlQejq:b-6l2v8w) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L128)):*
> ```C
> 
> int main(void) {
>     setup();
> ```

You won't need this redundant line if you move the test to `test/regression/`.
___
*[`libos/test/fs/close_range.c` line 136 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEFlGJA7NagR9Agtt4:-NiEFlGJA7NagR9Agtt5:b-ic29lb) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L136)):*
> ```C
> 
>     return 0;
> }
> ```

Please add a newline symbol here.

mwkmwkmwk (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1633)* status: 5 of 16 files reviewed, 21 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @kailun-qin, and @mkow)

___
*[`-- commits` line 7 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiE3y6s-L4mWrWqMYJg):*

Out of curiosity: why did you decide to implement this syscall exactly? Was it just a convenient syscall to start contributing to Gramine, or was there a particular need to implement it?
</blockquote></details>

Just seemed like an easy addition with some perf improvement.
___
*[`libos/include/libos_table.h` line 209 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiE4YMs1aig6DW-DoSk:-NiFyDIXBsowMj1R8nm_:b-ah9ksp) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/include/libos_table.h#L209)):*

Linux defines `flags` as unsigned int: https://elixir.bootlin.com/linux/v6.6-rc7/source/fs/open.c#L1595

Any reason you use `int`? Please note that we must conform to Linux kernel, not to Glibc wrappers. And even those use `unsigned int` -- https://man7.org/linux/man-pages/man2/close_range.2.html
</blockquote></details>

Okay, my mistake.
___
*[`libos/include/linux_abi/fs.h` line 14 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-Ni7g7_U6RrRek2q-SRT:-NiG78179YLXl3pG6qjx:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/include/linux_abi/fs.h#L14)):*

Please replace the Linux header with our own header. I guess you only need the flags, so please copy them from e.g. here: https://android.googlesource.com/kernel/common/+/refs/heads/android-mainline/include/uapi/linux/close_range.h

Gramine is in the process of removing all Linux headers taken from the host (where Gramine builds) system. There should be no dependency on the host in the LibOS headers. Check also with @mkow  on this effort, and see our GitHub issues and commit history.
</blockquote></details>

Done.
___
*[`libos/src/sys/libos_open.c` line 177 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEB-Xj2Pj_cr7REnPD:-NiFzTwk1KO9Ha9rH6Ph:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/src/sys/libos_open.c#L177)):*

ditto (type of `flags`)
</blockquote></details>

Done.
___
*[`libos/src/sys/libos_open.c` line 181 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEC7S2AOzaBSidlN4n:-NiFzaMTBYJPTfd8JhtK:bywq403) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/src/sys/libos_open.c#L181)):*

I wonder if this unsharing really works :) I don't think we ever tested this really, and probably normal apps don't use this at all.

But if it works, then looks like we have all the building blocks to implement `unshare(CLONE_FILES)`, which could be a good exercise for a Gramine dev.

UPDATE: Wow it works. I see now that you check it in your new test. Nice to know this. Definitely can implement `unshare(CLONE_FILES)` then.
</blockquote></details>

Yeah. It felt kind of weird implementing this given that we currently don't have any other way to unshare files, but it didn't seem to cause any trouble.
___
*[`libos/test/fs/close_range.c` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-Ni7hkwE54n0hJNU3jLn:-NiG2XeiCGzSDeZO6GHe:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L2)):*

+1. Please move under `test/regression`.

I know the directory names may be misleading, but the idea is as follows:
- `test/fs/` contains tests on **regular files** and their contents. A special feature of these tests is that the same test verifies all FS types Gramine offers: allowed (aka passthrough) files, trusted (aka hashed) files, protected (aka encrypted) files, in-memory-only (aka tmpfs) files.
- `test/regression/` contains miscellaneous tests.

Since your test doesn't fall under the category of "regular files", it should go under `test/regression/`.
</blockquote></details>

Done.
___
*[`libos/test/fs/close_range.c` line 4 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiECvek456T6sMHJxpx:-NiFzqMUCOVfa-era9ph:bk0954o) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L4)):*

Hm, I'm a bit unsure -- I know there are systems that use e.g. Linux 5.4 (with SGX patches on top), and so this test will fail on them as there was no `close_range()` at that version.

Do we want to verify this, and run this test only on Linux 5.9+ ?
</blockquote></details>

This shouldn't matter, as the implementation we're testing is provided by Gramine and not the underlying kernel, right?
___
*[`libos/test/fs/close_range.c` line 11 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEDBs02vffADYBOZlh:-NiFzyBSCGqZjDwNHMgl:bhc5hzg) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L11)):*

`i += 1`?
</blockquote></details>

Oops, spent too much time writing Rust.
___
*[`libos/test/fs/close_range.c` line 23 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEDw1s4Es-IyJqbNuM:-NiG-EqN4HqeXSnVVYYP:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L23)):*

In other cases (e.g. `fcntl()`) you check against `-1` explicitly, maybe do the same here?
</blockquote></details>

Done.
___
*[`libos/test/fs/close_range.c` line 39 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEDfQ-BZ4U7e6lAwz0:-NiG-N584ZUr6KchV55N:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L39)):*

Theoretically the returned flags can have some other bits set, so maybe replace this check with a bit-set check?
</blockquote></details>

Done.
___
*[`libos/test/fs/close_range.c` line 49 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEDlNFCnrjLxbl5gCQ:-NiG-NgE5nyQuiJxYxx7:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L49)):*

Theoretically the returned flags can have some other bits set, so maybe replace this check with a bit-not-set check?
</blockquote></details>

Done.
___
*[`libos/test/fs/close_range.c` line 56 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEE-nvDNcyuwRAcJyO:-NiG-QIQ5OkIya0m6wy5:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L56)):*

In other cases (e.g. `fcntl()`) you check against `-1` explicitly, maybe do the same here?
</blockquote></details>

Done.
___
*[`libos/test/fs/close_range.c` line 65 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEE5gyBjcNNsST3B5F:-NiG-SBz3shLEWDp6PgV:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L65)):*

Please remove empty line
</blockquote></details>

Done.
___
*[`libos/test/fs/close_range.c` line 87 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEEFV06KO6uF-u_bzY:-NiG-YRV6xWmOYute5kV:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L87)):*

Here and below: can you add the `test_` prefix? That's our typical convention, to distinguish where helper functions end and sub-test functions start.
</blockquote></details>

Done.
___
*[`libos/test/fs/close_range.c` line 89 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEEWpGD8s8vAGSpKFM:-NiG-ajiDrEON5-oB1yB:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L89)):*

Please change to `run_thread(/*flags=*/0)`. Otherwise I had to pause here and remember the signature.
</blockquote></details>

Done.
___
*[`libos/test/fs/close_range.c` line 90 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEEoln6t4wokDIxlBH:-NiG-crtBtWtRrackU6l:b-13s102) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L90)):*

Just to make sure: Below checks are because you performed `close_range()` in a separate thread, and you want to make sure that the same FD map state is reflected in this (main) thread?
</blockquote></details>

Yes, I want to make sure that plain `close_range` propagates to main thread, while version with `CLOSE_RANGE_UNSHARE` does not.
___
*[`libos/test/fs/close_range.c` line 97 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEFBKo7I_QwizeGfIm:-NiG-n_N8Ehq8WujgDAk:bf9kh5z) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L97)):*

This is confusing. `close_files()` seems to be used purely for cleanup of the test, but it still uses the under-test `close_range()` syscall. Isn't it contradictory? I think `close_files()` should have an explicit loop over FDs 3..9 and perform `close()` on each FD.
</blockquote></details>

It does provide a bit of extra test coverage, and is cross-verified by `open_files` making sure that the newly-opened files have expected FDs.
___
*[`libos/test/fs/close_range.c` line 116 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEFX3EBNPYe9s7oV0n:-NiG05r3ALHaMtQn3Mzk:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L116)):*

Please split into two lines. We never have this one-liner coding style.
</blockquote></details>

Done.
___
*[`libos/test/fs/close_range.c` line 123 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEF_fj5qnzI-HvVw3N:-NiG06LgDsj2stTkya6Z:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L123)):*

Please split into two lines. We never have this one-liner coding style.
</blockquote></details>

Done.
___
*[`libos/test/fs/close_range.c` line 128 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEFp6N4Ydo85DlQejp:-NiG2YToC5087d_HVQVk:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L128)):*

You won't need this redundant line if you move the test to `test/regression/`.
</blockquote></details>

Done.
___
*[`libos/test/fs/close_range.c` line 136 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEFlGJA7NagR9Agtt4:-NiG090d-i1WNFLh66Fr:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L136)):*

Please add a newline symbol here.
</blockquote></details>

Done.

dimakuv (changes requested):
> Reviewed 11 of 11 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1633)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin and @mwkmwkmwk)

___
*[`libos/include/libos_table.h` line 1 at r2](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiJ5vwyCh0zFpivkfyH:-NiJ5vwyCh0zFpivkfyI:bqvrdmj) ([raw file](https://github.com/gramineproject/gramine/blob/0884a46f0e9b37e7b35918ea2ec2899183b245d9/libos/include/libos_table.h#L1)):*
> ```objc
> 
> ```

Spurious empty line?
___
*[`libos/include/linux_abi/fs.h` line 55 at r2](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiJ64nk7onCuryLzgWk:-NiJ64nk7onCuryLzgWl:b-lhzo7b) ([raw file](https://github.com/gramineproject/gramine/blob/0884a46f0e9b37e7b35918ea2ec2899183b245d9/libos/include/linux_abi/fs.h#L55)):*
> ```objc
> 
> #define CLOSE_RANGE_UNSHARE	(1U << 1)
> #define CLOSE_RANGE_CLOEXEC	(1U << 2)
> ```

Our codebase doesn't use tabs. Please replace with spaces.
___
*[`libos/test/fs/close_range.c` line 4 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiECvek456T6sMHJxpx:-NiJ6W-qDN1eIOao3WFc:b-cg4d2m) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L4)):*
> This shouldn't matter, as the implementation we're testing is provided by Gramine and not the underlying kernel, right?

But the problem is: these tests are built just like regular applications -- on the host system, without using Gramine. I mean, Gramine just takes the binary and runs it in its own "Linux-like" environment. But **someone needs to build the application binary first**. This someone is the host (at least in our CI), and if the host has e.g. Linux 5.4, this test won't be able to run -- simply because it cannot be built in the first place.

I think it shouldn't be hard to add a bit of logic in our `meson.build` script to check the Linux version and skip this test if too-old.
___
*[`libos/test/fs/close_range.c` line 97 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiEFBKo7I_QwizeGfIm:-NiJ8Q6n6kusYDRaVq29:b1fgdll) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L97)):*

It does provide a bit of extra test coverage, and is cross-verified by `open_files` making sure that the newly-opened files have expected FDs.
</blockquote></details>

Ok, fair enough. The point about subsequent `open_files()` double-checking that the same FDs can be reused (because they were closed) is valid.
___
*[`libos/test/regression/close_range.c` line 15 at r2](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiJ7YBD2mx-gi1mn2fc:-NiJ7YBD2mx-gi1mn2fd:bmx2im2) ([raw file](https://github.com/gramineproject/gramine/blob/0884a46f0e9b37e7b35918ea2ec2899183b245d9/libos/test/regression/close_range.c#L15)):*
> ```C
>         int res = open("/dev/null", O_RDONLY);
>         if (res < 0) {
>             err(1, "open /dev/null");
> ```

In case you're interested, we have [`common.h`](https://github.com/gramineproject/gramine/blob/master/libos/test/regression/common.h) header, which you can use in regression tests. In this particular place, `CHECK()` macro is useful.

But I don't insist, as I actually prefer to have tests completely self-contained. (But other Gramine devs have/had the contrary opinions.)
___
*[`libos/test/regression/close_range.c` line 31 at r2](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiJ7ydL4-EDtePDlNb_:-NiJ7ydL4-EDtePDlNba:b-3wjxrl) ([raw file](https://github.com/gramineproject/gramine/blob/0884a46f0e9b37e7b35918ea2ec2899183b245d9/libos/test/regression/close_range.c#L31)):*
> ```C
> static void check_closed(int fd) {
>     if (fcntl(fd, F_GETFD) != -1 || errno != EBADF) {
>         errx(1, "file %d not closed", fd);
> ```

Why not `err()`?

mwkmwkmwk (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1633)* status: 14 of 16 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @kailun-qin)

___
*[`libos/include/libos_table.h` line 1 at r2](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiJ5vwyCh0zFpivkfyH:-NiKnQ_uEqTKD-vi9Sjr:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/0884a46f0e9b37e7b35918ea2ec2899183b245d9/libos/include/libos_table.h#L1)):*

Spurious empty line?
</blockquote></details>

Done.
___
*[`libos/include/linux_abi/fs.h` line 55 at r2](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiJ64nk7onCuryLzgWk:-NiKnUi8EBm94xEVuiCn:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/0884a46f0e9b37e7b35918ea2ec2899183b245d9/libos/include/linux_abi/fs.h#L55)):*

Our codebase doesn't use tabs. Please replace with spaces.
</blockquote></details>

Done.
___
*[`libos/test/fs/close_range.c` line 4 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiECvek456T6sMHJxpx:-NiKo0n79kguvao3frMm:b-q2kjo4) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L4)):*

> This shouldn't matter, as the implementation we're testing is provided by Gramine and not the underlying kernel, right?

But the problem is: these tests are built just like regular applications -- on the host system, without using Gramine. I mean, Gramine just takes the binary and runs it in its own "Linux-like" environment. But **someone needs to build the application binary first**. This someone is the host (at least in our CI), and if the host has e.g. Linux 5.4, this test won't be able to run -- simply because it cannot be built in the first place.

I think it shouldn't be hard to add a bit of logic in our `meson.build` script to check the Linux version and skip this test if too-old.
</blockquote></details>

What we actually want to check there would be *kernel header* and libc versions, not actual kernel, right?
___
*[`libos/test/regression/close_range.c` line 31 at r2](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiJ7ydL4-EDtePDlNb_:-NiKnbyK5-975zMxMFfG:b-oom9qm) ([raw file](https://github.com/gramineproject/gramine/blob/0884a46f0e9b37e7b35918ea2ec2899183b245d9/libos/test/regression/close_range.c#L31)):*

Why not `err()`?
</blockquote></details>

Because if this check fails in basically the only way possible (the file is actually open), errno will not be set.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1633)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin and @mwkmwkmwk)

___
*[`libos/test/fs/close_range.c` line 4 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiECvek456T6sMHJxpx:-NiYZhCYD_ZfnCYJVSzv:b-4rzdhl) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L4)):*
> What we actually want to check there would be _kernel header_ and libc versions, not actual kernel, right?

Yes, you can put it this way.

I guess what you're hinting at is:
- Inline required macros from `linux/close_range.h` in this test, if Linux < 5.9
- Use raw `syscall()` instead of the libc wrapper `close_range()`, if Linux < 5.9

Surely we can do it. This is a better approach than what I suggested in the beginning (hacking `meson.build` to skip this test).
___
*[`libos/test/regression/close_range.c` line 31 at r2](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiJ7ydL4-EDtePDlNb_:-NiY_70q2ZrUy03NGV8-:b3xr3p7) ([raw file](https://github.com/gramineproject/gramine/blob/0884a46f0e9b37e7b35918ea2ec2899183b245d9/libos/test/regression/close_range.c#L31)):*

Because if this check fails in basically the only way possible (the file is actually open), errno will not be set.
</blockquote></details>

Ah, sorry, you're right. The error condition here (among true non-EBADF errors) can be that `fcntl() >= 0`.

mwkmwkmwk (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1633)* status: 11 of 16 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @kailun-qin)

___
*[`libos/test/fs/close_range.c` line 4 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiECvek456T6sMHJxpx:-NkNC97F6PBvCyTG9qPI:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L4)):*

> What we actually want to check there would be _kernel header_ and libc versions, not actual kernel, right?

Yes, you can put it this way.

I guess what you're hinting at is:
- Inline required macros from `linux/close_range.h` in this test, if Linux < 5.9
- Use raw `syscall()` instead of the libc wrapper `close_range()`, if Linux < 5.9

Surely we can do it. This is a better approach than what I suggested in the beginning (hacking `meson.build` to skip this test).
</blockquote></details>

Done.

dimakuv (changes requested):
> Reviewed 5 of 5 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1633)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin and @mwkmwkmwk)

___
*[`libos/test/regression/close_range.c` line 17 at r5](https://reviewable.io/reviews/gramineproject/gramine/1633#-NkP7Che8s3KdBFB2iO6:-NkP7Che8s3KdBFB2iO7:bu7mar2) ([raw file](https://github.com/gramineproject/gramine/blob/94b37d5e451f39dc547ac2b84c2bdb8a0f2522c9/libos/test/regression/close_range.c#L17)):*
> ```C
> 
> #ifndef __NR_close_range
> #define __NR_close_range 436
> ```

Isn't this number correct only for x86 arch? I suggest to put it under an `ifdef __x86_64__`

Different archs have different syscall tables, and we can imagine that someday Gramine will be supported by other archs.
___
*[`libos/test/regression/close_range.c` line 20 at r5](https://reviewable.io/reviews/gramineproject/gramine/1633#-NkP7ibr3UsGHnCthwMq:-NkP7ibr3UsGHnCthwMr:btvyco5) ([raw file](https://github.com/gramineproject/gramine/blob/94b37d5e451f39dc547ac2b84c2bdb8a0f2522c9/libos/test/regression/close_range.c#L20)):*
> ```C
> #endif
> 
> // Likewise, don't rely on the glibc wrapper.
> ```

`glibc` is a bit misleading here. This test also works with musl. I would just rename to generic `libc`.
___
*[`libos/test/regression/close_range.c` line 22 at r5](https://reviewable.io/reviews/gramineproject/gramine/1633#-NkP7uV4AH5ohhCLtmCD:-NkP7uV4AH5ohhCLtmCE:b-wf8rc1) ([raw file](https://github.com/gramineproject/gramine/blob/94b37d5e451f39dc547ac2b84c2bdb8a0f2522c9/libos/test/regression/close_range.c#L22)):*
> ```C
> // Likewise, don't rely on the glibc wrapper.
> 
> int my_close_range(unsigned start, unsigned end, unsigned flags) {
> ```

We don't use the short version `unsigned` for the type. Please use the full form `unsigned int`.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1633)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin and @mwkmwkmwk)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1633#-NkP8Fvv53lEGx-9Q_oO:-NkP8Fvv53lEGx-9Q_oP:b-rrp4c8) (no related file):*
Please check why many jobs in CI failed.

___

mwkmwkmwk (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1633)* status: 15 of 16 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @kailun-qin)

___
*[`libos/test/regression/close_range.c` line 17 at r5](https://reviewable.io/reviews/gramineproject/gramine/1633#-NkP7Che8s3KdBFB2iO6:-NkPYLxD04zKTywKTRxp:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/94b37d5e451f39dc547ac2b84c2bdb8a0f2522c9/libos/test/regression/close_range.c#L17)):*

Isn't this number correct only for x86 arch? I suggest to put it under an `ifdef __x86_64__`

Different archs have different syscall tables, and we can imagine that someday Gramine will be supported by other archs.
</blockquote></details>

Done.
___
*[`libos/test/regression/close_range.c` line 20 at r5](https://reviewable.io/reviews/gramineproject/gramine/1633#-NkP7ibr3UsGHnCthwMq:-NkPYMFT1jbMmSOQIA5z:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/94b37d5e451f39dc547ac2b84c2bdb8a0f2522c9/libos/test/regression/close_range.c#L20)):*

`glibc` is a bit misleading here. This test also works with musl. I would just rename to generic `libc`.
</blockquote></details>

Done.
___
*[`libos/test/regression/close_range.c` line 22 at r5](https://reviewable.io/reviews/gramineproject/gramine/1633#-NkP7uV4AH5ohhCLtmCD:-NkPYMiI-Gy63p8y24--:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/94b37d5e451f39dc547ac2b84c2bdb8a0f2522c9/libos/test/regression/close_range.c#L22)):*

We don't use the short version `unsigned` for the type. Please use the full form `unsigned int`.
</blockquote></details>

Done.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1633)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin and @mwkmwkmwk)

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1633)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @kailun-qin and @mwkmwkmwk)

mwkmwkmwk (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1633)* status: 15 of 16 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1633#-NkP8Fvv53lEGx-9Q_oO:-NkVAtzg7CL11SoaTFgV:b-cdc2x7) (no related file):*

Please check why many jobs in CI failed.
</blockquote></details>

It seems to be fixed already? Not sure what that was about.

___
___
*[`libos/test/fs/close_range.c` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/1633#-Ni7hkwE54n0hJNU3jLn:-NkVAr1q3C5eXi3rMej3:bwrhh3o) ([raw file](https://github.com/gramineproject/gramine/blob/f885603dfaf0b5a3ae647dbb0777fbba03910cd2/libos/test/fs/close_range.c#L2)):*

Done.
</blockquote></details>

Also converted to use CHECK in a few places.
___
*[`libos/test/regression/close_range.c` line 15 at r2](https://reviewable.io/reviews/gramineproject/gramine/1633#-NiJ7YBD2mx-gi1mn2fc:-NkVApMz1JDo5aVgitYS:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/0884a46f0e9b37e7b35918ea2ec2899183b245d9/libos/test/regression/close_range.c#L15)):*

In case you're interested, we have [`common.h`](https://github.com/gramineproject/gramine/blob/master/libos/test/regression/common.h) header, which you can use in regression tests. In this particular place, `CHECK()` macro is useful.

But I don't insist, as I actually prefer to have tests completely self-contained. (But other Gramine devs have/had the contrary opinions.)
</blockquote></details>

Done.

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1633)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1633#-NkP8Fvv53lEGx-9Q_oO:-NkVHm5n9IL-Ob2kgMqd:b758is) (no related file):*

It seems to be fixed already? Not sure what that was about.
</blockquote></details>

Yes, you fixed it already. I'm pretty sure it was about the missing `static` word in your test.

___

mkow (changes requested):
> Reviewed 3 of 11 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1633)* status: 8 of 16 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @kailun-qin, and @mwkmwkmwk)

___
*[`libos/test/regression/close_range.c` line 28 at r9](https://reviewable.io/reviews/gramineproject/gramine/1633#-NyI5m4r0WLUPjJGK20C:-NyI5m4r0WLUPjJGK20D:b-uryre1) ([raw file](https://github.com/gramineproject/gramine/blob/22ba5314473b268fe756ac5d2dd57790eb8604e4/libos/test/regression/close_range.c#L28)):*
> ```C
> // Likewise, don't rely on the libc wrapper.
> 
> static int my_close_range(unsigned int start, unsigned int end, unsigned int flags) {
> ```

start-end is often inclusive-exclusive, let's ensure it's clear



_Suggestion:_
```C
unsigned int last
```
___
*[`libos/test/regression/close_range.c` line 33 at r9](https://reviewable.io/reviews/gramineproject/gramine/1633#-NyIKw3m5f_ZrIyFS4H0:-NyIKw3n6WD3upCXvu_w:b-g7rupc) ([raw file](https://github.com/gramineproject/gramine/blob/22ba5314473b268fe756ac5d2dd57790eb8604e4/libos/test/regression/close_range.c#L33)):*
> ```C
> 
> static void open_files(void) {
>     for (int i = 3; i <= 9; i++) {
> ```

That's seems risky, because it depends on the external environment. Could you close these before running each test?

mwkmwkmwk (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1633)* status: 7 of 16 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @kailun-qin, and @mkow)

___
*[`libos/test/regression/close_range.c` line 28 at r9](https://reviewable.io/reviews/gramineproject/gramine/1633#-NyI5m4r0WLUPjJGK20C:-NyIND6y7ae7J33xGPGI:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/22ba5314473b268fe756ac5d2dd57790eb8604e4/libos/test/regression/close_range.c#L28)):*

start-end is often inclusive-exclusive, let's ensure it's clear
</blockquote></details>

Done.
___
*[`libos/test/regression/close_range.c` line 33 at r9](https://reviewable.io/reviews/gramineproject/gramine/1633#-NyIKw3m5f_ZrIyFS4H0:-NyINCvfAVMP7I2lMmke:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/22ba5314473b268fe756ac5d2dd57790eb8604e4/libos/test/regression/close_range.c#L33)):*

That's seems risky, because it depends on the external environment. Could you close these before running each test?
</blockquote></details>

Done.

mkow (changes requested):
> Reviewed 2 of 12 files at r1, 2 of 2 files at r3, 8 of 8 files at r9, 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1633)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin and @mwkmwkmwk)

___
*[`libos/src/sys/libos_open.c` line 186 at r10](https://reviewable.io/reviews/gramineproject/gramine/1633#-NyLE944-oPHvpSAsaXB:-NyLE944-oPHvpSAsaXC:b-j8zmm8) ([raw file](https://github.com/gramineproject/gramine/blob/b3a1237f2dfa2a48e99b71fa4842feb7b50c7274/libos/src/sys/libos_open.c#L186)):*
> ```C
> 
>         int err = dup_handle_map(&new_map, thread->handle_map);
>         if (err) {
> ```

just for uniformity



_Suggestion:_
```C
if (err < 0) {
```
___
*[`libos/test/regression/close_range.c` line 60 at r10](https://reviewable.io/reviews/gramineproject/gramine/1633#-NyIQl5E6WzFfHC2nK-2:-NyIQl5E6WzFfHC2nK-3:bh3ddbq) ([raw file](https://github.com/gramineproject/gramine/blob/b3a1237f2dfa2a48e99b71fa4842feb7b50c7274/libos/test/regression/close_range.c#L60)):*
> ```C
> }
> 
> static void check_open_nocloexec(int fd) {
> ```

The analogous function above doesn't have that `open_`.



_Suggestion:_
```C
check_nocloexec
```

mwkmwkmwk (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1633)* status: 14 of 16 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin and @mkow)

___
*[`libos/src/sys/libos_open.c` line 186 at r10](https://reviewable.io/reviews/gramineproject/gramine/1633#-NyLE944-oPHvpSAsaXB:-NyQhFC-9XDJCfKlFcls:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/b3a1237f2dfa2a48e99b71fa4842feb7b50c7274/libos/src/sys/libos_open.c#L186)):*

just for uniformity
</blockquote></details>

Done.
___
*[`libos/test/regression/close_range.c` line 60 at r10](https://reviewable.io/reviews/gramineproject/gramine/1633#-NyIQl5E6WzFfHC2nK-2:-NyQhatTC1ZEypong0GH:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/b3a1237f2dfa2a48e99b71fa4842feb7b50c7274/libos/test/regression/close_range.c#L60)):*

The analogous function above doesn't have that `open_`.
</blockquote></details>

Done.

mkow (approved):
> Reviewed 2 of 2 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1633)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

kailun-qin (dismissed):
> Reviewed 3 of 11 files at r2, 2 of 2 files at r3, 7 of 8 files at r9, 2 of 2 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1633)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

___
*[`libos/test/regression/close_range.c` line 89 at r11](https://reviewable.io/reviews/gramineproject/gramine/1633#-NyR1I_m6T0q3rYsMJtd:-NyR1I_m6T0q3rYsMJte:bgwaaes) ([raw file](https://github.com/gramineproject/gramine/blob/b2ddfb7fbde134c8746119f2b7d0c380dafb5f7b/libos/test/regression/close_range.c#L89)):*
> ```C
> static void run_thread(int flags) {
>     pthread_t thr;
>     if ((errno = pthread_create(&thr, 0, close_range_thread, &flags))) {
> ```

well, why not just "check ret + errx"? anyway, I'm not blocking
___
*[`libos/test/regression/close_range.c` line 148 at r11](https://reviewable.io/reviews/gramineproject/gramine/1633#-NyR0tBl1Pmm6arDjg4K:-NyR0tBl1Pmm6arDjg4L:bz8g81x) ([raw file](https://github.com/gramineproject/gramine/blob/b2ddfb7fbde134c8746119f2b7d0c380dafb5f7b/libos/test/regression/close_range.c#L148)):*
> ```C
>     test_close_range_cloexec_unshare();
> 
>     printf("TEST OK\n");
> ```

-> `puts()`

mwkmwkmwk (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1633)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

___
*[`libos/test/regression/close_range.c` line 89 at r11](https://reviewable.io/reviews/gramineproject/gramine/1633#-NyR1I_m6T0q3rYsMJtd:-NyVpqKx2UIv0BVkmY8c:by1wl8y) ([raw file](https://github.com/gramineproject/gramine/blob/b2ddfb7fbde134c8746119f2b7d0c380dafb5f7b/libos/test/regression/close_range.c#L89)):*

well, why not just "check ret + errx"? anyway, I'm not blocking
</blockquote></details>

This has the advantage of actually printing the error code.
___
*[`libos/test/regression/close_range.c` line 148 at r11](https://reviewable.io/reviews/gramineproject/gramine/1633#-NyR0tBl1Pmm6arDjg4K:-NyVpMkT-eP1_wW62VzN:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/b2ddfb7fbde134c8746119f2b7d0c380dafb5f7b/libos/test/regression/close_range.c#L148)):*

-> `puts()`
</blockquote></details>

Done.

mkow (approved):
> Reviewed 1 of 1 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1633)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

kailun-qin (approved):
> Reviewed 1 of 1 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1633)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #1640, opened 2023-11-05
improved CI-Examples lighttpd configuration for better performance 

nniryanb:
improved CI-Examples lighttpd configuration for better performance

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1640)

### DIFF
--- a/CI-Examples/lighttpd/Makefile
+++ b/CI-Examples/lighttpd/Makefile
@@ -11,8 +11,8 @@ LIGHTTPD_MIRRORS ?= \
     https://download.lighttpd.net/lighttpd/releases-1.4.x \
     https://packages.gramineproject.io/distfiles
 
-HOST ?= 10.6.0.6
-PORT ?= 8030
+HOST ?= 127.0.0.1
+PORT ?= 8003
 
 ifeq ($(DEBUG),1)
 GRAMINE_LOG_LEVEL = debug
--- a/CI-Examples/lighttpd/Makefile
+++ b/CI-Examples/lighttpd/Makefile
@@ -11,8 +11,8 @@ LIGHTTPD_MIRRORS ?= \
     https://download.lighttpd.net/lighttpd/releases-1.4.x \
     https://packages.gramineproject.io/distfiles
 
-HOST ?= 127.0.0.1
-PORT ?= 8003
+HOST ?= 10.6.0.6
+PORT ?= 8030
 
 ifeq ($(DEBUG),1)
 GRAMINE_LOG_LEVEL = debug
--- a/CI-Examples/lighttpd/lighttpd-generic.conf
+++ b/CI-Examples/lighttpd/lighttpd-generic.conf
@@ -2,7 +2,7 @@ server.event-handler = "select"
 #server.event-handler = "linux-sysepoll"
 server.max-connections = 4096
 # If keep-alive is undesired, uncomment the following line.
-#server.max-keep-alive-requests = 0
+server.max-keep-alive-requests = 0
 
 index-file.names = ( "index.html", "index.php" )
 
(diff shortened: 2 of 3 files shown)


kailun-qin (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1640)* status: 0 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @nniryanb)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1640#-NiY8ETH5TPzl_dgdgsZ:-NiY8ETH5TPzl_dgdgs_:b-k5wz62) (no related file):*
@nniryanb Thanks for the PR!

However, I think we basically consider [CI-Examples](https://github.com/gramineproject/gramine/tree/master/CI-Examples) as the sample applications to demonstarte the usage/usability of Gramine and for CI testing (as the name implies), but not really for performance considerations/references.

___

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1640)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @nniryanb)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1640#-NiY8ETH5TPzl_dgdgsZ:-NiYg_JkEcomHS-ou4E_:b-1a88w2) (no related file):*

@nniryanb Thanks for the PR!

However, I think we basically consider [CI-Examples](https://github.com/gramineproject/gramine/tree/master/CI-Examples) as the sample applications to demonstarte the usage/usability of Gramine and for CI testing (as the name implies), but not really for performance considerations/references.
</blockquote></details>

I agree with @kailun-qin.

We *could* update `lighttpd-generic.conf`, but only if you elaborate why these changes are required.

We definitely don't want to update `lighttpd.manifest.template`, because our examples are supposed to be debuggable (`sgx.debug = true`) and with minimal memory footprint (`sgx.enclave_size = "256M"`).

Also, I don't think `sgx.max_threads = 254` makes any sense -- isn't lighttpd **single-threaded**?

@nniryanb If you think these changes are relevant in some situations, I suggest to create a discussion under https://github.com/gramineproject/gramine/discussions and post your patch there, explaining why this is required and for what cases.

___
___
*[`-- commits` line 11 at r1](https://reviewable.io/reviews/gramineproject/gramine/1640#-NiYgWkpEh6wUj3kb0sW):*
Btw, there is no Makefile patch in this PR.

nniryanb:
it simply maxes out at 300 to 400 concurrent connections and causes timeouts with the default settings, with this small change I got it to perform perfectly at 1000 concurrent GETs with less than 1 ms response times, nothing is wrong, simply an attempt to make the first experience of gramine better for the beginner.

nniryanb:
and yes I reversed the Makefile edit, because the docs all point to 127.0.0.1, however it is best to assign a specific ip to bind to a specific network interface when testing.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1640)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @nniryanb)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1640#-NiY8ETH5TPzl_dgdgsZ:-Ni_GYjq5HzgPCxjEWdl:bjtx5bc) (no related file):*
> it simply maxes out at 300 to 400 concurrent connections and causes timeouts with the default settings, with this small change I got it to perform perfectly at 1000 concurrent GETs with less than 1 ms response times, nothing is wrong, simply an attempt to make the first experience of gramine better for the beginner.

Sure, I don't mind the changes in `lighttpd-generic.conf`. Would you be willing to fix the PR to keep only those changes, and also update the commit message with the explanation why these changes are beneficial (rewording of what you said above)?

___

nniryanb:
nevermind.

PR #1645, opened 2023-11-08
[PAL/Linux-SGX] Thread initialization error handling fixes

mwkmwkmwk:
## Description of the changes

A TODO within the code is fixed, adding proper error reporting in host-side `clone_thread`.  Additionally, a segfault in error handling is fixed.

## How to test this PR?

This PR fixes a test added by #1451.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1645)

### DIFF
--- a/pal/src/host/linux-sgx/host_thread.c
+++ b/pal/src/host/linux-sgx/host_thread.c
@@ -195,10 +195,6 @@ int pal_thread_init(void* tcbptr) {
     unmap_tcs();
     ret = 0;
 out:
-#ifdef ASAN
-    asan_unpoison_region((uintptr_t)tcb->stack, THREAD_STACK_SIZE + ALT_STACK_SIZE);
-#endif
-    DO_SYSCALL(munmap, tcb->stack, THREAD_STACK_SIZE + ALT_STACK_SIZE);
     return ret;
 }
 
--- a/pal/src/host/linux-sgx/host_thread.c
+++ b/pal/src/host/linux-sgx/host_thread.c
@@ -189,12 +189,16 @@ int pal_thread_init(void* tcbptr) {
         return 0;
     }
 
+    __atomic_store_n(tcb->start_status_ptr, 0, __ATOMIC_RELAXED);
+
     /* not-first (child) thread, start it */
     ecall_thread_start();
 
     unmap_tcs();
     ret = 0;
 out:
+    if (ret != 0)
+        __atomic_store_n(tcb->start_status_ptr, ret, __ATOMIC_RELAXED);
     return ret;
 }
 
@@ -271,8 +275,9 @@ int clone_thread(void) {
     /* align child_stack to 16 */
     child_stack_top = ALIGN_DOWN_PTR(child_stack_top, 16);
 
-    // TODO: pal_thread_init() may fail during initialization (e.g. on TCS exhaustion), we should
-    // check its result (but this happens asynchronously, so it's not trivial to do).
+    int start_status = 1;
+    tcb->start_status_ptr = &start_status;
+
     ret = clone(pal_thread_init, child_stack_top,
                 CLONE_VM | CLONE_FS | CLONE_FILES | CLONE_SYSVSEM | CLONE_THREAD | CLONE_SIGHAND,
                 tcb, /*parent_tid=*/NULL, /*tls=*/NULL, /*child_tid=*/NULL, thread_exit);
@@ -281,7 +286,11 @@ int clone_thread(void) {
         DO_SYSCALL(munmap, stack, THREAD_STACK_SIZE + ALT_STACK_SIZE);
         return ret;
     }
-    return 0;
+
+    while ((ret = __atomic_load_n(&start_status, __ATOMIC_RELAXED)) == 1)
+        CPU_RELAX();
+
+    return ret;
 }
 
 int get_tid_from_tcs(void* tcs) {
--- a/pal/src/host/linux-sgx/pal_tcb.h
+++ b/pal/src/host/linux-sgx/pal_tcb.h
@@ -104,6 +104,7 @@ typedef struct pal_host_tcb {
     atomic_ulong async_signal_cnt; /* # of async signals, corresponds to # of SIGINT/SIGCONT/.. */
     uint64_t profile_sample_time;  /* last time sgx_profile_sample() recorded a sample */
     int32_t last_async_event;      /* last async signal, reported to the enclave on ocall return */
+    int* start_status_ptr;         /* pointer to return value of clone_thread */
 } PAL_HOST_TCB;
 
 extern void pal_host_tcb_init(PAL_HOST_TCB* tcb, void* stack, void* alt_stack);


mkow (changes requested):
> Reviewed 1 of 1 files at r2, 1 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1645)* status: 1 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @mwkmwkmwk)

___
*[`pal/src/host/linux-sgx/host_thread.c` line 278 at r3](https://reviewable.io/reviews/gramineproject/gramine/1645#-NikFm-g-w57jkr4aKG_:-NikFm-g-w57jkr4aKGa:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/c5ea34e8868a8444159eb299e72588a1df4b5ae0/pal/src/host/linux-sgx/host_thread.c#L278)):*
> ```C
>     child_stack_top = ALIGN_DOWN_PTR(child_stack_top, 16);
> 
>     int start_status  = 1;
> ```





_Suggestion:_
```C
int start_status = 1;
```

mwkmwkmwk (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1645)* status: 1 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @mkow)

___
*[`pal/src/host/linux-sgx/host_thread.c` line 278 at r3](https://reviewable.io/reviews/gramineproject/gramine/1645#-NikFm-g-w57jkr4aKG_:-NikG8CKALFJrjR4MQkA:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/c5ea34e8868a8444159eb299e72588a1df4b5ae0/pal/src/host/linux-sgx/host_thread.c#L278)):*
> ```C
>     child_stack_top = ALIGN_DOWN_PTR(child_stack_top, 16);
> 
>     int start_status  = 1;
> ```

Done.

mkow (approved):
> Reviewed 1 of 2 files at r3, 1 of 1 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1645)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

mkow:
Jenkins, test this please

llly (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1645)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @mwkmwkmwk)

___
*[`pal/src/host/linux-sgx/pal_tcb.h` line 107 at r5](https://reviewable.io/reviews/gramineproject/gramine/1645#-NimcKXuCmbARR7kzgrE:-NimcKXuCmbARR7kzgrF:bx5cs26) ([raw file](https://github.com/gramineproject/gramine/blob/70cbe536dad086ffef4a086ca2fae6438b535f5c/pal/src/host/linux-sgx/pal_tcb.h#L107)):*
> ```objc
>     uint64_t profile_sample_time;  /* last time sgx_profile_sample() recorded a sample */
>     int32_t last_async_event;      /* last async signal, reported to the enclave on ocall return */
>     int* start_status_ptr;         /* pointer to return value of clone_thread */
> ```

Why use a pointer instead of store value here?

kailun-qin (changes requested):
> Reviewed 1 of 1 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1645)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @mwkmwkmwk)

___
*[`pal/src/host/linux-sgx/host_thread.c` line 189 at r5](https://reviewable.io/reviews/gramineproject/gramine/1645#-Nimdj8z8Cp2WEvKR1aj:-Nimdj8z8Cp2WEvKR1ak:b-fj2ow3) ([raw file](https://github.com/gramineproject/gramine/blob/70cbe536dad086ffef4a086ca2fae6438b535f5c/pal/src/host/linux-sgx/host_thread.c#L189)):*
> ```C
>         /* only first thread doesn't have a stack (it uses the one provided by Linux); first
>          * thread calls ecall_enclave_start() instead of ecall_thread_start() so just exit */
>         return 0;
> ```

hmm.. what about the first thread then?
___
*[`pal/src/host/linux-sgx/host_thread.c` line 201 at r5](https://reviewable.io/reviews/gramineproject/gramine/1645#-NimdwMR7SP_iWfZusM0:-NimdwMR7SP_iWfZusM1:b-tntffe) ([raw file](https://github.com/gramineproject/gramine/blob/70cbe536dad086ffef4a086ca2fae6438b535f5c/pal/src/host/linux-sgx/host_thread.c#L201)):*
> ```C
> out:
>     if (ret != 0)
>         __atomic_store_n(tcb->start_status_ptr, ret, __ATOMIC_RELAXED);
> ```

why not set `ret` blindly?



_Code quote:_
```C
    if (ret != 0)
        __atomic_store_n(tcb->start_status_ptr, ret, __ATOMIC_RELAXED);
```

dimakuv (changes requested):
> Reviewed 1 of 1 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1645)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @kailun-qin and @mwkmwkmwk)

___
*[`-- commits` line 12 at r5](https://reviewable.io/reviews/gramineproject/gramine/1645#-NinIDXfDikRvythSNtN):*
This message is pretty vague. Maybe we should add a commit body that says smth like:
```
Previously, `clone_thread()` only returned errors if the current-thread (parent)
side of cloning failed. This commit adds error reporting also in the case if the
newly spawned thread (child) side of cloning fails.
```
___
*[`pal/src/host/linux-sgx/host_thread.c` line 189 at r5](https://reviewable.io/reviews/gramineproject/gramine/1645#-Nimdj8z8Cp2WEvKR1aj:-NinFKqk7n5uIXlOAPJG:b7iv5hz) ([raw file](https://github.com/gramineproject/gramine/blob/70cbe536dad086ffef4a086ca2fae6438b535f5c/pal/src/host/linux-sgx/host_thread.c#L189)):*

hmm.. what about the first thread then?
</blockquote></details>

@kailun-qin What do you mean? Why the first thread doesn't update `start_status_ptr`? That's because the first thread is NOT called from `clone_thread()`. Instead, this func is called from `load_enclave()`.
___
*[`pal/src/host/linux-sgx/host_thread.c` line 197 at r5](https://reviewable.io/reviews/gramineproject/gramine/1645#-NinEow1DCnsdVnF67B1:-NinEow1DCnsdVnF67B2:brkdbj0) ([raw file](https://github.com/gramineproject/gramine/blob/70cbe536dad086ffef4a086ca2fae6438b535f5c/pal/src/host/linux-sgx/host_thread.c#L197)):*
> ```C
>     ecall_thread_start();
> 
>     unmap_tcs();
> ```

Since we're here, shouldn't we also remove this one? `ecall_thread_start()` is not supposed to return.
___
*[`pal/src/host/linux-sgx/host_thread.c` line 201 at r5](https://reviewable.io/reviews/gramineproject/gramine/1645#-NinDaGP-UeX9B6lwvmk:-NinDaGP-UeX9B6lwvml:b-78dnc4) ([raw file](https://github.com/gramineproject/gramine/blob/24a581c2f8e9a5b992af6610a68d2a45c0004273/pal/src/host/linux-sgx/host_thread.c#L201)):*
> ```C
>     asan_unpoison_region((uintptr_t)tcb->stack, THREAD_STACK_SIZE + ALT_STACK_SIZE);
> #endif
>     DO_SYSCALL(munmap, tcb->stack, THREAD_STACK_SIZE + ALT_STACK_SIZE);
> ```

Regarding removal of this `munmap()` in the first commit of this PR -- the logic is non-trivial, so I'll leave the references for future reviewers:
1. This func `pal_thread_init()` is invoked from the `clone()` func.
2. This `clone()` is a wrapper written in asm and **not** the same as Glibc wrapper/Linux syscall equivalent. See [our code](https://github.com/gramineproject/gramine/blob/24a581c2f8e9a5b992af6610a68d2a45c0004273/pal/src/host/linux-common/arch/x86_64/syscall.S#L31-L78).
3. The most important part is the last argument in `clone()` which is the "thread exit" function pointer.
4. So if `pal_thread_init()` fails, then we end up in this func pointer, which is wired to `thread_exit()`, which will do the deallocation of the thread stack.
___
*[`pal/src/host/linux-sgx/host_thread.c` line 201 at r5](https://reviewable.io/reviews/gramineproject/gramine/1645#-NimdwMR7SP_iWfZusM0:-NinFdBvEuR0I_ZeePzp:b-1niwqc) ([raw file](https://github.com/gramineproject/gramine/blob/70cbe536dad086ffef4a086ca2fae6438b535f5c/pal/src/host/linux-sgx/host_thread.c#L201)):*

why not set `ret` blindly?
</blockquote></details>

+1, but more importantly -- I don't think this function ever returns in the good case. So `ret == 0` is simply impossible.

I would modify this func to reflect this, and test this hypothesis with our CI -- but I'm 99% sure that's the case.
___
*[`pal/src/host/linux-sgx/pal_tcb.h` line 107 at r5](https://reviewable.io/reviews/gramineproject/gramine/1645#-NimcKXuCmbARR7kzgrE:-NinGAGl1PoOVIk7nkkR:b-gpidgg) ([raw file](https://github.com/gramineproject/gramine/blob/70cbe536dad086ffef4a086ca2fae6438b535f5c/pal/src/host/linux-sgx/pal_tcb.h#L107)):*

Why use a pointer instead of store value here?
</blockquote></details>

If I understand correct, it **must** be the pointer. If we would have a value here, then this value would be allocated on the newly-created-thread stack. And if the thread creation fails, then that thread will do `thread_exit()` and unmap its own stack. So the value storage will be lost, and we'll have use-after-free.

So we need to allocate storage for this "thread status" value somewhere else. The obvious choice is the parent thread's stack. And in this case we need to communicate to the child thread the address which must contain this "thread status".

So to me everything looks correct and the only way possible. @mwkmwkmwk please correct me if I said something wrong.

kailun-qin (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1645)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @mwkmwkmwk)

___
*[`pal/src/host/linux-sgx/host_thread.c` line 189 at r5](https://reviewable.io/reviews/gramineproject/gramine/1645#-Nimdj8z8Cp2WEvKR1aj:-NinJVC54tQEKx41d079:b-u8v59b) ([raw file](https://github.com/gramineproject/gramine/blob/70cbe536dad086ffef4a086ca2fae6438b535f5c/pal/src/host/linux-sgx/host_thread.c#L189)):*

@kailun-qin What do you mean? Why the first thread doesn't update `start_status_ptr`? That's because the first thread is NOT called from `clone_thread()`. Instead, this func is called from `load_enclave()`.
</blockquote></details>

Yes, I understand this. But `start_status_ptr` is defined in `pal_host_tcb` and we don't even init this field (for the first thread)?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1645)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @kailun-qin and @mwkmwkmwk)

___
*[`pal/src/host/linux-sgx/host_thread.c` line 189 at r5](https://reviewable.io/reviews/gramineproject/gramine/1645#-Nimdj8z8Cp2WEvKR1aj:-NinOMuY3hZgk0GzoxJ7:b-zhoh23) ([raw file](https://github.com/gramineproject/gramine/blob/70cbe536dad086ffef4a086ca2fae6438b535f5c/pal/src/host/linux-sgx/host_thread.c#L189)):*

Yes, I understand this. But `start_status_ptr` is defined in `pal_host_tcb` and we don't even init this field (for the first thread)?
</blockquote></details>

Well, that's true. However, we do this for the main thread: https://github.com/gramineproject/gramine/blob/24a581c2f8e9a5b992af6610a68d2a45c0004273/pal/src/host/linux-sgx/host_main.c#L1051-L1060

So the TCB of the first thread is located inside the alt-stack-allocated region. And since we do `mmap()`, the whole TCB is zero-filled. Thus, `start_status_ptr == 0x0`, which I find reasonable.

We could of course make it very explicit by assigning this field in `pal_host_tcb_init()`, but to me this is an overkill. I'm still fine with the current state of the PR.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1645)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @mwkmwkmwk)

___
*[`pal/src/host/linux-sgx/host_thread.c` line 189 at r5](https://reviewable.io/reviews/gramineproject/gramine/1645#-Nimdj8z8Cp2WEvKR1aj:-NinRPH81vk-YWKDZQuo:b-wzdcuo) ([raw file](https://github.com/gramineproject/gramine/blob/70cbe536dad086ffef4a086ca2fae6438b535f5c/pal/src/host/linux-sgx/host_thread.c#L189)):*
> And since we do `mmap()`, the whole TCB is zero-filled. Thus, `start_status_ptr == 0x0`, which I find reasonable.

Ah Ok, if mmaped then it makes sense to me. I'm resolving.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1645)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv, @kailun-qin, and @mwkmwkmwk)

___
*[`pal/src/host/linux-sgx/host_thread.c` line 197 at r5](https://reviewable.io/reviews/gramineproject/gramine/1645#-NinEow1DCnsdVnF67B1:-Niou7X0BApGDCcPt6et:b-bl79mb) ([raw file](https://github.com/gramineproject/gramine/blob/70cbe536dad086ffef4a086ca2fae6438b535f5c/pal/src/host/linux-sgx/host_thread.c#L197)):*

Since we're here, shouldn't we also remove this one? `ecall_thread_start()` is not supposed to return.
</blockquote></details>

Is it? We actually had a discussion about this with mwk and we weren't sure. It's not declared as `noreturn`, it just happens that (probably) all exit paths from threads go through `PalThreadExit()`.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1645)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @kailun-qin and @mwkmwkmwk)

___
*[`pal/src/host/linux-sgx/host_thread.c` line 197 at r5](https://reviewable.io/reviews/gramineproject/gramine/1645#-NinEow1DCnsdVnF67B1:-NipTFKbB0JCNh_RA3Y1:bm8f070) ([raw file](https://github.com/gramineproject/gramine/blob/70cbe536dad086ffef4a086ca2fae6438b535f5c/pal/src/host/linux-sgx/host_thread.c#L197)):*

Is it? We actually had a discussion about this with mwk and we weren't sure. It's not declared as `noreturn`, it just happens that (probably) all exit paths from threads go through `PalThreadExit()`.
</blockquote></details>

Hm, ok, indeed, it's hard to argue about this case. Let's not deal with it now.
___
*[`pal/src/host/linux-sgx/host_thread.c` line 201 at r5](https://reviewable.io/reviews/gramineproject/gramine/1645#-NimdwMR7SP_iWfZusM0:-NipTMk49zk1fEOxmTwL:bv95982) ([raw file](https://github.com/gramineproject/gramine/blob/70cbe536dad086ffef4a086ca2fae6438b535f5c/pal/src/host/linux-sgx/host_thread.c#L201)):*

+1, but more importantly -- I don't think this function ever returns in the good case. So `ret == 0` is simply impossible.

I would modify this func to reflect this, and test this hypothesis with our CI -- but I'm 99% sure that's the case.
</blockquote></details>

Let me revert my comment. I guess we should keep the `ret == 0` path, see my other comment.

mwkmwkmwk (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1645)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @kailun-qin)

___
*[`-- commits` line 12 at r5](https://reviewable.io/reviews/gramineproject/gramine/1645#-NinIDXfDikRvythSNtN):*

This message is pretty vague. Maybe we should add a commit body that says smth like:
```
Previously, `clone_thread()` only returned errors if the current-thread (parent)
side of cloning failed. This commit adds error reporting also in the case if the
newly spawned thread (child) side of cloning fails.
```
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/host_thread.c` line 201 at r5](https://reviewable.io/reviews/gramineproject/gramine/1645#-NimdwMR7SP_iWfZusM0:-Niq7Kwl-tbeIV6Kzsp_:b9i9x6y) ([raw file](https://github.com/gramineproject/gramine/blob/70cbe536dad086ffef4a086ca2fae6438b535f5c/pal/src/host/linux-sgx/host_thread.c#L201)):*

Let me revert my comment. I guess we should keep the `ret == 0` path, see my other comment.
</blockquote></details>

In `ret == 0` case (if it is even possible), the store has already been performed right before `ecall_thread_start`, which unblocked the waiting thread. Since the waiting thread has already proceeded, the pointer is now dangling (it points to a variable on the parent's stack).

If we want to remove this conditional, we need to properly mark `ecall_thread_start` as noreturn and remove the other logic.
___
*[`pal/src/host/linux-sgx/pal_tcb.h` line 107 at r5](https://reviewable.io/reviews/gramineproject/gramine/1645#-NimcKXuCmbARR7kzgrE:-Niq7nKi1Z8S_dpHwA2R:bs228t) ([raw file](https://github.com/gramineproject/gramine/blob/70cbe536dad086ffef4a086ca2fae6438b535f5c/pal/src/host/linux-sgx/pal_tcb.h#L107)):*

If I understand correct, it **must** be the pointer. If we would have a value here, then this value would be allocated on the newly-created-thread stack. And if the thread creation fails, then that thread will do `thread_exit()` and unmap its own stack. So the value storage will be lost, and we'll have use-after-free.

So we need to allocate storage for this "thread status" value somewhere else. The obvious choice is the parent thread's stack. And in this case we need to communicate to the child thread the address which must contain this "thread status".

So to me everything looks correct and the only way possible. @mwkmwkmwk please correct me if I said something wrong.
</blockquote></details>

This is correct.

kailun-qin (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1645)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required) (waiting on @dimakuv)

___
*[`pal/src/host/linux-sgx/host_thread.c` line 201 at r5](https://reviewable.io/reviews/gramineproject/gramine/1645#-NimdwMR7SP_iWfZusM0:-NiqwUmy4pEUMvdT0Cg6:bgfd3jn) ([raw file](https://github.com/gramineproject/gramine/blob/70cbe536dad086ffef4a086ca2fae6438b535f5c/pal/src/host/linux-sgx/host_thread.c#L201)):*

In `ret == 0` case (if it is even possible), the store has already been performed right before `ecall_thread_start`, which unblocked the waiting thread. Since the waiting thread has already proceeded, the pointer is now dangling (it points to a variable on the parent's stack).

If we want to remove this conditional, we need to properly mark `ecall_thread_start` as noreturn and remove the other logic.
</blockquote></details>

Thanks, make sense to me.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1645)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #1651, opened 2023-11-13
Fix Coverity issues

jkr0103:
## Description of the changes 

PR fixes Coverity issue described below:

CID 328056: Infinite loop
The loop does not have a normal termination condition, so will continue until an abnormal condition arises.

In clear_pollable_event: Infinite loop with unsatisfiable exit condition (CWE-835)

loop_bottom Bottom of the loop.
loop_condition If ret == -4 is initially true then it will remain true

https://github.com/gramineproject/gramine/blob/master/libos/src/libos_pollable_event.c#L74-L86

## How to test this PR? 

Sanity run should do the job?

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1651)

### DIFF
--- a/libos/src/libos_pollable_event.c
+++ b/libos/src/libos_pollable_event.c
@@ -97,7 +97,7 @@ int clear_pollable_event(struct libos_pollable_event* event) {
     do {
         char buf[0x100];
         size_t size = sizeof(buf);
-        int ret = PalStreamRead(event->read_handle, /*offset=*/0, &size, buf);
+        ret = PalStreamRead(event->read_handle, /*offset=*/0, &size, buf);
         ret = pal_to_unix_errno(ret);
         if (ret == 0 && size == 0) {
             ret = -EINVAL;


mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1651)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @jkr0103)

___
*[`pal/src/host/linux-sgx/host_main.c` line 512 at r1](https://reviewable.io/reviews/gramineproject/gramine/1651#-Nj9wzEWCFlJmyIkEqfI:-Nj9wzEWCFlJmyIkEqfJ:bja6p0k) ([raw file](https://github.com/gramineproject/gramine/blob/d3e9b325608328d6290a4ea2c86154318ce165ad/pal/src/host/linux-sgx/host_main.c#L512)):*
> ```C
>             for (uint32_t t = 0; t < enclave->thread_num; t++) {
>                 struct pal_enclave_tcb* gs = data + g_page_size * t;
>                 memset(gs, 0, g_page_size);
> ```

Why did you remove this?

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1651)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

___
*[`libos/src/libos_pollable_event.c` line 100 at r1](https://reviewable.io/reviews/gramineproject/gramine/1651#-NjBh6X6320eiS7SyIus:-NjBh6X74s6B1NoqQhVK:bqonaa3) ([raw file](https://github.com/gramineproject/gramine/blob/d3e9b325608328d6290a4ea2c86154318ce165ad/libos/src/libos_pollable_event.c#L100)):*
> ```C
>         char buf[0x100];
>         size_t size = sizeof(buf);
>         int ret = PalStreamRead(event->read_handle, /*offset=*/0, &size, buf);
> ```

FYI: This looks like a typo. Please also note that `set_pollable_event()` function above has the exact same structure, and it **didn't** have this `int ret` inside the loop, which confirms the suspicion about this `int ret`.
___
*[`pal/src/host/linux-sgx/host_main.c` line 512 at r1](https://reviewable.io/reviews/gramineproject/gramine/1651#-Nj9wzEWCFlJmyIkEqfI:-NjBhPrgAnX5w1i2kB32:bcgaxxe) ([raw file](https://github.com/gramineproject/gramine/blob/d3e9b325608328d6290a4ea2c86154318ce165ad/pal/src/host/linux-sgx/host_main.c#L512)):*

Why did you remove this?
</blockquote></details>

This looks legit.

This particular removed `memset()` has a "bug" as described in the PR description (size mismatch between `gs` first argument and `g_page_size` last argument).

It could be fixed by casting to void: `memset((void*)gs, 0, g_page_size)`. But even this is not necessary, as this `gs` is taken from the `data` variable which is mmapped 5 lines above. And since mmap always returns zeroed-out memory, there is no need for an additional memset-to-zero. So looks fine to me.

kailun-qin (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1651)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1651)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @jkr0103)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/1651#-NjDVc-u8L0mWV8V1REB):*
This commit message is not very informative, we don't write e.g. "Fix bug found by mkow" but "Fix a bug causing malloc() to hang" - someone reading the commit messages doesn't care who found the bug, but what bug is actually fixed by this commit. Especially given that the first bug is legit (the second IMO should be dropped from the PR).
___
*[`pal/src/host/linux-sgx/host_main.c` line 512 at r1](https://reviewable.io/reviews/gramineproject/gramine/1651#-Nj9wzEWCFlJmyIkEqfI:-NjDW39CDLUQfAdH0W02:b-2avnl2) ([raw file](https://github.com/gramineproject/gramine/blob/d3e9b325608328d6290a4ea2c86154318ce165ad/pal/src/host/linux-sgx/host_main.c#L512)):*

This looks legit.

This particular removed `memset()` has a "bug" as described in the PR description (size mismatch between `gs` first argument and `g_page_size` last argument).

It could be fixed by casting to void: `memset((void*)gs, 0, g_page_size)`. But even this is not necessary, as this `gs` is taken from the `data` variable which is mmapped 5 lines above. And since mmap always returns zeroed-out memory, there is no need for an additional memset-to-zero. So looks fine to me.
</blockquote></details>

I don't understand. Where's the mismatch? The purpose of this memset is to zero the page which is given to this specific gs entry. The old code looks correct to me.
Also, the cast you suggest changes nothing? memset takes `void*`, so you can pass any pointer without casting.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1651)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @jkr0103 and @mkow)

___
*[`pal/src/host/linux-sgx/host_main.c` line 512 at r1](https://reviewable.io/reviews/gramineproject/gramine/1651#-Nj9wzEWCFlJmyIkEqfI:-NjDXKgf1Kx1XjBn4Wbn:b-3x6ldg) ([raw file](https://github.com/gramineproject/gramine/blob/d3e9b325608328d6290a4ea2c86154318ce165ad/pal/src/host/linux-sgx/host_main.c#L512)):*
> Also, the cast you suggest changes nothing? memset takes `void*`, so you can pass any pointer without casting.

It makes the Coverity static analyzer happy...

Anyway, my main point was that we can just remove this `memset()` as it is redundant. This solves two problems:
1. One less redundant code
2. Coverity is happy

So I don't see why not to remove this line.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1651)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @jkr0103)

___
*[`pal/src/host/linux-sgx/host_main.c` line 512 at r1](https://reviewable.io/reviews/gramineproject/gramine/1651#-Nj9wzEWCFlJmyIkEqfI:-NjDuRDi4_NcQ6VhPaL4:b-z85cdv) ([raw file](https://github.com/gramineproject/gramine/blob/d3e9b325608328d6290a4ea2c86154318ce165ad/pal/src/host/linux-sgx/host_main.c#L512)):*
> It makes the Coverity static analyzer happy...

That's not a good argument to change the logic of the code.

> So I don't see why not to remove this line.

I believe the original intention was to ensure that the whole struct is zeroed and we don't leave any garbage there (there won't be any warning if we miss initialization of some of the fields of `struct pal_enclave_tcb`).
Actually, there are some pointers in `struct pal_enclave_tcb` which this code doesn't initialize. I definitely don't want to risk introducing a security vuln just to make Coverity happy about some bogus finding.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1651)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @jkr0103 and @mkow)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/1651#-NjDVc-u8L0mWV8V1REB):*

This commit message is not very informative, we don't write e.g. "Fix bug found by mkow" but "Fix a bug causing malloc() to hang" - someone reading the commit messages doesn't care who found the bug, but what bug is actually fixed by this commit. Especially given that the first bug is legit (the second IMO should be dropped from the PR).
</blockquote></details>

I think this commit message is good enough:
```
[LibOS] Fix `ret` variable shadowing in `clear_pollable_event()`

There was a typo that accidentally declared `int ret` in a while-loop scope
and thus shadowed the `ret` variable in the function scope. This led to
the bogus loop (it would always execute only once).
```
___
*[`pal/src/host/linux-sgx/host_main.c` line 512 at r1](https://reviewable.io/reviews/gramineproject/gramine/1651#-Nj9wzEWCFlJmyIkEqfI:-NjECuQG8TfnP34Iix-_:b-3w253p) ([raw file](https://github.com/gramineproject/gramine/blob/d3e9b325608328d6290a4ea2c86154318ce165ad/pal/src/host/linux-sgx/host_main.c#L512)):*

> It makes the Coverity static analyzer happy...

That's not a good argument to change the logic of the code.

> So I don't see why not to remove this line.

I believe the original intention was to ensure that the whole struct is zeroed and we don't leave any garbage there (there won't be any warning if we miss initialization of some of the fields of `struct pal_enclave_tcb`).
Actually, there are some pointers in `struct pal_enclave_tcb` which this code doesn't initialize. I definitely don't want to risk introducing a security vuln just to make Coverity happy about some bogus finding.
</blockquote></details>

Ok, I don't have a strong feeling about it. I'm fine with dropping this change from the PR.

@jkr0103 Could you please revert this change? And mark it as False Positive in the Coverity report, explaining that we do this on purpose -- zeroing out the whole page with this struct.

jkr0103 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1651)* status: 1 of 2 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @kailun-qin, and @mkow)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/1651#-NjDVc-u8L0mWV8V1REB):*

I think this commit message is good enough:
```
[LibOS] Fix `ret` variable shadowing in `clear_pollable_event()`

There was a typo that accidentally declared `int ret` in a while-loop scope
and thus shadowed the `ret` variable in the function scope. This led to
the bogus loop (it would always execute only once).
```
</blockquote></details>

@dimakuv Could you please take care of it while merging the PR?
___
*[`pal/src/host/linux-sgx/host_main.c` line 512 at r1](https://reviewable.io/reviews/gramineproject/gramine/1651#-Nj9wzEWCFlJmyIkEqfI:-NjKaZGC7_lKJsjQ8oQS:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/d3e9b325608328d6290a4ea2c86154318ce165ad/pal/src/host/linux-sgx/host_main.c#L512)):*

Ok, I don't have a strong feeling about it. I'm fine with dropping this change from the PR.

@jkr0103 Could you please revert this change? And mark it as False Positive in the Coverity report, explaining that we do this on purpose -- zeroing out the whole page with this struct.
</blockquote></details>

Done.

mkow:
Jenkins, test this please

kailun-qin (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1651)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1651)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/1651#-NjDVc-u8L0mWV8V1REB):*
<details><summary><i>Previously, jkr0103 (Jitender Kumar) wrote…</i></summary><blockquote>

@dimakuv Could you please take care of it while merging the PR?
</blockquote></details>

Yes

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1651)* status: 1 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required) (waiting on @kailun-qin and @mkow)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/1651#-NjDVc-u8L0mWV8V1REB):*

Yes
</blockquote></details>

Done

BTW, I also updated the PR description to reflect that we dropped the "fix" for the other item.

dimakuv (approved):
> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1651)* status: all files reviewed, 1 unresolved discussion (waiting on @mkow)

dimakuv:
Jenkins, test this please

kailun-qin (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1651)* status: all files reviewed, 1 unresolved discussion (waiting on @mkow)

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1651)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #1681, opened 2023-12-13
[LibOS,Pal,common,tools] Improved protected files performance

### LINKED ISSUE #1599: Improve protected files performance

Current implementation of protected file is slow, especially for large files. At least the following 2 improvements can be applied.

1. Use file map to replace read/write a single node from/to disk.
2. Remove the `encrypted` part inside the `file_node_t` structure. The encrypted content can be directly retrieved from disk. The encryption/decryption functions can use the file mapped addresses instead so we can save the cache memory as well as the time used to copy it to/from disk.

Performance test on SGX SDK shows that applying these two changes can make reading 2 times fasters and writing 7 times faster.

yao-ji:
## Description of the changes 

Fixes #1599 

### LibOS
- Removed `cb_read`, `cb_write` callback. Add `cb_truncate` callback to truncate and remap file when file size changed.
- Redo the file map in the `rs_encrypted_file` function.

### Pal
- Add additional pal option flag `PAL_OPTION_ENCRYPTED_FILE` to indicate opening an encrypted file when using `PalStreamOpen`. 
- Add a flag `encrypted` and a pointer `addr` storing mapping address to the fields of `PAL_HANDLE`.
- Add pointer `addr` in `PAL_STREAM_ATTR`.
- In `file_open`, map the file and store address in `addr` field. Unmap the file in `file_destory`. Remap the file in `file_setlength`.
- Set and get `addr` field in `file_attrsetbyhdl` and `file_attrquerybyhdl`.

### common
Improve performance of `protected files` with the following changes:
- Replace reading/writing a single node with directly `memcpy` from/to the mapping address.
- Removed `encrypted` part inside `file_node_t` structure. The encrypted content is directly retrived from the file mapped memory.

### tools
Modify `pf_util` similar to the changes in LibOS and Pal.

### Performance test
In release mode, read is around 60% faster, and write is around 35% faster.

**Read**
<html>
<body">

File Size | 10KB | 100KB | 1MB | 10MB | 100MB | 1GB
-- | -- | -- | -- | -- | -- | --
Original(KB/s） | 96242 | 80569 | 81528 | 81259 | 80615 | 80004
Improved(KB/s) | 145229 | 134960 | 131873 | 130959 | 130208 | 129321
Percentage | 150.90% | 167.51% | 161.75% | 161.16% | 161.52% | 161.64%

</body>
</html>

**Write**
<html>
<body>

File Size | 10KB | 100KB | 1MB | 10MB | 100MB | 1GB
-- | -- | -- | -- | -- | -- | --
Original(KB/s） | 23699 | 26575 | 35626 | 40681 | 40257 | 39478
Improved(KB/s) | 34842 | 35779 | 50670 | 54829 | 54454 | 51478
Percentage | 147.02% | 134.63% | 142.23% | 134.78% | 135.27% | 130.40%

</body>
</html>

## How to test this PR? 

Using LibOS and Pal regression test. `gramine-sgx-pf-crypt` and `gramine-sgx-pf-tamper` can be used to test `pf_util`. 

Performance test result is from [iozone](https://www.iozone.org/) using `fread/fwrite`.

### DIFF
--- a/common/src/protected_files/protected_files.c
+++ b/common/src/protected_files/protected_files.c
@@ -11,8 +11,6 @@
 #include "api.h"
 
 /* Host callbacks */
-static pf_read_f     g_cb_read     = NULL;
-static pf_write_f    g_cb_write    = NULL;
 static pf_truncate_f g_cb_truncate = NULL;
 static pf_debug_f    g_cb_debug    = NULL;
 
@@ -60,6 +58,8 @@ static const char* g_pf_error_list[] = {
     [-PF_STATUS_CRYPTO_ERROR] = "Crypto error",
     [-PF_STATUS_CORRUPTED] = "File is corrupted",
     [-PF_STATUS_WRITE_TO_DISK_FAILED] = "Write to disk failed",
+    [-PF_STATUS_INVALID_ADDRESS] = "Invalid file mapping address",
+    [-PF_STATUS_INVALID_FILE_SIZE] = "File mapping size incorrect",
 };
 
 /* ipf prefix means "Intel protected files", these are functions from the SGX SDK implementation */
@@ -206,9 +206,45 @@ static bool ipf_update_all_data_and_mht_nodes(pf_context_t* pf) {
     pf_status_t status;
     void* data = lruc_get_first(pf->cache);
 
+    if (pf->addr == NULL) {
+        pf->last_error = PF_STATUS_INVALID_ADDRESS;
+        return false;
+    }
+
+    // find the largest node number in LRU cache
+    uint64_t max_node_number = 0;
+    while (data != NULL) {
+        if (((file_node_t*)data)->type == FILE_DATA_NODE_TYPE) {
+            file_node_t* data_node = (file_node_t*)data;
+            if (data_node->need_writing) {
+                max_node_number = MAX(max_node_number, data_node->physical_node_number);
+            }
+        }
+        data = lruc_get_next(pf->cache);
+    }
+
+    // truncate file to larger size
+    uint64_t max_file_size = PF_NODE_SIZE * (max_node_number + 1);
+    if (pf->real_file_size < max_file_size) {
+        void* ret_addr = NULL;
+        status = g_cb_truncate(pf->file, max_file_size, &ret_addr);
+        if (PF_FAILURE(status)) {
+            pf->last_error = status;
+            return false;
+        }
+        if (!ret_addr) {
+            pf->last_error = PF_STATUS_INVALID_ADDRESS;
+            return false;
+        }
+
+        pf->addr = ret_addr;
+        pf->real_file_size = max_file_size;
+    }
+
     // 1. encrypt the changed data
     // 2. set the IV+GMAC in the parent MHT
     // [3. set the need_writing flag for all the parents]
+    data = lruc_get_first(pf->cache);
     while (data != NULL) {
         if (((file_node_t*)data)->type == FILE_DATA_NODE_TYPE) {
             file_node_t* data_node = (file_node_t*)data;
@@ -219,17 +255,22 @@ static bool ipf_update_all_data_and_mht_nodes(pf_context_t* pf) {
                          .data_nodes_crypto[data_node->node_number % ATTACHED_DATA_NODES_COUNT];
 
                 if (!ipf_generate_random_key(pf, &gcm_crypto_data->key))
-                    goto out;
+                    return false;
 
                 // encrypt the data, this also saves the gmac of the operation in the mht crypto
                 // node
+                void* data_node_addr = (unsigned char*)pf->addr + 
+                                       PF_NODE_SIZE * data_node->physical_node_number;
                 status = g_cb_aes_gcm_encrypt(&gcm_crypto_data->key, &g_empty_iv, NULL, 0,  // aad
                                               data_node->decrypted.data.data, PF_NODE_SIZE,
-                                              data_node->encrypted.cipher, &gcm_crypto_data->gmac);
+                                              data_node_addr, &gcm_crypto_data->gmac);
                 if (PF_FAILURE(status)) {
                     pf->last_error = status;
-                    goto out;
+                    return false;
                 }
+
+                data_node->need_writing = false;
+                data_node->new_node = false;
 
 #ifdef DEBUG
                 file_node_t* file_mht_node = data_node->parent;
@@ -291,13 +332,18 @@ static bool ipf_update_all_data_and_mht_nodes(pf_context_t* pf) {
             goto out;
         }
 
+        void* mht_node_addr = (unsigned char*)pf->addr + 
+                                PF_NODE_SIZE * file_mht_node->physical_node_number;
         status = g_cb_aes_gcm_encrypt(&gcm_crypto_data->key, &g_empty_iv, NULL, 0,
                                       &file_mht_node->decrypted.mht, PF_NODE_SIZE,
-                                      &file_mht_node->encrypted.cipher, &gcm_crypto_data->gmac);
+                                      mht_node_addr, &gcm_crypto_data->gmac);
         if (PF_FAILURE(status)) {
             pf->last_error = status;
             goto out;
         }
+
+        file_mht_node->need_writing = false;
+        file_mht_node->new_node = false;
     }
 
     // update mht root gmac in the meta data node
@@ -307,47 +353,21 @@ static bool ipf_update_all_data_and_mht_nodes(pf_context_t* pf) {
     status = g_cb_aes_gcm_encrypt(&pf->encrypted_part_plain.mht_key, &g_empty_iv,
                                   NULL, 0,
                                   &pf->root_mht.decrypted.mht, PF_NODE_SIZE,
-                                  &pf->root_mht.encrypted.cipher,
+                                  (unsigned char*)pf->addr + PF_NODE_SIZE,
                                   &pf->encrypted_part_plain.mht_gmac);
     if (PF_FAILURE(status)) {
         pf->last_error = status;
         goto out;
     }
 
+    pf->root_mht.need_writing = false;
+    pf->root_mht.new_node = false;
+
     ret = true;
 
 out:
     free(mht_array);
     return ret;
-}
-
-static bool ipf_read_node(pf_context_t* pf, pf_handle_t handle, uint64_t node_number, void* buffer,
-                          uint32_t node_size) {
-    uint64_t offset = node_number * node_size;
-
-    pf_status_t status = g_cb_read(handle, buffer, offset, node_size);
-    if (PF_FAILURE(status)) {
-        pf->last_error = status;
-        return false;
-    }
-
-    return true;
-}
-
-static bool ipf_write_file(pf_context_t* pf, pf_handle_t handle, uint64_t offset, void* buffer,
-                           uint32_t size) {
-    pf_status_t status = g_cb_write(handle, buffer, offset, size);
-    if (PF_FAILURE(status)) {
-        pf->last_error = status;
-        return false;
-    }
-
-    return true;
-}
-
-static bool ipf_write_node(pf_context_t* pf, pf_handle_t handle, uint64_t node_number, void* buffer,
-                           uint32_t node_size) {
-    return ipf_write_file(pf, handle, node_number * node_size, buffer, node_size);
 }
 
 // this is a very 'specific' function, tied to the architecture of the file layout,
@@ -393,45 +413,6 @@ static void get_node_numbers(uint64_t offset, uint64_t* mht_node_number, uint64_
         *physical_data_node_number = _physical_data_node_number;
 }
 
-static bool ipf_write_all_changes_to_disk(pf_context_t* pf) {
-    if (pf->encrypted_part_plain.size > MD_USER_DATA_SIZE && pf->root_mht.need_writing) {
-        void* data = NULL;
-        uint8_t* data_to_write;
-        uint64_t node_number;
-        file_node_t* file_node;
-
-        for (data = lruc_get_first(pf->cache); data != NULL; data = lruc_get_next(pf->cache)) {
-            file_node = (file_node_t*)data;
-            if (!file_node->need_writing)
-                continue;
-
-            data_to_write = (uint8_t*)&file_node->encrypted;
-            node_number = file_node->physical_node_number;
-
-            if (!ipf_write_node(pf, pf->file, node_number, data_to_write, PF_NODE_SIZE)) {
-                return false;
-            }
-
-            file_node->need_writing = false;
-            file_node->new_node = false;
-        }
-
-        if (!ipf_write_node(pf, pf->file, /*node_number=*/1, &pf->root_mht.encrypted,
-                            PF_NODE_SIZE)) {
-            return false;
-        }
-
-        pf->root_mht.need_writing = false;
-        pf->root_mht.new_node = false;
-    }
-
-    if (!ipf_write_node(pf, pf->file, /*node_number=*/0, &pf->file_metadata, PF_NODE_SIZE)) {
-        return false;
-    }
-
-    return true;
-}
-
 static bool ipf_update_metadata_node(pf_context_t* pf) {
     pf_status_t status;
     pf_key_t key;
@@ -450,6 +431,36 @@ static bool ipf_update_metadata_node(pf_context_t* pf) {
         pf->last_error = status;
         return false;
     }
+
+    if (pf->addr == NULL) {
+        if (pf->real_file_size != 0) { // some error happened
+            pf->last_error = PF_STATUS_INVALID_ADDRESS;
+            return false;
+        }
+
+        // real file size is 0, truncate file to metadata node size
+        void* ret_addr = NULL;
+        status = g_cb_truncate(pf->file, PF_NODE_SIZE, &ret_addr);
+        if (PF_FAILURE(status)) {
+            pf->last_error = status;
+            return false;
+        }
+        if (!ret_addr) {
+            pf->last_error = PF_STATUS_INVALID_ADDRESS;
+            return false;
+        }
+
+        pf->addr = ret_addr;
+        pf->real_file_size = PF_NODE_SIZE;
+    }
+
+    if (pf->real_file_size < PF_NODE_SIZE) {
+        pf->last_error = PF_STATUS_INVALID_FILE_SIZE;
+        return false;
+    }
+
+    // write meta-data node to disk
+    memcpy(pf->addr, &pf->file_metadata, PF_NODE_SIZE);
 
     return true;
 }
@@ -475,13 +486,6 @@ static bool ipf_internal_flush(pf_context_t* pf) {
         // this is something that shouldn't happen, can't fix this...
         pf->file_status = PF_STATUS_CRYPTO_ERROR;
         DEBUG_PF("failed to update metadata nodes");
-        return false;
-    }
-
-    if (!ipf_write_all_changes_to_disk(pf)) {
-        pf->file_status = PF_STATUS_WRITE_TO_DISK_FAILED;
-
-        DEBUG_PF("failed to write changes to disk");
         return false;
     }
 
@@ -659,6 +663,18 @@ static file_node_t* ipf_read_data_node(pf_context_t* pf, uint64_t offset) {
     if (file_mht_node == NULL) // some error happened
         return NULL;
 
+    if (pf->addr == NULL) {
+        pf->last_error = PF_STATUS_INVALID_ADDRESS;
+        return NULL;
+    }
+
+    if (pf->real_file_size < PF_NODE_SIZE * (physical_node_number + 1)) {
+        pf->last_error = PF_STATUS_INVALID_FILE_SIZE;
+        return NULL;
+    }
+
+    void* file_data_node_addr = (unsigned char*)pf->addr + PF_NODE_SIZE * physical_node_number;
+
     file_data_node = calloc(1, sizeof(*file_data_node));
     if (!file_data_node) {
         pf->last_error = PF_STATUS_NO_MEMORY;
@@ -670,19 +686,13 @@ static file_node_t* ipf_read_data_node(pf_context_t* pf, uint64_t offset) {
     file_data_node->physical_node_number = physical_node_number;
     file_data_node->parent = file_mht_node;
 
-    if (!ipf_read_node(pf, pf->file, file_data_node->physical_node_number,
-                       file_data_node->encrypted.cipher, PF_NODE_SIZE)) {
-        free(file_data_node);
-        return NULL;
-    }
-
     gcm_crypto_data_t* gcm_crypto_data =
         &file_data_node->parent->decrypted.mht
              .data_nodes_crypto[file_data_node->node_number % ATTACHED_DATA_NODES_COUNT];
 
     // this function decrypt the data _and_ checks the integrity of the data against the gmac
     status = g_cb_aes_gcm_decrypt(&gcm_crypto_data->key, &g_empty_iv, NULL, 0,
-                                  file_data_node->encrypted.cipher, PF_NODE_SIZE,
+                                  file_data_node_addr, PF_NODE_SIZE,
                                   file_data_node->decrypted.data.data, &gcm_crypto_data->gmac);
 
     if (PF_FAILURE(status)) {
@@ -724,6 +734,18 @@ static file_node_t* ipf_read_mht_node(pf_context_t* pf, uint64_t mht_node_number
     if (parent_file_mht_node == NULL) // some error happened
         return NULL;
 
+    if (pf->addr == NULL) {
+        pf->last_error = PF_STATUS_INVALID_ADDRESS;
+        return NULL;
+    }
+
+    if (pf->real_file_size < PF_NODE_SIZE * (physical_node_number + 1)) {
+        pf->last_error = PF_STATUS_INVALID_FILE_SIZE;
+        return NULL;
+    }
+
+    void* file_mht_node_addr = (unsigned char*)pf->addr + PF_NODE_SIZE * physical_node_number;
+
     file_mht_node = calloc(1, sizeof(*file_mht_node));
     if (!file_mht_node) {
         pf->last_error = PF_STATUS_NO_MEMORY;
@@ -735,19 +757,13 @@ static file_node_t* ipf_read_mht_node(pf_context_t* pf, uint64_t mht_node_number
     file_mht_node->physical_node_number = physical_node_number;
     file_mht_node->parent               = parent_file_mht_node;
 
-    if (!ipf_read_node(pf, pf->file, file_mht_node->physical_node_number,
-                       file_mht_node->encrypted.cipher, PF_NODE_SIZE)) {
-        free(file_mht_node);
-        return NULL;
-    }
-
     gcm_crypto_data_t* gcm_crypto_data =
         &file_mht_node->parent->decrypted.mht
              .mht_nodes_crypto[(file_mht_node->node_number - 1) % CHILD_MHT_NODES_COUNT];
 
     // this function decrypt the data _and_ checks the integrity of the data against the gmac
     status = g_cb_aes_gcm_decrypt(&gcm_crypto_data->key, &g_empty_iv, NULL, 0,
-                                  file_mht_node->encrypted.cipher, PF_NODE_SIZE,
+                                  file_mht_node_addr, PF_NODE_SIZE,
                                   &file_mht_node->decrypted.mht, &gcm_crypto_data->gmac);
     if (PF_FAILURE(status)) {
         free(file_mht_node);
@@ -797,6 +813,7 @@ static bool ipf_init_fields(pf_context_t* pf) {
     ipf_init_root_mht(&pf->root_mht);
 
     pf->file           = NULL;
+    pf->addr           = NULL;
     pf->need_writing   = false;
     pf->file_status    = PF_STATUS_UNINITIALIZED;
     pf->last_error     = PF_STATUS_SUCCESS;
@@ -809,11 +826,18 @@ static bool ipf_init_fields(pf_context_t* pf) {
 static bool ipf_init_existing_file(pf_context_t* pf, const char* path) {
     pf_status_t status;
 
-    // read meta-data node
-    if (!ipf_read_node(pf, pf->file, /*node_number=*/0, (uint8_t*)&pf->file_metadata,
-                       PF_NODE_SIZE)) {
+    if (pf->addr == NULL) {
+        pf->last_error = PF_STATUS_INVALID_ADDRESS;
         return false;
     }
+
+    if (pf->real_file_size < PF_NODE_SIZE) {
+        pf->last_error = PF_STATUS_INVALID_FILE_SIZE;
+        return false;
+    }
+
+    // read meta-data node
+    memcpy(&pf->file_metadata, pf->addr, PF_NODE_SIZE);
 
     if (pf->file_metadata.plain_part.file_id != PF_FILE_ID) {
         // such a file exists, but it is not a protected file
@@ -854,15 +878,16 @@ static bool ipf_init_existing_file(pf_context_t* pf, const char* path) {
     }
 
     if (pf->encrypted_part_plain.size > MD_USER_DATA_SIZE) {
-        // read the root node of the mht
-        if (!ipf_read_node(pf, pf->file, /*node_number=*/1, &pf->root_mht.encrypted.cipher,
-                           PF_NODE_SIZE))
+        if (pf->real_file_size < PF_NODE_SIZE * 2) {
+            pf->last_error = PF_STATUS_INVALID_FILE_SIZE;
             return false;
+        }
 
+        // read the root node of the mht
         // this also verifies the root mht gmac against the gmac in the meta-data encrypted part
         status = g_cb_aes_gcm_decrypt(&pf->encrypted_part_plain.mht_key, &g_empty_iv,
                                       NULL, 0, // aad
-                                      &pf->root_mht.encrypted.cipher, PF_NODE_SIZE,
+                                      (unsigned char*)pf->addr + PF_NODE_SIZE, PF_NODE_SIZE,
                                       &pf->root_mht.decrypted.mht,
                                       &pf->encrypted_part_plain.mht_gmac);
         if (PF_FAILURE(status)) {
@@ -891,13 +916,6 @@ static void ipf_try_clear_error(pf_context_t* pf) {
             pf->file_status = PF_STATUS_SUCCESS;
     }
 
-    if (pf->file_status == PF_STATUS_WRITE_TO_DISK_FAILED) {
-        if (ipf_write_all_changes_to_disk(pf)) {
-            pf->need_writing = false;
-            pf->file_status = PF_STATUS_SUCCESS;
-        }
-    }
-
     if (pf->file_status == PF_STATUS_SUCCESS) {
         pf->last_error = PF_STATUS_SUCCESS;
     }
@@ -905,7 +923,8 @@ static void ipf_try_clear_error(pf_context_t* pf) {
 
 
 static pf_context_t* ipf_open(const char* path, pf_file_mode_t mode, bool create, pf_handle_t file,
-                              uint64_t real_size, const pf_key_t* kdk_key, pf_status_t* status) {
+                              void* addr, uint64_t real_size, const pf_key_t* kdk_key,
+                              pf_status_t* status) {
     *status = PF_STATUS_NO_MEMORY;
     pf_context_t* pf = calloc(1, sizeof(*pf));
 
@@ -949,6 +968,7 @@ static pf_context_t* ipf_open(const char* path, pf_file_mode_t mode, bool create
     }
 
     pf->file = file;
+    pf->addr = addr;
     pf->real_file_size = real_size;
     pf->mode = mode;
 
@@ -1170,12 +1190,9 @@ static bool ipf_close(pf_context_t* pf) {
 
 // public API
 
-void pf_set_callbacks(pf_read_f read_f, pf_write_f write_f, pf_truncate_f truncate_f,
-                      pf_aes_cmac_f aes_cmac_f, pf_aes_gcm_encrypt_f aes_gcm_encrypt_f,
-                      pf_aes_gcm_decrypt_f aes_gcm_decrypt_f, pf_random_f random_f,
-                      pf_debug_f debug_f) {
-    g_cb_read            = read_f;
-    g_cb_write           = write_f;
+void pf_set_callbacks(pf_truncate_f truncate_f, pf_aes_cmac_f aes_cmac_f,
+                      pf_aes_gcm_encrypt_f aes_gcm_encrypt_f, pf_aes_gcm_decrypt_f aes_gcm_decrypt_f,
+                      pf_random_f random_f, pf_debug_f debug_f) {
     g_cb_truncate        = truncate_f;
     g_cb_aes_cmac        = aes_cmac_f;
     g_cb_aes_gcm_encrypt = aes_gcm_encrypt_f;
@@ -1185,13 +1202,13 @@ void pf_set_callbacks(pf_read_f read_f, pf_write_f write_f, pf_truncate_f trunca
     g_initialized = true;
 }
 
-pf_status_t pf_open(pf_handle_t handle, const char* path, uint64_t underlying_size,
+pf_status_t pf_open(pf_handle_t handle, const char* path, uint64_t underlying_size, void* addr,
                     pf_file_mode_t mode, bool create, const pf_key_t* key, pf_context_t** context) {
     if (!g_initialized)
         return PF_STATUS_UNINITIALIZED;
 
     pf_status_t status;
-    *context = ipf_open(path, mode, create, handle, underlying_size, key, &status);
+    *context = ipf_open(path, mode, create, handle, addr, underlying_size, key, &status);
     return status;
 }
 
@@ -1243,7 +1260,7 @@ pf_status_t pf_set_size(pf_context_t* pf, uint64_t size) {
         void* data;
         char path[PATH_MAX_SIZE];
         size_t path_len;
-        pf_status_t status = g_cb_truncate(pf->file, 0);
+        pf_status_t status = g_cb_truncate(pf->file, 0, NULL);
         if (PF_FAILURE(status))
             return status;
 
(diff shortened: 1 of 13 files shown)


llly (commented):
> Reviewed 1 of 13 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1681)* status: 1 of 13 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @yao-ji)

___
*[`common/src/protected_files/protected_files.c` line 247 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-NnNihlZ2nQJKhz8YA_O:-NnNihlZ2nQJKhz8YA_P:b-fq5ew9) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/common/src/protected_files/protected_files.c#L247)):*
> ```C
>     // 2. set the IV+GMAC in the parent MHT
>     // [3. set the need_writing flag for all the parents]
>     data = lruc_get_first(pf->cache);
> ```

Can we save `data` of the first `lruc_get_first(pf->cache);` at the beginning of this function to avoid calling it again? I see a lot of them in this function.
___
*[`common/src/protected_files/protected_files.c` line 272 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-NncH3wn0a2mo8adqOzv:-NncH3wn0a2mo8adqOzw:b-86lji9) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/common/src/protected_files/protected_files.c#L272)):*
> ```C
>                 }
> 
>                 data_node->need_writing = false;
> ```

I suppose all ` data_node->need_writing` are set to `false`.  Following`dirty_count` will always be 0;

yao-ji (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1681)* status: 1 of 13 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @llly)

___
*[`common/src/protected_files/protected_files.c` line 272 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-NncH3wn0a2mo8adqOzv:-Nng2WdY81jR-JbimV71:bns0uiy) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/common/src/protected_files/protected_files.c#L272)):*

I suppose all ` data_node->need_writing` are set to `false`.  Following`dirty_count` will always be 0;
</blockquote></details>

Here we only write nodes with`FILE_DATA_NODE_TYPE`, while the following `dirty_count` records nodes with `FILE_MHT_NODE_TYPE`.

llly (commented):
> Reviewed 12 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1681)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @yao-ji)

___
*[`common/src/protected_files/protected_files.c` line 272 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-NncH3wn0a2mo8adqOzv:-NnrSW_DE8Liod5sbKiw:b-veg85v) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/common/src/protected_files/protected_files.c#L272)):*
<details><summary><i>Previously, yao-ji wrote…</i></summary><blockquote>

Here we only write nodes with`FILE_DATA_NODE_TYPE`, while the following `dirty_count` records nodes with `FILE_MHT_NODE_TYPE`.
</blockquote></details>

Get it. Then one `while` loop is enough to handle both types.
___
*[`libos/src/fs/libos_fs_encrypted.c` line 738 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-NnwSDXA-IcZN3LGigQI:-NnwSDXA-IcZN3LGigQJ:bqno4wp) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/libos/src/fs/libos_fs_encrypted.c#L738)):*
> ```C
>         assert(enc->pal_handle);
> 
>         /* Recreate file map: set handle->file.addr to NULL and call PalStreamSetLength */
> ```

Better to zero `handle->file.addr` in `handle_deserialize` function, then these code and `pal_attr.addr` are not needed.
___
*[`pal/include/pal/pal.h` line 329 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-NnwNpU894I5McuN8nJi:-NnwNpU894I5McuN8nJj:beitey9) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/pal/include/pal/pal.h#L329)):*
> ```objc
> #define PAL_OPTION_NONBLOCK        0x2
> #define PAL_OPTION_PASSTHROUGH     0x4 /*!< Disregard `sgx.{allowed,trusted}_files` */
> #define PAL_OPTION_ENCRYPTED_FILE  0x8 /*!< Open encrypted file */
> ```

Suggest name `PAL_OPTION_MAP_FILE`, which means will map the file to untrusted memory when opening.
___
*[`pal/src/host/linux-sgx/pal_files.c` line 417 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-Nnvaafn3nO4XYS4une3:-Nnvaafn3nO4XYS4une4:b62nn0k) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/pal/src/host/linux-sgx/pal_files.c#L417)):*
> ```C
>                 return unix_to_pal_error(ret);
>         }
>         
> ```

Remove spaces
___
*[`pal/src/host/linux-sgx/pal_host.h` line 53 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-NnwQ23U9hiw51KY_iog:-NnwQ23U9hiw51KY_ioh:b-7794z8) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/pal/src/host/linux-sgx/pal_host.h#L53)):*
> ```objc
>             char* realpath;
>             size_t total;
>             bool encrypted;                 /* flag to indicate encrypted files */
> ```

 Suggest name `mapped`, which means whether need to map file to untrusted memory.
___
*[`pal/src/host/linux-sgx/pal_host.h` line 54 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-NnwRTed8-4BdtFzBpb_:-NnwRTed8-4BdtFzBpba:bqbopha) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/pal/src/host/linux-sgx/pal_host.h#L54)):*
> ```objc
>             size_t total;
>             bool encrypted;                 /* flag to indicate encrypted files */
>             void* addr;                     /* mapped address, used only for encrypted files */
> ```

Can we reuse ` void* umem;  ` below?  It is also address mmapped in untrusted memory.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1681)* status: all files reviewed, 13 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @llly and @yao-ji)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1681#-Nnxqlpn6Vs61ecKz7Pw:-Nnxqlpn6Vs61ecKz7Px:b-cp27vx) (no related file):*
I **strongly insist** on keeping the old implementation too. In particular, `pf_read_f()` and `pf_write_f()` callbacks must be kept.

This has two important benefits:
1. I am currently **very unsure** about the security implications of this perf optimization. See my other comment.
   - Keeping the old implementation will allow to introduce a manifest switch `sys.encrypted_files.experimental__zero_copy = true|false`.
   - So that users can always choose the old implementation, which is known to be secure and battle tested.
   - In some future, when we are sure about the implementation and its security, we can remove this switch and use mmapped-based zero-copy optimization always.
2. This will allow **no changes** to `tools/sgx/common/pf_util.c`. This tooling is not performance critical, so I would prefer to not touch it, at least for now.
   - We won't need any modifications to `pf_util.c`. This will ease the review of this PR.
   - In some future, we may want to replace that implementation too. But currently there is no need. 

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1681#-Nnxrual01pgk_pUc53M:-Nnxrual01pgk_pUc53N:b-4atbys) (no related file):*
I propose to keep `pf_read_f()` and `pf_write_f()` callbacks.

With this in mind, why can't we encapsulate the mmap-based zero-copy optimization **inside** these two callbacks? This will tremendously simplify this PR. (We will still need to propagate the `addr` at which the file was mapped, but that's much less intrusive changes than the current state.)

___
___
*[`-- commits` line 8 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-NnxkZwC9jRezlW_B_gj):*
`memory`
___
*[`common/src/protected_files/protected_files.c` line 266 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-NnxpPPe85NUBM7a6hMS:-NnxpPPe85NUBM7a6hMT:bkohzim) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/common/src/protected_files/protected_files.c#L266)):*
> ```C
>                 status = g_cb_aes_gcm_encrypt(&gcm_crypto_data->key, &g_empty_iv, NULL, 0,  // aad
>                                               data_node->decrypted.data.data, PF_NODE_SIZE,
>                                               data_node_addr, &gcm_crypto_data->gmac);
> ```

This is scary!

You encrypt the input plaintext data (`data_node->decrypted.data.data`) directly into untrusted out-of-enclave memory (`data_node_addr`).

What is the guarantee that the encryption crypto doesn't accidentally keep some temporary data in this untrusted memory? The crypto logic is buried somewhere inside `g_cb_aes_gcm_encrypt()`, which in the end maps to `mbedtls_gcm_crypt_and_tag()`:
- https://github.com/gramineproject/gramine/blob/e282ee5a5faf6042973c68ecc6fa2185fcf71ea0/libos/src/fs/libos_fs_encrypted.c#L105-L107
- https://github.com/gramineproject/gramine/blob/e282ee5a5faf6042973c68ecc6fa2185fcf71ea0/common/src/crypto/adapters/mbedtls_adapter.c#L147-L149
- https://github.com/gramineproject/gramine/blob/e282ee5a5faf6042973c68ecc6fa2185fcf71ea0/common/src/crypto/adapters/mbedtls_adapter.c#L163-L164

Did you analyze the internal code of `mbedtls_gcm_crypt_and_tag()`? Or does the mbedTLS developer team provide such a guarantee?

---

Ditto for all other such places (both encryption and decryption of data).
___
*[`libos/src/fs/libos_fs_encrypted.c` line 738 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-NnwSDXA-IcZN3LGigQI:-NnxtjWU-K4wSLxtRk6V:b32v57) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/libos/src/fs/libos_fs_encrypted.c#L738)):*

Better to zero `handle->file.addr` in `handle_deserialize` function, then these code and `pal_attr.addr` are not needed.
</blockquote></details>

+1
___
*[`pal/include/pal/pal.h` line 329 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-NnwNpU894I5McuN8nJi:-NnxttgvB4rW8I2qHsOb:b7c9jz1) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/pal/include/pal/pal.h#L329)):*

Suggest name `PAL_OPTION_MAP_FILE`, which means will map the file to untrusted memory when opening.
</blockquote></details>

Btw, this is what is done for `sgx.trusted_files` already. I feel like we need to consolidate the similar logic of trusted and encrypted (protected) files before going deep into this PR. In other words, I feel like we need a series of preliminary PRs that e.g. introduce this `PAL_OPTION_MAP_FILE`.

We need to discuss this more.
___
*[`pal/src/host/linux/pal_files.c` line 204 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-NnxuNaU93t-FbedWYiU:-NnxuNaU93t-FbedWYiV:b-18ebqi) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/pal/src/host/linux/pal_files.c#L204)):*
> ```C
>         handle->file.addr = addr;
>         handle->file.total = length;
>     }
> ```

This is definitely a wrong place to do the remapping of encrypted files. I feel like this logic should be done inside LibOS, and using some additional existing PAL API, not `PalStreamSetLength()`.

Otherwise we're abusing the PAL API. And just generally -- this is a bad programming practice and a potential WTF place.

We'll need to discuss how to implement that correctly.
___
*[`pal/src/host/linux-sgx/pal_files.c` line 112 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-Nnxv041Ct5LbdptxsXz:-Nnxv041Ct5LbdptxsY-:b3fhwj1) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/pal/src/host/linux-sgx/pal_files.c#L112)):*
> ```C
>                 }
>             }
>         }
> ```

Our file logic in PAL got tremendously complicated. Sorry, we need to untangle our existing code before I will accept this PR. We just pile hack on top of another hack.

I think we need to:
1. First move trusted files logic inside LibOS (or at least in `common/` code).
2. Make the logic of "mmap into untrusted memory" generic for both trusted and encrypted files.
3. Make the distinction between three types of files (allowed, trusted and encrypted) very clear in PAL.
___
*[`pal/src/host/linux-sgx/pal_files.c` line 426 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-NnxvbLcDrUXZazfAZqU:-NnxvbLcDrUXZazfAZqV:bocuwch) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/pal/src/host/linux-sgx/pal_files.c#L426)):*
> ```C
>                 return unix_to_pal_error(ret);
>         }
>     }
> ```

ditto (we're abusing the PAL API)

kailun-qin (changes requested):
> Reviewed 13 of 13 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1681)* status: all files reviewed, 17 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @llly, and @yao-ji)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1681#-Nnxrual01pgk_pUc53M:-No9mi588OrgOsLsCrE_:bzexxqg) (no related file):*
> why can't we encapsulate the mmap-based zero-copy optimization **inside** these two callbacks?

+1, I have the same question - @yao-ji: what is the blocker here?

___
___
*[`-- commits` line 8 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-No9jKHxB0acdAqR2NBd):*
-> `retrieved`



_Code quote:_
```
retrived
```
___
*[`common/src/protected_files/protected_files.c` line 210 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-No9qda15rpJSQb5Ca7Y:-No9qda15rpJSQb5Ca7Z:b-7s7cw2) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/common/src/protected_files/protected_files.c#L210)):*
> ```C
> 
>     if (pf->addr == NULL) {
>         pf->last_error = PF_STATUS_INVALID_ADDRESS;
> ```

-> `PF_STATUS_INVALID_FILE_SIZE`



_Code quote:_
```C
PF_STATUS_INVALID_ADDRESS
```
___
*[`common/src/protected_files/protected_files.c` line 262 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-No9mA7e-nm-F1sEdrgB:-No9mA7e-nm-F1sEdrgC:b-65ctlz) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/common/src/protected_files/protected_files.c#L262)):*
> ```C
>                 // encrypt the data, this also saves the gmac of the operation in the mht crypto
>                 // node
>                 void* data_node_addr = (unsigned char*)pf->addr + 
> ```

pls remove trailing white space
___
*[`common/src/protected_files/protected_files.c` line 266 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-NnxpPPe85NUBM7a6hMS:-No9k5p92mfc1z2crcmr:blww299) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/common/src/protected_files/protected_files.c#L266)):*
> You encrypt the input plaintext data (`data_node->decrypted.data.data`) directly into untrusted out-of-enclave memory (`data_node_addr`).
> What is the guarantee that the encryption crypto doesn't accidentally keep some temporary data in this untrusted memory?

Yes, I think there is no guarantee there. It at least needs some thorough analysis, but better not to do this in-place.
___
*[`common/src/protected_files/protected_files.c` line 336 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-No9lfPC85NvPdF7L0qM:-No9lfPC85NvPdF7L0qN:bygrnlj) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/common/src/protected_files/protected_files.c#L336)):*
> ```C
> 
>         void* mht_node_addr = (unsigned char*)pf->addr + 
>                                 PF_NODE_SIZE * file_mht_node->physical_node_number;
> ```

pls align
___
*[`libos/src/fs/libos_fs_encrypted.c` line 738 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-NnwSDXA-IcZN3LGigQI:-No9o3YwDYauyR09G9I7:b32v57) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/libos/src/fs/libos_fs_encrypted.c#L738)):*

+1
</blockquote></details>

+1
___
*[`pal/include/pal/pal.h` line 329 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-NnwNpU894I5McuN8nJi:-No9oWO-BQCQxQMHRald:bs87ez9) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/pal/include/pal/pal.h#L329)):*
> this is what is done for `sgx.trusted_files` already. I feel like we need to consolidate the similar logic of trusted and encrypted (protected) files before going deep into this PR

+1
___
*[`pal/src/host/linux/pal_files.c` line 204 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-NnxuNaU93t-FbedWYiU:-No9l-4y89YN4A43zL2A:bsrrhq7) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/pal/src/host/linux/pal_files.c#L204)):*
> I feel like this logic should be done inside LibOS, and using some additional existing PAL API, not `PalStreamSetLength()`.

+1
___
*[`pal/src/host/linux-sgx/pal_host.h` line 54 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-NnwRTed8-4BdtFzBpb_:-No9pRt-BBxhJ6MwC3_i:b9b9m87) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/pal/src/host/linux-sgx/pal_host.h#L54)):*

Can we reuse ` void* umem;  ` below?  It is also address mmapped in untrusted memory.
</blockquote></details>

+1, I think we should discuss whether it's possible to consolidate the similar logic of trusted (where `umem` is used) and encrypted files.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1681)* status: all files reviewed, 17 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @kailun-qin, @llly, and @yao-ji)

___
*[`common/src/protected_files/protected_files.c` line 266 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-NnxpPPe85NUBM7a6hMS:-NoBDVgN2TvJ3x-kw4Fi:bkq6teb) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/common/src/protected_files/protected_files.c#L266)):*

> You encrypt the input plaintext data (`data_node->decrypted.data.data`) directly into untrusted out-of-enclave memory (`data_node_addr`).
> What is the guarantee that the encryption crypto doesn't accidentally keep some temporary data in this untrusted memory?

Yes, I think there is no guarantee there. It at least needs some thorough analysis, but better not to do this in-place.
</blockquote></details>

Yes, better not to do this in-place. We can simply have a scratch (temporary)  memory buffer of size `PF_NODE_SIZE`, and use it for all intermediate encryption/decryption steps.

This will lead to an additional memory copy, but at least the implementation will be security-safe. **You'll also need to re-do the evaluation**.

yao-ji (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1681)* status: 5 of 15 files reviewed, 17 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @kailun-qin, and @llly)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1681#-Nnxrual01pgk_pUc53M:-NpbgjQI0kiBCPwgt1Fh:bswwy56) (no related file):*

> why can't we encapsulate the mmap-based zero-copy optimization **inside** these two callbacks?

+1, I have the same question - @yao-ji: what is the blocker here?
</blockquote></details>

`mmap` can be moved inside read/write callbacks, but `munmap` cannot. It is clearer in logic to map/unmap file during file open/close.

After the fix, there is no security concern. I have updated the performance data. Adding a temp buffer makes around 3~6% downgrade, which I think is not significant. So we don't need to keep the old implementation.

___
___
*[`common/src/protected_files/protected_files.c` line 210 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-No9qda15rpJSQb5Ca7Y:-NpbfIcR2mWiqtFGmHXd:b-gm76el) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/common/src/protected_files/protected_files.c#L210)):*

-> `PF_STATUS_INVALID_FILE_SIZE`
</blockquote></details>

This is an address error. Why change to file size error?
___
*[`common/src/protected_files/protected_files.c` line 247 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-NnNihlZ2nQJKhz8YA_O:-Npbfowt4YuCMOW00fe4:bttcyo0) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/common/src/protected_files/protected_files.c#L247)):*

Can we save `data` of the first `lruc_get_first(pf->cache);` at the beginning of this function to avoid calling it again? I see a lot of them in this function.
</blockquote></details>

This function changes internal pointer which is also used by `lruc_get_next()`.
___
*[`common/src/protected_files/protected_files.c` line 262 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-No9mA7e-nm-F1sEdrgB:-NpbgFLGCeKdn3N0lZ2g:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/common/src/protected_files/protected_files.c#L262)):*

pls remove trailing white space
</blockquote></details>

Done.
___
*[`common/src/protected_files/protected_files.c` line 266 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-NnxpPPe85NUBM7a6hMS:-NpcI9EtBl27kmqW2AQi:bqrlgro) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/common/src/protected_files/protected_files.c#L266)):*

Yes, better not to do this in-place. We can simply have a scratch (temporary)  memory buffer of size `PF_NODE_SIZE`, and use it for all intermediate encryption/decryption steps.

This will lead to an additional memory copy, but at least the implementation will be security-safe. **You'll also need to re-do the evaluation**.
</blockquote></details>

Fixed by adding a temp buffer when doing encryption. Performance data is also updated, which indicates a 3~6% downgrade. My question is whether we have to do the same during decryption. The input untrusted memory is a const pointer.
___
*[`common/src/protected_files/protected_files.c` line 272 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-NncH3wn0a2mo8adqOzv:-NpbgMh_AZzBWZHz-bjD:b7rg8rf) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/common/src/protected_files/protected_files.c#L272)):*

Get it. Then one `while` loop is enough to handle both types.
</blockquote></details>

`data_node` should be updated before `mht_node`, since it saves key and gmac data into parent `mht_node`.
___
*[`common/src/protected_files/protected_files.c` line 336 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-No9lfPC85NvPdF7L0qM:-Npbge4kCJCl2VBA685y:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/common/src/protected_files/protected_files.c#L336)):*

pls align
</blockquote></details>

Done.
___
*[`libos/src/fs/libos_fs_encrypted.c` line 738 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-NnwSDXA-IcZN3LGigQI:-NpcLkWlB5VTs8tZtd_U:b-yvjnug) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/libos/src/fs/libos_fs_encrypted.c#L738)):*

+1
</blockquote></details>

Moved zero pointer logic into `handle_deserialize` function.

For now, we still need the remap logic. If we can come up with a PAL API that can do `mmap/munmap` untrusted memory, we can merge the remaining logic into `encrypted_file_internal_open()`, which makes more sense.
___
*[`pal/src/host/linux/pal_files.c` line 204 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-NnxuNaU93t-FbedWYiU:-NpcNU9Q3Rl5iwFiGCed:bfekxeh) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/pal/src/host/linux/pal_files.c#L204)):*

> I feel like this logic should be done inside LibOS, and using some additional existing PAL API, not `PalStreamSetLength()`.

+1
</blockquote></details>

Since map untrusted memory acts differently on Linux host and SGX host, I don't think it can be done inside LibOS using existing PAL API.

I think the better way is to create new PAL API to map/unmap untrusted memory. Then we can keep all the existing PAL file operations unchanged.
___
*[`pal/src/host/linux-sgx/pal_files.c` line 417 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-Nnvaafn3nO4XYS4une3:-NpcPaEo7yT4iJWDP48L:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/pal/src/host/linux-sgx/pal_files.c#L417)):*

Remove spaces
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/pal_host.h` line 53 at r1](https://reviewable.io/reviews/gramineproject/gramine/1681#-NnwQ23U9hiw51KY_iog:-NpcPeEn8lHLDhidWMrm:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/22b36daabb7d0f729559f566f291bf601d411e6b/pal/src/host/linux-sgx/pal_host.h#L53)):*

Suggest name `mapped`, which means whether need to map file to untrusted memory.
</blockquote></details>

Done.

llly (commented):
> Reviewed 4 of 10 files at r2.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1681)* status: 9 of 15 files reviewed, 17 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @kailun-qin, and @yao-ji)

___
*[`pal/include/pal/pal.h` line 329 at r2](https://reviewable.io/reviews/gramineproject/gramine/1681#-Npcan0gDs3_Pho3lpb2:-Npcan0gDs3_Pho3lpb3:b-hq7v03) ([raw file](https://github.com/gramineproject/gramine/blob/755660641339fc91a5fb47967593310e9998da6b/pal/include/pal/pal.h#L329)):*
> ```objc
> #define PAL_OPTION_NONBLOCK        0x2
> #define PAL_OPTION_PASSTHROUGH     0x4 /*!< Disregard `sgx.{allowed,trusted}_files` */
> #define PAL_OPTION_MAP_FILE        0x8 /*!< Map encrypted file to untrusted memory */
> ```

Better to remove `encrypted`. Pal can't handle and don't need to know file is encrypted or not.



_Code quote:_
```objc
Map encrypted file
```
___
*[`pal/src/host/linux/pal_files.c` line 78 at r2](https://reviewable.io/reviews/gramineproject/gramine/1681#-Npcj7ZQ-hbYgULU1cjD:-Npcj7ZQ-hbYgULU1cjE:bb7phvt) ([raw file](https://github.com/gramineproject/gramine/blob/755660641339fc91a5fb47967593310e9998da6b/pal/src/host/linux/pal_files.c#L78)):*
> ```C
>     hdl->file.seekable = !S_ISFIFO(st.st_mode);
> 
>     /* map file into memroy when open encrypted files */
> ```

typo



_Code quote:_
```C
memroy 
```
___
*[`pal/src/host/linux-sgx/pal_files.c` line 102 at r2](https://reviewable.io/reviews/gramineproject/gramine/1681#-NpcjJGg3e5QQym91v_o:-NpcjJGg3e5QQym91v_p:bb7phvt) ([raw file](https://github.com/gramineproject/gramine/blob/755660641339fc91a5fb47967593310e9998da6b/pal/src/host/linux-sgx/pal_files.c#L102)):*
> ```C
>         }
> 
>         /* map file into untrusted memroy when open encrypted files */
> ```

typo



_Code quote:_
```C
memroy 
```

mwkmwkmwk (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1681)* status: 9 of 15 files reviewed, 17 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @kailun-qin, @llly, and @yao-ji)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1681#-Nnxrual01pgk_pUc53M:-Nqde_1a4TdLawNaKDWR:bgs5844) (no related file):*
<details><summary><i>Previously, yao-ji wrote…</i></summary><blockquote>

`mmap` can be moved inside read/write callbacks, but `munmap` cannot. It is clearer in logic to map/unmap file during file open/close.

After the fix, there is no security concern. I have updated the performance data. Adding a temp buffer makes around 3~6% downgrade, which I think is not significant. So we don't need to keep the old implementation.
</blockquote></details>

I think `mmap` and `munmap` can be called just fine in libos `open`/`close` wrappers. The address can be kept within the structure pointed to by `pf_handle_t`. There is thus no need to modify the interface, or anything within `common/src/protected_files/` at all.

Keeping the current interface in place also ensures some security by construction — since only the explicitly requested read/writes can touch the untrusted memory. The callbacks should also take care of using the proper helpers (`sgx_copy_from_enclave`/`sgx_copy_to_enclave`) to access untrusted memory.

___

mkow (changes requested):
> Reviewed 1 of 10 files at r2.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1681)* status: 9 of 15 files reviewed, 20 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @kailun-qin, @llly, and @yao-ji)

___
*[`-- commits` line 2 at r2](https://reviewable.io/reviews/gramineproject/gramine/1681#-NqTRgrF-iH7pK_DC_t-):*




_Suggestion:_
```
Improve
```
___
*[`-- commits` line 2 at r2](https://reviewable.io/reviews/gramineproject/gramine/1681#-NqTRkOWB0rE5TARkvI9):*
I'd just skip the component tag, it touches practically the whole codebase.
___
*[`-- commits` line 4 at r2](https://reviewable.io/reviews/gramineproject/gramine/1681#-NqTSKAaD_abSflT0nuA):*




_Suggestion:_
```
This commit improves performance with the following changes:
```

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1681)* status: 9 of 15 files reviewed, 19 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin, @llly, @mwkmwkmwk, and @yao-ji)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1681#-Nnxrual01pgk_pUc53M:-Nr-XQPlCGl6DjsoG7Zt:b-qz8nwu) (no related file):*
> After the fix, there is no security concern. I have updated the performance data. Adding a temp buffer makes around 3~6% downgrade, which I think is not significant.

Thanks, at a first glance, looks good to me! Also the 3-6% drop in performance is tolerable.

> So we don't need to keep the old implementation.

I disagree with this. I'd like to keep the old read/write callbacks inside `protected_files/`. At least for the benefit of `pf_util.c`. But yes, we don't need to new manifest option (`sys.encrypted_files.experimental__zero_copy`) that I proposed above.

> The callbacks should also take care of using the proper helpers (`sgx_copy_from_enclave`/`sgx_copy_to_enclave`) to access untrusted memory.

@mwkmwkmwk I'm confused what you mean by "the callbacks". Do you mean that `libos_fs_encrypted.c` code must replace current `cb_read()` and `cb_write()` implementations to perform memcpy (more precisely, SGX-specific memory copy untrusted<->trusted) instead of `PalStreamRead()` and `PalStreamWrite()`?

- If I understand your proposal correctly, then how do you plan to use `sgx_copy_from_enclave()`/`sgx_copy_to_enclave()` inside the LibOS? That's prohibited (such things are encapsulated inside PAL).
- Or maybe you mean that LibOS still calls `PalStreamRead()` and `PalStreamWrite()` in its `cb_read()` and `cb_write()` callbacks, but these PAL implementations (on SGX PAL) see that there is untrusted memory as the source is already mmapped, so they will use that.

I like the second approach a lot. It also bodes well with Trusted Files.

> If we can come up with a PAL API that can do `mmap/munmap` untrusted memory, we can merge the remaining logic into `encrypted_file_internal_open()`, which makes more sense.

@yao-ji Why do we need to come up with a **separate API**? You introduced the `PAL_OPTION_MAP_FILE` flag, so why can't we do `PalStreamMap(..., PAL_OPTION_MAP_FILE)`? That seems enough.

---

I'd like to discuss and agree on the design and implementation steps during the next Gramine meeting. This async discussion is too slow and haphazard.

___

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1681)* status: 9 of 15 files reviewed, 19 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @kailun-qin, @llly, and @yao-ji)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1681#-Nnxrual01pgk_pUc53M:-Nr3NyG35bopr8cyJQ7G:bges6x5) (no related file):*
> Do you mean that `libos_fs_encrypted.c` code must replace current `cb_read()` and `cb_write()` implementations to perform memcpy (more precisely, SGX-specific memory copy untrusted<->trusted) instead of `PalStreamRead()` and `PalStreamWrite()`?

The biggest point here is that currently the PR just copies untrusted memory in and out with `memcpy()` instead of proper wrappers, which is insecure because of e.g. CVE-2022-21166. And the proposed solution is to design the code differently and go through the read/write operations, which will call `PalStreamRead()` and `PalStreamWrite()` respectively, which know how to handle the untrusted memory. And I think this is what you described in the second point.

___

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1681)* status: 9 of 15 files reviewed, 19 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin, @llly, @mkow, and @yao-ji)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1681#-Nnxrual01pgk_pUc53M:-Nr4RK5m3nA3k9nWnwWL:b-omr1fl) (no related file):*

> Do you mean that `libos_fs_encrypted.c` code must replace current `cb_read()` and `cb_write()` implementations to perform memcpy (more precisely, SGX-specific memory copy untrusted<->trusted) instead of `PalStreamRead()` and `PalStreamWrite()`?

The biggest point here is that currently the PR just copies untrusted memory in and out with `memcpy()` instead of proper wrappers, which is insecure because of e.g. CVE-2022-21166. And the proposed solution is to design the code differently and go through the read/write operations, which will call `PalStreamRead()` and `PalStreamWrite()` respectively, which know how to handle the untrusted memory. And I think this is what you described in the second point.
</blockquote></details>

@mkow Yes, I understand the problem with `memcpy()`. I agree with calling `PalStreamRead()` and `PalStreamWrite()` and doing proper copying inside of them. This will look very much like what Trusted Files do.

Anyway, we should discuss this PR and alternative design next week: https://github.com/gramineproject/gramine/discussions/1772

___

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1681)* status: 9 of 15 files reviewed, 23 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin, @llly, @mkow, and @yao-ji)

___
*[`common/src/protected_files/protected_files.c` line 273 at r2](https://reviewable.io/reviews/gramineproject/gramine/1681#-NreHz3P0V3Y4vqTG7Rq:-NreHz3P0V3Y4vqTG7Rr:bnhfsc) ([raw file](https://github.com/gramineproject/gramine/blob/755660641339fc91a5fb47967593310e9998da6b/common/src/protected_files/protected_files.c#L273)):*
> ```C
>                 }
> 
>                 memcpy(data_node_addr, temp_enc_node, PF_NODE_SIZE);
> ```

All these memcpy should be `sgx_copy_from_enclave()` or `sgx_copy_to_enclave()`.

Here and everywhere where we copy untrusted <-> trusted.

(But as mentioned in a top-level comment by @mwkmwkmwk and @mkow, it's worth trying a different design with read/write callbacks, and not here.)
___
*[`common/src/protected_files/protected_files.c` line 699 at r2](https://reviewable.io/reviews/gramineproject/gramine/1681#-NreHLDR8oRtT5yhTdeJ:-NreHLDR8oRtT5yhTdeK:b-wino2m) ([raw file](https://github.com/gramineproject/gramine/blob/755660641339fc91a5fb47967593310e9998da6b/common/src/protected_files/protected_files.c#L699)):*
> ```C
>     status = g_cb_aes_gcm_decrypt(&gcm_crypto_data->key, &g_empty_iv, NULL, 0,
>                                   file_data_node_addr, PF_NODE_SIZE,
>                                   file_data_node->decrypted.data.data, &gcm_crypto_data->gmac);
> ```

Why didn't you introduce a bounce buffer (`temp_enc_node`) here?
___
*[`common/src/protected_files/protected_files.c` line 770 at r2](https://reviewable.io/reviews/gramineproject/gramine/1681#-NreHS606l6xej5jtKHP:-NreHS606l6xej5jtKHQ:b-wino2m) ([raw file](https://github.com/gramineproject/gramine/blob/755660641339fc91a5fb47967593310e9998da6b/common/src/protected_files/protected_files.c#L770)):*
> ```C
>     status = g_cb_aes_gcm_decrypt(&gcm_crypto_data->key, &g_empty_iv, NULL, 0,
>                                   file_mht_node_addr, PF_NODE_SIZE,
>                                   &file_mht_node->decrypted.mht, &gcm_crypto_data->gmac);
> ```

Why didn't you introduce a bounce buffer (`temp_enc_node`) here?
___
*[`common/src/protected_files/protected_files.c` line 895 at r2](https://reviewable.io/reviews/gramineproject/gramine/1681#-NreHWD1B8k5bcJ--KR-:-NreHWD1B8k5bcJ--KR0:b-wino2m) ([raw file](https://github.com/gramineproject/gramine/blob/755660641339fc91a5fb47967593310e9998da6b/common/src/protected_files/protected_files.c#L895)):*
> ```C
>                                       (unsigned char*)pf->addr + PF_NODE_SIZE, PF_NODE_SIZE,
>                                       &pf->root_mht.decrypted.mht,
>                                       &pf->encrypted_part_plain.mht_gmac);
> ```

Why didn't you introduce a bounce buffer (`temp_enc_node`) here?

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1681)* status: 9 of 15 files reviewed, 24 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin, @llly, @mkow, and @yao-ji)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1681#-O3BYlp73cohkZxRuQZa:-O3BYlp73cohkZxRuQZb:b-sje83d) (no related file):*
@yao-ji @kailun-qin What is the status of this PR?

I believe this PR should go after the Trusted Files re-implementation (moving Trusted Files from PAL to LibOS). This reimplementation is done in PR https://github.com/gramineproject/gramine/pull/1812, so I would recommend to base the new version of this PR on top of #1812. This will save us from complex rebases afterwards.

Also, do you have the design ironed out now? Please re-check the discussion that we had in #1772.

___

Issue #1176, opened 2023-02-14
RFC: support symlinks in `sgx.trusted_files`


woju:
### Description of the feature

# how it would work

1) `gramine-sgx-sign`, upon encountering a symlink, would emit a symlink specification (see below for options) instead of measuring the contents of the file.

2) We'd have some `realpath(3)` equivalent in trusted PAL that would resolve the file paths to a canonical form (which does not need to be the `realpath` outside gramine, but `gramine-sgx-sign` would probably make it so anyway). This function will need to rely only on the contents of the manifest, not any outside info (esp. untrusted one like the runtime result of `readlink(2)` outside of gramine). 

# options for manifest syntax

## option 1: `type`

```toml
[[sgx.trusted_files]]
uri = "[URI]"
sha256 = "[SHA256]"
type = "symlink"
```

`type = "symlink"` means that `sha256` is over contents of `readlink(2)`, not over the contents of `read(2)`. During runtime, we call `readlink(2)` to untrusted kernel and check if the result matches.

## option 2: `symlink_to`

```toml
[[sgx.trusted_files]]
uri = "[URI]"
symlink_to = "..."
```

The path is symlink. Store the symlink contents directly in the manifest. No `readlink(2)` happens during runtime.

**NOTE:** this option is equivalent to fake bind-mounts, i.e. if you write fake symlink_to, underlying files don't need to exist on the filesystem outside of gramine.

## option 3: allow choice

`uri =` and `type = "symlink"` are mandatory, but whoever is writing manifest can choose if there's `symlink_to = ` or `sha256 = ` (exactly one of those is mandatory, specifying two or none is an error and causes gramine to exit !=0 with message when loading manifest).

### Why Gramine should implement it?

See discussion in gramineproject/gsc#120, but TLDR is, without this feature we're condemned to measure each file under every path it's accessible from. For one, newer distros are usrmerged, which means that there's links like `/bin -> usr/bin`. @dimakuv says on typical distro the size of the manifest explodes 40×.

mkow:
I like that option 1 is more consistent with the rest of our syntax, but I think we should go with option 2, because:
- Less performance overhead (no need to read the host filesystem, less metadata to keep, smaller manifest)
- More readable/debuggable entry in signed manifest - cleartext content instead of an obscure hash. Regular trusted files are hashed only because they are potentially big, so we don't want to embed the contents in the manifest. But symlinks are short, so we can paste them literally.

[...]

Issue #24, opened 2021-09-10
Filesystem refactoring


pwmarcz:
This describes the current state of my filesystem refactoring project.

Legend:

:heavy_check_mark: Done (merged to master)
:construction: In progress (usually has a PR open)
:star: Next (usually will be unlocked by current "in progress")

## Bug fixes and new features
:heavy_check_mark: gramineproject/graphene#952 Listing a directory doesn't show mountpoints
:heavy_check_mark: `lseek` on directories (PR gramineproject/graphene#2406)
:heavy_check_mark: Fix `/proc` code (will fix gramineproject/graphene#948, gramineproject/graphene#1387; PR gramineproject/graphene#2453)
:heavy_check_mark: Fix `lseek` overflow on big offsets (PR gramineproject/graphene#2478)
:heavy_check_mark: File locking (`fcntl`) (will fix gramineproject/graphene#437) (PR gramineproject/graphene#2481)
:heavy_check_mark: Fix `poll` on pseudo-handles (PR gramineproject/graphene#2498) (found in gramineproject/graphene#2419)
:heavy_check_mark: Fix crashes after `mknod` on path reuse (PR gramineproject/graphene#2499) (found in gramineproject/graphene#2419)
:construction: **Investigate the master list of issues (#1803)**: probably will lead to more minor bug fixes.
Symlinks (will fix gramineproject/graphene#516): as in-memory files emulated by Graphene

### Multi-process synchronization
:heavy_check_mark: Initial design and discussion (#2158)
:heavy_check_mark: Prototype in Python to clarify the API (https://github.com/pwmarcz/fs-demo)
:heavy_check_mark: Proof of concept: FD position sync (PR gramineproject/graphene#2264, gramineproject/graphene#2267)
:heavy_check_mark: File locking (`fcntl`) (will fix gramineproject/graphene#437) (PR gramineproject/graphene#281, ~~#2522~~)
:heavy_check_mark: Fix `mknod` crash (PR gramineproject/graphene#2499)
Synchronize dentry cache

### Dentry/low level filesystem cleanup
:heavy_check_mark: Initial design and discussion (issue gramineproject/graphene#2321)
:heavy_check_mark: Rewrite low-level dentry functions (PR gramineproject/graphene#2324)
:heavy_check_mark: Remove `ino` (PR gramineproject/graphene#2374)
:heavy_check_mark: Make `mode` and `type` functional (PR gramineproject/graphene#2379)
[...]

PR #1701, opened 2024-01-05
Add support for symlinks for regular files

BobbyAtFortanix:
AI and bank confidential applications use symlinks for regular files.

- added support for system calls: link, linkat, symlink & symlinkat
- added set_link() operation to libos_d_ops
- implemented set_link() for chroot, chroot/encrypted & tmp filesystems
- implemented follow_link() for chroot, chroot/encrypted & tmp filesystems
- added symlink support to the chroot/encrypted filesystem
- added hardlink and symlink support to tmp filesystem
- added lstat(), readlink() & link() operations to Pal's handle_ops
- implemented lstat(), readlink() & link() for linux & linux-sgx
- implemented corresponding lstat(), readlink() & link() outcalls for enclave and host
- implemented PalGetLinkStats(), PalReadLink() & PalCreateLink()
- added unit tests for the above changes

new changes can be tested by running the following:

```
cd libos/test/regression
gramine-test --sgx pytest -v -k "_link_ or _symlink_"

```
Signed-off-by: Bobby Marinov <Bobby.Marinov@fortanix.com>


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1701)* status: 0 of 30 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @BobbyAtFortanix)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1701#-NndDBjU5PZ_IY-PyXAi:-NndDBjU5PZ_IY-PyXAj:b-wsyoy9) (no related file):*
Quick cosmetic comments:
- Some files don't have your copyright (cases where you modified a lot) or have wrong-person copyright (the new LibOS test)
- There are no changes in our Documentation; surely we have some info in docs that conflicts with your symlinks implementation
- You didn't enable symlink-related LTP tests
- I think we have symlink-related LibOS regression and FS tests, but those parts are commented out (or something like this) -- they need to be re-enabled now

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1701#-NndE3Cm73a32kdvS55G:-NndE3Cm73a32kdvS55H:b-8e8l3k) (no related file):*
This symlink support is **completely pass-through** (except tmpfs files which are by design in-memory-only).

What is the threat model here? Can you provide a security analysis? Why did you choose this pass-through design?

How is this situation solved: https://github.com/gramineproject/gramine/issues/1176 ?

In [this meta-issue](https://github.com/gramineproject/gramine/issues/24), Pawel (ex-Gramine maintainer who wrote the whole FS subsystem) suggested the following:
> Symlinks (will fix [gramineproject/graphene#516](https://github.com/gramineproject/graphene/issues/516)): as in-memory files emulated by Graphene

Pawel's design suggestion contradicts yours: in-memory-only symlinks vs passthrough-symlinks. Why did you choose your design?

___

BobbyAtFortanix:
>     * Some files don't have your copyright (cases where you modified a lot) or have wrong-person copyright (the new LibOS test)

Fixed in the new commit

>     * There are no changes in our Documentation; surely we have some info in docs that conflicts with your symlinks implementation

Some additional information added. I hope it is helpful.

>     * You didn't enable symlink-related LTP tests

Enabled symlink-related tests in `ltp.cfg`. Some of them are still failing but it is not because of the lack of symlink support. For those, I have added a comment specifying what is missing.

>     * I think we have symlink-related LibOS regression and FS tests, but those parts are commented out (or something like this) -- they need to be re-enabled now

I could not figure out how to enable those nor how to run them.

> _[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1701#-NndE3Cm73a32kdvS55G:-NndE3Cm73a32kdvS55H:b-8e8l3k) (no related file):_ This symlink support is **completely pass-through** (except tmpfs files which are by design in-memory-only).

`chroot/encrypted` is also pass-through but the symlink there is protected by the encryption.
 
> What is the threat model here? Can you provide a security analysis? Why did you choose this pass-through design?

Our use-case scenario always uses an application inside a docker container. The files accessed by the application are hashed during container signing and if the hash does not match at run-time, we bug check.
The chroot (unencrypted) filesystem implementation uses the existing files in the application's container, including the symlinks. So for example, if in a container, /usr/bin/sh is a symlink to /usr/bin/dash, then there's an actual symlink on-disk in the container image. That symlink will function exactly the same when the application is run in the container. If we want a symlink to function the same when it's created by that app running the symlink syscall, then the symlink syscall implementation inside needs to make an actual symlink on disk.
For read-only chroot mounts, the filesystem hash includes the hashes of the symlinks, so symlinks are protected against modification or removal. Of course on read-only mounts, you can't make symlinks, so it doesn't matter how the symlink system call is implemented.
For read-write chroot mounts, no security is provided. An attacker can freely read or modify any data, including symlinks. These mounts should only be used for parts of the filesystem that do not contain sensitive data, and cannot compromise the security of the application if modified. So for example application configuration files should not be in writable chroot mounts even if the contents of the configuration files are not themselves secret. i.e. this implementation relies that the user that creates the container will use the `chroot/encrypted` mounts for sensitive information. 

> How is this situation solved: #1176 ?

The manifest solution can be used only for read-only symlinks that are known before the application is signed/ran. The proposed solution affects the cases where the application itself is creating symlinks during runtime.
 
> In [this meta-issue](https://github.com/gramineproject/gramine/issues/24), Pawel (ex-Gramine maintainer who wrote the whole FS subsystem) suggested the following:> 
> Symlinks (will fix [gramineproject/graphene#516](https://github.com/gramineproject/graphene/issues/516)): as in-memory files emulated by Graphene
> Pawel's design suggestion contradicts yours: in-memory-only symlinks vs passthrough-symlinks. Why did you choose your design?

Because some apps want to create symlinks at run time.

dimakuv (changes requested):
> I could not figure out how to enable those nor how to run them.

Check item 6 here: https://gramine.readthedocs.io/en/stable/devel/onboarding.html#fixing-a-bug

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1701#-:-NoSCsc-36K1jV0CwGg1:bdf390s)* status: 0 of 64 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @BobbyAtFortanix)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1701#-NndE3Cm73a32kdvS55G:-NoSD-Tm8Wk79D86YKWN:b8q0bkd) (no related file):*
Thanks for this initial analysis!

I guess my main question lies here:

> Because some apps want to create symlinks at run time.

What exactly do these apps want from symlinks? Can't these symlinks be created ephemerally, completely inside the enclave? If not, then why is it important to have these symlinks persistent across SGX enclaves (or separate runs of the same SGX enclave)?

> For read-only chroot mounts, the filesystem hash includes the hashes of the symlinks, so symlinks are protected against modification or removal.

Hm, are you sure that's how our manifest-generating Python scripts work? IIRC, our Python scripts simply resolve symlinks, so the hashes are NOT the symlink contents (string with a destination path) but the symlinked-to file contents.

___

BobbyAtFortanix:
I think I addressed all of the concerns from the previous review(s).
I've kept the sym-links support only for `chroot/encrypted` and `tmpfs` file systems. This way there could be no possible leak or interference with the enclave protected data.

dimakuv (commented):
Hi @BobbyAtFortanix, since the PR is very big, the Gramine team will need to put a priority on it, discuss the design and then do a deep-dive into the implementation. Would you be able to join our Gramine weekly meetings at some point, so we can discuss how and when we can proceed with this PR?

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1701#-:-Nr4YzBs2yi01QfaUboS:beg3tez)* status: 0 of 64 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @BobbyAtFortanix)

BobbyAtFortanix (commented):
I'll join the Gramine's 2024/02/27 (tomorrow) 7am PST meeting.

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1701#-:-NrcxIVl3WtdyWWeFc91:boavu61)* status: 0 of 64 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: )

BobbyAtFortanix (commented):
I hope this will help during the code review.
## Pal
* added a flag to `PalStreamOpen()` specifying if the handle created is only going to be used for deleting the file. In that case there is no need to go and open the file (`ocall`) before deleting it via handle-less path based `ocall`. It also allows us to delete link files that fail open with `ELOOP` error.
## LibOS
### libos/test
* added the basic symbolic & hard link tests for `chroot`(disabled), `chroot/encrypted` and `tmpfs`
### libos/src/sys/libos_file.c
* this is where the symlink/hardlink are created for all file systems
### libos/src/fs
* changes related to symlink & hardlink for `tmpfs` and  `chroot/encrypted`
##  Other notes
* Running non-file related LTP tests with `tmpfs` instead of `chroot` is failing with error. e.g. `setrlimit05` test. this can be verified by modifying the LTP manifest file and running the test.

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1701#-:-NrhF5CP5MMA1ar5Ih14:b2mbqmt)* status: 0 of 64 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: )

BobbyAtFortanix (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1701)* status: 0 of 64 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1701#-NndE3Cm73a32kdvS55G:-NsAOiwU4ViLlsikMixA:big352p) (no related file):*

Thanks for this initial analysis!

I guess my main question lies here:

> Because some apps want to create symlinks at run time.

What exactly do these apps want from symlinks? Can't these symlinks be created ephemerally, completely inside the enclave? If not, then why is it important to have these symlinks persistent across SGX enclaves (or separate runs of the same SGX enclave)?

> For read-only chroot mounts, the filesystem hash includes the hashes of the symlinks, so symlinks are protected against modification or removal.

Hm, are you sure that's how our manifest-generating Python scripts work? IIRC, our Python scripts simply resolve symlinks, so the hashes are NOT the symlink contents (string with a destination path) but the symlinked-to file contents.
</blockquote></details>

we don't know how the applications use symlinks. We have a batch file that runs the client application and collects all syscalls and pseudo-file opens and then creates a summary report. For security reasons the client send us that summary report.
In our Graphene(before it was renamed to Gramine) fork we have modified chroot to have read-only mounts. For those we generate a complete hash for everything under that mount so we can catch any attempt of modifying it without our knowledge. When a directory is enumerated we make sure that no files/dirs are added nor removed.
We also maintain a symlink directory entry instead of opening the symlink target. 

___

PR #1720, opened 2024-01-16
[curl] Re-enable proxy support in libcurl

### LINKED ISSUE #1719: [curl] Gramine's minimized `libcurl.a` doesn't have support for network proxies

I bumped into this problem when experimenting with the Intel Trust Authority (ITA) attestation service. I was trying the `ra-tls-mbedtls` example, and its `client` failed (note that the client doesn't even execute inside Gramine, it's a normal standalone program).

The `client` program is linked against the RA-TLS library. The RA-TLS library itself is statically linked against `libcurl.a`, which is a minimized version of libcurl that we have as a subproject.

At some point the Gramine team applied a `libcurl.a` minimization strategy as recommended here: https://github.com/curl/curl/blob/curl-8_4_0/docs/INSTALL.md#reducing-size

One recommendation from that web page is `--disable-proxy`. We applied it blindly to our curl build (in several places):
- https://github.com/gramineproject/gramine/blob/35f5d4b8b89bf713668bb1cfb5bae86c5d9ee387/subprojects/packagefiles/curl-8.4.0/compile.sh#L56?plain=1
- https://github.com/gramineproject/contrib/blob/eafc46ffb4c4782d83525d116a6450705bfa72eb/Integrations/azure/ra_tls_maa/subprojects/packagefiles/curl-7.84.0/compile.sh#L56
- (also in ITA PR that I was experimenting with: https://github.com/gramineproject/contrib/pull/59)

However, many companies have a proxy. This would imply that any app that would use our RA-TLS curl-based verification libraries (these are currently two: EPID and MAA) **fails** in such a company-proxy environment. In my case, the connection just hanged until the connection timeout (~60 seconds) triggered.

Some Curl references:
- https://github.com/curl/curl/blob/9034a16d974a826e1ba8ef7262646a941bb79fb2/docs/CURL-DISABLE.md?plain=1#L133
- https://github.com/curl/curl/blob/9034a16d974a826e1ba8ef7262646a941bb79fb2/configure.ac#L839-L854

### Solution

Remove the `--disable-proxy` build option.

Here is the current size of `libcurl.a`:
```
$ ls -lah build/subprojects/curl-7.84.0/
625K libcurl.a
```

Here is the new size of `libcurl.a` (when we re-enable proxy):
```
$ ls -lah build/subprojects/curl-7.84.0/
668K libcurl.a
```

I guess we can live with the `668 - 625 = 43` KB increase in size (~7% increase).

### Gramine commit hash

35f5d4b8b89bf713668bb1cfb5bae86c5d9ee387

dimakuv:
## Description of the changes 

For minimization purposes, libcurl is built with most features disabled. One of the disabled features was proxy support. This made libcurl-based dependencies, such as RA-TLS libs, unusable in proxy-restricted environments (e.g. with a company-wide proxy). This commit re-enables proxy support, which increases the size of `libcurl.a` by ~7%.

Fixes #1719. See it for detailed analysis.

## How to test this PR? 

Try RA-TLS with EPID in a proxy environment.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1720)

### DIFF
--- a/subprojects/packagefiles/curl-8.4.0/compile.sh
+++ b/subprojects/packagefiles/curl-8.4.0/compile.sh
@@ -29,6 +29,7 @@ cp -ar "$CURRENT_SOURCE_DIR" "$PRIVATE_DIR"
     # The list of configure options is selected based on:
     # https://github.com/curl/curl/blob/curl-8_4_0/docs/INSTALL.md#reducing-size
     ./configure                                     \
+        --enable-proxy                              \
         --disable-alt-svc                           \
         --disable-ares                              \
         --disable-cookies                           \
@@ -53,7 +54,6 @@ cp -ar "$CURRENT_SOURCE_DIR" "$PRIVATE_DIR"
         --disable-ntlm-wb                           \
         --disable-pop3                              \
         --disable-progress-meter                    \
-        --disable-proxy                             \
         --disable-pthreads                          \
         --disable-rtsp                              \
         --disable-shared                            \


kailun-qin (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1720)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1720)* status: 0 of 1 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1720#-NoMwTw10foSqI_zEn7F:-NoMwTw10foSqI_zEn7G:bdaprl9) (no related file):*
FYI: Updated based on comments from @woju in a companion PR https://github.com/gramineproject/contrib/pull/74

___

kailun-qin (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1720)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners

woju (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1720)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

kailun-qin (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1720)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

dimakuv:
I was asked if enabling a proxy is secure in our SGX threat model.

Let me copy my reply here, for context and so that we can discuss further:

References:
- https://security.stackexchange.com/questions/61334/does-an-https-proxy-encrypt-traffic-between-proxy-client-and-server-for-http-req
- https://security.stackexchange.com/questions/253951/does-a-http-proxy-ziproxy-squid-encrypt-communication-between-client-and-serve

Conclusions: there are 2 main types of proxies: (a) end-to-end client to server encryption, the proxy sees only the URL but the data is seen encrypted, (b) end-to-end client to proxy, then end-to-end proxy to server, proxy can see data in plaintext, this is typically used in corporate environments.

The second type of proxy is the concern. But I would argue that this is irrelevant anyway in our usage of Curl:
- The verifier in our examples is not inside the SGX enclave. It's a standalone program that runs on the trusted machine and talks to MAA/ITA, and the proxy should be also trusted in this scenario.
- If the verifier is put inside the SGX enclave, then we need to be very careful with proxies, that's true. But this becomes the responsibility of the SGX enclave developer -- proxy is specified only via environment variables, and the Gramine manifest fully controls the environment variables.

So I'd say "that's not a problem typically, and when it is a problem, it's not the Gramine core problem but the problem of the Gramine manifest writer".

PR #1721, opened 2024-01-16
[LibOS] Have async thread set a callback on libos_handle to take it off the list before it is freed. [v2]

stefanberger:
## Description of the changes 

This PR fixes issue https://github.com/gramineproject/gramine/issues/1525 where a PAL_HANDLE was freed even though this PAL_HANDLE still had an async event associated with it.  The solution I am using now is to add support for a callback to libos_handle where that callback is called before the libos_handle is freed once its reference counter has reached 0, thus avoiding the issue. The callback will be set for as long as the async thread has a libos_handle on its list and once it is removed from the list the callback will be set to NULL. 

This PR is improvement to https://github.com/gramineproject/gramine/pull/1585

## How to test this PR? 

A test case is provided.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1721)

### DIFF
--- a/libos/include/libos_handle.h
+++ b/libos/include/libos_handle.h
@@ -134,6 +134,8 @@ struct libos_epoll_handle {
     size_t last_returned_index;
 };
 
+typedef void (*libos_handle_free_callback_t)(struct libos_handle* handle);
+
 struct libos_handle {
     enum libos_handle_type type;
     bool is_dir;
@@ -155,6 +157,7 @@ struct libos_handle {
     bool created_by_process;
 
     refcount_t ref_count;
+    libos_handle_free_callback_t free_callback;
 
     struct libos_fs* fs;
     struct libos_dentry* dentry;
@@ -225,6 +228,16 @@ struct libos_handle {
 struct libos_handle* get_new_handle(void);
 void get_handle(struct libos_handle* hdl);
 void put_handle(struct libos_handle* hdl);
+
+/* Callback to call before freeing libos_handle */
+static inline libos_handle_free_callback_t
+set_handle_free_callback(struct libos_handle* hdl,
+                         libos_handle_free_callback_t cb)
+{
+    libos_handle_free_callback_t ret = hdl->free_callback;
+    hdl->free_callback = cb;
+    return ret;
+}
 
 /* Set handle to non-blocking or blocking mode. */
 int set_handle_nonblocking(struct libos_handle* hdl, bool on);
--- a/libos/include/libos_handle.h
+++ b/libos/include/libos_handle.h
@@ -157,7 +157,7 @@ struct libos_handle {
     bool created_by_process;
 
     refcount_t ref_count;
-    libos_handle_free_callback_t free_callback;
+    libos_handle_free_callback_t free_in_async_helper_thread_callback;
 
     struct libos_fs* fs;
     struct libos_dentry* dentry;
@@ -230,13 +230,9 @@ void get_handle(struct libos_handle* hdl);
 void put_handle(struct libos_handle* hdl);
 
 /* Callback to call before freeing libos_handle */
-static inline libos_handle_free_callback_t
-set_handle_free_callback(struct libos_handle* hdl,
-                         libos_handle_free_callback_t cb)
-{
-    libos_handle_free_callback_t ret = hdl->free_callback;
-    hdl->free_callback = cb;
-    return ret;
+static inline void set_handle_free_callback(struct libos_handle* hdl,
+                                            libos_handle_free_callback_t cb) {
+    hdl->free_in_async_helper_thread_callback = cb;
 }
 
 /* Set handle to non-blocking or blocking mode. */
--- a/libos/include/libos_utils.h
+++ b/libos/include/libos_utils.h
@@ -53,7 +53,7 @@ int create_pipe(char* name, char* uri, size_t size, PAL_HANDLE* hdl, bool use_vm
 
 /* Asynchronous event support */
 int init_async_worker(void);
-int64_t install_async_event(PAL_HANDLE object, unsigned long time,
+int64_t install_async_event(struct libos_handle* hdl, unsigned long time,
                             void (*callback)(IDTYPE caller, void* arg), void* arg);
 struct libos_thread* terminate_async_worker(void);
 
--- a/libos/src/bookkeep/libos_handle.c
+++ b/libos/src/bookkeep/libos_handle.c
@@ -512,8 +512,8 @@ void put_handle(struct libos_handle* hdl) {
         assert(hdl->epoll_items_count == 0);
         assert(LISTP_EMPTY(&hdl->epoll_items));
 
-        if (hdl->free_callback) {
-            hdl->free_callback(hdl);
+        if (hdl->free_in_async_helper_thread_callback) {
+            hdl->free_in_async_helper_thread_callback(hdl);
         }
 
         if (hdl->is_dir) {
--- a/libos/src/bookkeep/libos_handle.c
+++ b/libos/src/bookkeep/libos_handle.c
@@ -512,6 +512,10 @@ void put_handle(struct libos_handle* hdl) {
         assert(hdl->epoll_items_count == 0);
         assert(LISTP_EMPTY(&hdl->epoll_items));
 
+        if (hdl->free_callback) {
+            hdl->free_callback(hdl);
+        }
+
         if (hdl->is_dir) {
             clear_directory_handle(hdl);
         }
--- a/libos/src/libos_async.c
+++ b/libos/src/libos_async.c
@@ -23,12 +23,21 @@ struct async_event {
     LIST_TYPE(async_event) triggered_list;
     void (*callback)(IDTYPE caller, void* arg);
     void* arg;
-    PAL_HANDLE object;       /* handle (async IO) to wait on */
-    uint64_t expire_time_us; /* alarm/timer to wait on */
-    struct libos_handle* hdl;
+    struct libos_handle* hdl; /* hdl->pal_handle (async IO) to wait on */
+    uint64_t expire_time_us;  /* alarm/timer to wait on */
 };
 DEFINE_LISTP(async_event);
 static LISTP_TYPE(async_event) async_list;
+
+/* List of events thread needs to remove from async_list */
+DEFINE_LIST(remove_event);
+struct remove_event {
+    LIST_TYPE(remove_event) list;
+    struct libos_handle* hdl;
+    PAL_HANDLE done_notifier;
+};
+DEFINE_LISTP(remove_event);
+static LISTP_TYPE(remove_event) hdls_to_remove_from_list;
 
 /* Should be accessed with async_worker_lock held. */
 static enum { WORKER_NOTALIVE, WORKER_ALIVE } async_worker_state;
@@ -41,25 +50,66 @@ static struct libos_pollable_event install_new_event;
 
 static int create_async_worker(void);
 
-/* Remove a PAL_HANDLE from the list of monitored handles.
- * This function will be called before the libos_handle is freed.
- */
-static void remove_pal_handle(struct libos_handle* hdl) {
-    lock(&async_worker_lock);
-    if (async_worker_state == WORKER_NOTALIVE)
-        goto exit;
+static inline PAL_HANDLE get_pal_handle(struct libos_handle* hdl) {
+    if (!hdl)
+        return NULL;
+    return hdl->pal_handle;
+}
 
+
+/* Remove a libos_handle from the async_list; caller must hold async_worker_lock */
+static void async_list_remove_pal_handle(struct libos_handle* hdl) {
     struct async_event* tmp, *n;
+
     LISTP_FOR_EACH_ENTRY_SAFE(tmp, n, &async_list, list) {
-        if (tmp->object == hdl->pal_handle) {
+        if (get_pal_handle(tmp->hdl) == hdl->pal_handle) {
             LISTP_DEL(tmp, &async_list, list);
             free(tmp);
             break;
         }
     }
+}
 
-exit:
+/* Clean up the async_list by removing all handles on hdls_to_remove_from_list list */
+static void async_list_cleanup_pal_handles(void) {
+    struct remove_event* tmp, *n;
+
+    lock(&async_worker_lock);
+
+    LISTP_FOR_EACH_ENTRY_SAFE(tmp, n, &hdls_to_remove_from_list, list) {
+        async_list_remove_pal_handle(tmp->hdl);
+        LISTP_DEL(tmp, &hdls_to_remove_from_list, list);
+        PalEventSet(tmp->done_notifier);
+    }
+
     unlock(&async_worker_lock);
+}
+
+/* Remove a PAL_HANDLE from the list of monitored handles.
+ * This function will be called before the libos_handle is freed.
+ */
+static void remove_pal_handle(struct libos_handle* hdl) {
+    lock(&async_worker_lock);
+
+    if (async_worker_state == WORKER_NOTALIVE) {
+        async_list_remove_pal_handle(hdl);
+        unlock(&async_worker_lock);
+        return;
+    }
+
+    struct remove_event remove_event = {
+        .hdl = hdl,
+    };
+    INIT_LIST_HEAD(&remove_event, list);
+    LISTP_ADD_TAIL(&remove_event, &hdls_to_remove_from_list, list);
+    assert(!PalEventCreate(&remove_event.done_notifier, /*init_signaled=*/false, /*auto_clear=*/false));
+
+    unlock(&async_worker_lock);
+
+    set_pollable_event(&install_new_event);
+
+    while (PalEventWait(remove_event.done_notifier, /*timeout=*/NULL) < 0);
+    PalObjectDestroy(remove_event.done_notifier);
 }
 
 
@@ -96,18 +146,15 @@ int64_t install_async_event(struct libos_handle *hdl, uint64_t time_us,
         return -ENOMEM;
     }
 
-    PAL_HANDLE object = hdl ? hdl->pal_handle : NULL;
-
     event->callback       = callback;
     event->arg            = arg;
     event->caller         = get_cur_tid();
-    event->object         = object;
     event->expire_time_us = time_us ? now_us + time_us : 0;
     event->hdl            = hdl;
 
     lock(&async_worker_lock);
 
-    if (callback != &cleanup_thread && !object) {
+    if (callback != &cleanup_thread && !get_pal_handle(hdl)) {
         /* This is alarm() or setitimer() emulation, treat both according to
          * alarm() syscall semantics: cancel any pending alarm/timer. */
         struct async_event* tmp;
@@ -120,7 +167,7 @@ int64_t install_async_event(struct libos_handle *hdl, uint64_t time_us,
 
                 LISTP_DEL(tmp, &async_list, list);
                 if (tmp->hdl)
-                    assert(set_handle_free_callback(tmp->hdl, NULL) == remove_pal_handle);
+                    set_handle_free_callback(tmp->hdl, NULL);
                 free(tmp);
             }
         }
@@ -135,7 +182,7 @@ int64_t install_async_event(struct libos_handle *hdl, uint64_t time_us,
     }
 
     if (event->hdl)
-        assert(set_handle_free_callback(event->hdl, remove_pal_handle) == NULL);
+        set_handle_free_callback(event->hdl, remove_pal_handle);
 
     INIT_LIST_HEAD(event, list);
     LISTP_ADD_TAIL(event, &async_list, list);
@@ -239,7 +286,7 @@ static int libos_async_worker(void* arg) {
         bool other_event = false;
         LISTP_FOR_EACH_ENTRY_SAFE(tmp, n, &async_list, list) {
             /* repopulate `pals` with IO events and find the next expiring alarm/timer */
-            if (tmp->object) {
+            if (get_pal_handle(tmp->hdl)) {
                 if (pals_cnt == pals_max_cnt) {
                     /* grow `pals` to accommodate more objects */
                     PAL_HANDLE* tmp_pals = malloc(sizeof(*tmp_pals) * (1 + pals_max_cnt * 2));
@@ -271,7 +318,7 @@ static int libos_async_worker(void* arg) {
                     ret_events = tmp_ret_events;
                 }
 
-                pals[pals_cnt + 1]       = tmp->object;
+                pals[pals_cnt + 1]       = get_pal_handle(tmp->hdl);
                 pal_events[pals_cnt + 1] = PAL_WAIT_READ;
                 ret_events[pals_cnt + 1] = 0;
                 pals_cnt++;
@@ -319,6 +366,8 @@ static int libos_async_worker(void* arg) {
         }
         bool polled = ret == 0;
 
+        async_list_cleanup_pal_handles();
+
         ret = PalSystemTimeQuery(&now_us);
         if (ret < 0) {
             log_error("PalSystemTimeQuery failed with: %s", pal_strerror(ret));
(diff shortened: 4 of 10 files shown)


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1721)* status: 0 of 10 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @stefanberger)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1721#-NoWB-a57UxRN7VUJijE:-NoWB-a57UxRN7VUJijF:bdumxtn) (no related file):*
I definitely like this new fix, but see my comments.

___
___
*[`libos/include/libos_handle.h` line 160 at r1](https://reviewable.io/reviews/gramineproject/gramine/1721#-NoWB4iR626a94Ho_JgI:-NoWB4iR626a94Ho_JgJ:b9n5wvv) ([raw file](https://github.com/gramineproject/gramine/blob/e0472ecd28faaeec27da7b852a5feea300edb937/libos/include/libos_handle.h#L160)):*
> ```objc
> 
>     refcount_t ref_count;
>     libos_handle_free_callback_t free_callback;
> ```

Actually, the name is too generic. This field is used particularly to set up the Async helper thread deletion event.

Otherwise, imagine we'll have a handle that can be both added to the Async helper thread (as done in this PR) **and** it needs some e.g. file-system cleanup. Then there should be **two events** called, but we have only one field, and one of the event setups' will override the other, and we have a "lost event".

So, let's make it very clear and call it `free_in_async_helper_thread_callback`. Or something equally clear.
___
*[`libos/include/libos_handle.h` line 235 at r1](https://reviewable.io/reviews/gramineproject/gramine/1721#-NoWAwbJ8mLig9XrE42o:-NoWAwbJ8mLig9XrE42p:bjefn43) ([raw file](https://github.com/gramineproject/gramine/blob/e0472ecd28faaeec27da7b852a5feea300edb937/libos/include/libos_handle.h#L235)):*
> ```objc
> static inline libos_handle_free_callback_t
> set_handle_free_callback(struct libos_handle* hdl,
>                          libos_handle_free_callback_t cb)
> ```

This coding style is different from out classic style, please fix
___
*[`libos/include/libos_handle.h` line 239 at r1](https://reviewable.io/reviews/gramineproject/gramine/1721#-NoWA_sUCT9wgz3k7JW0:-NoWA_sUCT9wgz3k7JW1:b-v2qg70) ([raw file](https://github.com/gramineproject/gramine/blob/e0472ecd28faaeec27da7b852a5feea300edb937/libos/include/libos_handle.h#L239)):*
> ```objc
>     libos_handle_free_callback_t ret = hdl->free_callback;
>     hdl->free_callback = cb;
>     return ret;
> ```

Why do you need this return? Why not make the function `void`? Looks like unnecessary complication.
___
*[`libos/src/libos_async.c` line 26 at r1](https://reviewable.io/reviews/gramineproject/gramine/1721#-NoWAS6tD9aZFI1wkleP:-NoWAS6tD9aZFI1wkleQ:b-6kbhip) ([raw file](https://github.com/gramineproject/gramine/blob/e0472ecd28faaeec27da7b852a5feea300edb937/libos/src/libos_async.c#L26)):*
> ```C
>     void (*callback)(IDTYPE caller, void* arg);
>     void* arg;
>     PAL_HANDLE object;    /* handle (async IO) to wait on */
> ```

There is no need for this explicit `object` field now. You can replace it with `hdl->pal_handle` everywhere.
___
*[`libos/src/libos_async.c` line 56 at r1](https://reviewable.io/reviews/gramineproject/gramine/1721#-NoW8PVrC1Nd2C_VoMHJ:-NoW8PVrC1Nd2C_VoMHK:b6t7z1y) ([raw file](https://github.com/gramineproject/gramine/blob/e0472ecd28faaeec27da7b852a5feea300edb937/libos/src/libos_async.c#L56)):*
> ```C
>         if (tmp->object == hdl->pal_handle) {
>             LISTP_DEL(tmp, &async_list, list);
>             free(tmp);
> ```

This is a wrong logic, which leads to a **dangling pointer bug**.

Imagine we have two threads: one main thread and one Async helper thread.
- The main thread installs the Async I/O libos handle
- Async helper thread adds this libos handle's PAL handle to its list of handles-to-wait-on
- Async helper thread calls `PalStreamsWaitEvents()` and sleeps on it
- The main thread now deletes the libos handle and ends up in this `remove_pal_handle()`
- The main thread frees the object, and now the "PAL handle in the list of handles-to-wait-on" is a dangling pointer.
- Async helper thread will wake up in the future, and may accidentally touch this PAL handle and boom.

In other words, the object ownership here is incorrect: it is actually the Async helper thread who uses the LibOS handle object (and the embedded PAL handle object), so it must be the responsibility of the Async helper thread to get rid of this object (i.e. remove it from the `async_list`).

I think the solution is simple:
1. In this `remove_pal_handle()` callback you add this `hdl` to a list (you'll need a new global list "hlds-to-remove-from-list").
2. You call `set_pollable_event(&install_new_event)` to wake up the Async Helper thread.
3. You go into a busy loop waiting for the list to become empty.
4. You're done.

And in the Async helper thread context, you need the following:
1. Move your delete-from-async-list logic into `libos_async_worker()` function, somewhere after `PalStreamsWaitEvents()`.

All this logic should be correctly protected with the `async_worker_lock`.

Of course, the busy loop at step 3 is a potential performance problem. But I argue that this Async I/O deletion is so rare that we shouldn't care about performance of this logic.
___
*[`libos/src/libos_async.c` line 123 at r1](https://reviewable.io/reviews/gramineproject/gramine/1721#-NoWAfwP2onHRQQGSFHC:-NoWAfwP2onHRQQGSFHD:bgufiyq) ([raw file](https://github.com/gramineproject/gramine/blob/e0472ecd28faaeec27da7b852a5feea300edb937/libos/src/libos_async.c#L123)):*
> ```C
>                 LISTP_DEL(tmp, &async_list, list);
>                 if (tmp->hdl)
>                     assert(set_handle_free_callback(tmp->hdl, NULL) == remove_pal_handle);
> ```

What's the point of this? For alarms/timers, we don't have a libos handle, so you'll never get into here.

If you want to, you can add `assert(!tmp->hdl)` to be sure.

stefanberger (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1721)* status: 0 of 10 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv)

___
*[`libos/include/libos_handle.h` line 160 at r1](https://reviewable.io/reviews/gramineproject/gramine/1721#-NoWB4iR626a94Ho_JgI:-NoWaCmWDTxGyC3uRPKe:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/e0472ecd28faaeec27da7b852a5feea300edb937/libos/include/libos_handle.h#L160)):*

Actually, the name is too generic. This field is used particularly to set up the Async helper thread deletion event.

Otherwise, imagine we'll have a handle that can be both added to the Async helper thread (as done in this PR) **and** it needs some e.g. file-system cleanup. Then there should be **two events** called, but we have only one field, and one of the event setups' will override the other, and we have a "lost event".

So, let's make it very clear and call it `free_in_async_helper_thread_callback`. Or something equally clear.
</blockquote></details>

Done.
___
*[`libos/include/libos_handle.h` line 235 at r1](https://reviewable.io/reviews/gramineproject/gramine/1721#-NoWAwbJ8mLig9XrE42o:-NoWbeYV1M2dm-dAoV1C:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/e0472ecd28faaeec27da7b852a5feea300edb937/libos/include/libos_handle.h#L235)):*

This coding style is different from out classic style, please fix
</blockquote></details>

Done.
___
*[`libos/include/libos_handle.h` line 239 at r1](https://reviewable.io/reviews/gramineproject/gramine/1721#-NoWA_sUCT9wgz3k7JW0:-NoWatSdDS832KU5fDTK:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/e0472ecd28faaeec27da7b852a5feea300edb937/libos/include/libos_handle.h#L239)):*

Why do you need this return? Why not make the function `void`? Looks like unnecessary complication.
</blockquote></details>

Done.
___
*[`libos/src/libos_async.c` line 26 at r1](https://reviewable.io/reviews/gramineproject/gramine/1721#-NoWAS6tD9aZFI1wkleP:-NoWejfGDmfaKh5W9eSK:b4ohg6v) ([raw file](https://github.com/gramineproject/gramine/blob/e0472ecd28faaeec27da7b852a5feea300edb937/libos/src/libos_async.c#L26)):*

There is no need for this explicit `object` field now. You can replace it with `hdl->pal_handle` everywhere.
</blockquote></details>

Done. Using a static inline...
___
*[`libos/src/libos_async.c` line 56 at r1](https://reviewable.io/reviews/gramineproject/gramine/1721#-NoW8PVrC1Nd2C_VoMHJ:-NoWqL1c0owUu-hJN8yN:b1vnb3e) ([raw file](https://github.com/gramineproject/gramine/blob/e0472ecd28faaeec27da7b852a5feea300edb937/libos/src/libos_async.c#L56)):*

This is a wrong logic, which leads to a **dangling pointer bug**.

Imagine we have two threads: one main thread and one Async helper thread.
- The main thread installs the Async I/O libos handle
- Async helper thread adds this libos handle's PAL handle to its list of handles-to-wait-on
- Async helper thread calls `PalStreamsWaitEvents()` and sleeps on it
- The main thread now deletes the libos handle and ends up in this `remove_pal_handle()`
- The main thread frees the object, and now the "PAL handle in the list of handles-to-wait-on" is a dangling pointer.
- Async helper thread will wake up in the future, and may accidentally touch this PAL handle and boom.

In other words, the object ownership here is incorrect: it is actually the Async helper thread who uses the LibOS handle object (and the embedded PAL handle object), so it must be the responsibility of the Async helper thread to get rid of this object (i.e. remove it from the `async_list`).

I think the solution is simple:
1. In this `remove_pal_handle()` callback you add this `hdl` to a list (you'll need a new global list "hlds-to-remove-from-list").
2. You call `set_pollable_event(&install_new_event)` to wake up the Async Helper thread.
3. You go into a busy loop waiting for the list to become empty.
4. You're done.

And in the Async helper thread context, you need the following:
1. Move your delete-from-async-list logic into `libos_async_worker()` function, somewhere after `PalStreamsWaitEvents()`.

All this logic should be correctly protected with the `async_worker_lock`.

Of course, the busy loop at step 3 is a potential performance problem. But I argue that this Async I/O deletion is so rare that we shouldn't care about performance of this logic.
</blockquote></details>

Done. Using a PalEvent to wait for the async worker to indicate that it is done, though.
___
*[`libos/src/libos_async.c` line 123 at r1](https://reviewable.io/reviews/gramineproject/gramine/1721#-NoWAfwP2onHRQQGSFHC:-NoWqWIE3XydZxI7XQ4J:b-ohmpjq) ([raw file](https://github.com/gramineproject/gramine/blob/e0472ecd28faaeec27da7b852a5feea300edb937/libos/src/libos_async.c#L123)):*

What's the point of this? For alarms/timers, we don't have a libos handle, so you'll never get into here.

If you want to, you can add `assert(!tmp->hdl)` to be sure.
</blockquote></details>

Done. These asserts were due to the generic name of the field ...

stefanberger:
*[`libos/src/libos_async.c` line 105 at r2](https://reviewable.io/reviews/gramineproject/gramine/1721#-NoX5tfY8Xqw0-HWneOv:-NoX5tfY8Xqw0-HWneOw:b-wcpl1n) ([raw file](https://github.com/gramineproject/gramine/blob/7846718b7f76683ae2bd365fa0ed375d26c218ef/libos/src/libos_async.c#L105)):*
> ```C
>     INIT_LIST_HEAD(&remove_event, list);
>     LISTP_ADD_TAIL(&remove_event, &hdls_to_remove_from_list, list);
>     assert(!PalEventCreate(&remove_event.done_notifier, /*init_signaled=*/false, /*auto_clear=*/false));
> ```

I thought using an event was better than polling on an empty list but then there's this assert due to potential OOM condition...

stefanberger:
*[`libos/include/libos_handle.h` line 233 at r2](https://reviewable.io/reviews/gramineproject/gramine/1721#-NoX6enjBEjtWkDCuHvX:-NoX6enjBEjtWkDCuHvY:bix7hzq) ([raw file](https://github.com/gramineproject/gramine/blob/7846718b7f76683ae2bd365fa0ed375d26c218ef/libos/include/libos_handle.h#L233)):*
> ```objc
> 
> /* Callback to call before freeing libos_handle */
> static inline void set_handle_free_callback(struct libos_handle* hdl,
> ```

This function name may now also be a bit too generic though it becomes so long when I call it set_handle_free_async_helper_thread_callback().

stefanberger (commented):
> Reviewed 7 of 10 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1721)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

stefanberger (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1721)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`libos/include/libos_handle.h` line 233 at r2](https://reviewable.io/reviews/gramineproject/gramine/1721#-NoX6enjBEjtWkDCuHvX:-NpQLDbv8zs3bDCXcjOk:ba2525q) ([raw file](https://github.com/gramineproject/gramine/blob/7846718b7f76683ae2bd365fa0ed375d26c218ef/libos/include/libos_handle.h#L233)):*

This function name may now also be a bit too generic though it becomes so long when I call it set_handle_free_async_helper_thread_callback().
</blockquote></details>

@dimakuv Should I use the above mentioned function name as well?

stefanberger (commented):
Any more comments?

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1721#-:-NsKuaaB7dX_VyL3Z-Gb:bpcz3h5)* status: 5 of 10 files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

mkow:
*[`libos/src/libos_async.c` line 62 at r3](https://reviewable.io/reviews/gramineproject/gramine/1721#-O46LsDCDnOWNl9FDDat:-O46LsDCDnOWNl9FDDau:beziv4i) ([raw file](https://github.com/gramineproject/gramine/blob/b19ba95232546afe04becd1a2bed160c0ae36205/libos/src/libos_async.c#L62)):*
> ```C
> /* Remove a libos_handle from the async_list; caller must hold async_worker_lock */
> static void async_list_remove_pal_handle(struct libos_handle* hdl) {
>     struct async_event* tmp, *n;
> ```

Multiple pointers in a single line are not allowed, because of the left-alignment of our asterisks: https://gramine.readthedocs.io/en/stable/devel/coding-style.html#code-formatting

mkow:
*[`libos/src/libos_async.c` line 105 at r3](https://reviewable.io/reviews/gramineproject/gramine/1721#-O47nI2C-Houm2CP7aiV:-O47nI2C-Houm2CP7aiW:b-ezdvwt) ([raw file](https://github.com/gramineproject/gramine/blob/b19ba95232546afe04becd1a2bed160c0ae36205/libos/src/libos_async.c#L105)):*
> ```C
>     INIT_LIST_HEAD(&remove_event, list);
>     LISTP_ADD_TAIL(&remove_event, &hdls_to_remove_from_list, list);
>     assert(!PalEventCreate(&remove_event.done_notifier, /*init_signaled=*/false, /*auto_clear=*/false));
> ```

This call is completely erased from release builds.

mkow:
*[`libos/src/libos_async.c` line 105 at r3](https://reviewable.io/reviews/gramineproject/gramine/1721#-O47nI2C-Houm2CP7aiV:-O47nYEN8iXOBcf4BfFe:b-zen1qj) ([raw file](https://github.com/gramineproject/gramine/blob/b19ba95232546afe04becd1a2bed160c0ae36205/libos/src/libos_async.c#L105)):*

This call is completely erased from release builds.
</blockquote></details>

Have you tested this code in Release mode? It should just crash just a few lines below, if any test uses this function (if not, then then it means that this PR doesn't include enough tests).

kailun-qin (changes requested):
> Reviewed 3 of 10 files at r1, 2 of 3 files at r2, 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1721)* status: all files reviewed, 18 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @stefanberger)

___
*[`libos/src/libos_async.c` line 123 at r1](https://reviewable.io/reviews/gramineproject/gramine/1721#-NoWAfwP2onHRQQGSFHC:-O4E5rW87CHTM5SXUHEh:bfhqmln) ([raw file](https://github.com/gramineproject/gramine/blob/e0472ecd28faaeec27da7b852a5feea300edb937/libos/src/libos_async.c#L123)):*

Done. These asserts were due to the generic name of the field ...
</blockquote></details>

Not done?
___
*[`libos/src/libos_async.c` line 60 at r3](https://reviewable.io/reviews/gramineproject/gramine/1721#-O4Dz497CuWknjechx7T:-O4Dz497CuWknjechx7U:beqgzpj) ([raw file](https://github.com/gramineproject/gramine/blob/b19ba95232546afe04becd1a2bed160c0ae36205/libos/src/libos_async.c#L60)):*
> ```C
> 
> 
> /* Remove a libos_handle from the async_list; caller must hold async_worker_lock */
> ```

Can we add it as asserts?



_Code quote:_
```C
caller must hold async_worker_lock
```
___
*[`libos/src/libos_async.c` line 105 at r3](https://reviewable.io/reviews/gramineproject/gramine/1721#-O47nI2C-Houm2CP7aiV:-O4DwkDD8wL2aZxWiEAx:bd6zuzw) ([raw file](https://github.com/gramineproject/gramine/blob/b19ba95232546afe04becd1a2bed160c0ae36205/libos/src/libos_async.c#L105)):*

Have you tested this code in Release mode? It should just crash just a few lines below, if any test uses this function (if not, then then it means that this PR doesn't include enough tests).
</blockquote></details>

+1, the test only checks that no assertion occured (the one reported in the original issue: https://github.com/gramineproject/gramine/issues/1525) but does not check if the test would run successfully: https://github.com/gramineproject/gramine/blob/b19ba95232546afe04becd1a2bed160c0ae36205/libos/test/regression/test_libos.py#L1006
___
*[`libos/src/libos_async.c` line 115 at r3](https://reviewable.io/reviews/gramineproject/gramine/1721#-O4E8acz7VH9iUmpVuQ3:-O4E8acz7VH9iUmpVuQ4:b-rfki8w) ([raw file](https://github.com/gramineproject/gramine/blob/b19ba95232546afe04becd1a2bed160c0ae36205/libos/src/libos_async.c#L115)):*
> ```C
> }
> ```

redundant empty line?
___
*[`libos/src/libos_async.c` line 126 at r3](https://reviewable.io/reviews/gramineproject/gramine/1721#-O4E9wN-AkrBcFqEu4eo:-O4E9wN-AkrBcFqEu4ep:bhlabpn) ([raw file](https://github.com/gramineproject/gramine/blob/b19ba95232546afe04becd1a2bed160c0ae36205/libos/src/libos_async.c#L126)):*
> ```C
>  *   - alarm/timer events set object = NULL and time_us = microseconds
>  *     (time_us = 0 cancels all pending alarms/timers).
>  *   - async IO events set object = handle and time_us = 0.
> ```

Pls update this accordinly



_Code quote:_
```C
 *   - alarm/timer events set object = NULL and time_us = microseconds
 *     (time_us = 0 cancels all pending alarms/timers).
 *   - async IO events set object = handle and time_us = 0.
```
___
*[`libos/src/libos_async.c` line 131 at r3](https://reviewable.io/reviews/gramineproject/gramine/1721#-O4E9rg67OBTKywlmKiR:-O4E9rg67OBTKywlmKiS:b-mfwcti) ([raw file](https://github.com/gramineproject/gramine/blob/b19ba95232546afe04becd1a2bed160c0ae36205/libos/src/libos_async.c#L131)):*
> ```C
>  * or 0 for async IO events. On error, it returns a negated error code.
>  */
> int64_t install_async_event(struct libos_handle *hdl, uint64_t time_us,
> ```

Pls also update the comment here: https://github.com/gramineproject/gramine/blob/0c0971cd6930400b39b42eb92656b3968ae75698/libos/src/sys/libos_alarm.c#L69.
___
*[`libos/test/regression/async_task_shutdown.c` line 12 at r3](https://reviewable.io/reviews/gramineproject/gramine/1721#-O4DxkbbCaqzx6SW195j:-O4DxkbbCaqzx6SW195k:bfcmsrx) ([raw file](https://github.com/gramineproject/gramine/blob/b19ba95232546afe04becd1a2bed160c0ae36205/libos/test/regression/async_task_shutdown.c#L12)):*
> ```C
> #include <stdio.h>
> #include <sys/ioctl.h>
> #include <sys/wait.h>
> ```

Can we sort the headers?
___
*[`libos/test/regression/async_task_shutdown.c` line 13 at r3](https://reviewable.io/reviews/gramineproject/gramine/1721#-O4DylO748vXyQNe3-hm:-O4DylO748vXyQNe3-hn:bsksmf7) ([raw file](https://github.com/gramineproject/gramine/blob/b19ba95232546afe04becd1a2bed160c0ae36205/libos/test/regression/async_task_shutdown.c#L13)):*
> ```C
> #include <sys/ioctl.h>
> #include <sys/wait.h>
> ```

Can we follow the conventions for testing here (e.g., using `CHECK()` macro, checking "TEST OK" etc.)?
___
*[`libos/test/regression/test_libos.py` line 1004 at r3](https://reviewable.io/reviews/gramineproject/gramine/1721#-O4DxuZW6vsDusGRxfo3:-O4DxuZW6vsDusGRxfo4:bfsuqg7) ([raw file](https://github.com/gramineproject/gramine/blob/b19ba95232546afe04becd1a2bed160c0ae36205/libos/test/regression/test_libos.py#L1004)):*
> ```Python
> 
>     def test_160_async_task_shutdown(self):
>         for i in range(0, 200):
> ```

I understand the assertion happens occasionally, only when async helper thread executes into it -- but this would run the test 200 times which would take quite amount of time



_Code quote:_
```Python
for i in range(0, 200):
```
___
*[`libos/test/regression/test_libos.py` line 1006 at r3](https://reviewable.io/reviews/gramineproject/gramine/1721#-O4DxoAf5arJvnkMob8g:-O4DxoAf5arJvnkMob8h:b-dpmap) ([raw file](https://github.com/gramineproject/gramine/blob/b19ba95232546afe04becd1a2bed160c0ae36205/libos/test/regression/test_libos.py#L1006)):*
> ```Python
>         for i in range(0, 200):
>             _, stderr = self.run_binary(['async_task_shutdown'])
>             self.assertNotIn('assert failed', stderr)
> ```

I think we should also check the test actually completes successfully



_Code quote:_
```Python
self.assertNotIn('assert failed', stderr)
```

monavij:
@stefanberger Why did you close this PR? Is this issue resolved by some other PR?

stefanberger:
> @stefanberger Why did you close this PR? Is this issue resolved by some other PR?

I think https://github.com/gramineproject/gramine/pull/1992 solved it.

kailun-qin:
> I think https://github.com/gramineproject/gramine/pull/1992 solved it.

@jinengandhi-intel: can we check if https://github.com/gramineproject/gramine/pull/1992 resolves the original issue: https://github.com/gramineproject/gramine/issues/1525?

jinengandhi-intel (commented):
This issue was consistently seen when we ran Syzkaller based fuzzing, we plan to trigger the fuzzing runs once we code freeze for Gramine v1.8, which should be in a  day or two, I will keep an eye out for this and update the same here.

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1721#-:-O7b9dxo9FhkqYtizS7m:bxivzul)* status: all files reviewed, 18 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @stefanberger)

Issue #1714, opened 2024-01-11
[Encrypted FS] Use pre-allocacted free list instead of calloc/free for file nodes


kailun-qin:
### Description of the feature

Instead of using calloc/free directly for the Encrypted FS file nodes (e.g., MHT nodes, data nodes), we can introduce a free-list cache for them (i.e., memory chunks exclusively reserved for encrypted-files usage), as a perf optimization.

PoC (for testing and illustration purposes **ONLY**):
https://github.com/gramineproject/gramine/compare/master...kailun-qin:gramine:hack-file-node-free-list
> Note: We should implement fallbacks to proper malloc/calloc when running out of free nodes (instead of `BUG()` in the PoC).

TO BE DISCUSSED:
Should we expose the size of the free list as a new Gramine manifest option? For example, if the manifest e.g. sets it to 8GB, then these 8GB will be put aside as a memory pool for encrypted-files.

### Why Gramine should implement it?

To resolve perf bottlenecks seen on some workloads (e.g., RocksDB).

Specifically, RocksDB has a design of [leveled compaction](https://github.com/facebook/rocksdb/wiki/Leveled-Compaction). It operates on a key range and two levels, it opens the files that contains the ranges and iterate sequentially over the files and merges them into new sorted files where the stale keys are being omitted.

When this compaction process happened, RocksDB throughput was reduced by ~10x when running w/ Gramine's encrypted FS.

The ongoing optimization for encrypted FS (https://github.com/gramineproject/gramine/issues/1599, https://github.com/gramineproject/gramine/pull/1681) don't help much in this case. The perf analysis showed that the one of the dominant contributors was the memory allocation of file data nodes: https://github.com/gramineproject/gramine/blob/5d38bb14077cf87f1ddb3cef07f6f2ab70621276/common/src/protected_files/protected_files.c#L620

Gramine commit hash
1f72aaf8148a7712c5c8bad1da9839e2bcc5c633

dimakuv:
I reviewed this proposal privately before, and I'm fine with it. I think it's a simple and efficient optimization.

@kailun-qin Would it be possible to show some numbers on the RocksDB workload?

Alternatively, can we write a micro-benchmark that shows the benefit of this optimization? I guess a single large Protected File that is constantly read from and written to is enough.

> Should we expose the size of the free list as a new Gramine manifest option? For example, if the manifest e.g. sets it to 8GB, then these 8GB will be put aside as a memory pool for encrypted-files.

[...]

PR #1723, opened 2024-01-18
[common] Introduce pre-allocated node free list for encrypted files

kailun-qin:
## Description of the changes 
Allocation of the Encrypted FS file nodes (e.g., MHT nodes, data nodes) was found to be a performance bottleneck for certain workloads (e.g., RocksDB compaction).

Therefore, instead of using alloc/free directly on file nodes, this commit introduces a free list of pre-allocacted nodes reserved specifically for their usage, as a performance optimization that trades space for time. A new manifest option has been added to allow users to set and adjust the free list limit based on their applications.

Fixes https://github.com/gramineproject/gramine/issues/1714.

## How to test this PR? 

CI and RocksDB.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1723)

### DIFF
--- a/Documentation/manifest-syntax.rst
+++ b/Documentation/manifest-syntax.rst
@@ -1021,6 +1021,20 @@ Gramine:
    in the application is insecure. If you need to derive encryption keys from
    such a "doubly-used" key, you must apply a KDF.
 
+Encrypted files free list node limit
+
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+::
+
+    fs.limit.encrypted_files_node_free_list = [NUM]
+    (Default: 0)
+
+This syntax specifies a limit on the number of nodes in the pre-allocated
+encrypted files free list, as a performance optimization that trades space for
+time. By default, this optimization is disabled as the limit should be
+application-specific.
+
 .. _untrusted-shared-memory:
 
 Untrusted shared memory
--- a/Documentation/manifest-syntax.rst
+++ b/Documentation/manifest-syntax.rst
@@ -1021,6 +1021,8 @@ Gramine:
    in the application is insecure. If you need to derive encryption keys from
    such a "doubly-used" key, you must apply a KDF.
 
+.. _encrypted-files-free-list-node-limit:
+
 Encrypted files optimization: free list node limit
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
--- a/Documentation/manifest-syntax.rst
+++ b/Documentation/manifest-syntax.rst
@@ -1021,17 +1021,16 @@ Gramine:
    in the application is insecure. If you need to derive encryption keys from
    such a "doubly-used" key, you must apply a KDF.
 
-Encrypted files free list node limit
-
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Encrypted files optimization: free list node limit
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 ::
 
-    fs.limit.encrypted_files_node_free_list = [NUM]
+    fs.limits.encrypted_files_free_list_nodes = [NUM]
     (Default: 0)
 
 This syntax specifies a limit on the number of nodes in the pre-allocated
-encrypted files free list, as a performance optimization that trades space for
+encrypted files free list, as a performance optimization that trades memory for
 time. By default, this optimization is disabled as the limit should be
 application-specific.
 
--- a/Documentation/performance.rst
+++ b/Documentation/performance.rst
@@ -305,6 +305,17 @@ in Gramine:
    all IPC is transparently encrypted/decrypted using the TLS-PSK with AES-GCM
    crypto.
 
+Optimizations for file I/O
+--------------------------
+
+Allocation of the Encrypted FS file nodes (e.g., MHT nodes, data nodes) can be
+one source of overhead for certain workloads (e.g., RocksDB compaction). To
+address this performance bottleneck, instead of using alloc/free directly on
+file nodes, Gramine provides an optional free list of pre-allocacted nodes as an
+optimization that trades memory for time. This optimization is by default
+disabled, but can be enabled and tuned according to the needs of the application
+via the manifest option ``fs.limits.encrypted_files_node_free_list``.
+
 .. _choice_of_sgx_machine:
 
 Choice of SGX machine
--- a/Documentation/performance.rst
+++ b/Documentation/performance.rst
@@ -308,13 +308,20 @@ in Gramine:
 Optimizations for file I/O
 --------------------------
 
-Allocation of the Encrypted FS file nodes (e.g., MHT nodes, data nodes) can be
-one source of overhead for certain workloads (e.g., RocksDB compaction). To
-address this performance bottleneck, instead of using alloc/free directly on
-file nodes, Gramine provides an optional free list of pre-allocacted nodes as an
-optimization that trades memory for time. This optimization is by default
-disabled, but can be enabled and tuned according to the needs of the application
-via the manifest option ``fs.limits.encrypted_files_node_free_list``.
+An encrypted file is split into equally-sized chunks (nodes). The first node is
+the file metadata node, and the other nodes are stored as a variant of Merkle
+Hash Tree (MHT). In this MHT variant, MHT nodes host the crypto metadata like
+the encryption keys and MACs of the child nodes; data nodes host the actual
+encrypted data and are linked to MHT nodes.
+
+Allocation of the file nodes can be one source of overhead for certain workloads
+(e.g., RocksDB compaction). To address this performance bottleneck, instead of
+using alloc/free directly on file nodes, Gramine provides an optional free list
+of pre-allocated nodes as an optimization that trades memory for time. This
+optimization is by default disabled, but can be enabled and tuned according to
+the needs of the application via the manifest option
+``fs.limits.encrypted_files_node_free_list``. See
+:ref:`encrypted-files-free-list-node-limit` for more information.
 
 .. _choice_of_sgx_machine:
 
--- a/common/src/protected_files/protected_files.c
+++ b/common/src/protected_files/protected_files.c
@@ -515,7 +515,7 @@ pf_status_t pf_init_node_free_list(size_t limit_nodes) {
         struct pf_node_item* item = calloc(1, sizeof(struct pf_node_item));
         if (item == NULL) {
             log_error("Not enough memory for the encrypted files node free list. Please consider "
-                      "dicreasing the free list node limit via the manifest option "
+                      "decreasing the free list node limit via the manifest option "
                       "'fs.limits.encrypted_files_free_list_nodes'.");
             return PF_STATUS_NO_MEMORY;
         }
--- a/common/src/protected_files/protected_files.c
+++ b/common/src/protected_files/protected_files.c
@@ -572,8 +572,7 @@ static void ipf_free_node_item(struct pf_node_item* item) {
 static void ipf_free_node_item_from_data(file_node_t* data_ptr) {
     assert(data_ptr);
 
-    struct pf_node_item* item =
-        (struct pf_node_item*)((char*)data_ptr - offsetof(struct pf_node_item, file_node));
+    struct pf_node_item* item = container_of(data_ptr, struct pf_node_item, file_node);
 
     ipf_free_node_item(item);
 }
--- a/common/src/protected_files/protected_files.c
+++ b/common/src/protected_files/protected_files.c
@@ -7,8 +7,13 @@
 #include "protected_files.h"
 #include "protected_files_format.h"
 #include "protected_files_internal.h"
+#include "spinlock.h"
 
 #include "api.h"
+
+#ifdef IN_LIBOS
+#include "libos_thread.h"
+#endif
 
 /* Host callbacks */
 static pf_read_f     g_cb_read     = NULL;
@@ -490,6 +495,82 @@ static bool ipf_internal_flush(pf_context_t* pf) {
     return true;
 }
 
+struct pf_node_item {
+    struct pf_node_item* next;
+    file_node_t file_node;
+    bool from_free_list;
+};
+
+struct pf_node_item* g_pf_node_free_list = NULL;
+static spinlock_t g_pf_node_free_list_lock = INIT_SPINLOCK_UNLOCKED;
+
+pf_status_t pf_init_node_free_list(size_t limit_node_free_list) {
+    assert(!g_initialized);
+    struct pf_node_item* item;
+
+    for (size_t i = 0; i < limit_node_free_list; i++) {
+        item = calloc(1, sizeof(struct pf_node_item));
+        if (item == NULL) {
+            log_error("Not enough memory for the encrypted files node free list. Please consider "
+                      "dicreasing the limit in the manifest.");
+            return PF_STATUS_NO_MEMORY;
+        }
+
+        item->next = g_pf_node_free_list;
+        item->from_free_list = true;
+        g_pf_node_free_list = item;
+    }
+
+    return PF_STATUS_SUCCESS;
+}
+
+static struct pf_node_item* ipf_allocate_node(void) {
+    struct pf_node_item* item;
+
+    spinlock_lock(&g_pf_node_free_list_lock);
+    if (g_pf_node_free_list == NULL) {
+        spinlock_unlock(&g_pf_node_free_list_lock);
+
+        if (FIRST_TIME()) {
+            log_warning("No free file nodes available; using malloc as fallbacks. "
+                        "Please consider adjusting the limit of the encrypted files node free list "
+                        "in the manifest.");
+        }
+
+        item = calloc(1, sizeof(struct pf_node_item));
+        if (item == NULL)
+            return NULL;
+    } else {
+        item = g_pf_node_free_list;
+        g_pf_node_free_list = item->next;
+
+        spinlock_unlock(&g_pf_node_free_list_lock);
+        memset(&item->file_node, 0, sizeof(item->file_node));
+    }
+
+    return item;
+}
+
+static void ipf_free_node(struct pf_node_item* item) {
+    if (!item->from_free_list) {
+        free(item);
+    } else {
+        spinlock_lock(&g_pf_node_free_list_lock);
+
+        item->next = g_pf_node_free_list;
+        g_pf_node_free_list = item;
+
+        spinlock_unlock(&g_pf_node_free_list_lock);
+    }
+}
+
+static void ipf_free_node_from_data(file_node_t* data_ptr) {
+    struct pf_node_item* item =
+        (struct pf_node_item*)((char*)data_ptr - offsetof(struct pf_node_item, file_node));
+
+    ipf_free_node(item);
+}
+
 static file_node_t* ipf_get_mht_node(pf_context_t* pf, uint64_t offset) {
     file_node_t* file_mht_node;
     uint64_t mht_node_number;
@@ -529,12 +610,12 @@ static file_node_t* ipf_append_mht_node(pf_context_t* pf, uint64_t mht_node_numb
                                     // the '1' is for the mht node preceding every 96 data nodes
                                     mht_node_number * (1 + ATTACHED_DATA_NODES_COUNT);
 
-    file_node_t* new_file_mht_node = NULL;
-    new_file_mht_node = calloc(1, sizeof(*new_file_mht_node));
-    if (!new_file_mht_node) {
+    struct pf_node_item* new_file_mht_node_item = ipf_allocate_node();
+    if (!new_file_mht_node_item) {
         pf->last_error = PF_STATUS_NO_MEMORY;
         return NULL;
     }
+    file_node_t* new_file_mht_node = &(new_file_mht_node_item->file_node);
 
     new_file_mht_node->type = FILE_MHT_NODE_TYPE;
     new_file_mht_node->new_node = true;
@@ -543,14 +624,13 @@ static file_node_t* ipf_append_mht_node(pf_context_t* pf, uint64_t mht_node_numb
     new_file_mht_node->physical_node_number = physical_node_number;
 
     if (!lruc_add(pf->cache, new_file_mht_node->physical_node_number, new_file_mht_node)) {
-        free(new_file_mht_node);
+        ipf_free_node(new_file_mht_node_item);
         pf->last_error = PF_STATUS_NO_MEMORY;
         return NULL;
     }
 
     return new_file_mht_node;
 }
-
 
 static file_node_t* ipf_get_data_node(pf_context_t* pf, uint64_t offset) {
     file_node_t* file_data_node = NULL;
@@ -595,7 +675,7 @@ static file_node_t* ipf_get_data_node(pf_context_t* pf, uint64_t offset) {
             // before deleting the memory, need to scrub the plain secrets
             file_node_t* file_node = (file_node_t*)data;
             erase_memory(&file_node->decrypted, sizeof(file_node->decrypted));
-            free(file_node);
+            ipf_free_node_from_data(file_node);
         } else {
             if (!ipf_internal_flush(pf)) {
                 // error, can't flush cache, file status changed to error
@@ -615,13 +695,12 @@ static file_node_t* ipf_append_data_node(pf_context_t* pf, uint64_t offset) {
     if (file_mht_node == NULL) // some error happened
         return NULL;
 
-    file_node_t* new_file_data_node = NULL;
-
-    new_file_data_node = calloc(1, sizeof(*new_file_data_node));
-    if (!new_file_data_node) {
+    struct pf_node_item* new_file_data_node_item = ipf_allocate_node();
+    if (!new_file_data_node_item) {
         pf->last_error = PF_STATUS_NO_MEMORY;
         return NULL;
     }
+    file_node_t* new_file_data_node = &(new_file_data_node_item->file_node);
 
     uint64_t node_number, physical_node_number;
     get_node_numbers(offset, NULL, &node_number, NULL, &physical_node_number);
@@ -633,7 +712,7 @@ static file_node_t* ipf_append_data_node(pf_context_t* pf, uint64_t offset) {
     new_file_data_node->physical_node_number = physical_node_number;
 
     if (!lruc_add(pf->cache, new_file_data_node->physical_node_number, new_file_data_node)) {
-        free(new_file_data_node);
+        ipf_free_node(new_file_data_node_item);
         pf->last_error = PF_STATUS_NO_MEMORY;
         return NULL;
     }
@@ -659,11 +738,12 @@ static file_node_t* ipf_read_data_node(pf_context_t* pf, uint64_t offset) {
     if (file_mht_node == NULL) // some error happened
         return NULL;
 
-    file_data_node = calloc(1, sizeof(*file_data_node));
-    if (!file_data_node) {
+    struct pf_node_item* file_data_node_item = ipf_allocate_node();
+    if (!file_data_node_item) {
         pf->last_error = PF_STATUS_NO_MEMORY;
         return NULL;
     }
+    file_data_node = &(file_data_node_item->file_node);
 
     file_data_node->type = FILE_DATA_NODE_TYPE;
     file_data_node->node_number = data_node_number;
@@ -672,7 +752,7 @@ static file_node_t* ipf_read_data_node(pf_context_t* pf, uint64_t offset) {
 
     if (!ipf_read_node(pf, pf->file, file_data_node->physical_node_number,
                        file_data_node->encrypted.cipher, PF_NODE_SIZE)) {
-        free(file_data_node);
+        ipf_free_node(file_data_node_item);
         return NULL;
     }
 
@@ -686,7 +766,7 @@ static file_node_t* ipf_read_data_node(pf_context_t* pf, uint64_t offset) {
                                   file_data_node->decrypted.data.data, &gcm_crypto_data->gmac);
 
     if (PF_FAILURE(status)) {
-        free(file_data_node);
+        ipf_free_node(file_data_node_item);
         pf->last_error = status;
         if (status == PF_STATUS_MAC_MISMATCH)
             pf->file_status = PF_STATUS_CORRUPTED;
@@ -696,7 +776,7 @@ static file_node_t* ipf_read_data_node(pf_context_t* pf, uint64_t offset) {
     if (!lruc_add(pf->cache, file_data_node->physical_node_number, file_data_node)) {
         // scrub the plaintext data
         erase_memory(&file_data_node->decrypted, sizeof(file_data_node->decrypted));
-        free(file_data_node);
+        ipf_free_node(file_data_node_item);
         pf->last_error = PF_STATUS_NO_MEMORY;
         return NULL;
     }
@@ -724,11 +804,12 @@ static file_node_t* ipf_read_mht_node(pf_context_t* pf, uint64_t mht_node_number
     if (parent_file_mht_node == NULL) // some error happened
         return NULL;
 
-    file_mht_node = calloc(1, sizeof(*file_mht_node));
-    if (!file_mht_node) {
+    struct pf_node_item* file_mht_node_item = ipf_allocate_node();
+    if (!file_mht_node_item) {
         pf->last_error = PF_STATUS_NO_MEMORY;
         return NULL;
     }
+    file_mht_node = &(file_mht_node_item->file_node);
 
     file_mht_node->type                 = FILE_MHT_NODE_TYPE;
     file_mht_node->node_number          = mht_node_number;
@@ -737,7 +818,7 @@ static file_node_t* ipf_read_mht_node(pf_context_t* pf, uint64_t mht_node_number
 
     if (!ipf_read_node(pf, pf->file, file_mht_node->physical_node_number,
                        file_mht_node->encrypted.cipher, PF_NODE_SIZE)) {
-        free(file_mht_node);
+        ipf_free_node(file_mht_node_item);
         return NULL;
     }
 
@@ -750,7 +831,7 @@ static file_node_t* ipf_read_mht_node(pf_context_t* pf, uint64_t mht_node_number
                                   file_mht_node->encrypted.cipher, PF_NODE_SIZE,
                                   &file_mht_node->decrypted.mht, &gcm_crypto_data->gmac);
     if (PF_FAILURE(status)) {
-        free(file_mht_node);
+        ipf_free_node(file_mht_node_item);
         pf->last_error = status;
         if (status == PF_STATUS_MAC_MISMATCH)
             pf->file_status = PF_STATUS_CORRUPTED;
@@ -759,7 +840,7 @@ static file_node_t* ipf_read_mht_node(pf_context_t* pf, uint64_t mht_node_number
 
     if (!lruc_add(pf->cache, file_mht_node->physical_node_number, file_mht_node)) {
         erase_memory(&file_mht_node->decrypted, sizeof(file_mht_node->decrypted));
-        free(file_mht_node);
+        ipf_free_node(file_mht_node_item);
         pf->last_error = PF_STATUS_NO_MEMORY;
         return NULL;
     }
@@ -1151,7 +1232,7 @@ static bool ipf_close(pf_context_t* pf) {
     while ((data = lruc_get_last(pf->cache)) != NULL) {
         file_node_t* file_node = (file_node_t*)data;
         erase_memory(&file_node->decrypted, sizeof(file_node->decrypted));
-        free(file_node);
+        ipf_free_node_from_data(file_node);
         lruc_remove_last(pf->cache);
     }
 
@@ -1265,7 +1346,7 @@ pf_status_t pf_set_size(pf_context_t* pf, uint64_t size) {
         while ((data = lruc_get_last(pf->cache)) != NULL) {
             file_node_t* file_node = (file_node_t*)data;
             erase_memory(&file_node->decrypted, sizeof(file_node->decrypted));
-            free(file_node);
+            ipf_free_node_from_data(file_node);
             lruc_remove_last(pf->cache);
         }
 
--- a/common/src/protected_files/protected_files.c
+++ b/common/src/protected_files/protected_files.c
@@ -11,6 +11,9 @@
 
 #include "api.h"
 
+/* Function `get_cur_tid()` is forward declared as "static" in "spinlock.h" to keep the header
+ * independent. This requires that the function has been defined, which is actually done in
+ * "libos_thread.h". */
 #ifdef IN_LIBOS
 #include "libos_thread.h"
 #endif
@@ -501,18 +504,19 @@ struct pf_node_item {
     bool from_free_list;
 };
 
-struct pf_node_item* g_pf_node_free_list = NULL;
+static struct pf_node_item* g_pf_node_free_list = NULL;
+static bool g_pf_node_free_list_in_use = false;
 static spinlock_t g_pf_node_free_list_lock = INIT_SPINLOCK_UNLOCKED;
 
-pf_status_t pf_init_node_free_list(size_t limit_node_free_list) {
+pf_status_t pf_init_node_free_list(size_t limit_nodes) {
     assert(!g_initialized);
-    struct pf_node_item* item;
 
-    for (size_t i = 0; i < limit_node_free_list; i++) {
-        item = calloc(1, sizeof(struct pf_node_item));
+    for (size_t i = 0; i < limit_nodes; i++) {
+        struct pf_node_item* item = calloc(1, sizeof(struct pf_node_item));
         if (item == NULL) {
             log_error("Not enough memory for the encrypted files node free list. Please consider "
-                      "dicreasing the limit in the manifest.");
+                      "dicreasing the free list node limit via the manifest option "
+                      "'fs.limits.encrypted_files_free_list_nodes'.");
             return PF_STATUS_NO_MEMORY;
         }
 
@@ -521,54 +525,57 @@ pf_status_t pf_init_node_free_list(size_t limit_node_free_list) {
         g_pf_node_free_list = item;
     }
 
+    if (limit_nodes > 0)
+        g_pf_node_free_list_in_use = true;
+
     return PF_STATUS_SUCCESS;
 }
 
-static struct pf_node_item* ipf_allocate_node(void) {
-    struct pf_node_item* item;
-
+static struct pf_node_item* ipf_allocate_node_item(void) {
     spinlock_lock(&g_pf_node_free_list_lock);
+
     if (g_pf_node_free_list == NULL) {
         spinlock_unlock(&g_pf_node_free_list_lock);
 
-        if (FIRST_TIME()) {
-            log_warning("No free file nodes available; using malloc as fallbacks. "
-                        "Please consider adjusting the limit of the encrypted files node free list "
-                        "in the manifest.");
+        if (FIRST_TIME() && g_pf_node_free_list_in_use) {
+            log_warning("No free file nodes left; using malloc as fallback. "
+                        "Please consider adjusting the free list node limit via the manifest "
+                        "option 'fs.limits.encrypted_files_free_list_nodes'.");
         }
 
-        item = calloc(1, sizeof(struct pf_node_item));
-        if (item == NULL)
-            return NULL;
-    } else {
-        item = g_pf_node_free_list;
-        g_pf_node_free_list = item->next;
-
-        spinlock_unlock(&g_pf_node_free_list_lock);
-        memset(&item->file_node, 0, sizeof(item->file_node));
+        return calloc(1, sizeof(struct pf_node_item));
     }
+
+    struct pf_node_item* item = g_pf_node_free_list;
+    g_pf_node_free_list = item->next;
+
+    spinlock_unlock(&g_pf_node_free_list_lock);
+    memset(&item->file_node, 0, sizeof(item->file_node));
 
     return item;
 }
 
-static void ipf_free_node(struct pf_node_item* item) {
+static void ipf_free_node_item(struct pf_node_item* item) {
     if (!item->from_free_list) {
         free(item);
-    } else {
-        spinlock_lock(&g_pf_node_free_list_lock);
-
-        item->next = g_pf_node_free_list;
-        g_pf_node_free_list = item;
-
-        spinlock_unlock(&g_pf_node_free_list_lock);
+        return;
     }
+
+    spinlock_lock(&g_pf_node_free_list_lock);
+
+    item->next = g_pf_node_free_list;
+    g_pf_node_free_list = item;
+
+    spinlock_unlock(&g_pf_node_free_list_lock);
 }
 
-static void ipf_free_node_from_data(file_node_t* data_ptr) {
+static void ipf_free_node_item_from_data(file_node_t* data_ptr) {
+    assert(data_ptr);
+
     struct pf_node_item* item =
         (struct pf_node_item*)((char*)data_ptr - offsetof(struct pf_node_item, file_node));
 
-    ipf_free_node(item);
+    ipf_free_node_item(item);
 }
 
 static file_node_t* ipf_get_mht_node(pf_context_t* pf, uint64_t offset) {
@@ -610,12 +617,12 @@ static file_node_t* ipf_append_mht_node(pf_context_t* pf, uint64_t mht_node_numb
                                     // the '1' is for the mht node preceding every 96 data nodes
                                     mht_node_number * (1 + ATTACHED_DATA_NODES_COUNT);
 
-    struct pf_node_item* new_file_mht_node_item = ipf_allocate_node();
+    struct pf_node_item* new_file_mht_node_item = ipf_allocate_node_item();
     if (!new_file_mht_node_item) {
         pf->last_error = PF_STATUS_NO_MEMORY;
         return NULL;
     }
-    file_node_t* new_file_mht_node = &(new_file_mht_node_item->file_node);
+    file_node_t* new_file_mht_node = &new_file_mht_node_item->file_node;
 
     new_file_mht_node->type = FILE_MHT_NODE_TYPE;
     new_file_mht_node->new_node = true;
@@ -624,7 +631,7 @@ static file_node_t* ipf_append_mht_node(pf_context_t* pf, uint64_t mht_node_numb
     new_file_mht_node->physical_node_number = physical_node_number;
 
     if (!lruc_add(pf->cache, new_file_mht_node->physical_node_number, new_file_mht_node)) {
-        ipf_free_node(new_file_mht_node_item);
+        ipf_free_node_item(new_file_mht_node_item);
         pf->last_error = PF_STATUS_NO_MEMORY;
         return NULL;
     }
@@ -675,7 +682,7 @@ static file_node_t* ipf_get_data_node(pf_context_t* pf, uint64_t offset) {
             // before deleting the memory, need to scrub the plain secrets
             file_node_t* file_node = (file_node_t*)data;
             erase_memory(&file_node->decrypted, sizeof(file_node->decrypted));
-            ipf_free_node_from_data(file_node);
+            ipf_free_node_item_from_data(file_node);
         } else {
             if (!ipf_internal_flush(pf)) {
                 // error, can't flush cache, file status changed to error
@@ -695,7 +702,7 @@ static file_node_t* ipf_append_data_node(pf_context_t* pf, uint64_t offset) {
     if (file_mht_node == NULL) // some error happened
         return NULL;
 
-    struct pf_node_item* new_file_data_node_item = ipf_allocate_node();
+    struct pf_node_item* new_file_data_node_item = ipf_allocate_node_item();
     if (!new_file_data_node_item) {
         pf->last_error = PF_STATUS_NO_MEMORY;
         return NULL;
@@ -712,7 +719,7 @@ static file_node_t* ipf_append_data_node(pf_context_t* pf, uint64_t offset) {
     new_file_data_node->physical_node_number = physical_node_number;
 
     if (!lruc_add(pf->cache, new_file_data_node->physical_node_number, new_file_data_node)) {
-        ipf_free_node(new_file_data_node_item);
+        ipf_free_node_item(new_file_data_node_item);
         pf->last_error = PF_STATUS_NO_MEMORY;
         return NULL;
     }
@@ -738,7 +745,7 @@ static file_node_t* ipf_read_data_node(pf_context_t* pf, uint64_t offset) {
     if (file_mht_node == NULL) // some error happened
         return NULL;
 
-    struct pf_node_item* file_data_node_item = ipf_allocate_node();
+    struct pf_node_item* file_data_node_item = ipf_allocate_node_item();
     if (!file_data_node_item) {
         pf->last_error = PF_STATUS_NO_MEMORY;
         return NULL;
@@ -752,7 +759,7 @@ static file_node_t* ipf_read_data_node(pf_context_t* pf, uint64_t offset) {
 
     if (!ipf_read_node(pf, pf->file, file_data_node->physical_node_number,
                        file_data_node->encrypted.cipher, PF_NODE_SIZE)) {
-        ipf_free_node(file_data_node_item);
+        ipf_free_node_item(file_data_node_item);
         return NULL;
     }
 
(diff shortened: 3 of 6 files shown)


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1723)* status: all files reviewed, 26 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQu_eO-1edV2JrrfXD:-NoQu_eO-1edV2JrrfXE:bv8mpez) (no related file):*
Can we add a new section `Optimizations for file I/O` into https://gramine.readthedocs.io/en/stable/performance.html?

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoR0HRz2nyLulzodvqY:-NoR0HRz2nyLulzodvqZ:biyz4jz) (no related file):*
Great PR!

Can I ask for some benchmark results? With different limits (say with `0`, `50`, and whatever is a good limit for your benchmark).

___
___
*[`common/src/protected_files/protected_files.h` line 313 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQvnQK5wG5F9awbueq:-NoQvnQK5wG5F9awbuer:bq7xkbe) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/common/src/protected_files/protected_files.h#L313)):*
> ```objc
>  * \param limit_node_free_list  Limit number of nodes for the PF free list.
>  *
>  * \returns PF status.
> ```

We can add a note here that calling this function is optional -- if it's not called then no free list of file nodes will be created (i.e., this optimization will be disabled).

This will allow to revert changes in `pf_util.c`.
___
*[`common/src/protected_files/protected_files.h` line 315 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQuiyWEBbVVP4WG9ZL:-NoQuiyWEBbVVP4WG9ZM:bb7kry6) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/common/src/protected_files/protected_files.h#L315)):*
> ```objc
>  * \returns PF status.
>  */
> ```

Remove empty line
___
*[`common/src/protected_files/protected_files.c` line 15 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQxqo3BluejAryWo0f:-NoQxqo3BluejAryWo0g:bri532v) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/common/src/protected_files/protected_files.c#L15)):*
> ```C
> 
> #ifdef IN_LIBOS
> #include "libos_thread.h"
> ```

Why is this required?
___
*[`common/src/protected_files/protected_files.c` line 509 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQwyc5AmBLUom4168h:-NoQwyc5AmBLUom4168i:boqoxkx) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/common/src/protected_files/protected_files.c#L509)):*
> ```C
> pf_status_t pf_init_node_free_list(size_t limit_node_free_list) {
>     assert(!g_initialized);
>     struct pf_node_item* item;
> ```

Please move this declaration inside the loop.
___
*[`common/src/protected_files/protected_files.c` line 515 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQxJfZ55mQBLYqd_3q:-NoQxJf_0zWOXU_A-wEo:b-n4evp) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/common/src/protected_files/protected_files.c#L515)):*
> ```C
>         if (item == NULL) {
>             log_error("Not enough memory for the encrypted files node free list. Please consider "
>                       "dicreasing the limit in the manifest.");
> ```

Please add the exact name of the manifest option.
___
*[`common/src/protected_files/protected_files.c` line 527 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoR047x8OlV9_bXxntE:-NoR047x8OlV9_bXxntF:b-95eetn) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/common/src/protected_files/protected_files.c#L527)):*
> ```C
> }
> 
> static struct pf_node_item* ipf_allocate_node(void) {
> ```

Shouldn't it be `ipf_allocate_node_item()` ?
___
*[`common/src/protected_files/protected_files.c` line 535 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQybFND3g2yca69JnI:-NoQybFND3g2yca69JnJ:b-gast0d) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/common/src/protected_files/protected_files.c#L535)):*
> ```C
> 
>         if (FIRST_TIME()) {
>             log_warning("No free file nodes available; using malloc as fallbacks. "
> ```

`fallback`
___
*[`common/src/protected_files/protected_files.c` line 535 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQyfaV9jXHKB2HJiBI:-NoQyfaV9jXHKB2HJiBJ:b-omemiu) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/common/src/protected_files/protected_files.c#L535)):*
> ```C
> 
>         if (FIRST_TIME()) {
>             log_warning("No free file nodes available; using malloc as fallbacks. "
> ```

`available` -> `left`
___
*[`common/src/protected_files/protected_files.c` line 537 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQvdvo3ByA6dx8fDFa:-NoQvdvo3ByA6dx8fDFb:bst7txo) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/common/src/protected_files/protected_files.c#L537)):*
> ```C
>             log_warning("No free file nodes available; using malloc as fallbacks. "
>                         "Please consider adjusting the limit of the encrypted files node free list "
>                         "in the manifest.");
> ```

Please add the name of the manifest option here
___
*[`common/src/protected_files/protected_files.c` line 537 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQyLvt5y7FFr8CRiUc:-NoQyLvt5y7FFr8CRiUd:b2z3glm) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/common/src/protected_files/protected_files.c#L537)):*
> ```C
>             log_warning("No free file nodes available; using malloc as fallbacks. "
>                         "Please consider adjusting the limit of the encrypted files node free list "
>                         "in the manifest.");
> ```

Won't this warning be printed even when manifest doesn't contain this manifest option? It will be weird and unexpected for users. Please add a helper variable that memorizes whether the user requested this list at all.
___
*[`common/src/protected_files/protected_files.c` line 542 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQyqWj6ixvI-4uytwv:-NoQyqWj6ixvI-4uytww:be7wcb5) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/common/src/protected_files/protected_files.c#L542)):*
> ```C
>         item = calloc(1, sizeof(struct pf_node_item));
>         if (item == NULL)
>             return NULL;
> ```

Why not directly `return item;`

And you won't need the `else` branch -- the below code will be guaranteed to execute only if the free-list exists.

(Actually, you can even do `return calloc(...)` for brevity)
___
*[`common/src/protected_files/protected_files.c` line 554 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoR-o7D1AZzInITdvd1:-NoR-o7D1AZzInITdvd2:b1ujjbc) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/common/src/protected_files/protected_files.c#L554)):*
> ```C
> }
> 
> static void ipf_free_node(struct pf_node_item* item) {
> ```

Shouldn't it be `ipf_free_node_item()` ? Otherwise it reads as if you want to free `item->file_node`
___
*[`common/src/protected_files/protected_files.c` line 556 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQzbwWEpPm08cdrKcy:-NoQzbwWEpPm08cdrKcz:b-b9fhzh) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/common/src/protected_files/protected_files.c#L556)):*
> ```C
> static void ipf_free_node(struct pf_node_item* item) {
>     if (!item->from_free_list) {
>         free(item);
> ```

I don't like the `else` cascades where we can avoid them.

Can we rewrite to:
```
    if (!item->from_free_list) {
        free(item);
        return;

    spinlock_lock(&g_pf_node_free_list_lock);
    ...
```
___
*[`common/src/protected_files/protected_files.c` line 567 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoR-FmG3GpG1NlnsdlQ:-NoR-2mi3ZDHnoKSAyfr:b-hawjk0) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/common/src/protected_files/protected_files.c#L567)):*
> ```C
> }
> 
> static void ipf_free_node_from_data(file_node_t* data_ptr) {
> ```

Can we add `assert(data_ptr)`?

I'm scared when I see pointer arithmetic without additional checks. I know nothing can really go wrong, but let's add an assert.
___
*[`common/src/protected_files/protected_files.c` line 567 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoR-xI1-I6tCefD5cvM:-NoR-xI1-I6tCefD5cvN:b-p9k7q0) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/common/src/protected_files/protected_files.c#L567)):*
> ```C
> }
> 
> static void ipf_free_node_from_data(file_node_t* data_ptr) {
> ```

Shouldn't it be `ipf_free_node_item_from_data()` ? Otherwise it reads as if you want to free `item->file_node`
___
*[`common/src/protected_files/protected_files.c` line 618 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoR-bK_1PNIMpVEWYZM:-NoR-bK_1PNIMpVEWYZN:b-60gkze) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/common/src/protected_files/protected_files.c#L618)):*
> ```C
>         return NULL;
>     }
>     file_node_t* new_file_mht_node = &(new_file_mht_node_item->file_node);
> ```

There's no need in `()`. But I don't block as I understand that for some readers, this may be hard to figure out.
___
*[`Documentation/manifest-syntax.rst` line 1024 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQu7f-2IBtirRoF7S7:-NoQu7f-2IBtirRoF7S8:bp3owu8) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/Documentation/manifest-syntax.rst#L1024)):*
> ```reStructuredText
>    such a "doubly-used" key, you must apply a KDF.
> 
> Encrypted files free list node limit
> ```

I would add a word "optimization" somewhere here. 

Maybe `Encrypted files optimization: free list node limit`
___
*[`Documentation/manifest-syntax.rst` line 1025 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQtdNe7sOoFsrgkzm6:-NoQtdNe7sOoFsrgkzm7:b-ccltrz) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/Documentation/manifest-syntax.rst#L1025)):*
> ```reStructuredText
> 
> Encrypted files free list node limit
> ```

No need for an empty line -- I think RST will even incorrectly render in such case.
___
*[`Documentation/manifest-syntax.rst` line 1026 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQtasE9SfEgC_JbSW_:-NoQtasE9SfEgC_JbSWa:b-pps5zl) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/Documentation/manifest-syntax.rst#L1026)):*
> ```reStructuredText
> Encrypted files free list node limit
> 
> ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> ```

Too many `^` symbols (must end right below )
___
*[`Documentation/manifest-syntax.rst` line 1030 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQtYY7DdGA-3_9hqyc:-NoQtYY7DdGA-3_9hqyd:bvzgahj) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/Documentation/manifest-syntax.rst#L1030)):*
> ```reStructuredText
> ::
> 
>     fs.limit.encrypted_files_node_free_list = [NUM]
> ```

`limit` -> `limits`
___
*[`Documentation/manifest-syntax.rst` line 1030 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQuN-W6BnG36w8k022:-NoQuN-X3mdH9gBr9J-z:bogolo8) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/Documentation/manifest-syntax.rst#L1030)):*
> ```reStructuredText
> ::
> 
>     fs.limit.encrypted_files_node_free_list = [NUM]
> ```

Isn't a different order of words more readable? `encrypted_files_free_list_nodes`
___
*[`Documentation/manifest-syntax.rst` line 1034 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQtxja7KTpRuvwu51D:-NoQtxja7KTpRuvwu51E:bekxpks) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/Documentation/manifest-syntax.rst#L1034)):*
> ```reStructuredText
> 
> This syntax specifies a limit on the number of nodes in the pre-allocated
> encrypted files free list, as a performance optimization that trades space for
> ```

`space` -> `memory`?
___
*[`libos/src/fs/libos_fs_encrypted.c` line 288 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQwKVB5sTr7osrut8i:-NoQwKVB5sTr7osrut8j:bcnjres) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/libos/src/fs/libos_fs_encrypted.c#L288)):*
> ```C
>     if (limit_node_free_list_int64 < 0) {
>         log_error("'limit.encrypted_files_node_free_list' = %ld is negative",
>                   limit_node_free_list_int64);
> ```

I don't see a need to output the exact value -- it's obvious what the user should look at and fix.
___
*[`libos/src/fs/libos_fs_encrypted.c` line 294 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQwh7Z93Tzp5G8WUJh:-NoQwh7Z93Tzp5G8WUJi:bsw3cxf) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/libos/src/fs/libos_fs_encrypted.c#L294)):*
> ```C
>     ret = pf_init_node_free_list((size_t)limit_node_free_list_int64);
>     if (ret < 0)
>         return ret;
> ```

I would add a log error message. UPDATE: you have it inside the function itself, unblocking.
___
*[`tools/sgx/common/pf_util.c` line 223 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQw-C8BwbUZIvYHo3s:-NoQw-C8BwbUZIvYHo3t:bw0igjs) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/tools/sgx/common/pf_util.c#L223)):*
> ```C
>  * performance-sensitive so we set the limit of the node free list to 0 which by default disables
>  * the node free list optimization. */
> #define LIMIT_ENCRYPTED_FILES_NODE_FREE_LIST 0
> ```

Why not just revert all these changes? See my other comment.

If you want to leave a hint that we don't care about performance of PF utils, then we can put it in the commit message.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1723)* status: all files reviewed, 25 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoR0HRz2nyLulzodvqY:-NoSBXsVDcdN1UCNuD3x:b-yrm94h) (no related file):*
Bench setup: [RocksDB bench tool](https://github.com/cloudnoize/rocksdb_tool), release build, gramine-direct, estimated 2GB DB size

Limit nodes | 0 (opt disabled) | 128 | 256 | 500 | 1024 | 1280 (no fallback)
-- | -- | -- | -- | -- | -- | --
Multi average duration for 50 puts (µs) | 286.10| 211.22 | 188.72 | 170.85 | 150.88| 150.48

> Note：
The compaction process was not triggered in the above setup. For the bench results on RocksDB compaction, pls refer to https://github.com/gramineproject/gramine/issues/1714#issuecomment-1887367832.

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQu_eO-1edV2JrrfXD:-NoSEGeR06klldVc7VFj:b-896fix) (no related file):*

Can we add a new section `Optimizations for file I/O` into https://gramine.readthedocs.io/en/stable/performance.html?
</blockquote></details>

Done.

___
___
*[`common/src/protected_files/protected_files.h` line 313 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQvnQK5wG5F9awbueq:-NoSB3nL0iZrfS3Cf-EI:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/common/src/protected_files/protected_files.h#L313)):*

We can add a note here that calling this function is optional -- if it's not called then no free list of file nodes will be created (i.e., this optimization will be disabled).

This will allow to revert changes in `pf_util.c`.
</blockquote></details>

Done.
___
*[`common/src/protected_files/protected_files.h` line 315 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQuiyWEBbVVP4WG9ZL:-NoSB4Ie8bVk_n6_fJ5S:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/common/src/protected_files/protected_files.h#L315)):*

Remove empty line
</blockquote></details>

Done.
___
*[`common/src/protected_files/protected_files.c` line 15 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQxqo3BluejAryWo0f:-NoSB56T4f9e9v817_AX:b-33hszl) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/common/src/protected_files/protected_files.c#L15)):*

Why is this required?
</blockquote></details>

Explained in the comments.
___
*[`common/src/protected_files/protected_files.c` line 509 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQwyc5AmBLUom4168h:-NoSB7rD1hDGc_wAQGvC:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/common/src/protected_files/protected_files.c#L509)):*

Please move this declaration inside the loop.
</blockquote></details>

Done.
___
*[`common/src/protected_files/protected_files.c` line 515 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQxJfZ55mQBLYqd_3q:-NoSB8SZ7i3ySkE2UUxL:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/common/src/protected_files/protected_files.c#L515)):*

Please add the exact name of the manifest option.
</blockquote></details>

Done.
___
*[`common/src/protected_files/protected_files.c` line 527 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoR047x8OlV9_bXxntE:-NoSB9_h6uyuKDKy4RRv:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/common/src/protected_files/protected_files.c#L527)):*

Shouldn't it be `ipf_allocate_node_item()` ?
</blockquote></details>

Done.
___
*[`common/src/protected_files/protected_files.c` line 535 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQybFND3g2yca69JnI:-NoSBA4UBcGrsvvlYf5V:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/common/src/protected_files/protected_files.c#L535)):*

`fallback`
</blockquote></details>

Done.
___
*[`common/src/protected_files/protected_files.c` line 535 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQyfaV9jXHKB2HJiBI:-NoSBBVMBWrsrHd5T3PT:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/common/src/protected_files/protected_files.c#L535)):*

`available` -> `left`
</blockquote></details>

Done.
___
*[`common/src/protected_files/protected_files.c` line 537 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQvdvo3ByA6dx8fDFa:-NoSBCsb7u94n-Vh4gLF:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/common/src/protected_files/protected_files.c#L537)):*

Please add the name of the manifest option here
</blockquote></details>

Done.
___
*[`common/src/protected_files/protected_files.c` line 537 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQyLvt5y7FFr8CRiUc:-NoSBDAD35bGhOg_cxo9:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/common/src/protected_files/protected_files.c#L537)):*

Won't this warning be printed even when manifest doesn't contain this manifest option? It will be weird and unexpected for users. Please add a helper variable that memorizes whether the user requested this list at all.
</blockquote></details>

Done.
___
*[`common/src/protected_files/protected_files.c` line 542 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQyqWj6ixvI-4uytwv:-NoSBEBvBuEAfOjn5PC8:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/common/src/protected_files/protected_files.c#L542)):*

Why not directly `return item;`

And you won't need the `else` branch -- the below code will be guaranteed to execute only if the free-list exists.

(Actually, you can even do `return calloc(...)` for brevity)
</blockquote></details>

Done.
___
*[`common/src/protected_files/protected_files.c` line 554 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoR-o7D1AZzInITdvd1:-NoSBEnY1q5_s4hfHkZ6:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/common/src/protected_files/protected_files.c#L554)):*

Shouldn't it be `ipf_free_node_item()` ? Otherwise it reads as if you want to free `item->file_node`
</blockquote></details>

Done.
___
*[`common/src/protected_files/protected_files.c` line 556 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQzbwWEpPm08cdrKcy:-NoSBFsM4-GcMK9jT6c6:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/common/src/protected_files/protected_files.c#L556)):*

I don't like the `else` cascades where we can avoid them.

Can we rewrite to:
```
    if (!item->from_free_list) {
        free(item);
        return;

    spinlock_lock(&g_pf_node_free_list_lock);
    ...
```
</blockquote></details>

Done.
___
*[`common/src/protected_files/protected_files.c` line 567 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoR-FmG3GpG1NlnsdlQ:-NoSBGXWDtr5nRyOCg7O:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/common/src/protected_files/protected_files.c#L567)):*

Can we add `assert(data_ptr)`?

I'm scared when I see pointer arithmetic without additional checks. I know nothing can really go wrong, but let's add an assert.
</blockquote></details>

Done.
___
*[`common/src/protected_files/protected_files.c` line 567 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoR-xI1-I6tCefD5cvM:-NoSBHat2YgAvskqMMGF:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/common/src/protected_files/protected_files.c#L567)):*

Shouldn't it be `ipf_free_node_item_from_data()` ? Otherwise it reads as if you want to free `item->file_node`
</blockquote></details>

Done.
___
*[`Documentation/manifest-syntax.rst` line 1024 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQu7f-2IBtirRoF7S7:-NoSBKZNCzYAr5u9b5vA:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/Documentation/manifest-syntax.rst#L1024)):*

I would add a word "optimization" somewhere here. 

Maybe `Encrypted files optimization: free list node limit`
</blockquote></details>

Done.
___
*[`Documentation/manifest-syntax.rst` line 1025 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQtdNe7sOoFsrgkzm6:-NoSBLVa9hWvGuK2ufuy:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/Documentation/manifest-syntax.rst#L1025)):*

No need for an empty line -- I think RST will even incorrectly render in such case.
</blockquote></details>

Done.
___
*[`Documentation/manifest-syntax.rst` line 1026 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQtasE9SfEgC_JbSW_:-NoSBMKE11Cy5kmfjgcU:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/Documentation/manifest-syntax.rst#L1026)):*

Too many `^` symbols (must end right below )
</blockquote></details>

Done.
___
*[`Documentation/manifest-syntax.rst` line 1030 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQtYY7DdGA-3_9hqyc:-NoSBMfj09iIZV4WqwyV:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/Documentation/manifest-syntax.rst#L1030)):*

`limit` -> `limits`
</blockquote></details>

Done.
___
*[`Documentation/manifest-syntax.rst` line 1030 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQuN-W6BnG36w8k022:-NoSBN0NBsO0J-vMlmEL:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/Documentation/manifest-syntax.rst#L1030)):*

Isn't a different order of words more readable? `encrypted_files_free_list_nodes`
</blockquote></details>

Done.
___
*[`Documentation/manifest-syntax.rst` line 1034 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQtxja7KTpRuvwu51D:-NoSBNWyCpgmT2Ljbjdj:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/Documentation/manifest-syntax.rst#L1034)):*

`space` -> `memory`?
</blockquote></details>

Done.
___
*[`libos/src/fs/libos_fs_encrypted.c` line 288 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQwKVB5sTr7osrut8i:-NoSBP3c0xRg2DX-fzKZ:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/libos/src/fs/libos_fs_encrypted.c#L288)):*

I don't see a need to output the exact value -- it's obvious what the user should look at and fix.
</blockquote></details>

Done.
___
*[`tools/sgx/common/pf_util.c` line 223 at r1](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQw-C8BwbUZIvYHo3s:-NoSBQ5F7rRL3TKA1O_n:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/60ef6327da33866d983faa5abdd782a34dcc34f5/tools/sgx/common/pf_util.c#L223)):*

Why not just revert all these changes? See my other comment.

If you want to leave a hint that we don't care about performance of PF utils, then we can put it in the commit message.
</blockquote></details>

Done.

llly (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1723)* status: all files reviewed, 27 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @kailun-qin)

___
*[`common/src/protected_files/protected_files.c` line 576 at r2](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoV_1tyCcVshxM8bozR:-NoV_1tyCcVshxM8bozS:b-116vej) ([raw file](https://github.com/gramineproject/gramine/blob/ef510ddeba4669484764870ede34882497db6eaf/common/src/protected_files/protected_files.c#L576)):*
> ```C
> 
>     struct pf_node_item* item =
>         (struct pf_node_item*)((char*)data_ptr - offsetof(struct pf_node_item, file_node));
> ```

Can use `container_of` Macro
___
*[`libos/src/fs/libos_fs_encrypted.c` line 277 at r2](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoVaJsT7Gqzs2zBwEtW:-NoVaJsT7Gqzs2zBwEtX:b-mph36l) ([raw file](https://github.com/gramineproject/gramine/blob/ef510ddeba4669484764870ede34882497db6eaf/libos/src/fs/libos_fs_encrypted.c#L277)):*
> ```C
>     int ret;
> 
>     toml_table_t* manifest_fs = toml_table_in(g_manifest_root, "fs");
> ```

Need to make sure `manifest_fs` pointer is not `NULL` before using it.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1723)* status: 4 of 6 files reviewed, 27 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @llly)

___
*[`common/src/protected_files/protected_files.c` line 576 at r2](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoV_1tyCcVshxM8bozR:-NoVp8OT2hb5z6IvpDz7:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/ef510ddeba4669484764870ede34882497db6eaf/common/src/protected_files/protected_files.c#L576)):*

Can use `container_of` Macro
</blockquote></details>

Done.
___
*[`libos/src/fs/libos_fs_encrypted.c` line 277 at r2](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoVaJsT7Gqzs2zBwEtW:-NoVpCYpE-EeEF3rbUr-:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/ef510ddeba4669484764870ede34882497db6eaf/libos/src/fs/libos_fs_encrypted.c#L277)):*

Need to make sure `manifest_fs` pointer is not `NULL` before using it.
</blockquote></details>

Done.

dimakuv (changes requested):
> Reviewed 4 of 6 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1723)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin and @llly)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoR0HRz2nyLulzodvqY:-NoVzzt90dHQe_XQZsnJ:bqy5n7h) (no related file):*

Bench setup: [RocksDB bench tool](https://github.com/cloudnoize/rocksdb_tool), release build, gramine-direct, estimated 2GB DB size

Limit nodes | 0 (opt disabled) | 128 | 256 | 500 | 1024 | 1280 (no fallback)
-- | -- | -- | -- | -- | -- | --
Multi average duration for 50 puts (µs) | 286.10| 211.22 | 188.72 | 170.85 | 150.88| 150.48

> Note：
The compaction process was not triggered in the above setup. For the bench results on RocksDB compaction, pls refer to https://github.com/gramineproject/gramine/issues/1714#issuecomment-1887367832.
</blockquote></details>

The numbers look good to me.

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoQu_eO-1edV2JrrfXD:-NoW2Bfm9buIt799s8TR:bfvngdb) (no related file):*

Done.
</blockquote></details>

Thanks, but see my comments in there.

___
___
*[`-- commits` line 35 at r3](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoW-BuL99FOFTSGfjVf):*
`*/` -- delete this. Can be done during final rebase.
___
*[`common/src/protected_files/protected_files.c` line 518 at r3](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoW-_YNAEXdlqY8TABf:-NoW-_YNAEXdlqY8TABg:brpj0dw) ([raw file](https://github.com/gramineproject/gramine/blob/7759d3c71a4465d55408724ade4f2382752785ac/common/src/protected_files/protected_files.c#L518)):*
> ```C
>         if (item == NULL) {
>             log_error("Not enough memory for the encrypted files node free list. Please consider "
>                       "dicreasing the free list node limit via the manifest option "
> ```

typo: `decreasing`
___
*[`Documentation/performance.rst` line 310 at r3](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoW1IhD4iSkkv3xgiti:-NoW1IhD4iSkkv3xgitj:b1q30wa) ([raw file](https://github.com/gramineproject/gramine/blob/7759d3c71a4465d55408724ade4f2382752785ac/Documentation/performance.rst#L310)):*
> ```reStructuredText
> Optimizations for file I/O
> --------------------------
> ```

Would it be possible to quickly describe what are MHT nodes and data nodes? You can just mention that an encrypted file is split into equally-sized chunks (nodes). The first node is the header metadata node (header MHT node), and other nodes are stored as a Merkle tree, where the leaves are called "data nodes" (and host actual encrypted data) and non-leaves are MHT nodes (and host metadata like intermediate encryption keys).
___
*[`Documentation/performance.rst` line 314 at r3](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoW0lbG8iFGai9UzEBx:-NoW0lbH9FlR1R7AhQTl:b-vefaey) ([raw file](https://github.com/gramineproject/gramine/blob/7759d3c71a4465d55408724ade4f2382752785ac/Documentation/performance.rst#L314)):*
> ```reStructuredText
> one source of overhead for certain workloads (e.g., RocksDB compaction). To
> address this performance bottleneck, instead of using alloc/free directly on
> file nodes, Gramine provides an optional free list of pre-allocacted nodes as an
> ```

`pre-allocated` typo
___
*[`Documentation/performance.rst` line 317 at r3](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoW1t235a0LOe2_9Nnm:-NoW1t235a0LOe2_9Nnn:bucwg81) ([raw file](https://github.com/gramineproject/gramine/blob/7759d3c71a4465d55408724ade4f2382752785ac/Documentation/performance.rst#L317)):*
> ```reStructuredText
> optimization that trades memory for time. This optimization is by default
> disabled, but can be enabled and tuned according to the needs of the application
> via the manifest option ``fs.limits.encrypted_files_node_free_list``.
> ```

Can we add a link to this manifest option?
___
*[`libos/src/fs/libos_fs_encrypted.c` line 286 at r3](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoW1zjkCOH2WJ-qQuEI:-NoW1zjkCOH2WJ-qQuEJ:b-f5ij5l) ([raw file](https://github.com/gramineproject/gramine/blob/7759d3c71a4465d55408724ade4f2382752785ac/libos/src/fs/libos_fs_encrypted.c#L286)):*
> ```C
>                       &limit_nodes_int64);
>     if (ret < 0) {
>         log_error("Cannot parse 'limits.encrypted_files_free_list_nodes'!");
> ```

Seems I forgot previously: you must add the `fs.` prefix too
___
*[`libos/src/fs/libos_fs_encrypted.c` line 290 at r3](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoW22RH7BiK8vLRZyoD:-NoW22RH7BiK8vLRZyoE:b-f5ij5l) ([raw file](https://github.com/gramineproject/gramine/blob/7759d3c71a4465d55408724ade4f2382752785ac/libos/src/fs/libos_fs_encrypted.c#L290)):*
> ```C
>     }
>     if (limit_nodes_int64 < 0) {
>         log_error("Invalid 'limits.encrypted_files_free_list_nodes' value!");
> ```

Seems I forgot previously: you must add the `fs.` prefix too

kailun-qin (commented):
> Reviewed 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1723)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @llly)

___
*[`-- commits` line 35 at r3](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoW-BuL99FOFTSGfjVf):*

`*/` -- delete this. Can be done during final rebase.
</blockquote></details>

oops, done.
___
*[`common/src/protected_files/protected_files.c` line 518 at r3](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoW-_YNAEXdlqY8TABf:-NoW66Ml2jcw0getLjoY:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/7759d3c71a4465d55408724ade4f2382752785ac/common/src/protected_files/protected_files.c#L518)):*

typo: `decreasing`
</blockquote></details>

Done.
___
*[`Documentation/performance.rst` line 310 at r3](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoW1IhD4iSkkv3xgiti:-NoWTyJa8IDUAATYWpzm:beg8ejc) ([raw file](https://github.com/gramineproject/gramine/blob/7759d3c71a4465d55408724ade4f2382752785ac/Documentation/performance.rst#L310)):*

Would it be possible to quickly describe what are MHT nodes and data nodes? You can just mention that an encrypted file is split into equally-sized chunks (nodes). The first node is the header metadata node (header MHT node), and other nodes are stored as a Merkle tree, where the leaves are called "data nodes" (and host actual encrypted data) and non-leaves are MHT nodes (and host metadata like intermediate encryption keys).
</blockquote></details>

Done, w/ some text updates.
___
*[`Documentation/performance.rst` line 314 at r3](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoW0lbG8iFGai9UzEBx:-NoW6VaU4W_FjORixED6:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/7759d3c71a4465d55408724ade4f2382752785ac/Documentation/performance.rst#L314)):*

`pre-allocated` typo
</blockquote></details>

Done.
___
*[`Documentation/performance.rst` line 317 at r3](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoW1t235a0LOe2_9Nnm:-NoWTxLV6WJPXAxDZx_X:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/7759d3c71a4465d55408724ade4f2382752785ac/Documentation/performance.rst#L317)):*

Can we add a link to this manifest option?
</blockquote></details>

Done.
___
*[`libos/src/fs/libos_fs_encrypted.c` line 286 at r3](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoW1zjkCOH2WJ-qQuEI:-NoW6H32EIyJ3XOEKjq-:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/7759d3c71a4465d55408724ade4f2382752785ac/libos/src/fs/libos_fs_encrypted.c#L286)):*

Seems I forgot previously: you must add the `fs.` prefix too
</blockquote></details>

Done.
___
*[`libos/src/fs/libos_fs_encrypted.c` line 290 at r3](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoW22RH7BiK8vLRZyoD:-NoW6HQ0276L73HRcaLf:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/7759d3c71a4465d55408724ade4f2382752785ac/libos/src/fs/libos_fs_encrypted.c#L290)):*

Seems I forgot previously: you must add the `fs.` prefix too
</blockquote></details>

Done.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1723)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @llly)

___
*[`Documentation/performance.rst` line 310 at r3](https://reviewable.io/reviews/gramineproject/gramine/1723#-NoW1IhD4iSkkv3xgiti:-NoWXt_g8BHSAF-Qw1oY:bq8pv8f) ([raw file](https://github.com/gramineproject/gramine/blob/7759d3c71a4465d55408724ade4f2382752785ac/Documentation/performance.rst#L310)):*

Done, w/ some text updates.
</blockquote></details>

Great text, thanks!

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1723)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @llly)

___
*[`-- commits` line 49 at r4](https://reviewable.io/reviews/gramineproject/gramine/1723#-Nok3puK500RdHaQQu35):*
-> `allocated`, have to be fixed during rebase



_Code quote:_
```
allocacted
```

llly (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1723)* status: all files reviewed, 1 unresolved discussion, not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1723)* status: all files reviewed, 2 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1723#-NprDlw106RVTZtF18Up:-NprDlw23pnCfzrp_Mxk:b-yspib5) (no related file):*
This optimization is what all modern memory allocators do, it's just our own allocator which lacks it. So, shouldn't this be implemented as an optimization to our allocator, not just inside encrypted files? It would speed up everything, not only this subsystem.

___

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1723)* status: all files reviewed, 2 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1723#-NprDlw106RVTZtF18Up:-NpsE4PwDjSfEi2SORDv:br4lq0p) (no related file):*
> This optimization is what all modern memory allocators do, it's just our own allocator which lacks it. So, shouldn't this be implemented as an optimization to our allocator, not just inside encrypted files?

I am against this approach.

Ideally, we would not "add an optimization to our allocator", but instead I would vote for **replacing our allocator with a battle-tested allocator**.

In particular, we already had discussions (and I think a general agreement) that we should move to a proper libc (like Musl or maybe newlib), instead of copy-pasting or re-implementing the libc APIs. I would prefer this to "adding optimizations to our code".

Moreover, we already have such micro-optimizations sprinkled all over our Gramine code. Two optimizations that come to mind:
1. Using Linux-compatible structs for iovec objects for send/recv on the sockets. In the original networking sub-system by Borys, these LibOS objects were not-Linux-compatible, so the compatibility layer in PAL malloced and freed temporary helper objects. This resulted in huge perf overhead, so we just made iovec objects the same as in Linux, to side-steps memory allocation in PAL.
2. Using a scratch buffer for the first thread during `sendfile()`, again to not allocate and free every time. This was particularly beneficial for Nginx.

I'm sure we have many such micro-optimizations in different parts of Gramine. They are all quite small and self-contained.

I would prefer this approach:
1. Continue with micro-optimizations of different sub-systems. (This PR falls under this category.)
2. In some future, choose, design and implement the replacement for the memory allocator by a default allocator of some libc version. (Far away in future.)

I really don't like the rabbit hole of "let's fix/replace our current allocator", as this will take months/years to accomplish and test.

___

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1723)* status: all files reviewed, 2 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mwkmwkmwk)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1723#-NprDlw106RVTZtF18Up:-NpwF4SK90Ohh5HBZc9L:bekw03i) (no related file):*
> Ideally, we would not "add an optimization to our allocator", but instead I would vote for **replacing our allocator with a battle-tested allocator**.

I'm in favor of this, but do you mean that you want to do this instead of this PR?

> Moreover, we already have such micro-optimizations sprinkled all over our Gramine code. Two optimizations that come to mind:

Yup, that's a very good point. I hate these micro-optimizations for very specific parts of the codebase, we could just use a proper allocator and all of them would be good (or at least acceptable).

> I really don't like the rabbit hole of "let's fix/replace our current allocator", as this will take months/years to accomplish and test.

Hmm, but isn't it mostly the same algorithm as implemented in this PR? You only need to add buckets per size, but otherwise it would be the same?

btw, @mwkmwkmwk: How does allocation support looks like in nostd Rust? Can migrating a part of the codebase allow us to use some Rust allocator (but not sure how to get one in nostd) from C code and to remove our custom allocators?

___

mwkmwkmwk (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1723)* status: all files reviewed, 2 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1723#-NprDlw106RVTZtF18Up:-NpwOI5y3WIj4M6_FHKL:biuwdun) (no related file):*

> Ideally, we would not "add an optimization to our allocator", but instead I would vote for **replacing our allocator with a battle-tested allocator**.

I'm in favor of this, but do you mean that you want to do this instead of this PR?

> Moreover, we already have such micro-optimizations sprinkled all over our Gramine code. Two optimizations that come to mind:

Yup, that's a very good point. I hate these micro-optimizations for very specific parts of the codebase, we could just use a proper allocator and all of them would be good (or at least acceptable).

> I really don't like the rabbit hole of "let's fix/replace our current allocator", as this will take months/years to accomplish and test.

Hmm, but isn't it mostly the same algorithm as implemented in this PR? You only need to add buckets per size, but otherwise it would be the same?

btw, @mwkmwkmwk: How does allocation support looks like in nostd Rust? Can migrating a part of the codebase allow us to use some Rust allocator (but not sure how to get one in nostd) from C code and to remove our custom allocators?
</blockquote></details>

We could replace our custom allocator with an existing allocator, and it doesn't really matter much whether it is written in Rust or in C. Allocator interface is simple enough that it's just a bit of FFI plumbing.

From a short study of available crates, general Rust allocators tend to come in three variants:

- wrappers of heavyweight mmap-based allocators written in C, like jemalloc or mimalloc — not particularly interesting, as it just passes the buck
- simple allocators for embedded bare-metal Rust — these tend to come with an assumption of using a single static-sized memory area as a heap, which again makes them unsuitable
- simple allocators intended for WebAssembly, like `wee_alloc` — while the use case doesn't entirely match, this seems to be the closest to our needs; in particular, if given a constraint of "port gramine to some existing Rust allocator", I'd look into porting `wee_alloc`

Once we have something that implements the global allocator API, there's of course existing crates that we can use for optimizations on top of that, like `slab`.

___

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1723)* status: all files reviewed, 2 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1723#-NprDlw106RVTZtF18Up:-NpxVibC2MAodHNenlAD:b-8jnpb9) (no related file):*
> I'm in favor of this, but do you mean that you want to do this instead of this PR?

@mkow No, I want to merge this PR. My reasoning:
1. In the short term, we should continue with these micro-optimizations. Hence we should merge this PR.
2. In the long term, we should switch to some battle-tested allocator.

I strongly believe that doing item 2 will take us ~6 months. Our internal slab allocator is not for the faint of heart, and replacing it won't be a trivial task. Then we'll also need to measure overall performance of Gramine on different benchmarks, find all corner cases, fine-tune the allocator, choose the default parameters, and remove all micro-optimizations. That's a LOT of work.

> Hmm, but isn't it mostly the same algorithm as implemented in this PR? You only need to add buckets per size, but otherwise it would be the same?

What do you mean by buckets per size? We have a slab allocator that has 8 levels (up to 2KB sizes), and then the fall back to host-backed allocation for objects >2KB sized. Now, for encrypted files we have 4KB chunks -- so do we create new bucket size (new level) for them?

> simple allocators intended for WebAssembly, like `wee_alloc` — while the use case doesn't entirely match, this seems to be the closest to our needs; in particular, if given a constraint of "port gramine to some existing Rust allocator", I'd look into porting `wee_alloc`

I wouldn't mind this, sounds like a decent fit. Is this allocator no-stdlib? I vaguely remember that Rust didn't have a notion of nostdlib code at some point in time, not sure if this is fixed already or maybe it's not relevant to `wee_alloc` at all.

Also, wouldn't we want to consider a C-based simple allocator, like mimalloc? Two arguments:
1. Introducing a Rust compiler in the toolchain requires some effort from our side (probably not big enough, but we e.g. support CentOS 8 distros, and we guarantee manual builds of Gramine on such systems -- does Rust work on CentOS 8?).
2. A well-known and simple memory allocator shouldn't be susceptible to memory safety bugs, at least I don't remember any vulnerabilities concerning memory allocators. So why bother with a Rust-based one, if a C-based one is as good.

___

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1723)* status: all files reviewed, 2 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1723#-NprDlw106RVTZtF18Up:-Npxc3Q-8v9CIaiWVe4K:b-m9wzkq) (no related file):*
> -   In the short term, we should continue with these micro-optimizations. Hence we should merge this PR.
> -   In the long term, we should switch to some battle-tested allocator.

+1

> I strongly believe that doing item 2 will take us ~6 months. Our internal slab allocator is not for the faint of heart, and replacing it won't be a trivial task. Then we'll also need to measure overall performance of Gramine on different benchmarks, find all corner cases, fine-tune the allocator, choose the default parameters, and remove all micro-optimizations. That's a LOT of work.

Yeah, I think we actually did some similar PoCs before (about 1 year ago?) by replacing our internal slab allocator w/ `dlmalloc` and `mimalloc` respectively (for memchached perf optimization IIRC). But all these efforts were not moved fortward any further.

> wouldn't we want to consider a C-based simple allocator, like mimalloc?

+1, any special reason that we consdier a Rust allocator in this case?

___

mwkmwkmwk (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1723)* status: all files reviewed, 2 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1723#-NprDlw106RVTZtF18Up:-NpymnAK1SOgdPGz74Fc:bq6vqhy) (no related file):*
> Also, wouldn't we want to consider a C-based simple allocator, like mimalloc? Two arguments:
> 
> 1.  Introducing a Rust compiler in the toolchain requires some effort from our side (probably not big enough, but we e.g. support CentOS 8 distros, and we guarantee manual builds of Gramine on such systems -- does Rust work on CentOS 8?).

the reason this came up is because we were talking with @mkow about gradually porting gramine to Rust for other reasons, so we wouldn't be doing this just for the allocator

___

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1723)* status: all files reviewed, 2 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1723#-NprDlw106RVTZtF18Up:-NpzeOYv0L6xVHPL8KtR:bq08i0u) (no related file):*
> Hmm, but isn't it mostly the same algorithm as implemented in this PR? You only need to add buckets per size, but otherwise it would be the same?
> Now, for encrypted files we have 4KB chunks -- so do we create new bucket size (new level) for them?

The size of the file node in encrypted files is actually 8235B (including both the encrypted data and the plaintext data) rather than 4KB: https://github.com/gramineproject/gramine/blob/1cf1f46646646a3b9c6b371e67c80e945577456a/common/src/protected_files/protected_files_format.h#L115-L130

So adding a new level/slab size that could benefit encrypted files would be too random/customized, pls see a reference change here: https://github.com/gramineproject/gramine/pull/1763. We're asking internal team to test on this one for its effectiveness as well -- for completeness.

___

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1723)* status: all files reviewed, 2 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1723#-NprDlw106RVTZtF18Up:-Nq1bI_sERydpRIPj33m:bxjxo68) (no related file):*
> the reason this came up is because we were talking with [@mkow](https://github.com/mkow) about gradually porting gramine to Rust for other reasons, so we wouldn't be doing this just for the allocator

@mwkmwkmwk  @mkow I am also in favor of exploring Rust applicability to Gramine codebase, but I think there are two good starting-point candidates:
- the Linux PAL, as it is self-contained and rather simple, so a good playing ground
- The checkpoint/restore code in LibOS -- the current code makes me cry, so it needs to be rewritten anyway, and this feels like a good opportunity to re-write a bad subsystem with a new design and in a new language (also it is currently hard-core low-level pointer juggling, and I would like to see how Rust deals with such stuff; my impression that it will be a sad `unsafe` case but hopefully I'm mistaken)

> The size of the file node in encrypted files is actually 8235B (including both the encrypted data and the plaintext data) rather than 4KB:

@kailun-qin makes a good point -- the encrypted file node size is kinda random, and it doesn't fit the current slab allocator design.

___

Issue #1336, opened 2023-05-09
Fail on unknown options in Gramine manifest


dimakuv:
### Description of the feature

Currently Gramine silently ignores unknown-to-it manifest options. We must check the whole manifest file for unknown TOML keys and explicitly fail on them.

@monavij and @mkow agreed with this during the Gramine Tuesday meeting.

### Why Gramine should implement it?

To avoid problems like this: https://github.com/gramineproject/gramine/issues/1253#issuecomment-1538503201

PR #1726, opened 2024-01-19
`gramine-manifest-check` tool

woju:
## Description of the changes 

Linter for gramine manifest. Parser in the actual application is permissive, silently ignores unknown keys in manifest. This optional tool will warn about unknown options (probably mistyped real things!).

Fixes: #1336

## How to test this PR? 

Go get any manifest (and/or manifest.sgx) and feed it to this tool. If there's no output and 0 exit code, then manifest should be OK. If the manifest is not OK, exit code should be != 0 and some explanation should be written to stdout.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1726)

### DIFF
--- a/.ci/ubuntu20.04.dockerfile
+++ b/.ci/ubuntu20.04.dockerfile
@@ -62,6 +62,7 @@ RUN apt-get update && env DEBIAN_FRONTEND=noninteractive apt-get install -y \
     python3-pytest-xdist \
     python3-scipy \
     python3-sphinx-rtd-theme \
+    python3-voluptuous \
     shellcheck \
     sphinx-doc \
     sqlite3 \
--- a/.ci/ubuntu22.04.dockerfile
+++ b/.ci/ubuntu22.04.dockerfile
@@ -63,6 +63,7 @@ RUN apt-get update && env DEBIAN_FRONTEND=noninteractive apt-get install -y \
     python3-pytest-xdist \
     python3-scipy \
     python3-sphinx-rtd-theme \
+    python3-voluptuous \
     shellcheck \
     sphinx-doc \
     sqlite3 \
--- a/CI-Examples/helloworld/Makefile
+++ b/CI-Examples/helloworld/Makefile
@@ -25,6 +25,7 @@ helloworld.manifest: helloworld.manifest.template
 	gramine-manifest \
 		-Dlog_level=$(GRAMINE_LOG_LEVEL) \
 		$< $@
+	gramine-manifest-check $@
 
 # gramine-sgx-sign generates both a .sig file and a .manifest.sgx file. This is somewhat
 # hard to express properly in Make. The simple solution would be to use
@@ -47,6 +48,7 @@ sgx_sign: helloworld.manifest helloworld
 	gramine-sgx-sign \
 		--manifest $< \
 		--output $<.sgx
+	gramine-manifest-check $<.sgx
 
 ifeq ($(SGX),)
 GRAMINE = gramine-direct
--- a/Documentation/conf.py
+++ b/Documentation/conf.py
@@ -147,6 +147,7 @@ intersphinx_mapping = {
     'python': ('https://docs.python.org/3', None),
     'click': ('https://click.palletsprojects.com/en/latest', None),
     'cryptography': ('https://cryptography.io/en/latest', None),
+    'voluptuous': ('https://alecthomas.github.io/voluptuous/docs/_build/html', None),
 }
 
 # -- Options for HTML output -------------------------------------------------
@@ -185,6 +186,7 @@ man_pages = [
     ('manpages/gramine', 'gramine-sgx', 'Gramine', [author], 1),
     ('manpages/gramine-argv-serializer', 'gramine-argv-serializer', 'Serialize command line arguments', [author], 1),
     ('manpages/gramine-manifest', 'gramine-manifest', 'Gramine manifest preprocessor', [author], 1),
+    ('manpages/gramine-manifest-check', 'gramine-manifest-check', 'Gramine manifest schema validator', [author], 1),
     ('manpages/gramine-ratls', 'gramine-ratls', 'RA-TLS wrapper', [author], 1),
     ('manpages/gramine-sgx-gen-private-key', 'gramine-sgx-gen-private-key', 'Gramine SGX key generator', [author], 1),
     ('manpages/gramine-sgx-get-token', 'gramine-sgx-get-token', 'Gramine SGX Token generator', [author], 1),
--- a/Documentation/manpages/gramine-manifest-check.rst
+++ b/Documentation/manpages/gramine-manifest-check.rst
@@ -1,0 +1,28 @@
+.. program:: gramine-manifest-check
+.. _gramine-manifest-check:
+
+======================================================================
+:program:`gramine-manifest-check` -- Gramine manifest schema validator
+======================================================================
+
+Synopsis
+========
+
+:command:`gramine-manifest-check` [*MANIFEST-FILE*]
+
+Description
+===========
+
+The program :program:`gramine-manifest-check` is used to check manifests for
+compliance with builtin manifest schema.
+
+If the manifest contains entries that are not parsed by Gramine itself (possibly
+misspelled real options) or does not contain mandatory options,
+:program:`gramine-manifest-check` exits non-zero and short diagnostics
+describing the path into data structure will be output to standard error.
+If the manifest is OK, nothing is printed and the tool exits with return code 0.
+
+Note that options that are allowed and/or mandatory for default LibOS
+implementation (``libsysdb.so``) are considered allowed/mandatory in schema.
+Therefore, if you have another LibOS implementation (like it happens in PAL
+test suite), the check may be wrong.
--- a/Documentation/manpages/gramine-manifest-check.rst
+++ b/Documentation/manpages/gramine-manifest-check.rst
@@ -26,3 +26,7 @@ Note that options that are allowed and/or mandatory for default LibOS
 implementation (``libsysdb.so``) are considered allowed/mandatory in schema.
 Therefore, if you have another LibOS implementation (like it happens in PAL
 test suite), the check may be wrong.
+
+By default the check is already performed in :program:`gramine-manifest` (see
+:option:`gramine-manifest --check`). This standalone tool may be useful for
+example to validate existing manifests when updating Gramine version.
--- a/Documentation/manpages/gramine-manifest.rst
+++ b/Documentation/manpages/gramine-manifest.rst
@@ -23,6 +23,23 @@ Command line arguments
 
    Have a |~| variable available in the template.
 
+.. option:: --check
+
+   After rendering manifest from template, perform validation against manifest
+   schema to check for unknown manifest entries and/or missing mandatory
+   options. See :doc:`gramine-manifest-check` for more details.
+
+   The check is enabled by default. This option serves to re-enable the check
+   after :option:`--no-check`.
+
+   For the 1.7 release, only a |~| warning is issued and
+   :program:`gramine-manifest` proceeds to write the faulty manifest. In version
+   1.8 this will be a |~| hard error.
+
+.. option:: --no-check
+
+   Disable schema validation, as described above in :option:`--check`.
+
 Functions and constants available in templates
 ==============================================
 
--- a/Documentation/requirements.txt
+++ b/Documentation/requirements.txt
@@ -19,3 +19,6 @@ tomli==2.0.*
 
 # https://tracker.debian.org/pkg/python-tomli-w
 tomli-w==1.0.*
+
+# https://tracker.debian.org/pkg/voluptuous
+voluptuous==0.12.*
--- a/debian/control
+++ b/debian/control
@@ -21,6 +21,7 @@ Build-Depends: debhelper-compat (= 13),
  python3-sphinx-rtd-theme,
  python3-tomli (>= 1.1.0),
  python3-tomli-w (>= 0.4.0),
+ python3-voluptuous,
 #libunwind8,
 #python3-pytest,
 Standards-Version: 4.1.3
@@ -43,6 +44,7 @@ Depends:
  python3-pyelftools,
  python3-tomli (>= 1.1.0),
  python3-tomli-w (>= 0.4.0),
+ python3-voluptuous,
 Recommends:
  gramine-ratls-dcap,
  gramine-ratls-epid,
--- a/gramine.spec
+++ b/gramine.spec
@@ -43,6 +43,7 @@ Requires: python3-protobuf
 Requires: python3-pyelftools
 Requires: python3-tomli >= 1.1.0
 Requires: python3-tomli-w >= 0.4.0
+Requires: python3-voluptuous
 
 %global debug_package %{nil}
 %global __meson_auto_features disabled
--- a/libos/test/regression/fork_disallowed.manifest.template
+++ b/libos/test/regression/fork_disallowed.manifest.template
@@ -14,7 +14,6 @@ fs.mounts = [
 # must print a warning: "The app tried to create a subprocess, but this is disabled"
 sys.disallow_subprocesses = true
 
-sgx.nonpie_binary = true
 sgx.debug = true
 sgx.edmm_enable = {{ 'true' if env.get('EDMM', '0') == '1' else 'false' }}
 
--- a/packaging/alpine/APKBUILD
+++ b/packaging/alpine/APKBUILD
@@ -41,6 +41,7 @@ depends="
     py3-jinja2
     py3-tomli
     py3-tomli-w
+    py3-voluptuous
     "
 source="$pkgname-$pkgver.tar.gz"
 builddir="$srcdir/$pkgname-$_real_pkgver/"
--- a/pal/regression/tests.toml
+++ b/pal/regression/tests.toml
@@ -1,5 +1,11 @@
 binary_dir = "@GRAMINE_PKGLIBDIR@/tests/pal"
 
+# Disable checking manifest for conformance with official schema (`gramine-manifest --no-check`).
+# Official schema includes mandatory manifest entries for official libos (libsysdb.so), but what we
+# do here is, we run libos replacements and manifests for those don't include entries that are
+# normally required by libsysdb. Therefore we need to disable schema checking.
+gramine-manifest-no-check = true
+
 manifests = [
   "..Bootstrap",
   "avl_tree_test",
--- a/python/gramine-manifest
+++ b/python/gramine-manifest
@@ -4,6 +4,7 @@
 #                    Borys Popławski <borysp@invisiblethingslab.com>
 
 import click
+import voluptuous
 
 from graminelibos import Manifest
 
@@ -20,13 +21,25 @@ def validate_define(_ctx, _param, values):
 @click.command()
 @click.option('--string', '-c')
 @click.option('--define', '-D', multiple=True, callback=validate_define)
+@click.option('--check/--no-check', default=True,
+    help='check the manifest for correctness against builtin schema')
 @click.argument('infile', type=click.File('r'), required=False)
 @click.argument('outfile', type=click.File('wb'), default='-')
-def main(string, define, infile, outfile):
+@click.pass_context
+def main(ctx, string, define, infile, outfile, check):
     if not bool(string) ^ bool(infile):
         click.get_current_context().fail('specify exactly one of (infile, -c)')
     template = infile.read() if infile else string
     manifest = Manifest.from_template(template, define)
+
+    if check:
+        try:
+            manifest.check()
+        except voluptuous.MultipleInvalid as err:
+            click.echo(f'WARNING: error in manifest (after rendering): {err!s}', err=True)
+            # TODO after 1.7: make this a hard error
+            #ctx.exit(1)
+
     manifest.dump(outfile)
 
 if __name__ == '__main__':
--- a/python/gramine-manifest-check
+++ b/python/gramine-manifest-check
@@ -1,0 +1,34 @@
+#!/usr/bin/python3
+# SPDX-License-Identifier: LGPL-3.0-or-later
+# Copyright (C) 2024 Intel Corporation
+#                    Wojtek Porczyk <woju@invisiblethingslab.com>
+
+import sys
+
+import click
+import voluptuous
+
+try:
+    import tomllib
+except ImportError:
+    import tomli as tomllib
+
+from graminelibos.manifest_check import GramineManifestSchema
+
+@click.command()
+@click.argument('file', type=click.File('rb'), default='-')
+@click.pass_context
+def main(ctx, file):
+    with file:
+        try:
+            data = tomllib.load(file)
+        except tomllib.TOMLDecodeError as err:
+            ctx.fail(f'error parsing manifest: {err!s}')
+    try:
+        GramineManifestSchema(data)
+    except voluptuous.MultipleInvalid as err:
+        print(f'error in manifest: {err!s}')
+        ctx.exit(1)
+
+if __name__ == '__main__':
+    main() # pylint: disable=no-value-for-parameter
--- a/python/graminelibos/manifest.py
+++ b/python/graminelibos/manifest.py
@@ -20,6 +20,7 @@ import tomli
 import tomli_w
 
 from . import _env
+from .manifest_check import GramineManifestSchema
 
 DEFAULT_ENCLAVE_SIZE_NO_EDMM = '256M'
 DEFAULT_ENCLAVE_SIZE_WITH_EDMM = '1024G'  # 1TB; note that DebugInfo is at 1TB and ASan at 1.5TB
@@ -382,6 +383,15 @@ class Manifest:
 
     def dump(self, f):
         tomli_w.dump(self._manifest, f)
+
+    def check(self):
+        """Check the manifest against builtin schema
+
+        Raises:
+            voluptuous.error.MultipleInvalid: when check fails
+        """
+
+        return GramineManifestSchema(self._manifest)
 
     def expand_all_trusted_files(self, chroot=None):
         """Expand all trusted files entries.
--- a/python/graminelibos/manifest_check.py
+++ b/python/graminelibos/manifest_check.py
@@ -1,0 +1,137 @@
+# SPDX-License-Identifier: LGPL-3.0-or-later
+# Copyright (C) 2024 Intel Corporation
+#                    Wojtek Porczyk <woju@invisiblethingslab.com>
+
+from voluptuous import (
+    Any,
+    Required,
+    Schema,
+)
+
+# size (number + suffix)
+# TODO: write a better validator
+_size = str
+
+# masks for sgx.seal_key.*_mask fields
+# TODO: write a better validator
+_mask64 = str
+_mask32 = str
+
+# TODO: write a better validator
+_uri = str
+
+# fs.root and fs.mounts[] are almost the same, but fs.root does not contain path= key
+_fs_base = (
+    {
+        'type': 'chroot',
+        Required('uri'): _uri,
+    },
+    {
+        Required('type'): 'encrypted',
+        Required('uri'): _uri,
+        'key_name': str,
+    },
+    {
+        Required('type'): 'tmpfs',
+        'uri': str, # not _uri, this field is ignored by Gramine
+    },
+    {
+        Required('type'): 'untrusted_shm',
+        Required('uri'): _uri,
+    },
+)
+
+_fs_root = Any(*_fs_base)
+_fs_mount = Any(*({**d, Required('path'): str} for d in _fs_base))
+
+GramineManifestSchema = Schema({
+    Required('fs'): {
+        Required('mounts'): [_fs_mount],
+        'root': _fs_root,
+        'start_dir': str,
+        'insecure__keys': {str: str},
+    },
+
+    Required('libos'): {
+        Required('entrypoint'): str,
+        'check_invalid_pointers': bool,
+    },
+
+    Required('loader'): {
+        Required('entrypoint'): str,
+        'argv': [str],
+        'argv_src_file': str,
+        'env': {str: Any(str, {'value': str}, {'passthrough': True})},
+        'env_src_file': str,
+        'gid': int,
+        'insecure__use_cmdline_argv': bool,
+        'insecure__use_host_env': bool,
+        'insecure__disable_aslr': bool,
+        'log_file': str,
+        'log_level': Any('none', 'error', 'warning', 'debug', 'trace', 'all'),
+        'uid': int,
+    },
+
+    'sgx': {
+        'allowed_files': [str],
+        'cpu_features': {
+            Any('avx', 'avx512', 'amx'): Any('unspecified', 'disabled', 'required'),
+            Any('mpx', 'pkru'): Any('disabled', 'required'),
+        },
+        'debug': bool,
+        'edmm_enable': bool,
+        'enable_stats': bool,
+        'enclave_size': _size,
+        'file_check_policy': Any('strict', 'allow_all_but_log'),
+        'insecure__allow_memfaults_without_exinfo': bool,
+        'insecure__rpc_thread_num': int,
+        'isvprodid': int,
+        'isvsvn': int,
+        'max_threads': int,
+        'preheat_enclave': bool,
+        'profile': {
+            'enable': Any('none', 'main', 'all'),
+            'mode': Any('aex', 'ocall_inner', 'ocall_outer'),
+            'with_stack': bool,
+            'frequency': int,
+        },
+        'ra_client_linkable': bool,
+        'ra_client_spid': str,
+        'remote_attestation': Any('none', 'dcap', 'epid'),
+        'require_amx': bool,        # deprecated
+        'require_avx': bool,        # deprecated
+        'require_avx512': bool,     # deprecated
+        'require_exinfo': bool,     # deprecated
+        'require_mpx': bool,        # deprecated
+        'require_pkru': bool,       # deprecated
+        'seal_key': {
+            'flags_mask': _mask64,
+            'xfrm_mask': _mask64,
+            'misc_mask': _mask32,
+        },
+        # TODO: validator for sha256
+        'trusted_files': [Any(str, {'uri': _uri, 'sha256': str})],
+        'use_exinfo': bool,
+        'vtune_profile': bool,
+    },
+
+    'sys': {
+        'allowed_ioctls': [{
+            Required('request_code'): int,
+            'struct': str,
+        }],
+        'brk': {'max_size': _size},
+        'disallow_subprocesses': bool,
+        'enable_extra_runtime_domain_names_conf': bool,
+        'enable_sigterm_injection': bool,
+        'experimental__enable_flock': bool,
+        'insecure__allow_eventfd': bool,
+
+        # Description of this thing will be both very hard to write, and mostly useless, since
+        # majority of errors will be semantic (wrong offsets), not syntactic. We'll leave it almost
+        # not validated.
+        'ioctl_structs': {str: object},
+
+        'stack': {'size': _size},
+    },
+})
--- a/python/graminelibos/meson.build
+++ b/python/graminelibos/meson.build
@@ -15,6 +15,7 @@ python_src = [
     init_py,
     'gen_jinja_env.py',
     'manifest.py',
+    'manifest_check.py',
 ]
 
 if enable_tests
--- a/python/graminelibos/util_tests.py
+++ b/python/graminelibos/util_tests.py
@@ -67,6 +67,8 @@ class TestConfig:
 
         self.libc = data.get('libc', 'glibc')
 
+        self.no_check = data.get('gramine-manifest-no-check', False)
+
         self.arch_libdir = _CONFIG_SYSLIBDIR
 
         # Used by LTP, for `libstdbuf.so`
@@ -126,6 +128,7 @@ class TestConfig:
                      '-Dentrypoint=$ENTRYPOINT '
                      '-Dbinary_dir=$BINARY_DIR '
                      '-Dlibc=$GRAMINE_LIBC '
+                    f'{"--no-check " if self.no_check else ""}'
                      '$in $out'),
             description='manifest: $out'
         )
--- a/python/meson.build
+++ b/python/meson.build
@@ -4,6 +4,7 @@ subdir('graminelibos.dist-info')
 install_data([
     'gramine-gen-depend',
     'gramine-manifest',
+    'gramine-manifest-check',
 ], install_dir: get_option('bindir'))
 
 if enable_tests


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1726)* status: all files reviewed, 14 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @woju)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok7ZsBEn60EnIBrkpz:-Nok7ZsBEn60EnIBrkq-:bucfyf7) (no related file):*
CI fails with:
```
make: gramine-manifest-check: Command not found
```

Aren't you supposed to change some `meson.build` to install this tool?

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok7czOC6QrSD-irSwq:-Nok7czOC6QrSD-irSwr:bthw85p) (no related file):*
Can we add the call to this tool into `gramine-manifest`? It should be a new argument, and it should be disabled by default.

And we enable it for all our CI examples and for all our PAL/LibOS tests. This will we have the testing for both the new tool and for the manifest schema.

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok8BhA9KAmtczhaTa6:-Nok8BhA9KAmtczhaTa7:bclxxh1) (no related file):*
What about non-Debian distros? I guess you need to modify them too?

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok8F6r3gJOixNmUnRX:-Nok8F6r3gJOixNmUnRY:b-rp930v) (no related file):*
Brilliant PR!

___
___
*[`CI-Examples/helloworld/Makefile` line 28 at r1](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok7zwKE8K8sfeFanu2:-Nok7zwKE8K8sfeFanu3:b-5di2h3) ([raw file](https://github.com/gramineproject/gramine/blob/cfe1527d49aad8147a7e8024a3339aafd7e6ebf7/CI-Examples/helloworld/Makefile#L28)):*
> ```Makefile
> 		-Dlog_level=$(GRAMINE_LOG_LEVEL) \
> 		$< $@
> 	gramine-manifest-check $@
> ```

I suggest to incorporate the check into `gramine-manifest`
___
*[`CI-Examples/helloworld/Makefile` line 51 at r1](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok82Xf2YJs33bMklQ1:-Nok82Xf2YJs33bMklQ2:bal34av) ([raw file](https://github.com/gramineproject/gramine/blob/cfe1527d49aad8147a7e8024a3339aafd7e6ebf7/CI-Examples/helloworld/Makefile#L51)):*
> ```Makefile
> 		--manifest $< \
> 		--output $<.sgx
> 	gramine-manifest-check $<.sgx
> ```

Here I'm not sure, I would just remove this change.

We could check final `.sgx` manifests' schemas in LibOS regression tests, that should be enough coverage.
___
*[`python/graminelibos/manifest_check.py` line 23 at r1](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok2ZwU4gI3zKu3pEWd:-Nok2ZwU4gI3zKu3pEWe:b-wrp0wx) ([raw file](https://github.com/gramineproject/gramine/blob/cfe1527d49aad8147a7e8024a3339aafd7e6ebf7/python/graminelibos/manifest_check.py#L23)):*
> ```Python
>             },
>             {
>                 'type': Any('chroot', 'tmpfs', 'untrusted_shm'),
> ```

`tmpfs` doesn't have the `uri` (more correctly it's not required and always ignored), so it should be a separate item in the outer Any set
___
*[`python/graminelibos/manifest_check.py` line 30 at r1](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok3RKs6ZJeN_lq4TKw:-Nok3RKs6ZJeN_lq4TKx:b-rf3qj8) ([raw file](https://github.com/gramineproject/gramine/blob/cfe1527d49aad8147a7e8024a3339aafd7e6ebf7/python/graminelibos/manifest_check.py#L30)):*
> ```Python
>         'root': {
>             'type': Any('chroot', 'encrypted', 'tmpfs', 'untrusted_shm'),
>             'uri': str,
> ```

There is also `fs.root.key_name`: https://github.com/gramineproject/gramine/blob/a933017d35a7aa3330194521389838de6be53dc4/libos/src/fs/libos_fs.c#L128

Actually, `root` should be exactly like an item in the `fs.mounts` set. Is this possible to do in Voluptuous (declare the `fs.mounts` item description separately and use it in two places)?
___
*[`python/graminelibos/manifest_check.py` line 33 at r1](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok472p8K-mGSZki-x6:-Nok472p8K-mGSZki-x7:bpqarst) ([raw file](https://github.com/gramineproject/gramine/blob/cfe1527d49aad8147a7e8024a3339aafd7e6ebf7/python/graminelibos/manifest_check.py#L33)):*
> ```Python
>         },
>         'start_dir': str,
>         'insecure__keys': {str, str},
> ```

Is this a tuple syntax (array)? It should be a dict syntax:
```
fs.insecure__keys.[KEY_NAME] = "[32-character hex value]"
``` 
___
*[`python/graminelibos/manifest_check.py` line 42 at r1](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok4YpiB0eFGNon3Wky:-Nok4YpiB0eFGNon3Wkz:b1d4qj0) ([raw file](https://github.com/gramineproject/gramine/blob/cfe1527d49aad8147a7e8024a3339aafd7e6ebf7/python/graminelibos/manifest_check.py#L42)):*
> ```Python
>         Required('entrypoint'): str,
>         'argv': [str],
>         'argv_src_file': str,
> ```

Any way to specify a XOR in Voluptuous?

Here, `argv` and `argv_src_file` are mutually exclusive.
___
*[`python/graminelibos/manifest_check.py` line 63 at r1](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok5WEk5h3YDxNyRW5A:-Nok5WEk5h3YDxNyRW5B:ba6z8my) ([raw file](https://github.com/gramineproject/gramine/blob/cfe1527d49aad8147a7e8024a3339aafd7e6ebf7/python/graminelibos/manifest_check.py#L63)):*
> ```Python
>         'enable_stats': bool,
>         'enclave_size': _size,
>         'file_check_policy': Any('strict', 'allow_all_byt_log'),
> ```

Typo: `allow_all_but_log` (not `byt`)
___
*[`python/graminelibos/manifest_check.py` line 87 at r1](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok5yx21lfVQuOyljOx:-Nok5yx21lfVQuOyljOy:biomj3q) ([raw file](https://github.com/gramineproject/gramine/blob/cfe1527d49aad8147a7e8024a3339aafd7e6ebf7/python/graminelibos/manifest_check.py#L87)):*
> ```Python
>             'flags_mask': str,
>             'xfrm_mask': str,
>             'misc_mask': str,
> ```

These 3 have a specific format: https://gramine.readthedocs.io/en/stable/manifest-syntax.html#attribute-masks-for-sgx-sealing-key-derivation

Can we at least do the same as with `_size` -- create a global macro for them, and maybe later we'll add the proper formatting for them (a hex string with the `0x` prefix and exact number of digits)
___
*[`python/graminelibos/manifest_check.py` line 89 at r1](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok6Ejz45PiMgQVucEE:-Nok6Ejz45PiMgQVucEF:b-578g95) ([raw file](https://github.com/gramineproject/gramine/blob/cfe1527d49aad8147a7e8024a3339aafd7e6ebf7/python/graminelibos/manifest_check.py#L89)):*
> ```Python
>             'misc_mask': str,
>         },
>         'trusted_files': [Any(str, {'uri': str, 'sha256': str})],
> ```

ditto for `sha256` (macro with specific format)
___
*[`python/graminelibos/manifest_check.py` line 96 at r1](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok6hmFBvKxAdHTW3gh:-Nok6hmFBvKxAdHTW3gi:b-ub3135) ([raw file](https://github.com/gramineproject/gramine/blob/cfe1527d49aad8147a7e8024a3339aafd7e6ebf7/python/graminelibos/manifest_check.py#L96)):*
> ```Python
>         'allowed_ioctls': [{
>             Required('request_code'): int,
>             'struct': str,
> ```

This is also Required
___
*[`python/graminelibos/manifest_check.py` line 104 at r1](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok6sYN63DJcOewEOCC:-Nok6sYN63DJcOewEOCD:botmrrb) ([raw file](https://github.com/gramineproject/gramine/blob/cfe1527d49aad8147a7e8024a3339aafd7e6ebf7/python/graminelibos/manifest_check.py#L104)):*
> ```Python
>         'experimental__enable_flock': bool,
>         'insecure__allow_eventfd': bool,
>         'ioctl_structs': {str: [{str: object}]}, # XXX: I'm afraid description of this thing is above my pay grade
> ```

Does the `object` mean "whatever"? You don't need the inner `str:`, and just in general, why not:
```
'ioctl_structs': {str: object}
```

That should be enough.

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1726)* status: 5 of 10 files reviewed, 14 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok7ZsBEn60EnIBrkpz:-NoscSmU7JEjlefUpLZt:b-896fix) (no related file):*

CI fails with:
```
make: gramine-manifest-check: Command not found
```

Aren't you supposed to change some `meson.build` to install this tool?
</blockquote></details>

Done.

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok8BhA9KAmtczhaTa6:-NosdQAp2RUiQX-k1GbU:bto0xvw) (no related file):*

What about non-Debian distros? I guess you need to modify them too?
</blockquote></details>

Done. For now it's WIP, so wasn't a priority, those other distros are not tested in CI.

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok8F6r3gJOixNmUnRX:-NosdrJQBzC_iXA5whpE:b-ucfknh) (no related file):*

Brilliant PR!
</blockquote></details>

Thanks!

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok7czOC6QrSD-irSwq:-NosduihBNGzTat2cTMZ:b-gu8k61) (no related file):*

Can we add the call to this tool into `gramine-manifest`? It should be a new argument, and it should be disabled by default.

And we enable it for all our CI examples and for all our PAL/LibOS tests. This will we have the testing for both the new tool and for the manifest schema.
</blockquote></details>

That's a feature request. I'm not saying no, maybe I'll do it later, closer to finalising this PR. Plz keep this blocked.

(For the same reason I've left invocations in Makefile in helloworld example. I'll think what to do with examples later).

___
___
*[`python/graminelibos/manifest_check.py` line 23 at r1](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok2ZwU4gI3zKu3pEWd:-NoseNZm5nfC8_29dbnT:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/cfe1527d49aad8147a7e8024a3339aafd7e6ebf7/python/graminelibos/manifest_check.py#L23)):*

`tmpfs` doesn't have the `uri` (more correctly it's not required and always ignored), so it should be a separate item in the outer Any set
</blockquote></details>

Done.
___
*[`python/graminelibos/manifest_check.py` line 30 at r1](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok3RKs6ZJeN_lq4TKw:-NoseWVy0TmBD5rz93O9:b-93ueht) ([raw file](https://github.com/gramineproject/gramine/blob/cfe1527d49aad8147a7e8024a3339aafd7e6ebf7/python/graminelibos/manifest_check.py#L30)):*

There is also `fs.root.key_name`: https://github.com/gramineproject/gramine/blob/a933017d35a7aa3330194521389838de6be53dc4/libos/src/fs/libos_fs.c#L128

Actually, `root` should be exactly like an item in the `fs.mounts` set. Is this possible to do in Voluptuous (declare the `fs.mounts` item description separately and use it in two places)?
</blockquote></details>

Would be cleaner in JSON Schema, but here I can also do it, in global variable like `_size`.
___
*[`python/graminelibos/manifest_check.py` line 33 at r1](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok472p8K-mGSZki-x6:-NosjL8I0MsbjDpRt0qW:bemshpi) ([raw file](https://github.com/gramineproject/gramine/blob/cfe1527d49aad8147a7e8024a3339aafd7e6ebf7/python/graminelibos/manifest_check.py#L33)):*

Is this a tuple syntax (array)? It should be a dict syntax:
```
fs.insecure__keys.[KEY_NAME] = "[32-character hex value]"
```
</blockquote></details>

This was a `set` (1-element at that). Meant "a `set` consisting of instances of `str`"). A mistake, should have been `dict`, so `:` instead of `,`.
___
*[`python/graminelibos/manifest_check.py` line 42 at r1](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok4YpiB0eFGNon3Wky:-NoshKw243jDHhRb01X0:b-vhnck1) ([raw file](https://github.com/gramineproject/gramine/blob/cfe1527d49aad8147a7e8024a3339aafd7e6ebf7/python/graminelibos/manifest_check.py#L42)):*

Any way to specify a XOR in Voluptuous?

Here, `argv` and `argv_src_file` are mutually exclusive.
</blockquote></details>

Done.

It's ugly. Those are two separate validators, `All` means both of them needs to pass. Second one just checks this `Exclusive`, which I can't write in the first validator, because values have different types.

I think in JSON Schema this would be relatively nicer. It woudn't be prettier, but wouldn't stand out so much.
___
*[`python/graminelibos/manifest_check.py` line 63 at r1](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok5WEk5h3YDxNyRW5A:-NoshR6hAPGMobfobdB7:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/cfe1527d49aad8147a7e8024a3339aafd7e6ebf7/python/graminelibos/manifest_check.py#L63)):*

Typo: `allow_all_but_log` (not `byt`)
</blockquote></details>

Done.
___
*[`python/graminelibos/manifest_check.py` line 87 at r1](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok5yx21lfVQuOyljOx:-Nosi7wX91_ms6XIaOsG:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/cfe1527d49aad8147a7e8024a3339aafd7e6ebf7/python/graminelibos/manifest_check.py#L87)):*

These 3 have a specific format: https://gramine.readthedocs.io/en/stable/manifest-syntax.html#attribute-masks-for-sgx-sealing-key-derivation

Can we at least do the same as with `_size` -- create a global macro for them, and maybe later we'll add the proper formatting for them (a hex string with the `0x` prefix and exact number of digits)
</blockquote></details>

Done.
___
*[`python/graminelibos/manifest_check.py` line 89 at r1](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok6Ejz45PiMgQVucEE:-NosiDGd-TTecd2Uqcd_:bs593pp) ([raw file](https://github.com/gramineproject/gramine/blob/cfe1527d49aad8147a7e8024a3339aafd7e6ebf7/python/graminelibos/manifest_check.py#L89)):*

ditto for `sha256` (macro with specific format)
</blockquote></details>

This in only in 1 place, so probably won't be a type from global variable. I've added just a TODO for now.
___
*[`python/graminelibos/manifest_check.py` line 96 at r1](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok6hmFBvKxAdHTW3gh:-NosiVjp7TlhyeQNgwyP:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/cfe1527d49aad8147a7e8024a3339aafd7e6ebf7/python/graminelibos/manifest_check.py#L96)):*

This is also Required
</blockquote></details>

Done.
___
*[`python/graminelibos/manifest_check.py` line 104 at r1](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok6sYN63DJcOewEOCC:-NosiWb37OB5_Ka_6Bgl:b4u1rz5) ([raw file](https://github.com/gramineproject/gramine/blob/cfe1527d49aad8147a7e8024a3339aafd7e6ebf7/python/graminelibos/manifest_check.py#L104)):*

Does the `object` mean "whatever"? You don't need the inner `str:`, and just in general, why not:
```
'ioctl_structs': {str: object}
```

That should be enough.
</blockquote></details>

Done.

Yes. Giving a class as a validator means that the value needs to be an instance of this class, and since Python 3 everything is an `object`, so that's as "whatever" as it gets.

dimakuv (commented):
> Reviewed 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1726)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @woju)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok7czOC6QrSD-irSwq:-Noueu9OBTOqOzmaboqJ:bsreadq) (no related file):*

That's a feature request. I'm not saying no, maybe I'll do it later, closer to finalising this PR. Plz keep this blocked.

(For the same reason I've left invocations in Makefile in helloworld example. I'll think what to do with examples later).
</blockquote></details>

Yes, please think about this later, but after the discussion in the Gramine meeting, I'm neutral on this.

___
___
*[`CI-Examples/helloworld/Makefile` line 28 at r1](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok7zwKE8K8sfeFanu2:-NoueXCT7__cwxKBA6Qq:b-e3cpju) ([raw file](https://github.com/gramineproject/gramine/blob/cfe1527d49aad8147a7e8024a3339aafd7e6ebf7/CI-Examples/helloworld/Makefile#L28)):*

I suggest to incorporate the check into `gramine-manifest`
</blockquote></details>

As discussed in the meeting: maybe it makes more sense to keep it only as a separate tool (as done now). People anyway write scripts/Makefiles to hide the complexity of building a manifest, so the tool invocation can be plugged in there. I'm unblocking.
___
*[`CI-Examples/helloworld/Makefile` line 51 at r1](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok82Xf2YJs33bMklQ1:-NouelxPDIPMSIBbOFOK:b-2sglrm) ([raw file](https://github.com/gramineproject/gramine/blob/cfe1527d49aad8147a7e8024a3339aafd7e6ebf7/CI-Examples/helloworld/Makefile#L51)):*

Here I'm not sure, I would just remove this change.

We could check final `.sgx` manifests' schemas in LibOS regression tests, that should be enough coverage.
</blockquote></details>

ditto (unblocking)
___
*[`python/graminelibos/manifest_check.py` line 42 at r1](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok4YpiB0eFGNon3Wky:-Noucqr_2szMKZrFINJG:bl16okd) ([raw file](https://github.com/gramineproject/gramine/blob/cfe1527d49aad8147a7e8024a3339aafd7e6ebf7/python/graminelibos/manifest_check.py#L42)):*

Done.

It's ugly. Those are two separate validators, `All` means both of them needs to pass. Second one just checks this `Exclusive`, which I can't write in the first validator, because values have different types.

I think in JSON Schema this would be relatively nicer. It woudn't be prettier, but wouldn't stand out so much.
</blockquote></details>

Oh my. Can we revert this? Cognitive complexity of this outweighs the benefit.
___
*[`python/graminelibos/manifest_check.py` line 21 at r2](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nouc-5a50_MSUXdWQDX:-Nouc-5a50_MSUXdWQDY:b-td3k45) ([raw file](https://github.com/gramineproject/gramine/blob/62d2c37657e82f5a7a82a9d5dc4e2e52e2b0eaf3/python/graminelibos/manifest_check.py#L21)):*
> ```Python
>     },
>     {
>         'type': 'tmpfs',
> ```

This should also be `Required()`?
___
*[`python/graminelibos/manifest_check.py` line 26 at r2](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nouc1FQ85e5c52c5ciI:-Nouc1FQ85e5c52c5ciJ:b-6inis5) ([raw file](https://github.com/gramineproject/gramine/blob/62d2c37657e82f5a7a82a9d5dc4e2e52e2b0eaf3/python/graminelibos/manifest_check.py#L26)):*
> ```Python
>     },
>     {
>         'type': Any('chroot', 'untrusted_shm'),
> ```

This should also be `Required()`?

Or maybe even split into two items? Then we'll have a clear structure here -- every type of FS will have its own sub-schema. A bit redundant, but I think more readable.
___
*[`python/graminelibos/manifest_check.py` line 36 at r2](https://reviewable.io/reviews/gramineproject/gramine/1726#-NoucULv1pgLKy6EsCRT:-NoucULw7pH8BXHKP6KH:bfjyoue) ([raw file](https://github.com/gramineproject/gramine/blob/62d2c37657e82f5a7a82a9d5dc4e2e52e2b0eaf3/python/graminelibos/manifest_check.py#L36)):*
> ```Python
> _size = str
> 
> # masks in for sgx.seal_key.*_mask fields
> ```

`in for`? Use just one of them.

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1726)* status: 9 of 12 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok7czOC6QrSD-irSwq:-Nov3RcUBl4uXq_JaxnZ:b-eh9vox) (no related file):*

Yes, please think about this later, but after the discussion in the Gramine meeting, I'm neutral on this.
</blockquote></details>

I've added a check to gramine-manifest, but I'm still not sure. Let's see what happens in CI.

___
___
*[`python/graminelibos/manifest_check.py` line 42 at r1](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok4YpiB0eFGNon3Wky:-NouzNe1D6r1D60zGOj0:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/cfe1527d49aad8147a7e8024a3339aafd7e6ebf7/python/graminelibos/manifest_check.py#L42)):*

Oh my. Can we revert this? Cognitive complexity of this outweighs the benefit.
</blockquote></details>

Done.
___
*[`python/graminelibos/manifest_check.py` line 21 at r2](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nouc-5a50_MSUXdWQDX:-Nouxwaf47SvfDlrXWGJ:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/62d2c37657e82f5a7a82a9d5dc4e2e52e2b0eaf3/python/graminelibos/manifest_check.py#L21)):*

This should also be `Required()`?
</blockquote></details>

Done.
___
*[`python/graminelibos/manifest_check.py` line 26 at r2](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nouc1FQ85e5c52c5ciI:-NouxydY-8TVdR3ZsrV3:bkrq5bf) ([raw file](https://github.com/gramineproject/gramine/blob/62d2c37657e82f5a7a82a9d5dc4e2e52e2b0eaf3/python/graminelibos/manifest_check.py#L26)):*

This should also be `Required()`?

Or maybe even split into two items? Then we'll have a clear structure here -- every type of FS will have its own sub-schema. A bit redundant, but I think more readable.
</blockquote></details>

No, this shouldn't be `Required`: if type is missing, then it's `chroot`: https://github.com/gramineproject/gramine/blob/a933017d35a7aa3330194521389838de6be53dc4/libos/src/fs/libos_fs.c#L272.

Yes, at this point splitting is a very good idea.
___
*[`python/graminelibos/manifest_check.py` line 36 at r2](https://reviewable.io/reviews/gramineproject/gramine/1726#-NoucULv1pgLKy6EsCRT:-NouyvNGDapdPi3jgzhZ:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/62d2c37657e82f5a7a82a9d5dc4e2e52e2b0eaf3/python/graminelibos/manifest_check.py#L36)):*

`in for`? Use just one of them.
</blockquote></details>

Done.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1726)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

dimakuv (commented):
> Reviewed 6 of 6 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1726)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @woju)

___
*[`pal/regression/tests.toml` line 2 at r4](https://reviewable.io/reviews/gramineproject/gramine/1726#-NozWkfz0efxlY7fl_1R:-NozWkg-DG6NlJBrkIgy:b-ii6y2x) ([raw file](https://github.com/gramineproject/gramine/blob/7b6af3ddc7f3b942d5fc3db95fbf03a137bdda1c/pal/regression/tests.toml#L2)):*
> ```TOML
> binary_dir = "@GRAMINE_PKGLIBDIR@/tests/pal"
> no-check = true
> ```

This deserves a comment why we disable `gramine-manifest-check` for PAL tests (because they don't contain certain fields that we consider required in normal, LibOS-based, application manifests).
___
*[`python/graminelibos/manifest_check.py` line 44 at r4](https://reviewable.io/reviews/gramineproject/gramine/1726#-NozX3iWAY7qzAmyGYrw:-NozX3iWAY7qzAmyGYrx:b-twh356) ([raw file](https://github.com/gramineproject/gramine/blob/7b6af3ddc7f3b942d5fc3db95fbf03a137bdda1c/python/graminelibos/manifest_check.py#L44)):*
> ```Python
> 
> _fs_root = Any(*_fs_base)
> _fs_mount = Any(*({**d, Required('path'): str} for d in _fs_base))
> ```

Awesome solution! I even understand it.

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1726)* status: 13 of 15 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @woju)

___
*[`pal/regression/tests.toml` line 2 at r4](https://reviewable.io/reviews/gramineproject/gramine/1726#-NozWkfz0efxlY7fl_1R:-Nozl_Z3EPtXJhR-evSf:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/7b6af3ddc7f3b942d5fc3db95fbf03a137bdda1c/pal/regression/tests.toml#L2)):*

This deserves a comment why we disable `gramine-manifest-check` for PAL tests (because they don't contain certain fields that we consider required in normal, LibOS-based, application manifests).
</blockquote></details>

Done.
___
*[`python/graminelibos/manifest_check.py` line 96 at r1](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nok6hmFBvKxAdHTW3gh:-NozjmAz7BFdk3gDLHTC:b-32wbj4) ([raw file](https://github.com/gramineproject/gramine/blob/cfe1527d49aad8147a7e8024a3339aafd7e6ebf7/python/graminelibos/manifest_check.py#L96)):*

Done.
</blockquote></details>

@dimakuv is it really?

https://github.com/gramineproject/gramine/blob/36676a75e9897c9b7b183157a9d9fcdf2f668a13/libos/test/regression/device_ioctl.manifest.template#L61-L64
___
*[`python/graminelibos/manifest_check.py` line 44 at r4](https://reviewable.io/reviews/gramineproject/gramine/1726#-NozX3iWAY7qzAmyGYrw:-NozkDE2CQn_vtbU6S7a:b-ucfknh) ([raw file](https://github.com/gramineproject/gramine/blob/7b6af3ddc7f3b942d5fc3db95fbf03a137bdda1c/python/graminelibos/manifest_check.py#L44)):*

Awesome solution! I even understand it.
</blockquote></details>

Thanks!

dimakuv (commented):
> Reviewed 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1726)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @woju)

___
*[`Documentation/manpages/gramine-manifest-check.rst` line 28 at r6](https://reviewable.io/reviews/gramineproject/gramine/1726#-Np-o1nzApPwWMoUHI4T:-Np-o1o-EriuBDAuLorB:bvv2oic) ([raw file](https://github.com/gramineproject/gramine/blob/c95d611a7b212b8ff7bc2fdf05d82985157f0570/Documentation/manpages/gramine-manifest-check.rst#L28)):*
> ```reStructuredText
> implementation (``libsysdb.so``) are considered allowed/mandatory in schema.
> Therefore, if you have another LibOS implementation (like it happens in PAL
> test suite), the check may be wrong.
> ```

This is an internal detail that 99% of users will not understand... Do we really want it in this user-facing document?

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1726)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

___
*[`Documentation/manpages/gramine-manifest-check.rst` line 28 at r6](https://reviewable.io/reviews/gramineproject/gramine/1726#-Np-o1nzApPwWMoUHI4T:-Np05sZ052O3gMqMEmX-:bzid2ff) ([raw file](https://github.com/gramineproject/gramine/blob/c95d611a7b212b8ff7bc2fdf05d82985157f0570/Documentation/manpages/gramine-manifest-check.rst#L28)):*

This is an internal detail that 99% of users will not understand... Do we really want it in this user-facing document?
</blockquote></details>

I want this written somewhere, for completeness. I will forget this in a month.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1726)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

___
*[`Documentation/manpages/gramine-manifest-check.rst` line 28 at r6](https://reviewable.io/reviews/gramineproject/gramine/1726#-Np-o1nzApPwWMoUHI4T:-Np0BRSp8uulxZYNCCjv:b-dssqr5) ([raw file](https://github.com/gramineproject/gramine/blob/c95d611a7b212b8ff7bc2fdf05d82985157f0570/Documentation/manpages/gramine-manifest-check.rst#L28)):*

I want this written somewhere, for completeness. I will forget this in a month.
</blockquote></details>

Well, yeah, ok. Almost nobody (other than hard-core users) reads these manual pages anyway, so having such details is probably more useful than not.

dimakuv (approved):
> Reviewed 5 of 5 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1726)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

dimakuv (approved):
> Reviewed 1 of 6 files at r4.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1726)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

mkow (changes requested):
> Reviewed 4 of 6 files at r1, 2 of 5 files at r2, 3 of 6 files at r4, 1 of 2 files at r5, 1 of 1 files at r7, 5 of 5 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1726)* status: all files reviewed, 7 unresolved discussions (waiting on @woju)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nw6vnLS58q9vcfFp23v:-Nw6vnLS58q9vcfFp23w:b-gxeugd) (no related file):*
TODO before merge: check commit split.

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nw7316P2g2jHYuMTPHA:-Nw7316P2g2jHYuMTPHB:bsvny44) (no related file):*
This PR failed to build on RTD: https://readthedocs.org/projects/gramine/builds/24147011/, please check.

___
___
*[`Documentation/conf.py` line 150 at r8](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nw72WPR-Us3FDs6f_nd:-Nw72WPR-Us3FDs6f_ne:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/64f857fe332b3094a7ca4bdd4fa3bdd79b49b452/Documentation/conf.py#L150)):*
> ```Python
>     'click': ('https://click.palletsprojects.com/en/latest', None),
>     'cryptography': ('https://cryptography.io/en/latest', None),
>     'voluptuous': ('http://alecthomas.github.io/voluptuous/docs/_build/html', None),
> ```





_Suggestion:_
```Python
https://alecthomas.github.io/voluptuous/docs/_build/html
```
___
*[`pal/regression/tests.toml` line 7 at r8](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nw76OvHCDrDDG87U08O:-Nw76OvI5LDycMk8Ef9-:bt06wr9) ([raw file](https://github.com/gramineproject/gramine/blob/64f857fe332b3094a7ca4bdd4fa3bdd79b49b452/pal/regression/tests.toml#L7)):*
> ```TOML
> # do here is, we run libos replacements and manifests for those don't include entries that are
> # normally required by libsysdb. Therefore we need to disable schema checking.
> no-check = true
> ```

Just `no-check` as a global switch in `tests.toml` is too broad of a name IMO.



_Suggestion:_
```TOML
no-manifest-schema-check = true
```
___
*[`python/gramine-manifest-check` line 3 at r8](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nw78n8s9E1OeY04SBwb:-Nw78n8s9E1OeY04SBwc:b69dong) ([raw file](https://github.com/gramineproject/gramine/blob/64f857fe332b3094a7ca4bdd4fa3bdd79b49b452/python/gramine-manifest-check#L3)):*
> ```
> #!/usr/bin/python3
> # Copyright (C) 2024 Wojtek Porczyk <woju@invisiblethingslab.com>
> ```

This should be Intel.
ditto for the whole PR
___
*[`python/gramine-manifest-check` line 4 at r8](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nw79aKICx9Ai1ujL3Hk:-Nw79aKICx9Ai1ujL3Hl:b-pme2pg) ([raw file](https://github.com/gramineproject/gramine/blob/64f857fe332b3094a7ca4bdd4fa3bdd79b49b452/python/gramine-manifest-check#L4)):*
> ```
> # Copyright (C) 2024 Wojtek Porczyk <woju@invisiblethingslab.com>
> ```

I'd add a comment (or a doc paragraph) that this tool assumes that the input is a properly formatted TOML. Otherwise it prints an ugly stacktrace instead of a nice error to the user (which is probably ok here, because in normal flows it's being run on the output of `gramine-manifest`, which parses TOML and fails first (with an ugly stacktrace, so it should get a fix someday)).
___
*[`python/graminelibos/manifest_check.py` line 35 at r8](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nw7BhEa2kyLz-VKw4oM:-Nw7BhEa2kyLz-VKw4oN:b-i94u6w) ([raw file](https://github.com/gramineproject/gramine/blob/64f857fe332b3094a7ca4bdd4fa3bdd79b49b452/python/graminelibos/manifest_check.py#L35)):*
> ```Python
>     {
>         Required('type'): 'tmpfs',
>         'uri': str, # not _uri, this is ignored
> ```

It was a bit unclear to me at the first reading



_Suggestion:_
```Python
# not _uri, this field is ignored by Gramine
```

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1726)* status: 13 of 19 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nw7316P2g2jHYuMTPHA:-Nw8tuQK4LBUyMz74Yap:b-896fix) (no related file):*

This PR failed to build on RTD: https://readthedocs.org/projects/gramine/builds/24147011/, please check.
</blockquote></details>

Done.

___
___
*[`Documentation/conf.py` line 150 at r8](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nw72WPR-Us3FDs6f_nd:-Nw8twIP5meX_i1I_y_J:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/64f857fe332b3094a7ca4bdd4fa3bdd79b49b452/Documentation/conf.py#L150)):*
> ```Python
>     'click': ('https://click.palletsprojects.com/en/latest', None),
>     'cryptography': ('https://cryptography.io/en/latest', None),
>     'voluptuous': ('http://alecthomas.github.io/voluptuous/docs/_build/html', None),
> ```

Done.
___
*[`pal/regression/tests.toml` line 7 at r8](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nw76OvHCDrDDG87U08O:-Nw8xbLHEf3NDR1xPlAP:bywu1tp) ([raw file](https://github.com/gramineproject/gramine/blob/64f857fe332b3094a7ca4bdd4fa3bdd79b49b452/pal/regression/tests.toml#L7)):*

Just `no-check` as a global switch in `tests.toml` is too broad of a name IMO.
</blockquote></details>

I did `gramine-manifest-no-check` to match the cli option.
___
*[`python/gramine-manifest-check` line 3 at r8](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nw78n8s9E1OeY04SBwb:-Nw8ufmd0p8JV65JJmVM:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/64f857fe332b3094a7ca4bdd4fa3bdd79b49b452/python/gramine-manifest-check#L3)):*

This should be Intel.
ditto for the whole PR
</blockquote></details>

Done.
___
*[`python/gramine-manifest-check` line 4 at r8](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nw79aKICx9Ai1ujL3Hk:-Nw8wjov1oFpSALMr48t:bnfkp22) ([raw file](https://github.com/gramineproject/gramine/blob/64f857fe332b3094a7ca4bdd4fa3bdd79b49b452/python/gramine-manifest-check#L4)):*

I'd add a comment (or a doc paragraph) that this tool assumes that the input is a properly formatted TOML. Otherwise it prints an ugly stacktrace instead of a nice error to the user (which is probably ok here, because in normal flows it's being run on the output of `gramine-manifest`, which parses TOML and fails first (with an ugly stacktrace, so it should get a fix someday)).
</blockquote></details>

This is a trivial fix, so I'll fix here, and `gramine-manifest` we'll fix someday. Please file a bug if you care enough. :)
___
*[`python/graminelibos/manifest_check.py` line 35 at r8](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nw7BhEa2kyLz-VKw4oM:-Nw8xDDj5GSxKzPer_ZX:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/64f857fe332b3094a7ca4bdd4fa3bdd79b49b452/python/graminelibos/manifest_check.py#L35)):*

It was a bit unclear to me at the first reading
</blockquote></details>

Done.

dimakuv (dismissed):
> Reviewed 6 of 6 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1726)* status: all files reviewed, 7 unresolved discussions (waiting on @mkow and @woju)

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1726)* status: all files reviewed, 2 unresolved discussions (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1726#-Nw9fchN1oPH-r6MGniZ:-Nw9jbGpBDz6ikB6Vltj:b-896fix) (no related file):*

Please add "Fixes https://github.com/gramineproject/gramine/issues/1336" to the PR description on GitHub.
</blockquote></details>

Done.

___

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1726)* status: all files reviewed, 1 unresolved discussion (waiting on @woju)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1726)* status: all files reviewed, 1 unresolved discussion (waiting on @woju)

PR #1728, opened 2024-01-23
[LibOS] Add secure implementation of eventfd

dimakuv:
## Description of the changes 

The previous implementation of eventfd was insecure and disabled by default (enabled via `sys.insecure__allow_eventfd` manifest option). The new implementation of eventfd is secure but currently restricted to a single process. Thus, Gramine fails by default if eventfd is used and fork is issued. However, sometimes it is possible to use this secure implementation of eventfd in multi-process apps (if child processes do not use the parent's eventfd object and vice versa)~~; to support these cases, we introduce the `sys.experimental__allow_eventfd_fork` manifest option~~.

LibOS regression tests are added to test the secure implementation of eventfd. Memcached and Rust examples also use the secure eventfd now.

This eventfd implementation is an adaption of my tiny implementation of eventfd from Gramine-TDX PAL: https://github.com/gramineproject/gramine-tdx/blob/intel_tdx/pal/src/host/vm-common/pal_common_eventfd.c. A big difference is that I use host's eventfd object for notifications (e.g. for select/poll/epoll) in this PR.

## How to test this PR? 

LibOS tests included. Memcached and Rust Tokyo server are modified to not use the insecure eventfd version.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1728)

PR #1736, opened 2024-01-26
[LibOS] Do not perform `lock(&hdl->pos_lock)` on non-seeking handles

dimakuv:
## Description of the changes 

Handles can be file-like (allow seeking and thus have a file position) and stream-like (do not allow seeking and thus file position field `pos` is unused). Previously, LibOS emulations of `read()`, `write()`, `sendfile()`, etc. did not distinguish between file-like and stream-like handles and always acquired `hdl->pos_lock`.

This led to a bug on stream-like handles, in particular, on eventfd handles: `read()` and `write()` operations are blocking for eventfds, thus a blocking `read()` would acquire the `hdl->pos_lock` indefinitely and e.g. hang another thread that performs `write()` on same eventfd.

This commit fixes this bug by skipping this lock on stream-like handles. This has an additional performance benefit: stream-like handles like pipes, sockets and eventfds don't spend cycles in acquiring and releasing the lock inside the relevant (and very frequent) syscalls.

---

This bug was detected while working on https://github.com/gramineproject/gramine/pull/1728.

For context, this is the commit where this `pos_lock` was introduced: https://github.com/gramineproject/gramine/commit/e474134a4fd84fe72397e2f33a5a4e9950e06885#diff-d79d43998e924cff687e6a56d5f3b0643951f90b5b050730239d830e2f3e8b98

It's interesting to note that this bug does **not** exhibit itself on pipes, FIFOs and UNIX sockets (socketpairs). This is because these objects are unidirectional. I tried to write a test using a pipe or a socketpair first, but quickly realized that I can't write to and read from the same pipe/socket end even on normal Linux.

## How to test this PR? 

New LibOS regression test `eventfd_read_then_write` is added to test this bug fix.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1736)

### DIFF
--- a/libos/include/libos_handle.h
+++ b/libos/include/libos_handle.h
@@ -174,7 +174,12 @@ struct libos_handle {
      */
     struct libos_inode* inode;
 
-    /* Offset in file. Protected by `pos_lock`. */
+    /* Whether the handle is seekable (based on `type`). This field does not change, so reading it
+     * does not require holding any locks. Affects the `pos` field, see below. */
+    bool seekable;
+
+    /* Offset in file. Protected by `pos_lock`. Only meaningful for `seekable` handles; always zero
+     * for non-seekable handles. */
     file_off_t pos;
 
     /* This list contains `libos_epoll_item` objects this handle is part of. All accesses should be
@@ -221,7 +226,9 @@ struct libos_handle {
 
     /* Lock for handle position (`pos`). Intended for operations that change the position (e.g.
      * `read`, `seek` but not `pread`). This lock should be taken *before* `libos_handle.lock` and
-     * `libos_inode.lock`. */
+     * `libos_inode.lock`. Must be used *only* via maybe_lock_pos_handle() and
+     * maybe_unlock_pos_handle(); these functions make sure that the lock is acquired only on those
+     * handle types that are seekable (e.g. not on eventfds or pipes). */
     struct libos_lock pos_lock;
 };
 
@@ -295,3 +302,6 @@ int get_file_size(struct libos_handle* file, uint64_t* size);
 
 ssize_t do_handle_read(struct libos_handle* hdl, void* buf, size_t count);
 ssize_t do_handle_write(struct libos_handle* hdl, const void* buf, size_t count);
+
+void maybe_lock_pos_handle(struct libos_handle* hdl);
+void maybe_unlock_pos_handle(struct libos_handle* hdl);
--- a/libos/src/bookkeep/libos_handle.c
+++ b/libos/src/bookkeep/libos_handle.c
@@ -31,6 +31,16 @@ static MEM_MGR handle_mgr = NULL;
 
 #define INIT_HANDLE_MAP_SIZE 32
 
+void maybe_lock_pos_handle(struct libos_handle* hdl) {
+    if (hdl->seekable)
+        lock(&hdl->pos_lock);
+}
+
+void maybe_unlock_pos_handle(struct libos_handle* hdl) {
+    if (hdl->seekable)
+        unlock(&hdl->pos_lock);
+}
+
 int open_executable(struct libos_handle* hdl, const char* path) {
     struct libos_dentry* dent = NULL;
 
--- a/libos/src/fs/chroot/encrypted.c
+++ b/libos/src/fs/chroot/encrypted.c
@@ -196,6 +196,7 @@ static int chroot_encrypted_open(struct libos_handle* hdl, struct libos_dentry* 
     hdl->inode = dent->inode;
     get_inode(dent->inode);
     hdl->type = TYPE_CHROOT_ENCRYPTED;
+    hdl->seekable = true;
     hdl->pos = 0;
     return 0;
 }
@@ -232,6 +233,7 @@ static int chroot_encrypted_creat(struct libos_handle* hdl, struct libos_dentry*
     hdl->inode = dent->inode;
     get_inode(inode);
     hdl->type = TYPE_CHROOT_ENCRYPTED;
+    hdl->seekable = true;
     hdl->pos = 0;
     ret = 0;
 out:
--- a/libos/src/fs/chroot/fs.c
+++ b/libos/src/fs/chroot/fs.c
@@ -222,6 +222,7 @@ static int chroot_do_open(struct libos_handle* hdl, struct libos_dentry* dent, m
         uri = NULL;
 
         hdl->type = TYPE_CHROOT;
+        hdl->seekable = true;
         hdl->pos = 0;
         hdl->pal_handle = palhdl;
     } else {
--- a/libos/src/fs/libos_fs_pseudo.c
+++ b/libos/src/fs/libos_fs_pseudo.c
@@ -139,6 +139,7 @@ static int pseudo_open(struct libos_handle* hdl, struct libos_dentry* dent, int 
             }
 
             hdl->type = TYPE_STR;
+            hdl->seekable = true;
             mem_file_init(&hdl->info.str.mem, str, len);
             hdl->pos = 0;
             break;
@@ -146,6 +147,8 @@ static int pseudo_open(struct libos_handle* hdl, struct libos_dentry* dent, int 
 
         case PSEUDO_DEV: {
             hdl->type = TYPE_DEV;
+            hdl->seekable = true;
+            hdl->pos = 0;
             if (node->dev.dev_ops.open) {
                 ret = node->dev.dev_ops.open(hdl, dent, flags);
                 if (ret < 0)
@@ -482,11 +485,9 @@ static int pseudo_poll(struct libos_handle* hdl, int events, int* out_events) {
     switch (node->type) {
         case PSEUDO_STR: {
             assert(hdl->type == TYPE_STR);
-            lock(&hdl->pos_lock);
             lock(&hdl->lock);
             int ret = mem_file_poll(&hdl->info.str.mem, hdl->pos, events, out_events);
             unlock(&hdl->lock);
-            unlock(&hdl->pos_lock);
             return ret;
         }
 
--- a/libos/src/fs/libos_fs_util.c
+++ b/libos/src/fs/libos_fs_util.c
@@ -104,7 +104,10 @@ int generic_inode_hstat(struct libos_handle* hdl, struct stat* buf) {
 file_off_t generic_inode_seek(struct libos_handle* hdl, file_off_t offset, int origin) {
     file_off_t ret;
 
-    lock(&hdl->pos_lock);
+    if (!hdl->seekable)
+        return 0;
+
+    maybe_lock_pos_handle(hdl);
     lock(&hdl->inode->lock);
     file_off_t pos = hdl->pos;
     file_off_t size = hdl->inode->size;
@@ -115,7 +118,7 @@ file_off_t generic_inode_seek(struct libos_handle* hdl, file_off_t offset, int o
         ret = pos;
     }
     unlock(&hdl->inode->lock);
-    unlock(&hdl->pos_lock);
+    maybe_unlock_pos_handle(hdl);
     return ret;
 }
 
--- a/libos/src/fs/shm/fs.c
+++ b/libos/src/fs/shm/fs.c
@@ -76,6 +76,9 @@ static int shm_do_open(struct libos_handle* hdl, struct libos_dentry* dent, mode
     uri = NULL;
 
     hdl->type = TYPE_SHM;
+    hdl->seekable = true;
+    hdl->pos = 0;
+
     hdl->pal_handle = palhdl;
     ret = 0;
 
--- a/libos/src/fs/tmpfs/fs.c
+++ b/libos/src/fs/tmpfs/fs.c
@@ -136,6 +136,7 @@ static void tmpfs_do_open(struct libos_handle* hdl, struct libos_dentry* dent, i
     __UNUSED(flags);
 
     hdl->type = TYPE_TMPFS;
+    hdl->seekable = true;
     hdl->pos = 0;
 }
 
--- a/libos/src/sys/libos_file.c
+++ b/libos/src/sys/libos_file.c
@@ -481,7 +481,8 @@ long libos_syscall_sendfile(int out_fd, int in_fd, off_t* offset, size_t count) 
      * If `offset` is not NULL, we use `*offset` as starting offset for reading, and update
      * `*offset` afterwards (and keep the offset in input handle unchanged).
      *
-     * If `offset` is NULL, we use the offset in input handle, and update it afterwards.
+     * If `offset` is NULL, we use the offset in input handle, and update it afterwards (only if the
+     * input handle is seekable).
      */
     file_off_t pos_in = 0;
     if (offset) {
@@ -495,9 +496,9 @@ long libos_syscall_sendfile(int out_fd, int in_fd, off_t* offset, size_t count) 
             goto out;
         }
     } else {
-        lock(&in_hdl->pos_lock);
+        maybe_lock_pos_handle(in_hdl);
         pos_in = in_hdl->pos;
-        unlock(&in_hdl->pos_lock);
+        maybe_unlock_pos_handle(in_hdl);
     }
 
     if (!(out_hdl->acc_mode & MAY_WRITE)) {
@@ -523,9 +524,9 @@ long libos_syscall_sendfile(int out_fd, int in_fd, off_t* offset, size_t count) 
             break;
         }
 
-        lock(&out_hdl->pos_lock);
+        maybe_lock_pos_handle(out_hdl);
         ssize_t y = out_hdl->fs->fs_ops->write(out_hdl, buf, x, &out_hdl->pos);
-        unlock(&out_hdl->pos_lock);
+        maybe_unlock_pos_handle(out_hdl);
         if (y < 0) {
             ret = y;
             goto out_update;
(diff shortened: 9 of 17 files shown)


llly (commented):
> Reviewed 1 of 12 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1736)* status: 1 of 12 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`libos/src/fs/libos_fs_pseudo.c` line 485 at r1](https://reviewable.io/reviews/gramineproject/gramine/1736#-NpThSEu9q4uqh06iTSa:-NpThSEu9q4uqh06iTSb:b-711lmj) ([raw file](https://github.com/gramineproject/gramine/blob/322d9541131c9b76fc17b2f985fa7485f5a3f4ce/libos/src/fs/libos_fs_pseudo.c#L485)):*
> ```C
>         case PSEUDO_STR: {
>             assert(hdl->type == TYPE_STR);
>             lock_pos_handle(hdl);
> ```

Other pseudo functions don't lock the `pos_lock`. Callers lock `pos_lock` before calling them.
Should we also move this `lock_pos_handle(hdl);` to its caller function?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1736)* status: 1 of 13 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @llly)

___
*[`libos/src/fs/libos_fs_pseudo.c` line 485 at r1](https://reviewable.io/reviews/gramineproject/gramine/1736#-NpThSEu9q4uqh06iTSa:-NpU8mRF780NgqSTq5ra:b35wj2g) ([raw file](https://github.com/gramineproject/gramine/blob/322d9541131c9b76fc17b2f985fa7485f5a3f4ce/libos/src/fs/libos_fs_pseudo.c#L485)):*

Other pseudo functions don't lock the `pos_lock`. Callers lock `pos_lock` before calling them.
Should we also move this `lock_pos_handle(hdl);` to its caller function?
</blockquote></details>

Done. Good point.

llly (commented):
> Reviewed 10 of 12 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1736)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners

mkow (changes requested):
> Reviewed 7 of 12 files at r1.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1736)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1736#-NuB7eUw0k8jKPxemDOQ:-NuB7eUw0k8jKPxemDOR:b-q6n46v) (no related file):*
Hmm, how does file position concurrency works on Linux when a handle is accessed from many threads? I think it doesn't block like Gramine? Is this whole concept of locking the position for the whole duration of read/write actually right?

___
___
*[`libos/include/libos_handle.h` line 297 at r2](https://reviewable.io/reviews/gramineproject/gramine/1736#-NuBBDg8-ABdy0PLBWa4:-NuBBDg8-ABdy0PLBWa5:ba8cphy) ([raw file](https://github.com/gramineproject/gramine/blob/e129d57555bfc0a31d4163a1ee95eb4ace5b6242/libos/include/libos_handle.h#L297)):*
> ```objc
> ssize_t do_handle_write(struct libos_handle* hdl, const void* buf, size_t count);
> 
> void lock_pos_handle(struct libos_handle* hdl);
> ```

ditto for unlock



_Suggestion:_
```objc
maybe_lock_pos_handle
```
___
*[`libos/test/regression/eventfd_read_then_write.c` line 9 at r2](https://reviewable.io/reviews/gramineproject/gramine/1736#-NuAzV94366xFldexalN:-NuAzV95-U6qmi-JMbia:b-30o7gh) ([raw file](https://github.com/gramineproject/gramine/blob/e129d57555bfc0a31d4163a1ee95eb4ace5b6242/libos/test/regression/eventfd_read_then_write.c#L9)):*
> ```C
> #include <string.h>
> #include <unistd.h>
> #include <sys/eventfd.h>
> ```

Please sort.
___
*[`libos/test/regression/eventfd_read_then_write.c` line 27 at r2](https://reviewable.io/reviews/gramineproject/gramine/1736#-NuB17vH7V1DgGvSSPQS:-NuB17vI1BTC_-kiKBD4:b32vhv) ([raw file](https://github.com/gramineproject/gramine/blob/e129d57555bfc0a31d4163a1ee95eb4ace5b6242/libos/test/regression/eventfd_read_then_write.c#L27)):*
> ```C
>     while (__atomic_load_n(&g_read_thread_started, __ATOMIC_SEQ_CST) == false)
>         ;
>     sleep(1); /* make sure the other thread called the blocking read */
> ```

:(

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1736)* status: 5 of 13 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @llly and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1736#-NuB7eUw0k8jKPxemDOQ:-NuTwVKo3gsPJUl3YEaV:bck6zlb) (no related file):*

Hmm, how does file position concurrency works on Linux when a handle is accessed from many threads? I think it doesn't block like Gramine? Is this whole concept of locking the position for the whole duration of read/write actually right?
</blockquote></details>

Yes, file position (aka file offset) is atomic in `read()`, `write()`, `seek()`, etc.

It was actually a bug in the Linux kernel. It was fun to dig it out (you probably remember the old days of Linux before 3.14?):
- https://man7.org/linux/man-pages/man2/read.2.html#BUGS
- https://man7.org/linux/man-pages/man2/write.2.html#BUGS
- https://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html (Section 2.9.7)
- https://lore.kernel.org/lkml/53022DB1.4070805@gmail.com/ -- bug fixing thread

So the concept is correct.

___
___
*[`libos/include/libos_handle.h` line 297 at r2](https://reviewable.io/reviews/gramineproject/gramine/1736#-NuBBDg8-ABdy0PLBWa4:-NuTxPfE0XzdYA1dyNpr:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/e129d57555bfc0a31d4163a1ee95eb4ace5b6242/libos/include/libos_handle.h#L297)):*

ditto for unlock
</blockquote></details>

Done.
___
*[`libos/test/regression/eventfd_read_then_write.c` line 9 at r2](https://reviewable.io/reviews/gramineproject/gramine/1736#-NuAzV94366xFldexalN:-NuTyR7_9Tr5QL6LCt9k:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/e129d57555bfc0a31d4163a1ee95eb4ace5b6242/libos/test/regression/eventfd_read_then_write.c#L9)):*

Please sort.
</blockquote></details>

Done.
___
*[`libos/test/regression/eventfd_read_then_write.c` line 27 at r2](https://reviewable.io/reviews/gramineproject/gramine/1736#-NuB17vH7V1DgGvSSPQS:-NuTwvL_CvrFm4H8Zltr:bjlqd1m) ([raw file](https://github.com/gramineproject/gramine/blob/e129d57555bfc0a31d4163a1ee95eb4ace5b6242/libos/test/regression/eventfd_read_then_write.c#L27)):*

:(
</blockquote></details>

But there's no other way... Am I mistaken?

Or maybe you don't like that we wait for **a whole second**? Is your comment about replacing with `nanosleep()`?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1736)* status: 5 of 13 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @llly and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1736#-NuB7eUw0k8jKPxemDOQ:-NuU13-pDMZfHR3L0Xkz:b2rny62) (no related file):*

Yes, file position (aka file offset) is atomic in `read()`, `write()`, `seek()`, etc.

It was actually a bug in the Linux kernel. It was fun to dig it out (you probably remember the old days of Linux before 3.14?):
- https://man7.org/linux/man-pages/man2/read.2.html#BUGS
- https://man7.org/linux/man-pages/man2/write.2.html#BUGS
- https://pubs.opengroup.org/onlinepubs/9699919799/functions/V2_chap02.html (Section 2.9.7)
- https://lore.kernel.org/lkml/53022DB1.4070805@gmail.com/ -- bug fixing thread

So the concept is correct.
</blockquote></details>

And the commit by Linus itself: https://github.com/torvalds/linux/commit/9c225f2655e36a470c4f58dbbc99244c5fc7f2d4

___

mkow (changes requested):
> Reviewed 1 of 2 files at r2, 8 of 8 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1736)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @kailun-qin)

___
*[`libos/src/bookkeep/libos_handle.c` line 63 at r3](https://reviewable.io/reviews/gramineproject/gramine/1736#-NuWSzhPDOZVCMKHHycX:-NuWSzhPDOZVCMKHHycY:bcy2zra) ([raw file](https://github.com/gramineproject/gramine/blob/dc407f47c8a9d11ce9e7359559efdc486e64494c/libos/src/bookkeep/libos_handle.c#L63)):*
> ```C
>         else
>             unlock(&hdl->pos_lock);
>     }
> ```

I don't like that this is decided here, I think it would be cleaner if that was a property of a handle type, declared in a corresponding file. What do you think about adding `bool seekable` field to `struct libos_handle` and locking based on it's value? Should be roughly the same amount of code.
___
*[`libos/src/fs/libos_fs_pseudo.c` line 486 at r3](https://reviewable.io/reviews/gramineproject/gramine/1736#-NuWU2uj5kCKnW2kJ4Cf:-NuWU2uj5kCKnW2kJ4Cg:b5jklvn) ([raw file](https://github.com/gramineproject/gramine/blob/dc407f47c8a9d11ce9e7359559efdc486e64494c/libos/src/fs/libos_fs_pseudo.c#L486)):*
> ```C
>             assert(hdl->type == TYPE_STR);
>             lock(&hdl->lock);
>             int ret = mem_file_poll(&hdl->info.str.mem, hdl->pos, events, out_events);
> ```

This field is not atomic, I don't think we can read it without the lock?
// Update: see below.



_Code quote:_
```C
hdl->pos
```
___
*[`libos/src/fs/libos_fs_util.c` line 104 at r3](https://reviewable.io/reviews/gramineproject/gramine/1736#-NuWUIZdE-yMrQcM4tTz:-NuWUIZdE-yMrQcM4tU-:bbyypov) ([raw file](https://github.com/gramineproject/gramine/blob/dc407f47c8a9d11ce9e7359559efdc486e64494c/libos/src/fs/libos_fs_util.c#L104)):*
> ```C
> }
> 
> file_off_t generic_inode_seek(struct libos_handle* hdl, file_off_t offset, int origin) {
> ```

I guess here we could skip all the logic if we had `hdl->seekable == false`? And I guess it's also more in par with Linux, to always have pos==0 on non-seekable handles?
___
*[`libos/src/sys/libos_poll.c` line 134 at r3](https://reviewable.io/reviews/gramineproject/gramine/1736#-NuWV8-15h3LLhDaHjHM:-NuWV8-15h3LLhDaHjHN:b19kzbb) ([raw file](https://github.com/gramineproject/gramine/blob/dc407f47c8a9d11ce9e7359559efdc486e64494c/libos/src/sys/libos_poll.c#L134)):*
> ```C
>             maybe_lock_pos_handle(handle);
>             ret = handle->fs->fs_ops->poll(handle, events, &events);
>             maybe_unlock_pos_handle(handle);
> ```

Ah, I see, you moved the locking here. Why? (I'm not against, but would like to know the reason)
___
*[`libos/test/regression/eventfd_read_then_write.c` line 27 at r2](https://reviewable.io/reviews/gramineproject/gramine/1736#-NuB17vH7V1DgGvSSPQS:-NuWLvWh5FYSvbmoJJ8q:b-h3u36i) ([raw file](https://github.com/gramineproject/gramine/blob/e129d57555bfc0a31d4163a1ee95eb4ace5b6242/libos/test/regression/eventfd_read_then_write.c#L27)):*
> But there's no other way... Am I mistaken?

Hmm, yeah, I couldn't come up with anything else :/ Maybe @kailun-qin has an idea?

If not, then please at least change the comment, it's not "making sure", it's just a heuristic because we don't have a better idea how to assert that.

Also, it would be nice to make this test a bit more generic, i.e. testing concurrency, not just a single read and a single write. This should be (I think) just a pair of loops and checks on the eventfd values?

dimakuv:
Jenkins, retest Jenkins-SGX-Sanitizers please (unrelated `tests.test_sgx_sign.test_sign_from_pem_path` problem)

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1736)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`libos/src/fs/libos_fs_pseudo.c` line 120 at r5](https://reviewable.io/reviews/gramineproject/gramine/1736#-NuYus6aEbIp0bukmmIy:-NuZCPuG-9KT56KPaaON:bllaake) ([raw file](https://github.com/gramineproject/gramine/blob/4e13f396073c4ea03380218464380075feb5953c/libos/src/fs/libos_fs_pseudo.c#L120)):*

I only set `seekable = true` explicitly, in those cases where it's needed (regular files: chroot files, tmpfs files, shm dev files, encrypted files).

In all other cases `seekable = false` implicitly, and I decided not to clog our code with this (just like we don't set `pos = 0` in those cases).

Do you dislike this implicit logic?
</blockquote></details>

Ah, I see. I think it's ok, but I'd prefer to someday clean it up, i.e. make it clearer that all the structures' fields are always initialized.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1736)* status: 11 of 17 files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @llly and @mkow)

___
*[`-- commits` line 15 at r3](https://reviewable.io/reviews/gramineproject/gramine/1736#-NuYHw2u6R_BQSVGWJ2d):*

Will do: add that handles now have a new field `seekable`, and stream-like ones have it as `false` and file-like have it as `true`.
</blockquote></details>

Done

mkow (changes requested):
> Reviewed 6 of 6 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1736)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`libos/test/regression/test_libos.py` line 898 at r6](https://reviewable.io/reviews/gramineproject/gramine/1736#-NuZNhPw7QJ1zMyf_3qX:-NuZNhPw7QJ1zMyf_3qY:b9urueo) ([raw file](https://github.com/gramineproject/gramine/blob/79ff0e998dc5e256ae7ab25a67b11478460bc630/libos/test/regression/test_libos.py#L898)):*
> ```Python
>         self.assertIn('TEST OK', stdout)
> 
>     def test_071_eventfd_read_then_write(self):
> ```

Rebase fail here :) In the meantime another 071 test was added.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1736)* status: 16 of 17 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

___
*[`libos/test/regression/test_libos.py` line 898 at r6](https://reviewable.io/reviews/gramineproject/gramine/1736#-NuZNhPw7QJ1zMyf_3qX:-NuZOZL8A4PUdBraLdbX:bm59k45) ([raw file](https://github.com/gramineproject/gramine/blob/79ff0e998dc5e256ae7ab25a67b11478460bc630/libos/test/regression/test_libos.py#L898)):*

Rebase fail here :) In the meantime another 071 test was added.
</blockquote></details>

Done.

Let me also kill all Jenkins jobs, and restart it. Otherwise it will take ages.

kailun-qin (approved):
> Reviewed 4 of 12 files at r1, 3 of 8 files at r3, 9 of 10 files at r4, 1 of 1 files at r5, 6 of 6 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1736)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

___
*[`libos/test/regression/eventfd_read_then_write.c` line 27 at r2](https://reviewable.io/reviews/gramineproject/gramine/1736#-NuB17vH7V1DgGvSSPQS:-NuZRVxr5NIqgtJ56PiV:bdy0el2) ([raw file](https://github.com/gramineproject/gramine/blob/e129d57555bfc0a31d4163a1ee95eb4ace5b6242/libos/test/regression/eventfd_read_then_write.c#L27)):*

Done

1. I changed the comment.
2. I added 100 iterations, to test concurrency.
</blockquote></details>

yeah... I don't see a better solution w/o side effects

PR #1778, opened 2024-02-23
[PAL/Linux-SGX] Remove redundant checks

jkr0103:
## Description of the changes 

There are redundant checks performed inside the `sgx_copy_from_enclave_verified` and `sgx_copy_to_enclave_verified` functions as well as by the caller of these functions.

1-3% of performance improvement is seen with `Nginx` and `MySQL` workloads after removing the redundant checks.

This PR removes the redundant checks from caller.

## How to test this PR? 

Jenkins test should be enough

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1778)

### DIFF
--- a/pal/src/host/linux-sgx/enclave_framework.c
+++ b/pal/src/host/linux-sgx/enclave_framework.c
@@ -115,9 +115,6 @@ void* sgx_alloc_on_ustack(size_t size) {
 }
 
 void* sgx_copy_to_ustack(const void* ptr, size_t size) {
-    if (!sgx_is_completely_within_enclave(ptr, size)) {
-        return NULL;
-    }
     void* uptr = sgx_alloc_on_ustack(size);
     if (uptr) {
         sgx_copy_from_enclave_verified(uptr, ptr, size);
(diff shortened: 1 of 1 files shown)


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1778)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @jkr0103)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1778#-NrKtqYp63kfoPbhyTn2:-NrKtqYp63kfoPbhyTn3:bnhjziy) (no related file):*
This looks wrong. Inside the common helper functions `sgx_copy_to_enclave_verified()` and `sgx_copy_from_enclave_verified()` we have the following code:
```c
void sgx_copy_to_enclave_verified(void* ptr, const void* uptr, size_t size) {
    assert(sgx_is_valid_untrusted_ptr(uptr, size, /*alignment=*/1));
    assert(sgx_is_completely_within_enclave(ptr, size));
    ...
}

void sgx_copy_from_enclave_verified(void* uptr, const void* ptr, size_t size) {
    assert(sgx_is_completely_within_enclave(ptr, size));
    assert(sgx_is_valid_untrusted_ptr(uptr, size, /*alignment=*/1));
    ...
}
```

Notice the `assert()` surroundings. This means that such lines are **removed in release mode**. This renders `sgx_copy_to_enclave_verified()` and `sgx_copy_from_enclave_verified()` as insecure (not validating the enclave and buffer ranges), so the callers must do the validation themselves.

I agree that **in debug mode, we'll have redundancy**. But who cares about debug mode anyway.

___

PR #1779, opened 2024-02-23
[LibOS] Allow `unlink()` on corrupted encrypted files 

dimakuv:
## Description of the changes 

~~Previously, LibOS unconditionally set up the inode of the dentry on first lookup, for example, querying the size of the underlying file and saving it into inode fields. This setup of dentry's inode is not needed in cases like file removal (`unlink()`), since the inode will be immediately removed after it was set up. This was particularly painful for encrypted files (as they would open and decrypt the underlying file on inode setup) and tmpfs files (as they would allocate an underlying memory region for file contents on inode setup). Moreover, encrypted files were decrypted on lookup, just to get the real file size; this failed `unlink()` on corrupted or encrypted-with-old-key files.~~

~~This PR allows to skip the setup of the inode. Currently this is only used for the `unlink()` system call. This fixes the problematic case of encrypted files mentioned above and may improve tmpfs performance.~~

Currently, LibOS unconditionally sets up the inode of the dentry on first lookup, for example, querying the size of the underlying file and saving it into inode fields. This setup of dentry's inode is not needed in cases like file removal (`unlink()`), since the inode will be immediately removed after it was set up.

This pecularity made it impossible to remove (unlink) corrupted encrypted files by the app itself: already the lookup of the file would fail, disallowing to proceed to the removal operation. "Corrupted" may mean an attacker-modified file or a more benign case of encrypted-with-old-key file.

This PR allows to perform `unlink()` on such "corrupted" encrypted files (by keeping a "shallow" file object). A LibOS test was added.

## How to test this PR? 

A LibOS test was added.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1779)

### DIFF
--- a/libos/src/fs/chroot/encrypted.c
+++ b/libos/src/fs/chroot/encrypted.c
@@ -12,9 +12,15 @@
  *   key for files. Multiple mounts can use the same key. The list of keys is managed in
  *   `libos_fs_encrypted.c`.
  *
- * - Inodes (`libos_inode`, for regular files) hold a `libos_encrypted_file` object. This object
- *   lives as long as the inode, but is kept *open* only as long as there are `libos_handle` objects
- *   corresponding to it. We use `encrypted_file_{get,put}` operations to maintain that invariant.
+ * - Inodes (`libos_inode`, for regular files) hold a `libos_encrypted_file` object in `inode->data`
+ *   field. This object lives as long as the inode, but is kept *open* only as long as there are
+ *   `libos_handle` objects corresponding to it. We use `encrypted_file_{get,put}` operations to
+ *   maintain that invariant.
+ *
+ *   It is possible that inode has no corresponding `libos_encrypted_file` object, i.e.
+ *   `inode->data` is NULL. This happens if the encrypted file is corrupted; in this case we want to
+ *   allow at least some operations on the file (currently only `unlink()`), but all other
+ *   operations return -EACCES.
  *
  *   An open `libos_encrypted_file` object keeps an open PAL handle and associated data
  *   (`pf_context_t`), so that operations (read, write, truncate...) can be performed on the file.
@@ -152,18 +158,25 @@ static int chroot_encrypted_lookup(struct libos_dentry* dent) {
         struct libos_encrypted_files_key* key = dent->mount->data;
         ret = encrypted_file_open(uri, key, &enc);
         if (ret < 0) {
-            goto out;
-        }
+            if (ret == -EACCES) {
+                /* allow the inode to be created even if the underlying encrypted file is corrupted;
+                 * this is useful for unlinking a corrupted file */
+                inode->data = NULL;
+            } else {
+                goto out;
+            }
+        } else {
+            ret = encrypted_file_get_size(enc, &size);
+            encrypted_file_put(enc);
 
-        ret = encrypted_file_get_size(enc, &size);
-        encrypted_file_put(enc);
+            if (ret < 0) {
+                encrypted_file_destroy(enc);
+                goto out;
+            }
 
-        if (ret < 0) {
-            encrypted_file_destroy(enc);
-            goto out;
+            inode->data = enc;
+            inode->size = size;
         }
-        inode->data = enc;
-        inode->size = size;
     }
     dent->inode = inode;
     get_inode(inode);
@@ -185,6 +198,8 @@ static int chroot_encrypted_open(struct libos_handle* hdl, struct libos_dentry* 
 
     if (dent->inode->type == S_IFREG) {
         struct libos_encrypted_file* enc = dent->inode->data;
+        if (!enc)
+            return -EACCES;
 
         lock(&dent->inode->lock);
         ret = encrypted_file_get(enc);
@@ -325,6 +340,10 @@ static int chroot_encrypted_rename(struct libos_dentry* old, struct libos_dentry
     lock(&old->inode->lock);
 
     struct libos_encrypted_file* enc = old->inode->data;
+    if (!enc) {
+        ret = -EACCES;
+        goto out;
+    }
 
     ret = encrypted_file_get(enc);
     if (ret < 0)
@@ -341,6 +360,9 @@ out:
 static int chroot_encrypted_chmod(struct libos_dentry* dent, mode_t perm) {
     assert(locked(&g_dcache_lock));
     assert(dent->inode);
+
+    if (!dent->inode->data)
+        return -EACCES;
 
     char* uri = NULL;
 
@@ -374,6 +396,8 @@ static int chroot_encrypted_fchmod(struct libos_handle* hdl, mode_t perm) {
     assert(hdl->inode);
 
     struct libos_encrypted_file* enc = hdl->inode->data;
+    assert(enc);
+
     mode_t host_perm = HOST_PERM(perm);
     PAL_STREAM_ATTR attr = {.share_flags = host_perm};
     int ret = PalStreamAttributesSetByHandle(enc->pal_handle, &attr);
@@ -389,6 +413,7 @@ static int chroot_encrypted_flush(struct libos_handle* hdl) {
         return 0;
 
     struct libos_encrypted_file* enc = hdl->inode->data;
+    assert(enc);
 
     /* If there are any MAP_SHARED mappings for the file, this will write data to `enc` */
     int ret = msync_handle(hdl);
@@ -408,6 +433,7 @@ static int chroot_encrypted_close(struct libos_handle* hdl) {
         return 0;
 
     struct libos_encrypted_file* enc = hdl->inode->data;
+    assert(enc);
 
     lock(&hdl->inode->lock);
     encrypted_file_put(enc);
@@ -425,6 +451,8 @@ static ssize_t chroot_encrypted_read(struct libos_handle* hdl, void* buf, size_t
     }
 
     struct libos_encrypted_file* enc = hdl->inode->data;
+    assert(enc);
+
     size_t actual_count;
 
     lock(&hdl->inode->lock);
@@ -447,6 +475,8 @@ static ssize_t chroot_encrypted_write(struct libos_handle* hdl, const void* buf,
     }
 
     struct libos_encrypted_file* enc = hdl->inode->data;
+    assert(enc);
+
     size_t actual_count;
 
     lock(&hdl->inode->lock);
@@ -485,6 +515,7 @@ static int chroot_encrypted_truncate(struct libos_handle* hdl, file_off_t size) 
 
     int ret;
     struct libos_encrypted_file* enc = hdl->inode->data;
+    assert(enc);
 
     lock(&hdl->inode->lock);
     ret = encrypted_file_set_size(enc, size);
@@ -493,6 +524,16 @@ static int chroot_encrypted_truncate(struct libos_handle* hdl, file_off_t size) 
     unlock(&hdl->inode->lock);
 
     return ret;
+}
+
+static int chroot_encrypted_stat(struct libos_dentry* dent, struct stat* buf) {
+    assert(locked(&g_dcache_lock));
+    assert(dent->inode);
+
+    if (!dent->inode->data)
+        return -EACCES;
+
+    return generic_inode_stat(dent, buf);
 }
 
 struct libos_fs_ops chroot_encrypted_fs_ops = {
@@ -517,7 +558,7 @@ struct libos_d_ops chroot_encrypted_d_ops = {
     .lookup        = &chroot_encrypted_lookup,
     .creat         = &chroot_encrypted_creat,
     .mkdir         = &chroot_encrypted_mkdir,
-    .stat          = &generic_inode_stat,
+    .stat          = &chroot_encrypted_stat,
     .readdir       = &chroot_readdir, /* same as in `chroot` filesystem */
     .unlink        = &chroot_encrypted_unlink,
     .rename        = &chroot_encrypted_rename,
--- a/libos/test/regression/.gitignore
+++ b/libos/test/regression/.gitignore
@@ -3,6 +3,6 @@
 /trusted_testfile
 
 /tmp/*
-!/tmp/.dummy
+!/tmp/.gitkeep
 /tmp_enc/*
-!/tmp_enc/.dummy
+!/tmp_enc/.gitkeep
--- a/libos/test/regression/Makefile
+++ b/libos/test/regression/Makefile
@@ -17,5 +17,5 @@ clean:
 		testfile \
 		trusted_testfile \
 		asm/x86_64/__pycache__ \
-		tmp/* \
-		tmp_enc/*
+		tmp/*
+	find tmp_enc/ -not -path '*/.*' -type f -delete
--- a/libos/test/regression/manifest.template
+++ b/libos/test/regression/manifest.template
@@ -20,8 +20,8 @@ fs.mounts = [
 
   { type = "tmpfs", path = "/mnt/tmpfs" },
   { type = "encrypted", path = "/tmp_enc", uri = "file:tmp_enc", key_name = "my_custom_key" },
-  { type = "encrypted", path = "/encrypted_file_mrenclave.dat", uri = "file:encrypted_file_mrenclave.dat", key_name = "_sgx_mrenclave" },
-  { type = "encrypted", path = "/encrypted_file_mrsigner.dat", uri = "file:encrypted_file_mrsigner.dat", key_name = "_sgx_mrsigner" },
+  { type = "encrypted", path = "/tmp_enc/mrenclaves", uri = "file:tmp_enc/mrenclaves", key_name = "_sgx_mrenclave" },
+  { type = "encrypted", path = "/tmp_enc/mrsigners", uri = "file:tmp_enc/mrsigners", key_name = "_sgx_mrsigner" },
 ]
 
 sgx.max_threads = {{ '1' if env.get('EDMM', '0') == '1' else '16' }}
--- a/libos/test/regression/sealed_file.c
+++ b/libos/test/regression/sealed_file.c
@@ -19,8 +19,10 @@ int main(int argc, char** argv) {
     int ret;
     ssize_t bytes;
 
-    if (argc != 2)
-        errx(EXIT_FAILURE, "Usage: %s <protected file to create/validate>", argv[0]);
+    if (argc != 3 || (strcmp(argv[2], "unlink") && strcmp(argv[2], "nounlink"))) {
+        errx(EXIT_FAILURE, "Usage: %s <protected file to create/validate> unlink|nounlink",
+             argv[0]);
+    }
 
     ret = access(argv[1], F_OK);
     if (ret < 0) {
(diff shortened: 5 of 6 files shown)


mkow (changes requested):
> Reviewed 4 of 13 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1779)* status: 4 of 14 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`-- commits` line 11 at r2](https://reviewable.io/reviews/gramineproject/gramine/1779#-NsLghYfCBxOln2jjmpY):*
> Moreover, [...]

You just said exactly that in the previous sentence?
___
*[`libos/src/fs/tmpfs/fs.c` line 36 at r2](https://reviewable.io/reviews/gramineproject/gramine/1779#-NsLjj0X0pVGYyfmkeXu:-NsLjj0X0pVGYyfmkeXv:b-owo0ln) ([raw file](https://github.com/gramineproject/gramine/blob/274d0eeadeba1b381deb92d8b7f5659c2b634551/libos/src/fs/tmpfs/fs.c#L36)):*
> ```C
>         return -ENOMEM;
> 
>     if (!skip_inode_setup) {
> ```

Ehh, I'm not a fan of this, now we have some magic bool which leaves the `inode` struct in a partially-initialized state with some fields being initialized and some not. And it's not even documented which ones are which and `unlink` code actually uses that initialized subset...
___
*[`libos/test/regression/tmp_enc/mrenclaves/.dummy` line 0 at r2](https://reviewable.io/reviews/gramineproject/gramine/1779#-NsLijrA1JoWFx_imn9W:-NsLijrA1JoWFx_imn9X:b-22wpbz) ([raw file](https://github.com/gramineproject/gramine/blob/0ca3cbd5d24eb5e0ec1fe51995e5954dbbc47061/libos/test/regression/tmp_enc/mrenclaves/.dummy#L0)):*
I think I prefer `.gitkeep`, because it says why we need such a file.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1779)* status: 4 of 14 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

___
*[`libos/src/fs/tmpfs/fs.c` line 36 at r2](https://reviewable.io/reviews/gramineproject/gramine/1779#-NsLjj0X0pVGYyfmkeXu:-NsN4gbh2tnBzsJpaPO2:b6uh2pe) ([raw file](https://github.com/gramineproject/gramine/blob/274d0eeadeba1b381deb92d8b7f5659c2b634551/libos/src/fs/tmpfs/fs.c#L36)):*

Ehh, I'm not a fan of this, now we have some magic bool which leaves the `inode` struct in a partially-initialized state with some fields being initialized and some not. And it's not even documented which ones are which and `unlink` code actually uses that initialized subset...
</blockquote></details>

First of all, I agree that this is a bad coding practice. This was the first idea / workaround that came to my mind, but I also prefer now to change the logic.

After thinking about this, I propose the following:
1. We get rid of `xxx_setup_dentry()` functions:
   - It currently does two things: (a) creates the new inode + attaches it to the dentry, and (b) sets up the additional inode fields like size, create/modify time, etc.
   - The name is meaningless; a better name would be `xxx_create_and_attach_inode()`
   - Some FS implementations (Pseudo, Encrypted) already got rid of this helper function and just inline creation of the inode object; the rest FS implementations (Tmpfs, Chroot, Shm) have this helper function but could easily inline the logic.
2. We split what `xxx_setup_dentry()` func does into two parts:
   - The part of "create the new inode + attach it to dentry" is inlined into corresponding callers (like already done in Pseudo and Encrypted)
   - The part of "set up additional inode fields" is moved to separate helper functions, and this set up is performed **on demand** (lazily), instead of the current eager approach.
3. We clearly separate and document the fields that are immediately initialized in the inode object:
   - I'd like to note that we **must create the inode** during lookup, because inode represents the actual file (whereas dentry is more like a wrapper object that just keeps the info about path traversals)
   - But the inode only needs a small subset of fields during its creation: the FS type, the file type, the permissions, and maybe the uid/gid.
   - The rest inode fields are initialized first with "none" values (e.g. `size = -1` I guess?), and only set up on demand.

In this case, we get rid of this `skip_inode_setup` flag that we have to drag everywhere in this PR, and flip the logic to a more readable one: **instead of skipping set up of inodes in otherwise eager file lookup, we set up inode fields on demand**.

Then the problem of Encrypted and Tmpfs files -- as explained in the PR description -- will go away automatically.

@mkow @kailun-qin What do you think about this proposal?

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1779)* status: 4 of 14 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @kailun-qin)

___
*[`libos/src/fs/tmpfs/fs.c` line 36 at r2](https://reviewable.io/reviews/gramineproject/gramine/1779#-NsLjj0X0pVGYyfmkeXu:-NtNEZC6ALH77BghcbMs:b-si1a52) ([raw file](https://github.com/gramineproject/gramine/blob/274d0eeadeba1b381deb92d8b7f5659c2b634551/libos/src/fs/tmpfs/fs.c#L36)):*

First of all, I agree that this is a bad coding practice. This was the first idea / workaround that came to my mind, but I also prefer now to change the logic.

After thinking about this, I propose the following:
1. We get rid of `xxx_setup_dentry()` functions:
   - It currently does two things: (a) creates the new inode + attaches it to the dentry, and (b) sets up the additional inode fields like size, create/modify time, etc.
   - The name is meaningless; a better name would be `xxx_create_and_attach_inode()`
   - Some FS implementations (Pseudo, Encrypted) already got rid of this helper function and just inline creation of the inode object; the rest FS implementations (Tmpfs, Chroot, Shm) have this helper function but could easily inline the logic.
2. We split what `xxx_setup_dentry()` func does into two parts:
   - The part of "create the new inode + attach it to dentry" is inlined into corresponding callers (like already done in Pseudo and Encrypted)
   - The part of "set up additional inode fields" is moved to separate helper functions, and this set up is performed **on demand** (lazily), instead of the current eager approach.
3. We clearly separate and document the fields that are immediately initialized in the inode object:
   - I'd like to note that we **must create the inode** during lookup, because inode represents the actual file (whereas dentry is more like a wrapper object that just keeps the info about path traversals)
   - But the inode only needs a small subset of fields during its creation: the FS type, the file type, the permissions, and maybe the uid/gid.
   - The rest inode fields are initialized first with "none" values (e.g. `size = -1` I guess?), and only set up on demand.

In this case, we get rid of this `skip_inode_setup` flag that we have to drag everywhere in this PR, and flip the logic to a more readable one: **instead of skipping set up of inodes in otherwise eager file lookup, we set up inode fields on demand**.

Then the problem of Encrypted and Tmpfs files -- as explained in the PR description -- will go away automatically.

@mkow @kailun-qin What do you think about this proposal?
</blockquote></details>

Sounds good. One small change maybe: are the lazily-initialized fields always initialized together? I.e. do I understand correctly, that the whole inode structure can only be in 2 states: "early-initialized" and fully initialized? In this case, could we move these lazy fields into a separate structure and point to it from some inode field, which would be `NULL` in "early-initialized" state? This way we can be sure that nothing accidentally uses bogus values (like size == -1) for computations.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1779)* status: 4 of 14 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`libos/src/fs/tmpfs/fs.c` line 36 at r2](https://reviewable.io/reviews/gramineproject/gramine/1779#-NsLjj0X0pVGYyfmkeXu:-NuRRE4QBBYyD1vtjmOh:b133v7g) ([raw file](https://github.com/gramineproject/gramine/blob/274d0eeadeba1b381deb92d8b7f5659c2b634551/libos/src/fs/tmpfs/fs.c#L36)):*
Sounds good to me.

> are the lazily-initialized fields always initialized together?

I suspect not.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1779)* status: 4 of 14 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`libos/src/fs/tmpfs/fs.c` line 36 at r2](https://reviewable.io/reviews/gramineproject/gramine/1779#-NsLjj0X0pVGYyfmkeXu:-NuUQbuQ-u6fI4zUELMc:b-6c8sex) ([raw file](https://github.com/gramineproject/gramine/blob/274d0eeadeba1b381deb92d8b7f5659c2b634551/libos/src/fs/tmpfs/fs.c#L36)):*
> One small change maybe: are the lazily-initialized fields always initialized together?

Took a look at this -- no, the lazily-initialized fields are **not** initialized together.

So I don't think we could use your idea with a special `NULL` field that is lazily allocated.

I will instead continue with my original idea -- to just write a comment for `struct libos_inode` which fields are early initialized and which are initialized later (on `open()` actually).

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1779)* status: 0 of 21 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

___
*[`-- commits` line 11 at r2](https://reviewable.io/reviews/gramineproject/gramine/1779#-NsLghYfCBxOln2jjmpY):*

> Moreover, [...]

You just said exactly that in the previous sentence?
</blockquote></details>

I modified the commit message completely.
___
*[`libos/test/regression/sealed_file.c` line 43 at r1](https://reviewable.io/reviews/gramineproject/gramine/1779#-NrKsMr1A-ncxuwQUAhh:-NuZGXy4BxwCeKYQeKLu:bgdq47u) ([raw file](https://github.com/gramineproject/gramine/blob/41e17c074ff3a8565a99b1b971e1df931e7b2c14/libos/test/regression/sealed_file.c#L43)):*

This is not amazing. I was thinking whether I need to add this `skip_inode_setup = true` to the `access()` system call, but then decided against it. `access()` may legitimately setup the inode, and it will be used afterwards, e.g. in subsequent `open()`. So I kept this rather weird logic in the test.

This also satisfies the customer who found this bug.
</blockquote></details>

This comment is still relevant (other than `skip_inode_setup = true` part).
___
*[`libos/src/fs/tmpfs/fs.c` line 36 at r2](https://reviewable.io/reviews/gramineproject/gramine/1779#-NsLjj0X0pVGYyfmkeXu:-NuZGjng60OdsuHzSKlk:b-pkve1d) ([raw file](https://github.com/gramineproject/gramine/blob/274d0eeadeba1b381deb92d8b7f5659c2b634551/libos/src/fs/tmpfs/fs.c#L36)):*

> One small change maybe: are the lazily-initialized fields always initialized together?

Took a look at this -- no, the lazily-initialized fields are **not** initialized together.

So I don't think we could use your idea with a special `NULL` field that is lazily allocated.

I will instead continue with my original idea -- to just write a comment for `struct libos_inode` which fields are early initialized and which are initialized later (on `open()` actually).
</blockquote></details>

Ok, this turned out to be a total mess. 

I made a failed attempt at splitting into early-init and lazy-init parts of the inode. There's too much complexity, as many parts in Gramine's FS rely on the fact that e.g. `inode->size` is initialized immediately at lookup (i.e. at first "notice" of the file).

In particular, I got fed up with `generic_inode_stat()` and decided for a much more contained and pointed approach -- modifying only the Encrypted Files code. I force-pushed since everything is changed in this PR, except for the test.
___
*[`libos/test/regression/tmp_enc/mrenclaves/.dummy` line  at r2](https://reviewable.io/reviews/gramineproject/gramine/1779#-NsLijrA1JoWFx_imn9W:-NuYRXFpDeWp0wldxqLN:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/0ca3cbd5d24eb5e0ec1fe51995e5954dbbc47061/libos/test/regression/tmp_enc/mrenclaves/.dummy#L)):*

I think I prefer `.gitkeep`, because it says why we need such a file.
</blockquote></details>

Done.

mkow (changes requested):
> Reviewed 3 of 13 files at r1, 1 of 1 files at r2, 17 of 17 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1779)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`-- commits` line 10 at r3](https://reviewable.io/reviews/gramineproject/gramine/1779#-NuaMrbP16OpAqKYxZ7B):*




_Suggestion:_
```
peculiarity
```
___
*[`libos/src/fs/chroot/encrypted.c` line 163 at r3](https://reviewable.io/reviews/gramineproject/gramine/1779#-NuaWC8c65krmJAhwDJM:-NuaWC8c65krmJAhwDJN:b-g9l6pk) ([raw file](https://github.com/gramineproject/gramine/blob/97ee80a53cdfa6fbc739521f0e1168b2c58f100d/libos/src/fs/chroot/encrypted.c#L163)):*
> \`unlink\`ing

I'd rather just write `unlinking` or `unlink()-ing`.
___
*[`libos/src/fs/chroot/encrypted.c` line 342 at r3](https://reviewable.io/reviews/gramineproject/gramine/1779#-NuaX0QAE-JOSqLPnV4K:-NuaX0QAE-JOSqLPnV4L:bp63hjh) ([raw file](https://github.com/gramineproject/gramine/blob/97ee80a53cdfa6fbc739521f0e1168b2c58f100d/libos/src/fs/chroot/encrypted.c#L342)):*
> ```C
>     struct libos_encrypted_file* enc = old->inode->data;
>     if (!enc)
>         return -EACCES;
> ```

`old->inode->lock` and `new_uri` leak here
___
*[`libos/test/regression/sealed_file.c` line 24 at r3](https://reviewable.io/reviews/gramineproject/gramine/1779#-NuaTiLPELCCK-UBKQf5:-NuaTiLPELCCK-UBKQf6:bnszmuj) ([raw file](https://github.com/gramineproject/gramine/blob/97ee80a53cdfa6fbc739521f0e1168b2c58f100d/libos/test/regression/sealed_file.c#L24)):*
> ```C
>     if (argc != 3)
>         errx(EXIT_FAILURE, "Usage: %s <protected file to create/validate> <unlink?>", argv[0]);
> ```

Please also check that `argv[2]` is either `unlink` or `nounlink` (right now anything equals to `nounlink`)
___
*[`libos/test/regression/sealed_file.c` line 56 at r3](https://reviewable.io/reviews/gramineproject/gramine/1779#-NuaTA857C2fxd2VF9jO:-NuaTA857C2fxd2VF9jP:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/97ee80a53cdfa6fbc739521f0e1168b2c58f100d/libos/test/regression/sealed_file.c#L56)):*
> ```C
>             err(EXIT_FAILURE, "unlink failed");
> 
>         printf("UNLINK OK");
> ```





_Suggestion:_
```C
puts("UNLINK OK");
```

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1779)* status: 19 of 21 files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`-- commits` line 10 at r3](https://reviewable.io/reviews/gramineproject/gramine/1779#-NuaMrbP16OpAqKYxZ7B):*
Will do at final rebase
___
*[`libos/src/fs/chroot/encrypted.c` line 163 at r3](https://reviewable.io/reviews/gramineproject/gramine/1779#-NuaWC8c65krmJAhwDJM:-Nuc6UJ69Ml8OORBASgB:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/97ee80a53cdfa6fbc739521f0e1168b2c58f100d/libos/src/fs/chroot/encrypted.c#L163)):*

> \`unlink\`ing

I'd rather just write `unlinking` or `unlink()-ing`.
</blockquote></details>

Done.
___
*[`libos/src/fs/chroot/encrypted.c` line 342 at r3](https://reviewable.io/reviews/gramineproject/gramine/1779#-NuaX0QAE-JOSqLPnV4K:-Nuc6gkOCvS0DJxUQPXK:b-u661y0) ([raw file](https://github.com/gramineproject/gramine/blob/97ee80a53cdfa6fbc739521f0e1168b2c58f100d/libos/src/fs/chroot/encrypted.c#L342)):*

`old->inode->lock` and `new_uri` leak here
</blockquote></details>

Done, copy-paste fail
___
*[`libos/test/regression/sealed_file.c` line 24 at r3](https://reviewable.io/reviews/gramineproject/gramine/1779#-NuaTiLPELCCK-UBKQf5:-Nuc7J0o9NBtX18IGQup:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/97ee80a53cdfa6fbc739521f0e1168b2c58f100d/libos/test/regression/sealed_file.c#L24)):*

Please also check that `argv[2]` is either `unlink` or `nounlink` (right now anything equals to `nounlink`)
</blockquote></details>

Done.
___
*[`libos/test/regression/sealed_file.c` line 56 at r3](https://reviewable.io/reviews/gramineproject/gramine/1779#-NuaTA857C2fxd2VF9jO:-Nuc7SwOCQxKdwzB3UiD:b-ow196j) ([raw file](https://github.com/gramineproject/gramine/blob/97ee80a53cdfa6fbc739521f0e1168b2c58f100d/libos/test/regression/sealed_file.c#L56)):*
> ```C
>             err(EXIT_FAILURE, "unlink failed");
> 
>         printf("UNLINK OK");
> ```

Done. And in other places too.

kailun-qin (approved):
> Reviewed 2 of 13 files at r1, 1 of 1 files at r2, 16 of 17 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1779)* status: all files reviewed, 8 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1779)* status: all files reviewed, 4 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

dimakuv:
Jenkins, retest Jenkins-SGX-20.04-musl please (`tests.test_sgx_sign.test_sign_from_pem_path_with_passphrase` failed, unrelated)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1779)* status: 19 of 21 files reviewed, 1 unresolved discussion (waiting on @kailun-qin and @mkow)

___
*[`-- commits` line 10 at r3](https://reviewable.io/reviews/gramineproject/gramine/1779#-NuaMrbP16OpAqKYxZ7B):*

Will do at final rebase
</blockquote></details>

Done.

dimakuv:
Jenkins, retest Jenkins-SGX-20.04-apps please (Jenkins didn't trigger this job)

dimakuv:
Jenkins, retest Jenkins-SGX-20.04-musl please (Jenkins didn't trigger this job)

kailun-qin (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1779)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #1795, opened 2024-03-05
[LibOS] Use RW locks in the VMA tree

### LINKED ISSUE #1794: [LibOS] Use RW locks in the VMA tree
### Problem

Multi-threaded workloads with many syscalls stress the VMA subsystem a lot, because almost all syscalls verify their buffers for read/write access using the following functions:
- `is_user_memory_readable()`
- `is_user_memory_writable()`
- `is_user_string_readable()`
- `is_user_memory_writable_no_skip()`

All these functions call `test_user_memory()` helper: https://github.com/gramineproject/gramine/blob/4afc5509cdd7a91b391f83407860f5607c323600/libos/src/bookkeep/libos_signal.c#L393-L405

This helper in turn calls the `is_in_adjacent_user_vmas()` func: https://github.com/gramineproject/gramine/blob/4afc5509cdd7a91b391f83407860f5607c323600/libos/src/bookkeep/libos_vma.c#L1204-L1219

The important part is `spinlock_lock(&vma_tree_lock)` and `spinlock_unlock(&vma_tree_lock)`. On a multi-threaded app, this lock contention becomes the bottleneck.

Gramine introduced a workaround to sidestep this bottleneck, via the `libos.check_invalid_pointers` manifest option; it translates to the `g_check_invalid_ptrs` variable. However, this cannot be used in all cases:
- Some runtimes like Java rely on being able to check invalid pointers. Thus they cannot set `libos.check_invalid_pointers = false`; this would lead to Java apps failing.
- The `is_user_memory_writable_no_skip()` function does **not** honor the `libos.check_invalid_pointers` manifest option; this is because in certain situations Gramine really must decide whether the VMA is writable or read-only, see e.g. [the `ppoll()` case](https://github.com/gramineproject/gramine/blob/63ee38d5eb28f5a880a973c28de7be6f0dc6f290/libos/src/sys/libos_poll.c#L286-L290) which emulates how [Linux works](https://github.com/torvalds/linux/blob/cf1182944c7cc9f1c21a8a44e0d29abe12527412/fs/select.c#L356-L362).

### Solution

Use the RW lock that was previously introduced in the Gramine codebase: https://github.com/gramineproject/gramine/blob/master/libos/include/libos_rwlock.h

Example usage of this RW lock: https://github.com/gramineproject/gramine/commit/f0714505fb376534a9a83a6b963ae6c3bf98270a

### Benchmark results

TODO

dimakuv:
## Description of the changes 

Multi-threaded workloads with many syscalls stress the VMA subsystem of LibOS, because almost all syscalls verify their buffers for read/write access using the functions `is_user_memory_readable()`, `is_user_memory_writable()`, etc. All these functions end up in VMA-specific `is_in_adjacent_user_vmas()` that grabs a global VMA lock. On some multi-threaded apps like MongoDB, this lock contention becomes the performance bottleneck.

This commit tries to remove this bottleneck by switching from a spinlock to the Read-Write (RW) lock. The intuition is that most of the time, a read-only `is_in_adjacent_user_vmas()` func is called, which now uses the read lock.

Fixes #1794.

## How to test this PR? 

CI for functionality testing. Manual benchmarks for perf testing.

My quick tests on Memcached, Blender and iperf show no visible change in performance. This makes sense: these workloads use no more than 4 threads, so almost no contention.

**WIP: I asked to run big workloads.**

### DIFF
--- a/libos/src/bookkeep/libos_vma.c
+++ b/libos/src/bookkeep/libos_vma.c
@@ -18,11 +18,11 @@
 #include "libos_handle.h"
 #include "libos_internal.h"
 #include "libos_lock.h"
+#include "libos_rwlock.h"
 #include "libos_tcb.h"
 #include "libos_utils.h"
 #include "libos_vma.h"
 #include "linux_abi/memory.h"
-#include "spinlock.h"
 
 /* The amount of total memory usage, all accesses must be protected by `vma_tree_lock`. */
 static size_t g_total_memory_size = 0;
@@ -102,10 +102,57 @@ static bool cmp_addr_to_vma(void* addr, struct avl_tree_node* node) {
  * to be revisited as there might be some optimizations that would break due to it.
  */
 static struct avl_tree vma_tree = {.cmp = vma_tree_cmp};
-static spinlock_t vma_tree_lock = INIT_SPINLOCK_UNLOCKED;
+static struct libos_rwlock vma_tree_lock;
+static bool vma_tree_lock_created = false;
+
+static inline void vma_rwlock_read_lock(struct libos_rwlock* l) {
+    if (!vma_tree_lock_created)
+        return;
+    rwlock_read_lock(l);
+}
+
+static inline void vma_rwlock_read_unlock(struct libos_rwlock* l) {
+    if (!vma_tree_lock_created)
+        return;
+    rwlock_read_unlock(l);
+}
+
+static inline void vma_rwlock_write_lock(struct libos_rwlock* l) {
+    if (!vma_tree_lock_created)
+        return;
+    rwlock_write_lock(l);
+}
+
+static inline void vma_rwlock_write_unlock(struct libos_rwlock* l) {
+    if (!vma_tree_lock_created)
+        return;
+    rwlock_write_unlock(l);
+}
+
+#ifdef DEBUG
+static inline bool vma_rwlock_is_read_locked(struct libos_rwlock* l) {
+    if (!vma_tree_lock_created)
+        return true;
+    return rwlock_is_read_locked(l);
+}
+
+static inline bool vma_rwlock_is_write_locked(struct libos_rwlock* l) {
+    if (!vma_tree_lock_created)
+        return true;
+    return rwlock_is_write_locked(l);
+}
+#endif
+
+/* VMA code is supposed to use the vma_* wrappers of RW lock; hide the actual RW lock funcs */
+#define rwlock_is_write_locked(x) false
+#define rwlock_is_read_locked(x) false
+#define rwlock_read_lock(x) static_assert(false, "hidden func")
+#define rwlock_read_unlock(x) static_assert(false, "hidden func")
+#define rwlock_write_lock(x) static_assert(false, "hidden func")
+#define rwlock_write_unlock(x) static_assert(false, "hidden func")
 
 static void total_memory_size_add(size_t length) {
-    assert(spinlock_is_locked(&vma_tree_lock));
+    assert(vma_rwlock_is_write_locked(&vma_tree_lock));
 
     g_total_memory_size += length;
 
@@ -117,7 +164,7 @@ static void total_memory_size_add(size_t length) {
 }
 
 static void total_memory_size_sub(size_t length) {
-    assert(spinlock_is_locked(&vma_tree_lock));
+    assert(vma_rwlock_is_write_locked(&vma_tree_lock));
     assert(g_total_memory_size >= length);
 
     g_total_memory_size -= length;
@@ -131,29 +178,29 @@ static struct libos_vma* node2vma(struct avl_tree_node* node) {
 }
 
 static struct libos_vma* _get_next_vma(struct libos_vma* vma) {
-    assert(spinlock_is_locked(&vma_tree_lock));
+    assert(vma_rwlock_is_read_locked(&vma_tree_lock) || vma_rwlock_is_write_locked(&vma_tree_lock));
     return node2vma(avl_tree_next(&vma->tree_node));
 }
 
 static struct libos_vma* _get_prev_vma(struct libos_vma* vma) {
-    assert(spinlock_is_locked(&vma_tree_lock));
+    assert(vma_rwlock_is_read_locked(&vma_tree_lock) || vma_rwlock_is_write_locked(&vma_tree_lock));
     return node2vma(avl_tree_prev(&vma->tree_node));
 }
 
 static struct libos_vma* _get_last_vma(void) {
-    assert(spinlock_is_locked(&vma_tree_lock));
+    assert(vma_rwlock_is_read_locked(&vma_tree_lock) || vma_rwlock_is_write_locked(&vma_tree_lock));
     return node2vma(avl_tree_last(&vma_tree));
 }
 
 static struct libos_vma* _get_first_vma(void) {
-    assert(spinlock_is_locked(&vma_tree_lock));
+    assert(vma_rwlock_is_read_locked(&vma_tree_lock) || vma_rwlock_is_write_locked(&vma_tree_lock));
     return node2vma(avl_tree_first(&vma_tree));
 }
 
 /* Returns the vma that contains `addr`. If there is no such vma, returns the closest vma with
  * higher address. */
 static struct libos_vma* _lookup_vma(uintptr_t addr) {
-    assert(spinlock_is_locked(&vma_tree_lock));
+    assert(vma_rwlock_is_read_locked(&vma_tree_lock) || vma_rwlock_is_write_locked(&vma_tree_lock));
 
     struct avl_tree_node* node = avl_tree_lower_bound_fn(&vma_tree, (void*)addr, cmp_addr_to_vma);
     if (!node) {
@@ -176,7 +223,7 @@ typedef bool (*traverse_visitor)(struct libos_vma* vma, void* visitor_arg);
 // TODO: Probably other VMA functions could make use of this helper.
 static bool _traverse_vmas_in_range(uintptr_t begin, uintptr_t end, traverse_visitor visitor,
                                     void* visitor_arg) {
-    assert(spinlock_is_locked(&vma_tree_lock));
+    assert(vma_rwlock_is_read_locked(&vma_tree_lock) || vma_rwlock_is_write_locked(&vma_tree_lock));
     assert(begin <= end);
 
     if (begin == end)
@@ -228,7 +275,7 @@ static void split_vma(struct libos_vma* old_vma, struct libos_vma* new_vma, uint
  */
 static int _vma_bkeep_remove(uintptr_t begin, uintptr_t end, bool is_internal,
                              struct libos_vma** new_vma_ptr, struct libos_vma** vmas_to_free) {
-    assert(spinlock_is_locked(&vma_tree_lock));
+    assert(vma_rwlock_is_write_locked(&vma_tree_lock));
     assert(!new_vma_ptr || *new_vma_ptr);
     assert(IS_ALLOC_ALIGNED_PTR(begin) && IS_ALLOC_ALIGNED_PTR(end));
 
@@ -325,11 +372,11 @@ static void* _vma_malloc(size_t size) {
     if (ret < 0) {
         struct libos_vma* vmas_to_free = NULL;
 
-        spinlock_lock(&vma_tree_lock);
+        vma_rwlock_write_lock(&vma_tree_lock);
         /* Since we are freeing a range we just created, additional vma is not needed. */
         ret = _vma_bkeep_remove((uintptr_t)addr, (uintptr_t)addr + size, /*is_internal=*/true, NULL,
                                 &vmas_to_free);
-        spinlock_unlock(&vma_tree_lock);
+        vma_rwlock_write_unlock(&vma_tree_lock);
         if (ret < 0) {
             log_error("Removing a vma we just created failed: %s", unix_strerror(ret));
             BUG();
@@ -483,7 +530,7 @@ static struct libos_vma* alloc_vma(void) {
             BUG();
         }
 
-        spinlock_lock(&vma_tree_lock);
+        vma_rwlock_write_lock(&vma_tree_lock);
         /* Currently `tmp_vma` is always used (added to `vma_tree`), but this assumption could
          * easily be changed (e.g. if we implement VMAs merging).*/
         struct avl_tree_node* node = &tmp_vma.tree_node;
@@ -493,7 +540,7 @@ static struct libos_vma* alloc_vma(void) {
             avl_tree_swap_node(&vma_tree, node, &vma_migrate->tree_node);
             vma_migrate = NULL;
         }
-        spinlock_unlock(&vma_tree_lock);
+        vma_rwlock_write_unlock(&vma_tree_lock);
 
         if (vma_migrate) {
             free_mem_obj_to_mgr(vma_mgr, vma_migrate);
@@ -541,7 +588,7 @@ static void free_vmas_freelist(struct libos_vma* vma) {
 }
 
 static int _bkeep_initial_vma(struct libos_vma* new_vma) {
-    assert(spinlock_is_locked(&vma_tree_lock));
+    assert(vma_rwlock_is_write_locked(&vma_tree_lock));
 
     struct libos_vma* tmp_vma = _lookup_vma(new_vma->begin);
     if (tmp_vma && tmp_vma->begin < new_vma->end) {
@@ -596,7 +643,7 @@ int init_vma(void) {
     }
     assert(1 + idx == ARRAY_SIZE(init_vmas));
 
-    spinlock_lock(&vma_tree_lock);
+    vma_rwlock_write_lock(&vma_tree_lock);
     int ret = 0;
     /* First of init_vmas is reserved for later usage. */
     for (size_t i = 1; i < ARRAY_SIZE(init_vmas); i++) {
@@ -622,7 +669,7 @@ int init_vma(void) {
         log_debug("Initial VMA region 0x%lx-0x%lx (%s) bookkeeped", init_vmas[i].begin,
                   init_vmas[i].end, init_vmas[i].comment);
     }
-    spinlock_unlock(&vma_tree_lock);
+    vma_rwlock_write_unlock(&vma_tree_lock);
     /* From now on if we return with an error we might leave a structure local to this function in
      * vma_tree. We do not bother with removing them - this is initialization of VMA subsystem, if
      * it fails the whole application startup fails and we should never call any of functions in
@@ -671,6 +718,11 @@ int init_vma(void) {
         return -ENOMEM;
     }
 
+    if (!rwlock_create(&vma_tree_lock)) {
+        return -ENOMEM;
+    }
+    vma_tree_lock_created = true;
+
     /* Now we need to migrate temporary initial vmas. */
     struct libos_vma* vmas_to_migrate_to[ARRAY_SIZE(init_vmas)];
     for (size_t i = 0; i < ARRAY_SIZE(vmas_to_migrate_to); i++) {
@@ -680,7 +732,7 @@ int init_vma(void) {
         }
     }
 
-    spinlock_lock(&vma_tree_lock);
+    vma_rwlock_write_lock(&vma_tree_lock);
     for (size_t i = 0; i < ARRAY_SIZE(init_vmas); i++) {
         /* Skip empty areas. */
         if (init_vmas[i].begin == init_vmas[i].end) {
@@ -690,7 +742,7 @@ int init_vma(void) {
         avl_tree_swap_node(&vma_tree, &init_vmas[i].tree_node, &vmas_to_migrate_to[i]->tree_node);
         vmas_to_migrate_to[i] = NULL;
     }
-    spinlock_unlock(&vma_tree_lock);
+    vma_rwlock_write_unlock(&vma_tree_lock);
 
     for (size_t i = 0; i < ARRAY_SIZE(vmas_to_migrate_to); i++) {
         if (vmas_to_migrate_to[i]) {
@@ -702,7 +754,7 @@ int init_vma(void) {
 }
 
 static void _add_unmapped_vma(uintptr_t begin, uintptr_t end, struct libos_vma* vma) {
-    assert(spinlock_is_locked(&vma_tree_lock));
+    assert(vma_rwlock_is_write_locked(&vma_tree_lock));
 
     vma->begin  = begin;
     vma->end    = end;
@@ -733,7 +785,7 @@ int bkeep_munmap(void* addr, size_t length, bool is_internal, void** tmp_vma_ptr
 
     struct libos_vma* vmas_to_free = NULL;
 
-    spinlock_lock(&vma_tree_lock);
+    vma_rwlock_write_lock(&vma_tree_lock);
     int ret = _vma_bkeep_remove((uintptr_t)addr, (uintptr_t)addr + length, is_internal,
                                 vma2 ? &vma2 : NULL, &vmas_to_free);
     if (ret >= 0) {
@@ -741,7 +793,7 @@ int bkeep_munmap(void* addr, size_t length, bool is_internal, void** tmp_vma_ptr
         *tmp_vma_ptr = (void*)vma1;
         vma1 = NULL;
     }
-    spinlock_unlock(&vma_tree_lock);
+    vma_rwlock_write_unlock(&vma_tree_lock);
 
     free_vmas_freelist(vmas_to_free);
     if (vma1) {
@@ -765,10 +817,10 @@ void bkeep_remove_tmp_vma(void* _vma) {
 
     assert(vma->flags == (VMA_INTERNAL | VMA_UNMAPPED));
 
-    spinlock_lock(&vma_tree_lock);
+    vma_rwlock_write_lock(&vma_tree_lock);
     avl_tree_delete(&vma_tree, &vma->tree_node);
     total_memory_size_sub(vma->end - vma->begin);
-    spinlock_unlock(&vma_tree_lock);
+    vma_rwlock_write_unlock(&vma_tree_lock);
 
     free_vma(vma);
 }
@@ -776,10 +828,10 @@ void bkeep_remove_tmp_vma(void* _vma) {
 void bkeep_convert_tmp_vma_to_user(void* _vma) {
     struct libos_vma* vma = (struct libos_vma*)_vma;
 
-    spinlock_lock(&vma_tree_lock);
+    vma_rwlock_write_lock(&vma_tree_lock);
     assert(vma->flags == (VMA_INTERNAL | VMA_UNMAPPED));
     vma->flags &= ~VMA_INTERNAL;
-    spinlock_unlock(&vma_tree_lock);
+    vma_rwlock_write_unlock(&vma_tree_lock);
 }
 
 static bool is_file_prot_matching(struct libos_handle* file_hdl, int prot) {
@@ -816,7 +868,7 @@ int bkeep_mmap_fixed(void* addr, size_t length, int prot, int flags, struct libo
 
     struct libos_vma* vmas_to_free = NULL;
 
-    spinlock_lock(&vma_tree_lock);
+    vma_rwlock_write_lock(&vma_tree_lock);
     int ret = 0;
     if (flags & MAP_FIXED_NOREPLACE) {
         struct libos_vma* tmp_vma = _lookup_vma(new_vma->begin);
@@ -831,7 +883,7 @@ int bkeep_mmap_fixed(void* addr, size_t length, int prot, int flags, struct libo
         avl_tree_insert(&vma_tree, &new_vma->tree_node);
         total_memory_size_add(new_vma->end - new_vma->begin);
     }
-    spinlock_unlock(&vma_tree_lock);
+    vma_rwlock_write_unlock(&vma_tree_lock);
 
     free_vmas_freelist(vmas_to_free);
     if (vma1) {
@@ -853,7 +905,7 @@ static void vma_update_prot(struct libos_vma* vma, int prot) {
 
 static int _vma_bkeep_change(uintptr_t begin, uintptr_t end, int prot, bool is_internal,
                              struct libos_vma** new_vma_ptr1, struct libos_vma** new_vma_ptr2) {
-    assert(spinlock_is_locked(&vma_tree_lock));
+    assert(vma_rwlock_is_write_locked(&vma_tree_lock));
     assert(IS_ALLOC_ALIGNED_PTR(begin) && IS_ALLOC_ALIGNED_PTR(end));
     assert(begin < end);
 
@@ -982,10 +1034,10 @@ int bkeep_mprotect(void* addr, size_t length, int prot, bool is_internal) {
         return -ENOMEM;
     }
 
-    spinlock_lock(&vma_tree_lock);
+    vma_rwlock_write_lock(&vma_tree_lock);
     int ret = _vma_bkeep_change((uintptr_t)addr, (uintptr_t)addr + length, prot, is_internal, &vma1,
                                 &vma2);
-    spinlock_unlock(&vma_tree_lock);
+    vma_rwlock_write_unlock(&vma_tree_lock);
 
     if (vma1) {
         free_vma(vma1);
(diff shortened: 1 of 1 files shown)


vasanth-intel (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1795)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1795#-NsS6zjq-moWeiefgPFH:-NsS6zjr2LgYvJGZPPOh:bnaxr9k) (no related file):*
We don't see any problems with the PR and it improves perf for some workloads (MySQL and MariaDB) and doesn't degrade perf for other workloads like NginX, Tensorflow, SpecPower, Tensorflow Serving and Openvino(Latency).

___

dimakuv (commented):
> Dismissed @vasanth-intel from a discussion.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1795)* status: 0 of 1 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1795#-NsS6zjq-moWeiefgPFH:-NsSOi_CBdv_K18nk6X7:bvcwdnu) (no related file):*
<details><summary><i>Previously, vasanth-intel wrote…</i></summary><blockquote>

We don't see any problems with the PR and it improves perf for some workloads (MySQL and MariaDB) and doesn't degrade perf for other workloads like NginX, Tensorflow, SpecPower, Tensorflow Serving and Openvino(Latency).
</blockquote></details>

Thank you @vasanth-intel for the performance evaluation!

We're done with internal testing (on the Intel side); this PR is moved from Draft to Ready for Review.

___

dimakuv:
Jenkins, test this please (just for sanity)

kailun-qin (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1795)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`libos/src/bookkeep/libos_vma.c` line 108 at r1](https://reviewable.io/reviews/gramineproject/gramine/1795#-NsDoGxr9hJSdC8sWbo6:-O35fGfO2kUqV8IBXcKa:b-h509g7) ([raw file](https://github.com/gramineproject/gramine/blob/c208c3cd58ab34d36fd7c2d9e2b73e704153d5a7/libos/src/bookkeep/libos_vma.c#L108)):*

It's important to use these wrappers because at the very startup, we don't have the lock because it wasn't yet created. But at startup we have only one thread, so the lock would be redundant anyway.

Note that we can't create the lock as the very first step, because creating the lock itself requires the memory subsystem (VMA) to be fully initialized. So we disable the locking first, init the VMA subsystem, then create the lock and only then the VMA can be used in thread-safe manner.
</blockquote></details>

can we add these explanations into the comments?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1795)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin)

___
*[`libos/src/bookkeep/libos_vma.c` line 108 at r1](https://reviewable.io/reviews/gramineproject/gramine/1795#-NsDoGxr9hJSdC8sWbo6:-O36uBdS9qL1Md4ofVgs:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/c208c3cd58ab34d36fd7c2d9e2b73e704153d5a7/libos/src/bookkeep/libos_vma.c#L108)):*

can we add these explanations into the comments?
</blockquote></details>

Done.

kailun-qin (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1795)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1795)* status: all files reviewed, 1 unresolved discussion, not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1795#-O87QvbsC1kYWF7g1CVw:-O87Qvbt1T85thChW5JV:bfum9as) (no related file):*
@jkr0103 Could you point to the instructions on how to run those MongoDB and/or MySQL experiments that you did?

___

jkr0103:
Instructions for MySQL:
---------------------------------------------------------------
sudo chown -R $USER:$USER /var/lib/mysql-files
sudo systemctl stop mysql.service
sudo mkdir /var/run/mysqld && sudo chown -R $USER:$USER /var/run/mysqld
sudo mkdir /var/run/mysql-data && sudo chown -R $USER:$USER /var/run/mysql-data
mysqld --initialize-insecure --datadir=/var/run/mysql-data

sudo ln -s /etc/apparmor.d/usr.sbin.mysqld /etc/apparmor.d/disable/
sudo apparmor_parser -R /etc/apparmor.d/usr.sbin.mysqld
Check status if MySQL is still loaded: sudo aa-status

sudo vim /etc/security/limits.conf
* soft nofile 65535
* hard nofile 65535

gramine-sgx mysqld --skip-log-bin --datadir /var/run/mysql-data numactl -N 0,1 -l gramine-sgx mysqld --skip-log-bin --datadir /var/run/mysql-data

Sysbench Run:
-------------------------------
sudo apt install -y sysbench
sudo mysqladmin -h 127.0.0.1 -P 3306 create sbtest

sysbench --db-driver=mysql --mysql-host=127.0.0.1 --mysql-port=3306 --mysql-user=root --mysql-db=sbtest --time=90 \
--report-interval=5 oltp_read_write --tables=2 --table_size=100000 prepare

sysbench --db-driver=mysql --mysql-host=127.0.0.1 --mysql-port=3306 --mysql-user=root --mysql-db=sbtest --time=300 \
--report-interval=5 oltp_read_write --tables=2 --table_size=100000 --threads=64 run

======================================================================
MongoDB Run:
-------------------------
git clone https://github.com/mongodb/mongo-perf.git
cd mongo-perf

wget https://repo.mongodb.org/apt/ubuntu/dists/focal/mongodb-org/5.0/multiverse/binary-amd64/mongodb-org-shell_5.0.21_amd64.deb
 
sudo dpkg -i mongodb-org-shell_5.0.21_amd64.deb
 
echo "deb http://security.ubuntu.com/ubuntu focal-security main" | sudo tee /etc/apt/sources.list.d/focal-security.list
sudo apt-get update

sudo apt-get install libssl1.1

gramine-sgx mongod --nounixsocket --dbpath /var/run/db

Benchmark Run:
---------------------
~/examples/mongodb/mongo-perf$ python3 benchrun.py -f testcases/complex_update.js  -t 64

dimakuv (commented):
@jkr0103 Where can we take the MySQL and MongoDB makefiles/manifests for Gramine?

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1795#-:-O87c3ar7I4jS4EQS2Rx:bvte0uh)* status: all files reviewed, 1 unresolved discussion, not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners

jkr0103:
You may find MySQL manifest amd makefiles here https://github.com/gramineproject/examples/pull/28/files#diff-3280189625504cf8039a0453946ea1b585f74319278ffd7abb72e4e71d019f57

For MongoDB:
https://github.com/gramineproject/examples/tree/master/mongodb

@vasanth-intel Please update here if it differ in your setup.

chiache (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1795)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`libos/src/bookkeep/libos_vma.c` line 144 at r3](https://reviewable.io/reviews/gramineproject/gramine/1795#-OF8cl-hDkXehpnaSO11:-OF8cl-hDkXehpnaSO12:by5gg4e) ([raw file](https://github.com/gramineproject/gramine/blob/e5c76915c19cb87224f9b1763b616bcced8df68e/libos/src/bookkeep/libos_vma.c#L144)):*
> ```C
>     rwlock_write_unlock(l);
> }
> ```

Since only `vma_tree_lock` will be locked and unlocked here, why not remove the parameter to these functions and dedicate them to `vma_tree_lock`?
___
*[`libos/src/bookkeep/libos_vma.c` line 149 at r3](https://reviewable.io/reviews/gramineproject/gramine/1795#-OF8c6_C3iDJ5Dhi1xDA:-OF8c6_C3iDJ5Dhi1xDB:bb5849m) ([raw file](https://github.com/gramineproject/gramine/blob/e5c76915c19cb87224f9b1763b616bcced8df68e/libos/src/bookkeep/libos_vma.c#L149)):*
> ```C
>     if (!vma_tree_lock_created)
>         return true;
>     return rwlock_is_read_locked(l);
> ```

To me, if a thread is holding the write lock, it should be able to perform a read-only operation atomically. So, IMO, the condition should be `rwlock_is_read_locked(l) or rwlock_is_write_locked(l). 
___
*[`libos/src/bookkeep/libos_vma.c` line 158 at r3](https://reviewable.io/reviews/gramineproject/gramine/1795#-OF8_r0VBx9UcJ-msMzp:-OF8_r0VBx9UcJ-msMzq:bjau7vp) ([raw file](https://github.com/gramineproject/gramine/blob/e5c76915c19cb87224f9b1763b616bcced8df68e/libos/src/bookkeep/libos_vma.c#L158)):*
> ```C
> }
> #endif
> ```

Shouldn't `vma_rwlock_is_read_locked` and `vma_rwlock_is_write_locked` be defined even if `DEBUG` is not?

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1795)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @chiache and @dimakuv)

___
*[`libos/src/bookkeep/libos_vma.c` line 149 at r3](https://reviewable.io/reviews/gramineproject/gramine/1795#-OF8c6_C3iDJ5Dhi1xDA:-OFSMwzpDpVZbRY_637T:b-9h6727) ([raw file](https://github.com/gramineproject/gramine/blob/e5c76915c19cb87224f9b1763b616bcced8df68e/libos/src/bookkeep/libos_vma.c#L149)):*

To me, if a thread is holding the write lock, it should be able to perform a read-only operation atomically. So, IMO, the condition should be `rwlock_is_read_locked(l) or rwlock_is_write_locked(l).
</blockquote></details>

If we decide so, then the function name will require changing, otherwise it's misleading.
___
*[`libos/src/bookkeep/libos_vma.c` line 158 at r3](https://reviewable.io/reviews/gramineproject/gramine/1795#-OF8_r0VBx9UcJ-msMzp:-OFSN3hh7tv3vDb7cmE3:b-z8gr7x) ([raw file](https://github.com/gramineproject/gramine/blob/e5c76915c19cb87224f9b1763b616bcced8df68e/libos/src/bookkeep/libos_vma.c#L158)):*

Shouldn't `vma_rwlock_is_read_locked` and `vma_rwlock_is_write_locked` be defined even if `DEBUG` is not?
</blockquote></details>

No, these functions are inherently race'y and should be used only inside asserts (i.e. only in debug builds), I don't think there's any legitimate production use for them.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1795)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @chiache and @mkow)

___
*[`libos/src/bookkeep/libos_vma.c` line 144 at r3](https://reviewable.io/reviews/gramineproject/gramine/1795#-OF8cl-hDkXehpnaSO11:-OFzt74W1fSr2tzUEXnp:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/e5c76915c19cb87224f9b1763b616bcced8df68e/libos/src/bookkeep/libos_vma.c#L144)):*

Since only `vma_tree_lock` will be locked and unlocked here, why not remove the parameter to these functions and dedicate them to `vma_tree_lock`?
</blockquote></details>

Done.
___
*[`libos/src/bookkeep/libos_vma.c` line 149 at r3](https://reviewable.io/reviews/gramineproject/gramine/1795#-OF8c6_C3iDJ5Dhi1xDA:-OFzt8Q-7FoWsjcKmbQC:b-38fqb0) ([raw file](https://github.com/gramineproject/gramine/blob/e5c76915c19cb87224f9b1763b616bcced8df68e/libos/src/bookkeep/libos_vma.c#L149)):*

If we decide so, then the function name will require changing, otherwise it's misleading.
</blockquote></details>

pls check if this is something you'd expect

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1795)* status: 0 of 1 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @chiache and @mkow)

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1795#-OFzvv2M3QliakZcfrtl:-OFzvv2M3QliakZcfrtm:bno2nhx) (no related file):*
I have to rebase to revive the CI (hopefully this won't disrupt your reviews).

efu39 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1795)* status: 0 of 1 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @chiache, @dimakuv, @kailun-qin, and @mkow)

___
*[`libos/src/bookkeep/libos_vma.c` line 697 at r6](https://reviewable.io/reviews/gramineproject/gramine/1795#-OG0eSzY5haLtyn_YeDd:-OG0eSzY5haLtyn_YeDe:b-n1zpqa) ([raw file](https://github.com/gramineproject/gramine/blob/50d524641fe76fafb198eb997abb093e5d769e3e/libos/src/bookkeep/libos_vma.c#L697)):*
> ```C
>     assert(1 + idx == ARRAY_SIZE(init_vmas));
> 
>     vma_rwlock_write_lock();
> ```

rwlock_create() is not yet done here. 



_Code quote:_
```C
vma_rwlock_write_lock();
```
___
*[`libos/src/bookkeep/libos_vma.c` line 775 at r6](https://reviewable.io/reviews/gramineproject/gramine/1795#-OG0d3sl8Dc1cRMv2SlM:-OG0d3sl8Dc1cRMv2SlN:b-ep35y0) ([raw file](https://github.com/gramineproject/gramine/blob/50d524641fe76fafb198eb997abb093e5d769e3e/libos/src/bookkeep/libos_vma.c#L775)):*
> ```C
>         return -ENOMEM;
>     }
>     vma_tree_lock_created = true;
> ```

Should these be moved to front before line 697 where vma_rwlock_write_lock() is invoked earlier?



_Code quote:_
```C
    if (!rwlock_create(&vma_tree_lock)) {
        return -ENOMEM;
    }
    vma_tree_lock_created = true;
```

kailun-qin (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1795)* status: 0 of 1 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @chiache, @dimakuv, @efu39, and @mkow)

___
*[`libos/src/bookkeep/libos_vma.c` line 697 at r6](https://reviewable.io/reviews/gramineproject/gramine/1795#-OG0eSzY5haLtyn_YeDd:-OG3rXDY8oI9v1zJdZq5:b-as8xjo) ([raw file](https://github.com/gramineproject/gramine/blob/50d524641fe76fafb198eb997abb093e5d769e3e/libos/src/bookkeep/libos_vma.c#L697)):*
<details><summary><i>Previously, efu39 (Erica Fu) wrote…</i></summary><blockquote>

rwlock_create() is not yet done here.
</blockquote></details>

Yes, I think this is what Dmitrii's comments were describing: https://github.com/gramineproject/gramine/blob/50d524641fe76fafb198eb997abb093e5d769e3e/libos/src/bookkeep/libos_vma.c#L114-L121

Well, upon rereading this, I don't understand the limitation and don't recall why we cannot create `vma_tree_lock` at the very beginning of the LibOS startup or why creating this lock requires the memory subsystem to be fully initialized. I'll need to double-check.

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1795)* status: 0 of 1 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @chiache, @efu39, and @mkow)

___
*[`libos/src/bookkeep/libos_vma.c` line 697 at r6](https://reviewable.io/reviews/gramineproject/gramine/1795#-OG0eSzY5haLtyn_YeDd:-OLhWGbE9GXcK_e9Q6IS:bquh2nb) ([raw file](https://github.com/gramineproject/gramine/blob/50d524641fe76fafb198eb997abb093e5d769e3e/libos/src/bookkeep/libos_vma.c#L697)):*

Yes, I think this is what Dmitrii's comments were describing: https://github.com/gramineproject/gramine/blob/50d524641fe76fafb198eb997abb093e5d769e3e/libos/src/bookkeep/libos_vma.c#L114-L121

Well, upon rereading this, I don't understand the limitation and don't recall why we cannot create `vma_tree_lock` at the very beginning of the LibOS startup or why creating this lock requires the memory subsystem to be fully initialized. I'll need to double-check.
</blockquote></details>

I rechecked and didn't get the limitation. There seemed to be no issues during testing when creating the lock early enough in the init. Let me know if I misunderstood anything.
___
*[`libos/src/bookkeep/libos_vma.c` line 775 at r6](https://reviewable.io/reviews/gramineproject/gramine/1795#-OG0d3sl8Dc1cRMv2SlM:-OLhVujF7SOnHC_tQOOp:bwd8vv8) ([raw file](https://github.com/gramineproject/gramine/blob/50d524641fe76fafb198eb997abb093e5d769e3e/libos/src/bookkeep/libos_vma.c#L775)):*
<details><summary><i>Previously, efu39 (Erica Fu) wrote…</i></summary><blockquote>

Should these be moved to front before line 697 where vma_rwlock_write_lock() is invoked earlier?
</blockquote></details>

Not relevant any more.

kailun-qin (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1795)* status: 0 of 1 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @chiache, @efu39, and @mkow)

___
*[`libos/src/bookkeep/libos_vma.c` line 697 at r6](https://reviewable.io/reviews/gramineproject/gramine/1795#-OG0eSzY5haLtyn_YeDd:-OLhousc5LHs9WWUTEN7:bftvnhz) ([raw file](https://github.com/gramineproject/gramine/blob/50d524641fe76fafb198eb997abb093e5d769e3e/libos/src/bookkeep/libos_vma.c#L697)):*

I rechecked and didn't get the limitation. There seemed to be no issues during testing when creating the lock early enough in the init. Let me know if I misunderstood anything.
</blockquote></details>

Ok, it turns out my understanding is wrong (though I can't reproduce on my setup); I've reverted back to the previous version.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1795)* status: all files reviewed, 10 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @chiache, @dimakuv, @efu39, and @kailun-qin)

___
*[`libos/src/bookkeep/libos_vma.c` line 106 at r1](https://reviewable.io/reviews/gramineproject/gramine/1795#-NsDorcc6TUTHDb0sio8:-OPQv3Yn0p0otypiTkgQ:b-c84fc9) ([raw file](https://github.com/gramineproject/gramine/blob/c208c3cd58ab34d36fd7c2d9e2b73e704153d5a7/libos/src/bookkeep/libos_vma.c#L106)):*

Technically all these variables must have the prefix `g_`. But I didn't want to add this unrelated change in this PR.
</blockquote></details>

Please add `g_` to the global variables refactored in this PR (at least `vma_tree_lock` and `vma_tree_lock_created`).
___
*[`libos/src/bookkeep/libos_vma.c` line 697 at r6](https://reviewable.io/reviews/gramineproject/gramine/1795#-OG0eSzY5haLtyn_YeDd:-OPR1i8H5IccD6mX_Zi-:brfw8x9) ([raw file](https://github.com/gramineproject/gramine/blob/50d524641fe76fafb198eb997abb093e5d769e3e/libos/src/bookkeep/libos_vma.c#L697)):*

Ok, it turns out my understanding is wrong (though I can't reproduce on my setup); I've reverted back to the previous version.
</blockquote></details>

But the code up to the point of `rwlock_create(&vma_tree_lock)` doesn't need to take the lock? Could we just remove the lock calls from it? This way we wouldn't need this weird and rather dangerous semantics, that if the lock wasn't initialized then everything succeeds silently, without actually taking it.
In case there are some function which we really need to use which have assert, then we could split them into 2, one checking the lock and calling the other, and the other just doing the implementation without checking the lock.
___
*[`libos/src/bookkeep/libos_vma.c` line 123 at r8](https://reviewable.io/reviews/gramineproject/gramine/1795#-OPR2XFG9vZ-gAB7RHIE:-OPR2XFG9vZ-gAB7RHIF:b-weiaf) ([raw file](https://github.com/gramineproject/gramine/blob/3a0be69ebce839fd98cb0a2c92a450eb6500a24c/libos/src/bookkeep/libos_vma.c#L123)):*
> ```C
>  * the VMA subsystem can be used in thread-safe manner.
>  */
> static inline void vma_rwlock_read_lock(void) {
> ```

I'd drop `inline` and let the compiler decide what's better.



_Code quote:_
```C
static inline
```
___
*[`libos/src/bookkeep/libos_vma.c` line 124 at r8](https://reviewable.io/reviews/gramineproject/gramine/1795#-OPR2KQ-3H2_jNLcZyiT:-OPR2KQ-3H2_jNLcZyiU:bhovv4r) ([raw file](https://github.com/gramineproject/gramine/blob/3a0be69ebce839fd98cb0a2c92a450eb6500a24c/libos/src/bookkeep/libos_vma.c#L124)):*
> ```C
>  */
> static inline void vma_rwlock_read_lock(void) {
>     if (!vma_tree_lock_created)
> ```

This should be `assert(vma_tree_lock_created)` after applying my suggestion from below.
___
*[`libos/src/bookkeep/libos_vma.c` line 1720 at r8](https://reviewable.io/reviews/gramineproject/gramine/1795#-OPR83Ya3VOfl10z5hec:-OPR83Ya3VOfl10z5hed:b24ikeb) ([raw file](https://github.com/gramineproject/gramine/blob/3a0be69ebce839fd98cb0a2c92a450eb6500a24c/libos/src/bookkeep/libos_vma.c#L1720)):*
> ```C
> /* returns whether prot_refresh_vma() must be applied on a VMA */
> static bool vma_update_valid_end(struct libos_vma* vma, void* _args) {
>     assert(vma_rwlock_is_read_or_write_locked());
> ```

Hmm, I think this actually modifies VMAs, not only reads?
```C
vma->valid_end = vma->begin + valid_length;
```

It seems to be a bug in this PR, because `dump_vmas_with_buf()` only takes a _read_ lock, not write? It takes a read-only lock, but passes non-const `struct libos_vma*` to the visitor function. It should either be read-only and pass `const struct libos_vma*`, or take a read-write lock and then pass `struct libos_vma*`.

kailun-qin (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1795)* status: 0 of 1 files reviewed, 10 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @chiache, @dimakuv, @efu39, and @mkow)

___
*[`libos/src/bookkeep/libos_vma.c` line 106 at r1](https://reviewable.io/reviews/gramineproject/gramine/1795#-NsDorcc6TUTHDb0sio8:-OPdrDlt2lEkydYWM6nG:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/c208c3cd58ab34d36fd7c2d9e2b73e704153d5a7/libos/src/bookkeep/libos_vma.c#L106)):*

Please add `g_` to the global variables refactored in this PR (at least `vma_tree_lock` and `vma_tree_lock_created`).
</blockquote></details>

Done.
___
*[`libos/src/bookkeep/libos_vma.c` line 697 at r6](https://reviewable.io/reviews/gramineproject/gramine/1795#-OG0eSzY5haLtyn_YeDd:-OPdrp0D9bSaVMZnQFDN:b-rcuob8) ([raw file](https://github.com/gramineproject/gramine/blob/50d524641fe76fafb198eb997abb093e5d769e3e/libos/src/bookkeep/libos_vma.c#L697)):*

But the code up to the point of `rwlock_create(&vma_tree_lock)` doesn't need to take the lock? Could we just remove the lock calls from it? This way we wouldn't need this weird and rather dangerous semantics, that if the lock wasn't initialized then everything succeeds silently, without actually taking it.
In case there are some function which we really need to use which have assert, then we could split them into 2, one checking the lock and calling the other, and the other just doing the implementation without checking the lock.
</blockquote></details>

I like the idea, but removing lock calls from some functions may not be that straightforward?
E.g., `create_mem_mgr()` (before `rwlock_create(&vma_tree_lock)` and thus doesn't need to take the lock): https://github.com/gramineproject/gramine/blob/6bde81410d399bee26727631e2bdab0cce707a08/common/include/memmgr.h#L136-L137
would call into `bkeep_mmap_any_in_range()` where the lock call inside is hard to untangle?
https://github.com/gramineproject/gramine/blob/6bde81410d399bee26727631e2bdab0cce707a08/libos/src/bookkeep/libos_vma.c#L352-L357
https://github.com/gramineproject/gramine/blob/6bde81410d399bee26727631e2bdab0cce707a08/libos/src/bookkeep/libos_vma.c#L1116
___
*[`libos/src/bookkeep/libos_vma.c` line 123 at r8](https://reviewable.io/reviews/gramineproject/gramine/1795#-OPR2XFG9vZ-gAB7RHIE:-OPdrIK4AT_FcWK_TYYI:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/3a0be69ebce839fd98cb0a2c92a450eb6500a24c/libos/src/bookkeep/libos_vma.c#L123)):*

I'd drop `inline` and let the compiler decide what's better.
</blockquote></details>

Done.
___
*[`libos/src/bookkeep/libos_vma.c` line 1720 at r8](https://reviewable.io/reviews/gramineproject/gramine/1795#-OPR83Ya3VOfl10z5hec:-OPdrav3-vGVxJXlgrzA:bpvzhvz) ([raw file](https://github.com/gramineproject/gramine/blob/3a0be69ebce839fd98cb0a2c92a450eb6500a24c/libos/src/bookkeep/libos_vma.c#L1720)):*

Hmm, I think this actually modifies VMAs, not only reads?
```C
vma->valid_end = vma->begin + valid_length;
```

It seems to be a bug in this PR, because `dump_vmas_with_buf()` only takes a _read_ lock, not write? It takes a read-only lock, but passes non-const `struct libos_vma*` to the visitor function. It should either be read-only and pass `const struct libos_vma*`, or take a read-write lock and then pass `struct libos_vma*`.
</blockquote></details>

Good catch! Updated.

kailun-qin (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1795)* status: 0 of 1 files reviewed, 10 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @chiache, @dimakuv, @efu39, and @mkow)

___
*[`libos/src/bookkeep/libos_vma.c` line 697 at r6](https://reviewable.io/reviews/gramineproject/gramine/1795#-OG0eSzY5haLtyn_YeDd:-OPdz81RB28xmS4-AeqH:b-50jdby) ([raw file](https://github.com/gramineproject/gramine/blob/50d524641fe76fafb198eb997abb093e5d769e3e/libos/src/bookkeep/libos_vma.c#L697)):*

I like the idea, but removing lock calls from some functions may not be that straightforward?
E.g., `create_mem_mgr()` (before `rwlock_create(&vma_tree_lock)` and thus doesn't need to take the lock): https://github.com/gramineproject/gramine/blob/6bde81410d399bee26727631e2bdab0cce707a08/common/include/memmgr.h#L136-L137
would call into `bkeep_mmap_any_in_range()` where the lock call inside is hard to untangle?
https://github.com/gramineproject/gramine/blob/6bde81410d399bee26727631e2bdab0cce707a08/libos/src/bookkeep/libos_vma.c#L352-L357
https://github.com/gramineproject/gramine/blob/6bde81410d399bee26727631e2bdab0cce707a08/libos/src/bookkeep/libos_vma.c#L1116
</blockquote></details>

Besides, could you pls remind me why there's a limitation that "creating this lock itself requires the memory subsystem (VMA) to be fully initialized"?

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1795)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @chiache, @dimakuv, @efu39, and @kailun-qin)

___
*[`libos/src/bookkeep/libos_vma.c` line 697 at r6](https://reviewable.io/reviews/gramineproject/gramine/1795#-OG0eSzY5haLtyn_YeDd:-OPfOEal9XhtM2mk_EDw:bngjlop) ([raw file](https://github.com/gramineproject/gramine/blob/50d524641fe76fafb198eb997abb093e5d769e3e/libos/src/bookkeep/libos_vma.c#L697)):*
> Besides, could you pls remind me why there's a limitation that "creating this lock itself requires the memory subsystem (VMA) to be fully initialized"?

No idea, it's you who said it doesn't work if you move the creation to the earlier place? `_PalEventCreate()` uses `calloc()`, but I'm don't remember which allocator it uses, maybe this one and that's the problem?

donporter (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1795)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @chiache, @dimakuv, @efu39, @kailun-qin, and @mkow)

___
*[`libos/src/bookkeep/libos_vma.c` line 697 at r6](https://reviewable.io/reviews/gramineproject/gramine/1795#-OG0eSzY5haLtyn_YeDd:-ORH8X4r8-sz-vhxiIJq:b4ll8j9) ([raw file](https://github.com/gramineproject/gramine/blob/50d524641fe76fafb198eb997abb093e5d769e3e/libos/src/bookkeep/libos_vma.c#L697)):*

> Besides, could you pls remind me why there's a limitation that "creating this lock itself requires the memory subsystem (VMA) to be fully initialized"?

No idea, it's you who said it doesn't work if you move the creation to the earlier place? `_PalEventCreate()` uses `calloc()`, but I'm don't remember which allocator it uses, maybe this one and that's the problem?
</blockquote></details>

FWIW, I don't see this dependency either.  What happens if you just try to create/initialize the lock here?

PR #1796, opened 2024-03-07
[PAL/Linux-SGX] Fix Trusted Files degenerating to Allowed Files on fork

dimakuv:
## Description of the changes 

Previously, Trusted Files feature had the following bug: after fork, the metadata of currently-opened-in-parent-process TFs (SHA256 hashes for each chunk of the file) was not available in the child SGX enclave. This effectively degenerated all currently-opened TFs into Allowed Files, and thus the child enclave lost integrity guarantees in these TFs. An attacker could substitute the TF contents on the storage with arbitrary code/data, and the child enclave would not notice this.

This bug was exposed only in the child process (child SGX enclave) and only in these scenarios:
- fork without execve: the child continues reading from one of the file descriptors pointing to a trusted file opened before `fork()`;
- fork with execve: the child reads from an inherited file descriptor pointing to a trusted file.

This PR also adds a test to verify this bug is fixed.

## How to test this PR? 

LibOS test (based on a GDB script) is added.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1796)

### DIFF
--- a/common/include/arch/x86_64/cpu.h
+++ b/common/include/arch/x86_64/cpu.h
@@ -102,7 +102,10 @@ static inline void wrfsbase(uint64_t addr) {
         :: "D"(addr) : "memory");
 }
 
-static inline noreturn void die_or_inf_loop(void) {
+/* This function must be breakable for debugging and GDB integration scripts (e.g. see
+ * `fork_and_access_file.gdb` in LibOS regression tests). */
+__attribute__((__noinline__)) __attribute__((unused))
+static noreturn void die_or_inf_loop(void) {
     __asm__ volatile (
         "1: \n"
         "ud2 \n"
--- a/libos/test/regression/fork_and_access_file.c
+++ b/libos/test/regression/fork_and_access_file.c
@@ -1,0 +1,46 @@
+/* SPDX-License-Identifier: LGPL-3.0-or-later */
+/* Copyright (C) 2024 Intel Corporation */
+
+#include <err.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/wait.h>
+#include <unistd.h>
+
+#include "common.h"
+#include "rw_file.h"
+
+#define FILENAME     "fork_and_access_file_testfile"
+#define MAX_BUF_SIZE 256
+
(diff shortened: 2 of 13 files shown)


kailun-qin (dismissed):
> Reviewed 13 of 13 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1796)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1796)* status: 11 of 13 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @kailun-qin)

___
*[`pal/src/host/linux-sgx/pal_host.h` line 56 at r3](https://reviewable.io/reviews/gramineproject/gramine/1796#-NsN_Mc88XLC5Fs-G7KL:-NsN_Mc88XLC5Fs-G7KM:bqxhbzg) ([raw file](https://github.com/gramineproject/gramine/blob/d82e885cb4adbcec876c78a0f745559cdb26aee4/pal/src/host/linux-sgx/pal_host.h#L56)):*
> ```objc
>             /* below fields are used only for trusted files */
>             sgx_chunk_hash_t* chunk_hashes; /* array of hashes of file chunks */
>             void* umem;                     /* valid only when chunk_hashes != NULL and size > 0 */
> ```

In a previous iteration, I forgot that `umem` can be legitimately NULL if the underlying Trusted File is empty (then it cannot be mmapped, since there are no contents to mmap).

My misunderstanding led to me checking `assert(handle->file.umem)` everywhere. And on our Python tests it failed, because Python has some trusted files like `__init__.py` which are empty. So I fixed my asserts to only check `umem` when necessary.

I expanded the comment to not forget about this in the future.

mkow (approved):
> Reviewed 10 of 13 files at r1, 1 of 1 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1796)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

Issue #466, opened 2022-03-21
[LibOS] time zone is incorrect in syscall `gettimeofday()` or glibc `localtime()` or bash command `date`


llly:
## Description of the problem
Current Gramine doesn't take care of time zone. The requests for time with time zone is incorrect and inconsistent. 

## Steps to reproduce
1. Choose time zone using `timedatectl set-timezone` on Ubuntu.
2.  execute sample 
```
time_t timep;
time (&timep);
printf("%s\n",asctime(localtime(&timep)));

struct timeval tv;
struct timezone tz= {60, 1};
gettimeofday (&tv , &tz);
printf("tz_minuteswest: %d\n", tz.tz_minuteswest);
```
3. execute `bash -c date` in CI-Example.

## Expected results
```
Mon Mar 21 14:39:45 2022
tz_minuteswest: -480
Mon Mar 21 14:40:21 CST 2022
```
## Actual results

```
Mon Mar 21 06:40:45 2022
tz_minuteswest: 60
Mon Mar 21 06:41:41 UTC 2022
```
## Additional information
We can see that  all time become UTC time, while `tz_minuteswest` doesn't reflect UTC time.
## Gramine commit hash
e37f773d4d910799a854b90334f2ff8eb6f411bd

lejunzhu:
If you add this to CI-Example bash manifest.template, the result is correct:

```
...
loader.env.TZ = ":/etc/localtime"
...
fs.mount.etc.type = "chroot"
fs.mount.etc.path = "/etc"
fs.mount.etc.uri = "file:/etc"
...
sgx.allowed_files = [
...
  "file:/etc/localtime",
]
```

Similar happens to the C example. 

However, tz_minuteswest is always 60. Not sure why.

lejunzhu:
Looking at shim_do_gettimeofday(), tz is never touched, hence the initial value 60 is always there. Is it because this argument is obsolete or a bug?

llly:
>`loader.env.TZ = ":/etc/localtime"` 

Yes, The workaround works for localtime and date. Native app doesn't need a TZ env to get time zone, OS doesn't provide TZ env, neither.

lejunzhu:
> > `loader.env.TZ = ":/etc/localtime"`
> 
> Yes, The workaround works for localtime and date. Native app doesn't need a TZ env to get time zone, OS doesn't provide TZ env, neither.

TZ env is only needed for the bash -c case. C sample code runs fine without it.

[...]

PR #1834, opened 2024-04-02
Pal sgx fast clock gettimeofday

### LINKED ISSUE #1799: [PAL-SGX] alternative gettimeofday() implementation for gramine

Looking at the gettimeofday() implementation used by gramine, there are many similarities with an implementation we did for "native" sgx apps that we develop where I work. But, we also took some different approaches listed below.

I have working and highly tested code I can share with you, if you're interested in my implementation. It's currently written in c++ but if there's interest I will be happy to port it to c and add it to gramine.

- our implementation also uses RDTSC when it's available, and falls back to OCALL if there's no other choice.
- we don't rely on cpuid heuristics to read the clock frequency (gramine has bare-metal and hypervisor implementation, which is partial).
- instead, we calculate the clock frequency ourselves.
  - for this we need two "ground truth" readings of timeval + tsc pairs, and we can calculate the clock frequency by counting "how many cycles passed in a given time".
  - we don't need to "sleep" for this to work, we use consecutive calls to gettimeofday() and the time elapsed between them to make the calculation.
  - the clock frequency is periodically refreshed and recalibrated (by performing 2 "extra" ocalls that are far enough apart), in order to keep close to the host's time.
- the code was written to give consistent time between all threads in a thread safe and lockless manner (using atomics)
   - there is a "pair" of global ground-truth states which we switch between in round-robin
   - a single thread gets to "update and advance" the state

This code was stressed and tested on azure, I know in the past gramine had some issues with the fast-path not working on azure?

Please let me know if this is interesting to you.

1. do not rely on "fixed" clock_frequency of the cpu and different hypervisor implementations for reporting it in cpuid
2. there are probably some targets that *should* support this fast-path gettimeofday() using rdtsc, but currently use the slow path
3. this implementation is completely lockless - gramine currently uses a spin lock to update the ground truth.

Out implementation is measured to run in ~47 cycles (with -O3 compile flag) which is faster than the ~60 cycles it takes to run gettimeofday() natively on the host!

Also, as an unrelated note, we do handle getting the (deprecated) timezone struct when re\calibrating our ground-truth, just in case some old code uses this parameter. I saw an issue regarding this still open on gramine.

jonathan-sha:
## Description of the changes 

This is a re-implementation of _palSystemTimeQuery in a way that doesn't require steady TSC feature and works on previously unsupported hypervisors.

See full explanation and discussion in #1799

Fixes #1799

## How to test this PR? 

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1834)


pal/src/host/linux-sgx/utils/fast_clock.c
```diff
@@ -0,0 +1,362 @@
...
+    fast_clock_desc_t disabled_desc = advance_state(descriptor, FC_STATE_RDTSC_DISABLED, false);
+    desc_atomic_store(fast_clock, disabled_desc, __ATOMIC_RELEASE);
+}
+
+static inline fast_clock_timepoint_t* get_timepoint(fast_clock_t* fast_clock, fast_clock_desc_t descriptor)
+{
+    return &fast_clock->time_points[timepoint_index(descriptor)];
+}
+
+static bool is_rdtsc_available(void) {
```
jonathan-sha, line 226: Not sure if this method is needed - can we rely on the RDTSC emulation feature?

Maybe we can do the following - have a global `g_is_rdtsc_emulated` which is set under the `FIRST_TIME()` guard in the emulator. This method will simply call `get_tsc()` then check the global. We can then completely remove this logic from `pal_sgx_main`.

pal/src/host/linux-sgx/utils/fast_clock.h
```diff
@@ -0,0 +1,79 @@
...
+typedef struct fast_clock_s
+{
+    fast_clock_desc_t atomic_descriptor;
+    fast_clock_timepoint_t time_points[FC_FLAGS_NUM_TIMEPOINTS];
+} fast_clock_t;
+
+extern fast_clock_t g_fast_clock;
+
+int fast_clock__get_time(fast_clock_t* fast_clock, uint64_t* time_micros, bool force_new_timepoint);
+void fast_clock__get_timezone(const fast_clock_t* fast_clock, int* tz_minutewest, int* tz_dsttime);
```
jonathan-sha, line 71: missing implementation

pal/src/host/linux-sgx/utils/fast_clock.h
```diff
@@ -0,0 +1,79 @@
...
+bool fast_clock__is_enabled(const fast_clock_t* fast_clock);
+void fast_clock__disable(fast_clock_t* fast_clock);
+
+#ifdef __cplusplus
+} /* extern int */
```
jonathan-sha, line 76: typo - extern "C"

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1834)* status: 0 of 11 files reviewed, 28 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @jonathan-sha)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui6Zr1-2fOipiHl49G:-Nui6Zr1-2fOipiHl49H:b-xxq0e3) (no related file):*
I just briefly looked at the PR and left some comments. Didn't go deep into the logic itself.

Please fix according to the comments, and then we could start the review in earnest.

___
___
*[`pal/src/host/linux-sgx/enclave_ocalls.h` line 92 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui41nP8JYuBwW_Gw4q:-Nui41nQ0SG_Ff-3mTHk:bb288uj) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/enclave_ocalls.h#L92)):*
> ```objc
> int ocall_futex(uint32_t* uaddr, int op, int val, uint64_t* timeout_us);
> 
> int ocall_reset_time(uint64_t* microsec, uint64_t* tsc, int* tz_minuteswest, int* tz_dsttime);
> ```

Can we remove the `tz_` args and handling from this PR? You don't use them here anyway, and it's harder to review.
___
*[`pal/src/host/linux-sgx/enclave_ocalls.c` line 1810 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui48m_B9uyg7IjNUrN:-Nui48m_B9uyg7IjNUrO:b-o3dehg) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/enclave_ocalls.c#L1810)):*
> ```C
>         *microsec_ptr = MAX(microsec, expected_microsec);
>         if (tsc_ptr != NULL) {
>             *tsc_ptr = COPY_UNTRUSTED_VALUE(&ocall_gettime_args->tsc);
> ```

Do you have some sanity checks on the returned-by-host `tsc` value? It must be at least non-decreasing. We have a similar check for returned-by-host `microsec`. This is to prevent attacks.
___
*[`pal/src/host/linux-sgx/enclave_ocalls.c` line 1825 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui4Ils6eTgvz6LNCQs:-Nui4Ils6eTgvz6LNCQt:b-8slbc3) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/enclave_ocalls.c#L1825)):*
> ```C
> {
>     return ocall_reset_time(microsec_ptr, NULL, NULL, NULL);
> }
> ```

That's a bad security practice -- you added an additional attack surface (a new OCALL) for no good reason. Just expand existing `ocall_gettime` to have an additional argument `tsc_ptr`, and remove `ocall_reset_time()`.
___
*[`pal/src/host/linux-sgx/host_ocalls.c` line 608 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui3gtW3bSvs7AXM0sE:-Nui3gtW3bSvs7AXM0sF:bdqn4ji) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/host_ocalls.c#L608)):*
> ```C
>     struct timeval tv;
>     struct timezone tz;
>     unsigned long long tsc = __rdtsc();
> ```

We have `get_tsc()` which does the same. Please remove `x86intrin.h`
___
*[`pal/src/host/linux-sgx/pal_main.c` line 561 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui4c8V6o2Vdkf76zap:-Nui4c8V6o2Vdkf76zaq:bn8xn5p) ([raw file](https://github.com/gramineproject/gramine/blob/0ee48eebb24c38bdad193fa9945a3b26d4a89dc6/pal/src/host/linux-sgx/pal_main.c#L561)):*
> ```C
>                     "not expose corresponding CPUID leaves). This degrades performance.");
>     }
> }
> ```

Why did you remove `print_warning_on_invariant_tsc()`?
___
*[`pal/src/host/linux-sgx/pal_misc.c` line 319 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui4goHAXmlioS0i3zV:-Nui4goHAXmlioS0i3zW:b4cbi4m) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/pal_misc.c#L319)):*
> ```C
>     /* hypervisor-specific CPUID leaf functions (0x40000000 - 0x400000FF) start here */
>     {.leaf = 0x40000000, .zero_subleaf = true, .cache = true},  /* CPUID Info */
>     {.leaf = 0x40000010, .zero_subleaf = true, .cache = true},  /* VMWare-style Timing Info */
> ```

These 2 leaves won't be needed now, right? We can remove them.
___
*[`pal/src/host/linux-sgx/utils/fast_clock.h` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui4po26N4JaaLnopTS:-Nui4po26N4JaaLnopTT:b-1did7g) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.h#L2)):*
> ```objc
> #ifndef _FAST_CLOCK_H_
> #define _FAST_CLOCK_H_
> ```

We use `#pragma once`
___
*[`pal/src/host/linux-sgx/utils/fast_clock.h` line 9 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui4n4pEh8cHWb1tQWx:-Nui4n4pEh8cHWb1tQWy:bf00apt) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.h#L9)):*
> ```objc
> #ifdef __cplusplus
> extern "C" {
> #endif
> ```

There is nothing C++ in our codebase, please remove these guards.
___
*[`pal/src/host/linux-sgx/utils/fast_clock.h` line 13 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui4ux38R1aRsxtYqfE:-Nui4ux38R1aRsxtYqfF:b-ffh4s4) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.h#L13)):*
> ```objc
> 
> enum fast_clock_state_e
> {
> ```

We have a different coding style (based on Google C/C++ coding style), see https://gramine.readthedocs.io/en/stable/devel/coding-style.html
___
*[`pal/src/host/linux-sgx/utils/fast_clock.h` line 19 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui542W86c5YrrS93qN:-Nui542W86c5YrrS93qO:bjt93fs) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.h#L19)):*
> ```objc
>     FC_STATE_INIT,
> 
>     FC_STATE_RDTSC_DISABLED,
> ```

I don't like the `FC_`, it's not immediately obvious what it is.

Hm, is this enum really needed to be in the header? It looks like it can be embedded in the source file itself. Then it won't be visible in other compilation units, and I'd be fine with `FC_` since it will become local-to-single-file.
___
*[`pal/src/host/linux-sgx/utils/fast_clock.h` line 39 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui5K1g7WVt4Gk8MSVc:-Nui5K1g7WVt4Gk8MSVd:b-ob8dqp) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.h#L39)):*
> ```objc
> typedef union
> {
>     #pragma pack(push, 1)
> ```

Why you need this bit-compression?
___
*[`pal/src/host/linux-sgx/utils/fast_clock.c` line 1 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui5QueAKV-8EogrIX-:-Nui5QueAKV-8EogrIX0:b-x8kcj4) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.c#L1)):*
> ```C
> #include <string.h>
> ```

You should use `"api.h"`
___
*[`pal/src/host/linux-sgx/utils/fast_clock.c` line 10 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui5WWdE-9mzH3QpzdD:-Nui5WWdE-9mzH3QpzdE:bg05uf2) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.c#L10)):*
> ```C
> 
> /**
>  * FastClock
> ```

Please reformat to 100-chars-per-line.
___
*[`pal/src/host/linux-sgx/utils/fast_clock.c` line 28 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui5aRnBrRpBjIMOir9:-Nui5aRnBrRpBjIMOirA:b2qfgut) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.c#L28)):*
> ```C
>  * *** Implementation ***
>  *
>  * FastClock is implemented as a state machine. This was done since we don't have a good portable way to get the cpu clock frequency.
> ```

It would be nicer if you would ASCII-draw the state machine with transitions.
___
*[`pal/src/host/linux-sgx/utils/fast_clock.c` line 81 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui5h4S3To2EVzd7h5f:-Nui5h4S3To2EVzd7h5g:b7nvhjs) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.c#L81)):*
> ```C
> #ifndef SEC_USEC
> #define SEC_USEC                        ((uint64_t)1000000)
> #endif
> ```

We have some macro for this in `api.h`, please check (I don't remember the name)
___
*[`pal/src/host/linux-sgx/utils/fast_clock.c` line 84 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui5lBT35dS0W0me0vW:-Nui5lBU8JR9MAqgcaDE:b3sbu8z) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.c#L84)):*
> ```C
> 
> #define RDTSC_CALIBRATION_TIME          ((uint64_t)1 * SEC_USEC)
> #define RDTSC_RECALIBRATION_INTERVAL    ((uint64_t)120 * SEC_USEC)
> ```

How did you decide on these values? Please add a comment.
___
*[`pal/src/host/linux-sgx/utils/fast_clock.c` line 139 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui60lq5GCIkEmCm7cF:-Nui60lq5GCIkEmCm7cG:b-35h9si) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.c#L139)):*
> ```C
> static inline long reset_timepoint(fast_clock_timepoint_t* timepoint)
> {
>     int ret = ocall_reset_time(&timepoint->t0_usec, &timepoint->tsc0, &timepoint->tz_minutewest, &timepoint->tz_dsttime);
> ```

Need to analyze security and validate received-from-host values
___
*[`pal/src/host/linux-sgx/utils/fast_clock.c` line 148 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui683m19NF-S0KSwbr:-Nui683m19NF-S0KSwbs:bjjct5e) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.c#L148)):*
> ```C
> }
> 
> static inline fast_clock_desc_t desc_atomic_load(const fast_clock_t* fast_clock, int mo)
> ```

What's the point in this helper function? It brings no benefit.
___
*[`pal/src/host/linux-sgx/utils/fast_clock.c` line 155 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui6BGM3YmSNWz9ii0X:-Nui6BGM3YmSNWz9ii0Y:bjjct5e) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.c#L155)):*
> ```C
> }
> 
> static inline void desc_atomic_store(fast_clock_t* fast_clock, fast_clock_desc_t new_desc, int mo)
> ```

What's the point in this helper function? It brings no benefit.
___
*[`pal/src/host/linux-sgx/utils/fast_clock.c` line 164 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui6FTi0xlVceOUlq2m:-Nui6FTi0xlVceOUlq2n:b-adfubf) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.c#L164)):*
> ```C
> static inline int handle_state_rdtsc(fast_clock_t* fast_clock, fast_clock_desc_t descriptor, uint64_t* time_usec, bool force_new_timepoint);
> static inline int handle_state_rdtsc_recalibrate(fast_clock_t* fast_clock, fast_clock_desc_t descriptor, uint64_t* time_usec);
> static int handle_state_rdtsc_disabled(uint64_t* time_usec);
> ```

This is ugly, why can't you define functions in the order of dependency?
___
*[`pal/src/host/linux-sgx/utils/fast_clock.c` line 166 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui6JgiBGwORUbJQ7Jo:-Nui6JgiBGwORUbJQ7Jp:b-vbhoib) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.c#L166)):*
> ```C
> static int handle_state_rdtsc_disabled(uint64_t* time_usec);
> 
> int fast_clock__get_time(fast_clock_t* fast_clock, uint64_t* time_usec, bool force_new_timepoint)
> ```

We don't use the `__` style. Just one `_` is enough.
___
*[`pal/src/host/linux-sgx/utils/fast_clock.c` line 290 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui65A3EPi-MOI5AIMH:-Nui65A3EPi-MOI5AIMI:b-35h9si) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.c#L290)):*
> ```C
>     // all callers in this state will perform an OCALL - no need to set the change_state_guard before OCALLing
>     uint64_t tmp_tsc = 0;
>     int ret = ocall_reset_time(time_usec, &tmp_tsc, NULL, NULL);
> ```

Need to analyze security and validate received-from-host values
___
*[`pal/src/host/linux-sgx/utils/fast_clock.c` line 350 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui65tj2nAipfqkjNLN:-Nui65tj2nAipfqkjNLO:b-35h9si) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.c#L350)):*
> ```C
> 
>     uint64_t tsc = 0;
>     int ret = ocall_reset_time(time_usec, &tsc, NULL, NULL);
> ```

Need to analyze security and validate received-from-host values
___
*[`pal/src/host/linux-sgx/utils/fast_clock.c` line 362 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui6WnT3vgOZwWJdDg1:-Nui6WnT3vgOZwWJdDg2:b-tgj1ck) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.c#L362)):*
> ```C
> 
>     return ret;
> }
> ```

Please add newlines at the end of each file.

jonathan-sha:
Thanks for the review!

> _[`pal/src/host/linux-sgx/enclave_ocalls.h` line 92 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui41nP8JYuBwW_Gw4q:-Nui41nQ0SG_Ff-3mTHk:bb288uj) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/enclave_ocalls.h#L92)):_
> 
> > ```objc
> > int ocall_futex(uint32_t* uaddr, int op, int val, uint64_t* timeout_us);
> > 
> > int ocall_reset_time(uint64_t* microsec, uint64_t* tsc, int* tz_minuteswest, int* tz_dsttime);
> > ```
> 
> Can we remove the `tz_` args and handling from this PR? You don't use them here anyway, and it's harder to review.

yes, I can remove this. There's currently an "unimplemented" comment in the gettimeofday libos method, as well as issue #466 - do you prefer I fix this as well (at least partially, for gettimeofday())?

> _[`pal/src/host/linux-sgx/enclave_ocalls.c` line 1825 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui4Ils6eTgvz6LNCQs:-Nui4Ils6eTgvz6LNCQt:b-8slbc3) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/enclave_ocalls.c#L1825)):_
> 
> > ```c
> > {
> >     return ocall_reset_time(microsec_ptr, NULL, NULL, NULL);
> > }
> > ```
> 
> That's a bad security practice -- you added an additional attack surface (a new OCALL) for no good reason. Just expand existing `ocall_gettime` to have an additional argument `tsc_ptr`, and remove `ocall_reset_time()`.

This is what I did actually, I renamed `ocall_gettime` to `ocall_reset_time` and added the new parameters. I kept `ocall_gettime` for compatibility with the rest of the code. I can rename this back to `ocall_gettime` and fix accordingly.
Regarding caching sanity checking rdtsc - I read somewhere that a privileged user can reset the TSC counter, which makes it inherently unsafe. But it seems like I was wrong (?). I will add two `get_tsc` calls before and after the ocall and validate that the host-side tsc is between the two. Do you think I still need to check that it monotonically increases (this is guaranteed in x86)?

> _[`pal/src/host/linux-sgx/host_ocalls.c` line 608 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui3gtW3bSvs7AXM0sE:-Nui3gtW3bSvs7AXM0sF:bdqn4ji) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/host_ocalls.c#L608)):_
> 
> > ```c
> >     struct timeval tv;
> >     struct timezone tz;
> >     unsigned long long tsc = __rdtsc();
> > ```
> 
> We have `get_tsc()` which does the same. Please remove `x86intrin.h`

This is the host code (outside of sgx) - should I still use the enclave reimplementation of rdtsc?

> _[`pal/src/host/linux-sgx/pal_main.c` line 561 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui4c8V6o2Vdkf76zap:-Nui4c8V6o2Vdkf76zaq:bn8xn5p) ([raw file](https://github.com/gramineproject/gramine/blob/0ee48eebb24c38bdad193fa9945a3b26d4a89dc6/pal/src/host/linux-sgx/pal_main.c#L561)):_
> 
> > ```c
> >                     "not expose corresponding CPUID leaves). This degrades performance.");
> >     }
> > }
> > ```
> 
> Why did you remove `print_warning_on_invariant_tsc()`?

I need to rework the initialization a bit. I'll re-add this print, but I'm not sure we need it to happen in the main function.

> _[`pal/src/host/linux-sgx/utils/fast_clock.h` line 39 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui5K1g7WVt4Gk8MSVc:-Nui5K1g7WVt4Gk8MSVd:b-ob8dqp) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.h#L39)):_
> 
> > ```objc
> > typedef union
> > {
> >     #pragma pack(push, 1)
> > ```
> 
> Why you need this bit-compression?

I changed this to a bitfield in my company's codebase (I'll pull in these changes). I just wanted to make sure this struct fits in a uint32_t which we can load \ store atomically.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1834)* status: 0 of 11 files reviewed, 28 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @jonathan-sha)

___
*[`pal/src/host/linux-sgx/enclave_ocalls.h` line 92 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui41nP8JYuBwW_Gw4q:-NuwwxTYBDIdKm20klu5:b-ura4lp) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/enclave_ocalls.h#L92)):*
> do you prefer I fix this as well (at least partially, for gettimeofday())?

Sorry, don't understand your proposal. Fix how?

To make it clear, I suggest the following:
1. Remove in this PR everything about timezones.
2. Create a follow-up PR to this one that adds timezones (and thus fixes #466).
___
*[`pal/src/host/linux-sgx/enclave_ocalls.c` line 1825 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui4Ils6eTgvz6LNCQs:-NuwxAhq7N9kScswqpuV:bwwhmmg) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/enclave_ocalls.c#L1825)):*
> I kept `ocall_gettime` for compatibility with the rest of the code.

No reason to keep compatibility. Just change the code. (These are internal Gramine functions, so can be changed without caring about compatibility issues.)
___
*[`pal/src/host/linux-sgx/host_ocalls.c` line 608 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui3gtW3bSvs7AXM0sE:-NuwxLKq9EPSNCC6Ka2F:b-rnkl9d) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/host_ocalls.c#L608)):*
<details><summary><i>Previously, jonathan-sha (Jonathan Shamir) wrote…</i></summary><blockquote>

This is the host code (outside of sgx) - should I still use the enclave reimplementation of rdtsc?
</blockquote></details>

We have this func in `cpu.h`: https://github.com/gramineproject/gramine/blob/fcdd1d5f84724fe0044db7025ef636532a83f710/common/include/arch/x86_64/cpu.h#L66

The `cpu.h` header is independent from enclave/non-enclave environment. So yes, just use it.
___
*[`pal/src/host/linux-sgx/utils/fast_clock.h` line 39 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui5K1g7WVt4Gk8MSVc:-NuwxdzN1rpQg_qgVntc:b-94d64) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.h#L39)):*
<details><summary><i>Previously, jonathan-sha (Jonathan Shamir) wrote…</i></summary><blockquote>

I changed this to a bitfield in my company's codebase (I'll pull in these changes). I just wanted to make sure this struct fits in a uint32_t which we can load \ store atomically. 
</blockquote></details>

I see. You can change to a bitfield here too.

Also, please add a `static_assert(sizeof (fast_clock_desc_t) == sizeof(uint64_t))` to make it explicit. Plus add a comment that this struct must be accessed with atomic ops.

jonathan-sha:
@dimakuv I updated the MR, fixed most of your comments.

Two big open issues that I would like to get feedback on -
1. `is_rdtsc_available` - since gramine has rdtsc emulation, I think we can take this logic out of fast-clock, and move this into the error handler for rdtsc emulation. I can implement a new `init_gettime()` method that calls `get_tsc()`, then checks a global to see if we're in emulation mode, and `fast_clock_enable()` only if we're not. We can then remove all the cpuid logic from the init state of fast-clock (and bring back the warning log I removed).
2. should we update the gettimeofday() tests? I'm not sure they're elaborate enough. In my company's codebase we have cycle calculation, comparison with native \ ocall \ fast-clock implementation, multi-threaded stress, checking time drift with host. Currently gramine has multi-threaded stress, but it's iterations based and not time based (so fast-clock won't advance through all it's states). We should add a 5-minute stress test.

mkow (changes requested):
Replied to 1. under the same discussion in the code.

> should we update the gettimeofday() tests?

We could add/improve some, but I'm not sure if it's worth comparing to the host - I'm afraid that they may fail randomly if we get unlucky with a CI worker load and we'll end up debugging CI failures which are just flaky tests, not real bugs.

> We should add a 5-minute stress test.

No, we shouldn't :) 5 minutes is really a lot of CI time spent on a single feature.

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1834#-:-NvMu4kRBZO5IfquNW3b:bxpomiz)* status: 0 of 12 files reviewed, 31 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @jonathan-sha)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui6Zr1-2fOipiHl49G:-NvMtw8F7FTI3S8cTFio:bckaza3) (no related file):*

I just briefly looked at the PR and left some comments. Didn't go deep into the logic itself.

Please fix according to the comments, and then we could start the review in earnest.
</blockquote></details>

Same. Will do a more throughout review later.

___
___
*[`-- commits` line 3 at r2](https://reviewable.io/reviews/gramineproject/gramine/1834#-NvMuX0C0362iXaW_F0t):*




_Suggestion:_
```
[PAL/Linux-SGX] Add ocall_reset_clock which returns tsc and timezone info
```
___
*[`-- commits` line 7 at r2](https://reviewable.io/reviews/gramineproject/gramine/1834#-NvMuf2DEKg7UlxpMo4L):*
I don't think there's a point in splitting this change into 4 commits, as far as I understand they can't/shouldn't be used separately, so they aren't really independent of each other.
___
*[`-- commits` line 19 at r2](https://reviewable.io/reviews/gramineproject/gramine/1834#-NvMuCI79jfRxRmxvbKl):*
Please use fixup commits when adding fixups, see https://gramine.readthedocs.io/en/latest/devel/contributing.html#pr-life-cycle. And in general, please read that contributing guide :)
___
*[`pal/src/host/linux-sgx/enclave_ocalls.h` line 92 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui41nP8JYuBwW_Gw4q:-NvMv1_S5W-p8JpdBp2k:b-l73ljr) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/enclave_ocalls.h#L92)):*

> do you prefer I fix this as well (at least partially, for gettimeofday())?

Sorry, don't understand your proposal. Fix how?

To make it clear, I suggest the following:
1. Remove in this PR everything about timezones.
2. Create a follow-up PR to this one that adds timezones (and thus fixes #466).
</blockquote></details>

+1, I'd skip the timezones problem for now and focus on merging the core functionality first.
___
*[`pal/src/host/linux-sgx/utils/fast_clock.c` line 226 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#gh-1548130275:-NvMw5CJB7hy7r3-yM1O:b-a39imh) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.c#L226)):*
<details><summary><i>Previously, jonathan-sha (Jonathan Shamir) wrote…</i></summary><blockquote>

Not sure if this method is needed - can we rely on the RDTSC emulation feature?

Maybe we can do the following - have a global `g_is_rdtsc_emulated` which is set under the `FIRST_TIME()` guard in the emulator. This method will simply call `get_tsc()` then check the global. We can then completely remove this logic from `pal_sgx_main`.
</blockquote></details>

Yeah, sounds like a much better idea IMO.

jonathan-sha:
Hey, it's been a while. I would like to bring this mr back to life if you're interested. I fixed all the above comments.

@dimakuv @mkow

mkow:
Yeah, I remember about it, but I was quite busy recently with other stuff. I have some more time now, so I'll continue the review now :)

mkow:
@jonathan-sha: Could you reply to all the discussions in Reviewable? There are a few of them from Dmitrii which you left unanswered and it's not obvious whether you fixed the thing pointed out there or not.

jonathan-sha (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1834)* status: 0 of 12 files reviewed, 29 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @mkow)

___
*[`-- commits` line 7 at r2](https://reviewable.io/reviews/gramineproject/gramine/1834#-NvMuf2DEKg7UlxpMo4L):*

I don't think there's a point in splitting this change into 4 commits, as far as I understand they can't/shouldn't be used separately, so they aren't really independent of each other.
</blockquote></details>

I squashed all commits
___
*[`pal/src/host/linux-sgx/enclave_ocalls.h` line 92 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui41nP8JYuBwW_Gw4q:-O2dh1KfAB_ae9IKV-1m:bewpmj4) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/enclave_ocalls.h#L92)):*

+1, I'd skip the timezones problem for now and focus on merging the core functionality first.
</blockquote></details>

timezone handling was removed
___
*[`pal/src/host/linux-sgx/enclave_ocalls.c` line 1825 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui4Ils6eTgvz6LNCQs:-O2dhUD472Nzjni-jMKM:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/enclave_ocalls.c#L1825)):*

> I kept `ocall_gettime` for compatibility with the rest of the code.

No reason to keep compatibility. Just change the code. (These are internal Gramine functions, so can be changed without caring about compatibility issues.)
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/host_ocalls.c` line 608 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui3gtW3bSvs7AXM0sE:-O2dhlYDD7ikledFBH7d:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/host_ocalls.c#L608)):*

We have this func in `cpu.h`: https://github.com/gramineproject/gramine/blob/fcdd1d5f84724fe0044db7025ef636532a83f710/common/include/arch/x86_64/cpu.h#L66

The `cpu.h` header is independent from enclave/non-enclave environment. So yes, just use it.
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/pal_main.c` line 561 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui4c8V6o2Vdkf76zap:-O2dhqUZ4A2kb6ksjKL0:b-y83n23) ([raw file](https://github.com/gramineproject/gramine/blob/0ee48eebb24c38bdad193fa9945a3b26d4a89dc6/pal/src/host/linux-sgx/pal_main.c#L561)):*
<details><summary><i>Previously, jonathan-sha (Jonathan Shamir) wrote…</i></summary><blockquote>

I need to rework the initialization a bit. I'll re-add this print, but I'm not sure we need it to happen in the main function.
</blockquote></details>

I added print_warnings_on_disabled_clock_emulation above
___
*[`pal/src/host/linux-sgx/pal_misc.c` line 319 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui4goHAXmlioS0i3zV:-O2di6E1Cgmyv1prtOeU:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/pal_misc.c#L319)):*

These 2 leaves won't be needed now, right? We can remove them.
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/utils/fast_clock.h` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui4po26N4JaaLnopTS:-O2di74_9uLyfPpevZ8T:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.h#L2)):*

We use `#pragma once`
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/utils/fast_clock.h` line 9 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui4n4pEh8cHWb1tQWx:-O2di7vl3qt4_YmxhoIX:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.h#L9)):*

There is nothing C++ in our codebase, please remove these guards.
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/utils/fast_clock.h` line 13 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui4ux38R1aRsxtYqfE:-O2di8itBuyR5L4UzOG8:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.h#L13)):*

We have a different coding style (based on Google C/C++ coding style), see https://gramine.readthedocs.io/en/stable/devel/coding-style.html
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/utils/fast_clock.h` line 19 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui542W86c5YrrS93qN:-O2di9nC-mN6TvAAizMD:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.h#L19)):*

I don't like the `FC_`, it's not immediately obvious what it is.

Hm, is this enum really needed to be in the header? It looks like it can be embedded in the source file itself. Then it won't be visible in other compilation units, and I'd be fine with `FC_` since it will become local-to-single-file.
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/utils/fast_clock.h` line 39 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui5K1g7WVt4Gk8MSVc:-O2diBbQCh0Db-PSeCRA:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.h#L39)):*

I see. You can change to a bitfield here too.

Also, please add a `static_assert(sizeof (fast_clock_desc_t) == sizeof(uint64_t))` to make it explicit. Plus add a comment that this struct must be accessed with atomic ops.
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/utils/fast_clock.c` line 1 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui5QueAKV-8EogrIX-:-O2diFJY5bfcvTgbL4Du:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.c#L1)):*

You should use `"api.h"`
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/utils/fast_clock.c` line 10 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui5WWdE-9mzH3QpzdD:-O2diHch8SbnOHRL8Z_9:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.c#L10)):*

Please reformat to 100-chars-per-line.
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/utils/fast_clock.c` line 28 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui5aRnBrRpBjIMOir9:-O2diInRBSK5GlpuL9n-:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.c#L28)):*

It would be nicer if you would ASCII-draw the state machine with transitions.
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/utils/fast_clock.c` line 81 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui5h4S3To2EVzd7h5f:-O2diN6pEC06PxHypQrU:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.c#L81)):*

We have some macro for this in `api.h`, please check (I don't remember the name)
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/utils/fast_clock.c` line 84 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui5lBT35dS0W0me0vW:-O2diRY27tgBg-fDeCYy:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.c#L84)):*

How did you decide on these values? Please add a comment.
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/utils/fast_clock.c` line 139 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui60lq5GCIkEmCm7cF:-O2diVcqAPk7HVJ5cAdL:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.c#L139)):*

Need to analyze security and validate received-from-host values
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/utils/fast_clock.c` line 148 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui683m19NF-S0KSwbr:-O2diXIv4V432wx-XdOj:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.c#L148)):*

What's the point in this helper function? It brings no benefit.
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/utils/fast_clock.c` line 155 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui6BGM3YmSNWz9ii0X:-O2diZ5s-QjAR_RA072y:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.c#L155)):*

What's the point in this helper function? It brings no benefit.
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/utils/fast_clock.c` line 164 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui6FTi0xlVceOUlq2m:-O2diZww8a8Dmrjrh1i9:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.c#L164)):*

This is ugly, why can't you define functions in the order of dependency?
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/utils/fast_clock.c` line 166 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui6JgiBGwORUbJQ7Jo:-O2di_ax357npgtsHDUl:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.c#L166)):*

We don't use the `__` style. Just one `_` is enough.
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/utils/fast_clock.c` line 226 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#gh-1548130275:-O2did4K4SbIrMLz4ZTq:bb71zzx) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.c#L226)):*

Yeah, sounds like a much better idea IMO.
</blockquote></details>

I removed the previous implementation of this - `is_rdtsc_available` used to check using cpuid if rdtsc is legal in an sgx enclave on this machine. Since gramine has the tsc emulation feature, what I currently do is check if the emulation was triggered.
In linux_pal_main we still have the `get_tsc()` early during start up. If rdtsc isn't supported, this will trigger a `fast_clock_disable` on the global `g_fast_clock`. This method is kept so we can transition to DISABLED in case we already had a `gettimeofday` in flight (during linux_pal_main for some reason) or if the non-global g_fast_clock is used (unit tests etc).

On second thought - another implementation could be to start off as DISABLED, and implement `fast_clock_enable` method. This could be done in the pal_main after triggering `get_tsc()`, so we separate responsibilities. fast-clock will assume that rdtsc is allowed and pal_main will enable fast-clock only after checking that rdtsc isn't emulated. 
___
*[`pal/src/host/linux-sgx/utils/fast_clock.c` line 290 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui65A3EPi-MOI5AIMH:-O2dleCp4rujX9qVjsNZ:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.c#L290)):*

Need to analyze security and validate received-from-host values
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/utils/fast_clock.c` line 350 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui65tj2nAipfqkjNLN:-O2dlf5F2R5_Pc7TpHM-:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.c#L350)):*

Need to analyze security and validate received-from-host values
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/utils/fast_clock.c` line 362 at r1](https://reviewable.io/reviews/gramineproject/gramine/1834#-Nui6WnT3vgOZwWJdDg1:-O2dll5XARTv5qowqlaF:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/9679f960c185cce962fc85e1ef92a42846f272d3/pal/src/host/linux-sgx/utils/fast_clock.c#L362)):*

Please add newlines at the end of each file.
</blockquote></details>

Done.

mkow (changes requested):
> Reviewed 1 of 11 files at r1, 3 of 11 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1834)* status: 4 of 12 files reviewed, 30 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @jonathan-sha)

___
*[`-- commits` line 2 at r3](https://reviewable.io/reviews/gramineproject/gramine/1834#-O2bJaQjAgKLrlf2EWsg):*
This one-liner is much too long, see https://commit.style/ (no need to strictly follow the 50-char limit, but 134 characters is much too much :) Please move the details to the commit message body.
___
*[`pal/src/host/linux-sgx/enclave_ocalls.c` line 1769 at r3](https://reviewable.io/reviews/gramineproject/gramine/1834#-O4RDA9-E6vcyANogLdU:-O4RDA9-E6vcyANogLdV:b-h6w8ch) ([raw file](https://github.com/gramineproject/gramine/blob/664de3ddc19edbc3fedf529da23b2aec86d996bb/pal/src/host/linux-sgx/enclave_ocalls.c#L1769)):*
> ```C
> }
> 
> int ocall_gettime(uint64_t* microsec_ptr, uint64_t* tsc_ptr) {
> ```

Sorry, old code, we must have missed this when unifying naming conventions.



_Suggestion:_
```C
uint64_t* out_microsec, uint64_t* out_tsc
```
___
*[`pal/src/host/linux-sgx/enclave_ocalls.c` line 1814 at r3](https://reviewable.io/reviews/gramineproject/gramine/1834#-O4RDmDhE4Smqkv4c3bo:-O4RDmDhE4Smqkv4c3bp:b-fxmrb1) ([raw file](https://github.com/gramineproject/gramine/blob/664de3ddc19edbc3fedf529da23b2aec86d996bb/pal/src/host/linux-sgx/enclave_ocalls.c#L1814)):*
> ```C
>     }
> 
>     /* detect malicious host - time and tsc must monotonically increase */
> ```

Couldn't we just use `tsc_before_ocall` for `new_tsc`? The OCALL delay would cancel out, if it's roughtly constant (but and don't know if it is). This would considerably simplify the logic here.
___
*[`pal/src/host/linux-sgx/meson.build` line 89 at r3](https://reviewable.io/reviews/gramineproject/gramine/1834#-O4REWF-05_0nWtgzKwy:-O4REWF-05_0nWtgzKwz:b96snvu) ([raw file](https://github.com/gramineproject/gramine/blob/664de3ddc19edbc3fedf529da23b2aec86d996bb/pal/src/host/linux-sgx/meson.build#L89)):*
> ```Meson
>     'pal_streams.c',
>     'pal_threading.c',
>     'utils/fast_clock.c',
> ```

I don't think we need a directory (especially with such a vague name), there would only be a single file there.



_Suggestion:_
```Meson
'pal_rdtsc_clock.c',
```

jonathan-sha:
Hey, it's been a while since I last talked here.
I wanted to get this pr in, but I left the company I was working at. I'm still interested in moving this forward if you want this.
Unfortunately, I no longer have access to machines with sgx, so I can't properly test the code. Is there any way you can help me with this? I really like this pr and would love to move this forward if I can.

woju (commented):
It's being worked on. All machines we have today are located more or less in trusted networks, where I'm not comfortable giving access to for people who are not Maintainers, but the two new machines (@mkow you know, which ones) will be placed in public colocation, so OK to share access to (in a VM most likely).

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1834#-:-OKjvNkvEQMilnYRnK_r:bx12ggk)* status: 4 of 12 files reviewed, 30 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @jonathan-sha, and @mkow)

kailun-qin (commented):
@jonathan-sha: Just checking in -- are you still interested in moving this PR forward? Thanks!

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1834#-:-ORrzmnS4bRh3ibF0YO-:b-hbxrnd)* status: 4 of 12 files reviewed, 30 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @jonathan-sha, and @mkow)

PR #1859, opened 2024-04-24
[LibOS] Add `sys.debug__mock_syscalls = [ ... ]` manifest option

dimakuv:
## Description of the changes 

This PR adds the manifest syntax ~~`sys.disallowed_syscalls = [ ... ]`~~ ~~`sys.mock_syscalls = [ ... ]`~~ `sys.debug_mock_syscalls = [ ... ]` to specify system calls that will return a specified value when executed in Gramine.

## How to test this PR? 

Added a LibOS regression test.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1859)


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1859)* status: 0 of 12 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1859#-NwFEBEO--9n0qirnWki:-NwFEBEO--9n0qirnWkj:btk6t7f) (no related file):*
This is a patch that I was experimenting with (for reasons not relevant to this PR). I decided to create the PR out of what I have. No idea if this is useful... Any feedback welcome.

___
___
*[`Documentation/manifest-syntax.rst` line 397 at r1](https://reviewable.io/reviews/gramineproject/gramine/1859#-NwFEQ29CEoHcLbbz4Bu:-NwFEQ29CEoHcLbbz4Bv:b-qk94si) ([raw file](https://github.com/gramineproject/gramine/blob/0504ab3a832e24202fea3288d8193de7fe311bfe/Documentation/manifest-syntax.rst#L397)):*
> ```reStructuredText
> ::
> 
>     sgx.disallowed_syscalls = [
> ```

Alternatives:
1. Provide an allowlist `sys.allowed_syscalls = [ ... ]`. This will be more security-expected and more in line with e.g. seccomp profiles. Downside: much harder for end users.
2. Provide both this denylist and the allowlist manifest options. At init time, Gramine fails if detects them both.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1859)* status: 0 of 12 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

___
*[`Documentation/manifest-syntax.rst` line 399 at r2](https://reviewable.io/reviews/gramineproject/gramine/1859#-NwPCPKt6qX2XH4TOeCT:-NwPCPKt6qX2XH4TOeCU:b-d8o22c) ([raw file](https://github.com/gramineproject/gramine/blob/8820f1035f85a50080b93185782fcb854993ba2f/Documentation/manifest-syntax.rst#L399)):*
> ```reStructuredText
>     sys.disallowed_syscalls = [
>       "syscall_name",
>       "syscall_name",
> ```

One additional option would be to not simply return `-ENOSYS`, but allow to specify what int to return:
```
    sys.fixed_syscalls = [
      { name = "syscall_name1", return = -38},  # returns -ENOSYS
      { name = "syscall_name2", return = 0},    # returns 0, i.e. dummy impl
```

oshogbo (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1859)* status: 0 of 12 files reviewed, 9 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`Documentation/manifest-syntax.rst` line 397 at r2](https://reviewable.io/reviews/gramineproject/gramine/1859#-NxCDN3E0vZW4vDLiSkn:-NxCDN3E0vZW4vDLiSko:b3mywyq) ([raw file](https://github.com/gramineproject/gramine/blob/8820f1035f85a50080b93185782fcb854993ba2f/Documentation/manifest-syntax.rst#L397)):*
> ```reStructuredText
> ::
> 
>     sys.disallowed_syscalls = [
> ```

Should this also be read from a file similarly?
___
*[`libos/src/libos_parser.c` line 1653 at r2](https://reviewable.io/reviews/gramineproject/gramine/1859#-NxCDwrT5k_GTCYSoL-B:-NxCDwrT5k_GTCYSoL-C:b78p052) ([raw file](https://github.com/gramineproject/gramine/blob/8820f1035f85a50080b93185782fcb854993ba2f/libos/src/libos_parser.c#L1653)):*
> ```C
> 
> unsigned long get_syscall_number(const char* name) {
>     assert(LIBOS_SYSCALL_BOUND == ARRAY_SIZE(syscall_parser_table));
> ```

Shouldn't this be a static_assert beyond this function?
___
*[`libos/src/libos_parser.c` line 1658 at r2](https://reviewable.io/reviews/gramineproject/gramine/1859#-NxCDexH848eF4Wr__dd:-NxCDexH848eF4Wr__de:b-exnbep) ([raw file](https://github.com/gramineproject/gramine/blob/8820f1035f85a50080b93185782fcb854993ba2f/libos/src/libos_parser.c#L1658)):*
> ```C
>         if (!syscall_parser_table[i].name)
>             continue;
>         if (strcmp(name, syscall_parser_table[i].name) == 0) {
> ```

Do we need `sysno`?

```
 if (strcmp(name, syscall_parser_table[i].name) == 0) {
  return i;
}
...

return LIBOS_SYSCALL_BOUND;
}
```

Also shouldn't this indicate the error more clearly?
___
*[`libos/src/libos_syscalls.c` line 114 at r2](https://reviewable.io/reviews/gramineproject/gramine/1859#-NxCEH7l4arc8HQNcaX-:-NxCEH7l4arc8HQNcaX0:b11qmc3) ([raw file](https://github.com/gramineproject/gramine/blob/8820f1035f85a50080b93185782fcb854993ba2f/libos/src/libos_syscalls.c#L114)):*
> ```C
>         if (!toml_disallowed_syscall_raw) {
>             log_error("Invalid disallowed syscall in manifest at index %ld", i);
>             return -EINVAL;
> ```

goto after return?
___
*[`libos/src/libos_syscalls.c` line 121 at r2](https://reviewable.io/reviews/gramineproject/gramine/1859#-NxCEHrS46uhiKOIYasj:-NxCEHrS46uhiKOIYask:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/8820f1035f85a50080b93185782fcb854993ba2f/libos/src/libos_syscalls.c#L121)):*
> ```C
>         if (ret < 0) {
>             log_error("Invalid disallowed syscall in manifest at index %ld (not a string)", i);
>             return -EINVAL;
> ```

ditto
___
*[`libos/src/libos_syscalls.c` line 129 at r2](https://reviewable.io/reviews/gramineproject/gramine/1859#-NxCEIYN6gEVzWTDVlF3:-NxCEIYN6gEVzWTDVlF4:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/8820f1035f85a50080b93185782fcb854993ba2f/libos/src/libos_syscalls.c#L129)):*
> ```C
>             log_error("Unrecognized disallowed syscall `%s` in manifest at index %ld",
>                       toml_disallowed_syscall_str, i);
>             return -EINVAL;
> ```

ditto

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1859)* status: 0 of 12 files reviewed, 9 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @oshogbo)

___
*[`Documentation/manifest-syntax.rst` line 397 at r2](https://reviewable.io/reviews/gramineproject/gramine/1859#-NxCDN3E0vZW4vDLiSkn:-NxD8-wW9TKsk4Uu6pdh:b5n9gt4) ([raw file](https://github.com/gramineproject/gramine/blob/8820f1035f85a50080b93185782fcb854993ba2f/Documentation/manifest-syntax.rst#L397)):*

Should this also be read from a file similarly?
</blockquote></details>

What would be the benefit of having these syscalls in a separate file, rather than directly in the manifest?
___
*[`libos/src/libos_parser.c` line 1653 at r2](https://reviewable.io/reviews/gramineproject/gramine/1859#-NxCDwrT5k_GTCYSoL-B:-NxD8cmY-yBI6B0Y9Vl3:b-ppj781) ([raw file](https://github.com/gramineproject/gramine/blob/8820f1035f85a50080b93185782fcb854993ba2f/libos/src/libos_parser.c#L1653)):*

Shouldn't this be a static_assert beyond this function?
</blockquote></details>

Done partially. It indeed needs to be a static assert, but I don't know what you mean by "beyond this function". I add this assert here to make it clear to readers that these two are the same limits.
___
*[`libos/src/libos_parser.c` line 1658 at r2](https://reviewable.io/reviews/gramineproject/gramine/1859#-NxCDexH848eF4Wr__dd:-NxD8mQP0D0ebI-oMJsr:b2osulh) ([raw file](https://github.com/gramineproject/gramine/blob/8820f1035f85a50080b93185782fcb854993ba2f/libos/src/libos_parser.c#L1658)):*

Do we need `sysno`?

```
 if (strcmp(name, syscall_parser_table[i].name) == 0) {
  return i;
}
...

return LIBOS_SYSCALL_BOUND;
}
```

Also shouldn't this indicate the error more clearly?
</blockquote></details>

Done partially.

What do you mean by "indicate the error more clearly"? Replace with `int get_syscall_number(const char* name, unsigned long* out_sysno)`? I can do it, though the current version looked sufficiently clear to me.
___
*[`libos/src/libos_syscalls.c` line 114 at r2](https://reviewable.io/reviews/gramineproject/gramine/1859#-NxCEH7l4arc8HQNcaX-:-NxD9J-XEgiQ-9KYdn1v:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/8820f1035f85a50080b93185782fcb854993ba2f/libos/src/libos_syscalls.c#L114)):*

goto after return?
</blockquote></details>

Done.
___
*[`libos/src/libos_syscalls.c` line 121 at r2](https://reviewable.io/reviews/gramineproject/gramine/1859#-NxCEHrS46uhiKOIYasj:-NxD9JkZ3o08iO-Y2L5x:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/8820f1035f85a50080b93185782fcb854993ba2f/libos/src/libos_syscalls.c#L121)):*

ditto
</blockquote></details>

Done.
___
*[`libos/src/libos_syscalls.c` line 129 at r2](https://reviewable.io/reviews/gramineproject/gramine/1859#-NxCEIYN6gEVzWTDVlF3:-NxD9K31--nU9iEDCAIr:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/8820f1035f85a50080b93185782fcb854993ba2f/libos/src/libos_syscalls.c#L129)):*

ditto
</blockquote></details>

Done.

oshogbo (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1859)* status: 0 of 12 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`Documentation/manifest-syntax.rst` line 397 at r2](https://reviewable.io/reviews/gramineproject/gramine/1859#-NxCDN3E0vZW4vDLiSkn:-NxH7JPu7oFdRTPMd7BE:bm4acum) ([raw file](https://github.com/gramineproject/gramine/blob/8820f1035f85a50080b93185782fcb854993ba2f/Documentation/manifest-syntax.rst#L397)):*

What would be the benefit of having these syscalls in a separate file, rather than directly in the manifest?
</blockquote></details>

I was thinking about docker seccomp policies, and they are mainly defined as an outside file. So my idea was that somebody might want to define a more generic policy. But this was just a random idea.
___
*[`libos/src/libos_parser.c` line 1653 at r2](https://reviewable.io/reviews/gramineproject/gramine/1859#-NxCDwrT5k_GTCYSoL-B:-NxH7_Ct8EogFHpjM-OX:biwa7y7) ([raw file](https://github.com/gramineproject/gramine/blob/8820f1035f85a50080b93185782fcb854993ba2f/libos/src/libos_parser.c#L1653)):*

Done partially. It indeed needs to be a static assert, but I don't know what you mean by "beyond this function". I add this assert here to make it clear to readers that these two are the same limits.
</blockquote></details>

"Beyond this function" I thought about doing it above or where the syscall_parse_table/IBOS_SYSCALL_BOUND is defined. But here makes sense as well.
___
*[`libos/src/libos_parser.c` line 1658 at r2](https://reviewable.io/reviews/gramineproject/gramine/1859#-NxCDexH848eF4Wr__dd:-NxH7o-D49ANMbJbyTnc:b1x07gn) ([raw file](https://github.com/gramineproject/gramine/blob/8820f1035f85a50080b93185782fcb854993ba2f/libos/src/libos_parser.c#L1658)):*

Done partially.

What do you mean by "indicate the error more clearly"? Replace with `int get_syscall_number(const char* name, unsigned long* out_sysno)`? I can do it, though the current version looked sufficiently clear to me.
</blockquote></details>

Yes, my idea was to not use a LIBOS_SYSCALL_BOUND as a error msg, as the name does not suggest that something went wrong. And doing it as you proposed. Your decision.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1859)* status: 0 of 12 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @oshogbo)

___
*[`Documentation/manifest-syntax.rst` line 397 at r2](https://reviewable.io/reviews/gramineproject/gramine/1859#-NxCDN3E0vZW4vDLiSkn:-NxHG1Oe1MH398B3qnat:b-lo7s0) ([raw file](https://github.com/gramineproject/gramine/blob/8820f1035f85a50080b93185782fcb854993ba2f/Documentation/manifest-syntax.rst#L397)):*
> I was thinking about docker seccomp policies, and they are mainly defined as an outside file.

That's simply because there is no "main configuration file" in Docker. Gramine on the other side has such a file, so seems reasonable to put all the configuration in one file. (The only potential exceptions are envvars and cmdline args, but that's because they may be sent in encrypted form.)
___
*[`libos/src/libos_parser.c` line 1653 at r2](https://reviewable.io/reviews/gramineproject/gramine/1859#-NxCDwrT5k_GTCYSoL-B:-NxHHIQC3VeXQbSbegDj:b6186ot) ([raw file](https://github.com/gramineproject/gramine/blob/8820f1035f85a50080b93185782fcb854993ba2f/libos/src/libos_parser.c#L1653)):*

"Beyond this function" I thought about doing it above or where the syscall_parse_table/IBOS_SYSCALL_BOUND is defined. But here makes sense as well.
</blockquote></details>

I looked at where else I can easily add this `static_assert`, but I don't see another good place:
- in `libos_parser.c`: it's already obvious that the size of the table is equal to the bound (from array definition)
- in `libos_defs.h`: there we don't know what `syscall_parser_table` is, so not possible.

So keeping as is.d
___
*[`libos/src/libos_parser.c` line 1658 at r2](https://reviewable.io/reviews/gramineproject/gramine/1859#-NxCDexH848eF4Wr__dd:-NxHGiSCBjd1ozbfLdWi:bb6un1j) ([raw file](https://github.com/gramineproject/gramine/blob/8820f1035f85a50080b93185782fcb854993ba2f/libos/src/libos_parser.c#L1658)):*

Yes, my idea was to not use a LIBOS_SYSCALL_BOUND as a error msg, as the name does not suggest that something went wrong. And doing it as you proposed. Your decision.
</blockquote></details>

Done

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1859)* status: 0 of 14 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

___
*[`Documentation/manifest-syntax.rst` line 397 at r1](https://reviewable.io/reviews/gramineproject/gramine/1859#-NwFEQ29CEoHcLbbz4Bu:-Ny-D2-V4QCsc_GdSU5c:b-opfa1c) ([raw file](https://github.com/gramineproject/gramine/blob/0504ab3a832e24202fea3288d8193de7fe311bfe/Documentation/manifest-syntax.rst#L397)):*

Alternatives:
1. Provide an allowlist `sys.allowed_syscalls = [ ... ]`. This will be more security-expected and more in line with e.g. seccomp profiles. Downside: much harder for end users.
2. Provide both this denylist and the allowlist manifest options. At init time, Gramine fails if detects them both.
</blockquote></details>

Done, see discussion in https://github.com/gramineproject/gramine/discussions/1876
___
*[`Documentation/manifest-syntax.rst` line 399 at r2](https://reviewable.io/reviews/gramineproject/gramine/1859#-NwPCPKt6qX2XH4TOeCT:-Ny-D5ZB6wnL2Md3boVH:b-qt9f3l) ([raw file](https://github.com/gramineproject/gramine/blob/8820f1035f85a50080b93185782fcb854993ba2f/Documentation/manifest-syntax.rst#L399)):*

One additional option would be to not simply return `-ENOSYS`, but allow to specify what int to return:
```
    sys.fixed_syscalls = [
      { name = "syscall_name1", return = -38},  # returns -ENOSYS
      { name = "syscall_name2", return = 0},    # returns 0, i.e. dummy impl
```
</blockquote></details>

Done, replaced with `sys.mocked_syscalls` as proposed during the Gramine meeting: https://github.com/gramineproject/gramine/discussions/1876

llly (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1859)* status: 0 of 14 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`Documentation/manifest-syntax.rst` line 418 at r4](https://reviewable.io/reviews/gramineproject/gramine/1859#-Ny3SHgKAwaAqUUJC9CE:-Ny3SHgKAwaAqUUJC9CF:b-u8kko1) ([raw file](https://github.com/gramineproject/gramine/blob/4a5ce3e99388cd7306ba5be0f61992bf8fbfba36/Documentation/manifest-syntax.rst#L418)):*
> ```reStructuredText
>     sys.mock_syscalls = [
>       { name = "eventfd", return = -38 },
>       { name = "eventfd", return = -38 },
> ```

eventfd2



_Code quote:_
```reStructuredText
eventfd
```
___
*[`Documentation/manifest-syntax.rst` line 428 at r4](https://reviewable.io/reviews/gramineproject/gramine/1859#-Ny3HCcBB8gUC7ONbkb-:-Ny3HCcBB8gUC7ONbkb0:b1nv1yw) ([raw file](https://github.com/gramineproject/gramine/blob/4a5ce3e99388cd7306ba5be0f61992bf8fbfba36/Documentation/manifest-syntax.rst#L428)):*
> ```reStructuredText
>    (2) it is also used to create threads in the same process. The
>    ``sys.disallow_subprocesses`` manifest option disables only the first usage,
>    whereas ``sys.mock_syscalls = [ name = "clone", ...]`` disables both usages.
> ```

{name = "clone"}



_Code quote:_
```reStructuredText
name = "clone"
```

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1859)* status: 0 of 14 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @llly)

___
*[`Documentation/manifest-syntax.rst` line 418 at r4](https://reviewable.io/reviews/gramineproject/gramine/1859#-Ny3SHgKAwaAqUUJC9CE:-Ny4GPhr9tK49RhmQy-i:baqbn6u) ([raw file](https://github.com/gramineproject/gramine/blob/4a5ce3e99388cd7306ba5be0f61992bf8fbfba36/Documentation/manifest-syntax.rst#L418)):*

eventfd2
</blockquote></details>

Done, oops, thanks for catching this!
___
*[`Documentation/manifest-syntax.rst` line 428 at r4](https://reviewable.io/reviews/gramineproject/gramine/1859#-Ny3HCcBB8gUC7ONbkb-:-Ny4G_ZI1AwTdmcvx6QB:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/4a5ce3e99388cd7306ba5be0f61992bf8fbfba36/Documentation/manifest-syntax.rst#L428)):*

{name = "clone"}
</blockquote></details>

Done.

llly (commented):
> Reviewed 13 of 14 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1859)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners

mkow (changes requested):
> Reviewed 10 of 14 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1859)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`-- commits` line 19 at r5](https://reviewable.io/reviews/gramineproject/gramine/1859#-NyPy6_bBuvIkZQtPXIn):*
I think we shouldn't encourage such things, it's not semantically correct (`vhangup()` is not a semantical no-op).
___
*[`Documentation/manifest-syntax.rst` line 406 at r5](https://reviewable.io/reviews/gramineproject/gramine/1859#-NyPzjgO6d2RwFZqYlv4:-NyPzjgO6d2RwFZqYlv5:b-p0tb4o) ([raw file](https://github.com/gramineproject/gramine/blob/a645087851e83dffc85ae9aa7da3510267f4cc18/Documentation/manifest-syntax.rst#L406)):*
> ```reStructuredText
> This syntax specifies the system calls that are mocked when executed in
> Gramine (i.e. they return a specified value without any other side effects).
> If ``return`` field is skipped, then the default value is ``0`` (no-op).
> ```

I'd also warn the users that returning a success but skipping side-effects may introduce bugs to the app, if it expects them (e.g. mocking `futex` may lead to silent introduction of race conditions).
___
*[`libos/src/libos_parser.c` line 1659 at r5](https://reviewable.io/reviews/gramineproject/gramine/1859#-NyQ2Iyk8hrqGPXvJT43:-NyQ2Iyk8hrqGPXvJT44:blyzyvc) ([raw file](https://github.com/gramineproject/gramine/blob/a645087851e83dffc85ae9aa7da3510267f4cc18/libos/src/libos_parser.c#L1659)):*
> ```C
> int get_syscall_number(const char* name, unsigned long* out_sysno) {
>     static_assert(LIBOS_SYSCALL_BOUND == ARRAY_SIZE(syscall_parser_table), "oops");
>     assert(out_sysno);
> ```

Why asserting this, but not asserting `name`? I'd just remove it.
___
*[`libos/src/libos_syscalls.c` line 153 at r5](https://reviewable.io/reviews/gramineproject/gramine/1859#-NyQ7yZY3wjjpkpF9f6q:-NyQ7yZY3wjjpkpF9f6r:bw1vous) ([raw file](https://github.com/gramineproject/gramine/blob/a645087851e83dffc85ae9aa7da3510267f4cc18/libos/src/libos_syscalls.c#L153)):*
> ```C
> 
>         /* add syscall to the table of mocked syscalls */
>         libos_mock_syscall_table[sysno].is_mocked = true;
> ```

Please check if `sysno` fits in the table.

mwkmwkmwk (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1859)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1859#-NyQUuiiBzFZoQoO-qT2:-NyQUuiiBzFZoQoO-qT3:b-vnqsva) (no related file):*
I don't think doing it like this is a good idea for UX reasons: it exposes what is essentially a debug-level option for a user-level need.

The "user-level" need here is executing `sched_yield` within the enclave, which is a fundamentally sound thing to do. It can be a supported option, in the "we provide user support when you use it and something breaks" sense.

However, the proposed generic interface is debug-level: it's a sharp tool that can easily be used to completely break syscall semantics, and make the entire thing unsound. We cannot provide support for using this option.

While I'm not completely against having such debug-level options, I believe they should be clearly marked as such (`sys.debug__mock_syscalls`?), and not recommended for production usage. Separately, I believe we should provide a simple user-facing option like `sys.sched_yield_noop = true`.

___

mkow (commented):
> Reviewed 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1859)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @kailun-qin, and @woju)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1859#-NyQUuiiBzFZoQoO-qT2:-NzVUPr9BHRV1IGGShD5:bpo52gc) (no related file):*
> I don't like the prefix `debug__`, can we use something else? It's more like a `pro__` feature, or maybe even `risky__` or `experimental__`? [@mkow](https://github.com/mkow) [@kailun-qin](https://github.com/kailun-qin) [@woju](https://github.com/woju) Would be good to hear your opinion on the possible name.

Hmm, honestly, I don't like any of them :grimacing:

___

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1859)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @woju)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1859#-NyQUuiiBzFZoQoO-qT2:-NzVzhSC8k2hCF3VrhK_:be8y4dh) (no related file):*

> I don't like the prefix `debug__`, can we use something else? It's more like a `pro__` feature, or maybe even `risky__` or `experimental__`? [@mkow](https://github.com/mkow) [@kailun-qin](https://github.com/kailun-qin) [@woju](https://github.com/woju) Would be good to hear your opinion on the possible name.

Hmm, honestly, I don't like any of them :grimacing:
</blockquote></details>

I slightly prefer `pro__` or `adv__` (to indicate that it's an advaned feature for users w/ a specific level of expertise, and state in the manifest doc that it should be applied with caution).

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1859)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @woju)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1859#-NyQUuiiBzFZoQoO-qT2:-NzYb_Vw8Nc0Z-4URfgE:b-8pnowj) (no related file):*

I slightly prefer `pro__` or `adv__` (to indicate that it's an advaned feature for users w/ a specific level of expertise, and state in the manifest doc that it should be applied with caution).
</blockquote></details>

I think maybe `experimental__`? It is a feature for experimenting, and it's not particularly for production, so the prefix seems to make sense. Besides, we already have other features with this prefix.

___

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1859)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mwkmwkmwk)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1859#-NyQUuiiBzFZoQoO-qT2:-Nz_-DlSDPiX35HRre1w:b-yhh08u) (no related file):*

I think maybe `experimental__`? It is a feature for experimenting, and it's not particularly for production, so the prefix seems to make sense. Besides, we already have other features with this prefix.
</blockquote></details>

`sys.sched_yield_noop = <bool>` is a good way to fix the immediate problem.

However, about generic "mock any syscall interface": I disagree this is not recommended for prod, we already have `sys.ioctl_structs` which is similar in that the user is supposed to specify low level behaviour in a sort of DSL. Similar complexity, although possibly less of a footgun.

I also don't think we should escalate prefixes to options, because what will we do next, `sys.very_risky__`? (And on related note, I heard frontend developers will have `!very-important` in CSS 2025). Having this unprefixed is an OK thing to do, just document this with scary warning in `manifest-syntax.rst` that messing with this option is not something to be done lightly.

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1859)* status: 13 of 14 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow and @mwkmwkmwk)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1859#-NyQUuiiBzFZoQoO-qT2:-NzcXP4V76XtBNRrncAM:b6gay5j) (no related file):*

`sys.sched_yield_noop = <bool>` is a good way to fix the immediate problem.

However, about generic "mock any syscall interface": I disagree this is not recommended for prod, we already have `sys.ioctl_structs` which is similar in that the user is supposed to specify low level behaviour in a sort of DSL. Similar complexity, although possibly less of a footgun.

I also don't think we should escalate prefixes to options, because what will we do next, `sys.very_risky__`? (And on related note, I heard frontend developers will have `!very-important` in CSS 2025). Having this unprefixed is an OK thing to do, just document this with scary warning in `manifest-syntax.rst` that messing with this option is not something to be done lightly.
</blockquote></details>

Done.

I agree with @woju's reasoning. Currently I stick to `sys.mock_syscalls` (without any prefix) and I added a warning in the documentation.

___

mkow (dismissed):
> Reviewed 3 of 14 files at r4, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1859)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @mwkmwkmwk)

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1859)* status: all files reviewed, 3 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mwkmwkmwk)

___
*[`Documentation/manifest-syntax.rst` line 441 at r7](https://reviewable.io/reviews/gramineproject/gramine/1859#-Nzm-hiOERJtzgpTFXEa:-Nzm-hiP1dHifUiMIy4c:bib8cf) ([raw file](https://github.com/gramineproject/gramine/blob/81c8ece993128bcf659ef0d5d6308f026b8dcd15/Documentation/manifest-syntax.rst#L441)):*
> ```reStructuredText
>    implemented in Gramine, and limiting syscalls exposed to the app.
> 
> .. warning ::
> ```

Can you make the enclave insecure using this construct? Like, leak secrets, not just crash. If not, I think this might be `.. caution::`, not `.. warning::` (but N/B). IIUC only *attempting* to use this as sort of sandbox is a security footgun, but this is already the subject of the previous admonition (which is correctly a warning).

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1859)* status: all files reviewed, 3 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @mwkmwkmwk and @woju)

___
*[`Documentation/manifest-syntax.rst` line 441 at r7](https://reviewable.io/reviews/gramineproject/gramine/1859#-Nzm-hiOERJtzgpTFXEa:-NzmQOIE5W2bpwIYbJVU:bqm2joc) ([raw file](https://github.com/gramineproject/gramine/blob/81c8ece993128bcf659ef0d5d6308f026b8dcd15/Documentation/manifest-syntax.rst#L441)):*

Can you make the enclave insecure using this construct? Like, leak secrets, not just crash. If not, I think this might be `.. caution::`, not `.. warning::` (but N/B). IIUC only *attempting* to use this as sort of sandbox is a security footgun, but this is already the subject of the previous admonition (which is correctly a warning).
</blockquote></details>

Yes, I think it could make the enclave insecure. Like, if the application expects some side effects from a certain system call, but this system call is mocked to always return `0` (success), then the app may end up in a weird state and start doing smth weird, including leaking secrets.

Thus, I would prefer to keep the `warning`.

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1859)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mwkmwkmwk)

___
*[`Documentation/manifest-syntax.rst` line 441 at r7](https://reviewable.io/reviews/gramineproject/gramine/1859#-Nzm-hiOERJtzgpTFXEa:-NzmUECV-PepVt3er0VN:b-4g5mqw) ([raw file](https://github.com/gramineproject/gramine/blob/81c8ece993128bcf659ef0d5d6308f026b8dcd15/Documentation/manifest-syntax.rst#L441)):*

Yes, I think it could make the enclave insecure. Like, if the application expects some side effects from a certain system call, but this system call is mocked to always return `0` (success), then the app may end up in a weird state and start doing smth weird, including leaking secrets.

Thus, I would prefer to keep the `warning`.
</blockquote></details>

I understand. If so, then yes, this should be a `.. warning::`.

mwkmwkmwk (commented):
> Reviewed 6 of 14 files at r4, 2 of 3 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1859)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1859)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @mwkmwkmwk)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1859#-NyQUuiiBzFZoQoO-qT2:-O0SucAdCuK97fBAyQD1:b-pxk9xq) (no related file):*

Done.

I agree with @woju's reasoning. Currently I stick to `sys.mock_syscalls` (without any prefix) and I added a warning in the documentation.
</blockquote></details>

Ok, we are currently stuck. May I propose to got with `debug_` prefix, as this one seems most logical to me? @mkow @woju @kailun-qin ?

___

kailun-qin (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1859)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mwkmwkmwk)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1859#-NyQUuiiBzFZoQoO-qT2:-O0T9ZGn6iFOGzaCDyXi:b-ynxyd3) (no related file):*

Ok, we are currently stuck. May I propose to got with `debug_` prefix, as this one seems most logical to me? @mkow @woju @kailun-qin ?
</blockquote></details>

I'm fine w/ the debug prefix :)

___

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1859)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mwkmwkmwk)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1859#-NyQUuiiBzFZoQoO-qT2:-O0fJF7D91fw8Dapnt9c:b-ox1a1v) (no related file):*

I'm fine w/ the debug prefix :)
</blockquote></details>

Fine with me.

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1859)* status: 10 of 14 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @llly, @mkow, and @mwkmwkmwk)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1859#-NyQUuiiBzFZoQoO-qT2:-O0hA-cY2MhZb7WQ_NcT:b-896fix) (no related file):*

Fine with me.
</blockquote></details>

Done.

___
___
*[`-- commits` line 2 at r7](https://reviewable.io/reviews/gramineproject/gramine/1859#-O0h9oU84oDAeAGa6XzO):*
Will do: change everywhere in commit message to `sys.debug__mock_syscalls`

mkow (approved):
> Reviewed 4 of 4 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1859)* status: all files reviewed, 3 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mwkmwkmwk)

mwkmwkmwk (commented):
> Reviewed 2 of 4 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1859)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1859)* status: 7 of 14 files reviewed, 2 unresolved discussions (waiting on @llly, @mkow, and @mwkmwkmwk)

___
*[`-- commits` line 19 at r5](https://reviewable.io/reviews/gramineproject/gramine/1859#-NyPy6_bBuvIkZQtPXIn):*

Will do: I will remove this example from the commit message
</blockquote></details>

Done.
___
*[`-- commits` line 2 at r7](https://reviewable.io/reviews/gramineproject/gramine/1859#-O0h9oU84oDAeAGa6XzO):*

Will do: change everywhere in commit message to `sys.debug__mock_syscalls`
</blockquote></details>

Done.

kailun-qin (approved):
> Reviewed 3 of 14 files at r4, 4 of 4 files at r8, 7 of 7 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1859)* status: all files reviewed, 2 unresolved discussions (waiting on @mkow)

mkow (approved):
> Reviewed 7 of 7 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1859)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

mkow:
Jenkins, retest this please (CI supposedly should work again now)

woju:
Jenkins, test this please

woju:
Jenkins, test this please

woju:
Jenkins, test this please

woju:
Jenkins, test this please

woju:
Jenkins, test this please

woju:
Jenkins, test this please

woju:
Jenkins, test check-python-platlib please

woju:
Jenkins, test pkg-rpm-almalinux9

woju:
Jenkins, test pkg-rpm-almalinux9 please

woju:
Jenkins, test linux-sgx-vm-gcc-release please

woju:
Jenkins, test Jenkins-SGX-EDMM please

woju:
Jenkins, test Jenkins-SGX-Sanitizers please

woju:
Jenkins, test Jenkins-SGX-20.04-apps please

woju:
Jenkins, test Jenkins-SGX-EDMM please

woju:
Jenkins, test Jenkins-SGX-20.04  please

woju:
Jenkins, test Jenkins-SGX-20.04 please

oshogbo:
Jenkins, test Jenkins-SGX-20.04

oshogbo (commented):
Jenkins, test Jenkins-SGX-20.04 please

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1859#-:-O29axHhARX1MaDooW5w:b-4qh941)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

oshogbo:
Jenkins, test Jenkins-SGX-20.04 please

oshogbo:
Jenkins, test Jenkins-SGX-20.04 please

oshogbo:
Jenkins, test Jenkins-SGX-20.04-apps please

PR #1874, opened 2024-05-07
[LibOS] Fix dentry of open file handles related to a rename

g2flyer:
During `rename`, the `dentry` of the corresponding handle is not updated, which e.g., makes `fchown` not work properly.   I did discover that during work on issue #1835 in Draft PR #1856 where i have to be able to figure out stale handles.   A corresponding test-case is added as `test_rename_fchown_fchmod` in `rename_unlink`.  An additional rename related test case `test_rename_follow` test the "flip-side" of `test_rename_replace` (the one which originally triggered me finding the bug).

_BTW: the original (in)validation tests from here are moved now to  PR #1875.  To still have these tests, this PR is currently rebased on the that PRs branch rather than master and currently marked as draft.  Will remove the draft once #1875 is merged and i can rebase to master_ 

## Description of the changes 

To fix that, `do_rename` iterates over open file handles and fixes any which was was pointing to the source-dentry of the rename.

PS: I'm not very "fluent" on the overall locking strategy. I didn't find anything obvious where my code violates invarients but definitely good to cross-check that  (a) using sufficents locks and (b) the implied lock prioties  `g_dcache.lock` >  `libos_handle_map.lock` > `libos_handle.lock` does not violate anything. 

## How to test this PR? 

I tested by adding the two new test-cases and then running the usual regression tests.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1874)

PR #1538, opened 2023-09-06
[LibOS] Fix `ENOENT` error in `fchmod` on unlinked file

sahason:
## Description of the changes 
Fix `ENOENT` error when `fchmod` is called on unlinked file

## How to test this PR? 
`rename_unlink` regression test is extended to validate this implementation.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1538)

PR #1875, opened 2024-05-07
[LibOS] Fix `ENOENT` error in `fchown` on unlinked file

dimakuv:
## Description of the changes 

This is conceptually similar to the commit ["[LibOS] Fix `ENOENT` error in `fchmod` on unlinked file"](https://github.com/gramineproject/gramine/commit/ac8fc770d1004dbcaf43a5b805a7928b7c4fb90f).

This PR was created while reviewing #1874.

## How to test this PR? 

Added a sub-test to the `rename_unlink` LibOS regression test.

If you want to run this sub-test on vanilla Linux, change UID and GID in `fchown()` to some reasonable IDs available on your system.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1875)

### DIFF
--- a/libos/src/sys/libos_file.c
+++ b/libos/src/sys/libos_file.c
@@ -222,7 +222,7 @@ long libos_syscall_fchmod(int fd, mode_t mode) {
     /* This isn't documented, but that's what Linux does. */
     mode_t perm = mode & 07777;
 
-    int ret = 0;
+    int ret;
     if (!hdl->inode) {
         ret = -ENOENT;
         goto out;
@@ -239,6 +239,7 @@ long libos_syscall_fchmod(int fd, mode_t mode) {
     hdl->inode->perm = perm;
     unlock(&hdl->inode->lock);
 
+    ret = 0;
 out:
     put_handle(hdl);
     return ret;
@@ -250,8 +251,6 @@ long libos_syscall_chown(const char* path, uid_t uid, gid_t gid) {
 
 long libos_syscall_fchownat(int dfd, const char* filename, uid_t uid, gid_t gid, int flags) {
     __UNUSED(flags);
-    __UNUSED(uid);
-    __UNUSED(gid);
 
     if (!is_user_string_readable(filename))
         return -EFAULT;
@@ -282,30 +281,23 @@ out:
 }
 
 long libos_syscall_fchown(int fd, uid_t uid, gid_t gid) {
-    __UNUSED(uid);
-    __UNUSED(gid);
-
     struct libos_handle* hdl = get_fd_handle(fd, NULL, NULL);
     if (!hdl)
         return -EBADF;
 
     int ret;
-    struct libos_dentry* dent = hdl->dentry;
-
-    lock(&g_dcache_lock);
-    if (!dent || !dent->inode) {
+    if (!hdl->inode) {
         ret = -ENOENT;
         goto out;
     }
 
-    lock(&dent->inode->lock);
-    dent->inode->uid = (uid == (uid_t)-1) ? dent->inode->uid : uid;
-    dent->inode->gid = (gid == (gid_t)-1) ? dent->inode->gid : gid;
-    unlock(&dent->inode->lock);
+    lock(&hdl->inode->lock);
+    hdl->inode->uid = (uid == (uid_t)-1) ? hdl->inode->uid : uid;
+    hdl->inode->gid = (gid == (gid_t)-1) ? hdl->inode->gid : gid;
+    unlock(&hdl->inode->lock);
 
     ret = 0;
 out:
-    unlock(&g_dcache_lock);
     put_handle(hdl);
     return ret;
 }
--- a/libos/test/regression/meson.build
+++ b/libos/test/regression/meson.build
@@ -99,6 +99,7 @@ tests = {
     'pthread_set_get_affinity': {},
     'readdir': {},
     'rename_unlink': {},
+    'rename_unlink_fchown': {},
     'run_test': {
         'include_directories': include_directories(
             # for `gramine_entry_api.h`
--- a/libos/test/regression/rename_unlink.c
+++ b/libos/test/regression/rename_unlink.c
@@ -1,10 +1,11 @@
 /* SPDX-License-Identifier: LGPL-3.0-or-later */
-/* Copyright (C) 2021 Intel Corporation
+/* Copyright (C) 2024 Intel Corporation
  *                    Paweł Marczewski <pawel@invisiblethingslab.com>
+ *                    Michael Steiner <michael.steiner@intel.com>
  */
 
 /*
- * Tests for renaming and deleting files. Mostly focus on cases where a file is still open.
+ * Tests for renaming and deleting files. Mostly focused on cases where a file is still open.
  */
 
 #define _DEFAULT_SOURCE /* fchmod */
@@ -177,6 +178,46 @@ static void test_rename_replace(const char* path1, const char* path2) {
         err(1, "unlink %s", path2);
 }
 
+static void test_rename_follow(const char* path1, const char* path2) {
+    printf("%s...\n", __func__);
+
+    int fd = create_file(path1, message1, message1_len);
+
+    if (rename(path1, path2) != 0)
+        err(1, "rename");
+
+    should_not_exist(path1);
+    should_exist(path2, message1_len);
+
+    if (lseek(fd, 0, SEEK_SET) != 0)
+        err(1, "lseek");
+
+    ssize_t n = posix_fd_write(fd, message2, message2_len);
+    if (n < 0)
+        errx(1, "posix_fd_write failed");
+    if ((size_t)n != message2_len)
+        errx(1, "wrote less bytes than expected");
+
+    static_assert(sizeof(message1) < sizeof(message2), "message1 must be smaller than message2");
+    should_contain("file opened before it's renamed", fd, message2, message2_len);
+
+    if (close(fd) != 0)
+        err(1, "close %s", path2);
+
+    fd = open(path2, O_RDONLY, 0);
+    if (fd < 0)
+        err(1, "open %s", path2);
+
+    /* we expect `fd` to point to new data, even though we changed data via old fd after rename */
+    should_contain("file opened after it's renamed", fd, message2, message2_len);
+
+    if (close(fd) != 0)
+        err(1, "close %s", path2);
+
+    if (unlink(path2) != 0)
+        err(1, "unlink %s", path2);
+}
+
 static void test_rename_open_file(const char* path1, const char* path2) {
     printf("%s...\n", __func__);
 
@@ -271,6 +312,7 @@ int main(int argc, char* argv[]) {
     test_rename_same_file(path1);
     test_simple_rename(path1, path2);
     test_rename_replace(path1, path2);
+    test_rename_follow(path1, path2);
     test_rename_open_file(path1, path2);
     test_unlink_and_recreate(path1);
     test_unlink_and_write(path1);
--- a/libos/test/regression/rename_unlink_fchown.c
+++ b/libos/test/regression/rename_unlink_fchown.c
@@ -1,0 +1,139 @@
+/* SPDX-License-Identifier: LGPL-3.0-or-later */
+/* Copyright (C) 2024 Intel Corporation
+ *                    Michael Steiner <michael.steiner@intel.com>
+ */
+
+/*
+ * Tests for fchown after renaming and deleting files. Mostly focused on cases where a file is still
+ * open. These tests are separate from other renaming/deleting tests in `rename_unlink.c` because
+ * these tests require a root user to perform fchown with arbitrary user/group.
+ */
+
+#define _DEFAULT_SOURCE /* fchmod */
+
+#include <assert.h>
+#include <err.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+#include "common.h"
+#include "rw_file.h"
+
+static const char message[] = "first message\n";
+static const size_t message_len = sizeof(message) - 1;
+
+static void should_not_exist(const char* path) {
+    struct stat statbuf;
+
+    if (stat(path, &statbuf) == 0)
+        errx(1, "%s unexpectedly exists", path);
+    if (errno != ENOENT)
+        err(1, "stat %s", path);
+}
+
+static void check_statbuf(const char* desc, struct stat* statbuf, size_t size) {
+    assert(!OVERFLOWS(off_t, size));
+
+    if (!S_ISREG(statbuf->st_mode))
+        errx(1, "%s: wrong mode (0o%o)", desc, statbuf->st_mode);
+    if (statbuf->st_size != (off_t)size)
+        errx(1, "%s: wrong size (%lu)", desc, statbuf->st_size);
+}
+
+static void check_ownership_and_permissions(const char* path, struct stat* statbuf, uid_t uid,
+                                            gid_t gid, mode_t permissions) {
+    if (statbuf->st_uid != uid || statbuf->st_gid != gid)
+        errx(1, "wrong ownership of file %s", path);
+    if ((statbuf->st_mode & 0777) != permissions)
+        errx(1, "wrong permissions of file %s", path);
+}
+
+static void should_exist(const char* path, size_t size) {
+    struct stat statbuf;
+    CHECK(stat(path, &statbuf));
+    check_statbuf(path, &statbuf, size);
+}
+
+static int create_file(const char* path, const char* str, size_t len) {
+    int fd = CHECK(open(path, O_RDWR | O_CREAT | O_TRUNC, 0600));
+
+    ssize_t n = posix_fd_write(fd, str, len);
+    if (n < 0)
+        errx(1, "posix_fd_write %s", path);
+    if ((size_t)n != len)
+        errx(1, "written less bytes than expected into %s", path);
+
+    return fd;
+}
+
+static void test_rename_fchown_fchmod(const char* path1, const char* path2) {
+    printf("%s...\n", __func__);
+
+    int fd = create_file(path1, message, message_len);
+
+    if (fchown(fd, /*owner=*/123, /*group=*/123) != 0) /* dummy owner/group just for testing */
+        err(1, "fchown before rename");
+    if (fchmod(fd, 0660) != 0) /* note: no "other users" mode bits */
+        err(1, "fchmod before rename");
+
+    struct stat st;
+    CHECK(stat(path1, &st));
+    check_ownership_and_permissions(path1, &st, /*uid=*/123, /*gid=*/123, /*permissions=*/0660);
+
+    CHECK(rename(path1, path2));
+
+    should_not_exist(path1);
+    should_exist(path2, message_len);
+
+    if (fchown(fd, /*owner=*/321, /*group=*/321) != 0) /* different dummy owner/group */
+        err(1, "fchown after rename");
+    if (fchmod(fd, 0666) != 0) /* note: now with "other users" mode bits */
+        err(1, "fchmod after rename");
+
+    CHECK(stat(path2, &st));
+    check_ownership_and_permissions(path2, &st, /*uid=*/321, /*gid=*/321, /*permissions=*/0666);
+
+    CHECK(close(fd));
+    CHECK(unlink(path2));
+}
+
+static void test_unlink_fchown(const char* path) {
+    printf("%s...\n", __func__);
+
+    int fd = create_file(path, /*message=*/NULL, /*len=*/0);
+
+    CHECK(unlink(path));
+    should_not_exist(path);
+
+    if (fchown(fd, /*owner=*/123, /*group=*/123) != 0) /* dummy owner/group just for testing */
+        err(1, "fchown after file removal");
+
+    /* note that file was created with permissions 0600, see create_file() */
+    struct stat st;
+    CHECK(fstat(fd, &st));
+    check_ownership_and_permissions(path, &st, /*uid=*/123, /*gid=*/123, /*permissions=*/0600);
+
+    CHECK(close(fd));
+}
+
+int main(int argc, char* argv[]) {
+    setbuf(stdout, NULL);
+    setbuf(stderr, NULL);
+
+    if (argc != 3)
+        errx(1, "Usage: %s <path1> <path2>", argv[0]);
+
+    const char* path1 = argv[1];
+    const char* path2 = argv[2];
+
+    test_rename_fchown_fchmod(path1, path2);
+    test_unlink_fchown(path1);
+    puts("TEST OK");
+    return 0;
+}
--- a/libos/test/regression/test_libos.py
+++ b/libos/test/regression/test_libos.py
@@ -705,7 +705,7 @@ class TC_30_Syscall(RegressionTestCase):
 
         self.assertIn('TEST OK', stdout)
 
-    def test_033_rename_unlink_chroot(self):
+    def test_033a_rename_unlink_chroot(self):
         file1 = 'tmp/file1'
         file2 = 'tmp/file2'
         try:
@@ -716,7 +716,7 @@ class TC_30_Syscall(RegressionTestCase):
                     os.unlink(path)
         self.assertIn('TEST OK', stdout)
 
-    def test_034_rename_unlink_pf(self):
+    def test_033b_rename_unlink_pf(self):
         os.makedirs('tmp/pf', exist_ok=True)
         file1 = 'tmp/pf/file1'
         file2 = 'tmp/pf/file2'
@@ -728,7 +728,7 @@ class TC_30_Syscall(RegressionTestCase):
                     os.unlink(path)
         self.assertIn('TEST OK', stdout)
 
-    def test_035_rename_unlink_enc(self):
+    def test_033c_rename_unlink_enc(self):
         os.makedirs('tmp_enc', exist_ok=True)
         file1 = 'tmp_enc/file1'
         file2 = 'tmp_enc/file2'
@@ -744,10 +744,55 @@ class TC_30_Syscall(RegressionTestCase):
                     os.unlink(path)
         self.assertIn('TEST OK', stdout)
 
-    def test_036_rename_unlink_tmpfs(self):
+    def test_033d_rename_unlink_tmpfs(self):
         file1 = '/mnt/tmpfs/file1'
         file2 = '/mnt/tmpfs/file2'
         stdout, _ = self.run_binary(['rename_unlink', file1, file2])
+        self.assertIn('TEST OK', stdout)
+
+    def test_034a_rename_unlink_fchown_chroot(self):
+        file1 = 'tmp/file1'
+        file2 = 'tmp/file2'
+        try:
+            stdout, _ = self.run_binary(['rename_unlink_fchown', file1, file2])
+        finally:
+            for path in [file1, file2]:
+                if os.path.exists(path):
+                    os.unlink(path)
+        self.assertIn('TEST OK', stdout)
+
+    def test_034b_rename_unlink_fchown_pf(self):
+        os.makedirs('tmp/pf', exist_ok=True)
+        file1 = 'tmp/pf/file1'
+        file2 = 'tmp/pf/file2'
+        try:
+            stdout, _ = self.run_binary(['rename_unlink_fchown', file1, file2])
+        finally:
+            for path in [file1, file2]:
+                if os.path.exists(path):
+                    os.unlink(path)
+        self.assertIn('TEST OK', stdout)
+
+    def test_034c_rename_unlink_fchown_enc(self):
+        os.makedirs('tmp_enc', exist_ok=True)
+        file1 = 'tmp_enc/file1'
+        file2 = 'tmp_enc/file2'
+        # Delete the files: the test overwrites them anyway, but it may fail if they are malformed.
+        for path in [file1, file2]:
+            if os.path.exists(path):
+                os.unlink(path)
+        try:
+            stdout, _ = self.run_binary(['rename_unlink_fchown', file1, file2])
+        finally:
+            for path in [file1, file2]:
+                if os.path.exists(path):
+                    os.unlink(path)
+        self.assertIn('TEST OK', stdout)
+
+    def test_034d_rename_unlink_fchown_tmpfs(self):
+        file1 = '/mnt/tmpfs/file1'
+        file2 = '/mnt/tmpfs/file2'
+        stdout, _ = self.run_binary(['rename_unlink_fchown', file1, file2])
         self.assertIn('TEST OK', stdout)
 
     def test_040_futex_bitset(self):
--- a/libos/test/regression/tests.toml
+++ b/libos/test/regression/tests.toml
@@ -98,6 +98,7 @@ manifests = [
   "pthread_set_get_affinity",
   "readdir",
   "rename_unlink",
+  "rename_unlink_fchown",
   "run_test",
   "rwlock",
   "sched",
--- a/libos/test/regression/tests_musl.toml
+++ b/libos/test/regression/tests_musl.toml
@@ -100,6 +100,7 @@ manifests = [
   "pthread_set_get_affinity",
   "readdir",
   "rename_unlink",
+  "rename_unlink_fchown",
   "run_test",
   "rwlock",
   "sched",


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1875)* status: 0 of 2 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1875#-NxHW3bq2L3DQQdzZWid:-NxHW3bq2L3DQQdzZWie:b-87uk1a) (no related file):*
FYI: I double checked that vanilla Linux kernel correctly `fchown()`s the file even after it was unlinked.

___
*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1875#-NxHWBo80gDPnq1b1mly:-NxHWBo80gDPnq1b1mlz:bz5sk3n) (no related file):*
This PR is modelled exactly after #1538.

___
___
*[`libos/src/sys/libos_file.c` line 242 at r1](https://reviewable.io/reviews/gramineproject/gramine/1875#-NxHWMSCCCepaKNsPLlp:-NxHWMSCCCepaKNsPLlq:baffvoz) ([raw file](https://github.com/gramineproject/gramine/blob/dfd5621da4bfc9ac32381c4a85181f4d687c2ee9/libos/src/sys/libos_file.c#L242)):*
> ```C
>     unlock(&hdl->inode->lock);
> 
>     ret = 0;
> ```

FYI: This is just a random improvement of the related code.

oshogbo (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1875#-:-NxHYEug9Z1dC5WomexN:bnfp4nl)* status: 0 of 2 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

g2flyer (commented):
As mentioned in discussion of #1874 i can confirm that this PR fixes the `fchown` issue identified in #1874 and in more minimal (and complete :-)) ways than #1874.   I wonder though whether the new test-vectors from #1874 might be better add directly to this PR rather than in a separate one?

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1875#-:-NxJSSEo866he3HRBCBt:bs7hdz9)* status: 0 of 2 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

dimakuv (commented):
Done, added your sub-tests too.

I force-pushed since Mariusz finished his review, others didn't start it, and I wanted to not forget to add Michael Steiner as a co-author in the commit message.

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1875#-:-NxLyLmEE8iw1iWI9Xgy:b34eekx)* status: 0 of 2 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

mkow (changes requested):
> Reviewed 1 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1875)* status: 1 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`libos/test/regression/rename_unlink.c` line 230 at r2](https://reviewable.io/reviews/gramineproject/gramine/1875#-NyLucKb20Eg7TbLrxsi:-NyLucKb20Eg7TbLrxsj:b-ahgd34) ([raw file](https://github.com/gramineproject/gramine/blob/642b21c6d387caf7659013ec347b1947518e270b/libos/test/regression/rename_unlink.c#L230)):*
> ```C
>         err(1, "create %s", path1);
> 
>     if (fchown(fd, /*owner=*/123, /*group=*/123) != 0) /* dummy owner/group just for testing */
> ```

This fails when run natively, could you change this test somehow so that it also works natively? It would make it easier to verify whether everything in Gramine works exactly as on Linux.

g2flyer:
> This fails when run natively, could you change this test somehow so that it also works natively? It would make it easier to verify whether everything in Gramine works exactly as on Linux.

In my original version from that did run natively as long as you run it as root (i actually had also a corresponding comment) and i think that is somewhat unavoidable.  I've used different ids but i think it still works with 123, the ids do not necessarily have to exist in /etc/passwd et al for it to work.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1875)* status: 1 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`libos/test/regression/rename_unlink.c` line 230 at r2](https://reviewable.io/reviews/gramineproject/gramine/1875#-NyLucKb20Eg7TbLrxsi:-NyM-SAsCnWDvpz9AHG1:bruidd3) ([raw file](https://github.com/gramineproject/gramine/blob/642b21c6d387caf7659013ec347b1947518e270b/libos/test/regression/rename_unlink.c#L230)):*
@g2flyer: Please reply using Reviewable, so that your reply lands in a proper thread ;)

> In my original version from that did run natively as long as you run it as root (i actually had also a corresponding comment) and i think that is somewhat unavoidable. I've used different ids but i think it still works with 123, the ids do not necessarily have to exist in `/etc/passwd` et al for it to work.

In the future we'd like to run these tests in CI also natively, and that requirement would prevent this. It should be possible to change it, maybe `chown()` to the current user? (no-op, but should still trigger all the logic).

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1875)* status: 1 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`libos/test/regression/rename_unlink.c` line 230 at r2](https://reviewable.io/reviews/gramineproject/gramine/1875#-NyLucKb20Eg7TbLrxsi:-NyMnAsc6xmraO0SZNYv:bqhimay) ([raw file](https://github.com/gramineproject/gramine/blob/642b21c6d387caf7659013ec347b1947518e270b/libos/test/regression/rename_unlink.c#L230)):*
> We don't run CI containerized? Otherwise running as "root" is not really a problem? Using your own id (and essentially not changing the ids) wouldn't really work, e.g., in `test_rename_fchown_fchmod` to detect all potential issues? In particular, i don't think you would have detected the original bug?

Ah, right, then it should work in CI, it will just be annoying when running the tests manually, outside of CI.
@dimakuv, @woju, @kailun-qin, @oshogbo: what do you think? Especially, is there a better way to implement this test?

@g2flyer: and please again, use Reviewable for comments :) When you comment on GitHub your comments are not linked to the discussion thread and land in a different place, without the context.

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1875)* status: 1 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv, @g2flyer, @kailun-qin, and @oshogbo)

___
*[`libos/test/regression/rename_unlink.c` line 230 at r2](https://reviewable.io/reviews/gramineproject/gramine/1875#-NyLucKb20Eg7TbLrxsi:-NyPNG7-3oLV0i1E0d-I:bj3uqev) ([raw file](https://github.com/gramineproject/gramine/blob/642b21c6d387caf7659013ec347b1947518e270b/libos/test/regression/rename_unlink.c#L230)):*

> We don't run CI containerized? Otherwise running as "root" is not really a problem? Using your own id (and essentially not changing the ids) wouldn't really work, e.g., in `test_rename_fchown_fchmod` to detect all potential issues? In particular, i don't think you would have detected the original bug?

Ah, right, then it should work in CI, it will just be annoying when running the tests manually, outside of CI.
@dimakuv, @woju, @kailun-qin, @oshogbo: what do you think? Especially, is there a better way to implement this test?

@g2flyer: and please again, use Reviewable for comments :) When you comment on GitHub your comments are not linked to the discussion thread and land in a different place, without the context.
</blockquote></details>

If you're non-root (and without `CAP_FOWNER`), `fchmod` should succeed if gid matches current euid and that euid (user) is a member of the group that is the param of the operation. So you could make it this way:

1) `geteuid()`
2) if euid is `0` pick something, maybe 65534:65534 (nobody:nobody)
3) `getgroups()`
4) if there is only one group and matches the existing file, skip the test,
5) pick first group that does not match the group of the target file,
6) `fchown(fd, geteuid(), <group from point 5>)`.

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1875)* status: 1 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv, @g2flyer, @kailun-qin, and @oshogbo)

___
*[`libos/test/regression/rename_unlink.c` line 230 at r2](https://reviewable.io/reviews/gramineproject/gramine/1875#-NyLucKb20Eg7TbLrxsi:-NyQRQBX3nEtttiDhvSg:b6z9on) ([raw file](https://github.com/gramineproject/gramine/blob/642b21c6d387caf7659013ec347b1947518e270b/libos/test/regression/rename_unlink.c#L230)):*

If you're non-root (and without `CAP_FOWNER`), `fchmod` should succeed if gid matches current euid and that euid (user) is a member of the group that is the param of the operation. So you could make it this way:

1) `geteuid()`
2) if euid is `0` pick something, maybe 65534:65534 (nobody:nobody)
3) `getgroups()`
4) if there is only one group and matches the existing file, skip the test,
5) pick first group that does not match the group of the target file,
6) `fchown(fd, geteuid(), <group from point 5>)`.
</blockquote></details>

... if *uid* matches current euid

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1875)* status: 1 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @g2flyer, @kailun-qin, and @oshogbo)

___
*[`libos/test/regression/rename_unlink.c` line 230 at r2](https://reviewable.io/reviews/gramineproject/gramine/1875#-NyLucKb20Eg7TbLrxsi:-NzSrFIW09XonIahv0Hy:b-2a7ke7) ([raw file](https://github.com/gramineproject/gramine/blob/642b21c6d387caf7659013ec347b1947518e270b/libos/test/regression/rename_unlink.c#L230)):*

... if *uid* matches current euid
</blockquote></details>

@mkow We already have a bunch of tests like this, that don't run on native Linux... E.g. tests that verify our synthetic values inside `/proc`, tests that operate on UIDs/GIDs, etc.

Running this test manually would require adding 123 and 321 users. Wouldn't that be enough for manual tests, to ask the user to create such users (possible in a separate user namespace).

g2flyer:
> Running this test manually would require adding 123 and 321 users. Wouldn't that be enough for manual tests, to ask the user to create such users (possible in a separate user namespace).

You don't need to create these users (in sense of adding them to passwd/groups/shadow and home directories et al).  It still works even if they do not exist as long as you run them as root (which seems kindof conceptually required as these test-cases really make sense mostly for root/setuid commands).    (I validated it for this program but i've seen that also in lots of past cases when you untar tar-files with unknown identities/groups ...)

g2flyer (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1875)* status: 1 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @kailun-qin and @oshogbo)

___
*[`libos/test/regression/rename_unlink.c` line 230 at r2](https://reviewable.io/reviews/gramineproject/gramine/1875#-NyLucKb20Eg7TbLrxsi:-NzTqYjM3SldOLmBPetu:b-y3osmx) ([raw file](https://github.com/gramineproject/gramine/blob/642b21c6d387caf7659013ec347b1947518e270b/libos/test/regression/rename_unlink.c#L230)):*

@mkow We already have a bunch of tests like this, that don't run on native Linux... E.g. tests that verify our synthetic values inside `/proc`, tests that operate on UIDs/GIDs, etc.

Running this test manually would require adding 123 and 321 users. Wouldn't that be enough for manual tests, to ask the user to create such users (possible in a separate user namespace).
</blockquote></details>

You don't need to create these users (in sense of adding them to passwd/groups/shadow and home directories et al).  It still works even if they do not exist as long as you run them as root (which seems kindof conceptually required as these test-cases really make sense mostly for root/setuid commands).    (I validated it for this program but i've seen that also in lots of past cases when you untar tar-files with unknown identities/groups ...)

PS: sorry for first replying directly in github: Still too much muscle memories from other projects where you just click in email and directly reply from github ...

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1875)* status: 1 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @kailun-qin and @oshogbo)

___
*[`libos/test/regression/rename_unlink.c` line 230 at r2](https://reviewable.io/reviews/gramineproject/gramine/1875#-NyLucKb20Eg7TbLrxsi:-NzcXtJ22C-hnCG2-hTu:b-qfaz9y) ([raw file](https://github.com/gramineproject/gramine/blob/642b21c6d387caf7659013ec347b1947518e270b/libos/test/regression/rename_unlink.c#L230)):*
<details><summary><i>Previously, g2flyer (Michael Steiner) wrote…</i></summary><blockquote>

You don't need to create these users (in sense of adding them to passwd/groups/shadow and home directories et al).  It still works even if they do not exist as long as you run them as root (which seems kindof conceptually required as these test-cases really make sense mostly for root/setuid commands).    (I validated it for this program but i've seen that also in lots of past cases when you untar tar-files with unknown identities/groups ...)

PS: sorry for first replying directly in github: Still too much muscle memories from other projects where you just click in email and directly reply from github ...
</blockquote></details>

I agree with @g2flyer -- this test is conceptually a root-user test. So it will work even on vanilla Linux (but under root). This is similar to how certain LTP tests require running them as root.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1875)* status: 1 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @kailun-qin and @oshogbo)

___
*[`libos/test/regression/rename_unlink.c` line 230 at r2](https://reviewable.io/reviews/gramineproject/gramine/1875#-NyLucKb20Eg7TbLrxsi:-NzedB9C8djK0nHI09OT:b-1pqpea) ([raw file](https://github.com/gramineproject/gramine/blob/642b21c6d387caf7659013ec347b1947518e270b/libos/test/regression/rename_unlink.c#L230)):*
> [@mkow](https://github.com/mkow) We already have a bunch of tests like this, that don't run on native Linux... E.g. tests that verify our synthetic values inside `/proc`, tests that operate on UIDs/GIDs, etc.

Yeah, but these are explicitly testing something Gramine-only or explicitly root-only. This is a generic rename and unlink test.

> I agree with [@g2flyer](https://github.com/g2flyer) -- this test is conceptually a root-user test. So it will work even on vanilla Linux (but under root).

My idea/question was to change this test so it still tests this functionality, but in some other way which doesn't require root. So, it's not possible? Is woju's solution is too complicated?

g2flyer (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1875)* status: 1 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @kailun-qin and @oshogbo)

___
*[`libos/test/regression/rename_unlink.c` line 230 at r2](https://reviewable.io/reviews/gramineproject/gramine/1875#-NyLucKb20Eg7TbLrxsi:-Nzer432-f70czpoRZiu:b-e45wi) ([raw file](https://github.com/gramineproject/gramine/blob/642b21c6d387caf7659013ec347b1947518e270b/libos/test/regression/rename_unlink.c#L230)):*

> [@mkow](https://github.com/mkow) We already have a bunch of tests like this, that don't run on native Linux... E.g. tests that verify our synthetic values inside `/proc`, tests that operate on UIDs/GIDs, etc.

Yeah, but these are explicitly testing something Gramine-only or explicitly root-only. This is a generic rename and unlink test.

> I agree with [@g2flyer](https://github.com/g2flyer) -- this test is conceptually a root-user test. So it will work even on vanilla Linux (but under root).

My idea/question was to change this test so it still tests this functionality, but in some other way which doesn't require root. So, it's not possible? Is woju's solution is too complicated?
</blockquote></details>

What Wojtek outlines seems the best you can get as non-root. That said, is that really worth the additional complexity/obscuring of the actual test you want to check and the potential risk that due to other issues you actually might only run restricted tests in the actual CI?  My take would have been that these tests you run natively exactly and only to initially validate that they correctly interpret POSIX behavior (at least Linux's interpretation thereof ;-)  and afterwards wouldn't really touch them anymore?  So should really happen only rarely and then running as root is really not a problem?   But I guess not worth a hill to die on, this discussion probably already took way longer than it should have  ...

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1875)* status: 1 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @kailun-qin and @oshogbo)

___
*[`libos/test/regression/rename_unlink.c` line 230 at r2](https://reviewable.io/reviews/gramineproject/gramine/1875#-NyLucKb20Eg7TbLrxsi:-NzgWPvlEM--7E1W-dqL:b-c6ytu8) ([raw file](https://github.com/gramineproject/gramine/blob/642b21c6d387caf7659013ec347b1947518e270b/libos/test/regression/rename_unlink.c#L230)):*
<details><summary><i>Previously, g2flyer (Michael Steiner) wrote…</i></summary><blockquote>

What Wojtek outlines seems the best you can get as non-root. That said, is that really worth the additional complexity/obscuring of the actual test you want to check and the potential risk that due to other issues you actually might only run restricted tests in the actual CI?  My take would have been that these tests you run natively exactly and only to initially validate that they correctly interpret POSIX behavior (at least Linux's interpretation thereof ;-)  and afterwards wouldn't really touch them anymore?  So should really happen only rarely and then running as root is really not a problem?   But I guess not worth a hill to die on, this discussion probably already took way longer than it should have  ...
</blockquote></details>

I agree with @g2flyer -- it feels too complicated. I's prefer to keep it simple. I guess we could also just assume that native versions will run in a "root" Docker container (then no changes to the test are needed)? This seems like a fair and simple assumption.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1875)* status: 1 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @kailun-qin and @oshogbo)

___
*[`libos/test/regression/rename_unlink.c` line 230 at r2](https://reviewable.io/reviews/gramineproject/gramine/1875#-NyLucKb20Eg7TbLrxsi:-Nzjq2gX7uz53Sx6HxWa:bxc2ov6) ([raw file](https://github.com/gramineproject/gramine/blob/642b21c6d387caf7659013ec347b1947518e270b/libos/test/regression/rename_unlink.c#L230)):*
> My take would have been that these tests you run natively exactly and only to initially validate that they correctly interpret POSIX behavior (at least Linux's interpretation thereof ;-) and afterwards wouldn't really touch them anymore?

That's what we do right now (or at least me when reviewing new tests), but the tests are being changed from time to time and it's always better if you can assert some properties automatically in CI.
So, it's a trade-off between the amount of stuff tested in-gramine and how much native compatibility we can assert in CI.

One idea to improve the trade-off in this PR: maybe you could refactor-out the root-only tests from this test to a separate one? The subtests from the old `rename_unlink` test didn't require root, so it would be nice to be able to still run them without root.

> I guess we could also just assume that native versions will run in a "root" Docker container (then no changes to the test are needed)? This seems like a fair and simple assumption.

Dunno if that will work well, but it's some idea to try. @oshogbo is looking into this, he can probably say something more on that.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1875)* status: 1 of 7 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin and @oshogbo)

___
*[`libos/test/regression/rename_unlink.c` line 230 at r2](https://reviewable.io/reviews/gramineproject/gramine/1875#-NyLucKb20Eg7TbLrxsi:-NzleLeo7fWpiMo8F2ug:bxipktr) ([raw file](https://github.com/gramineproject/gramine/blob/642b21c6d387caf7659013ec347b1947518e270b/libos/test/regression/rename_unlink.c#L230)):*
> One idea to improve the trade-off in this PR: maybe you could refactor-out the root-only tests from this test to a separate one?

Good idea, done.

Please note that there is one sub-test `test_rename_follow` that was added in this PR, but this test actually doesn't have to do anything with `fchown` support (the point of this PR). So I don't know -- should I create a separate PR that adds this `test_rename_follow` sub-test separately? Or we can close our eyes on this and merge as-is?

mkow (changes requested):
> Reviewed 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1875)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`libos/test/regression/rename_unlink.c` line 230 at r2](https://reviewable.io/reviews/gramineproject/gramine/1875#-NyLucKb20Eg7TbLrxsi:-Nzn32DC5ZsAy8yLNbJ4:b-ngqvsq) ([raw file](https://github.com/gramineproject/gramine/blob/642b21c6d387caf7659013ec347b1947518e270b/libos/test/regression/rename_unlink.c#L230)):*

> One idea to improve the trade-off in this PR: maybe you could refactor-out the root-only tests from this test to a separate one?

Good idea, done.

Please note that there is one sub-test `test_rename_follow` that was added in this PR, but this test actually doesn't have to do anything with `fchown` support (the point of this PR). So I don't know -- should I create a separate PR that adds this `test_rename_follow` sub-test separately? Or we can close our eyes on this and merge as-is?
</blockquote></details>

I'm fine with keeping it here, it's just a test.
___
*[`libos/test/regression/rename_unlink.c` line 203 at r3](https://reviewable.io/reviews/gramineproject/gramine/1875#-Nzn5C1d3tFDn73lZHS2:-Nzn5C1d3tFDn73lZHS3:b-klx10y) ([raw file](https://github.com/gramineproject/gramine/blob/042a1401daac3cfe07ab291f730308d43ff8ceba/libos/test/regression/rename_unlink.c#L203)):*
> ```C
>         errx(1, "wrote less bytes than expected");
> 
>     should_contain("file opened before it's renamed", fd, message2, message2_len);
> ```

Please `static_assert` that `message2` is longer than `message1` (otherwise this check doesn't work, right? `message1` won't be fully overwritten).
___
*[`libos/test/regression/rename_unlink_fchown.c` line 7 at r3](https://reviewable.io/reviews/gramineproject/gramine/1875#-NzpFAMB-RUxD_0gh1PG:-NzpFAMB-RUxD_0gh1PH:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/042a1401daac3cfe07ab291f730308d43ff8ceba/libos/test/regression/rename_unlink_fchown.c#L7)):*
> ```C
> 
> /*
>  * Tests for fchown after renaming and deleting files. Mostly focus on cases where a file is still
> ```





_Suggestion:_
```C
Mostly focused
```
___
*[`libos/test/regression/rename_unlink_fchown.c` line 28 at r3](https://reviewable.io/reviews/gramineproject/gramine/1875#-NzpJr9SB9QB3KssUVUB:-NzpJr9T2a0SYCwvV64P:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/042a1401daac3cfe07ab291f730308d43ff8ceba/libos/test/regression/rename_unlink_fchown.c#L28)):*
> ```C
> #include "rw_file.h"
> 
> static const char message1[] = "first message\n";
> ```





_Suggestion:_
```C
message
```
___
*[`libos/test/regression/rename_unlink_fchown.c` line 29 at r3](https://reviewable.io/reviews/gramineproject/gramine/1875#-NzpJuHN69orHoOFpc9h:-NzpJuHN69orHoOFpc9i:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/042a1401daac3cfe07ab291f730308d43ff8ceba/libos/test/regression/rename_unlink_fchown.c#L29)):*
> ```C
> 
> static const char message1[] = "first message\n";
> static const size_t message1_len = sizeof(message1) - 1;
> ```





_Suggestion:_
```C
message_len
```
___
*[`libos/test/regression/rename_unlink_fchown.c` line 89 at r3](https://reviewable.io/reviews/gramineproject/gramine/1875#-NzpIwmc9ze3ncFCA6wY:-NzpIwmc9ze3ncFCA6wZ:bpw5t3l) ([raw file](https://github.com/gramineproject/gramine/blob/042a1401daac3cfe07ab291f730308d43ff8ceba/libos/test/regression/rename_unlink_fchown.c#L89)):*
> ```C
>     if (st.st_uid != 123 || st.st_gid != 123)
>         err(1, "wrong ownership of file %s", path1);
>     if (st.st_mode & S_IRWXO)
> ```

That's a weird check, why not checking whether the perms match _exactly_ what the file was chmoded to?
___
*[`libos/test/regression/rename_unlink_fchown.c` line 107 at r3](https://reviewable.io/reviews/gramineproject/gramine/1875#-NzpJ61z-8_vijtl1XQb:-NzpJ61z-8_vijtl1XQc:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/042a1401daac3cfe07ab291f730308d43ff8ceba/libos/test/regression/rename_unlink_fchown.c#L107)):*
> ```C
>     if (st.st_uid != 321 || st.st_gid != 321)
>         err(1, "wrong ownership of (renamed) file %s", path2);
>     if (!(st.st_mode & S_IRWXO))
> ```

ditto
___
*[`libos/test/regression/rename_unlink_fchown.c` line 129 at r3](https://reviewable.io/reviews/gramineproject/gramine/1875#-NzpJW4e72RKT9UnLy2u:-NzpJW4e72RKT9UnLy2v:b-vbg670) ([raw file](https://github.com/gramineproject/gramine/blob/042a1401daac3cfe07ab291f730308d43ff8ceba/libos/test/regression/rename_unlink_fchown.c#L129)):*
> ```C
>     if (fchown(fd, /*owner=*/123, /*group=*/123) != 0) /* dummy owner/group just for testing */
>         err(1, "fchown");
> ```

Could you verify with `fstat` that the perms were actually changed?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1875)* status: 5 of 7 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`libos/test/regression/rename_unlink.c` line 203 at r3](https://reviewable.io/reviews/gramineproject/gramine/1875#-Nzn5C1d3tFDn73lZHS2:-Nzt03I7DTZM6g42YFwY:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/042a1401daac3cfe07ab291f730308d43ff8ceba/libos/test/regression/rename_unlink.c#L203)):*

Please `static_assert` that `message2` is longer than `message1` (otherwise this check doesn't work, right? `message1` won't be fully overwritten).
</blockquote></details>

Done.
___
*[`libos/test/regression/rename_unlink_fchown.c` line 7 at r3](https://reviewable.io/reviews/gramineproject/gramine/1875#-NzpFAMB-RUxD_0gh1PG:-Nzt0FV4-uI6agnOFQW9:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/042a1401daac3cfe07ab291f730308d43ff8ceba/libos/test/regression/rename_unlink_fchown.c#L7)):*
> ```C
> 
> /*
>  * Tests for fchown after renaming and deleting files. Mostly focus on cases where a file is still
> ```

Done.
___
*[`libos/test/regression/rename_unlink_fchown.c` line 28 at r3](https://reviewable.io/reviews/gramineproject/gramine/1875#-NzpJr9SB9QB3KssUVUB:-Nzt0MNF6qKvofMGBAln:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/042a1401daac3cfe07ab291f730308d43ff8ceba/libos/test/regression/rename_unlink_fchown.c#L28)):*
> ```C
> #include "rw_file.h"
> 
> static const char message1[] = "first message\n";
> ```

Done.
___
*[`libos/test/regression/rename_unlink_fchown.c` line 29 at r3](https://reviewable.io/reviews/gramineproject/gramine/1875#-NzpJuHN69orHoOFpc9h:-Nzt0MjI-a077CJSH7bC:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/042a1401daac3cfe07ab291f730308d43ff8ceba/libos/test/regression/rename_unlink_fchown.c#L29)):*
> ```C
> 
> static const char message1[] = "first message\n";
> static const size_t message1_len = sizeof(message1) - 1;
> ```

Done.
___
*[`libos/test/regression/rename_unlink_fchown.c` line 89 at r3](https://reviewable.io/reviews/gramineproject/gramine/1875#-NzpIwmc9ze3ncFCA6wY:-Nzt1sId8QWl82I1CdKW:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/042a1401daac3cfe07ab291f730308d43ff8ceba/libos/test/regression/rename_unlink_fchown.c#L89)):*

That's a weird check, why not checking whether the perms match _exactly_ what the file was chmoded to?
</blockquote></details>

Done.
___
*[`libos/test/regression/rename_unlink_fchown.c` line 107 at r3](https://reviewable.io/reviews/gramineproject/gramine/1875#-NzpJ61z-8_vijtl1XQb:-Nzt1smnDDWUWCuzSv3V:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/042a1401daac3cfe07ab291f730308d43ff8ceba/libos/test/regression/rename_unlink_fchown.c#L107)):*

ditto
</blockquote></details>

Done.
___
*[`libos/test/regression/rename_unlink_fchown.c` line 129 at r3](https://reviewable.io/reviews/gramineproject/gramine/1875#-NzpJW4e72RKT9UnLy2u:-Nzt2PUX-h7F4hQrQpBZ:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/042a1401daac3cfe07ab291f730308d43ff8ceba/libos/test/regression/rename_unlink_fchown.c#L129)):*

Could you verify with `fstat` that the perms were actually changed?
</blockquote></details>

Done.

mkow (changes requested):
> Reviewed 1 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1875)* status: 6 of 7 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`libos/test/regression/rename_unlink.c` line 203 at r3](https://reviewable.io/reviews/gramineproject/gramine/1875#-Nzn5C1d3tFDn73lZHS2:-NzuGbblErNwH_4hyw8-:b-29jgpt) ([raw file](https://github.com/gramineproject/gramine/blob/042a1401daac3cfe07ab291f730308d43ff8ceba/libos/test/regression/rename_unlink.c#L203)):*

Done.
</blockquote></details>

I mean, please assert here - with the assert being placed 170 lines above from here we won't know why we have this weird limitation after merging this PR.
Here we have the exact line which requires that.
___
*[`libos/test/regression/rename_unlink_fchown.c` line 89 at r3](https://reviewable.io/reviews/gramineproject/gramine/1875#-NzpIwmc9ze3ncFCA6wY:-NzuH2D6AulWVh7dQzmD:b69cp55) ([raw file](https://github.com/gramineproject/gramine/blob/042a1401daac3cfe07ab291f730308d43ff8ceba/libos/test/regression/rename_unlink_fchown.c#L89)):*

Done.
</blockquote></details>

Hmm, sorry for another round, but this is extremely unreadable now... Can we switch to octal notation?
I generally agree with https://lwn.net/Articles/696227/ and our styleguide on this, but we don't have `PERM_*` here, so I'd go with octal.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1875)* status: 5 of 7 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`libos/test/regression/rename_unlink.c` line 203 at r3](https://reviewable.io/reviews/gramineproject/gramine/1875#-Nzn5C1d3tFDn73lZHS2:-NzyQ3QeCw--9Y4QAWw1:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/042a1401daac3cfe07ab291f730308d43ff8ceba/libos/test/regression/rename_unlink.c#L203)):*

I mean, please assert here - with the assert being placed 170 lines above from here we won't know why we have this weird limitation after merging this PR.
Here we have the exact line which requires that.
</blockquote></details>

Done.
___
*[`libos/test/regression/rename_unlink_fchown.c` line 89 at r3](https://reviewable.io/reviews/gramineproject/gramine/1875#-NzpIwmc9ze3ncFCA6wY:-NzyRqWL1RIbBpxiVSRW:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/042a1401daac3cfe07ab291f730308d43ff8ceba/libos/test/regression/rename_unlink_fchown.c#L89)):*

Hmm, sorry for another round, but this is extremely unreadable now... Can we switch to octal notation?
I generally agree with https://lwn.net/Articles/696227/ and our styleguide on this, but we don't have `PERM_*` here, so I'd go with octal.
</blockquote></details>

Done.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1875)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`libos/test/regression/rename_unlink_fchown.c` line 89 at r3](https://reviewable.io/reviews/gramineproject/gramine/1875#-NzpIwmc9ze3ncFCA6wY:-Nzzbb-Z86A6g6jPS-GB:bshz8zh) ([raw file](https://github.com/gramineproject/gramine/blob/042a1401daac3cfe07ab291f730308d43ff8ceba/libos/test/regression/rename_unlink_fchown.c#L89)):*

Done.
</blockquote></details>

Why the casts though? Were they needed?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1875)* status: 6 of 7 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`libos/test/regression/rename_unlink_fchown.c` line 89 at r3](https://reviewable.io/reviews/gramineproject/gramine/1875#-NzpIwmc9ze3ncFCA6wY:-O-0-hGN8kNlg4jb1CVM:bd1dqcj) ([raw file](https://github.com/gramineproject/gramine/blob/042a1401daac3cfe07ab291f730308d43ff8ceba/libos/test/regression/rename_unlink_fchown.c#L89)):*

Why the casts though? Were they needed?
</blockquote></details>

Done. Removed them. I thought it was cleaner, but can remove them as well, sure.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1875)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

___
*[`libos/test/regression/rename_unlink_fchown.c` line 89 at r3](https://reviewable.io/reviews/gramineproject/gramine/1875#-NzpIwmc9ze3ncFCA6wY:-O-0kkHNAg0_EfdKA1-E:btwgz70) ([raw file](https://github.com/gramineproject/gramine/blob/042a1401daac3cfe07ab291f730308d43ff8ceba/libos/test/regression/rename_unlink_fchown.c#L89)):*

Done. Removed them. I thought it was cleaner, but can remove them as well, sure.
</blockquote></details>

I think this is well known and idiomatic "Linux C", no need for the loud casts IMO.

mkow (approved):
> Reviewed 4 of 4 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1875)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

kailun-qin (changes requested):
> Reviewed 1 of 2 files at r1, 1 of 2 files at r5, 1 of 1 files at r6, 4 of 4 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1875)* status: all files reviewed, 3 unresolved discussions (waiting on @dimakuv)

___
*[`libos/test/regression/rename_unlink.c` line 186 at r7](https://reviewable.io/reviews/gramineproject/gramine/1875#-O-3pT-IDeVckjeqy5ES:-O-3pT-IDeVckjeqy5ET:b-t2yq0j) ([raw file](https://github.com/gramineproject/gramine/blob/4045c0f983ee376f2883fbebe3e8a616fe7362f2/libos/test/regression/rename_unlink.c#L186)):*
> ```C
>     int fd = create_file(path1, message1, message1_len);
>     if (fd < 0)
>         err(1, "create %s", path1);
> ```

it's handled in `create_file()`, do we still need this?



_Code quote:_
```C
    if (fd < 0)
        err(1, "create %s", path1);
```
___
*[`libos/test/regression/rename_unlink_fchown.c` line 11 at r7](https://reviewable.io/reviews/gramineproject/gramine/1875#-O-3qk6YDr0A6RKlZ1HA:-O-3qk6YDr0A6RKlZ1HB:bvpm2oc) ([raw file](https://github.com/gramineproject/gramine/blob/4045c0f983ee376f2883fbebe3e8a616fe7362f2/libos/test/regression/rename_unlink_fchown.c#L11)):*
> ```C
>  * these tests require a root user to perform fchown with arbitrary user/group.
>  */
> ```

As this is a newly-added test file, should we follow our convention (e.g., using `CHECK()` etc)?
___
*[`libos/test/regression/rename_unlink_fchown.c` line 77 at r7](https://reviewable.io/reviews/gramineproject/gramine/1875#-O-3o2vODLBANii0px0E:-O-3o2vODLBANii0px0F:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/4045c0f983ee376f2883fbebe3e8a616fe7362f2/libos/test/regression/rename_unlink_fchown.c#L77)):*
> ```C
>     int fd = create_file(path1, message, message_len);
>     if (fd < 0)
>         err(1, "create %s", path1);
> ```

ditto



_Code quote:_
```C
    if (fd < 0)
        err(1, "create %s", path1);
```
___
*[`libos/test/regression/rename_unlink_fchown.c` line 90 at r7](https://reviewable.io/reviews/gramineproject/gramine/1875#-O-3oniM97htU6i3Uj3T:-O-3oniM97htU6i3Uj3U:b1vzksl) ([raw file](https://github.com/gramineproject/gramine/blob/4045c0f983ee376f2883fbebe3e8a616fe7362f2/libos/test/regression/rename_unlink_fchown.c#L90)):*
> ```C
>         err(1, "wrong ownership of file %s", path1);
>     if ((st.st_mode & 0777) != 0660)
>         err(1, "wrong permissions of file %s", path1);
> ```

Is it better that we refactor this into a `check_ownership_and_permissions()` helper?



_Code quote:_
```C
    if (stat(path1, &st) != 0)
        err(1, "Failed to stat file %s", path1);
    if (st.st_uid != 123 || st.st_gid != 123)
        err(1, "wrong ownership of file %s", path1);
    if ((st.st_mode & 0777) != 0660)
        err(1, "wrong permissions of file %s", path1);
```
___
*[`libos/test/regression/rename_unlink_fchown.c` line 108 at r7](https://reviewable.io/reviews/gramineproject/gramine/1875#-O-3pVdZ3p_HNkLyqvEE:-O-3pVdZ3p_HNkLyqvEF:b-7ropln) ([raw file](https://github.com/gramineproject/gramine/blob/4045c0f983ee376f2883fbebe3e8a616fe7362f2/libos/test/regression/rename_unlink_fchown.c#L108)):*
> ```C
>         err(1, "wrong ownership of (renamed) file %s", path2);
>     if ((st.st_mode & 0777) != 0666)
>         err(1, "wrong permissions of (renamed) file %s", path2);
> ```

ditto



_Code quote:_
```C
    if (stat(path2, &st) != 0)
        err(1, "Failed to stat (renamed) file %s", path2);
    if (st.st_uid != 321 || st.st_gid != 321)
        err(1, "wrong ownership of (renamed) file %s", path2);
    if ((st.st_mode & 0777) != 0666)
        err(1, "wrong permissions of (renamed) file %s", path2);
```
___
*[`libos/test/regression/rename_unlink_fchown.c` line 152 at r7](https://reviewable.io/reviews/gramineproject/gramine/1875#-O-3qT-a8HKO3Wq9h3AO:-O-3qT-a8HKO3Wq9h3AP:bfmikry) ([raw file](https://github.com/gramineproject/gramine/blob/4045c0f983ee376f2883fbebe3e8a616fe7362f2/libos/test/regression/rename_unlink_fchown.c#L152)):*
> ```C
>     test_rename_fchown_fchmod(path1, path2);
>     test_unlink_fchown(path1);
>     printf("TEST OK\n");
> ```

-> puts



_Code quote:_
```C
printf
```

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1875)* status: 5 of 7 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @kailun-qin and @mkow)

___
*[`libos/test/regression/rename_unlink.c` line 186 at r7](https://reviewable.io/reviews/gramineproject/gramine/1875#-O-3pT-IDeVckjeqy5ES:-O-4tpKt9avsPY97XvPY:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/4045c0f983ee376f2883fbebe3e8a616fe7362f2/libos/test/regression/rename_unlink.c#L186)):*

it's handled in `create_file()`, do we still need this?
</blockquote></details>

Done.
___
*[`libos/test/regression/rename_unlink_fchown.c` line 11 at r7](https://reviewable.io/reviews/gramineproject/gramine/1875#-O-3qk6YDr0A6RKlZ1HA:-O-4ww9K10R1eLPevZbc:b-91d599) ([raw file](https://github.com/gramineproject/gramine/blob/4045c0f983ee376f2883fbebe3e8a616fe7362f2/libos/test/regression/rename_unlink_fchown.c#L11)):*

As this is a newly-added test file, should we follow our convention (e.g., using `CHECK()` etc)?
</blockquote></details>

Done

Note that I added `CHECK()` only in those places where we don't need an explicit clear error message. In other places, I kept the `err()` way, to have a better message.
___
*[`libos/test/regression/rename_unlink_fchown.c` line 77 at r7](https://reviewable.io/reviews/gramineproject/gramine/1875#-O-3o2vODLBANii0px0E:-O-4tzjR4OHhV-zKWIhv:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/4045c0f983ee376f2883fbebe3e8a616fe7362f2/libos/test/regression/rename_unlink_fchown.c#L77)):*

ditto
</blockquote></details>

Done.
___
*[`libos/test/regression/rename_unlink_fchown.c` line 90 at r7](https://reviewable.io/reviews/gramineproject/gramine/1875#-O-3oniM97htU6i3Uj3T:-O-4x4Sc4taToCMvOiT6:b2rc9ym) ([raw file](https://github.com/gramineproject/gramine/blob/4045c0f983ee376f2883fbebe3e8a616fe7362f2/libos/test/regression/rename_unlink_fchown.c#L90)):*

Is it better that we refactor this into a `check_ownership_and_permissions()` helper?
</blockquote></details>

Done. I didn't add `stat()` in there, since we can have `stat` or `fstat` syscalls, so it would be cumbersome to choose which calls needs to be done.
___
*[`libos/test/regression/rename_unlink_fchown.c` line 108 at r7](https://reviewable.io/reviews/gramineproject/gramine/1875#-O-3pVdZ3p_HNkLyqvEE:-O-4xCAC5h8VhevA8M7R:bb6un1j) ([raw file](https://github.com/gramineproject/gramine/blob/4045c0f983ee376f2883fbebe3e8a616fe7362f2/libos/test/regression/rename_unlink_fchown.c#L108)):*

ditto
</blockquote></details>

Done
___
*[`libos/test/regression/rename_unlink_fchown.c` line 152 at r7](https://reviewable.io/reviews/gramineproject/gramine/1875#-O-3qT-a8HKO3Wq9h3AO:-O-4xDIYCTuxoNFSwteN:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/4045c0f983ee376f2883fbebe3e8a616fe7362f2/libos/test/regression/rename_unlink_fchown.c#L152)):*

-> puts
</blockquote></details>

Done.

kailun-qin (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1875)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: Intel)

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1875)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

dimakuv:
Jenkins, retest Jenkins-SGX-22.04 please

dimakuv:
Jenkins, retest Jenkins-SGX-24.04 please

PR #1877, opened 2024-05-08
[PAL/Linux-SGX] Re-init attestation key after AESM restarted

dimakuv:
## Description of the changes 

Gramine uses the AESM daemon to request SGX Quotes from the Quoting Enclave. AESM runs in the background and may be restarted. For some reason, after a restart AESM "forgets" the platform's attestation key and requires an additional INIT_QUOTE_REQUEST before it is able to generate SGX Quotes again. This behavior is documented in Intel SGX PSW/DCAP software, but Gramine didn't implement this and thus failed with "AESM service returned error 42".

This PR complies with the SGX PSW/DCAP documentation and adds the send of INIT_QUOTE_REQUEST + retry of SGX quote retrieval.

For documentation, see e.g. this comment: https://github.com/intel/linux-sgx/blob/80a6625c497056c43e78d993e414ca99a9efed5c/common/inc/sgx_uae_quote_ex.h#L139-L140

## How to test this PR? 

This was detected by a customer.

A simple reproducer is like this:
```diff
diff --git a/CI-Examples/python/scripts/sgx-quote.py b/CI-Examples/python/scripts/sgx-quote.py
index 8a79b4f1..3c979553 100644
--- a/CI-Examples/python/scripts/sgx-quote.py
+++ b/CI-Examples/python/scripts/sgx-quote.py
@@ -5,27 +5,32 @@

 import os
 import sys
+import time

-if not os.path.exists("/dev/attestation/quote"):
-    print("Cannot find `/dev/attestation/quote`; "
-          "are you running under SGX, with remote attestation enabled?")
-    sys.exit(1)
-
-with open('/dev/attestation/attestation_type') as f:
-    print(f"Detected attestation type: {f.read()}")
-
-with open("/dev/attestation/user_report_data", "wb") as f:
-    f.write(b'\0'*64)
-
-with open("/dev/attestation/quote", "rb") as f:
-    quote = f.read()
-
-print(f"Extracted SGX quote with size = {len(quote)} and the following fields:")
-print(f"  ATTRIBUTES.FLAGS: {quote[96:104].hex()}  [ Debug bit: {quote[96] & 2 > 0} ]")
-print(f"  ATTRIBUTES.XFRM:  {quote[104:112].hex()}")
-print(f"  MRENCLAVE:        {quote[112:144].hex()}")
-print(f"  MRSIGNER:         {quote[176:208].hex()}")
-print(f"  ISVPRODID:        {quote[304:306].hex()}")
-print(f"  ISVSVN:           {quote[306:308].hex()}")
-print(f"  REPORTDATA:       {quote[368:400].hex()}")
-print(f"                    {quote[400:432].hex()}")
+while True:
+    if not os.path.exists("/dev/attestation/quote"):
+        print("Cannot find `/dev/attestation/quote`; "
+                "are you running under SGX, with remote attestation enabled?")
+        sys.exit(1)
+
+    with open('/dev/attestation/attestation_type') as f:
+        print(f"Detected attestation type: {f.read()}")
+
+    with open("/dev/attestation/user_report_data", "wb") as f:
+        f.write(b'\0'*64)
+
+    with open("/dev/attestation/quote", "rb") as f:
+        quote = f.read()
+
+    print(f"Extracted SGX quote with size = {len(quote)} and the following fields:")
+    print(f"  ATTRIBUTES.FLAGS: {quote[96:104].hex()}  [ Debug bit: {quote[96] & 2 > 0} ]")
+    print(f"  ATTRIBUTES.XFRM:  {quote[104:112].hex()}")
+    print(f"  MRENCLAVE:        {quote[112:144].hex()}")
+    print(f"  MRSIGNER:         {quote[176:208].hex()}")
+    print(f"  ISVPRODID:        {quote[304:306].hex()}")
+    print(f"  ISVSVN:           {quote[306:308].hex()}")
+    print(f"  REPORTDATA:       {quote[368:400].hex()}")
+    print(f"                    {quote[400:432].hex()}")
+
+    sys.stdout.flush()
+    time.sleep(5)
```

Apply the above patch, run it in one terminal, and in another terminal do `sudo systemctl restart aesmd`. Without this PR, Gramine will fail. With this PR, Gramine will continue execution of the endless loop.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1877)

### DIFF
--- a/pal/src/host/linux-sgx/enclave_ocalls.c
+++ b/pal/src/host/linux-sgx/enclave_ocalls.c
@@ -2107,7 +2107,8 @@ int ocall_ioctl(int fd, unsigned int cmd, unsigned long arg) {
 }
 
 int ocall_get_quote(const sgx_spid_t* spid, bool linkable, const sgx_report_t* report,
-                    const sgx_quote_nonce_t* nonce, char** quote, size_t* quote_len) {
+                    const sgx_quote_nonce_t* nonce, char** quote, size_t* quote_len,
+                    sgx_target_info_t* qe_targetinfo, bool* qe_targetinfo_set) {
     int retval;
     struct ocall_get_quote* ocall_quote_args;
     char* buf = NULL;
@@ -2127,6 +2128,10 @@ int ocall_get_quote(const sgx_spid_t* spid, bool linkable, const sgx_report_t* r
         COPY_VALUE_TO_UNTRUSTED(&ocall_quote_args->is_epid, false);
         memset(&ocall_quote_args->spid, 0, sizeof(ocall_quote_args->spid)); /* for sanity */
     }
+
+    /* for sanity */
+    COPY_VALUE_TO_UNTRUSTED(&ocall_quote_args->qe_targetinfo_set, false);
+    memset(&ocall_quote_args->qe_targetinfo, 0, sizeof(ocall_quote_args->qe_targetinfo));
 
     memcpy(&ocall_quote_args->report, report, sizeof(*report));
     memcpy(&ocall_quote_args->nonce, nonce, sizeof(*nonce));
@@ -2177,6 +2182,12 @@ int ocall_get_quote(const sgx_spid_t* spid, bool linkable, const sgx_report_t* r
 
             *quote     = buf;
             *quote_len = len;
+
+            *qe_targetinfo_set = false;
+            if (quote_copied.qe_targetinfo_set) {
+                *qe_targetinfo_set = true;
+                memcpy(qe_targetinfo, &quote_copied.qe_targetinfo, sizeof(*qe_targetinfo));
+            }
         }
     }
 
--- a/pal/src/host/linux-sgx/enclave_ocalls.h
+++ b/pal/src/host/linux-sgx/enclave_ocalls.h
@@ -112,14 +112,18 @@ int ocall_ioctl(int fd, unsigned int cmd, unsigned long arg);
 /*!
  * \brief Execute untrusted code in PAL to obtain a quote from the Quoting Enclave.
  *
- * \param      spid       Software provider ID (SPID); if NULL then DCAP/ECDSA is used.
- * \param      linkable   Quote type (linkable vs unlinkable); ignored if DCAP/ECDSA is used.
- * \param      report     Enclave report to be sent to the Quoting Enclave.
- * \param      nonce      16B nonce to be included in the quote for freshness; ignored if
- *                        DCAP/ECDSA is used.
- * \param[out] quote      Quote returned by the Quoting Enclave (allocated via malloc() in this
- *                        function; the caller gets the ownership of the quote).
- * \param[out] quote_len  Length of the quote returned by the Quoting Enclave.
+ * \param      spid               Software provider ID (SPID); if NULL then DCAP is used.
+ * \param      linkable           Quote type (linkable vs unlinkable); ignored if DCAP is used.
+ * \param      report             Enclave report to be sent to the Quoting Enclave.
+ * \param      nonce              16B nonce to be included in the quote for freshness; ignored if
+ *                                DCAP is used.
+ * \param[out] quote              Quote returned by the Quoting Enclave (allocated via mmap() in
+ *                                this function; the caller gets the ownership of the quote).
+ * \param[out] quote_len          Length of the quote returned by the Quoting Enclave.
+ * \param[out] qe_targetinfo      Retrieved Quoting Enclave's target info; buffer must be submitted
+ *                                by the caller.
+ * \param[out] qe_targetinfo_set  Whether Quoting Enclave's target info (above) was set; if true the
+ *                                caller must update the corresponding qe_targetinfo global var.
  *
  * \returns 0 on success, negative Linux error code otherwise.
  *
@@ -127,7 +131,8 @@ int ocall_ioctl(int fd, unsigned int cmd, unsigned long arg);
  * returned quote corresponds to this enclave or whether its contents make sense).
  */
 int ocall_get_quote(const sgx_spid_t* spid, bool linkable, const sgx_report_t* report,
-                    const sgx_quote_nonce_t* nonce, char** quote, size_t* quote_len);
+                    const sgx_quote_nonce_t* nonce, char** quote, size_t* quote_len,
+                    sgx_target_info_t* qe_targetinfo, bool* qe_targetinfo_set);
 
 int ocall_edmm_restrict_pages_perm(uint64_t addr, size_t count, uint64_t prot);
 int ocall_edmm_modify_pages_type(uint64_t addr, size_t count, uint64_t type);
--- a/pal/src/host/linux-sgx/enclave_platform.c
+++ b/pal/src/host/linux-sgx/enclave_platform.c
@@ -4,6 +4,8 @@
 #include "pal_linux.h"
 #include "pal_linux_error.h"
 
+/* this function assumes proper synchronization by callers, such that
+ * g_pal_linuxsgx_state.qe_targetinfo is not racey */
 int sgx_get_quote(const sgx_spid_t* spid, const sgx_quote_nonce_t* nonce,
                   const sgx_report_data_t* report_data, bool linkable, char** quote,
                   size_t* quote_len) {
@@ -18,10 +20,18 @@ int sgx_get_quote(const sgx_spid_t* spid, const sgx_quote_nonce_t* nonce,
         return -PAL_ERROR_DENIED;
     }
 
-    ret = ocall_get_quote(spid, linkable, &report, nonce, quote, quote_len);
+    sgx_target_info_t qe_targetinfo;
+    bool qe_targetinfo_set = false;
+
+    ret = ocall_get_quote(spid, linkable, &report, nonce, quote, quote_len, &qe_targetinfo,
+                          &qe_targetinfo_set);
     if (ret < 0) {
         log_error("Failed to get quote");
         return unix_to_pal_error(ret);
     }
+
+    if (qe_targetinfo_set)
+        memcpy(&g_pal_linuxsgx_state.qe_targetinfo, &qe_targetinfo, sizeof(qe_targetinfo));
+
     return 0;
 }
--- a/pal/src/host/linux-sgx/host_ocalls.c
+++ b/pal/src/host/linux-sgx/host_ocalls.c
@@ -716,10 +716,25 @@ static long sgx_ocall_ioctl(void* args) {
 
 static long sgx_ocall_get_quote(void* args) {
     struct ocall_get_quote* ocall_quote_args = args;
-    return retrieve_quote(ocall_quote_args->is_epid ? &ocall_quote_args->spid : NULL,
-                          ocall_quote_args->linkable, &ocall_quote_args->report,
-                          &ocall_quote_args->nonce, &ocall_quote_args->quote,
-                          &ocall_quote_args->quote_len);
+
+    int ret;
+    ret = retrieve_quote(ocall_quote_args->is_epid ? &ocall_quote_args->spid : NULL,
+                         ocall_quote_args->linkable, &ocall_quote_args->report,
+                         &ocall_quote_args->nonce, &ocall_quote_args->quote,
+                         &ocall_quote_args->quote_len);
+    if (ret == -EAGAIN) {
+        /* special case, perform init_quoting_enclave_targetinfo() and retry */
+        sgx_target_info_t dummy_qe_targetinfo = {0};
+        ret = init_quoting_enclave_targetinfo(ocall_quote_args->is_epid, &dummy_qe_targetinfo);
+        if (ret < 0)
+            return ret;
+
+        ret = retrieve_quote(ocall_quote_args->is_epid ? &ocall_quote_args->spid : NULL,
+                             ocall_quote_args->linkable, &ocall_quote_args->report,
+                             &ocall_quote_args->nonce, &ocall_quote_args->quote,
+                             &ocall_quote_args->quote_len);
+    }
+    return ret;
 }
 
 static long sgx_ocall_edmm_modify_pages_type(void* _args) {
--- a/pal/src/host/linux-sgx/host_ocalls.c
+++ b/pal/src/host/linux-sgx/host_ocalls.c
@@ -715,8 +715,10 @@ static long sgx_ocall_ioctl(void* args) {
 }
 
 static long sgx_ocall_get_quote(void* args) {
-    int ret;
     struct ocall_get_quote* ocall_quote_args = args;
+
+    int ret;
+    ocall_quote_args->qe_targetinfo_set = false;
 
     int retries = 0;
     while (retries < 5) {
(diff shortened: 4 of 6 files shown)


mkow (changes requested):
btw. I would be a bit easier to test if you just pasted the new version of `sgx-quote.py` as is, instead of the diff - it replaces the whole script anyways ;)

> Reviewed 1 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1877#-:-NxQ0_CL-yApQn1TZeal:b-gb34gf)* status: 1 of 2 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1877#-NxQ3xHZEWUJ8tWKrbGO:-NxQ3xHZEWUJ8tWKrbGP:b-36zci) (no related file):*
> Apply the above patch, run it in one terminal, and in another terminal do `sudo systemctl restart aesmd`. Without this PR, Gramine will fail.

It doesn't fail for me on a client CPU with EPID attestation. What's the exact setup to reproduce this?

___
___
*[`pal/src/host/linux-sgx/host_ocalls.c` line 725 at r1](https://reviewable.io/reviews/gramineproject/gramine/1877#-NxQ4fIv-zGQTrDMm5Tl:-NxQ4fIv-zGQTrDMm5Tm:b-6kdd9b) ([raw file](https://github.com/gramineproject/gramine/blob/e87639d1149ece1e0a6d35c4852c264453f13432/pal/src/host/linux-sgx/host_ocalls.c#L725)):*
> ```C
>                          &ocall_quote_args->nonce, &ocall_quote_args->quote,
>                          &ocall_quote_args->quote_len);
>     if (ret == -EAGAIN) {
> ```

I'd prefer a do-while instead of this copy-paste. Also, I think it should actually be a loop, not just a single retry (in case we race with another restart).
___
*[`pal/src/host/linux-sgx/host_platform.c` line 22 at r1](https://reviewable.io/reviews/gramineproject/gramine/1877#-NxQ5BzU80dCpP3aAxzt:-NxQ5BzV3-6PMY7MrS1k:bb52gai) ([raw file](https://github.com/gramineproject/gramine/blob/e87639d1149ece1e0a6d35c4852c264453f13432/pal/src/host/linux-sgx/host_platform.c#L22)):*
> ```C
> /* retrieving the SGX quote may return error 42, which means that the attestation key is not
>  * available and AESM service must re-generate the key; when Gramine sees such error, it performs a
>  * dummy INIT_QUOTE_REQUEST and then re-tries retrieving the SGX quote */
> ```

I'd say that this explanation belongs to the function retrying the get-quote request, not to the constant.

kailun-qin (changes requested):
> Reviewed 1 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1877)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1877#-NxQ3xHZEWUJ8tWKrbGO:-NxQR_q_7LtJHKrIsvEF:b-yneeyj) (no related file):*

> Apply the above patch, run it in one terminal, and in another terminal do `sudo systemctl restart aesmd`. Without this PR, Gramine will fail.

It doesn't fail for me on a client CPU with EPID attestation. What's the exact setup to reproduce this?
</blockquote></details>

Yes, the issue was originally reported on a DCAP setup.

I think this specific error (`AESM_ATT_KEY_NOT_INITIALIZED`) and the issue should only be seen on DCAP where an attestation key re-generation in QE is involved. While w/ EPID, the attestation key is actually provisioned to and sealed by QVE and later unsealed by QE (during init and get quote) where such keygen process is irrelevant.

___
___
*[`pal/src/host/linux-sgx/host_ocalls.c` line 725 at r1](https://reviewable.io/reviews/gramineproject/gramine/1877#-NxQ4fIv-zGQTrDMm5Tl:-NxQJNHW8tqgoY-apAII:bovb0t2) ([raw file](https://github.com/gramineproject/gramine/blob/e87639d1149ece1e0a6d35c4852c264453f13432/pal/src/host/linux-sgx/host_ocalls.c#L725)):*

I'd prefer a do-while instead of this copy-paste. Also, I think it should actually be a loop, not just a single retry (in case we race with another restart).
</blockquote></details>

+1, I think better with a retry limit
___
*[`pal/src/host/linux-sgx/host_platform.c` line 305 at r1](https://reviewable.io/reviews/gramineproject/gramine/1877#-NxQLjGDCv8tMDUuVoSe:-NxQLjGE2Dw1XSkHe-hg:b-1njmlz) ([raw file](https://github.com/gramineproject/gramine/blob/e87639d1149ece1e0a6d35c4852c264453f13432/pal/src/host/linux-sgx/host_platform.c#L305)):*
> ```C
> 
>         Response__GetQuoteResponse* r = res->getquoteres;
>         if (r->errorcode == AESM_ATT_KEY_NOT_INITIALIZED) {
> ```

I think the PR does not apply to EPID, pls see the discussions above.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1877)* status: 0 of 2 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @kailun-qin, and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1877#-NxQ3xHZEWUJ8tWKrbGO:-NxWKWNu-5YiOMyftmut:bxmxo7l) (no related file):*

Yes, the issue was originally reported on a DCAP setup.

I think this specific error (`AESM_ATT_KEY_NOT_INITIALIZED`) and the issue should only be seen on DCAP where an attestation key re-generation in QE is involved. While w/ EPID, the attestation key is actually provisioned to and sealed by QVE and later unsealed by QE (during init and get quote) where such keygen process is irrelevant.
</blockquote></details>

Yes, this was reported with DCAP, and I also reproduced on a DCAP machine. I didn't have an EPID machine around, so I just assumed it has the same issue.

I reverted the changes to the EPID paths now. Thanks Kailun for explanations.

___
___
*[`-- commits` line 10 at r1](https://reviewable.io/reviews/gramineproject/gramine/1877#-NxWKjDC3bnqXTJQ5hj5):*
Add somewhere here that this doesn't concern EPID attestation. Only DCAP.
___
*[`pal/src/host/linux-sgx/host_ocalls.c` line 725 at r1](https://reviewable.io/reviews/gramineproject/gramine/1877#-NxQ4fIv-zGQTrDMm5Tl:-NxWMnsY7Ve6pNZEhOKL:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/e87639d1149ece1e0a6d35c4852c264453f13432/pal/src/host/linux-sgx/host_ocalls.c#L725)):*

+1, I think better with a retry limit
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/host_platform.c` line 22 at r1](https://reviewable.io/reviews/gramineproject/gramine/1877#-NxQ5BzU80dCpP3aAxzt:-NxWMpdE5hlDsZN0UG4r:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/e87639d1149ece1e0a6d35c4852c264453f13432/pal/src/host/linux-sgx/host_platform.c#L22)):*

I'd say that this explanation belongs to the function retrying the get-quote request, not to the constant.
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/host_platform.c` line 305 at r1](https://reviewable.io/reviews/gramineproject/gramine/1877#-NxQLjGDCv8tMDUuVoSe:-NxWKxjS3EplECYCiPf0:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/e87639d1149ece1e0a6d35c4852c264453f13432/pal/src/host/linux-sgx/host_platform.c#L305)):*

I think the PR does not apply to EPID, pls see the discussions above.
</blockquote></details>

Done.

llly (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1877)* status: 0 of 2 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @kailun-qin, and @mkow)

___
*[`pal/src/host/linux-sgx/host_ocalls.c` line 736 at r2](https://reviewable.io/reviews/gramineproject/gramine/1877#-NxWUhZY3WzLMnUXq_uY:-NxWUhZY3WzLMnUXq_uZ:bnx40wq) ([raw file](https://github.com/gramineproject/gramine/blob/f563707af2d9169599bc64978311e048fe598d70/pal/src/host/linux-sgx/host_ocalls.c#L736)):*
> ```C
>          * a dummy INIT_QUOTE_REQUEST and then re-try retrieving the SGX quote.
>          */
>         sgx_target_info_t dummy_qe_targetinfo = {0};
> ```

New `qe_targetinfo` should not by dummy.
QE can be upgraded after aesm_service restart.
Must store it to`g_pal_linuxsgx_state.qe_targetinfo`.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1877)* status: 0 of 6 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin and @mkow)

___
*[`pal/src/host/linux-sgx/host_ocalls.c` line 736 at r2](https://reviewable.io/reviews/gramineproject/gramine/1877#-NxWUhZY3WzLMnUXq_uY:-NxWl2yL1GRhhOsrxKD1:b3wz92j) ([raw file](https://github.com/gramineproject/gramine/blob/f563707af2d9169599bc64978311e048fe598d70/pal/src/host/linux-sgx/host_ocalls.c#L736)):*

New `qe_targetinfo` should not by dummy.
QE can be upgraded after aesm_service restart.
Must store it to`g_pal_linuxsgx_state.qe_targetinfo`.
</blockquote></details>

I tried to implement this in r3 of this PR, but I now realize that if the flow is more complicated and my current PR still fails if QE Target Info was modified:
1. SGX Report must be generated inside the SGX enclave from the QE Target Info (in trusted PAL).
2. This SGX Report is put inside the GET_QUOTE request to AESM (in untrusted PAL).
3. AESM calls into Quoting Enclave (QE).
4. QE checks that QE Target Info inside the GET_QUOTE's SGX Report corresponds to its own Target Info.
5. Only if Target Infos match, then QE generates the SGX Quote.
6. AESM returns the SGX Quote to Gramine's untrusted PAL, or (if AESM restarted) returns an error 42.
7. If Gramine's untrusted PAL gets error 42, Gramine must get the new QE Target Info (in untrusted PAL, by talking to AESM), then jump into the SGX enclave (in trusted PAL) and restart the whole sequence from step 1.

The problem with my current PR is that we retry the loop only in untrusted PAL (basically in step 2). In fact, we must wrap the whole sequence of steps 1-7 into a while loop.

So my "hack" of `ocall_get_quote()` returning a new QE Target Info happens too late -- we may get into a situation when AESM updated QE Target Info, so it won't generate a new SGX Quote but return some error "hey, Gramine, you sent me an old QE Target Info, I don't trust you".

@mkow @kailun-qin @llly My current idea is to introduce a new OCALL like `ocall_update_qe_targetinfo()` which will be called if `ocall_get_quote()` returns -EAGAIN (corresponds to AESM error 42). Does this sounds reasonable to you?

llly (commented):
> Reviewed 3 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1877)* status: 3 of 6 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @kailun-qin, and @mkow)

___
*[`pal/src/host/linux-sgx/enclave_platform.c` line 14 at r3](https://reviewable.io/reviews/gramineproject/gramine/1877#-Nx_QQfi6MH1NJ6-pyK-:-Nx_QQfi6MH1NJ6-pyK0:bb703n6) ([raw file](https://github.com/gramineproject/gramine/blob/87ea2e27dcf73121114379e8798ada16fab09fb6/pal/src/host/linux-sgx/enclave_platform.c#L14)):*
> ```C
>     /* must align all arguments to sgx_report() so that EREPORT doesn't complain */
>     __sgx_mem_aligned sgx_report_t report;
>     __sgx_mem_aligned sgx_target_info_t targetinfo = g_pal_linuxsgx_state.qe_targetinfo;
> ```

Mark
___
*[`pal/src/host/linux-sgx/host_ocalls.c` line 736 at r2](https://reviewable.io/reviews/gramineproject/gramine/1877#-NxWUhZY3WzLMnUXq_uY:-Nx_QrTD8iHdIvXSuY9e:b-52o6ju) ([raw file](https://github.com/gramineproject/gramine/blob/f563707af2d9169599bc64978311e048fe598d70/pal/src/host/linux-sgx/host_ocalls.c#L736)):*

I tried to implement this in r3 of this PR, but I now realize that if the flow is more complicated and my current PR still fails if QE Target Info was modified:
1. SGX Report must be generated inside the SGX enclave from the QE Target Info (in trusted PAL).
2. This SGX Report is put inside the GET_QUOTE request to AESM (in untrusted PAL).
3. AESM calls into Quoting Enclave (QE).
4. QE checks that QE Target Info inside the GET_QUOTE's SGX Report corresponds to its own Target Info.
5. Only if Target Infos match, then QE generates the SGX Quote.
6. AESM returns the SGX Quote to Gramine's untrusted PAL, or (if AESM restarted) returns an error 42.
7. If Gramine's untrusted PAL gets error 42, Gramine must get the new QE Target Info (in untrusted PAL, by talking to AESM), then jump into the SGX enclave (in trusted PAL) and restart the whole sequence from step 1.

The problem with my current PR is that we retry the loop only in untrusted PAL (basically in step 2). In fact, we must wrap the whole sequence of steps 1-7 into a while loop.

So my "hack" of `ocall_get_quote()` returning a new QE Target Info happens too late -- we may get into a situation when AESM updated QE Target Info, so it won't generate a new SGX Quote but return some error "hey, Gramine, you sent me an old QE Target Info, I don't trust you".

@mkow @kailun-qin @llly My current idea is to introduce a new OCALL like `ocall_update_qe_targetinfo()` which will be called if `ocall_get_quote()` returns -EAGAIN (corresponds to AESM error 42). Does this sounds reasonable to you?
</blockquote></details>

Reasonable for me.
I add a "Mark" comment in `enclave_platform.c`. It uses `g_pal_linux.sgx_state.qe_targetinfo` to create a report.
I think you can add a loop there that if `ocall_get_quote` returns 42, call`ocall_update_qe_targetinfo()` then go back to create a new report and call `ocall_get_quote` again until success.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1877)* status: 3 of 6 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @kailun-qin and @mkow)

___
*[`-- commits` line 10 at r1](https://reviewable.io/reviews/gramineproject/gramine/1877#-NxWKjDC3bnqXTJQ5hj5):*

Add somewhere here that this doesn't concern EPID attestation. Only DCAP.
</blockquote></details>

Done in #1881 
___
*[`pal/src/host/linux-sgx/host_ocalls.c` line 736 at r2](https://reviewable.io/reviews/gramineproject/gramine/1877#-NxWUhZY3WzLMnUXq_uY:-NxlBcQM4YzRIheBr9AS:bhmdhk1) ([raw file](https://github.com/gramineproject/gramine/blob/f563707af2d9169599bc64978311e048fe598d70/pal/src/host/linux-sgx/host_ocalls.c#L736)):*

Reasonable for me.
I add a "Mark" comment in `enclave_platform.c`. It uses `g_pal_linux.sgx_state.qe_targetinfo` to create a report.
I think you can add a loop there that if `ocall_get_quote` returns 42, call`ocall_update_qe_targetinfo()` then go back to create a new report and call `ocall_get_quote` again until success.
</blockquote></details>

Thank you @llly !

I re-created this patch as #1881 (as the implementation drastically changed). I'm blocking this PR, and it will be closed automatically after #1881 is merged.

PR #1881, opened 2024-05-13
[PAL/Linux-SGX] Re-init attestation key after AESM restarted

dimakuv:
## Description of the changes 

Gramine uses the AESM daemon to request SGX Quotes from the Quoting Enclave. AESM runs in the background and may be restarted. For some reason, after a restart AESM "forgets" the platform's attestation key and requires an additional INIT_QUOTE_REQUEST before it is able to generate SGX Quotes again. Moreover, the Quoting Enclave's Target Info may change at this point, so Gramine must update its cached QE Target Info. This behavior is applicable only to DCAP attestation and is documented in Intel SGX PSW/DCAP software. Gramine didn't implement this behavior and thus failed with "AESM service returned error 42".

This PR complies with the SGX PSW/DCAP documentation and adds the send of INIT_QUOTE_REQUEST + update of QE Target Info + retry of SGX quote retrieval.

This is a re-creation of #1877, with a better design. Closes #1877.

## How to test this PR? 

Change `CI-Examples/python/scripts/sgx-quote.py` to this:
```python
import os
import sys
import time

while True:
    if not os.path.exists("/dev/attestation/quote"):
        print("Cannot find `/dev/attestation/quote`; "
                "are you running under SGX, with remote attestation enabled?")
        sys.exit(1)

    with open('/dev/attestation/attestation_type') as f:
        print(f"Detected attestation type: {f.read()}")

    with open("/dev/attestation/user_report_data", "wb") as f:
        f.write(b'\0'*64)

    with open("/dev/attestation/quote", "rb") as f:
        quote = f.read()

    print(f"Extracted SGX quote with size = {len(quote)} and the following fields:")
    print(f"  ATTRIBUTES.FLAGS: {quote[96:104].hex()}  [ Debug bit: {quote[96] & 2 > 0} ]")
    print(f"  ATTRIBUTES.XFRM:  {quote[104:112].hex()}")
    print(f"  MRENCLAVE:        {quote[112:144].hex()}")
    print(f"  MRSIGNER:         {quote[176:208].hex()}")
    print(f"  ISVPRODID:        {quote[304:306].hex()}")
    print(f"  ISVSVN:           {quote[306:308].hex()}")
    print(f"  REPORTDATA:       {quote[368:400].hex()}")
    print(f"                    {quote[400:432].hex()}")

    sys.stdout.flush()
    time.sleep(5)
```

Apply the above Python script, run it in one terminal, and in another terminal do `sudo systemctl restart aesmd`. Without this PR, Gramine will fail. With this PR, Gramine will continue execution of the endless loop.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1881)

### DIFF
--- a/pal/src/host/linux-sgx/enclave_ocalls.c
+++ b/pal/src/host/linux-sgx/enclave_ocalls.c
@@ -2187,6 +2187,45 @@ out:
     return retval;
 }
 
+int ocall_get_qe_targetinfo(bool is_epid, sgx_target_info_t* qe_targetinfo) {
+    int retval;
+    struct ocall_get_qe_targetinfo* ocall_qe_ti_args;
+
+    void* old_ustack = sgx_prepare_ustack();
+    ocall_qe_ti_args = sgx_alloc_on_ustack_aligned(sizeof(*ocall_qe_ti_args),
+                                                   alignof(*ocall_qe_ti_args));
+    if (!ocall_qe_ti_args) {
+        sgx_reset_ustack(old_ustack);
+        return -EPERM;
+    }
+
+    COPY_VALUE_TO_UNTRUSTED(&ocall_qe_ti_args->is_epid, is_epid);
+    memset(&ocall_qe_ti_args->qe_targetinfo, 0, sizeof(ocall_qe_ti_args->qe_targetinfo));
+
+    do {
+        retval = sgx_exitless_ocall(OCALL_GET_QE_TARGETINFO, ocall_qe_ti_args);
+    } while (retval == -EINTR);
+
+    if (retval < 0 && retval != -EACCES && retval != -EINVAL && retval != -ENOMEM &&
+            retval != -EPERM && retval != -EAGAIN && retval != -ECONNREFUSED) {
+        /* OCALL_GET_QE_TARGETINFO OCALL may return many error codes, but we sanitize all error
+         * codes except the above (most important ones) because PAL/LibOS logic doesn't care about
+         * specific errors */
+        retval = -EPERM;
+    }
+
+    if (!retval) {
+        if (!sgx_copy_to_enclave(qe_targetinfo, sizeof(*qe_targetinfo),
+                                 &ocall_qe_ti_args->qe_targetinfo,
+                                 sizeof(ocall_qe_ti_args->qe_targetinfo))) {
+            retval = -EPERM;
+        }
+    }
+
+    sgx_reset_ustack(old_ustack);
+    return retval;
+}
+
 int ocall_sched_setaffinity(void* tcs, unsigned long* cpu_mask, size_t cpu_mask_len) {
     int retval = 0;
     struct ocall_sched_setaffinity* ocall_sched_args;
--- a/pal/src/host/linux-sgx/enclave_ocalls.h
+++ b/pal/src/host/linux-sgx/enclave_ocalls.h
@@ -129,6 +129,20 @@ int ocall_ioctl(int fd, unsigned int cmd, unsigned long arg);
 int ocall_get_quote(const sgx_spid_t* spid, bool linkable, const sgx_report_t* report,
                     const sgx_quote_nonce_t* nonce, char** quote, size_t* quote_len);
 
+/*!
+ * \brief Execute untrusted code in PAL to obtain Target Info of the Quoting Enclave (QE).
+ *
+ * \param      is_epid            True if EPID is used, false if DCAP/ECDSA is used.
+ * \param[out] qe_targetinfo      Retrieved Quoting Enclave's target info; buffer must be provided
+ *                                by the caller.
+ *
+ * \returns 0 on success, negative Linux error code otherwise.
+ *
+ * The obtained QE Target Info is not validated in any way (i.e., this function does not check
+ * whether Target Info contents make sense).
+ */
+int ocall_get_qe_targetinfo(bool is_epid, sgx_target_info_t* qe_targetinfo);
+
 int ocall_edmm_restrict_pages_perm(uint64_t addr, size_t count, uint64_t prot);
 int ocall_edmm_modify_pages_type(uint64_t addr, size_t count, uint64_t type);
 int ocall_edmm_remove_pages(uint64_t addr, size_t count);
--- a/pal/src/host/linux-sgx/enclave_platform.c
+++ b/pal/src/host/linux-sgx/enclave_platform.c
@@ -1,27 +1,90 @@
 /* SPDX-License-Identifier: LGPL-3.0-or-later */
 /* Copyright (C) 2019, Texas A&M University */
 
+#include "api.h"
+#include "enclave_api.h"
 #include "pal_linux.h"
 #include "pal_linux_error.h"
+#include "sgx_arch.h"
+
+static sgx_target_info_t g_qe_targetinfo; /* received from untrusted host, use carefully */
+static spinlock_t g_qe_targetinfo_lock = INIT_SPINLOCK_UNLOCKED;
+
+int init_qe_targetinfo(void* uptr_qe_targetinfo) {
+    /* no need to acquire g_qe_targetinfo_lock at this point since there is only a single thread */
+    if (!sgx_copy_to_enclave(&g_qe_targetinfo,
+                             sizeof(g_qe_targetinfo),
+                             uptr_qe_targetinfo,
+                             sizeof(sgx_target_info_t))) {
+        return -PAL_ERROR_DENIED;
+    }
+    return 0;
+}
 
 int sgx_get_quote(const sgx_spid_t* spid, const sgx_quote_nonce_t* nonce,
                   const sgx_report_data_t* report_data, bool linkable, char** quote,
                   size_t* quote_len) {
-    /* must align all arguments to sgx_report() so that EREPORT doesn't complain */
-    __sgx_mem_aligned sgx_report_t report;
-    __sgx_mem_aligned sgx_target_info_t targetinfo = g_pal_linuxsgx_state.qe_targetinfo;
-    __sgx_mem_aligned sgx_report_data_t _report_data = *report_data;
+    int ret;
+    int retries = 0;
+    while (retries < 5) {
+        /* must align all arguments to sgx_report() so that EREPORT doesn't complain */
+        __sgx_mem_aligned sgx_target_info_t targetinfo;
+        __sgx_mem_aligned sgx_report_data_t _report_data;
+        __sgx_mem_aligned sgx_report_t report;
 
-    int ret = sgx_report(&targetinfo, &_report_data, &report);
-    if (ret) {
-        log_error("Failed to get enclave report");
-        return -PAL_ERROR_DENIED;
+        bool new_qe_targetinfo = false;
+        if (retries) {
+            /* new attempt, need to update QE target info */
+            ret = ocall_get_qe_targetinfo(/*is_epid=*/!!spid, &targetinfo);
+            if (ret < 0) {
+                log_error("Failed to get QE target info (error code=%d)", ret);
+                return unix_to_pal_error(ret);
+            }
+            new_qe_targetinfo = true;
+        }
+
+        spinlock_lock(&g_qe_targetinfo_lock);
+        if (new_qe_targetinfo) {
+            /* got new QE target info into the local stack var, also update the global var */
+            g_qe_targetinfo = targetinfo;
+        } else {
+            /* use old QE target info stored in global var, copy it into local stack var */
+            targetinfo = g_qe_targetinfo;
+        }
+        spinlock_unlock(&g_qe_targetinfo_lock);
+
+        _report_data = *report_data;
+        ret = sgx_report(&targetinfo, &_report_data, &report);
+        if (ret) {
+            log_error("Failed to get enclave report (error code=%d)", ret);
+            return -PAL_ERROR_DENIED;
+        }
+
+        /*
+         * In DCAP, retrieving the SGX quote may return error AESM_ATT_KEY_NOT_INITIALIZED (42),
+         * which means that the attestation key is not available and AESM service must re-generate
+         * the key. When Gramine sees such error, it must perform a new INIT_QUOTE_REQUEST and then
+         * re-try retrieving the SGX quote. Note that after INIT_QUOTE_REQUEST, the targetinfo of
+         * Quoting Enclave (QE) may change and must be updated in SGX enclave (in g_qe_targetinfo).
+         *
+         * In our OCALLs, AESM_ATT_KEY_NOT_INITIALIZED is transformed into EAGAIN, and
+         * INIT_QUOTE_REQUEST corresponds to ocall_get_qe_targetinfo().
+         */
+        ret = ocall_get_quote(spid, linkable, &report, nonce, quote, quote_len);
+        if (ret < 0 && ret != -EAGAIN) {
+            log_error("Failed to get quote (error code=%d)", ret);
+            return unix_to_pal_error(ret);
+        }
+
+        if (!ret) {
+            /* success */
+            return 0;
+        }
+
+        assert(ret == -EAGAIN);
+        retries++;
     }
 
-    ret = ocall_get_quote(spid, linkable, &report, nonce, quote, quote_len);
-    if (ret < 0) {
-        log_error("Failed to get quote");
-        return unix_to_pal_error(ret);
-    }
-    return 0;
+    log_error("Failed to get quote after %d attempts", retries);
+    return -PAL_ERROR_DENIED;
 }
--- a/pal/src/host/linux-sgx/host_ocalls.c
+++ b/pal/src/host/linux-sgx/host_ocalls.c
@@ -722,6 +722,12 @@ static long sgx_ocall_get_quote(void* args) {
                           &ocall_quote_args->quote_len);
 }
 
+static long sgx_ocall_get_qe_targetinfo(void* args) {
+    struct ocall_get_qe_targetinfo* ocall_qe_ti_args = args;
+    return init_quoting_enclave_targetinfo(ocall_qe_ti_args->is_epid,
+                                           &ocall_qe_ti_args->qe_targetinfo);
+}
+
 static long sgx_ocall_edmm_modify_pages_type(void* _args) {
     struct ocall_edmm_modify_pages_type* args = _args;
     return edmm_modify_pages_type(args->addr, args->count, args->type);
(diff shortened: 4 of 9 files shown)


llly (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1881#-:-Nxp7ba60y1xBl5xYpU6:bnfp4nl)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

mkow (changes requested):
> Reviewed 5 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1881)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/1881#-NyH9W7l02crLc1V6HvY):*




_Suggestion:_
```
Re-init attestation key after an AESM restart
```
___
*[`-- commits` line 15 at r1](https://reviewable.io/reviews/gramineproject/gramine/1881#-NyHGXdWB8pWWTYLuFgc):*
ditto later in this sentence



_Suggestion:_
```
sending
```
___
*[`pal/src/host/linux-sgx/enclave_ocalls.h` line 136 at r1](https://reviewable.io/reviews/gramineproject/gramine/1881#-NyHIeFu0XXtjx5jpNv8:-NyHIeFu0XXtjx5jpNv9:b9j56u0) ([raw file](https://github.com/gramineproject/gramine/blob/95dbd9c574fc001ca10963baa956c4b84687f128/pal/src/host/linux-sgx/enclave_ocalls.h#L136)):*
> ```objc
>  *
>  * \param      is_epid            True if EPID is used, false if DCAP/ECDSA is used.
>  * \param[out] qe_targetinfo      Retrieved Quoting Enclave's target info; buffer must be submitted
> ```

I'd drop this note altogether, isn't it obvious? It's a single-level pointer, so there's no way for the function to allocate a buffer by itself and return it.



_Suggestion:_
```objc
buffer must be provided
```
___
*[`pal/src/host/linux-sgx/enclave_ocalls.h` line 141 at r1](https://reviewable.io/reviews/gramineproject/gramine/1881#-NyHJ0o26dkDY0ee4Urz:-NyHJ0o26dkDY0ee4Us-:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/95dbd9c574fc001ca10963baa956c4b84687f128/pal/src/host/linux-sgx/enclave_ocalls.h#L141)):*
> ```objc
>  * \returns 0 on success, negative Linux error code otherwise.
>  *
>  * The obtained QE Target Info is not validated in any way (i.e., this function does not check
> ```





_Suggestion:_
```objc
this function does not check whether
```
___
*[`pal/src/host/linux-sgx/enclave_ocalls.h` line 144 at r1](https://reviewable.io/reviews/gramineproject/gramine/1881#-NyHJ5Y-2LaDjXeSw_9x:-NyHJ5Y-2LaDjXeSw_9y:bw93ym4) ([raw file](https://github.com/gramineproject/gramine/blob/95dbd9c574fc001ca10963baa956c4b84687f128/pal/src/host/linux-sgx/enclave_ocalls.h#L144)):*
> ```objc
>  * Target Info contents make sense).
>  */
> int ocall_get_qe_targetinfo(bool is_epid, sgx_target_info_t* qe_targetinfo);
> ```

Double checking if I understand the security model here correctly: qe_targetinfo is completely untrusted, and thus we can't assume that we're talking directly to a real QE (we can be talking to a fake one, or be MitM-ed).
This is not a problem, because only the authentic QE has access to the keys to generate a valid quote (so, the fake QE can't), and our report targets whatever we got in `qe_targetinfo` and the MitMer can't re-encrypt it and then send it to a real QE (due to how EREPORT works).
Is this right?
___
*[`pal/src/host/linux-sgx/enclave_ocalls.c` line 2199 at r1](https://reviewable.io/reviews/gramineproject/gramine/1881#-NyHzv0z8zpqVLLHjynw:-NyHzv0z8zpqVLLHjynx:ballemf) ([raw file](https://github.com/gramineproject/gramine/blob/95dbd9c574fc001ca10963baa956c4b84687f128/pal/src/host/linux-sgx/enclave_ocalls.c#L2199)):*
> ```C
>     if (!ocall_qe_ti_args) {
>         sgx_reset_ustack(old_ustack);
>         return -ENOMEM;
> ```

Almost all other places return `-EPERM` in this case.
___
*[`pal/src/host/linux-sgx/enclave_platform.c` line 8 at r1](https://reviewable.io/reviews/gramineproject/gramine/1881#-NyI1VuKCVva3WpbVrem:-NyI1VuKCVva3WpbVren:b1sxylx) ([raw file](https://github.com/gramineproject/gramine/blob/95dbd9c574fc001ca10963baa956c4b84687f128/pal/src/host/linux-sgx/enclave_platform.c#L8)):*
> ```C
> 
> /* this function assumes proper synchronization by callers, such that
>  * g_pal_linuxsgx_state.qe_targetinfo is not racey */
> ```

This should be documented in the header, not here. Also, it's not only about calling this function multiple times concurrently, but also about anything else which could be reading `g_pal_linuxsgx_state.qe_targetinfo` at the same time.
___
*[`pal/src/host/linux-sgx/enclave_platform.c` line 8 at r1](https://reviewable.io/reviews/gramineproject/gramine/1881#-NyI1bYp3x-yEZMYyOry:-NyI1bYp3x-yEZMYyOrz:b-rjsh0h) ([raw file](https://github.com/gramineproject/gramine/blob/95dbd9c574fc001ca10963baa956c4b84687f128/pal/src/host/linux-sgx/enclave_platform.c#L8)):*
> ```C
> 
> /* this function assumes proper synchronization by callers, such that
>  * g_pal_linuxsgx_state.qe_targetinfo is not racey */
> ```

Do we actually have that synchronization in place? Also, it's quite a surprising/risky interface. Why not just use a lock?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1881)* status: 3 of 7 files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @llly, and @mkow)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/1881#-NyH9W7l02crLc1V6HvY):*
Will do during final rebase
___
*[`-- commits` line 15 at r1](https://reviewable.io/reviews/gramineproject/gramine/1881#-NyHGXdWB8pWWTYLuFgc):*

ditto later in this sentence
</blockquote></details>

Will do during final rebase
___
*[`pal/src/host/linux-sgx/enclave_ocalls.h` line 136 at r1](https://reviewable.io/reviews/gramineproject/gramine/1881#-NyHIeFu0XXtjx5jpNv8:-NzSvpLQ7UrgdhT644kc:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/95dbd9c574fc001ca10963baa956c4b84687f128/pal/src/host/linux-sgx/enclave_ocalls.h#L136)):*

I'd drop this note altogether, isn't it obvious? It's a single-level pointer, so there's no way for the function to allocate a buffer by itself and return it.
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/enclave_ocalls.h` line 141 at r1](https://reviewable.io/reviews/gramineproject/gramine/1881#-NyHJ0o26dkDY0ee4Urz:-NzSvtbW-rz8Q2sz9Gy9:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/95dbd9c574fc001ca10963baa956c4b84687f128/pal/src/host/linux-sgx/enclave_ocalls.h#L141)):*
> ```objc
>  * \returns 0 on success, negative Linux error code otherwise.
>  *
>  * The obtained QE Target Info is not validated in any way (i.e., this function does not check
> ```

Done.
___
*[`pal/src/host/linux-sgx/enclave_ocalls.h` line 144 at r1](https://reviewable.io/reviews/gramineproject/gramine/1881#-NyHJ5Y-2LaDjXeSw_9x:-NzStA0s6U-3X1KNYS03:bzc2tlp) ([raw file](https://github.com/gramineproject/gramine/blob/95dbd9c574fc001ca10963baa956c4b84687f128/pal/src/host/linux-sgx/enclave_ocalls.h#L144)):*
This is correct. Though one of your phrases is vague:

> the MitMer can't re-encrypt it and then send it to a real QE

What do you mean by re-encrypt? The MitMer can't replace the measurements in the intercepted SGX Report (like MRENCLAVE), because then the MAC of the SGX Report won't correspond to the overridden values. And the MitMer cannot generate the new MAC because she doesn't have the corresponding SGX key (with which the SGX Report was signed). And if the MitMer would call EREPORT instruction, then she won't be able to use victim-enclave measurements, because EREPORT + EINIT disallow this.
___
*[`pal/src/host/linux-sgx/enclave_ocalls.c` line 2199 at r1](https://reviewable.io/reviews/gramineproject/gramine/1881#-NyHzv0z8zpqVLLHjynw:-NzStv3_3bNWiKyjq1PR:b-tv6scw) ([raw file](https://github.com/gramineproject/gramine/blob/95dbd9c574fc001ca10963baa956c4b84687f128/pal/src/host/linux-sgx/enclave_ocalls.c#L2199)):*

Almost all other places return `-EPERM` in this case.
</blockquote></details>

Done, though this doesn't sound right. But here it's not important which error code we return, so whatever.
___
*[`pal/src/host/linux-sgx/enclave_platform.c` line 8 at r1](https://reviewable.io/reviews/gramineproject/gramine/1881#-NyI1bYp3x-yEZMYyOry:-NzSuM8X-CwxO1YJ0BZ7:b-2ui8x9) ([raw file](https://github.com/gramineproject/gramine/blob/95dbd9c574fc001ca10963baa956c4b84687f128/pal/src/host/linux-sgx/enclave_platform.c#L8)):*

Do we actually have that synchronization in place? Also, it's quite a surprising/risky interface. Why not just use a lock?
</blockquote></details>

We always had this assumption: https://github.com/gramineproject/gramine/blob/929bb9d156597bca0694d36fe10559555322d7a1/libos/src/fs/dev/attestation.c#L13-L15

Even if we would need a lock, it shouldn't be at this level, but rather inside LibOS.
___
*[`pal/src/host/linux-sgx/enclave_platform.c` line 8 at r1](https://reviewable.io/reviews/gramineproject/gramine/1881#-NyI1VuKCVva3WpbVrem:-NzSunT666VCbs4uKXJD:b-p2142v) ([raw file](https://github.com/gramineproject/gramine/blob/95dbd9c574fc001ca10963baa956c4b84687f128/pal/src/host/linux-sgx/enclave_platform.c#L8)):*
Done.

> it's not only about calling this function multiple times concurrently, but also about anything else which could be reading `g_pal_linuxsgx_state.qe_targetinfo` at the same time

Yes, but the only function that operates on this global var's field is this one (`sgx_get_quote()`). Do you want me to add something additionally?

woju:
Jenkins, test this please

mkow (changes requested):
> Reviewed 3 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1881)* status: 6 of 7 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @llly)

___
*[`pal/src/host/linux-sgx/enclave_platform.c` line 8 at r1](https://reviewable.io/reviews/gramineproject/gramine/1881#-NyI1VuKCVva3WpbVrem:-NzYB45F2IEjnd4p9iXn:b-29lx22) ([raw file](https://github.com/gramineproject/gramine/blob/95dbd9c574fc001ca10963baa956c4b84687f128/pal/src/host/linux-sgx/enclave_platform.c#L8)):*

Done.

> it's not only about calling this function multiple times concurrently, but also about anything else which could be reading `g_pal_linuxsgx_state.qe_targetinfo` at the same time

Yes, but the only function that operates on this global var's field is this one (`sgx_get_quote()`). Do you want me to add something additionally?
</blockquote></details>

Yes. Right now that line is:
```C
sgx_target_info_t qe_targetinfo; /* received from untrusted host, use carefully */
```

This comment sounds like this field was received from the host and was constant afterwards. I can easily see how we could add some functionality using this and not notice that it can change.

Actually, maybe it would be cleaner if we moved that field from `struct pal_linuxsgx_state` to a private variable in `enclave_platform.c` and initialize in on `ECALL_ENCLAVE_START`? (via calling something like `initialize_quoting()` which would copy the enclave start argument into that private, global variable)
It should be obvious then that this variable is only for the use of `sgx_quote()` and would be much harder to misuse.
This way you could also lock accesses to it properly and remove that tricky assumption from `sgx_get_quote()`.
___
*[`pal/src/host/linux-sgx/enclave_platform.c` line 8 at r1](https://reviewable.io/reviews/gramineproject/gramine/1881#-NyI1bYp3x-yEZMYyOry:-NzYNE4H3rEynzqdJCrp:bi73xjx) ([raw file](https://github.com/gramineproject/gramine/blob/95dbd9c574fc001ca10963baa956c4b84687f128/pal/src/host/linux-sgx/enclave_platform.c#L8)):*

We always had this assumption: https://github.com/gramineproject/gramine/blob/929bb9d156597bca0694d36fe10559555322d7a1/libos/src/fs/dev/attestation.c#L13-L15

Even if we would need a lock, it shouldn't be at this level, but rather inside LibOS.
</blockquote></details>

See above. I think it's obvious from the comment you quoted that e.g. setting `/dev/attestation/user_report_data` and reading `/dev/attestation/quote` cannot be concurrent, but it would still be surprising to me that I cannot _read_ `/dev/attestation/quote` concurrently (because it's just reading).

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1881)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`pal/src/host/linux-sgx/enclave_platform.c` line 14 at r3](https://reviewable.io/reviews/gramineproject/gramine/1881#-NzYovTL4cnupIRXe0RI:-NzYovTL4cnupIRXe0RJ:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/002e4efe9b9614248675f1265c19e21fdd65bbd8/pal/src/host/linux-sgx/enclave_platform.c#L14)):*
> ```C
> 
> int init_qe_targetinfo(void* uptr_qe_targetinfo) {
>     /* no need to acquire g_qe_targetinfo_lock at this point since there is a single thread */
> ```





_Suggestion:_
```C
since there is only a single thread */
```
___
*[`pal/src/host/linux-sgx/enclave_platform.c` line 34 at r3](https://reviewable.io/reviews/gramineproject/gramine/1881#-NzYw9Q_3asMMyGGGgYJ:-NzYw9Q_3asMMyGGGgYK:b-ulrc07) ([raw file](https://github.com/gramineproject/gramine/blob/002e4efe9b9614248675f1265c19e21fdd65bbd8/pal/src/host/linux-sgx/enclave_platform.c#L34)):*
> ```C
>         if (retries) {
>             /* new attempt, need to update QE target info */
>             ret = ocall_get_qe_targetinfo(/*is_epid=*/!!spid, &g_qe_targetinfo);
> ```

nitpick: you could OCALL into a local variable, and then only `memcpy` under the spinlock
...which you actually do anyways, below, so just swap these two :)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1881)* status: 8 of 9 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`pal/src/host/linux-sgx/enclave_platform.c` line 14 at r3](https://reviewable.io/reviews/gramineproject/gramine/1881#-NzYovTL4cnupIRXe0RI:-NzcSsAF4X_x1RTFNadn:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/002e4efe9b9614248675f1265c19e21fdd65bbd8/pal/src/host/linux-sgx/enclave_platform.c#L14)):*
> ```C
> 
> int init_qe_targetinfo(void* uptr_qe_targetinfo) {
>     /* no need to acquire g_qe_targetinfo_lock at this point since there is a single thread */
> ```

Done.
___
*[`pal/src/host/linux-sgx/enclave_platform.c` line 34 at r3](https://reviewable.io/reviews/gramineproject/gramine/1881#-NzYw9Q_3asMMyGGGgYJ:-NzcU8ET5qkRQpziyKS8:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/002e4efe9b9614248675f1265c19e21fdd65bbd8/pal/src/host/linux-sgx/enclave_platform.c#L34)):*

nitpick: you could OCALL into a local variable, and then only `memcpy` under the spinlock
...which you actually do anyways, below, so just swap these two :)
</blockquote></details>

Done.

kailun-qin (approved):
> Reviewed 3 of 6 files at r1, 2 of 4 files at r2, 3 of 4 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1881)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1881)* status: all files reviewed, 2 unresolved discussions (waiting on @mkow)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/1881#-NyH9W7l02crLc1V6HvY):*

Will do during final rebase
</blockquote></details>

Done.
___
*[`-- commits` line 15 at r1](https://reviewable.io/reviews/gramineproject/gramine/1881#-NyHGXdWB8pWWTYLuFgc):*

Will do during final rebase
</blockquote></details>

Done.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1881)* status: all files reviewed, 2 unresolved discussions (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1881#-NzgUuNJCL55MpjnlTjB:-NzgUuNJCL55MpjnlTjC:bxi76s5) (no related file):*
After final rebase, tested this PR one last time on a DCAP system -- everything works.

___

kailun-qin (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1881)* status: all files reviewed, 2 unresolved discussions (waiting on @mkow)

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1881)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #1911, opened 2024-06-17
DONTMERGE [common] Update ciphersuite used for securing pipe/UDS connections

dimakuv:
## Description of the changes 

Previously, Gramine-SGX used MBEDTLS_TLS_PSK_WITH_AES_128_GCM_SHA256 ciphersuite when creating TLS-secured pipes and Unix Domain Sockets. This ciphersuite is considered deprecated, thus we switch to MBEDTLS_TLS_ECDHE_PSK_WITH_AES_128_CBC_SHA256 (with Elliptic Curve DHE key exchange). This is recommended for Forward Security.

Unfortunately mbedTLS v3.6.0 doesn't provide an ECDHE PSK based ciphersuite that has AES-GCM, but only the one with AES-CBC. For the lack of options, we use the AES-CBC one.

**DONTMERGE:**
- This hampers performance of applications that use pipes/UDSes a lot.
- Serialization of TLS sessions (during fork) is broken because our patch only works for GCM mode (only checkpoints CTR counter).

There is an alternative PR that uses DHE: https://github.com/gramineproject/gramine/pull/1918

More info see also in https://github.com/Mbed-TLS/mbedtls/issues/8170

## How to test this PR? 

CI is enough.

CI currently fails on forking apps, because serialization of TLS sessions doesn't work correctly. That's because of our [minimalistic patch](https://github.com/gramineproject/gramine/blob/master/subprojects/packagefiles/mbedtls/gramine.patch).

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1911)

PR #1832, opened 2024-04-02
[mbedtls] Upgrade mbedTLS to version 3.6.0

kailun-qin:
## Description of the changes 
This commit introduces the following changes along with the upgrade of mbedTLS to version 3.6.0:
* updating the sonames of mbedTLS-produced libs;
* using the uploaded release asset from the release tag instead of the GitHub-generated one to avoid the involvement of Git submodules;
* updating the subproject name from `mbedtls-mbedtls-` to `mbedtls-` in accordance with the above asset change;
* initializing PSA crypto in `ra-tls-mbedtls` example and `secret-prov` libs, required for TLS 1.3 -- enabled by default since this mbedTLS version.

This mbedTLS version includes fixes for CVE-2024-28755, CVE-2024-28836 and CVE-2024-28960.

## How to test this PR? 
CI

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1832)

PR #1918, opened 2024-06-20
DONTMERGE [common] Update ciphersuite used for securing pipe/UDS connections

dimakuv:
## Description of the changes 

Previously, Gramine-SGX used MBEDTLS_TLS_PSK_WITH_AES_128_GCM_SHA256 ciphersuite when creating TLS-secured pipes and Unix Domain Sockets. This ciphersuite is considered deprecated, thus we switch to MBEDTLS_TLS_DHE_PSK_WITH_AES_128_GCM_SHA256. This is recommended for perfect forward secrecy.

Unfortunately mbedTLS v3.6.0 doesn't provide an ECDHE PSK based ciphersuite that has AES-GCM, but only the one with AES-CBC. For the lack of options, we use the DHE PSK with AES-GCM.

DONTMERGE:
- This hampers performance of applications that use pipes/UDSes a lot.

This is an alternative to https://github.com/gramineproject/gramine/pull/1911.

More info see also in https://github.com/Mbed-TLS/mbedtls/issues/8170

## How to test this PR? 

CI is enough.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1918)

### DIFF
--- a/common/src/crypto/adapters/mbedtls_adapter.c
+++ b/common/src/crypto/adapters/mbedtls_adapter.c
@@ -343,7 +343,7 @@ int lib_SSLInit(LIB_SSL_CONTEXT* ssl_ctx, int stream_fd, bool is_server, const u
 
     memset(ssl_ctx, 0, sizeof(*ssl_ctx));
 
-    ssl_ctx->ciphersuites[0] = MBEDTLS_TLS_PSK_WITH_AES_128_GCM_SHA256;
+    ssl_ctx->ciphersuites[0] = MBEDTLS_TLS_DHE_PSK_WITH_AES_128_GCM_SHA256;
     memset(&ssl_ctx->ciphersuites[1], 0, sizeof(ssl_ctx->ciphersuites[1]));
 
     ssl_ctx->pal_recv_cb = pal_recv_cb;
--- a/subprojects/packagefiles/mbedtls/include/mbedtls/config-pal.h
+++ b/subprojects/packagefiles/mbedtls/include/mbedtls/config-pal.h
@@ -38,7 +38,6 @@
 #define MBEDTLS_HAVE_X86_64
 #endif
 #define MBEDTLS_HKDF_C
-#define MBEDTLS_KEY_EXCHANGE_PSK_ENABLED
 #define MBEDTLS_MD_C
 #define MBEDTLS_NET_C
 #define MBEDTLS_NO_PLATFORM_ENTROPY
@@ -49,9 +48,15 @@
 #define MBEDTLS_PLATFORM_ZEROIZE_ALT
 #define MBEDTLS_RSA_C
 #define MBEDTLS_SHA256_C
-#define MBEDTLS_SSL_CIPHERSUITES MBEDTLS_TLS_PSK_WITH_AES_128_GCM_SHA256
 #define MBEDTLS_SSL_CLI_C
 #define MBEDTLS_SSL_CONTEXT_SERIALIZATION
 #define MBEDTLS_SSL_PROTO_TLS1_2
 #define MBEDTLS_SSL_SRV_C
 #define MBEDTLS_SSL_TLS_C
+
+/* below features are to implement DHE-PSK based secure-pipe sessions */
+#define MBEDTLS_KEY_EXCHANGE_DHE_PSK_ENABLED
+/* We use DHE (slightly weaker than ECDHE) because it's the strongest PSK ciphersuite that offers
+ * AES-128 and GCM; note that there is no GCM or CCM support for ECDHE-PSK yet. See issue
+ * https://github.com/Mbed-TLS/mbedtls/issues/8170 that tracks ECDHE-AES-GCM support status. */
+#define MBEDTLS_SSL_CIPHERSUITES MBEDTLS_TLS_DHE_PSK_WITH_AES_128_GCM_SHA256


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1918)* status: 0 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "DONTMERGE" found in commit messages' one-liners

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1918#-O-ojBhm8g41PrS9LeGI:-O-ojBhm8g41PrS9LeGJ:b-gw3y3b) (no related file):*
Functionality-wise everything works.

But performance-wise this PR significantly deteriorates performance:
```
$ gramineproject/gramine/libos/test/regression$ SGX=1 gramine-test pytest -v

With DHE (branch dimakuv/change-mbedtls-key-exchange-to-dhe): 355.70 sec
With RSA (current master): 291.19 sec
```

I'm really surprised by these numbers, since DHE is only involved during TLS handshake, and the rest is done using the same AES-GCM crypto. Is it really so that DHE handshake is that much slower than RSA?

I think with such performance overheads, this change is a no-no.

___

SabaEskandarian:
I had a chance to talk to @donporter about this yesterday, so I thought I'd chime in. 

I think the cost may be coming from the fact that you've switched from a cipher suite that uses only symmetric key encryption to one that uses public key encryption. TLS_PSK_WITH_AES_128_GCM_SHA256 just has the parties use a pre-shared key with no RSA or Diffie-Hellman key exchange involved. TLS_DHE_PSK_WITH_AES_128_GCM_SHA256 does all the things that were being done with just PSK but additionally does an ephemeral Diffie-Hellman key exchange to get forward secrecy and prevent dictionary attacks. Going from no public key crypto to using public key crypto is going to be costly -- [wolfSSL reports](https://www.wolfssl.com/when-to-use-pre-shared-key-psk-cipher-suites-2/) a 2x slowdown using RSA vs just PSK.

I think the best performance you could hope for with forward secrecy would be ECDHE with AES_128_GCM, but this combination doesn't seem to be supported by mbedTLS. Maybe you could ask for it to be added? Elliptic curve diffie hellman gets the same security level with smaller field elements, so it saves on communication/computation costs versus DHE, and AES-GCM benefits from hardware acceleration and provides built-in ciphertext integrity, saving the cost of separately computing MACs on messages compared to AES-CBC (mentioned in the alternative proposal). Even if mbedTLS adds ECDHE with AES-GCM, there will still be a performance hit.

If you're in the process of changing TLS stuff, it might be worth thinking about moving to TLS 1.3, which drops support for older cipher suites and provides opportunities for faster connection setup (although paying something for forward secrecy remains unavoidable).

mkow:
> I think the best performance you could hope for with forward secrecy would be [...]

But I don't think we gain anything from forward secrecy? The symmetric keys we use are ephemeral anyways, and exchanged via hardware means. As far as I understand we just got forced into it, because mbedtls is dropping all the combinations without forward secrecy, right? So, if we want to request anything from them, then it would be to preserve some non-forward-secrecy PSK combination.

> If you're in the process of changing TLS stuff, it might be worth thinking about moving to TLS 1.3 [...]

I think we're already on 1.3, see https://github.com/gramineproject/gramine/pull/1832.

dimakuv (commented):
Thank you @SabaEskandarian very much for your inputs!

> I think the cost may be coming from the fact that you've switched from a cipher suite that uses only symmetric key encryption to one that uses public key encryption. TLS_PSK_WITH_AES_128_GCM_SHA256 just has the parties use a pre-shared key with no RSA or Diffie-Hellman key exchange involved. TLS_DHE_PSK_WITH_AES_128_GCM_SHA256 does all the things that were being done with just PSK but additionally does an ephemeral Diffie-Hellman key exchange to get forward secrecy and prevent dictionary attacks. 

Yes, thank you, that was my mistake. You're right that the TLS_PSK_WITH_AES_128_GCM_SHA256 ciphersuite (currently used in Gramine) doesn't use public key encryption.

> Going from no public key crypto to using public key crypto is going to be costly -- [wolfSSL reports](https://www.wolfssl.com/when-to-use-pre-shared-key-psk-cipher-suites-2/) a 2x slowdown using RSA vs just PSK.

Thanks for this link. I didn't expect such a huge slowdown. This may explain the overheads I observe in this PR.

> I think the best performance you could hope for with forward secrecy would be ECDHE with AES_128_GCM, but this combination doesn't seem to be supported by mbedTLS. Maybe you could ask for it to be added? Elliptic curve diffie hellman gets the same security level with smaller field elements, so it saves on communication/computation costs versus DHE, and AES-GCM benefits from hardware acceleration and provides built-in ciphertext integrity, saving the cost of separately computing MACs on messages compared to AES-CBC (mentioned in the alternative proposal). Even if mbedTLS adds ECDHE with AES-GCM, there will still be a performance hit.

Thanks, I added your response to mbedTLS's discussion: https://github.com/Mbed-TLS/mbedtls/issues/8170

> But I don't think we gain anything from forward secrecy? The symmetric keys we use are ephemeral anyways, and exchanged via hardware means. As far as I understand we just got forced into it, because mbedtls is dropping all the combinations without forward secrecy, right? So, if we want to request anything from them, then it would be to preserve some non-forward-secrecy PSK combination.

Yes, I agree that our particular use case doesn't need *additional* forward secrecy. I think we could continue using TLS_PSK_WITH_AES_128_GCM_SHA256. (Again, not a crypto expert.) I'll add your comment to mbedTLS issue.

> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1918#-:-O0Nzv_Y-7rbYv2lGWHE:b-o18xqy)* status: 0 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "DONTMERGE" found in commit messages' one-liners

mpg:
> MBEDTLS_TLS_PSK_WITH_AES_128_GCM_SHA256 [...] is considered deprecated

I'm surprised. Can you point to the document that deprecates it?

> As far as I understand we just got forced into it, because mbedtls is dropping all the combinations without forward secrecy, right?

Speaking as an Mbed TLS maintainer, I don't think we are! We are dropping the (plain) [RSA and RSA-PSK](https://github.com/Mbed-TLS/mbedtls/issues/8170) key exchanges for several reasons, including reliance on PKCS#1 v1.5 encryption which is notoriously hard to implement securely (see the long series of attacks starting with Bleichenbacher's, the latest of which being [MARVIN](https://people.redhat.com/~hkario/marvin/)), being officially deprecated, and lacking forward secrecy _despite_ incurring the cost of asymmetric crypto.

This last bit is important: it is well understood within the TLS community that some sometimes you can't afford the computational (and memory) costs of asymmetric crypto, or simply don't need any of the things it provides, and the PSK key exchange is there for that. I'll note that TLS 1.3, which removed a lot of old stuff (including RSA PKCS#1 v1.5, and RSA encryption entirely - only RSA-PSS signatures are kept), did keep a PSK key exchange mode, alongside Ephemeral-PSK and Ephemeral.

Mbed TLS implements all 3 keys exchange modes in TLS 1.3, including (non-ephemeral) PSK, and I don't think we have any intention of removing the pure-PSK key exchange in TLS 1.2 (as opposed to RSA-PSK which will be dropped due to the RSA part, and DHE-PSK dropped due to [the DH part in 1.2](https://github.com/Mbed-TLS/mbedtls/issues/5278) - but we'll keep FFDH in 1.3).

dimakuv:
As explained by @mpg in https://github.com/gramineproject/gramine/pull/1918#issuecomment-2196387016, there is no deprecation of the currently-used `MBEDTLS_TLS_PSK_WITH_AES_128_GCM_SHA256`. Closing this PR.

PR #1934, opened 2024-07-03
[PAL/Linux-SGX] Add locking around read/write on encrypted pipes

dimakuv:
## Description of the changes 

Pipes in Linux-SGX PAL are automatically encrypted using mbedTLS's SSL/TLS sessions (aka contexts). By design, an mbedTLS SSL/TLS context is assumed to be used within a single thread and thus does not use any locking. In Gramine, though, the pipe and its associated SSL/TLS context can be used in multiple threads. Without protecting pipe read/write operations with a lock, mbedTLS internal handling of the context would exhibit data races, leading to e.g. -EACCES in Gramine.

This bug was detected on a private workload. This bug can be considered a regression, introduced in [commit "[LibOS] Do not perform lock(&hdl->pos_lock) on non-seekable handles"](https://github.com/gramineproject/gramine/commit/ec06c32366672b3dfc19f539d16d37c6449ab4ca). That commit, among other things, removed the locking around LibOS-level pipe read/write operations; the locks were supposed to guard only "file position", which is meaningless for pipes and that's why that commit removed the locking. However, that locking (by accident) also protected the mbedTLS SSL/TLS context, and removing it resulted in some workloads failing (pipe reads/writes returning -EACCES) due to internal mbedTLS data races.

See for the mbedTLS threading discussion e.g. this thread: https://github.com/Mbed-TLS/mbedtls/issues/564

## How to test this PR? 

Detected on a private workload. ~~I don't know if it's worth it to write a LibOS test that stresses the data race.~~

Added a new LibOS test:
- without this PR:
```
~/gramineproject/gramine/libos/test/regression$ gramine-sgx pipe_race
pipe_race: error at write(pipefds[1], buf, strlen(buf)) (line 45): Permission denied: Permission denied
```

- with this PR:
```
~/gramineproject/gramine/libos/test/regression$ gramine-sgx pipe_race
TEST OK
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1934)

### DIFF
--- a/libos/test/regression/meson.build
+++ b/libos/test/regression/meson.build
@@ -88,6 +88,7 @@ tests = {
     'pipe': {},
     'pipe_nonblocking': {},
     'pipe_ocloexec': {},
+    'pipe_race': {},
     'poll': {},
     'poll_closed_fd': {},
     'poll_many_types': {},
--- a/libos/test/regression/pipe_race.c
+++ b/libos/test/regression/pipe_race.c
@@ -1,0 +1,81 @@
+/* SPDX-License-Identifier: LGPL-3.0-or-later */
+/* Copyright (C) 2024 Intel Corporation */
+
+/*
+ * Pipes in e.g. Linux-SGX PAL are automatically encrypted using mbedTLS's SSL/TLS sessions (aka
+ * contexts). By design, an mbedTLS SSL/TLS context is assumed to be used within a single thread and
+ * thus does not use any locking. In Gramine, though, the pipe and its associated SSL/TLS context
+ * can be used in multiple threads. Without protecting pipe read/write operations with a lock,
+ * mbedTLS internal handling of the context would exhibit data races, leading to e.g. -EACCES in
+ * Gramine. This test checks that locking is correct and that no data races occur.
+ */
+
+#define _XOPEN_SOURCE 700
+#include <pthread.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
+#include "common.h"
+
+#define WRITER_THREADS 4
+#define ITERATIONS     100000
+#define BUF_SIZE       64 /* must be safe to not overflow, see users of this macro */
+
+static uint32_t threads_started  = 0;
+static uint32_t threads_finished = 0;
+
+static void pthread_check(int x) {
+    if (x) {
+        errx(1, "pthread failed with %d", x);
+    }
+}
+
+static void* writer_thread_func(void* arg) {
+    int* pipefds = arg;
+    uint32_t writer_thread_idx = __atomic_fetch_add(&threads_started, 1, __ATOMIC_SEQ_CST);
+
+    char buf[BUF_SIZE] = {0};
+    CHECK(sprintf(buf, "WRITER THREAD %u\n", writer_thread_idx));
+
+    for (uint64_t i = 0; i < ITERATIONS; i++)
+        CHECK(write(pipefds[1], buf, strlen(buf)));
+
+    if (__atomic_add_fetch(&threads_finished, 1, __ATOMIC_SEQ_CST) == WRITER_THREADS) {
+        /* last writer thread closes the write end of pipe */
+        CHECK(close(pipefds[1]));
+    }
+    return NULL;
+}
+
+int main(int argc, char** argv) {
+    int pipefds[2];
+    CHECK(pipe(pipefds));
+
+    pthread_t th[WRITER_THREADS];
+    for (int i = 0; i < WRITER_THREADS; i++)
+        pthread_check(pthread_create(&th[i], NULL, writer_thread_func, pipefds));
+
+    while (__atomic_load_n(&threads_started, __ATOMIC_SEQ_CST) != WRITER_THREADS)
+        ;
+
+    char buf[BUF_SIZE] = {0};
+    while (true) {
+        ssize_t bytes_read = CHECK(read(pipefds[0], buf, sizeof(buf) - 1));
+        if (!bytes_read)
+            break;
+#ifdef DEBUG_TEST /* declare to see the test output during debugging */
+        printf("%s", buf);
+#endif
+    }
+    CHECK(close(pipefds[0]));
+
+    for (int i = 0; i < WRITER_THREADS; i++)
+        pthread_check(pthread_join(th[i], NULL));
+
+    puts("TEST OK");
+    return 0;
+}
--- a/libos/test/regression/test_libos.py
+++ b/libos/test/regression/test_libos.py
@@ -1504,6 +1504,10 @@ class TC_80_Socket(RegressionTestCase):
         stdout, _ = self.run_binary(['pipe_ocloexec'])
         self.assertIn('TEST OK', stdout)
 
+    def test_093_pipe_race(self):
+        stdout, _ = self.run_binary(['pipe_race'])
+        self.assertIn('TEST OK', stdout)
+
     def test_095_mkfifo(self):
         try:
             stdout, _ = self.run_binary(['mkfifo'], timeout=60)
--- a/libos/test/regression/tests.toml
+++ b/libos/test/regression/tests.toml
@@ -87,6 +87,7 @@ manifests = [
   "pipe",
   "pipe_nonblocking",
   "pipe_ocloexec",
+  "pipe_race",
   "poll",
   "poll_closed_fd",
   "poll_many_types",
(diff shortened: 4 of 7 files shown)


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1934)* status: 0 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1934#-O11Y0ax2s_KSlbZVSg0:-O11Y0ax2s_KSlbZVSg1:b-j8tzdu) (no related file):*
I was informed that the fix works on that private workload. So the change itself is good.

I think I still need to create a LibOS test. Blocking myself on this.

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1934)* status: 0 of 7 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/1934#-O11Y0ax2s_KSlbZVSg0:-O1GUvCM3gqQbnGDf44N:bb6un1j) (no related file):*

I was informed that the fix works on that private workload. So the change itself is good.

I think I still need to create a LibOS test. Blocking myself on this.
</blockquote></details>

Done

___

mkow (changes requested):
> Reviewed 2 of 2 files at r1, 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1934)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`pal/src/host/linux-sgx/pal_host.h` line 74 at r2](https://reviewable.io/reviews/gramineproject/gramine/1934#-O1m9LRRBFPL3UK3bdHX:-O1m9LRRBFPL3UK3bdHY:b-rjvw3k) ([raw file](https://github.com/gramineproject/gramine/blob/38c61db002c27736095e8e9fd87899a2184b6361/pal/src/host/linux-sgx/pal_host.h#L74)):*
> ```objc
>              * Gramine, though, the pipe and its associated SSL/TLS context can be used in multiple
>              * threads. Without protecting pipe read/write operations with a lock, mbedTLS internal
>              * handling of the context would exhibit data races, leading to e.g. -EACCES in Gramine.
> ```

I'd skip that, it's a random fallout of many levels of error propagation and may change in the future (assuming the bug would be re-introduced).



_Code quote:_
```objc
, leading to e.g. -EACCES in Gramine.
```
___
*[`pal/src/host/linux-sgx/pal_host.h` line 90 at r2](https://reviewable.io/reviews/gramineproject/gramine/1934#-O1m9l9DCThF0aNGkOA0:-O1m9l9DCThF0aNGkOA1:b-6ortet) ([raw file](https://github.com/gramineproject/gramine/blob/38c61db002c27736095e8e9fd87899a2184b6361/pal/src/host/linux-sgx/pal_host.h#L90)):*
> ```objc
>              * Since we're in PAL layer, we can only use spinlocks (not e.g. futexes). It is
>              * suboptimal (for blocking operations), but the hope is that multiple threads
>              * simultaneously operating on the pipe is a rare scenario.
> ```

Theoretically we could fix this by introducing `try_lock()` and returning `-EAGAIN` if it fails, but not sure if it's worth the effort.
___
*[`pal/src/host/linux-sgx/pal_pipes.c` line 176 at r2](https://reviewable.io/reviews/gramineproject/gramine/1934#-O1m0D-zEQMkmqd5qjQ0:-O1m0D-zEQMkmqd5qjQ1:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/38c61db002c27736095e8e9fd87899a2184b6361/pal/src/host/linux-sgx/pal_pipes.c#L176)):*
> ```C
>  * our end of the pipe. Typically, `pipesrv` handle is not needed after this and can be closed.
>  */
> static int pipe_waitforclient(PAL_HANDLE handle, PAL_HANDLE* client, pal_stream_options_t options) {
> ```





_Suggestion:_
```C
out_client
```
___
*[`pal/src/host/linux-sgx/pal_pipes.c` line 248 at r2](https://reviewable.io/reviews/gramineproject/gramine/1934#-O1m05ON88JnoEYtw0sd:-O1m05ON88JnoEYtw0se:undefined) ([raw file](https://github.com/gramineproject/gramine/blob/38c61db002c27736095e8e9fd87899a2184b6361/pal/src/host/linux-sgx/pal_pipes.c#L248)):*
> ```C
>  * The other end of the pipe is typically of type `pipecli`.
>  */
> static int pipe_connect(PAL_HANDLE* handle, const char* name, pal_stream_options_t options) {
> ```





_Suggestion:_
```C
out_handle
```

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1934)* status: 5 of 7 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`pal/src/host/linux-sgx/pal_host.h` line 74 at r2](https://reviewable.io/reviews/gramineproject/gramine/1934#-O1m9LRRBFPL3UK3bdHX:-O1p72eF0SBI01N2eypO:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/38c61db002c27736095e8e9fd87899a2184b6361/pal/src/host/linux-sgx/pal_host.h#L74)):*

I'd skip that, it's a random fallout of many levels of error propagation and may change in the future (assuming the bug would be re-introduced).
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/pal_host.h` line 90 at r2](https://reviewable.io/reviews/gramineproject/gramine/1934#-O1m9l9DCThF0aNGkOA0:-O1p6STf0VyJ-rfaoKPt:b60a13j) ([raw file](https://github.com/gramineproject/gramine/blob/38c61db002c27736095e8e9fd87899a2184b6361/pal/src/host/linux-sgx/pal_host.h#L90)):*

Theoretically we could fix this by introducing `try_lock()` and returning `-EAGAIN` if it fails, but not sure if it's worth the effort.
</blockquote></details>

Yeah, I also think it is not worth the effort.
___
*[`pal/src/host/linux-sgx/pal_pipes.c` line 176 at r2](https://reviewable.io/reviews/gramineproject/gramine/1934#-O1m0D-zEQMkmqd5qjQ0:-O1p7UnTA2SuIuuuI8wB:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/38c61db002c27736095e8e9fd87899a2184b6361/pal/src/host/linux-sgx/pal_pipes.c#L176)):*
> ```C
>  * our end of the pipe. Typically, `pipesrv` handle is not needed after this and can be closed.
>  */
> static int pipe_waitforclient(PAL_HANDLE handle, PAL_HANDLE* client, pal_stream_options_t options) {
> ```

Done.
___
*[`pal/src/host/linux-sgx/pal_pipes.c` line 248 at r2](https://reviewable.io/reviews/gramineproject/gramine/1934#-O1m05ON88JnoEYtw0sd:-O1p8-buCp3W5YQz11VI:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/38c61db002c27736095e8e9fd87899a2184b6361/pal/src/host/linux-sgx/pal_pipes.c#L248)):*
> ```C
>  * The other end of the pipe is typically of type `pipecli`.
>  */
> static int pipe_connect(PAL_HANDLE* handle, const char* name, pal_stream_options_t options) {
> ```

Done.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1934)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

kailun-qin (approved):
> Reviewed 5 of 5 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1934)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1934)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #1944, opened 2024-07-23
DONTMERGE Rework of PR 1857

dimakuv:
## Description of the changes 

It turned out to be easier to try to implement this myself.

## How to test this PR? 

I was trying Redis in Debug mode (works) and in Release mode (works).

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1944)

### DIFF
--- a/CI-Examples/redis/redis-server.manifest.template
+++ b/CI-Examples/redis/redis-server.manifest.template
@@ -3,6 +3,8 @@
 
 # Redis manifest file example
 
+sgx.enable_stats = true
+
 ################################## GRAMINE ####################################
 
 # Entrypoint binary which Gramine invokes.
--- a/pal/src/host/linux-sgx/enclave_framework.c
+++ b/pal/src/host/linux-sgx/enclave_framework.c
@@ -67,26 +67,18 @@ bool sgx_is_valid_untrusted_ptr(const void* _addr, size_t size, size_t alignment
 }
 
 /*
- * When DEBUG is enabled, we run sgx_profile_sample() during asynchronous enclave exit (AEX), which
+ * We run some functions (e.g. sgx_profile_sample()) during asynchronous enclave exit (AEX), which
  * uses the stack. Make sure to update URSP so that the AEX handler does not overwrite the part of
  * the stack that we just allocated.
  *
  * (Recall that URSP is an outside stack pointer, saved by EENTER and restored on AEX by the SGX
  * hardware itself.)
  */
-#ifdef DEBUG
-
 #define UPDATE_USTACK(_ustack)                           \
     do {                                                 \
         SET_ENCLAVE_TCB(ustack, _ustack);                \
         GET_ENCLAVE_TCB(gpr)->ursp = (uint64_t)_ustack;  \
     } while(0)
-
-#else
-
-#define UPDATE_USTACK(_ustack) SET_ENCLAVE_TCB(ustack, _ustack)
-
-#endif
 
 void* sgx_prepare_ustack(void) {
     void* old_ustack = GET_ENCLAVE_TCB(ustack);
--- a/pal/src/host/linux-sgx/enclave_ocalls.c
+++ b/pal/src/host/linux-sgx/enclave_ocalls.c
@@ -136,8 +136,9 @@ static long sgx_exitless_ocall(uint64_t code, void* ocall_args) {
         }
     }
 
+    long result = COPY_UNTRUSTED_VALUE(&req->result);
     sgx_reset_ustack(old_ustack);
-    return COPY_UNTRUSTED_VALUE(&req->result);
+    return result;
 }
 
 __attribute_no_sanitize_address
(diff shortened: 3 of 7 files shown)


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1944)* status: 0 of 5 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "DONTMERGE" and "fixup! " found in commit messages' one-liners

___
*[`pal/src/host/linux-sgx/host_entry.S` line 82 at r1](https://reviewable.io/reviews/gramineproject/gramine/1944#-O2eJTkfE_tsW3B1rGLp:-O2eJTkfE_tsW3B1rGLq:b-4y1kn8) ([raw file](https://github.com/gramineproject/gramine/blob/05b7adb0e4511f5952acd4a19460f28863eeddd6/pal/src/host/linux-sgx/host_entry.S#L82)):*
> ```
>     #        RPC request objects, as the AEX flow here. It just so happens that the total size of
>     #        allocated RPC request objects is less than 128B (red zone size), so I reuse this const.
>     #        This double-stack-usage needs a proper fix.
> ```

These are the references:
- https://github.com/gramineproject/gramine/blob/951795e12bc25ca16ae76d429e3b2a84cd84a811/pal/src/host/linux-sgx/enclave_framework.c#L95-L96
- https://github.com/gramineproject/gramine/blob/951795e12bc25ca16ae76d429e3b2a84cd84a811/pal/src/host/linux-sgx/enclave_ocalls.c#L54-L55

Note how the red zone is accounted for in the preparation of the untrusted stack, and then the RPC request object is allocated from below that red zone. That's why we need two red zones in this workaround here.

This problem is unrelated to this PR really. Fundamentally this AEX handler and the OCALL logic re-use the same stack, and can do it at the same time. We were just lucky that the AEX handler didn't use the stack, or used the stack but only in DEBUG mode.

I feel like the OCALL stack should be a **different memory region** than the stack. Guess that's the correct solution.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1944)* status: 0 of 7 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " and "DONTMERGE" found in commit messages' one-liners

___
*[`pal/src/host/linux-sgx/host_entry.S` line 75 at r2](https://reviewable.io/reviews/gramineproject/gramine/1944#-O2i_uGk-i46EfO_4r2_:-O2i_uGlAHHsD9EzqEp4:b-voac8q) ([raw file](https://github.com/gramineproject/gramine/blob/a4aeacf6b7e95241f243bdb55fce6be710112a34/pal/src/host/linux-sgx/host_entry.S#L75)):*
> ```
> 
>     # Inform that we are in AEX profiling code
>     movb $1, %gs:PAL_HOST_TCB_IN_AEX_PROF
> ```

I think the name of this variable/macro should be changed, dropping `_PROF`. Now AEX doesn't do only profiling but more.

Also, this asm line should be moved before the previous (`lock inc`) line.

dimakuv:
Jenkins, test this please

dimakuv:
All code was moved into the main PR #1857, so I'm closing this "proof of concept" PR

dimakuv:
*[`pal/src/host/linux-sgx/host_entry.S` line 82 at r1](https://reviewable.io/reviews/gramineproject/gramine/1944#-O2eJTkfE_tsW3B1rGLp:-O99-AOS5sc09u-NCJfk:b-5bdxcz) ([raw file](https://github.com/gramineproject/gramine/blob/05b7adb0e4511f5952acd4a19460f28863eeddd6/pal/src/host/linux-sgx/host_entry.S#L82)):*

These are the references:
- https://github.com/gramineproject/gramine/blob/951795e12bc25ca16ae76d429e3b2a84cd84a811/pal/src/host/linux-sgx/enclave_framework.c#L95-L96
- https://github.com/gramineproject/gramine/blob/951795e12bc25ca16ae76d429e3b2a84cd84a811/pal/src/host/linux-sgx/enclave_ocalls.c#L54-L55

Note how the red zone is accounted for in the preparation of the untrusted stack, and then the RPC request object is allocated from below that red zone. That's why we need two red zones in this workaround here.

This problem is unrelated to this PR really. Fundamentally this AEX handler and the OCALL logic re-use the same stack, and can do it at the same time. We were just lucky that the AEX handler didn't use the stack, or used the stack but only in DEBUG mode.

I feel like the OCALL stack should be a **different memory region** than the stack. Guess that's the correct solution.
</blockquote></details>

This comment is wrong and stale. The problem was in other two places, see the changes in this PR in files `enclave_framework.c` and `enclave_ocalls.c`.

Note that the latter place (actual bug) was fixed in a separate PR: https://github.com/gramineproject/gramine/pull/1955

PR #1955, opened 2024-07-26
[PAL/Linux-SGX] Read exitless-OCALL result before resetting the stack

dimakuv:
## Description of the changes 

Previously, there was a data race that the exitless-OCALL logic read the result of the OCALL from the stack *after* the code reset the stack. However, this stack is shared with the AEX flows. As of now, the AEX flows use the stack only in debug mode, so this data race flew under the radar for a long time. What can happen is that right-before reading the result of the exitless OCALL, the enclave thread is interrupted, an AEX logic is executed and modifies the values on the stack, then the exitless-OCALL logic is resumed, and the enclave thread reads an AEX-modified OCALL result value.

Future commits (e.g., AEX-Notify) will introduce more AEX flows and expose this data race. So let's fix it now.

This was detected while working on #1857 and #1944. The same problem is exhibited also in AEX-Notify PR: #1531. 

## How to test this PR? 

Run PAL and LibOS regression tests in a loop on e.g. PR #1944 (commenting out this fix first). They have Exitless tests, so they fail periodically.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1955)

### DIFF
--- a/pal/src/host/linux-sgx/enclave_ocalls.c
+++ b/pal/src/host/linux-sgx/enclave_ocalls.c
@@ -136,8 +136,12 @@ static long sgx_exitless_ocall(uint64_t code, void* ocall_args) {
         }
     }
 
+    /* important to copy req->result before resetting the stack, otherwise it may be overwritten;
+     * this enclave's stack is also used in AEX flows, see host_entry.S:async_exit_pointer() */
+    long result = COPY_UNTRUSTED_VALUE(&req->result);
     sgx_reset_ustack(old_ustack);
-    return COPY_UNTRUSTED_VALUE(&req->result);
+
+    return result;
 }
 
 __attribute_no_sanitize_address


mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1955)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`pal/src/host/linux-sgx/enclave_ocalls.c` line 143 at r1](https://reviewable.io/reviews/gramineproject/gramine/1955#-O2ogg8W0RZFiiwAJCrx:-O2ohame7gKT1pp2JDe8:b-k9fkin) ([raw file](https://github.com/gramineproject/gramine/blob/2ddaae06b6cc6e3388696c138a38f7317e1995a5/pal/src/host/linux-sgx/enclave_ocalls.c#L143)):*

Hmm, have you seen the comment at enclave_framework.c:69? How does it relate to this case?
https://github.com/gramineproject/gramine/blob/master/pal/src/host/linux-sgx/enclave_framework.c#L69-L89
</blockquote></details>

Fixed link:
https://github.com/gramineproject/gramine/blob/ca715b9ca1addd4d846975d5f4a2d1d58804f2b9/pal/src/host/linux-sgx/enclave_framework.c#L69-L89

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1955)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

___
*[`pal/src/host/linux-sgx/enclave_ocalls.c` line 143 at r1](https://reviewable.io/reviews/gramineproject/gramine/1955#-O2ogg8W0RZFiiwAJCrx:-O2xD-Kq1uwr3ZJPfmn9:bxexij1) ([raw file](https://github.com/gramineproject/gramine/blob/2ddaae06b6cc6e3388696c138a38f7317e1995a5/pal/src/host/linux-sgx/enclave_ocalls.c#L143)):*

Fixed link:
https://github.com/gramineproject/gramine/blob/ca715b9ca1addd4d846975d5f4a2d1d58804f2b9/pal/src/host/linux-sgx/enclave_framework.c#L69-L89
</blockquote></details>

Yes, I've seen that comment. If you look at my other PR (https://github.com/gramineproject/gramine/pull/1944), I modify those lines in that PR.

To answer your question, that code indeed relates to this bug fix here. The linked code modifies the Untrusted RSP value (ursp) that is restored during the AEX flow. So the bug is like this (in debug mode, as the linked code is only currently relevant in debug mode, but my PR #1944 changes that):
1. `sgx_exitless_ocall()` is done with its logic and performs `sgx_reset_ustack()`.
2. The `UPDATE_STACK()` macro modifies `ursp` as part of resetting the stack.
3. The enclave thread is interrupted, and the AEX flow is triggered.
4. The AEX flow restores `ursp`, which points to the top of the stack (because of reset on step 2).
5. The [AEX landing pad](https://github.com/gramineproject/gramine/blob/a173a4fb02b25ae9929f2f90c51dc0080fc28781/pal/src/host/linux-sgx/host_entry.S#L66) [pushes stuff](https://github.com/gramineproject/gramine/blob/a173a4fb02b25ae9929f2f90c51dc0080fc28781/pal/src/host/linux-sgx/host_entry.S#L77) on this stack, overwriting the `req` object from step 1.
6. Enclave thread is ERESUME'd and performs the last line in this func: `return COPY_UNTRUSTED_VALUE(&req->result)`. Unfortunately, `req->result` was overwritten in step 5, and the thread reads some garbage.

Hope this clarifies the problem. You can also notice that to trigger this bug, you need to be really unlucky:
a. Needs Exitless feature enabled (in the manifest)
b. Needs Gramine to be built in DEBUG mode
c. For faster reproducibility, needs SGX profiling enabled (in the manifest), otherwise not enough "stack push overwrite" events
d. For faster reproducibility, needs a lot of AEXes happening

Note that our CI tests a) only minimally, and doesn't test c) at all I believe. At the same time, that's a truly rare event, I don't think we want a regression test for this.

kailun-qin (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1955)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1955)* status: all files reviewed, 1 unresolved discussion, not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`pal/src/host/linux-sgx/enclave_ocalls.c` line 143 at r1](https://reviewable.io/reviews/gramineproject/gramine/1955#-O2ogg8W0RZFiiwAJCrx:-O3Tp8yN4RPeGWUGVZe8:btasoya) ([raw file](https://github.com/gramineproject/gramine/blob/2ddaae06b6cc6e3388696c138a38f7317e1995a5/pal/src/host/linux-sgx/enclave_ocalls.c#L143)):*

Yes, I've seen that comment. If you look at my other PR (https://github.com/gramineproject/gramine/pull/1944), I modify those lines in that PR.

To answer your question, that code indeed relates to this bug fix here. The linked code modifies the Untrusted RSP value (ursp) that is restored during the AEX flow. So the bug is like this (in debug mode, as the linked code is only currently relevant in debug mode, but my PR #1944 changes that):
1. `sgx_exitless_ocall()` is done with its logic and performs `sgx_reset_ustack()`.
2. The `UPDATE_STACK()` macro modifies `ursp` as part of resetting the stack.
3. The enclave thread is interrupted, and the AEX flow is triggered.
4. The AEX flow restores `ursp`, which points to the top of the stack (because of reset on step 2).
5. The [AEX landing pad](https://github.com/gramineproject/gramine/blob/a173a4fb02b25ae9929f2f90c51dc0080fc28781/pal/src/host/linux-sgx/host_entry.S#L66) [pushes stuff](https://github.com/gramineproject/gramine/blob/a173a4fb02b25ae9929f2f90c51dc0080fc28781/pal/src/host/linux-sgx/host_entry.S#L77) on this stack, overwriting the `req` object from step 1.
6. Enclave thread is ERESUME'd and performs the last line in this func: `return COPY_UNTRUSTED_VALUE(&req->result)`. Unfortunately, `req->result` was overwritten in step 5, and the thread reads some garbage.

Hope this clarifies the problem. You can also notice that to trigger this bug, you need to be really unlucky:
a. Needs Exitless feature enabled (in the manifest)
b. Needs Gramine to be built in DEBUG mode
c. For faster reproducibility, needs SGX profiling enabled (in the manifest), otherwise not enough "stack push overwrite" events
d. For faster reproducibility, needs a lot of AEXes happening

Note that our CI tests a) only minimally, and doesn't test c) at all I believe. At the same time, that's a truly rare event, I don't think we want a regression test for this.
</blockquote></details>

Yeah, I understand the race here, but I don't understand the comment I linked to. It seems to describe _exactly this scenario_ which you're fixing in this PR, but saying that the codepath in `DEBUG` handles it correctly. If that would be the case, then you wouldn't need a change here, just remove the `ifdef`.

So, I'm not getting what scenario is that comment describing.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1955)* status: all files reviewed, 1 unresolved discussion, not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

___
*[`pal/src/host/linux-sgx/enclave_ocalls.c` line 143 at r1](https://reviewable.io/reviews/gramineproject/gramine/1955#-O2ogg8W0RZFiiwAJCrx:-O3WHw5d2RlK8Jxi7_Ib:b6iqfvz) ([raw file](https://github.com/gramineproject/gramine/blob/2ddaae06b6cc6e3388696c138a38f7317e1995a5/pal/src/host/linux-sgx/enclave_ocalls.c#L143)):*

Yeah, I understand the race here, but I don't understand the comment I linked to. It seems to describe _exactly this scenario_ which you're fixing in this PR, but saying that the codepath in `DEBUG` handles it correctly. If that would be the case, then you wouldn't need a change here, just remove the `ifdef`.

So, I'm not getting what scenario is that comment describing.
</blockquote></details>

No, the linked comment describes a general fix, not the exact (buggy) scenario that I described.

The linked comment explains that there are two actors that use uRSP (untrusted stack of the thread): the AEX event and the OCALL `sgx_alloc_on_ustack()` family of functions.

As a side note, currently in Gramine, only the DEBUG mode cares about the untrusted stack during the AEX events, because currently the RELEASE mode does **not** use the stack (it executes very simple assembly).

So, the linked comment explains that in DEBUG mode, the uRSP value must be modified in OCALL ustack-allocating functions, otherwise AEX events would overwrite the prepared-for-OCALL objects and arguments. That's all that the linked comment does.

What this comment and corresponding code **don't account for** is the case when AEX happens while OCALL objects/arguments are still used. That's just a bug in the code, and the comment talks about a related scenario, but doesn't cover this bug (and it shouldn't cover this bug, because well it's a bug).

Also note that we **can** trigger this bug in DEBUG mode, it's just that we need to enable Exitless, have a lot of AEXes, and need SGX profiling enabled, and run it for several hours in a loop.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1955)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

___
*[`pal/src/host/linux-sgx/enclave_ocalls.c` line 143 at r1](https://reviewable.io/reviews/gramineproject/gramine/1955#-O2ogg8W0RZFiiwAJCrx:-O3XCzPN9JcQvUsbx1Mp:bmye5a5) ([raw file](https://github.com/gramineproject/gramine/blob/2ddaae06b6cc6e3388696c138a38f7317e1995a5/pal/src/host/linux-sgx/enclave_ocalls.c#L143)):*

No, the linked comment describes a general fix, not the exact (buggy) scenario that I described.

The linked comment explains that there are two actors that use uRSP (untrusted stack of the thread): the AEX event and the OCALL `sgx_alloc_on_ustack()` family of functions.

As a side note, currently in Gramine, only the DEBUG mode cares about the untrusted stack during the AEX events, because currently the RELEASE mode does **not** use the stack (it executes very simple assembly).

So, the linked comment explains that in DEBUG mode, the uRSP value must be modified in OCALL ustack-allocating functions, otherwise AEX events would overwrite the prepared-for-OCALL objects and arguments. That's all that the linked comment does.

What this comment and corresponding code **don't account for** is the case when AEX happens while OCALL objects/arguments are still used. That's just a bug in the code, and the comment talks about a related scenario, but doesn't cover this bug (and it shouldn't cover this bug, because well it's a bug).

Also note that we **can** trigger this bug in DEBUG mode, it's just that we need to enable Exitless, have a lot of AEXes, and need SGX profiling enabled, and run it for several hours in a loop.
</blockquote></details>

Ok, I see, thanks!

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1955)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

kailun-qin (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1955)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #1183, opened 2023-02-15
`meson devenv`, part 1

woju:
## Description of the changes 

This is part 1 of `meson devenv`, which allows to run Gramine without installing (just from builddir). It's a QoL improvement for people who don't have development setup and don't want to install and fiddle with `PATH` and `PYTHONPATH` (`meson devenv` can be understood as a machinery that does it for them).

Note this is for developers of gramine proper, not developers of apps that would run inside gramine. People who are not in business of modifying gramine should still install gramine from packages or at most compile and install, as usual.

Next on the horizon is building documentation for `sgx_sign.py` (which requires module with offsets, which is compiled, so docs can't be extracted without building).

## How to test this PR? 

There are docs, plz follow them. Test `gramine-direct`, `gramine-sgx`, both without and with `GDB=1`. Run some `CI-Examples/`.

PR #1988, opened 2024-09-02
[Scripts] Add activate.sh

woju:
## Description of the changes 

This is mostly QoL feature for Gramine developers. If you setup the build with nonstandard `--prefix`, like you do for development, it's quite a chore to set all environment variables for Gramine to work. This script does this automatically.

The script is never installed, but is present in meson's build directory. After you install gramine, you can just run this script to get a valid environment. 

Until we can do something about devenv (which would not need even installing; cf. #1183), this should be a reasonably convenient way to reliably get a testing environment.

## How to test this PR? 

```sh
meson setup build/ --prefix ... -Dsgx=enabled -Ddirect=enabled <other options>
meson compile -C build/
meson install -C build/
build/activate.sh
cd CI-Examples/helloworld
make SGX=1
gramine-sgx helloworld # gramine-sgx should be in $PATH after activate.sh
```

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -311,3 +311,14 @@ if enable_libgomp
 endif
 
 run_target('clang-format', command: [meson_clang_format_prog])
+
+conf = configuration_data()
+conf.set('PATH', get_option('prefix') / get_option('bindir'))
+conf.set('LD_LIBRARY_PATH', get_option('prefix') / get_option('libdir'))
+conf.set('PKG_CONFIG_PATH', get_option('prefix') / get_option('libdir') / 'pkgconfig')
+conf.set('PYTHONPATH', python3_platlib)
+configure_file(
+    input: 'scripts/activate.sh.in',
+    output: 'activate.sh',
+    configuration: conf,
+)
--- a/scripts/activate.sh.in
+++ b/scripts/activate.sh.in
@@ -1,0 +1,24 @@
+#!/bin/sh
+
+set -xe
+
+case "@PATH@" in
+@*)
+    echo this script is intended for running from builddir >&2
+    exit 1
+    ;;
+esac
+
+pathmunge() {
+    eval "var=\$$1"
+    case "$var" in *"$2"*) return;; esac
+    eval "$1=\$2\${$1:+:\$$1}"
+    eval "export $1"
+}
+
+pathmunge PATH "@PATH@"
+pathmunge LD_LIBRARY_PATH "@LD_LIBRARY_PATH@"
+pathmunge PKG_CONFIG_PATH "@PKG_CONFIG_PATH@"
+pathmunge PYTHONPATH "@PYTHONPATH@"
+
+exec $SHELL


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1988)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "WIP" found in commit messages' one-liners (waiting on @woju)

___
*[`meson.build` line 317 at r1](https://reviewable.io/reviews/gramineproject/gramine/1988#-O5mkNzn8YRd9qsUzJ_G:-O5mkNzn8YRd9qsUzJ_H:ba20dlm) ([raw file](https://github.com/gramineproject/gramine/blob/2f4644a6b5ef989aa92e007491333482cccb6506/meson.build#L317)):*
> ```Meson
> conf = configuration_data()
> conf.set('PATH', get_option('prefix') / get_option('bindir'))
> conf.set('LD_LIBRARY_PATH', get_option('prefix') / get_option('libdir'))
> ```

Why is `LD_LIBRARY_PATH` needed? I never set it in my environments, and is this even a correct set up?
___
*[`scripts/activate.sh.in` line 6 at r1](https://reviewable.io/reviews/gramineproject/gramine/1988#-O5mlYWGCOaoliD-_MLT:-O5mlYWGCOaoliD-_MLU:bvrrbe0) ([raw file](https://github.com/gramineproject/gramine/blob/2f4644a6b5ef989aa92e007491333482cccb6506/scripts/activate.sh.in#L6)):*
> ```
> 
> case "@PATH@" in
> @*)
> ```

I don't understand this case. How does it work, what does it protect against?
___
*[`scripts/activate.sh.in` line 20 at r1](https://reviewable.io/reviews/gramineproject/gramine/1988#-O5mlcHBAN_lTZ8_sEx2:-O5mlcHBAN_lTZ8_sEx3:b-l81ltl) ([raw file](https://github.com/gramineproject/gramine/blob/2f4644a6b5ef989aa92e007491333482cccb6506/scripts/activate.sh.in#L20)):*
> ```
> 
> pathmunge PATH "@PATH@"
> pathmunge LD_LIBRARY_PATH "@LD_LIBRARY_PATH@"
> ```

ditto (really needed?)

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1988)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "WIP" found in commit messages' one-liners (waiting on @dimakuv and @woju)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/1988#-O5mq0v73ZDDXYJs8Ivl):*
Blocking until non-WIP.
@woju: Shouldn't this be a draft PR, if it's WIP?



_Code quote:_
```
WIP 
```
___
*[`scripts/activate.sh.in` line 7 at r1](https://reviewable.io/reviews/gramineproject/gramine/1988#-O5mqQvc9poYK2jVlrf-:-O5mqQvc9poYK2jVlrf0:bbwikmc) ([raw file](https://github.com/gramineproject/gramine/blob/2f4644a6b5ef989aa92e007491333482cccb6506/scripts/activate.sh.in#L7)):*
> ```
> case "@PATH@" in
> @*)
>     echo this script is intended for running from builddir >&2
> ```

(not sure about the exact wording, but I'd mention that it's not only about location, but also processing by Meson, otherwise people may interpret this as "I'll just copy the script over there and it will work")



_Suggestion:_
```
This script is intended for running from builddir after being templated by Meson
```
___
*[`scripts/activate.sh.in` line 15 at r1](https://reviewable.io/reviews/gramineproject/gramine/1988#-O5mrVF03hB-wT4va-Se:-O5mrVF14KprlXS7-hMk:b-yz5n09) ([raw file](https://github.com/gramineproject/gramine/blob/2f4644a6b5ef989aa92e007491333482cccb6506/scripts/activate.sh.in#L15)):*
> ```
>     eval "var=\$$1"
>     case "$var" in *"$2"*) return;; esac
>     eval "$1=\$2\${$1:+:\$$1}"
> ```

Could you explain why we need this magic instead of just something like `export X="@X@"`?

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1988)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), "WIP" found in commit messages' one-liners (waiting on @woju)

___
*[`scripts/activate.sh.in` line 6 at r1](https://reviewable.io/reviews/gramineproject/gramine/1988#-O5mlYWGCOaoliD-_MLT:-O5n0lTDBxcasnHLkZZH:b-xivica) ([raw file](https://github.com/gramineproject/gramine/blob/2f4644a6b5ef989aa92e007491333482cccb6506/scripts/activate.sh.in#L6)):*

It protects against running `./scripts/activate.sh.in` vs `build/activate.sh`. `activate.sh.in` needs to be executable, because `configure_file()` will preserve file permissions.
</blockquote></details>

A bit weird use of `case` (why not just an `if` check?), but ok

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1988)* status: 1 of 2 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @mkow)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/1988#-O5mq0v73ZDDXYJs8Ivl):*

Blocking until non-WIP.
@woju: Shouldn't this be a draft PR, if it's WIP?
</blockquote></details>

Done.
___
*[`scripts/activate.sh.in` line 6 at r1](https://reviewable.io/reviews/gramineproject/gramine/1988#-O5mlYWGCOaoliD-_MLT:-O5n6qjX9UDBiUMcnpoW:bqunsl8) ([raw file](https://github.com/gramineproject/gramine/blob/2f4644a6b5ef989aa92e007491333482cccb6506/scripts/activate.sh.in#L6)):*

A bit weird use of `case` (why not just an `if` check?), but ok
</blockquote></details>

`sh`'s `if [`/`if test` does not support `*` matching, because `[`/`test` is a command (in contrast to bash's non-portable `if [[`, which is a syntax). This is the portable variant of `if [[ "@PATH@" = @* ]]`.

Why I can't do this with `if`: because I can't write `if "@PATH@" = @PATH@` -- that is always true whether meson already did @-formatting or not (for the formatted case both sides are substituted). I need something that I can put on the right side which won't be substituted. Maybe I could do `if test "@PATH@" = "@"PATH@`, and then answer a bunch of questions about POSIX shell quoting and wordsplitting rules, but I'm not even sure if meson wouldn't barf for invalid `"PATH` substitution, I haven't checked.
___
*[`scripts/activate.sh.in` line 7 at r1](https://reviewable.io/reviews/gramineproject/gramine/1988#-O5mqQvc9poYK2jVlrf-:-O5n9zBA-B7hxt0jtI_M:bvy711z) ([raw file](https://github.com/gramineproject/gramine/blob/2f4644a6b5ef989aa92e007491333482cccb6506/scripts/activate.sh.in#L7)):*
> "I'll just copy the script over there and it will work"

Yeah, good point. Done.
___
*[`scripts/activate.sh.in` line 15 at r1](https://reviewable.io/reviews/gramineproject/gramine/1988#-O5mrVF03hB-wT4va-Se:-O5n4VEGDtekpukAA-kV:b41uioa) ([raw file](https://github.com/gramineproject/gramine/blob/2f4644a6b5ef989aa92e007491333482cccb6506/scripts/activate.sh.in#L15)):*

Could you explain why we need this magic instead of just something like `export X="@X@"`?
</blockquote></details>

Sure. Generally, all this is needed, because sh is hard if you care about correctness in edge cases. In particular:

Suppose you already has something in `LD_LIBRARY_PATH`, then you don't want to have this overwritten (so you can't just `X="@X@"`). `${:+}` substitution is needed for variables that are not yet set, which is most of the time, so you don't have an extra colon.

`case` is needed in two cases: for when you accidentally run this script twice, so you won't change the variable twice, and sometimes not all variables need to be set, for example with `--prefix=/usr/local` (which is the default) `PYTHONPATH` doesn't need to be set, and get-python-platlib script returns empty, so `case "$var" in *""*)` is true and this function is noop.

That this whole thing is function is needed, because you need `return` to work. I guess I could do this in `for` with `continue` instead, but will be less readable, because I'd need to set temporary variables with the substitutions from meson and then process them in the loop.

`eval` is needed, because the name of the variable is an argument to this function (or would be a loop parameter in `for`+`continue` variant — no difference here).

I wanted three `eval`s and separate `case`instead of one `eval` with `case` inside, to not have to think about quoting `"`-s inside `case` statement. Those would be tricky to get right (there are already too much tricky things in this function that need to be gotten exactly right).

Did I miss any magic that still needs explanation?

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1988)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

___
*[`scripts/activate.sh.in` line 15 at r1](https://reviewable.io/reviews/gramineproject/gramine/1988#-O5mrVF03hB-wT4va-Se:-O5nJJUz53K3rSq0magi:b-985ve4) ([raw file](https://github.com/gramineproject/gramine/blob/2f4644a6b5ef989aa92e007491333482cccb6506/scripts/activate.sh.in#L15)):*
> Did I miss any magic that still needs explanation?

No, but you reminded me why I do all such wrappers in Python ;)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1988)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #1995, opened 2024-09-04
[PAL/Linux-SGX] Remove printing of thread-local SGX statistics

dimakuv:
## Description of the changes 

Previously, when manifest contained `sgx.enable_stats = true`, Gramine-SGX printed not the process-wide SGX statistics, but also statistics for each thread. These thread-local SGX statistics only litter the output as (1) there are helper Gramine threads that do not correspond to any application thread, (2) there is no way to correlate the printed thread-local stats to application threads/workload execution (as stats simply print the opaque TID of the thread).

This PR removes thread-local stats and keeps only process-wide ones.

## How to test this PR? 

The Helloworld CI example (with `sgx.enable_stats = true`):

- With this PR:
```
gramineproject/gramine/CI-Examples/helloworld$ gramine-sgx helloworld
Gramine is starting. Parsing TOML manifest file, this may take some time...
----- SGX enclave loading time =     412964 microseconds -----
-----------------------------------------------------------------------------------------------------------------------
Gramine detected the following insecure configurations:

  - sgx.debug = true                           (this is a debug enclave)

Gramine will continue application execution, but this configuration must not be used in production!
-----------------------------------------------------------------------------------------------------------------------

Hello, world
----- Total SGX stats for process 273134 -----
  # of EENTERs:        317
  # of EEXITs:         241
  # of AEXs:           2175
  # of sync signals:   72
  # of async signals:  0
```

- Without this PR:
```
gramineproject/gramine/CI-Examples/helloworld$ gramine-sgx helloworld
Gramine is starting. Parsing TOML manifest file, this may take some time...
----- SGX enclave loading time =     410795 microseconds -----
-----------------------------------------------------------------------------------------------------------------------
Gramine detected the following insecure configurations:

  - sgx.debug = true                           (this is a debug enclave)

Gramine will continue application execution, but this configuration must not be used in production!
-----------------------------------------------------------------------------------------------------------------------

----- SGX stats for thread 323750 -----
  # of EENTERs:        15
  # of EEXITs:         14
  # of AEXs:           19
  # of sync signals:   0
  # of async signals:  0
----- SGX stats for thread 323751 -----                                                                               [0/1860]
  # of EENTERs:        15
  # of EEXITs:         14
  # of AEXs:           0
  # of sync signals:   0
  # of async signals:  0
----- SGX stats for thread 323753 -----
  # of EENTERs:        15
  # of EEXITs:         14
  # of AEXs:           0
  # of sync signals:   0
  # of async signals:  0
Hello, world
[...]

PR #1996, opened 2024-09-04
[PAL/Linux-SGX] Print SGX stats on SIGUSR1 and reset them

### LINKED ISSUE #1711: [PAL/Linux-SGX] Allow to dump current SGX/perf stats on a signal

Currently, Gramine-SGX has two perf analysis tools:
1. Trivial stats on SGX events (EENTER, EEXIT, etc.)
2. Advanced stats similar to `perf record`

Both these tools **have a limitation**: they start collecting stats when Gramine-SGX starts and end collecting stats when Gramine-SGX terminates.

This limits the ability to analyze performance of long-living applications. For example, if MySQL runs under Gramine-SGX, then we may want to analyze only the stats during "hot runs", when a particular client with a particular workload connects to the MySQL server. But because of the current limitation, we will have a lot of noise because stats also contain the startup events, the termination events, and other non-relevant events (like clients that pre-populate the database).

### Proposal 1: dump stats on a signal

We choose a signal that serves as a hint to Gramine to dump the currently collected statistics, e.g. `SIGUSR1`. For simplicity, we block this signal on all threads of the process bar the main thread (so `SIGUSR1` is guaranteed to always lend in Thread 1).

When the signal arrives, we dump SGX stats similar to this:
https://github.com/gramineproject/gramine/blob/1f72aaf8148a7712c5c8bad1da9839e2bcc5c633/pal/src/host/linux-sgx/host_ocalls.c#L33-L44

**NOTE**: We will need to dump stats on all currently executing threads, and this will require finding a way to iterate through all threads and summing up their stats. Should be doable.

The output can be like this:
```c
   static uint32_t g_user_signal_number = 0;
   log_always("----- SGX stats for process %d (on user signal %u) -----\n"
                   "  # of EENTERs:        %lu\n", ... g_user_signal_number++);
   ...
```

Now the MySQL example can be done like this:
1. Start MySQL in Gramine
2. Do/wait for the initialization to finish
3. Right-before starting the workload client, send SIGUSR1 to Gramine
4. Gramine dumps the current stats
5. Right-after ending the workload client, send SIGUSR1 to Gramine
6. Gramine dumps the current stats
7. Finish the run

Now we have two sets of stats, collected at steps 4 and 6. We subtract 4-stats from 6-stats (we can subtract only process-wide stats), and we get the statistic on SGX events during the client workload -- exactly what we wanted.

### Proposal 2: reset stats on a signal

Same as Proposal 1, but **set all stats to zero**. This will be easier for end user to read, but Proposal 1 (which requires the "differential" analysis of stats) seems more flexible and easier to implement.

I am in favor of Proposal 1.

### What about `perf record` stats?

Perf record style (advanced) stats are much more complicated, see:
- https://github.com/gramineproject/gramine/blob/master/pal/src/host/linux-sgx/host_perf_data.c
- https://github.com/gramineproject/gramine/blob/master/pal/src/host/linux-sgx/host_profile.c

Here the problem is that we create a `perf.data` file, initialize it with some header, and add events one by one into it. So it's unclear what we can do when a SIGUSR1 signal arrives -- can we seal the current file and start a new file? This adheres to Proposal 2. I don't know how to make it work with Proposal 1...

Someone needs to learn how this can be achieved -- `perf record` surely allows such things, so it must be accounted for in perf internal formats.

One can start with the simpler SGX stats though, and leave the `perf record` stats for later implementation.

Useful for perf analysis.

dimakuv:
## Description of the changes 

This commit adds support to dump and reset SGX-related statistics interactively, using SIGUSR1 signal. This helps to collect SGX-related statistics only for a particular period, e.g. skipping the Gramine startup and application initialization time and concentrating only on the actual application processing.

The printed-out statistics are not precise, as the "stats collecting" thread may run in parallel with other threads that update the statistics. However, this imprecise implementation is simple and enough for perf analysis.

This commit also breaks compatibility: SGX statistics can only be collected and printed when Gramine is built in debug or debugoptimized mode. However, this should not affect users as SGX stats is tailored for manual debugging and profiling sessions.

Applied on top of #1995 

Closes #1857

Fixes #1711 

## How to test this PR? 

Check instructions in https://github.com/gramineproject/gramine/pull/1857

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1996)

### DIFF
--- a/Documentation/manifest-syntax.rst
+++ b/Documentation/manifest-syntax.rst
@@ -1293,6 +1293,10 @@ This syntax specifies whether to enable SGX enclave-specific statistics:
    includes creating the enclave, adding enclave pages, measuring them and
    initializing the enclave.
 
+For this option to take effect, Gramine must be compiled with
+``--buildtype=debug`` or ``--buildtype=debugoptimized``. Otherwise (if built in
+release mode), Gramine will exit with an error.
+
 .. warning::
    This option is insecure and cannot be used with production enclaves
    (``sgx.debug = false``). If a production enclave is started with this option
--- a/Documentation/performance.rst
+++ b/Documentation/performance.rst
@@ -18,11 +18,13 @@ Enabling SGX enclave stats
 
 See also :ref:`perf` below for installing ``perf``.
 
-Enable statistics using ``sgx.enable_stats = true`` manifest option. Now your
-graminized application correctly reports performance counters. This is useful
-when using e.g. ``perf stat`` to collect performance statistics. This manifest
-option also forces Gramine to dump SGX-related information on each enclave exit.
-Here is an example:
+Enable statistics using ``sgx.enable_stats = true`` manifest option (note that
+Gramine must be compiled with ``--buildtype=debug`` or
+``--buildtype=debugoptimized`` for this option to work). Now your graminized
+application correctly reports performance counters. This is useful when using
+e.g. ``perf stat`` to collect performance statistics. This manifest option also
+forces Gramine to dump SGX-related information on each enclave exit. Here is an
+example:
 
 ::
 
@@ -89,6 +91,12 @@ How to read this output:
    itself, this particular output is not interesting. In reality, performance
    counters should be compared against "golden runs" to deduce any interesting
    trends.
+
+It is also possible to dump and reset SGX-related statistics interactively, using
+``SIGUSR1`` signal. This helps to collect SGX-related statistics only for a
+particular period, e.g. skipping the Gramine startup and application
+initialization time and concentrating only on the actual application processing.
+Send ``SIGUSR1`` using command ``kill -SIGUSR1 -<PGID>``.
 
 Effects of system calls / ocalls
 --------------------------------
--- a/libos/test/regression/fork_and_access_file.manifest.template
+++ b/libos/test/regression/fork_and_access_file.manifest.template
@@ -11,6 +11,9 @@ sgx.max_threads = {{ '1' if env.get('EDMM', '0') == '1' else '16' }}
 sgx.debug = true
 sgx.edmm_enable = {{ 'true' if env.get('EDMM', '0') == '1' else 'false' }}
 
+# this is only to test that `sgx.enable_stats` works (it can only be specified for debug-mode tests)
+sgx.enable_stats = true
+
 sgx.trusted_files = [
   "file:{{ gramine.runtimedir(libc) }}/",
   "file:{{ binary_dir }}/{{ entrypoint }}",
--- a/libos/test/regression/multi_pthread.manifest.template
+++ b/libos/test/regression/multi_pthread.manifest.template
@@ -12,7 +12,6 @@ sgx.max_threads = {{ '1' if env.get('EDMM', '0') == '1' else '8' }}
 
 sgx.debug = true
 sgx.edmm_enable = {{ 'true' if env.get('EDMM', '0') == '1' else 'false' }}
-sgx.enable_stats = true
 
 sgx.trusted_files = [
   "file:{{ gramine.runtimedir(libc) }}/",
--- a/libos/test/regression/multi_pthread_exitless.manifest.template
+++ b/libos/test/regression/multi_pthread_exitless.manifest.template
@@ -13,7 +13,6 @@ sgx.insecure__rpc_thread_num = 8
 
 sgx.debug = true
 sgx.edmm_enable = {{ 'true' if env.get('EDMM', '0') == '1' else 'false' }}
-sgx.enable_stats = true
 
 sgx.trusted_files = [
   "file:{{ gramine.runtimedir(libc) }}/",
--- a/pal/regression/Thread2.manifest.template
+++ b/pal/regression/Thread2.manifest.template
@@ -1,6 +1,5 @@
 loader.entrypoint.uri = "file:{{ binary_dir }}/{{ entrypoint }}"
 
 sgx.max_threads = {{ '1' if env.get('EDMM', '0') == '1' else '2' }}
-sgx.enable_stats = true
 sgx.debug = true
 sgx.edmm_enable = {{ 'true' if env.get('EDMM', '0') == '1' else 'false' }}
--- a/pal/regression/Thread2_exitless.manifest.template
+++ b/pal/regression/Thread2_exitless.manifest.template
@@ -4,6 +4,5 @@ loader.entrypoint.uri = "file:{{ binary_dir }}/{{ entrypoint }}"
 
 sgx.max_threads = {{ '1' if env.get('EDMM', '0') == '1' else '2' }}
 sgx.insecure__rpc_thread_num = 2
-sgx.enable_stats = true
 sgx.debug = true
 sgx.edmm_enable = {{ 'true' if env.get('EDMM', '0') == '1' else 'false' }}
--- a/pal/src/host/linux-sgx/host_entry.S
+++ b/pal/src/host/linux-sgx/host_entry.S
@@ -16,6 +16,7 @@
 
     .extern tcs_base
     .extern g_in_aex_profiling
+    .extern maybe_dump_and_reset_stats
 
     .global sgx_ecall
     .type sgx_ecall, @function
@@ -102,6 +103,8 @@ async_exit_pointer:
     movq %rbx, %rdi
     call sgx_profile_sample_aex
 
+    call maybe_dump_and_reset_stats
+
     # Restore stack
     movq %rbp, %rsp
     .cfi_def_cfa_register %rsp
@@ -168,6 +171,8 @@ sgx_raise:
     # Call sgx_profile_sample_ocall_inner with RDX (pointer to in-enclave context)
     movq %rdx, %rdi
     call sgx_profile_sample_ocall_inner
+
+    call maybe_dump_and_reset_stats
 
     # Restore RDI
     movq -8(%rbp), %rdi
--- a/pal/src/host/linux-sgx/host_exception.c
+++ b/pal/src/host/linux-sgx/host_exception.c
@@ -17,6 +17,7 @@
 #include "debug_map.h"
 #include "host_internal.h"
 #include "pal_rpc_queue.h"
+#include "pal_tcb.h"
 #include "sigreturn.h"
 #include "sigset.h"
 #include "ucontext.h"
@@ -188,6 +189,11 @@ static void handle_sigusr1(int signum, siginfo_t* info, struct ucontext* uc) {
     __UNUSED(info);
     __UNUSED(uc);
 
+    if (g_sgx_enable_stats) {
+        PAL_HOST_TCB* tcb = pal_get_host_tcb();
+        __atomic_store_n(&tcb->reset_stats, true, __ATOMIC_RELAXED);
+    }
+
     if (g_pal_enclave.profile_enable) {
         __atomic_store_n(&g_trigger_profile_reinit, true, __ATOMIC_RELEASE);
     }
@@ -268,3 +274,16 @@ void pal_describe_location(uintptr_t addr, char* buf, size_t buf_size) {
 #endif
     default_describe_location(addr, buf, buf_size);
 }
+
+#ifdef DEBUG
+/* called on each AEX and OCALL (in normal context), see host_entry.S */
+void maybe_dump_and_reset_stats(void) {
+    if (!g_sgx_enable_stats)
+        return;
+
+    PAL_HOST_TCB* tcb = pal_get_host_tcb();
+    if (__atomic_exchange_n(&tcb->reset_stats, false, __ATOMIC_RELAXED) == true) {
+        collect_and_print_sgx_stats();
+    }
+}
+#endif
--- a/pal/src/host/linux-sgx/host_internal.h
+++ b/pal/src/host/linux-sgx/host_internal.h
@@ -122,7 +122,6 @@ int clone_thread(void* dynamic_tcs);
 
 int create_tcs_mapper(void* tcs_base, unsigned int thread_num);
 int pal_thread_init(void* tcbptr);
-void map_tcs(unsigned int tid);
 void unmap_my_tcs(void);
 int current_enclave_thread_cnt(void);
 void thread_exit(int status);
--- a/pal/src/host/linux-sgx/host_main.c
+++ b/pal/src/host/linux-sgx/host_main.c
@@ -753,6 +753,14 @@ static int parse_loader_config(char* manifest, struct pal_enclave* enclave_info,
         goto out;
     }
 
+#ifndef DEBUG
+    if (g_sgx_enable_stats) {
+        log_error("'sgx.enable_stats = true' is specified in non-debug mode, this is disallowed");
+        ret = -EINVAL;
+        goto out;
+    }
+#endif
+
     ret = toml_string_in(manifest_root, "sgx.sigfile", &dummy_sigfile_str);
     if (ret < 0 || dummy_sigfile_str) {
         log_error("sgx.sigfile is not supported anymore. Please update your manifest according to "
--- a/pal/src/host/linux-sgx/host_ocalls.c
+++ b/pal/src/host/linux-sgx/host_ocalls.c
@@ -31,9 +31,8 @@
 rpc_queue_t* g_rpc_queue = NULL; /* pointer to untrusted queue */
 
 static noreturn void process_exit(int exitcode) {
-    update_sgx_stats(/*do_print=*/true);
-
 #ifdef DEBUG
+    update_sgx_stats_on_exit(/*do_print=*/true);
     sgx_profile_finish();
 #endif
 
@@ -73,7 +72,9 @@ static long sgx_ocall_exit(void* args) {
         process_exit((int)ocall_exit_args->exitcode);
     }
 
-    update_sgx_stats(/*do_print=*/false);
+#ifdef DEBUG
+    update_sgx_stats_on_exit(/*do_print=*/false);
+#endif
     thread_exit((int)ocall_exit_args->exitcode);
     return 0;
 }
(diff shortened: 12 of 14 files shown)


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1996)* status: 0 of 14 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

___
*[`pal/src/host/linux-sgx/host_exception.c` line 194 at r1](https://reviewable.io/reviews/gramineproject/gramine/1996#-O5woH8n6hpu8YpIBglj:-O5woH8n6hpu8YpIBglk:b-gf6k26) ([raw file](https://github.com/gramineproject/gramine/blob/0e1e3a74db98122ff24049b6ce23b01e29c7f675/pal/src/host/linux-sgx/host_exception.c#L194)):*
> ```C
>     if (g_sgx_enable_stats) {
>         PAL_HOST_TCB* tcb = pal_get_host_tcb();
>         __atomic_store_n(&tcb->reset_stats, true, __ATOMIC_RELAXED);
> ```

It is possible that two SIGUSR1 signals arrive one after another (or even on different threads). This is benign because the whole `collect_and_print_sgx_stats()` function is protected with a global lock.
___
*[`pal/src/host/linux-sgx/host_thread.c` line 19 at r1](https://reviewable.io/reviews/gramineproject/gramine/1996#-O5woUGI0TAcSfURiexU:-O5woUGI0TAcSfURiexV:bt0pxov) ([raw file](https://github.com/gramineproject/gramine/blob/0e1e3a74db98122ff24049b6ce23b01e29c7f675/pal/src/host/linux-sgx/host_thread.c#L19)):*
> ```C
>     unsigned int    tid;
>     sgx_arch_tcs_t* tcs;
>     PAL_HOST_TCB*   tcb;
> ```

This is a bit controversial: a "SGX-stats-collecting" thread (that received SIGUSR1) is able to peek into the thread local storage of other threads. This is typically considered a bad smell (one thread reads local data of another thread), but here I think it's a reasonable trade-off: most of the accesses to the thread-local SGX counters are done on EEXIT and AEX events by the thread itself, so the memory access should be as simple as possible and as fast as possible. If we would move everything in a shared array, then we would have false cache sharing and complex memory management of the shared array.

mkow (changes requested):
> Reviewed 14 of 14 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1996)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`-- commits` line 15 at r1](https://reviewable.io/reviews/gramineproject/gramine/1996#-O6vEXop4moN_GQ0x3NY):*




_Suggestion:_
```
SGX statistics can now only be
```
___
*[`pal/src/host/linux-sgx/host_exception.c` line 285 at r1](https://reviewable.io/reviews/gramineproject/gramine/1996#-O6zvkRv54t55HEnC2mu:-O6zvkRv54t55HEnC2mv:b-3m7fgd) ([raw file](https://github.com/gramineproject/gramine/blob/0e1e3a74db98122ff24049b6ce23b01e29c7f675/pal/src/host/linux-sgx/host_exception.c#L285)):*
> ```C
> 
>     PAL_HOST_TCB* tcb = pal_get_host_tcb();
>     if (__atomic_load_n(&tcb->reset_stats, __ATOMIC_RELAXED) == false)
> ```

Maybe just atomic exchange with `false`? Would be one line less and effectively work the same.
___
*[`pal/src/host/linux-sgx/host_thread.c` line 19 at r1](https://reviewable.io/reviews/gramineproject/gramine/1996#-O5woUGI0TAcSfURiexU:-O6zwvKe6ZR0yZXxDJty:b6x3o70) ([raw file](https://github.com/gramineproject/gramine/blob/0e1e3a74db98122ff24049b6ce23b01e29c7f675/pal/src/host/linux-sgx/host_thread.c#L19)):*

This is a bit controversial: a "SGX-stats-collecting" thread (that received SIGUSR1) is able to peek into the thread local storage of other threads. This is typically considered a bad smell (one thread reads local data of another thread), but here I think it's a reasonable trade-off: most of the accesses to the thread-local SGX counters are done on EEXIT and AEX events by the thread itself, so the memory access should be as simple as possible and as fast as possible. If we would move everything in a shared array, then we would have false cache sharing and complex memory management of the shared array.
</blockquote></details>

Could you put that in a comment here? Or better, to the loop iterating over TCBs?
___
*[`pal/src/host/linux-sgx/host_thread.c` line 75 at r1](https://reviewable.io/reviews/gramineproject/gramine/1996#-O6zyZyO-OZ6DHaJAc0W:-O6zyZyO-OZ6DHaJAc0X:b-z0a5km) ([raw file](https://github.com/gramineproject/gramine/blob/0e1e3a74db98122ff24049b6ce23b01e29c7f675/pal/src/host/linux-sgx/host_thread.c#L75)):*
> ```C
>     /* there is a small window when the thread's counters may be updated in-between reading and
>      * resetting these counters -- some SGX events will be lost; we ignore this as the number of
>      * lost events is negligible for perf analysis purposes */
> ```

Why not atomic exchange with 0 then?
___
*[`pal/src/host/linux-sgx/host_thread.c` line 84 at r1](https://reviewable.io/reviews/gramineproject/gramine/1996#-O6zrgug3TCh1_KIJqhA:-O6zrgug3TCh1_KIJqhB:b-w17gmh) ([raw file](https://github.com/gramineproject/gramine/blob/0e1e3a74db98122ff24049b6ce23b01e29c7f675/pal/src/host/linux-sgx/host_thread.c#L84)):*
> ```C
> }
> 
> /* this function is called only on thread/process exit (never in the middle of thread exec) */
> ```

Then I'd add a suffix `on_thread_exit` to it (if it's important to its semantics).

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1996)* status: 10 of 14 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[`-- commits` line 15 at r1](https://reviewable.io/reviews/gramineproject/gramine/1996#-O6vEXop4moN_GQ0x3NY):*
Will do during final rebase
___
*[`pal/src/host/linux-sgx/host_exception.c` line 285 at r1](https://reviewable.io/reviews/gramineproject/gramine/1996#-O6zvkRv54t55HEnC2mu:-O73k8it56TJVUL_AAYU:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/0e1e3a74db98122ff24049b6ce23b01e29c7f675/pal/src/host/linux-sgx/host_exception.c#L285)):*

Maybe just atomic exchange with `false`? Would be one line less and effectively work the same.
</blockquote></details>

Done.
___
*[`pal/src/host/linux-sgx/host_thread.c` line 19 at r1](https://reviewable.io/reviews/gramineproject/gramine/1996#-O5woUGI0TAcSfURiexU:-O73lIaS-8T2XZ6vHQOv:b-nnvmd2) ([raw file](https://github.com/gramineproject/gramine/blob/0e1e3a74db98122ff24049b6ce23b01e29c7f675/pal/src/host/linux-sgx/host_thread.c#L19)):*

Could you put that in a comment here? Or better, to the loop iterating over TCBs?
</blockquote></details>

Done. Put in the loop.
___
*[`pal/src/host/linux-sgx/host_thread.c` line 75 at r1](https://reviewable.io/reviews/gramineproject/gramine/1996#-O6zyZyO-OZ6DHaJAc0W:-O73mCXG9qOxox2v9WSl:bz1wmeo) ([raw file](https://github.com/gramineproject/gramine/blob/0e1e3a74db98122ff24049b6ce23b01e29c7f675/pal/src/host/linux-sgx/host_thread.c#L75)):*

Why not atomic exchange with 0 then?
</blockquote></details>

Done. True, I was being lazy.
___
*[`pal/src/host/linux-sgx/host_thread.c` line 84 at r1](https://reviewable.io/reviews/gramineproject/gramine/1996#-O6zrgug3TCh1_KIJqhA:-O73lbC53clGGkSaiXXd:bap2yby) ([raw file](https://github.com/gramineproject/gramine/blob/0e1e3a74db98122ff24049b6ce23b01e29c7f675/pal/src/host/linux-sgx/host_thread.c#L84)):*

Then I'd add a suffix `on_thread_exit` to it (if it's important to its semantics).
</blockquote></details>

Done. It's not only thread exit but also process exit, so I decided to add `_on_exit`.

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1996)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[`pal/src/host/linux-sgx/host_thread.c` line 75 at r1](https://reviewable.io/reviews/gramineproject/gramine/1996#-O6zyZyO-OZ6DHaJAc0W:-O74LDbf7oc1JDQhUt4K:b-mzdr4e) ([raw file](https://github.com/gramineproject/gramine/blob/0e1e3a74db98122ff24049b6ce23b01e29c7f675/pal/src/host/linux-sgx/host_thread.c#L75)):*

Done. True, I was being lazy.
</blockquote></details>

It's even shorter now :)

kailun-qin (approved):
> Reviewed 9 of 14 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1996)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

kailun-qin (approved):
> Reviewed 1 of 14 files at r1.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1996)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1996)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1996)* status: 13 of 14 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

___
*[`pal/src/host/linux-sgx/host_thread.c` line 73 at r4](https://reviewable.io/reviews/gramineproject/gramine/1996#-O7_QBnm59NdvrcQRcgk:-O7_QBnm59NdvrcQRcgl:b-kg01w5) ([raw file](https://github.com/gramineproject/gramine/blob/6b5a59f4137403de942190ea291dd4872cc6b39b/pal/src/host/linux-sgx/host_thread.c#L73)):*
> ```C
>                                               __ATOMIC_RELAXED);
>     g_async_signal_cnt += __atomic_exchange_n((uint64_t*)&tcb->async_signal_cnt, 0,
>                                               __ATOMIC_RELAXED);
> ```

Not happy about the explicit cast, but otherwise Clang complains (GCC doesn't):
```
  ../pal/src/host/linux-sgx/host_thread.c:67:27: error: address argument to atomic operation must be a pointer to integer or pointer ('atomic_ulong *' (aka '_Atomic(unsigned long) *') invalid)
      g_eenter_cnt       += __atomic_exchange_n(&tcb->eenter_cnt, 0, __ATOMIC_RELAXED);
```

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1996)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`pal/src/host/linux-sgx/host_thread.c` line 73 at r4](https://reviewable.io/reviews/gramineproject/gramine/1996#-O7_QBnm59NdvrcQRcgk:-O7_wpD-1lg2iJJh4Be6:b-ghg9xt) ([raw file](https://github.com/gramineproject/gramine/blob/6b5a59f4137403de942190ea291dd4872cc6b39b/pal/src/host/linux-sgx/host_thread.c#L73)):*

Not happy about the explicit cast, but otherwise Clang complains (GCC doesn't):
```
  ../pal/src/host/linux-sgx/host_thread.c:67:27: error: address argument to atomic operation must be a pointer to integer or pointer ('atomic_ulong *' (aka '_Atomic(unsigned long) *') invalid)
      g_eenter_cnt       += __atomic_exchange_n(&tcb->eenter_cnt, 0, __ATOMIC_RELAXED);
```
</blockquote></details>

`atomic_ulong` is `_Atomic unsigned long`, not `_Atomic uint64_t`

kailun-qin (commented):
> Reviewed 6 of 7 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1996)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

___
*[`pal/src/host/linux-sgx/host_thread.c` line 73 at r4](https://reviewable.io/reviews/gramineproject/gramine/1996#-O7_QBnm59NdvrcQRcgk:-O7a7dcY9dGCWBZZ74w3:b6ursmc) ([raw file](https://github.com/gramineproject/gramine/blob/6b5a59f4137403de942190ea291dd4872cc6b39b/pal/src/host/linux-sgx/host_thread.c#L73)):*
> Not happy about the explicit cast

+1 (UB from standard), but it looks to be the easiest way

> `atomic_ulong` is `_Atomic unsigned long`, not `_Atomic uint64_t`

+1

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1996)* status: 13 of 14 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @kailun-qin and @mkow)

___
*[`pal/src/host/linux-sgx/host_thread.c` line 73 at r4](https://reviewable.io/reviews/gramineproject/gramine/1996#-O7_QBnm59NdvrcQRcgk:-O7c15i7CearZtZ1wk5g:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/6b5a59f4137403de942190ea291dd4872cc6b39b/pal/src/host/linux-sgx/host_thread.c#L73)):*

> Not happy about the explicit cast

+1 (UB from standard), but it looks to be the easiest way

> `atomic_ulong` is `_Atomic unsigned long`, not `_Atomic uint64_t`

+1
</blockquote></details>

Done.

dimakuv:
Jenkins, retest Jenkins-20.04 please (unrelated issue, see description below)

On Jenkins-20.04, `gramine-direct ppoll01` LTP test failed:
```
/home/jenkins/workspace/graphene-20.04/libos/test/ltp/src/testcases/kernel/syscalls/ppoll/ppoll01.c:288: TFAIL: ret: 0, exp: -1, ret_errno: SUCCESS (0), exp_errno: EINTR (4)
```

I can't reproduce it locally, and I have never seen this error before. Moreover, this error has nothing to do with the current PR, as the current PR does *not* modify LTP or `gramine-direct` (aka Linux PAL) in any way.

kailun-qin (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1996)* status: all files reviewed, 1 unresolved discussion, not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1996)* status: all files reviewed, 1 unresolved discussion, not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @kailun-qin)

___
*[`pal/src/host/linux-sgx/host_thread.c` line 73 at r4](https://reviewable.io/reviews/gramineproject/gramine/1996#-O7_QBnm59NdvrcQRcgk:-O7eBI7i2STS3-K40vqP:bm5qz6e) ([raw file](https://github.com/gramineproject/gramine/blob/6b5a59f4137403de942190ea291dd4872cc6b39b/pal/src/host/linux-sgx/host_thread.c#L73)):*

Done.
</blockquote></details>

Ah, wait, you should just use C11 functions for that instead: https://en.cppreference.com/w/c/atomic. Then no cast will be needed.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1996)* status: 13 of 14 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @kailun-qin and @mkow)

___
*[`pal/src/host/linux-sgx/host_thread.c` line 73 at r4](https://reviewable.io/reviews/gramineproject/gramine/1996#-O7_QBnm59NdvrcQRcgk:-O7er8H88uvMONx9-BV4:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/6b5a59f4137403de942190ea291dd4872cc6b39b/pal/src/host/linux-sgx/host_thread.c#L73)):*

Ah, wait, you should just use C11 functions for that instead: https://en.cppreference.com/w/c/atomic. Then no cast will be needed.
</blockquote></details>

Done.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1996)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

kailun-qin (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1996)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #2012, opened 2024-09-30
[LibOS] Exclude `/etc/resolv.conf` from allowed/trusted files

### LINKED ISSUE #2014: GSC build with latest master fails on `/etc/resolv.conf` mountpoint when `sys.enable_extra_runtime_domain_names_conf = true`

The commit [aef087f ](https://github.com/gramineproject/gramine/commit/aef087f9c844fbab176333cb76f9fc5d6e04c039) moved the handling of allowed and trusted files into the LibOS layer. This change introduced a bug related to the handling of the /etc/resolv.conf file when the sys.enable_extra_runtime_domain_names_conf option is set to true.

**Details:**
When sys.enable_extra_runtime_domain_names_conf is set to true, Gramine generates /etc/resolv.conf from scratch, ignoring whether the file was marked as allowed or trusted.
If the file is marked as allowed, it is found on the host but ignored and replaced with a generated version.
If the file is marked as trusted and its hash does not match the one specified in the manifest, Gramine's lookup logic detects the mismatch and fails on startup.

For more details, refer https://github.com/gramineproject/gramine/pull/2012

### Steps to reproduce

Build gsc workload with latest gramine, after adding `sys.enable_extra_runtime_domain_names_conf = true` into manifest file

### Expected results

Workload should be running successfully

### Actual results

/bin/bash: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)
bash: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)
[P1:T1:] error: Looking up mountpoint /etc/resolv.conf failed: Operation not permitted (EPERM)
[P1:T1:] error: libos_init() failed in init_mount: Operation not permitted (EPERM)

### Gramine commit hash

aef087f

dimakuv:
## Description of the changes 

Commit aef087f "[LibOS] Move trusted and allowed files logic to LibOS" moved all handling of allowed/trusted files into the LibOS layer. This led to a subtle bug: if `sys.enable_extra_runtime_domain_names_conf` is set to true, Gramine will generate `/etc/resolv.conf` from scratch, ignoring whether the file was marked as allowed or trusted. At the same time, Gramine still needs to look up this file, and as part of this look up Gramine will check whether it was marked as allowed or trusted.

The bug is benign if the file was not marked as allowed or trusted at all: Gramine's lookup logic will return a perfectly valid ENOENT code. The bug is also benign if the file was marked as allowed: the file will be found on the host but ignored and immediately replaced with a generated version. The bug manifests itself, however, if the file was marked as trusted and its hash does not correspond to the one specified in the manifest. In such case, Gramine's lookup logic detects that the file is trusted and that its hashes do not match, and Gramine fails on startup.

This bug is subtle because it needs three things to happen: (1) `/etc/resolv.conf` ends up in `sgx.trusted_files`, (2) the hashes in the host environment and in the manifest do not match, (3) the manifest contains `sys.enable_extra_runtime_domain_names_conf = true`. The first prerequisite may happen if the user puts a blanket `/etc/` dir in the manifest. The second prerequisite may happen e.g. in Docker containers where `/etc/resolv.conf` is re-generated by Docker every time.

This commit adds the corner case of excluding `/etc/resolv.conf` from allowed/trusted file checks, if the manifest contains `sys.enable_extra_runtime_domain_names_conf = true`. This silent exclusion is in line with the current behavior of this manifest option.

Code references:
- https://github.com/gramineproject/gramine/blob/988e6b8c82c5e317758f16034199eb13fdc9c2b3/Documentation/manifest-syntax.rst#domain-names-configuration
- https://github.com/gramineproject/gramine/blob/988e6b8c82c5e317758f16034199eb13fdc9c2b3/libos/src/fs/etc/fs.c#L135-L149
- https://github.com/gramineproject/gramine/blob/988e6b8c82c5e317758f16034199eb13fdc9c2b3/libos/src/fs/libos_fs.c#L543-L546
- https://github.com/gramineproject/gramine/blob/988e6b8c82c5e317758f16034199eb13fdc9c2b3/libos/src/fs/chroot/fs.c#L338
- https://github.com/gramineproject/gramine/blob/988e6b8c82c5e317758f16034199eb13fdc9c2b3/libos/src/fs/chroot/fs.c#L115

Reported by @anjalirai-intel and @jinengandhi-intel on Curated Apps/GSC. Kudos to them.

Fixes #2014 

## How to test this PR? 

~~Not trivial to reproduce.~~ Updated one LibOS regression test to check this corner case (without this PR, the test fails on startup).

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/2012)


mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/2012)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[`libos/src/fs/chroot/fs.c` line 116 at r1](https://reviewable.io/reviews/gramineproject/gramine/2012#-O83NXSTCi79XChxVo6J:-O83NXSTCi79XChxVo6K:b-n9f690) ([raw file](https://github.com/gramineproject/gramine/blob/2408a67b6ecbe7ea602bd03b62ed4c9a0ff3f4d6/libos/src/fs/chroot/fs.c#L116)):*
> ```C
> }
> 
> static bool is_normalized_path_equal_to_etc_resolv_conf(const char* path) {
> ```

That name sounds like the argument is supposed to already be normalized.



_Code quote:_
```C
is_normalized_path_equal_to_etc_resolv_conf
```
___
*[`libos/src/fs/chroot/fs.c` line 128 at r1](https://reviewable.io/reviews/gramineproject/gramine/2012#-O83VDub0sjojNumpgiC:-O83VDub0sjojNumpgiD:bthpt1k) ([raw file](https://github.com/gramineproject/gramine/blob/2408a67b6ecbe7ea602bd03b62ed4c9a0ff3f4d6/libos/src/fs/chroot/fs.c#L128)):*
> ```C
>     }
> 
>     bool result = strcmp(norm_path, "/etc/resolv.conf") == 0;
> ```

You can shorten (and also make it more explicit and readable) by moving this path to an argument - the call would be `normalize_and_compare_path(path, "/etc/resolv.conf")`



_Code quote:_
```C
"/etc/resolv.conf"
```

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/2012)* status: 1 of 2 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[`libos/src/fs/chroot/fs.c` line 116 at r1](https://reviewable.io/reviews/gramineproject/gramine/2012#-O83NXSTCi79XChxVo6J:-O863Xd_9BWgDpV9oBR2:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/2408a67b6ecbe7ea602bd03b62ed4c9a0ff3f4d6/libos/src/fs/chroot/fs.c#L116)):*

That name sounds like the argument is supposed to already be normalized.
</blockquote></details>

Done.
___
*[`libos/src/fs/chroot/fs.c` line 128 at r1](https://reviewable.io/reviews/gramineproject/gramine/2012#-O83VDub0sjojNumpgiC:-O863Wi81wFMI5KPdg4o:b-896fix) ([raw file](https://github.com/gramineproject/gramine/blob/2408a67b6ecbe7ea602bd03b62ed4c9a0ff3f4d6/libos/src/fs/chroot/fs.c#L128)):*

You can shorten (and also make it more explicit and readable) by moving this path to an argument - the call would be `normalize_and_compare_path(path, "/etc/resolv.conf")`
</blockquote></details>

Done.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/2012)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

kailun-qin (approved):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/2012)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

PR #1991, opened 2024-09-03
Add log_trace for raw syscalls

adarshan-intel:
## Description of the changes 

- Logging every raw syscall helps identify how often they occur. If it's too many, then that might cause performance issues, but if it's rare, we can ignore the performance angle
- Printing messages for every syscall captures information about multiple raw syscalls rather than just the first one.
- Printing the syscall address helps to find the library to which the syscall belongs without modifying Gramine code
 

## How to test this PR? 
- To build gramine, use meson and ninja commands. Then, for any workload (eg tested on pytorch), modify the manifest file to change log_level to trace, and you will see the raw syscalls.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/1991)

PR #2015, opened 2024-10-04
[PAL/{Linux,Linux-SGX}] Removing buffer and pal_describe_location function

adarshan-intel:
## Description of the Changes

PR #1991 exposes a segmentation fault with the use of `pal_describe_location` as in below error log:
 
```
(host_exception.c:141:handle_sync_signal) error: Unexpected segmentation fault (SIGSEGV) occurred inside untrusted PAL (libc-2.31.so+0x22941 (addr = 0x7f62e5f98941))
```
 
Using the above addresses we could find the functions abort->hlt from objdump of libc.
 
We tried to debug the issue using GDB to figure out the cause of abort call inside libc but issue did not reproduced with GDB even after multiple tries.
 
Issue is consistent with Pytorch workload using curated apps (with debug build) without GDB, on a Azure VM, rarely on our lab servers.
 
Also tested by increasing the ENCLAVE_SIG_STACK_SIZE and ALT_STACK_SIZE size to 256KB but that also did not help.
 
Likely some race in Gramine when using `pal_describe_location` which we are not hitting with GDB as GDB makes the execution slow.
 
We want to remove this function call in coming release and analyze this issue after the release.

Fixes gramineproject/contrib#87 

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/2015)

### DIFF
--- a/pal/src/host/linux-sgx/pal_exception.c
+++ b/pal/src/host/linux-sgx/pal_exception.c
@@ -1,8 +1,6 @@
 /* SPDX-License-Identifier: LGPL-3.0-or-later */
 /* Copyright (C) 2014 Stony Brook University
  *               2020 Intel Labs
- *               2024 Intel Corporation
- *                    Kailun Qin <kailun.qin@intel.com>
  */
 
 /*
@@ -18,7 +16,6 @@
 #include "pal.h"
 #include "pal_internal.h"
 #include "pal_linux.h"
-#include "pal_sgx.h"
 
 #define ADDR_IN_PAL(addr) ((void*)(addr) > TEXT_START && (void*)(addr) < TEXT_END)
 
@@ -232,9 +229,10 @@ static bool handle_ud(sgx_cpu_context_t* uc, int* out_event_num) {
             log_always("Emulating a raw syscall instruction. This degrades performance, consider"
                        " patching your application to use Gramine syscall API.");
         }
-        char buf[LOCATION_BUF_SIZE];
-        pal_describe_location(uc->rip, buf, sizeof(buf));
-        log_trace("Emulating raw syscall instruction with number %lu at address %s", uc->rax, buf);
+        /* FIXME: better to use `pal_describe_location()` (see example below), but it leads to a
+ *        non-deterministic bug on some workloads */
+        log_trace("Emulating raw syscall instruction with number %lu at address %p",
+                  uc->rax, (void*)uc->rip);
         return false;
     } else if (is_in_out(instr) && !has_lock_prefix(instr)) {
         /*
@@ -260,17 +258,6 @@ static bool handle_ud(sgx_cpu_context_t* uc, int* out_event_num) {
     log_warning("Unknown or illegal instruction executed at %s", buf);
     return false;
 }
-
-/* TODO: remove this once the Linux kernel is patched. */
-#ifndef LINUX_KERNEL_SGX_EDMM_DATA_RACES_PATCHED
-static bool is_eaccept_instr(sgx_cpu_context_t* uc) {
-    /* instruction must be ENCLU and leaf must be EACCEPT */
-    uint8_t* instr = (uint8_t*)uc->rip;
-    if (instr[0] == 0x0f && instr[1] == 0x01 && instr[2] == 0xd7 && uc->rax == EACCEPT)
-        return true;
-    return false;
-}
-#endif
 
 /* perform exception handling inside the enclave */
 void _PalExceptionHandler(uint32_t trusted_exit_info_,
@@ -440,15 +427,8 @@ void _PalExceptionHandler(uint32_t trusted_exit_info_,
         _PalProcessExit(1);
     }
 
-    bool async_event = event_num == PAL_EVENT_QUIT || event_num == PAL_EVENT_INTERRUPTED;
-    bool memfault_with_edmm = !is_synthetic_gp && event_num == PAL_EVENT_MEMFAULT &&
-                              g_pal_linuxsgx_state.edmm_enabled;
-
-    /* in PAL, and event isn't asynchronous (i.e., synchronous exception) or memory fault with EDMM
-     * enabled (which could happen legitimately when some syscalls try to access user buffers
-     * that're created using mappings with `MAP_NORESERVE`) -- this should be handled later in the
-     * lazy-allocation logic */
-    if (ADDR_IN_PAL(uc->rip) && !async_event && !memfault_with_edmm) {
+    /* in PAL, and event isn't asynchronous (i.e., synchronous exception) */
+    if (ADDR_IN_PAL(uc->rip) && event_num != PAL_EVENT_QUIT && event_num != PAL_EVENT_INTERRUPTED) {
         char buf[LOCATION_BUF_SIZE];
         pal_describe_location(uc->rip, buf, sizeof(buf));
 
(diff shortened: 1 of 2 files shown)


mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/2015)* status: 0 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), one-liner is not in imperative mood (waiting on @adarshan-intel)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/2015#-O8Mr-Kq7-tT4ulpQ-u5:-O8Mr-Kq7-tT4ulpQ-u6:b-oi687q) (no related file):*
Why does this fix the issue? What was the reason for the segfault?

___

jkr0103 (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/2015)* status: 0 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), one-liner is not in imperative mood (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/2015#-O8Mr-Kq7-tT4ulpQ-u5:-O8Q3vCE7L_CKxk4os3L:b489mvc) (no related file):*

Why does this fix the issue? What was the reason for the segfault?
</blockquote></details>

Below is the error log with SIGSEGV:
```
(host_exception.c:141:handle_sync_signal) error: Unexpected segmentation fault (SIGSEGV) occurred inside untrusted PAL (libc-2.31.so+0x22941 (addr = 0x7f62e5f98941))
```

Using the above addresses we could get the functions `abort->hlt` from objdump of libc.

We tried to debug the issue using GDB to figure out the cause of `abort` call inside `libc` but issue is not reproducing with GDB, tried multiple times but always reproduces without GDB, seems like some `race` in Gramine when using `pal_describe_location` only with specific configuration.

Hence we plan to remove this function call in coming release and analyze this issue after the release.

___

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/2015)* status: 0 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), one-liner is not in imperative mood (waiting on @jkr0103)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/2015#-O8Mr-Kq7-tT4ulpQ-u5:-O8R1rLcCeNZ3OZUtjKa:be24qat) (no related file):*
<details><summary><i>Previously, jkr0103 (Jitender Kumar) wrote…</i></summary><blockquote>

Below is the error log with SIGSEGV:
```
(host_exception.c:141:handle_sync_signal) error: Unexpected segmentation fault (SIGSEGV) occurred inside untrusted PAL (libc-2.31.so+0x22941 (addr = 0x7f62e5f98941))
```

Using the above addresses we could get the functions `abort->hlt` from objdump of libc.

We tried to debug the issue using GDB to figure out the cause of `abort` call inside `libc` but issue is not reproducing with GDB, tried multiple times but always reproduces without GDB, seems like some `race` in Gramine when using `pal_describe_location` only with specific configuration.

Hence we plan to remove this function call in coming release and analyze this issue after the release.
</blockquote></details>

Please debug this, it may be something serious. This PR sounds like randomly changing code until it stops crashing...

___

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/2015)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), one-liner is not in imperative mood (waiting on @adarshan-intel and @jkr0103)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/gramineproject/gramine/2015#-O8_gCS7Czsqqestq82B):*
This should be like this:
```
[PAL/{Linux,Linux-SGX}] Do not try to describe RIP location on syscall instruction

Commit b6a2d79 ("[PAL/{Linux,Linux-SGX}] Add trace log for raw syscalls")
added a debug print for every encountered raw syscall instruction. Each print
describes what system call number is invoked and at which address in the binary.
The address is fed to the helper function `pal_describe_location(uc->rip)` which
tries to find the binary + function name and put them in human-readable form
into the provided buffer.

For some reason, the snippet used in that commit (allocating a 128-byte buffer
on signal-handling stack and calling `pal_describe_location()`) leads to a
non-deterministic memory corruption on some workloads. That bug is hardly
reproducible, and it is not fixed by e.g. increasing the signal stack size inside
Gramine. The `pal_describe_location()` control path also seems correct.

As the true root cause for this bug is not yet found, this commit introduces
a workaround: temporarily removing the stack-allocated buffer and the
correspoding `pal_describe_location()` call, and instead printing the raw
RIP value.

Signed-off-by: Adarsh Anand <adarsh.anand@intel.com>
```

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/2015)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), one-liner is not in imperative mood (waiting on @adarshan-intel and @mkow)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/2015#-O8Mr-Kq7-tT4ulpQ-u5:-O8_jVIO44JHo04eyllO:b51sg0w) (no related file):*
> Please debug this, it may be something serious. This PR sounds like randomly changing code until it stops crashing...

We tried to debug this, but failed. The issue was (for some reason) reproducible only in a particular Ubuntu 20.04 Docker container on a particular machine.

The issue seems to be a memory corruption, as it is not deterministic. In fact, the message that Jitender and Anjali put here (`
error: Unexpected segmentation fault (SIGSEGV) occurred inside untrusted PAL`) seems to me misleading, as this error is just a manifestation of Glibc's `abort()` function that performs a HLT instruction -- and Gramine is not expecting the untrusted runtime's Glibc to execute such HLT instructions.

Anyway, experimenting with different stack sizes didn't help at all. This was my first guess: a stack buffer overflow, since we allocate a 128-byte buffer on a signal-handling stack. However, this is not the root cause.

Next theory was (and still is) that there is some incorrect locking/synchronization inside `pal_describe_location()`. But inspecting that function, I don't see anything wrong -- all sensitive operations are protected by locks. It does feel though like iterating through a GDB-specific list of loaded binaries has a bug somewhere, but I couldn't pinpoint where. This would explain both requirements for this bug: (1) the bug manifests only in debug build of Gramine, and (2) the bug is a memory corruption that seems to happen in the untrusted PAL.

This can probably be debugged in 2-3 days by a senior dev, but I am personally not taking any such requests anymore. And the time for the new Gramine release is approaching, so my recommendation would be to merge this PR, and properly debug later.

___

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/2015)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), one-liner is not in imperative mood (waiting on @adarshan-intel and @mkow)

___
*[`pal/src/host/linux/pal_exception.c` line 94 at r1](https://reviewable.io/reviews/gramineproject/gramine/2015#-O8_mF7A8mVhZBrrjOVs:-O8_mF7A8mVhZBrrjOVt:b-jshxbm) ([raw file](https://github.com/gramineproject/gramine/blob/b239fbb5258b55db61eff5094db0262c949bb6e0/pal/src/host/linux/pal_exception.c#L94)):*
> ```C
>                        " patching your application to use Gramine syscall API.");
>         }
>         log_trace("Emulating raw syscall instruction with number %d at address %p", info->si_syscall,
> ```

Please add above this line a comment so that we don't forget in the future:
```
/* FIXME: better to use `pal_describe_location()` (see example below), but it leads to a
 *        non-deterministic bug on some workloads */
```
___
*[`pal/src/host/linux-sgx/pal_exception.c` line 235 at r1](https://reviewable.io/reviews/gramineproject/gramine/2015#-O8_mihT59yHIt6PFRgS:-O8_mihT59yHIt6PFRgT:b-4cf746) ([raw file](https://github.com/gramineproject/gramine/blob/b239fbb5258b55db61eff5094db0262c949bb6e0/pal/src/host/linux-sgx/pal_exception.c#L235)):*
> ```C
>                        " patching your application to use Gramine syscall API.");
>         }
>         log_trace("Emulating raw syscall instruction with number %lu at address %p",
> ```

ditto (addd comment)

adarshan-intel:
This PR has been moved to #2017

PR #2025, opened 2024-10-14
[PAL/Linux-SGX] AEX-Notify 1/5: Prepare AEX handler to call C code in release mode

dimakuv:
## Description of the changes 

**Part 1 in AEX-Notify series**.

Previously, the AEX handler only called C code (more specific to our use case, the code that works on the untrusted-thread stack) for SGX profiling logic and only when Gramine was built in debug mode.

As a preparation for AEX-Notify support, this commit modifies the AEX handler such that it can call C code when Gramine is built in release mode. Also, the AEX handler logic is not restricted to SGX profiling only, so the `_PROF`/`_prof` suffixes in related function and variable names are removed.

This commit also adds two small prerequisites for AEX-Notify: (1) `eenter_pointer` helper so that signal handling logic can learn whether an exception happened at EENTER instruction, and (2) dummy function `maybe_raise_pending_signal` that will be populated in a later commit.

See also related PRs and discussions:
- #1530 
- #1531 
- #1948 

## How to test this PR? 

CI is enough.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/2025)

PR #2023, opened 2024-10-10
[PAL/Linux] Log a message on `mmap()` returning `-ENOMEM`

kailun-qin:
## Description of the changes 

Linux can return `-ENOMEM` if a request to `mmap()` exceeds the process's maximum number of mappings. After commit c0a2765 "[LibOS,PAL/Linux-SGX] Add EDMM lazy allocation support", this can occur during the checkpoint/restore on the Linux PAL. In such case, each committed page is sent separately and Linux creates a mapping for each one, which can result in a large number of mappings.

This commit makes this failure more visible and provides a hint to the user on what can be done on the host.

Related to https://github.com/gramineproject/examples/issues/107.

## How to test this PR? 

CI.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/2023)

PR #2032, opened 2024-10-15
[PAL/Linux-SGX] Inject signals into enclave in regular context

dimakuv:
## Description of the changes 

Previously, if the enclave was interrupted by a sync signal (e.g., SIGILL) or an async signal (e.g., SIGTERM), then the untrusted-runtime signal handler injected the signal directly into the enclave. In particular, untrusted runtime ran in the signal-handling context and called `sgx_raise()` that would perform EENTER to enter the in-enclave stage-1 signal handler, then EEXIT to exit the enclave back into the untrusted-runtime signal-handling context, and then untrusted runtime would perform sigreturn to go back to untrusted-runtime regular context, jumping to AEP (Asynchronous Exit Pointer). In AEP, ERESUME was called to resume the enclave execution from the stage-2 signal handler.

In other words, the following invariants held:
- In-enclave stage-1 signal handler (in SSA 1) always executed in the signal-handling context of the untrusted runtime.
- In-enclave stage-2 signal handler (in SSA 0) always executed in regular context of the untrusted runtime.

As a preparation for AEX Notify support, this commit breaks the above strong coupling of contexts: in-enclave stage-1 signal handler must execute in regular context of the untrusted runtime.

In particular, this commit changes signal-handling logic as follows: instead of immediately delivering a sync/async signal into the enclave, the untrusted runtime's signal handler memorizes the signal in a thread-local variable `last_sync_signal`/`last_async_signal` and returns. When host kernel returns back to regular context from the signal handler, it jumps to the AEP, which is augmented with a new logic: checking whether there is any signal pending (variables `last_sync_signal` or `last_async_signal` are not zero). If there is a pending signal, the new AEP logic performs EENTER, so that in-enclave stage-1 handler executes. After the stage-1 handler is done, it performs EEXIT, and the AEP logic finalizes with ERESUME as usual. At this point the flow is the same as was previously implemented: the enclave is resumed in the in-enclave stage-2 handler.

There is one corner case: an async signal can arrive while the enclave is executing the stage-1 handler (in SSA 1). In this case, an async signal flow is triggered in untrusted runtime, and the AEP after the async signal will try to EENTER, but since there's already SSA 1 executing inside the enclave and SSA 2 is forbidden by SGX hardware, this (nested) EENTER will raise a #GP fault which translates into SIGSEGV and is delivered to the untrusted runtime's signal handler. We augment the SIGSEGV (aka PAL_EVENT_MEMFAULT) signal handler to catch this particular case and ignore it: the async signal is re-memorized in `last_async_signal` variable but cannot be delivered right now. This async signal will be delivered on some later AEX event.

- #1530 
- #1531 
- #1948 

## How to test this PR? 

CI is enough.

### DIFF
--- a/pal/src/host/linux-sgx/host_exception.c
+++ b/pal/src/host/linux-sgx/host_exception.c
@@ -93,25 +93,76 @@ static bool interrupted_in_aex(void) {
 }
 
 static void handle_sync_signal(int signum, siginfo_t* info, struct ucontext* uc) {
-    enum pal_event event = signal_to_pal_event(signum);
-
     __UNUSED(info);
+
+    enum pal_event event = signal_to_pal_event(signum);
+    uint64_t rip = ucontext_get_ip(uc);
 
     /* send dummy signal to RPC threads so they interrupt blocked syscalls */
     if (g_rpc_queue)
         for (size_t i = 0; i < g_rpc_queue->rpc_threads_cnt; i++)
             DO_SYSCALL(tkill, g_rpc_queue->rpc_threads[i], SIGUSR2);
 
+    if (event == PAL_EVENT_MEMFAULT && interrupted_in_aex() && rip == (uint64_t)&eenter_pointer) {
+        /*
+         * This is a #GP on EENTER instruction inside sgx_raise(), called during AEX handling by
+         * maybe_raise_pending_signal(). This implies that some async signal arrived and was
+         * injected by AEX logic while the enclave thread is being executed in CSSA=1 (stage-1
+         * exception handler).
+         *
+         * We ignore this #GP fault by skipping EENTER. This newly arrived async signal will be
+         * delivered at some later AEX event, when the enclave thread starts executing in CSSA=0.
+         *
+         * Since last_async_event was reset to NO_EVENT before sgx_raise(), we must restore it to
+         * this failed-to-deliver async signal. We extract async signal number from RDI register.
+         * See also maybe_raise_pending_signal().
+         */
+        enum pal_event faulted_event = uc->uc_mcontext.rdi; /* convention, see .Lcssa1_exception */
+        if (faulted_event != PAL_EVENT_INTERRUPTED && faulted_event != PAL_EVENT_QUIT) {
+            log_error("#GP on EENTER instruction not because of async signal, impossible!");
+            BUG();
+        }
+        if (pal_get_host_tcb()->last_async_event != PAL_EVENT_QUIT) {
+            /* Do not overwrite `PAL_EVENT_QUIT`. For explanation, see handle_async_signal(). */
+            pal_get_host_tcb()->last_async_event = faulted_event;
+        }
+
+        ucontext_set_ip(uc, rip + /*sizeof(ENCLU)=*/3); /* skip EENTER */
+        return;
+    }
+
     if (interrupted_in_enclave(uc)) {
-        /* exception happened in app/LibOS/trusted PAL code, handle signal inside enclave */
+        /*
+         * Exception happened in app/LibOS/trusted PAL code, mark this sync signal as pending. This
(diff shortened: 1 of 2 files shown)


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/2032)* status: 0 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/2032#-O9EMRD12nYbafokI6OM:-O9EMRD12nYbafokI6ON:bprxfke) (no related file):*
Must be applied on top of #2025. Blocking.

___

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/2032)* status: 0 of 2 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

___
*[`-- commits` line 21 at r1](https://reviewable.io/reviews/gramineproject/gramine/2032#-O9EWAlg7b4fQWtnmeW6):*
Here and everywhere: the correct spelling is `AEX-Notify`, with a dash. Change.
___
*[`pal/src/host/linux-sgx/host_exception.c` line 141 at r1](https://reviewable.io/reviews/gramineproject/gramine/2032#-O9EWCQx6MFsD7NtUrMY:-O9EWCQx6MFsD7NtUrMZ:bh4b0lq) ([raw file](https://github.com/gramineproject/gramine/blob/bca2d41431cb7a1de7555c00282bf70051d4e3cc/pal/src/host/linux-sgx/host_exception.c#L141)):*
> ```C
>          *
>          * We do not deliver the signal immediately to the enclave (but instead mark it as pending)
>          * because we want to support AEX Notify hardware feature in SGX. In particular, AEX Notify
> ```

Here and everywhere: the correct spelling is `AEX-Notify`, with a dash. Change.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/2032)* status: 0 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

___
*[`-- commits` line 21 at r1](https://reviewable.io/reviews/gramineproject/gramine/2032#-O9EWAlg7b4fQWtnmeW6):*

Here and everywhere: the correct spelling is `AEX-Notify`, with a dash. Change.
</blockquote></details>

Done
___
*[`pal/src/host/linux-sgx/host_exception.c` line 141 at r1](https://reviewable.io/reviews/gramineproject/gramine/2032#-O9EWCQx6MFsD7NtUrMY:-O9n4Nze2Emg14-onHhJ:bb6un1j) ([raw file](https://github.com/gramineproject/gramine/blob/bca2d41431cb7a1de7555c00282bf70051d4e3cc/pal/src/host/linux-sgx/host_exception.c#L141)):*

Here and everywhere: the correct spelling is `AEX-Notify`, with a dash. Change.
</blockquote></details>

Done

dimakuv:
Jenkins, retest Jenkins-Direct-24.04-Sanitizers please (`ppoll01` LTP test failed, unrelated to this PR as this PR only changes `gramine-sgx` and not `gramine-direct`). Also, this LTP failure is like this one: https://github.com/gramineproject/gramine/pull/1996#issuecomment-2373469631

dimakuv:
Jenkins, retest Jenkins-SGX-22.04-EDMM please (`cryptography.exceptions.InternalError: Unknown OpenSSL error`). Not related to this PR, see https://github.com/gramineproject/gramine/pull/2023#issuecomment-2417416345

PR #2034, opened 2024-10-15
[PAL/Linux-SGX] Add AEX-Notify enabling code

dimakuv:
## Description of the changes 

This is a preparatory PR for AEX-Notify support. This PR:
- Introduces the `sgx.experimental_enable_aex_notify` manifest option.
- Adds architectural flags/bits for SECS, TCS, SSA data structures.
- Adds a Gramine startup check whether AEX-Notify hardware feature is supported by the platform.
- Adds dynamic enablement/disablement of AEX-Notify feature per enclave thread (enable on thread creation, disable on thread termination).

## How to test this PR? 

CI is enough to test that old flows are not broken.

Testing new manifest option `sgx.experimental_enable_aex_notify` manually:
- Add `sgx.experimental_enable_aex_notify = true` to HelloWorld CI example
- Test on different platforms

#### Platform with no AEX-Notify hardware feature

**TBD**

#### Platform with AEX-Notify hardware feature but old Linux

(Linux kernel received AEX-Notify support in v6.2. Below I'm testing with Linux v5.15.)

```
gramine/CI-Examples/helloworld$ gramine-sgx helloworld

Gramine is starting. Parsing TOML manifest file, this may take some time...
error: Enclave initialization IOCTL failed: Permission denied (EACCES)
error: Initializing enclave failed: Permission denied (EACCES)
error: load_enclave() failed with error: Permission denied (EACCES)
```

If you wonder why Linux v5.15 enclave init fails with EACCES, check these sources:
- https://elixir.bootlin.com/linux/v5.15/source/arch/x86/kernel/cpu/sgx/ioctl.c#L508
- https://elixir.bootlin.com/linux/v5.15/source/arch/x86/kernel/cpu/sgx/ioctl.c#L113

And compare with Linux v6.2:
- https://elixir.bootlin.com/linux/v6.2/source/arch/x86/kernel/cpu/sgx/ioctl.c#L497
- https://elixir.bootlin.com/linux/v6.2/source/arch/x86/kernel/cpu/sgx/ioctl.c#L114
- https://elixir.bootlin.com/linux/v6.2/source/arch/x86/include/asm/sgx.h#L141

#### Platform with AEX-Notify hardware feature and new Linux

**TBD**

PR #2036, opened 2024-10-16
[PAL/Linux-SGX] AEX-Notify 4/5: Do not clobber RBX reg in stage-1 signal handler

dimakuv:
## Description of the changes 

**Part 4 in AEX-Notify series.**

If the enclave is entered with CSSA=1, then the enclave starts executing the stage-1 signal handler asssembly code. This assembly code prepares the context for the stage-2 signal handler, which will be executed in regular context with CSSA=0 and using C code.

This stage-1 signal handler uses the RBX register as a base pointer to the SSA[0].GPRSGX region. Closer to the end of the stage-1 handler flow, in particular before the `.Lcssa1_exception_eexit` label, this SSA[0] base pointer is not needed anymore, so the RBX register is used for other purposes. In particular, RBX is used to hold the stashed RDX value (where-to-exit address in untrusted runtime).

However, a future commit that will introduce AEX-Notify flows inside the enclave needs to access the SSA[0].GPRSGX region at this stage (in the `.Lcssa1_exception_eexit` label). So the RBX register must not be overwritten. Thus, this preparatory commit does not clobber RBX, but instead stashes RDX into another unused register, R10.

- #1530 
- #1531 
- #1948 

## How to test this PR? 

CI is enough.

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/2036)

### DIFF
--- a/pal/src/host/linux-sgx/enclave_entry.S
+++ b/pal/src/host/linux-sgx/enclave_entry.S
@@ -531,8 +531,8 @@ enclave_entry:
     movq %rdi, SGX_GPR_RIP(%rbx)
 
     # copy the whole SSA[0].XSAVE region to the CPU context's XSAVE on stack;
-    # __restore_xregs / __save_xregs clobber RDX so need to stash it in RBX
-    movq %rdx, %rbx
+    # __restore_xregs / __save_xregs clobber RDX so need to stash it in R10
+    movq %rdx, %r10
     movq %gs:SGX_SSA, %rdi
     leaq 1f(%rip), %r11
     jmp __restore_xregs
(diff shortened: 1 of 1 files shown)


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/gramine/2036)* status: 0 of 18 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

*[a discussion](https://reviewable.io/reviews/gramineproject/gramine/2036#-O9n7jBa4blRMTMF3HBG:-O9n7jBa4blRMTMF3HBH:bise3hs) (no related file):*
Must be applied on top of #2034. Blocking.

___

PR #2088, opened 2025-01-28
AlmaLinux 10

woju:
## Description of the changes 

RHEL 10 is expected around end of January 2025. 10-Beta was released in December 2024, so let's try to build packages for that.

Do not merge before AlmaLinux 10 final release (not only RHEL 10).

## How to test this PR? 

CI. After we finalise this PR, we need to drop the commit that disables other pipelines.


woju:
Jenkins, test Jenkins-Direct-22.04 please

PR #2161, opened 2025-10-14
Allow for tcb migration of sealed files

### LINKED ISSUE #855: sgx_get_seal_key() and protected filesystem are not backwards compatible
## Description of the problem

SGX key derivation gets SVN fields in the request (ISVSVN, CPUSVN and CONFIGSVN). This mechanism provides backwards compatibility and anti-rollback protection as follows:
1. Each combination of SVNs generates a different key
2. If all SVNs are less or equal the values of the currently running enclave, a key is returned
3. If at least one SVN value is greater than the currently running enclave, key derivation fails

Current implementation of sgx_get_seal_key() uses the current enclave's SVNs (ISV, CPU and CONFIG) in a hardcoded manner. Consequently, Gramine does not provide backwards compatibility mechanism: if the enclave or the platform were updated, previously sealed blob will no longer be accessible.

This is problematic not only for user code, but could also break backwards compatibility of protected filesystem: if a file was written with one combinations of SVNs, it will no longer be accessible after they were changed (scenario and expected vs. actual results are below).

I think the following changes will be required:
1. Protected file format: the file's header will have to include either the entire key request or selected parameters (SVNs are required, but other fields may be needed as well)
2. Adding some more parameters to sgx_get_seal_key() may be needed (either all or some of the key request parameters). PalGetSpecialKey() may have to be modified as well, but I'm not sure if and how
3. What should happen if data is written into a file that was protected with older SVNs? Writing with the old SVNs is a security concern, because in that case potentially vulnerable enclave/platform may disclose the new data. Re-encrypting the entire file may be a significant performance impact, but it happens only once per SVN upgrade

## Steps to reproduce

Setup:
1. Create two enclaves: enclave _A_ with `ISVSVN=1` and enclave _B_ with `ISVSVN=2`, such that both enclave access the same protected files and both are signed with the same key
5. In both enclaves, mount the protected filesystem with MRSIGNER-based key
6. Write file X with enclave _A_
7. Write file Y with enclave _B_

Test flow:
1. Enclave _A_: read file X
4. Enclave _A_: read file Y
5. Enclave _B_: Read file X
6. Enclave _B_: Read file Y

## Expected results

1. Success
3. Fail
4. Success
5. Success

## Actual results

1. Success
2. Fail
3. **Fail**
4. Success

njeans:
## Description of the changes 
- This does automatic migration of files that were sealed with an different CPU SVN. The current CPU SVN of the platform is saved in the mounted directory in `gramine.tcb_info` file. On startup if the current CPU SVN is different the files mounted in the directory are re-encrypted with the sealing key of the current CPU SVN. This is basically the design proposed in the issue by @szymek156 https://github.com/gramineproject/gramine/issues/855#issuecomment-2404948161. Did not include MRSigner based keys from the issue discussion it seems like migration can be done by signing a new enclave with the MRSigner key.
- Also adds interface to get the current CPU SVN from the `/dev/attestation/cpu_svn` file and get the cooresponding key with `/dev/attestation/keys/svn/_sgx_mrenclave/<cpu_svn>`
- I created a test by allowing the debug enclave to set the cpu svn by updating `/dev/attestation/cpu_svn` file. On the first run the enclave sets an older cpu svn and seals some files. The `gramine.tcb_info` file it set to the old svn. On the second run the files should have been automatically migrated. The enclave checks if the file contents match.

Fixes #855 

## How to test this PR? 
```
cd /gramine/libos/test/regression
gramine-test --sgx pytest -v -k test_230_keys
gramine-test --sgx pytest -v -k test_054_sealed_file_tcb_migration
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/gramine/2161)

### DIFF
--- a/Documentation/devel/encfiles.rst
+++ b/Documentation/devel/encfiles.rst
@@ -542,3 +542,14 @@ Additional details
     file data. Therefore, the usual NIST limits on the total number of
     invocations of the encryption operation with the same key would be reached
     much slower.
+
+- TCB migration: Gramine supports migration between different CPU SVN versions
+  by enabling it using the ``allow_tcb_migration`` mount parameter in the 
+  Gramine manifest. This feature allows encrypted files to be accessed even when 
+  the CPU SVN has changed after applying microcode updates. The current CPU SVN 
+  is stored in the mount directory in a file named ``gramine.tcb_info``. On 
+  startup if the CPU SVN has changed, Gramine will unseal the files using the 
+  old CPU SVN and reseal them with the current CPU SVN. This feature can negatively
+  impact integrity of files, as it allows files from a platform with a lower 
+  security level to be used. It is recommended to use this feature only in controlled
+  environments where the security implications are understood.
--- a/Documentation/devel/features.md
+++ b/Documentation/devel/features.md
@@ -1314,6 +1314,8 @@ grows with time, as Gramine adds functionality required by real-world workloads.
       - ☑ `/dev/attestation/keys/<key_name>` <sup>[25](#attestation)</sup>
       - ☑ `/dev/attestation/keys/_sgx_mrenclave` <sup>[25](#attestation)</sup>
       - ☑ `/dev/attestation/keys/_sgx_mrsigner` <sup>[25](#attestation)</sup>
+      - ☑ `/dev/attestation/keys/svn/_sgx_mrenclave/<cpu_svn>` <sup>[25](#attestation)</sup>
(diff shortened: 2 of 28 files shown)


monavij:
Thanks a lot for contributing this PR for supporting an important feature in gramine. Before we review this, it will be great if you can come present high level design of this feature in our community meeting (Tuesdays at 7am pacific time)

njeans:
Hi yes I can do that

njeans:
How do I join the meeting?

